Issue #1321, opened 2017-10-02
In --catchup-complete mode no progress is written to output when applying ledgers


vogel:
When running stellar-core in --catchup-complete mode progress is nicely displayed when downloading and verifying ledgers, downloading and applying buckets and downloading ledgers. There is nothing displayed when applying ledgers. It is the longest operation of them all and displaying nothing can make an impression that nothing is happening.

That needs to be fixed.

PR #1322, opened 2017-10-02
print log on apply buckets and apply ledgers in commandline mode

vogel:
Force displaying log messages from CatchupManager when a bulk of buckets
or ledgers are applied.

This ensures that even in commandline mode updates are printed on
screen, when there is no 'consensus received' event that calls
logAndUpdateCatchupStatus in LedgerManagerImpl::valueExternalized.

Fixes: #1321


MonsieurNicolas:
r+ 3644e625356f5bb196e04dc1ced12702e9680d7c

PR #1324, opened 2017-10-02
Ledger entries validations

MonsieurNicolas:
This PR makes a few minor adjustments to validations performed when saving and generating ledger entries.

I found those when using a better fuzzer to generate test data.

I do not think that this has any impact on the product as validations are performed by various operations - so this is more of a cosmetic change that only impacts test code (but that can help fix bad test generators)

### DIFF
--- a/src/ledger/AccountFrame.cpp
+++ b/src/ledger/AccountFrame.cpp
@@ -103,10 +103,23 @@ AccountFrame::normalize()
 bool
 AccountFrame::isValid()
 {
+    bool res = (mEntry.lastModifiedLedgerSeq <= INT32_MAX);
     auto const& a = mAccountEntry;
-    return isString32Valid(a.homeDomain) && a.balance >= 0 &&
-           std::is_sorted(a.signers.begin(), a.signers.end(),
-                          &AccountFrame::signerCompare);
+
+    res = res && isString32Valid(a.homeDomain);
+    res = res && a.balance >= 0;
+    res = res && (a.seqNum <= INT64_MAX);
+    res = res && (std::adjacent_find(a.signers.begin(), a.signers.end(),
+                                     [](Signer const& s1, Signer const& s2) {
+                                         return !signerCompare(s1, s2);
+                                     }) == a.signers.end());
+    res = res && ((a.flags & ~MASK_ACCOUNT_FLAGS) == 0);
+    res = res && (a.numSubEntries <= INT32_MAX);
+    res = res &&
+          std::all_of(a.signers.begin(), a.signers.end(), [](Signer const& s) {
+              return (s.weight <= UINT8_MAX) && (s.weight != 0);
+          });
+    return res;
 }
 
 bool
--- a/src/ledger/LedgerTestUtils.cpp
+++ b/src/ledger/LedgerTestUtils.cpp
@@ -35,22 +35,24 @@ clampHigh(T high, T& v)
     }
 }
 
+// mutate string such that it doesn't contain control characters
+// and is at least minSize characters long
 template <typename T>
 void
-stripControlCharacters(T& s)
+replaceControlCharacters(T& s, int minSize)
 {
     std::locale loc("C");
-
-    for (auto it = s.begin(); it != s.end();)
+    if (s.size() < minSize)
+    {
+        s.resize(minSize);
+    }
+    for (auto it = s.begin(); it != s.end(); it++)
     {
         char c = static_cast<char>(*it);
         if (c < 0 || std::iscntrl(c))
         {
-            it = s.erase(it);
-        }
-        else
-        {
-            it++;
+            auto b = autocheck::generator<char>{}(autocheck::detail::nalnums);
+            *it = b;
         }
     }
 }
@@ -69,8 +71,12 @@ makeValid(AccountEntry& a)
         a.balance = -a.balance;
     }
 
-    stripControlCharacters(a.homeDomain);
-    a.inflationDest.activate() = PubKeyUtils::random();
+    replaceControlCharacters(a.homeDomain, 0);
+
+    if (a.inflationDest)
+    {
+        *a.inflationDest = PubKeyUtils::random();
+    }
 
     std::sort(a.signers.begin(), a.signers.end(), &AccountFrame::signerCompare);
     a.signers.erase(
@@ -78,12 +84,15 @@ makeValid(AccountEntry& a)
         a.signers.end());
     for (auto& s : a.signers)
     {
+        s.weight = s.weight & UINT8_MAX;
         if (s.weight == 0)
         {
             s.weight = 100;
         }
     }
     a.numSubEntries = (uint32)a.signers.size();
+    a.seqNum = a.seqNum & INT64_MAX;
+    a.flags = a.flags & MASK_ACCOUNT_FLAGS;
 }
 
 void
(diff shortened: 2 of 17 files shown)


src/ledger/AccountFrame.cpp
```diff
@@ -104,9 +104,12 @@ bool
...
     auto const& a = mAccountEntry;
-    return isString32Valid(a.homeDomain) && a.balance >= 0 &&
-           std::is_sorted(a.signers.begin(), a.signers.end(),
-                          &AccountFrame::signerCompare);
+    bool res = isString32Valid(a.homeDomain);
+    res = res && a.balance >= 0;
+    res = res && std::is_sorted(a.signers.begin(), a.signers.end(),
+                                &AccountFrame::signerCompare);
+    res = res && ((a.flags & ~MASK_ACCOUNT_FLAGS) == 0);
+    return res;
```
jonjove: This does not implement all of the checks that are applied in makeValid, including:
- signers must be unique
- weights of signers cannot be 0

src/ledger/LedgerTestUtils.cpp
```diff
@@ -102,26 +104,34 @@ makeValid(TrustLineEntry& tl)
...
     o.buying.type(ASSET_TYPE_CREDIT_ALPHANUM4);
     strToAssetCode(o.buying.alphaNum4().assetCode, "EUR");
 
-    clampLow<int64_t>(0, o.amount);
-    clampLow(0, o.price.n);
+    if (o.amount < 0)
+    {
+        o.amount = -o.amount;
+    }
+    clampLow(1, o.price.n);
```
jonjove: Using clampLow instead of absolute value reduces the entropy of the overall price, giving a significant probability that price = price.n / price.d == 1.

src/ledger/LedgerTestUtils.cpp
```diff
@@ -102,26 +104,34 @@ makeValid(TrustLineEntry& tl)
     clampLow<int64_t>(0, tl.balance);
```
jonjove: clampLow for balance and limit are redundant considering that balance and limit are already converted to absolute value. This PR is an opportunity to fix this

src/ledger/OfferFrame.cpp
```diff
@@ -120,7 +120,16 @@ OfferFrame::getFlags() const
 bool
 OfferFrame::isValid(OfferEntry const& oe)
 {
-    return isAssetValid(oe.buying) && isAssetValid(oe.selling);
+    bool res = isAssetValid(oe.buying);
+    // to avoid SQL problems (not a protocol limitation)
+    res = res && (oe.offerID <= INT64_MAX);
```
jonjove: If this isn't a protocol limitation, should we actually be checking this here? We aren't performing analogous checks in other *Frame::isValid functions for values that are BIGINT.

src/ledger/OfferFrame.cpp
```diff
@@ -120,7 +120,16 @@ OfferFrame::getFlags() const
...
+
+    res = res && (oe.amount > 0);
+    res = res && (oe.price.n > 0);
+    res = res && (oe.price.d >= 1);
+    res = res && isAssetValid(oe.selling);
```
jonjove: It might be neater (and less brainpower to maintain in the future) to keep both isAssetValid checks together, since they are analogous. This is purely a question of style though

src/ledger/AccountFrame.cpp
MonsieurNicolas: * is_sorted does ensure that signers are unique (strict comparison)
* weight - yep, I am adding this check

src/ledger/LedgerTestUtils.cpp
MonsieurNicolas: fixed

src/ledger/LedgerTestUtils.cpp
MonsieurNicolas: fixed

src/ledger/OfferFrame.cpp
MonsieurNicolas: Yes we should check - I've removed the comment that is confusing:
At the protocol level, we want those to be unsigned - yet most current implementations won't be able to deal with values that are bigger than max int64; this is not a problem for now as the sun should have exploded by the time we reach those type of numbers.

I've done an extra pass on the isValid functions to address this widely.

src/ledger/OfferFrame.cpp
MonsieurNicolas: fixed

src/ledger/AccountFrame.cpp
```diff
@@ -103,12 +103,19 @@ AccountFrame::normalize()
 bool
 AccountFrame::isValid()
 {
+    bool res = (mEntry.lastModifiedLedgerSeq <= INT32_MAX);
```
jonjove: Why not check that the account seqNum is less than INT64_MAX?

MonsieurNicolas: this is ledger sequence number, a 32 bit number

src/ledger/OfferFrame.cpp
```diff
@@ -118,24 +118,25 @@ OfferFrame::getFlags() const
 }
 
 bool
-OfferFrame::isValid(OfferEntry const& oe)
+OfferFrame::isValid(LedgerEntry const& le)
```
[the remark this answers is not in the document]
MonsieurNicolas: this is an internal method - I see that for `OfferFrame` it's in the public namespace, I am moving it to private

src/ledger/AccountFrame.cpp
MonsieurNicolas: yikes, you are correct, every word is important `non-descending` means that it's using `!(a < b)` internally

src/ledger/AccountFrame.cpp
jonjove: Sorry this comment was pretty unclear in retrospect. What I meant was, why not also check that AccountEntry::seqNum (of type SequenceNumber, which is an alias for uint64) is less than INT64_MAX? The check as your wrote it for LedgerEntry::lastModifiedLedgerSeq is indeed correct.

src/ledger/OfferFrame.cpp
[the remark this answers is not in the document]
jonjove: I see that it is also public in `TrustFrame`

src/ledger/LedgerTestUtils.cpp
```diff
@@ -70,20 +70,27 @@ makeValid(AccountEntry& a)
...
     for (auto& s : a.signers)
     {
+        s.weight = s.weight & UINT8_MAX;
         if (s.weight == 0)
         {
             s.weight = 100;
         }
     }
     a.numSubEntries = (uint32)a.signers.size();
+    a.seqNum = a.seqNum & INT64_MAX;
```
jonjove: Does the C++ standard guarantee that it is well defined to do (uint64_t) & (int64_t)? There are several places in this PR where such operations occur.

MonsieurNicolas: yes, it's just a bitwise operation

src/ledger/OfferFrame.cpp
```diff
@@ -246,7 +256,7 @@ OfferFrame::loadOffers(StatementContext& prep,
             }
         }
 
-        if (!isValid(oe))
+        if (!isValid(le))
         {
             throw std::runtime_error("Invalid asset");
```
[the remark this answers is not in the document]
MonsieurNicolas: nope, I think it was probably vestigial that we used to only check for assets in this part of the code

src/ledger/LedgerTestUtils.cpp
jonjove: But my understanding is that bitwise operations act on the representation of the data rather than the interpretation, and the C++ standard admits multiple representations of signed integers (although I am under the impression that almost all modern implementations are two's complement). However, I do not know if it is always guaranteed that a positive signed integer will have the same representation as an unsigned integer of the same number of bytes.

MonsieurNicolas: positive numbers yes (they don't say "positive", rather "non-negative"):
per standard 3.9.1
 > The range of non-negative values of a signed integer type is a subrange of the corresponding unsigned integer type, and the value representation of each corresponding signed/unsigned type shall be the same.

jonjove:
Looks good to me now. Can you squash/rebase and then I'll trigger the test and merge bot?

jonjove:
r+ ecdb4ab

MonsieurNicolas:
r+ ecdb4ab06d7acbeaf3b59b34dfe328ac5b70a984

MonsieurNicolas:
@latobarita : retry

MonsieurNicolas:
r+ 57940769a438ef3a81d6f0f648ae9b383c315f2c

MonsieurNicolas:
r+ 94ace775914b6b3fe7bfb23ec3f679fd71c16809

MonsieurNicolas:
r+ 32fe4f8

MonsieurNicolas:
r+ 19de9c258e60aabc9121d4c28443ac16fdbb4dcd

MonsieurNicolas:
(I rebased again)

MonsieurNicolas:
looks like it hit the flaky test that is on master right now, I made a change to the dot reporter so that we get some context on what is going on.

MonsieurNicolas:
r+ 8c02e21f7c769cadb70171c585a1f058ed8ccc89

PR #1328, opened 2017-10-03
re-introduce assert on double delete (ledger delta)

MonsieurNicolas:
this was removed by accident in https://github.com/stellar/stellar-core/commit/8dc10bf12e537365ad6e85d148d288842548f94b

a double delete in a ledger delta is the definite manifestation of buggy code upstream, in this case this commit would just hide the bug fixed in https://github.com/stellar/stellar-core/commit/ab8dca2e9570fb6d73fb03f7708f8b39b9a6daff


jedmccaleb:
r+ 50e6169

PR #1335, opened 2017-10-13
Database cache metrics

vogel:
It lets us check how database cache is performing. It is doing much better that I've expected.


MonsieurNicolas:
leaving this out for now (we have big changes coming up in this area, and we first need higher level metrics) - we have already too many metrics that we don't really use.

Issue #1332, opened 2017-10-06
Check if quick restart of node does not require catchup


vogel:
Our code should support rejoining network without catchup if only few ledgers have been closed between stop and start of a node.

Check if that works.

vogel:
It works only when number of missing nodes is 2 or less. That leaves 10 seconds for restart with connection to at least one peer that has messages signed by our quorum.

MonsieurNicolas:
Oh I see.
I was thinking we could send more by default, but maybe we can do something a little better:
when we see that we have gaps between our LCL and the SCP messages, we could request peers for LCL+1 and it should allow to quickly close the gap if peers have the data.

What do you think?

vogel:
I think we should just send all data we have, without limit - we only have MAX_SLOTS_TO_REMEMBER last slots anyway. And its value is only 4. We could ncrease this o 12 - that would give full 60 seconds of time for restart and reconnection.

vogel:
I also think that if we had LCL field in hello message we could only send needed envelopes and not all of them...

MonsieurNicolas:
> I think we should just send all data we have, without limit - we only have MAX_SLOTS_TO_REMEMBER last slots anyway. And its value is only 4. We could ncrease this o 12 - that would give full 60 seconds of time for restart and reconnection.

sounds good

> I also think that if we had LCL field in hello message we could only send needed envelopes and not all of them...
changing the HELLO message is a bit of a pain from an upgrade point of view, but I agree would be better. Let's do the simpler version for now.

MonsieurNicolas:
Scratch that, let's do something closer to the HELLO change by triggering a similar exchange at the business logic layer: we should increase the number of ledgers in memory but we should not send out all of this data every time we connect (not scalable). I think that a node just needs to have immediate information on what its newly connected peer has and make a decision in order to ask for specific ledgers. That way we move to something closer to "only send large data if I was asked to".

Right now we have one side of the equation `Peer::sendGetScpState(uint32 ledgerSeq)` , we just need to give the client information so that it can request what it needs.

MonsieurNicolas:
was fixed in #1336

PR #1336, opened 2017-10-13
Test quick restart

vogel:
It seems that stellar-core can survive restart/disconnection only if the next connection comes withing 2 ledgers (about 10 seconds).

It is controlled by HerderImpl::sendSCPStateToPeer::nbLedgers constant (it is 1 more than number of missing ledgers).

Tests for: #1332


vogel:
I think we should also limit number of GET_SCP_STATE request that we allow from one peer to one (just ignore next ones).
If not, this can be a surface for a DOS.

MonsieurNicolas:
mechanism for dealing with this type of DOS is currently to use trusted peers (that can downgraded to regular peers), so I think it's fine.

MonsieurNicolas:
r+ 4edc16f21e21cd2492a75f4c82168cc886a1936c

vogel:
Rebased on current master, so it will compile

PR #1344, opened 2017-10-19
Refactor herder

vogel:
2 classes extracted from Herder/HerderImpl.

This is work in progress.


MonsieurNicolas:
cool, can you add intermediate commits with just the code moves? (they won't compile, but that's fine) Right now it's really hard to tell what actually changed without pulling a side by side diff of the old vs new code.

MonsieurNicolas:
r+ 43fc77fc487f3e0996dfd9848c50cac2ab6723af

PR #1345, opened 2017-10-19
ensure all needed buckets are retained at shutdown

### LINKED ISSUE #1342: catchup-at doesn't always leave the database in a consistent state
# Issue type
bug

# Version found
https://github.com/stellar/stellar-core/commit/d625b738ae3543347d2f5684155c707192c000e9

# Your environment and setup
running from the command line

# Issue description

do the following (against current public network):
1. run `stellar-core.exe --conf prod.cfg --catchup-at 13980579`
2. run `stellar-core.exe --conf prod.cfg --catchup-at 13980581`

## Actual result
step 1 claims that it succeded but step 2 triggered the "missing buckets" work, which leads me to think that step 1 didn't leave the database in a good state.

```
2017-10-18T13:57:55.834 GANOG [default INFO] Loading last known ledger
2017-10-18T13:57:55.840 GANOG [Ledger WARNING] Some buckets are missing in 'c:/prod/buckets'.
2017-10-18T13:57:55.842 GANOG [Ledger WARNING] Attempting to recover from the history store.
2017-10-18T13:57:55.844 GANOG [History INFO] Starting RepairMissingBucketsWork
```

## Expected result
It should have been able to download and replay the 2 extra ledgers that I was asking for.

vogel:
Fixes #1342
Also probably fixes some other issues where people were reporting missing buckets at starts, and these were not buckets from any snapshot (as they were not present in history archives).

### DIFF
--- a/src/ledger/LedgerManagerImpl.cpp
+++ b/src/ledger/LedgerManagerImpl.cpp
@@ -1010,6 +1010,13 @@ LedgerManagerImpl::storeCurrentLedger()
         has.resolveAnyReadyFutures();
     }
 
+    // we will need these buckets after restart
+    for (auto const& bucket : has.allBuckets())
+    {
+        mApp.getBucketManager()
+            .getBucketByHash(hexToBin256(bucket))
+            ->setRetain(true);
+    }
     mApp.getPersistentState().setState(PersistentState::kHistoryArchiveState,
                                        has.toString());
 }
--- a/src/bucket/BucketManager.h
+++ b/src/bucket/BucketManager.h
@@ -97,5 +97,8 @@ class BucketManager : NonMovableOrCopyable
     // Restart from a saved state: find and attach all buckets in `has`, set
     // current BL.
     virtual void assumeState(HistoryArchiveState const& has) = 0;
+
+    // Ensure all needed buckets are retained
+    virtual void shutdown() = 0;
 };
 }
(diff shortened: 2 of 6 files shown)


MonsieurNicolas:
this is confusing: isn't that method just doing garbage collection? The only thing it seems to do is remove buckets from `mSharedBuckets`.
Also, why is it explicitly being not run during catchup in the first place?

vogel:
I think I have found (an obvious) place to retains all required buckets. I'm keeping previous commit as it seems to be a good idea to do a bucket GC at end of run, even in case of commandline catchup.

src/ledger/LedgerManagerImpl.cpp
```diff
@@ -1010,6 +1010,13 @@ LedgerManagerImpl::storeCurrentLedger()
         has.resolveAnyReadyFutures();
     }
 
+    // we will need these buckets after restart
+    for (auto const& bucket : has.allBuckets())
+    {
+        mApp.getBucketManager()
+            .getBucketByHash(hexToBin256(bucket))
+            ->setRetain(true);
+    }
```
MonsieurNicolas: nice, yes this is probably it

MonsieurNicolas:
r+ b65954283a58dd95f0e0d5a4736dc41b15392a8d

PR #1346, opened 2017-10-19
make nomination a little more strict on validation

MonsieurNicolas:
note: this has no downstream effect as we use sets to compute values.

I found this after doing an audit of how we use `std::is_sorted` in the code.


jonjove:
r+ 5ef83f9

PR #1349, opened 2017-10-20
Master weight tests

MonsieurNicolas:
added a couple tests that were missing for checking master weight behavior
also fixed a multisig test that was not checking proper transaction type


src/transactions/TxEnvelopeTests.cpp
```diff
@@ -199,16 +199,51 @@ TEST_CASE("txenvelope", "[tx][envelope]")
...
+            auto tx = a1.tx({payment(root, 1000)});
+            // only sign with s2 (med)
+            tx->getEnvelope().signatures.clear();
+            tx->addSignature(s2);
+
+            ThresholdSetter th2;
+            th2.masterWeight = make_optional<int>(0);
+            auto setOptionsTx = a1.tx({setOptions(nullptr, nullptr, nullptr,
+                                                  &th2, nullptr, nullptr)});
+            setOptionsTx->addSignature(s2);
```
vogel: this transaction is never executed - it is only created as transaction object, you need to call applyCheck to execute it

MonsieurNicolas: ah yes - I missed that the setup had the `masterweight` with a weight of `high`.
I reworked the test a bit, should be closer to what I originally wanted.

vogel:
r+ 1d63cd764576443dcf7baf15659b5c7763da3dd5

PR #1350, opened 2017-10-21
Test debug improvements

MonsieurNicolas:
First commit adds a new field to tests accounts so that it's easy to map a test account's `StrKey` representation (used in sql and logs) with the binary public key used in most of the code.

Second commit ensures that we cleanup the sql cache - this unlocks the database, which allows to perform writes into the database from outside of the process (sqlite) without getting a "database is locked" error message (which happens when attempting to modify it as the cache creates an implicit transaction).


jonjove:
r+ ead7590

MonsieurNicolas:
r+ a71c0107781ba8e82a6f0bebe59ba6cab537a04e

PR #1351, opened 2017-10-23
re-introduced check in scp for well formed quorum set

MonsieurNicolas:
check must be included in scp library in order to keep it as much as possible standalone
this was removed by accident in 033ab7abbd789c4b2d942e7ef2b868d62c2478be


vogel:
OK. I removed this, because it is no longer possible to mark envelope with not sane quorum set as ready (as PendingEnvelopes discards such quorums sets early, and isFullyFetched() always returns false for such envelopes).

MonsieurNicolas:
it ran into some intermittent failure that is in master right now. No idea where it failed... I am adding some instrumentation to help pinpoint.

MonsieurNicolas:
r+ 17786c449e101606667abbafbfac1c279e2d49cf

PR #1353, opened 2017-10-24
simple test reporter fixes

MonsieurNicolas:
fixes crash that could occur when in test reporter when running into bad test cases

### DIFF
--- a/src/test/SimpleTestReporter.h
+++ b/src/test/SimpleTestReporter.h
@@ -56,7 +56,7 @@ struct SimpleTestReporter : public ConsoleReporter
 
         if (result.isOk())
             return true;
-        assertionStarting(mLastAssertInfo);
+        ConsoleReporter::assertionStarting(mLastAssertInfo);
         return ConsoleReporter::assertionEnded(_assertionStats);
     }
 
--- a/src/test/SimpleTestReporter.h
+++ b/src/test/SimpleTestReporter.h
@@ -32,6 +32,7 @@ struct SimpleTestReporter : public ConsoleReporter
     void
     testCaseStarting(TestCaseInfo const& ti) override
     {
+        ConsoleReporter::testCaseStarting(ti);
         stream << "\"" << ti.name << "\" " << ti.lineInfo << std::endl;
     }
 
--- a/src/test/SimpleTestReporter.h
+++ b/src/test/SimpleTestReporter.h
@@ -36,9 +36,15 @@ struct SimpleTestReporter : public ConsoleReporter
     }
 
     void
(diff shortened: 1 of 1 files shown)


src/test/SimpleTestReporter.h
```diff
@@ -36,9 +36,21 @@ struct SimpleTestReporter : public ConsoleReporter
...
     void
-    assertionStarting(AssertionInfo const& ai) override
+    sectionStarting(SectionInfo const& _sectionInfo) override
     {
         printDot();
+        ConsoleReporter::sectionStarting(_sectionInfo);
+    }
+
+    void
+    sectionEnded(SectionStats const& _sectionStats) override
```
jonjove: This function isn't doing anything but forwarding to the function that is being overloaded `ConsoleReporter::sectionEnded`. This is the default behavior of a virtual function that does not have an override, so why provide it explicitly?

src/test/SimpleTestReporter.h
```diff
@@ -36,9 +36,21 @@ struct SimpleTestReporter : public ConsoleReporter
...
+    {
+        ConsoleReporter::sectionEnded(_sectionStats);
+    }
+
+    void
+    assertionStarting(AssertionInfo const& ai) override
```
jonjove: Perhaps I'm missing something, but this function doesn't seem to do anything anymore. `assertionStarting` does nothing but store the argument in the private variable `mLastAssertInfo`. But the only place `mLastAssertInfo` is used is as the argument to `assertionStarting`. If this is expected, can you explain what is happening here?

jonjove: Why do we no longer want to print a dot on `assertionStarting`?

src/test/SimpleTestReporter.h
MonsieurNicolas: ah yes, that's a left over from a refactor. removing

src/test/SimpleTestReporter.h
MonsieurNicolas: `mLastAssertInfo` is used in `assertionEnded`. The parameter to `Ended` doesn't have the right information (like the line number). In essence, what this reporter does is filter the events that are passed down to the console reporter, which is why we simulate the right sequence of start/end for the console reporter so that it has the right data to do its reporting.

MonsieurNicolas: dot on `assertionStarting` turned out to be completely useless (almost never triggers), now it's doing it with each section which is much better

jonjove:
r+ 659b941

PR #1355, opened 2017-10-24
fixed intermittent crash on Windows (object lifetime issue)

### LINKED ISSUE #1354: Windows: crash on shutdown after running catchup-at
Sometimes after running catchup, the app crashes. It looks like it's related to files in use.

For example:
`stellar-core.exe --catchup-at 10000`

MonsieurNicolas:
resolves #1354

Problem was really that `*Work` instances were keeping other `Work`
references alive via the `WorkParent` class, which defeated the
deterministic (field declaration) order of the `*Work` class.

On Windows this would create situations where resource de-allocation of
file system related objects were racing against various destructors.

This change forces all references to children to be released so that the only references left to related Work items should be member variables.

### DIFF
--- a/src/catchup/ApplyBucketsWork.cpp
+++ b/src/catchup/ApplyBucketsWork.cpp
@@ -54,6 +54,7 @@ ApplyBucketsWork::ApplyBucketsWork(
 
 ApplyBucketsWork::~ApplyBucketsWork()
 {
+    clearChildren();
 }
 
 BucketList&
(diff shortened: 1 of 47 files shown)


MonsieurNicolas:
there might be a more elegant fix, but I suspect it involves making deeper changes to `WorkParent` and `WorkManager`

vogel:
r+ cb91733f6541e7fcb9645f46cf9cbb6abf70dadb

PR #1358, opened 2017-10-26
Fix simulation timing

### LINKED ISSUE #1357: Herder tests failing on Mac
It looks like the tests introduced in #1336 are not passing on Mac for unknown reason.

MonsieurNicolas:
The simulation condition check was scheduled at the exact same moment than the timer for closing a ledger, the order of execution was therefore implementation dependent.

Solution is to simply schedule the check timer slightly offset from what everything else does.

As most code deals with timers in seconds, an offset of a few milliseconds guarantees that it won't trigger at the exact same moment.

resolves #1357


src/herder/HerderTests.cpp
```diff
@@ -882,27 +882,25 @@ TEST_CASE("quick restart", "[herder][quickRestart]")
...
-    REQUIRE(currentListenerLedger() >= currentLedger - 1);
+    REQUIRE(currentListenerLedger() == currentLedger);
 
     // disconnect listener
     simulation->dropConnection(validatorKey.getPublicKey(),
                                listenerKey.getPublicKey());
 
     auto static const SMALL_GAP = Herder::MAX_SLOTS_TO_REMEMBER;
-    auto static const BIG_GAP = Herder::MAX_SLOTS_TO_REMEMBER + 1;
+    auto static const BIG_GAP = Herder::MAX_SLOTS_TO_REMEMBER + 2;
```
jonjove: Why `+ 2` instead of `+ 1` here? Isn't the idea to test for the smallest gap that should fail?

I tried running it with my suggested change here and saw that "does not work when gap is big" failed in that situation. Why does it fail when the gap is bigger than `Herder::MAX_SLOTS_TO_REMEMBER`? @vogel

vogel: It looks like we can change SMALL_GAP is MAX_SLOTS_TO_REMEMBER + 1 and keep BIG_GAP as MAX_SLOTS_TO_REMEMBER + 2.

I think it is because of SCP::purgeSlots implementation and getSCP().purgeSlots(slotIndex - MAX_SLOTS_TO_REMEMBER); call.

It seems to keep last MAX_SLOTS_TO_REMEMBER + 1 slots instead of MAX_SLOTS_TO_REMEMBER.

We could adjust purgeSlots to have if (it->first <= maxSlotIndex) instead of if (it->first < maxSlotIndex) and these offsets would make sense again.

MonsieurNicolas: yeah I didn't want to mix up PRs. I updated the code with a comment to clarify what those gaps are

MonsieurNicolas:
updated @jonjove

src/herder/HerderTests.cpp
vogel: I think we should keep current behavior, as it is correct.

Each node has current slot and remembers MAX_SLOTS_TO_REMEMBER. So being able to recover from MAX_SLOTS_TO_REMEMBER + 1 gap seems reasonable.

jonjove:
r+ d0741c6

PR #1360, opened 2017-10-26
Remove "Editable" Versions of Application and LedgerManager

jonjove:
Removed `ApplicationEditableVersion` and `LedgerManagerEditableVersion` as they were unnecessary. Also changed the way that tests create Applications so that they all use a new function `createTestApplication` which updates to the desired version.


src/test/TestUtils.cpp
```diff
@@ -3,54 +3,25 @@
...
-{
-    mLedgerManager =
-        make_unique<LedgerManagerEditableVersion>(*this, cfg.DESIRED_BASE_FEE);
-    mHerder = Herder::create(*this); // need to recreate
-    newDB();
-}
-
-LedgerManagerEditableVersion&
-ApplicationEditableVersion::getLedgerManager()
+setCurrentLedgerVersion(LedgerManager& lm, uint32_t currentLedgerVersion)
```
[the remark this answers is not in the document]
jonjove: Good idea. There is already a function named `test` so I added this function to the namespace `testutil`.

src/transactions/InflationTests.cpp
```diff
@@ -198,7 +195,6 @@ doInflation(Application& app, int ledgerVersion, int nbAccounts,
     }
     LedgerManager& lm = app.getLedgerManager();
     LedgerHeader& cur = lm.getCurrentLedgerHeader();
-    cur.feePool = 10000;
```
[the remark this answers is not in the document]
jonjove: Good point. Done.

src/test/TestUtils.h
```diff
@@ -10,28 +10,13 @@
...
+Application::pointer
+createTestApplication(VirtualClock& clock, Config const& cfg);
 
-  private:
-    uint32_t mBaseFee;
-    uint32_t mCurrentLedgerVersion;
-};
-
-class ApplicationEditableVersion : public ApplicationImpl
+namespace testutil
```
vogel: we already have txtest namespace for testing transaction, maybe use testutil for that as well?

src/transactions/ChangeTrustTests.cpp
```diff
@@ -131,7 +129,7 @@ TEST_CASE("change trust", "[tx][changetrust]")
             validateTrustLineIsConst();
         });
 
-        for_versions_from(3, app, [&] {
+        for_versions_from(3, *app, [&] {
```
vogel: wouldn't it be simpler to change what for_versions_*** accepts from reference to pointer?

src/test/TestUtils.cpp
```diff
@@ -3,54 +3,25 @@
...
-    mHerder = Herder::create(*this); // need to recreate
-    newDB();
-}
-
-LedgerManagerEditableVersion&
-ApplicationEditableVersion::getLedgerManager()
+testutil::setCurrentLedgerVersion(LedgerManager& lm, uint32_t currentLedgerVersion)
 {
-    return static_cast<LedgerManagerEditableVersion&>(*mLedgerManager);
+    lm.getCurrentLedgerHeader().ledgerVersion = currentLedgerVersion;
```
vogel: I really don't like the idea that this is even legal in our code to do that.... We need better encapsulation.

src/test/TestUtils.h
jonjove: I think it is better to keep `txtest` and `testutil` separate. If there are functions which are not specific to transaction tests in `txtest` then we should move those to `testutil`. But having a namespace which has the utility functions for a given set of tests is a reasonable way to organize things in my opinion. I have even followed this approach in the tests for invariants that I have been writing, where there is a namespace for some of the test-specific functionality.

src/transactions/ChangeTrustTests.cpp
jonjove: This was an approach I considered, but since I was already changing `app.` to `app->` and `app` to `*app` in so many places I thought I would just do it everywhere. If you and @MonsieurNicolas prefer I change the interface of `for_versions` then I am happy to do so. Another approach would be to just write code like
```
auto appPtr = createTestApplication(...);
auto& app = *app;
```
I like this solution less, but it would result in a smaller change. Let me know what you both prefer.

MonsieurNicolas: yeah I'm fine with the way it's done right now: passing a pointer implies that you can pass null for some reason which is not the case.

src/test/TestUtils.cpp
MonsieurNicolas: Ideally this would be done by cleaning up how we use `LedgerHeaderFrame` and stop passing around just LedgerHeaders... but this is a fight for another day.

src/test/TestUtils.h
vogel: OK, that is reasonable.

src/transactions/ChangeTrustTests.cpp
vogel: I'm fine with both ways, I was just wondering if you really like adding hundreds of '*' into code ;)

MonsieurNicolas:
r+ a38ab07aa75c5d8172b5368ef160a52baa9fad7a

PR #1361, opened 2017-10-27
inherit environment when spawning child processes

### LINKED ISSUE #1291: Processes executed by stellar-core do not inherit the environment from stellar-core on all platforms
For example if stellar-core is executed with AWS env vars like:

AWS_ACCESS_KEY_ID=xxx
AWS_SECRET_ACCESS_KEY=xxx
AWS_DEFAULT_REGION=us-east-2 aws

these should be passed to sub processes.

MonsieurNicolas:
on system other than Windows, we were clearing the environment for the child process.

resolves #1291


vogel:
r+ a0f5830939f4e4925480e38245b3bf11a9089b03

PR #1363, opened 2017-10-30
Bump libsodium to 1.0.15

MonsieurNicolas:


MonsieurNicolas:
In particular, here are some changes that seem relevant since the last bump:
 * crypto_sign_ed25519_pk_to_curve25519() now rejects points that are not on the curve, or not in the main subgroup.
 * Further changes have been made to ensure that smart compilers will not optimize out code that we don't want to be optimized.

MonsieurNicolas:
please hold on this one, I am going to remove the customization that is not required anymore

MonsieurNicolas:
done, we now point to the unmodified 1.0.15. Thanks @jedisct1 for pointing this out

PR #1366, opened 2017-11-02
fix configurations used for fuzzing

MonsieurNicolas:
currently, attempting to run `make fuzz` will crash right away - it turns out that the passphrase was not set on the config files.

This PR uses test configs as a base for the fuzzer apps


vogel:
Looks OK for me. Should we wait with merge for release?

vogel:
r+ e2ab22f46eefb1362297d0dde7263d7caf2818c9

PR #1375, opened 2017-11-06
add --all-versions options to test runs

vogel:
Now it can be selected at runtime if tests are to be run againsts all
protocol versions or only against current.

### DIFF
--- a/src/test/test.cpp
+++ b/src/test/test.cpp
@@ -161,12 +161,19 @@ test(int argc, char* const* argv, el::Level ll,
 
     auto unusedTokens = session.unusedTokens();
     using namespace Clara;
-    gTestAllVersions =
-        std::find_if(std::begin(unusedTokens), std::end(unusedTokens),
-                     [](Parser::Token const& token) {
-                         return token.type == Parser::Token::LongOpt &&
-                                token.data == "all-versions";
-                     }) != std::end(unusedTokens);
+
+    for (auto const& token : unusedTokens)
+    {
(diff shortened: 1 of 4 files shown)


src/test/test.cpp
```diff
@@ -151,15 +152,30 @@ test(int argc, char* const* argv, el::Level ll,
...
+    Session session{};
+    auto r =
+        session.applyCommandLine(argc, argv, Session::OnUnusedOptions::Ignore);
+    if (r != 0)
+        return r;
+
+    auto unusedTokens = session.unusedTokens();
+    using namespace Clara;
+    gTestAllVersions =
+        std::find_if(std::begin(unusedTokens), std::end(unusedTokens),
```
MonsieurNicolas: I'd rather have an explicit loop over `unusedToken` and error if we don't recognize one of them

vogel: Updated.

Note that Catch 2.0 has much better API for adding own command line arguments - https://github.com/catchorg/Catch2/blob/master/docs/own-main.md#adding-your-own-command-line-options

MonsieurNicolas: cool. Ah yes 2.0 seems promising, but it won't be cheap. I see that they have a bunch of breaking changes like removed `Catch::toString` . We can probably consider switching to it in a few months when they have hammered out the bugs.

MonsieurNicolas:
r+ eaa4a6628474558bd2d6445e8089afeb2802efef

PR #1378, opened 2017-11-08
Fixed issue where virtual members of Application were called from its constructor

jonjove:
As the subsystems of `Application` were being created in the constructor of `ApplicationImpl`, they were calling virtual functions. This issue is fixed in this PR, and some related clean-up was also done simultaneously.

### DIFF
--- a/src/main/Application.h
+++ b/src/main/Application.h
@@ -4,6 +4,7 @@
 // under the Apache License, Version 2.0. See the COPYING file at the root
 // of this distribution or at http://www.apache.org/licenses/LICENSE-2.0
 
+#include "main/Config.h"
 #include "xdr/Stellar-types.h"
 #include <memory>
 #include <string>
@@ -21,7 +22,6 @@ namespace stellar
 {
 
 class VirtualClock;
-class Config;
 class TmpDirManager;
 class LedgerManager;
 class BucketManager;
@@ -39,6 +39,9 @@ class CommandHandler;
 class WorkManager;
 class BanManager;
 class StatusManager;
+
+class Application;
+void validateNetworkPassphrase(std::shared_ptr<Application> app);
 
 /*
  * State of a single instance of the stellar-core application.
@@ -150,6 +153,8 @@ class Application
 
     virtual ~Application(){};
 
+    virtual void initialize() = 0;
+
     // Return the time in seconds since the POSIX epoch, according to the
     // VirtualClock this Application is bound to. Convenience method.
     virtual uint64_t timeNow() = 0;
@@ -257,6 +262,18 @@ class Application
     // copy made of `cfg`.
     static pointer create(VirtualClock& clock, Config const& cfg,
                           bool newDB = true);
+    template <typename T>
+    static std::shared_ptr<T> create(VirtualClock& clock, Config const& cfg,
+                          bool newDB = true)
+    {
+        auto ret = std::make_shared<T>(clock, cfg);
+        ret->initialize();
+        if (newDB || cfg.DATABASE.value == "sqlite3://:memory:")
+            ret->newDB();
+        validateNetworkPassphrase(ret);
+
+        return ret;
+    }
 
   protected:
     Application()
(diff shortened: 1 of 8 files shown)


src/overlay/OverlayManagerTests.cpp
```diff
@@ -75,29 +77,37 @@ class OverlayManagerTests
...
-            mOverlayManager = make_shared<OverlayManagerStub>(*this);
-            newDB();
         }
+
         virtual OverlayManagerStub&
         getOverlayManager() override
         {
-            return *mOverlayManager;
+            auto& overlay = ApplicationImpl::getOverlayManager();
+            return dynamic_cast<OverlayManagerStub&>(overlay);
```
MonsieurNicolas: the use of dynamic_cast makes it look like you don't know the type, I'd prefer to use static_cast instead. On that note I don't even think we need to override that method.

src/test/TestUtils.h
```diff
@@ -10,14 +10,22 @@
...
-createTestApplication(VirtualClock& clock, Config const& cfg);
-
 namespace testutil
 {
 void setCurrentLedgerVersion(LedgerManager& lm, uint32_t currentLedgerVersion);
 }
 
+template <typename T = ApplicationImpl>
+Application::pointer
+createTestApplication(VirtualClock& clock, Config const& cfg)
```
MonsieurNicolas: having those implementations in the header files makes me a bit sad (and will cause us to recompile way too many things when updating the implementation of the factories). What do you think of doing something like https://isocpp.org/wiki/faq/templates#templates-defn-vs-decl instead?

src/overlay/ItemFetcherTests.cpp
```diff
@@ -43,18 +44,21 @@ class ApplicationStub : public ApplicationImpl
...
-        mHerder = std::make_shared<HerderStub>(*this);
-        newDB();
     }
 
     virtual HerderStub&
     getHerder() override
     {
-        return *mHerder;
+        auto& herder = ApplicationImpl::getHerder();
+        return dynamic_cast<HerderStub&>(herder);
```
MonsieurNicolas: same re: dynamic_cast that we don't use in our code base

src/main/Application.h
```diff
@@ -257,6 +262,19 @@ class Application
     // copy made of `cfg`.
     static pointer create(VirtualClock& clock, Config const& cfg,
                           bool newDB = true);
+    template <typename T>
+    static pointer create(VirtualClock& clock, Config const& cfg,
```
MonsieurNicolas: if you were returning std::shared_ptr<T> instead of pointer, it would save you a bunch of work elsewhere

vogel: Github is removing <> from ::shared_ptr&lt;T> in comments.

vogel:
I don't think this is best possible fix for that problem (but it is simplest one and at least acceptable).

We now got 2-phase initialization, which is not very cool way to initialize classes (constructors loose their meaning). I think we could go different way here - do not depend on Application in all the classes we have.

For example, Database constructor could look like this:
Database(MetricsRegistry, VirtualClock, PersistentState, Config) - because it uses those four classes. It does not use anything more from Application. This way we could make dependencies clear and it would force us to remove cyclic dependencies between classes. It is a lot of work, but if we agree that this is a good way to go, we could start implementing that step by step along other changes.

And then we would not have to worry about calling virtual methods from contructors.

src/test/TestUtils.h
MonsieurNicolas: it's also that we depend on the compiler being smart to not generate bloated code. I guess we'll see if the stellar-core binaries become too big we can try newer versions of the compilers.

MonsieurNicolas:
I'm happy with this PR the way it is then. You can squash/rebase and we'll get it in.

MonsieurNicolas:
r+ eba2fef53c8cbacdd487688c3210b53c1e11460c

PR #1379, opened 2017-11-13
Upgrade fixes

### LINKED ISSUE #1368: PREFERRED_UPGRADE_DATETIME is not properly respected
there are a couple problems with the current implementation:
* the local node's time is used to decide for voting - this may cause different validators to have a different opinion even though they are really agreeing to the change
    * the network time should be used instead
* the only place where we check for the time is in `HerderImpl::prepareUpgrades` (used as nomination value)
    * this may cause the node to vote for an upgrade that was requested by a different node, causing the switch to occur ahead of schedule

vogel:
Fixes #1368

This fix requires introducing two modes of value validation - hard for nomination and soft for ballot.
During nomination node does not want to vote for upgrades that are not in its PREFERRED_UPGRADE_DATETIME. But if consensus is made with that upgrade, it is accepted if upgrade values other that this datetime are OK.


src/herder/HerderImpl.cpp
```diff
@@ -767,19 +767,27 @@ HerderImpl::triggerNextLedger(uint32_t ledgerSeqToTrigger)
                               lcl.header.scpValue);
 }
 
+bool
+HerderImpl::timeForUpgrade() const
```
MonsieurNicolas: I don't think you need to add this helper: it's probably much simpler to change the way PREFERRED_UPGRADE_DATETIME is done by configuring it by default to have the value of epoch (instead of using an optional field)

src/herder/HerderSCPDriver.cpp
```diff
@@ -322,7 +331,8 @@ HerderSCPDriver::validateValueHelper(uint64_t slotIndex,
 }
 
 SCPDriver::ValidationLevel
-HerderSCPDriver::validateValue(uint64_t slotIndex, Value const& value)
+HerderSCPDriver::validateValue(uint64_t slotIndex, Value const& value,
```
MonsieurNicolas: let's not make changes to the definition of a "valid" value in SCP (see my comment in the PR for rationale)

MonsieurNicolas:
The changes in SCP don't seem to make sense to me: the only reason you had to do this is because you made the switch to network time after making that change (so local nodes are very likely to have a different opinion of what "valid" is at a given time, worst their opinion changes during SCP round).
I'd rather keep the notion of what a "valid" value is, simple.

The way I see it: when using network time as upgrade time, the node is explicitly configured to not want to upgrade before that agreed upon time. If other nodes decided to go ahead with the upgrade, it's a violation of the validator's intent, therefore I do not have a problem with the node crashing/dropping off the network.

src/herder/HerderSCPDriver.cpp
vogel: Reason of changing that was not switch from local time to network time, but my understanding of meaning of  PREFERRED_UPGRADE_DATETIME. I've understand it as: "I will vote for upgrade starting at that time, but if my quorum votes for it earlier, I'll accept that".

I've restored validateValue to previous state, but maybe we should consider changing PREFERRED_UPGRADE_DATETIME name to something more close to it current behavior, like EARLIEST_UPGRADE_DATETIME?

vogel:
Changed to account for Nicolas suggestions.

vogel:
It requires tests updates due to new findings.

PR #1384, opened 2017-11-14
Fix double fee checking

### LINKED ISSUE #1220: Allow merge when account has minimum balance + one tx fee
Reported by @stanford-scs 

> So I had an account with 20.00001 XLM, and tried to merge with another account.  The merge failed with insufficient funds.  That's kind of annoying.  Really any account should be able to merge even if it has below the minimum, with the transaction fee just deducted from the amount credited to the destination account.

vogel:
Fixes #1220

### DIFF
--- a/src/main/Config.h
+++ b/src/main/Config.h
@@ -31,7 +31,7 @@ class Config : public std::enable_shared_from_this<Config>
     std::string expandNodeID(std::string const& s) const;
 
   public:
-    static const int CURRENT_LEDGER_PROTOCOL_VERSION = 8;
+    static const int CURRENT_LEDGER_PROTOCOL_VERSION = 9;
 
     typedef std::shared_ptr<Config> pointer;
 
--- a/src/transactions/TransactionFrame.cpp
+++ b/src/transactions/TransactionFrame.cpp
@@ -296,8 +296,16 @@ TransactionFrame::commonValid(SignatureChecker& signatureChecker,
         return false;
     }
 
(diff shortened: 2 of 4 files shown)


MonsieurNicolas:
let's use #1387 to get master back in a clean state wrt clang-format

src/main/Config.h
```diff
@@ -31,7 +31,7 @@ class Config : public std::enable_shared_from_this<Config>
     std::string expandNodeID(std::string const& s) const;
 
   public:
-    static const int CURRENT_LEDGER_PROTOCOL_VERSION = 8;
+    static const int CURRENT_LEDGER_PROTOCOL_VERSION = 9;
```
MonsieurNicolas: please make this change in its own commit as it's very significant

src/transactions/TransactionFrame.cpp
```diff
@@ -296,8 +296,13 @@ TransactionFrame::commonValid(SignatureChecker& signatureChecker,
         return false;
     }
 
+    auto balanceAfter =
```
MonsieurNicolas: add a comment to explain that fees are captured in a different step when applying

vogel:
All suggestions implemented

MonsieurNicolas:
looks good but it has conflicts with master probably due to clang-format

vogel:
updated

MonsieurNicolas:
r+ 53308debfada0fbadf6f40757a7ed547e50b1d3c

Issue #1381, opened 2017-11-14
Missing invariant files in vs project files


MonsieurNicolas:
As per reported in #1380

PR #1387, opened 2017-11-15
Visual Studio project update

MonsieurNicolas:
This PR addresses a few things missed in the last few PRs:
* clang-format not always run (I am opening an issue to enforce this in travis - I'm getting tired of it)
* warnings introduced (one of the changes I did was to stop using `size_t` as the type for bucketlist levels - there is no reason we ever need such large numbers and we had a few places in code that were assuming `uint32_t` anyways)
* visual studio project drift

of note there was a few bad `REQUIRE` (see last commit) that trips (for the right reason) Visual Studio - this probably needs to be redone somehow.

### DIFF
--- a/src/invariant/BucketListIsConsistentWithDatabaseTests.cpp
+++ b/src/invariant/BucketListIsConsistentWithDatabaseTests.cpp
@@ -276,11 +276,6 @@ class ApplyBucketsWorkAddEntry : public ApplyBucketsWork
         CLOG(INFO, "Invariant") << "Try to add @ " << mFromLedgerSeq;
     }
 
-    ~ApplyBucketsWorkAddEntry()
-    {
-        REQUIRE(mAdded);
-    }
-
     Work::State
     onSuccess() override
     {
@@ -338,7 +333,12 @@ class ApplyBucketsWorkAddEntry : public ApplyBucketsWork
                 mAdded = true;
             }
         }
-        return ApplyBucketsWork::onSuccess();
+        auto r = ApplyBucketsWork::onSuccess();
+        if (r == WORK_SUCCESS)
+        {
+            REQUIRE(mAdded);
+        }
+        return r;
     }
 };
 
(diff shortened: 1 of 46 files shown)


MonsieurNicolas:
resolve #1381

src/invariant/BucketListIsConsistentWithDatabaseTests.cpp
```diff
@@ -278,7 +278,7 @@ class ApplyBucketsWorkAddEntry : public ApplyBucketsWork
 
     ~ApplyBucketsWorkAddEntry()
     {
-        REQUIRE(mAdded);
+        // REQUIRE(mAdded);
```
jonjove: The only issue with this REQUIRE (and the corresponding ones for Delete and Modify) is the possibility of throwing from a constructor, right? These REQUIRE statements are used to check that the test actually did make the required alteration to the data, so I consider them too important to simply comment out.

The solution that I propose is to check this after `ApplyBucketsWork::onSuccess` returns `WORK_SUCCESS`. For example, we could make the following change at line 341 (and corresponding changes at lines 386 and 498):
```
auto abwRes = ApplyBucketsWork::onSuccess();
if (abwRes == WORK_SUCCESS)
{
    REQUIRE(mAdded);
}
return abwRes;
```
It would be even better from a code duplication perspective to make a function out of the above code snippet since it will be used 3 times.

This change would still cause the REQUIRE statements to occur after ApplyBucketsWork is complete, but would avoid putting the REQUIRE in the constructor. In fact, at this point we could simply use the default destructor instead of providing our own. Let me know if you see any other issues with this solution.

MonsieurNicolas: I moved the checks per your suggestion. I didn't create a function for this though as it seems overkill.

jonjove:
r+ bb92809

PR #1389, opened 2017-11-15
enforce that clang-format was run in travis

### LINKED ISSUE #1388: enforce that clang-format was properly run in travis
We should be bouncing PRs that don't meet our coding standard:
right now a PR will pass validations in travis even though `clang-format` or `make format` was not run.

We should add a check in `travis.sh` to ensure that `make format` is a no-op.

MonsieurNicolas:
resolves #1388 

like title says

this can only be committed on a clean master (so depends on #1387 as of right now)


MonsieurNicolas:
Ignore this, script is wrong

PR #1390, opened 2017-11-15
Restart nodes test and simulation improvements

### LINKED ISSUE #1382: Improve simulation tests
The simulation tests right now are fairly limited as each app instance in the simulation shares the same timer.

Because of this:
* we can't simulate instance restarts (app kills the io service on shutdown)
* nodes may starve other nodes, especially when time is set to virtual (they share the same asio queue, so if a node keeps posting events, it will process work ahead of nodes that rely on timers)

MonsieurNicolas:
resolves #1382 

this PR changes the way the `Simulation` class schedules work between its instances:
before this PR, a single clock was shared (among with its IOService), this PR allocates a clock and an IOService for each of the nodes.
There are many reasons why a shared IOService was bad, here are a few:
* the app destructor kills its ioservice - basically killing one app, kills all other apps
* work between apps bleeds between apps, causing all sorts of strange behaviors that would not otherwise occur (for example: a "busy" app that reposts work on the queue will end up preempting a lot of work that other nodes could be doing)

This allows in particular (also part of this PR), to restart nodes in a simulation by destroying/recreating nodes. As part of those resilience tests, I introduced a new "CustomA" simulation type that somewhat  resembles production.

Added benefit: if one day we feel adventurous enough, changing the simulation code to run on multiple threads should be fairly straightforward.

I also fixed a few bugs in the LoopbackPeer code that was causing crashes/undefined behavior, see the list of commits for more details.

### DIFF
--- a/src/scp/NominationProtocol.cpp
+++ b/src/scp/NominationProtocol.cpp
@@ -397,6 +397,8 @@ NominationProtocol::processEnvelope(SCPEnvelope const& envelope)
                     {
                         mVotes.emplace(newVote);
                         modified = true;
+                        mSlot.getSCPDriver().nominatingValue(
+                            mSlot.getSlotIndex(), newVote);
                     }
                 }
 
@@ -442,8 +444,8 @@ NominationProtocol::nominate(Value const& value, Value const& previousValue,
                              bool timedout)
 {
     if (Logging::logDebug("SCP"))
-        CLOG(DEBUG, "SCP") << "NominationProtocol::nominate "
-                           << mSlot.getSCP().getValueString(value);
+        CLOG(DEBUG, "SCP") << "NominationProtocol::nominate (" << mRoundNumber
+                           << ") " << mSlot.getSCP().getValueString(value);
 
     bool updated = false;
 
--- a/src/simulation/Simulation.cpp
+++ b/src/simulation/Simulation.cpp
@@ -40,6 +40,8 @@ Simulation::Simulation(Mode mode, Hash const& networkID,
 
 Simulation::~Simulation()
 {
+    // kills all connections
+    mLoopbackConnections.clear();
     // destroy all nodes first
     mNodes.clear();
 
--- a/src/simulation/Simulation.cpp
+++ b/src/simulation/Simulation.cpp
@@ -107,6 +107,44 @@ Simulation::getNodeIDs()
 }
 
 void
+Simulation::removeNode(NodeID const& id)
+{
+    auto it = mNodes.find(id);
+    if (it != mNodes.end())
+    {
+        auto node = it->second;
+        mNodes.erase(it);
+        if (mMode == OVER_LOOPBACK)
+        {
+            dropAllConnections(id);
+        }
+        node.mApp->gracefulStop();
+        while (node.mClock->crank(false) > 0)
+            ;
+    }
+}
+
+void
+Simulation::dropAllConnections(NodeID const& id)
+{
+    if (mMode == OVER_LOOPBACK)
+    {
+        mLoopbackConnections.erase(
+            std::remove_if(mLoopbackConnections.begin(),
+                           mLoopbackConnections.end(),
+                           [&](std::shared_ptr<LoopbackPeerConnection> c) {
+                               return c->getAcceptor()->getPeerID() == id ||
+                                      c->getInitiator()->getPeerID() == id;
+                           }),
+            mLoopbackConnections.end());
+    }
+    else
+    {
+        throw std::runtime_error("can only drop connections over loopback");
+    }
+}
+
+void
 Simulation::addPendingConnection(NodeID const& initiator,
                                  NodeID const& acceptor)
 {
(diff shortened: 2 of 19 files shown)


src/simulation/Simulation.h
```diff
@@ -57,7 +57,7 @@ class Simulation : public LoadGenerator
 
     // returns true if all nodes have externalized
     // triggers and exception if a node externalized higher than num+maxSpread
-    bool haveAllExternalized(SequenceNumber num, uint32 maxSpread);
```
vogel: what about adding LedgerNumber typedef?

src/scp/NominationProtocol.cpp
```diff
@@ -397,7 +397,8 @@ NominationProtocol::processEnvelope(SCPEnvelope const& envelope)
                     {
                         mVotes.emplace(newVote);
                         modified = true;
-                        mSlot.getSCPDriver().nominatingValue(mSlot.getSlotIndex(), newVote);
```
vogel: I think that one should go with previous commit

src/simulation/Simulation.cpp
```diff
@@ -39,6 +40,10 @@ Simulation::Simulation(Mode mode, Hash const& networkID,
 
 Simulation::~Simulation()
 {
+    for (auto& p : mNodes)
+    {
+        p.second.second.reset();
```
vogel: this is unreadable, what and why are we resetting here?

src/simulation/Simulation.cpp
```diff
@@ -47,47 +52,59 @@ Simulation::~Simulation()
...
-VirtualClock&
-Simulation::getClock()
+void
+Simulation::setCurrentTime(VirtualClock::time_point t)
 {
-    return mClock;
+    mClock.setCurrentTime(t);
+    for (auto& p : mNodes)
+    {
+        p.second.first->setCurrentTime(t);
```
vogel: as above, very cryptic line

src/simulation/Simulation.cpp
```diff
@@ -187,36 +205,146 @@ Simulation::stopAllNodes()
...
+        {
+            i++;
+        }
+    } while (i < nbTicks);
     return count;
 }
 
 bool
-Simulation::haveAllExternalized(SequenceNumber num, uint32 maxSpread)
+Simulation::haveAllExternalized(uint32 num, uint32 maxSpread)
```
vogel: this line should be in 1dfab15

src/simulation/Simulation.h
```diff
@@ -89,14 +89,19 @@ class Simulation : public LoadGenerator
...
     void addTCPConnection(NodeID initiator, NodeID acception);
 
+    bool mVirtualClockMode;
     VirtualClock mClock;
     Mode mMode;
     int mConfigCount;
     Application::pointer mIdleApp;
-    std::map<NodeID, Application::pointer> mNodes;
+    std::map<NodeID,
+             std::pair<std::shared_ptr<VirtualClock>, Application::pointer>>
```
vogel: simple

struct Node { std::shared_ptr<VirtualClock> clock; Application::pointer app; } instead of pair would make a lot of code in that commit clear, instead of refering to second.first or second.second all the time

src/simulation/Simulation.h
MonsieurNicolas: We could but that would require a code base sweep, I'll create an issue to discuss this - there are probably other types that we should typedef

src/simulation/Simulation.h
MonsieurNicolas: fair point, I'm updating

MonsieurNicolas:
@vogel I made your suggested changes. I also rebased on top of master now that we've done clang-format, so this is ready to merge

vogel:
This does not compile anymore. Here is error I'm getting:

```
simulation/Simulation.cpp: In member function ‘void stellar::Simulation::removeNode(const NodeID&)’:
simulation/Simulation.cpp:124:24: error: ‘struct stellar::Simulation::Node’ has no member named ‘second’
         auto& app = *p.second;
                        ^~~~~~
simulation/Simulation.cpp:126:26: error: ‘struct stellar::Simulation::Node’ has no member named ‘first’
         auto& clock = *p.first;
                          ^~~~~
simulation/Simulation.cpp:130:11: error: ‘struct stellar::Simulation::Node’ has no member named ‘second’
         p.second.reset();
           ^~~~~~
```

vogel:
r+ affc4b28fa3f3fb7059ec01027b5fbcab081bad9

vogel:
It seems travis does not support generic lambdas:

```
[1moverlay/OverlayManagerImpl.cpp:286:29: [0m[0;1;31merror: [0m[1m'auto' not allowed in lambda parameter[0m
                        [&](auto& p) { return p.get() == peer; });
```

MonsieurNicolas:
ah yes we didn't enable C++14 yet. Not sure why it could build on my machine (clang). I am refreshing this

MonsieurNicolas:
r+ b520dd2a3482a58d2198b0a6ea8e448f1a44b1c5

MonsieurNicolas:
One of the herder test needed to be updated: it was making (bad) assumptions on the state of the nodes, in that it was assuming that all messages had been processed.

MonsieurNicolas:
r+ 8536bc911281a2002b2ca39cfdc0f55f6ec96b71

PR #1392, opened 2017-11-15
Operations invariants

jonjove:
Moved all transaction-level invariants to the operation-level and added several new operation-level invariants as well. Also allowed invariants to specify whether they should be treated strictly or not.


src/invariant/AccountSubEntriesCountIsValidTests.cpp
```diff
@@ -0,0 +1,318 @@
...
+#include "test/TestUtils.h"
+#include "test/test.h"
+#include <random>
+#include <xdrpp/autocheck.h>
+
+using namespace stellar;
+using namespace stellar::InvariantTestUtils;
+
+static LedgerEntry
+generateRandomAccountWithNoSubEntries(uint32_t ledgerSeq)
```
MonsieurNicolas: same thing than earlier - I think this is overkill : we have a better chance of reaching high code coverage manually

src/invariant/ConservationOfLumens.cpp
```diff
@@ -0,0 +1,123 @@
...
+                                                   nullptr);
+            });
+    deltaBalances += std::accumulate(
+            delta.modifiedBegin(), delta.modifiedEnd(), static_cast<int64_t>(0),
+            [this] (int64_t lhs, LedgerDelta::ModifiedLedgerEntry const& rhs) {
+                return lhs + calculateDeltaBalance(&rhs.current->mEntry,
+                                                   &rhs.previous->mEntry);
+            });
+    deltaBalances += std::accumulate(
+            delta.deletedBegin(), delta.deletedEnd(), static_cast<int64_t>(0),
```
MonsieurNicolas: you don't need to cast `0` to int64 as it's a bigger type

src/invariant/ConservationOfLumensTests.cpp
```diff
@@ -0,0 +1,321 @@
...
+    std::vector<EntryFrame::pointer> result;
+    std::transform(entries.begin(), entries.end(), std::back_inserter(result),
+        [] (LedgerEntry const& le) {
+            return EntryFrame::FromXDR(le);
+        });
+    return result;
+}
+
+UpdateList
+generateUpdateList(std::vector<EntryFrame::pointer> const& current,
```
MonsieurNicolas: same comment than earlier: I am not sure going the random route for this particular class of code is the most elegant way to go about it. Basically this code even randomized doesn't really tell us if the invariant will work all the time as the type of changes it's doing are always of the same kind (and it's the same than a simple fixed test)

MonsieurNicolas:
you'll also have to rebase on top of https://github.com/stellar/stellar-core/pull/1393 instead of current master

src/invariant/Invariant.h
```diff
@@ -25,6 +25,8 @@ class Invariant
 
     virtual std::string getName() const = 0;
 
+    virtual bool isStrict() const = 0;
```
vogel: I think making this one virtual is a bit of overkill. What about mStrict field that is set in constructor and returned here?

src/main/ApplicationImpl.cpp
```diff
@@ -108,7 +108,7 @@ ApplicationImpl::initialize()
     mBucketManager = BucketManager::create(*this);
     mCatchupManager = CatchupManager::create(*this);
     mHistoryManager = HistoryManager::create(*this);
-    mInvariantManager = InvariantManager::create(*this);
+    mInvariantManager = createInvariantManager();
```
vogel: We should start thinking of injecting these dependencies into application more seriously. Maybe this PR could start that with passing InvariantManager to Application?

src/invariant/InvariantTestUtils.cpp
```diff
@@ -0,0 +1,83 @@
+// Copyright 2014 Stellar Development Foundation and contributors. Licensed
```
vogel: I think we should use 2017 here

src/invariant/MinimumAccountBalanceTests.cpp
```diff
@@ -0,0 +1,151 @@
...
+    cfg.INVARIANT_CHECKS = {"MinimumAccountBalance"};
+
+    for (uint32_t i = 0; i < 10; ++i)
+    {
+        VirtualClock clock;
+        Application::pointer app = createTestApplication(clock, cfg);
+
+        auto le = generateRandomAccount(2);
+        le = updateAccountWithRandomBalance(le, *app, gen, true, 0);
+        store(*app, false, {{EntryFrame::FromXDR(le), nullptr}});
```
vogel: What do you think of changing `store` signature to return bool?
It would work like this:
- if code in store does not throw, return true
- if code in store throws InvariantDoesNotHold, return false
- if code in store throws anything else, let it propagate

Then call site would look like this:
REQUIRE(store(*app, {{EntryFrame::FromXDR(le), nullptr}}));

And for test case below:
REQUIRE(!store(*app, {{EntryFrame::FromXDR(le), nullptr}}));

src/invariant/AccountSubEntriesCountIsValid.cpp
```diff
@@ -0,0 +1,224 @@
...
+
+void
+AccountSubEntriesCountIsValid::updateChangedSubEntriesCount(
+        std::map<AccountID, int32_t>& numSubEntriesChange,
+        std::map<AccountID, int32_t>& calculatedSubEntriesChange,
+        std::map<AccountID, int32_t>& signersChange,
+        LedgerEntry const* current,
+        LedgerEntry const* previous) const
+{
+    assert(current || previous);
```
vogel: `auto e = current ? current : previous;` would make lots of below code simpler

src/invariant/ChangedAccountsSubentriesCountIsValid.cpp
```diff
@@ -1,148 +0,0 @@
...
-                               "{} in accounts table vs {} calculated",
-                               KeyUtils::toStrKey(account),
-                               subentries.inAccountsTable,
-                               subentries.calculated);
-        }
-    }
-
-    for (auto const& account : getDeletedAccounts(delta))
-    {
-        auto subentries = numberOfSubentries(account, mDb);
```
vogel: NumberOfSubentries seems to not be used, can be deleted now

But maybe we should keep invariant-by-db checking code and call it at checkpoint boundaries?

src/invariant/LedgerEntryIsValid.cpp
```diff
@@ -0,0 +1,225 @@
+// Copyright 2014 Stellar Development Foundation and contributors. Licensed
```
vogel: I think it should be 2017

src/invariant/LedgerEntryIsValid.cpp
```diff
@@ -0,0 +1,225 @@
...
+        return checkIsValid(le.data.data());
+    default:
+        return "LedgerEntry has invalid type";
+    }
+}
+
+std::string
+LedgerEntryIsValid::checkIsValid(AccountEntry const& ae) const
+{
+    if (ae.balance < 0)
```
vogel: This will give us only first error even if account has multiple issues.
I think we should return all errors if possible.

src/invariant/LedgerEntryIsValid.cpp
```diff
@@ -0,0 +1,225 @@
...
+    {
+        return fmt::format("Account balance ({}) cannot be negative",
```
vogel: It should say "Account balance ({}) is negative", as it is listing errors here, not conditions, so the wording is consistent.

src/invariant/ConservationOfLumens.cpp
vogel: wouldn't that make std::accumulate return just int32_t, which can be invalid?

src/invariant/ConservationOfLumens.cpp
```diff
@@ -0,0 +1,123 @@
...
+            [this] (int64_t lhs, LedgerDelta::DeletedLedgerEntry const& rhs) {
+                return lhs + calculateDeltaBalance(nullptr,
+                                                   &rhs.previous->mEntry);
+            });
+
+    if (result.tr().type() == INFLATION)
+    {
+        int64_t inflationAmount = std::accumulate(
```
vogel: This name is misleading - inflation amount is sum of inflation payouts for accounts + a payout to fee pool, so maybe change that to inflationPayouts?

src/invariant/TotalCoinsEqualsBalancesPlusFeePool.cpp
```diff
@@ -1,58 +0,0 @@
...
-{
-    auto& lh = delta.getHeader();
-    if (lh.ledgerVersion <= 7) // due to bugs in previous versions
-    {
-        return {};
-    }
-
-    auto ledgerTotalCoins = lh.totalCoins;
-    auto feePool = lh.feePool;
-    auto databaseTotalCoins = sumOfBalances(mDb);
```
vogel: sumOfBalances is not used anymore and can be removed, but maybe we should consider calling those db-invariants on checkpoint boundaries?

src/invariant/ConservationOfLumens.cpp
MonsieurNicolas: ah yes you are right, the template uses that arg's type to derive the type of the accumulator

src/invariant/ChangedAccountsSubentriesCountIsValid.cpp
vogel: I mean checking content of DB to check if invariants are met. Your new code only check ledger delta. I think it could be useful to check from time to time that DB content is correct.

jonjove: I believe that this is handled by the `CacheIsConsistentWithDatabase` invariant, which was changed in this PR to be checked in `checkOnOperationApply`. The `CacheIsConsistentWithDatabase` invariant checks that the contents of the `LedgerDelta` are also reflected in the database (but not the converse). 

It is possible that there are other changes in the database (for example, made by an external process) that would not be caught by this method. If this is what you want to check for, something that @MonsieurNicolas and I discussed, then this will require more changes in how `CacheIsConsistentWithDatabase` works. But one of the primary design goals for the invariants was to make as few of them access the database as possible, so any additional database checking will likely be added in `CacheIsConsistentWithDatabase`. Let me know if this is what you meant, or if you had something else in mind.

src/invariant/LedgerEntryIsValid.cpp
MonsieurNicolas: first failure is fine: we're logging the offending entry anyways

src/main/CommandHandler.cpp
```diff
@@ -456,6 +457,12 @@ CommandHandler::info(std::string const& params, std::string& retStr)
         info["quorum"] = q["slots"];
     }
 
+    Json::Value invariantFailures = mApp.getInvariantManager().getInformation();
```
MonsieurNicolas: I think this got moved to a different file in a recent commit, otherwise this is fine

src/invariant/InvariantManagerImpl.cpp
```diff
@@ -148,10 +170,13 @@ InvariantManagerImpl::enableInvariant(std::string const& name)
...
 InvariantManagerImpl::onInvariantFailure(
         std::shared_ptr<Invariant> invariant,
-        std::string const& message) const
+        std::string const& message,
+        uint32_t ledger) const
 {
-    mMetricsRegistry.NewCounter({"invariant", "does-not-hold",
+    mMetricsRegistry.NewCounter({"invariant", "does-not-hold", "count",
                                  invariant->getName()}).inc();
+    mMetricsRegistry.NewCounter({"invariant", "does-not-hold", "on-ledger",
```
MonsieurNicolas: I'd prefer to not use metrics as the mechanism to capture a snapshot for "info" (the `on-ledger` metric is just bloating metrics):
The way we propagate status in the history subsystem for example, it by simply saving the last failure in a member variable. Here all you'd need to do is store the `<message, ledger>` pair in a `mLastFailure` member. Doing like this will not tie us up to the limitations of the metrics subsystem while at the same time keeping the door open to storing more than the very last failure.

MonsieurNicolas:
r+ 02e2aba49abc0b6cf41eeeef48f902bdb0c67e5b

MonsieurNicolas:
r+ 22d01ca92b715d4e5c2d618983ba1ec592ddd87a

MonsieurNicolas:
turns out that it was compile errors with older versions of compilers clang-5.0 (and gcc)

MonsieurNicolas:
r+ 4f774f54f050de80c3db5c543db36fac3cea6f62

MonsieurNicolas:
r+ 64b25ddc8ed75e183ff953d09e7a98840c1aabac

PR #1393, opened 2017-11-15
enforce that clang-format was run in PR from travis

### LINKED ISSUE #1388: enforce that clang-format was properly run in travis

MonsieurNicolas:
resolves #1388

like title says

this moves to clang-5.0 (updates the config to the current schema)
and fixes a bug in the Makefile that made it that on some machines it would ignore our custom config.

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -2,48 +2,61 @@
 Language:        Cpp
 # BasedOnStyle:  LLVM
 AccessModifierOffset: -2
-AlignAfterOpenBracket: true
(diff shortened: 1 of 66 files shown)


jonjove:
r+ 724bdfc

PR #1318, opened 2017-09-21
Resolves #1307 - BucketListIsConsistentWithDatabase Invariant

jonjove:
Introduces BucketListIsConsistentWithDatabase invariant which resolves #1307. This invariant implements the newly introduced checkOnBucketApply to ensure that the database matches the bucket, after it was applied, for all ledgers starting from the first ledger considered in the bucket. Merge after #1317

PR #1399, opened 2017-11-17
CATCHUP_MINIMAL Cannot Move Database into Invalid State

### LINKED ISSUE #1300: Specific order of CATCHUP_MINIMAL operations can move database into invalid state
Imagine account GABC that is created on ledger 2000 and removed on ledger 3000.
We can imagine set of buckets for following checkpoints:
* 2017 - set of buckets contains LIVEENTRY for GABC
* 3007 - set of buckets contains LIVEENTRY for GABC and DEADENTRY for GABC
* 10047 - set of buckets does not contain any entry for GABC (buckets got merged in meantime)

This mean that doing CATCHUP_MINIMAL at 2017, then stopping stellar-core and doing next CATCHUP_MINIMAL at 10047 leaves entry for GABC in database even if it does not exists in real ledger.

Fortunately INVARIANT_CHECK_BALANCE enabled in configuration will crash node immediately and wont allow this bug to propagate.

jonjove:
Added a test showing that applying buckets before and after a DEADENTRY cancels a LIVEENTRY cannot move the database into an invalid state. The initial issue was fixed by #1318.

Resolves #1300

### DIFF
--- a/src/bucket/BucketList.cpp
+++ b/src/bucket/BucketList.cpp
@@ -152,6 +152,22 @@ BucketLevel::snap()
     return mSnap;
 }
 
+BucketListDepth::BucketListDepth(uint32_t numLevels) : mNumLevels(numLevels)
+{
+}
+
+BucketListDepth&
+BucketListDepth::operator=(uint32_t numLevels)
+{
+    mNumLevels = numLevels;
+    return *this;
+}
+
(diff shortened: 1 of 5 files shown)


MonsieurNicolas:
r+ 1f13c8478ac6b99aeffef224a9c4ba37375017b4

PR #1401, opened 2017-11-18
Fixed `createTestApplication` that was not using config as expected

MonsieurNicolas:
A previous change attempted to force test applications to use their configuration to update a few ledger header fields.
The problem is that this didn't work as the current ledger was created and stored in the database before updating the fields.

This PR fixes that by taking a different approach: allow changing the genesis ledger in test code to use values from the configuration.

### DIFF
--- a/src/main/Config.h
+++ b/src/main/Config.h
@@ -31,7 +31,7 @@ class Config : public std::enable_shared_from_this<Config>
     std::string expandNodeID(std::string const& s) const;
 
   public:
-    static const int CURRENT_LEDGER_PROTOCOL_VERSION = 9;
+    static const int CURRENT_LEDGER_PROTOCOL_VERSION;
 
     typedef std::shared_ptr<Config> pointer;
 
--- a/src/main/Config.h
+++ b/src/main/Config.h
@@ -105,6 +105,10 @@ class Config : public std::enable_shared_from_this<Config>
     // this should only be enabled when testing as it's a security issue
     bool ALLOW_LOCALHOST_FOR_TESTING;
 
+    // Set to use config file values for genesis ledger
+    // not setable in config file - only tests are allowed to do this
+    bool USE_CONFIG_FOR_GENESIS;
+
     // This is the number of failures you want to be able to tolerate.
     // You will need at least 3f+1 nodes in your quorum set.
     // If you don't have enough in your quorum set to tolerate the level you
--- a/src/test/TestUtils.h
+++ b/src/test/TestUtils.h
@@ -30,10 +30,9 @@ template <typename T = ApplicationImpl>
 std::shared_ptr<T>
 createTestApplication(VirtualClock& clock, Config const& cfg)
 {
-    auto app = Application::create<T>(clock, cfg);
-    auto& lm = app->getLedgerManager();
-    lm.getCurrentLedgerHeader().baseFee = cfg.DESIRED_BASE_FEE;
-    testutil::setCurrentLedgerVersion(lm, cfg.LEDGER_PROTOCOL_VERSION);
+    Config c2(cfg);
+    c2.USE_CONFIG_FOR_GENESIS = true;
+    auto app = Application::create<T>(clock, c2);
     return app;
 }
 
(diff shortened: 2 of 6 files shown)


src/test/TestUtils.h
```diff
@@ -30,10 +30,9 @@ template <typename T = ApplicationImpl>
 std::shared_ptr<T>
 createTestApplication(VirtualClock& clock, Config const& cfg)
 {
-    auto app = Application::create<T>(clock, cfg);
-    auto& lm = app->getLedgerManager();
-    lm.getCurrentLedgerHeader().baseFee = cfg.DESIRED_BASE_FEE;
-    testutil::setCurrentLedgerVersion(lm, cfg.LEDGER_PROTOCOL_VERSION);
+    Config c2(cfg);
```
jonjove: I generally dislike this pattern. `cfg` is being passed by const-reference then immediately copied in the function body. I think it would be better to let `cfg` be passed by value, and remove `c2` entirely. Better to let the compiler do the work for you, in my opinion. (See the next comment before you fix this though)

src/main/Config.h
```diff
@@ -105,6 +105,9 @@ class Config : public std::enable_shared_from_this<Config>
     // this should only be enabled when testing as it's a security issue
     bool ALLOW_LOCALHOST_FOR_TESTING;
 
+    // Set to use config file values for genesis ledger
+    // not setable in config file - only tests are allowed to do this
+    bool USE_CONFIG_FOR_GENESIS;
```
jonjove: Small comment about style, but the whitespace here doesn't respect the convention used in the rest of this file.

src/test/TestUtils.h
```diff
@@ -30,10 +30,9 @@ template <typename T = ApplicationImpl>
...
+    c2.USE_CONFIG_FOR_GENESIS = true;
```
jonjove: Do we actually want to do it this way? Another option that I see is to update `getTestConfig` in test/test.cpp to have `USE_CONFIG_FOR_GENESIS = true` in all test configurations. The advantage to this won't be realized until #1392 is merged, as that PR changes the type of `Application` that is created by `createTestApplication`. It would therefore be ideal if all tests (possibly excluding simulation tests) could go through `createTestApplication`, which will be impossible for the `LedgerHeader` tests the way you implemented it. 

There is still an issue that `createTestApplication` doesn't support `newDb = false` since I didn't think it made sense to create an application with an existing database but then immediately change the ledger version. This means that there are still some places in the tests where we call `Application::create` instead of `createTestApplication` because we need to use an existing database. Your change will eliminate this issue, so I will fix this `newDb` issue in #1392 once this PR is merged.

src/main/Config.h
MonsieurNicolas: which whitespace?

src/test/TestUtils.h
MonsieurNicolas: I thought about that too, I didn't want to introduce a dependency between test configuration (that works in all scenarios) vs what a test application is as we're not creating test applications all the time (simulation is just one of those places like you pointed out).

#1392 should be merged in days, so I am not sure making additional changes to the code to support the interim state really makes sense.
Trying to switch all test cases to use test application is not in scope for this PR that was only looking at fixing the behavior of that factory function.

src/main/Config.h
jonjove: I realize my previous comment wasn't specific enough, sorry. I was referring to the lack of a blank line between lines 110 and 111. The rest of this file has (mostly) followed the convention of configuration description, configuration variable definition, blank line, repeat.

src/test/TestUtils.h
jonjove: Alright, I am comfortable to accept this solution in the interim since it does resolve the issue that this PR is addressing. But it might become apparent when I do finish going through the process of changing to `createTestApplication` everywhere that we need to do something different. We can handle that when we come to it though. Given that we will do it this way for now, it would be good to address the comment directly above though.

MonsieurNicolas:
updated PR - missing <CR> in config file

jonjove:
r+ 44213f0

PR #1402, opened 2017-11-20
Upgrade fixes

### LINKED ISSUE #1367: Make DESIRED_BASE_RESERVE configurable
Right now `DESIRED_BASE_RESERVE`'s value is hardcoded in the internal config object.

Instead, we should make it configurable (similar to `DESIRED_BASE_FEE`), and have validators vote on its value.

### LINKED ISSUE #1368: PREFERRED_UPGRADE_DATETIME is not properly respected

### LINKED ISSUE #1369: Expose drifting config to validator operator
Right now a validator will vote for making changes to the network according to its config file (for example, to change the base fee).

The problem is that if an operator is not aware that a change was made by other validators, its node will vote for its local config value in the background even though it may not be what the operator wants.

We should make it obvious to a node operator that there is a difference between the node's config file and the network's opinion:
* we should log that the node is voting for a change
* we should display in `info` that the node is not configured like the rest of the network (and voting for a change)

### LINKED ISSUE #1370: Change PREFERRED_UPGRADE_DATETIME to apply to all ledger header changes
Right now the setting is only used to schedule protocol updates.
Instead it should apply to any change that requires validator operators to agree to a change as arguably fields like `DESIRED_BASE_FEE` and `DESIRED_BASE_RESERVE` have a lot more impact on the network than a protocol version upgrade.

This will close a loophole: right now the last validator that votes for a change gets to control the timing of a network wide behavior change which may have implications beyond the network.

vogel:
Fixes #1367
Fixes #1368 
Fixes #1369
Fixes #1370


src/herder/UpgradesTests.cpp
```diff
@@ -367,6 +367,19 @@ TEST_CASE("simulate upgrades", "[herder][upgrades]")
                               LedgerManager::GENESIS_LEDGER_BASE_FEE * 4,
                               LedgerManager::GENESIS_LEDGER_MAX_TX_SIZE * 4};
 
+    SECTION("oscillation")
```
MonsieurNicolas: I imagine that you won't include this commit in the final PR as the next commit is making this irrelevant?

src/herder/UpgradesTests.cpp
```diff
@@ -282,48 +253,42 @@ testValidateUpgrades(VirtualClock::time_point preferredUpgradeDatetime,
...
-                                      toUpgradeType(makeBaseFeeUpgrade(200)),
+                                      toUpgradeType(makeBaseFeeUpgrade(100)),
                                       ledgerUpgradeType));
     }
 
     SECTION("too small fee")
     {
         REQUIRE(!Upgrades{cfg}.isValid(checkTime,
-                                       toUpgradeType(makeBaseFeeUpgrade(49)),
+                                       toUpgradeType(makeBaseFeeUpgrade(99)),
```
MonsieurNicolas: you should collapse those tests: the only thing we're testing is that it's different

src/herder/Upgrades.cpp
```diff
@@ -69,18 +69,13 @@ Upgrades::isValid(uint64_t closeTime, UpgradeType const& upgrade,
     case LEDGER_UPGRADE_BASE_FEE:
     {
         uint32 newFee = lupgrade.newBaseFee();
-        // allow fee to move within a 2x distance from the one we have in our
-        // config
-        res = (newFee >= mCfg.DESIRED_BASE_FEE * .5) &&
-              (newFee <= mCfg.DESIRED_BASE_FEE * 2);
+        res = (newFee == mCfg.DESIRED_BASE_FEE);
```
MonsieurNicolas: you're missing another change in this PR: moving to using `timeForUpgrade` for all changes. Right now it's only enforced for `LEDGER_PROTOCOL_VERSION`, which means that nodes would upgrade fees etc before upgrading the protocol version.

MonsieurNicolas: this may also simplify some tests depending on the sequencing of commits in the PR

vogel: There was different issue for that one, so I made separate PR with that.

vogel:
Updated with suggestions and merged with 2 other upgrades PRs

src/ledger/LedgerManagerImpl.h
```diff
@@ -86,9 +86,7 @@ class LedgerManagerImpl : public LedgerManager
     uint64_t secondsSinceLastLedgerClose() const override;
     void syncMetrics() override;
 
-    void startNewLedger(uint32_t protocolVersion, int64_t balance,
-                        uint32_t baseFee, uint32_t baseReserve,
-                        uint32_t maxTxSetSize);
+    void startNewLedger(LedgerHeader genesisLedger);
```
MonsieurNicolas: use a `const&` as there is no reason to copy this object around

src/herder/Upgrades.cpp
```diff
@@ -74,6 +74,24 @@ Upgrades::applyTo(LedgerUpgrade const& upgrade, LedgerHeader& header)
     }
 }
 
+std::string
+Upgrades::toString(LedgerUpgrade const& upgrade)
+{
+    switch (upgrade.type())
+    {
+    case LEDGER_UPGRADE_VERSION:
+        return fmt::format("VERSION={0}", upgrade.newLedgerVersion());
```
MonsieurNicolas: I think we now use `PROTOCOL_VERSION` now that the string is going to be used in various context (so that it matches what we use in Config for all fields)

src/herder/HerderImpl.cpp
```diff
@@ -762,6 +765,25 @@ HerderImpl::triggerNextLedger(uint32_t ledgerSeqToTrigger)
         }
     }
 
+    if (!upgrades.empty())
+    {
+        auto message =
+            fmt::format("Proposing herder configuration upgrades: {0}; network "
```
MonsieurNicolas: it would be cleaner to either use the same names than `Upgrades::toString` (for example `BASE_FEE` instead of `baseFee`) or better yet, use `Upgrades::toString` on the network values.

src/herder/HerderImpl.cpp
```diff
@@ -762,6 +765,25 @@ HerderImpl::triggerNextLedger(uint32_t ledgerSeqToTrigger)
...
+                        "values for herder are: ledgerVersion={1}, "
+                        "baseFee={2}, baseReserve={3}, maxTxSetSize={4}",
+                        Upgrades::toString(upgrades), lcl.header.ledgerVersion,
+                        lcl.header.baseFee, lcl.header.baseReserve,
+                        lcl.header.maxTxSetSize);
+        CLOG(WARNING, "Herder") << message;
```
MonsieurNicolas: I think it should be `INFO` (INFO > WARNING), we want this to be nagging/highly visible

src/ledger/LedgerManagerImpl.h
vogel: Ledger delta constructor requires non-const LedgerHeader and it is used in startNewLedger

vogel:
upgrade with Nicolas suggestions

src/ledger/LedgerManagerImpl.cpp
```diff
@@ -212,7 +212,7 @@ LedgerManagerImpl::startNewLedger()
         ledger.maxTxSetSize = cfg.DESIRED_MAX_TX_PER_LEDGER;
     }
 
-    startNewLedger(ledger);
+    startNewLedger(std::move(ledger));
```
MonsieurNicolas: I think it's clearer to just have `startNewLedger` take a const reference: the only reason to pass around by value is if the function is going to make a copy anyways for some reason, which is not the case here.

MonsieurNicolas: nm - didn't see your other comment on the same issue that ledgerDelta requires a non const

MonsieurNicolas:
if you want I can help with the documentation bits @vogel  - let me know

vogel:
I've added some documentation. Not sure if that is enough.

MonsieurNicolas:
r+ a2d34589b53b94323aee9d8503cc18e021fa989e

PR #1407, opened 2017-11-20
Desired base reserve configuration option

### LINKED ISSUE #1367: Make DESIRED_BASE_RESERVE configurable

vogel:
Fix #1367


swetakedia:
@vogel - Is this fixed? If yes, which version we need to upgrade. Also it will be helpful if you can share the steps to upgrade a stellar core - when built using the source code & not docker.

Note: I am at present using  "build" : "v9.0.0-2-g51aa007-dirty"

I want to update core so that the base_reserve is 05 XLM instead of 20 XLM.

MonsieurNicolas:
The network is already running with 0.5 XLM.
you can read the release notes for 9.0.1 that covered this. https://github.com/stellar/stellar-core/releases/tag/v9.0.1

In general you should aim to run the latest stable release (currently 9.1.0)

Issue #1309, opened 2017-09-14
Write invariant to validate LedgerDelta against SQL


MonsieurNicolas:
Currently we check that all changes included in the LedgerDelta are consistent with the state in SQL with the invariant `CacheIsConsistentWithDatabase`.

We should also check that changes in SQL map to LedgerDelta entries. If done well, this could also allow to catch real time invalid updates to SQL done from outside the core process.

MonsieurNicolas:
this was made obsolete by `LedgerState`

PR #1413, opened 2017-11-27
Expose drifting configuration

### LINKED ISSUE #1369: Expose drifting config to validator operator

vogel:
This one depends on #1402
Fixes #1369


MonsieurNicolas:
not sure I understand this PR: the commit that takes care of #1309 is already in #1402 but this PR has additional commits related to upgrade vs catchup

PR #1415, opened 2017-11-27
Report last history checkpoint

vogel:
First part of implementation of parallel catchup tests.

### DIFF
--- a/src/main/main.cpp
+++ b/src/main/main.cpp
@@ -14,6 +14,7 @@
 #include "crypto/SecretKey.h"
 #include "database/Database.h"
 #include "history/HistoryManager.h"
+#include "historywork/GetHistoryArchiveStateWork.h"
 #include "ledger/LedgerManager.h"
 #include "lib/http/HttpClient.h"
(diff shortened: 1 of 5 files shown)


src/main/main.cpp
```diff
@@ -347,6 +356,48 @@ catchupTo(Config const& cfg, uint32_t to)
...
+        auto state = HistoryArchiveState{};
+        auto& wm = app->getWorkManager();
+        auto getHistoryArchiveStateWork =
+            wm.addWork<GetHistoryArchiveStateWork>(
+                "get-history-archive-state-work", state);
+        wm.advanceChildren();
+        while (clock.crank(true) && !wm.allChildrenDone())
+            ;
+
+        auto ok = getHistoryArchiveStateWork->getState() == Work::WORK_SUCCESS;
```
MonsieurNicolas: I think that it's probably better to output this to a file to make parsing by scripts easier.

MonsieurNicolas:
r+ 9738e580d685104a2444c7cca3457a95ab63d624

Issue #1418, opened 2017-11-28
Improve our command-line interface and code


vogel:
Current support for command-line arguments is all put into main.cpp as big spaghetti code. We should fix that and also improve our output.

What I think needs to be done:
* there is need for a better command-line-arguments parsing library, I vote for Clara for Catch2, as we are already using Catch and upgrade to Catch2 will come in future
* separate command-line handling code (extract class or a whole directory for them)
* look at making input/output consistent between commands
* ensure that stdout contains only what we are printf/couting and stderr contains all the logging
* allow graphquorum and other commands to output to stdout
* check if each command returns proper exit code to OS
* check that commands that do not use DB do not require it
* check that commands are only initializing what is needed (there is no need to setup full Application each time)
* ... probably more

MonsieurNicolas:
I don't think we'd be addressing all of those issues in the same PR.

I do like moving to Clara - not because it's used in Catch2 but because it looks very simple is c++11 compliant, is a single header and allows to do a first pass at cleaning things up. We should open an issue just for that.

I don't know if there is a dependency on this for using stdout exclusively for command results (which would allow piping outputs). I think we should open an issue also for this, I imagine that as part of this - we'd be moving to output formats that are either human readable or machine readable (using json for example).

The other changes, I don't have an opinion yet as I don't see pros/cons.

MonsieurNicolas:
Now that Catch2 is in the tree, we can probably overhaul the command line parsing.

Ideally we would do something similar to what git's command line does and give each "subcommand" its own options, so we could do things like:
`./stellar-core -catchup -from X -to Y`

Next step for this issue is to actually write the list of commands and options for each.

vogel:
After looking at out command line implementation I've learned that:
- in reality most of our command line arguments are de facto "subcommands"
- only few items are proper arguments:
  - conf
  - base64
  - output-file
  - metric
  - ll (for log level)
  - netid (used only in signtxn)
- then there is a lot of very simple "subcommands"
  - convertid
  - checkquorum
  - dumpxdr
  - printtxn
  - signtxn
  - loadxdr
  - fuzz
  - genfuzz
  - genseed
  - graphquorum
  - help
  - inferquorum
  - offlineinfo
  - report-last-history-checkpoint
  - sec2pub
  - forcescp (it was suprising, I was expecting this to be an argument)
  - newdb (as above)
  - newhist (as above)
  - version
- and only a few more complicated ones:
[...]

PR #1422, opened 2017-11-29
Output catchup result

vogel:


src/main/main.cpp
```diff
@@ -340,28 +342,43 @@ catchup(Config const& cfg, uint32_t to, uint32_t count)
...
+catchupTo(Config const& cfg, uint32_t to, Json::Value& catchupInfo)
 {
-    return catchup(cfg, to, std::numeric_limits<uint32_t>::max());
+    return catchup(cfg, to, std::numeric_limits<uint32_t>::max(), catchupInfo);
+}
+
+static void
+writeCatchupInfo(Json::Value const& catchupInfo, std::string const& outputFile)
+{
+    std::string filename = outputFile.empty() ? "catchupInfo.json" : outputFile;
```
MonsieurNicolas: I don't know if this makes sense: I'd rather have a behavior consistent across all command lines that use "--output-file".
Something in the lines of a common function that takes a string and where the default filename is "-" (ie stdout) - so if people want to output to a file, they have to specify a file name. I know that breaks the current contract for "graphquorum" but it seems worth it instead of having different implementations and behaviors for different command line options.

vogel: Behavior of `--output-file` is consistent:

writeCatchupInfo:
```
std::string filename = outputFile.empty() ? "catchupInfo.json" : outputFile;
```

reportLastHistoryCheckpoint:
```
std::string filename =
            outputFile.empty() ? "lasthistorycheckpoint.json" : outputFile;
```

writeQuorumGraph:
```
std::string filename = outputFile.empty() ? "quorumgraph.dot" : outputFile;
```

Currently we cannot support `-` as stdout, as all of our logging information goes to stdout instead of stderr (I've covered that in #1418). Should I do #1418 first so we can implement `-`?

MonsieurNicolas: what I am saying is that the behavior is not consistent as the default filename depends on the command (so you can't output the default filename in the help for `--output-file` and you can't refactor the `output-file` functionality into its own function).

I don't think there is a dependency on #1418 for this:
what I am suggesting is that when people run from a shell, outputting the result to stdout allows people to see what the command did and if they want to use the output for something else, they can just use the `--output-file` option to a temp file for example.

Cleaning up stdout vs stderr will allow people to pipe commands (and not use a temp file), which is a different thing.

vogel: So what should I do here? Implement #1427 and add support for `-`? Change default name of file for output to output.json for all commands? Or add LOG output for all of them?

MonsieurNicolas:
r+ 2c4e74c29ecbbf931fb91d302f1593763f618f65

vogel:
bors: retry

PR #1433, opened 2017-12-04
Catchup at twice

### LINKED ISSUE #1404: Doing --catchup-at twice with the same non-checkpoint argument can damage retained buckets
To reproduce:

```
./stellar-core --newdb
./stellar-core --catchup-at 10070000
./stellar-core --catchup-at 10070000
```
It seems that some buckets are not properly retained at second run (which should do nothing about buckets).

vogel:
Fixes #1404

### DIFF
--- a/src/bucket/BucketManagerImpl.cpp
+++ b/src/bucket/BucketManagerImpl.cpp
@@ -351,6 +351,13 @@ BucketManagerImpl::assumeState(HistoryArchiveState const& has)
         mBucketList.getLevel(i).setNext(has.currentBuckets.at(i).next);
     }
     mBucketList.restartMerges(mApp);
+    for (uint32_t i = 0; i < BucketList::kNumLevels; ++i)
+    {
+        if (mBucketList.getLevel(i).getNext().hasOutputHash())
+        {
+            mBucketList.getLevel(i).getNext().resolve()->setRetain(true);
+        }
+    }
 }
 
 void
--- a/src/bucket/BucketManagerImpl.cpp
+++ b/src/bucket/BucketManagerImpl.cpp
@@ -335,6 +335,19 @@ BucketManagerImpl::checkForMissingBucketsFiles(HistoryArchiveState const& has)
 }
 
 void
+BucketManagerImpl::retainAll(HistoryArchiveState const& has)
+{
+    for (auto const& bucket : has.allBuckets())
+    {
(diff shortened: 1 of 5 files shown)


src/bucket/BucketManagerImpl.cpp
```diff
@@ -351,6 +351,13 @@ BucketManagerImpl::assumeState(HistoryArchiveState const& has)
         mBucketList.getLevel(i).setNext(has.currentBuckets.at(i).next);
     }
     mBucketList.restartMerges(mApp);
+    for (uint32_t i = 0; i < BucketList::kNumLevels; ++i)
```
MonsieurNicolas: 1. why do we need to do this?
2. if it does, I think you may have a race as this is basically racing with the merges that `restartMerges` started

vogel:
new version should be much more clear

src/bucket/BucketManagerImpl.cpp
```diff
@@ -350,14 +363,10 @@ BucketManagerImpl::assumeState(HistoryArchiveState const& has)
         mBucketList.getLevel(i).setSnap(snap);
         mBucketList.getLevel(i).setNext(has.currentBuckets.at(i).next);
     }
+
+    // we will need these buckets after restart
+    retainAll(has);
```
MonsieurNicolas: this should no-op... The loop right before is supposed to do just that:
```
mBucketList.getLevel(i).setCurr(curr);
mBucketList.getLevel(i).setSnap(snap);
```
where `setCurr` and `setSnap` are calling `setRetain(true)` on basically all buckets in `has`

MonsieurNicolas: oh I see there are the future buckets in there as well

MonsieurNicolas: mmm, it looks like it's the same change than what you did in `LedgerManagerImpl::storeCurrentLedger()` (so logic should be shared).

I think this change will fix this specific issue, as GC will take care of things, but it seems we're chasing a larger underlying problem and playing wack-a-mole.

Is it maybe that `restartMerges` only marks `retain(true)` for buckets involved in new merges but doesn't actually do anything if a merge was already completed?

MonsieurNicolas:
r+ fc73e2a1f2dc5a841c464a339f31e71749a3e6eb

PR #1434, opened 2017-12-04
Upgrade basenh

### LINKED ISSUE #1424: Upgrade basen.h library
Upgrade basen library from `https://github.com/azawadzki/base-n/blob/master/include/basen.hpp`.
New version does not use undefined behavior in extract_partial_bits and extract_overlapping_bits,

vogel:
Fixes #1424

I'm not 100% happy with that one. I'll try to get 2 fix commits into main basen.h repository.


MonsieurNicolas:
yeah it looks like a lot of change for not much benefits (I am worried that there are 0 tests in upstream)

vogel:
I'll look for alternative implementations.

vogel:
It looks like the only other viable option is https://github.com/tplgy/cppcodec (or I can just hack our current basen with few fixes for UB).

cppcodec:
* is MIT licenced
* is header-only
* has tests (using Catch)
* has nice API

MonsieurNicolas:
this other one comes with too many things - basen is simple. I think that we should just contribute a few unit tests to that project

vogel:
I propose something different:
1. first add tests for API that we use in stellar-core
2. hide that API behing some wrapper (similar to BasenWrapper.h, like Codecs.h) and do not use basen directly (wrapper can be inline, so it won't add to runtime)
3. upgrade basen.h to newest versions (we will have our tests then, so it won't break)
4. submit our tests and fixes to basen.h upstream
5. even if it gets accepted, leave them in stellar-core just in case we switch our implementation sometime

That way we can update/change codecs library whenever we wish.

What do you think?

MonsieurNicolas:
yeah that sounds like a plan, I think the only test we have right now is
`TEST_CASE("base64 tests", "[crypto]")`
which uses random data, we may want to add a small corpus of test data, to test various byte length from 1 byte to `n` with `n=7` for buffer.

MonsieurNicolas:

vogel:
Updated with new approach

lib/util/basen.h
```diff
@@ -54,165 +54,164 @@ void decode_b64(Iter1 start, Iter1 end, Iter2 out);
...
 void decode(Iter1 start, Iter1 end, Iter2 out)
 {
     Iter1 iter = start;
-    int output_current_bit = 0;
+    size_t output_current_bit = 0;
     char buffer = 0;
-    int trailing_equal = 0;
 
     while (iter != end) {
-        if (*iter == '=') {
```
MonsieurNicolas: I don't understand why this passes tests: if we do not deal with the trailing '=' explicitly, when are we writing to the `out` buffer? It seems we would just trigger the error condition that just skips the rest of the loop

vogel: It just ignores it in the if (value == Error) part, and then continues to the next iteration.

Previous version does almost the same with the difference than trailing_equal is computed and then used on line 200.

That line does not exist in upstream (and never existed), so it must have been added by you in 860ef61989b955a0f147e797d8895635a2fc65b5. If so, then we need tests to cover case where it is executed and reintroduce these changes here (and make PR upstream).

MonsieurNicolas:
@vogel finally got to this one - things seem fine. You need to rebase it though

MonsieurNicolas:
r+ c3ab30e857d0afb6fe0cb70b32b15c09051c5bab

MonsieurNicolas:
r+ c9ae0227ab560a21f68ee9ffbebd496034d5568c

Issue #1295, opened 2017-08-31
Investigate large number of buckets in bucket directory


s-a-y:
Currently all bucket files are stored in 1 flat directory:
```
/buckets
    bucket-002d4e82245617c72c5eedbdd7ebda6a76d0ccf4bc714fd3eccb57e4cd7dd5e1.xdr  
    bucket-54789b62f64896e05aa457d743e9a4a08ed3f665a21a56c578955fd646e521c4.xdr  
    bucket-a986b1bcc9fd737ab2dd529b86db957ada76fcb66bac758dbfb19c76fe242bbf.xdr
    bucket-004106780e8ff5466dbaa6b99a95f373587abc176808c0c9b26194ed47da54cf.xdr
```
On most filesystems (except maybe ext4) it causes performance issues when you have 10,000 files and more. It just blows away my terminal when I'm trying to list files or calculate storage.

We could use subdirectories to improve situation, similar to what varnish is doing:
```
/buckets/0/0/2/d4e82245617c72c5eedbdd7ebda6a76d0ccf4bc714fd3eccb57e4cd7dd5e1.xdr  
/buckets/0/0/4/106780e8ff5466dbaa6b99a95f373587abc176808c0c9b26194ed47da54cf.xdr
/buckets/5/4/7/89b62f64896e05aa457d743e9a4a08ed3f665a21a56c578955fd646e521c4.xdr  
/buckets/a/9/8/6b1bcc9fd737ab2dd529b86db957ada76fcb66bac758dbfb19c76fe242bbf.xdr
```

s-a-y:
that was full validator. 
And yes, I found that your history archives do have directory split, so I guess this issue can be closed. You can open separate one for garbage collector, but I don't have anymore information on that at the moment.

MonsieurNicolas:
to be clear, I was talking about the local `Buckets` folder: I can see a situation, for example when the node is running in `CATCHUP_COMPLETE` mode (replaying all of history), which creates a huge backlog in the publish queue - then I think we'd create all those buckets in the same folder. So this may be a legitimate concern.

s-a-y:
Yes, that's exactly my use-case, since I was struggling to setup publishing to archive and buckets folder became too big, and eventually I had to throw away these nodes, because it just stopped working.

MonsieurNicolas:
I found that some of our own nodes have way too many bucket files in that folder. We'll look at this in more detail in our next release.

MonsieurNicolas:
[...]

PR #1440, opened 2017-12-07
reduce number of buckets retained during catchup

vogel:

This fixes part of #1295


vogel:
We not longer need to keep from removing unreferenced buckets during catchup, for 2 reasons:
1. removing unreferenced buckets is now much faster, as it does not go to database each time (lots of data is cached)
2. catchup itself is less prone to failures

MonsieurNicolas:
r+ 3f5d5fd49b065ea0df846fe837880677b92df50c

PR #1441, opened 2017-12-07
always emit previous state in meta

### LINKED ISSUE #1420: change the meta data format to be easier to process
Right now the meta data format exported in `txhistory` and `txfeehistory` (used by Horizon for example), requires complex logic to import state as changes are "linked" between operations and transactions:
for example, with a txset = [tx1, tx2], if tx1 changes a ledger entry A, tx1's meta will have both the "before" (version N) and "after" (version N+1) value of A (as modified by tx1), and if tx2 also modifies A its meta data will only have the new value (version N+2) as version N+1 was created in the same transaction set.
Instead, we should attach the before and after values for each transaction and operation: while it can be slightly less efficient (from a disk space point of view), it makes the logic outside of core extremely simple (and therefore less likely to be wrong).

MonsieurNicolas:
this simplifies ingestion logic in consumers of the meta data; old version required to keep track of updates over transactions and operations

resolves #1420

### DIFF
--- a/src/ledger/LedgerDelta.cpp
+++ b/src/ledger/LedgerDelta.cpp
@@ -262,13 +262,9 @@ LedgerDelta::addCurrentMeta(LedgerEntryChanges& changes,
     auto it = mPrevious.find(key);
     if (it != mPrevious.end())
     {
-        // if the old value is from a previous ledger we emit it
         auto const& e = it->second->mEntry;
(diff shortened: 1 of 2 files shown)


vogel:
r+ e45546ced97a40ebc1c79f3e5e3621e3b4d033be

jonjove:
r+ fe955a9

PR #1447, opened 2017-12-09
Overlay connections fix

### LINKED ISSUE #1278: Better system for preferred peers
Should be able to enter in an IP and a public key for preferred peers in the .cfg

Should prioritize connecting to the preferred peers. 

Also investigate how the peer list is being handled right now. It seems like the old peers can clog up a server and prevent it from joining the network for a long time.

MonsieurNicolas:
resolves #1278 

This PR fixes a few problems with the overlay code:
* we were not loading enough peers from the peers table, causing the node to sometimes not attempt to make more connections
* preferred peers may get buried behind other peers, causing the previous logic to not see them

the first problem is taken care of by loading batches of peers until enough new peers are loaded from the peers table.

the second problem is solved by forcing preferred peers to have a next attempt time of 0 (epoch), which puts them at the top of the list


MonsieurNicolas:
closing this; I am replacing it with just the changes for preferred peers

PR #1449, opened 2017-12-10
Preferred peers fix

### LINKED ISSUE #1278: Better system for preferred peers

MonsieurNicolas:
resolves #1278 

Preferred peers may get buried behind other peers, causing the previous logic to not see them.

This PR fixes that problem by forcing preferred peers to have a next attempt time of 0 (epoch), which puts them at the top of the list


vogel:
r+ 1add7a9

vogel:
@latobarita: retry

vogel:
@latobarita: retry

Issue #1385, opened 2017-11-14
Alpine: build fails


otokarev:
To reproduce run (`docker build ` with [Dockerfile](https://gist.github.com/otokarev/c3455ac8c0c07d88848578930e4ecff7)):
```
curl https://gist.githubusercontent.com/otokarev/c3455ac8c0c07d88848578930e4ecff7/raw/dcfc085a7fa2b61590b7eeff32e2cbbe3e10eabd/Dockerfile.alpine | docker build -
```
it will result in:
```
 In file included from ../lib/asio/include/asio/impl/src.hpp:23:0,
                 from asio/src/asio.cpp:61:
../lib/asio/include/asio/impl/error_code.ipp: In member function 'virtual std::__cxx11::string asio::detail::system_category::message(int) const':
../lib/asio/include/asio/impl/error_code.ipp:110:22: error: could not convert 'strerror_r(value, ((char*)(& buf)), sizeof (buf))' from 'int' to 'std::__cxx11::string {aka std::__cxx11::basic_string<char>}'
     return strerror_r(value, buf, sizeof(buf));
            ~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~

```

PR #1451, opened 2017-12-10
Replace asio with submodule

santegoeds:
Fixes issue #1385

The Alpine build fails due to a bug in dependency `lib/asio`.

The issues was raised and fixed in the upstream repo, see https://github.com/chriskohlhoff/asio/issues/94

This PR replaces the `lib/asio` dependency with a submodule and fixes the build scripts.

So far I've only been able to verify the build on Linux (including Alpine) but I do have access to a Windows machine to should be able to test that one once I have installed Visual Studio.

I don't have access to OSX so would appreciate it if someone can check there.


MonsieurNicolas:
Glad you found the commit that fixes this issue!
I was going to skim though the changes myself to see if anything like this was there.
Let me see when we can get this merged as it will require deeper testing as asio is quite central to stellar-core.

MonsieurNicolas:
Also, I think that you should instead point asio to https://github.com/chriskohlhoff/asio/releases/tag/asio-1-10-8 which is the latest stable release. It includes the fix we're looking for.

MonsieurNicolas:
Anybody with a Mac can confirm that this works for them? Build & tests pass @jonjove maybe?

jonjove:
Confirmed that it builds and tests pass on Mac.

MonsieurNicolas:
I didn't forget about this change - I think it uncovered a bug, at least on Windows. I'll report back here when I am done with the investigation.

santegoeds:
Thanks for the heads-up. :-) I thought you might be waiting for me to confirm it was working on Windows.

I've been trying to build it there but have had little spare  time and didn't manage to build xdrpp yet (a brief attempt using msys2 failed)

MonsieurNicolas:
I didn't have time to try this, but there seems to be something missing/broken wrt defines.
If you look at https://github.com/stellar/stellar-core/blob/master/lib/asio/src/asio.cpp we modified it to match https://github.com/stellar/stellar-core/blob/master/src/util/asio.h but when moving to this version, the changes in `asio.cpp` got dropped, so we end up with a potential mismatch.

I don't know if this is Windows specific - I suspect we should move all those defines to a common place:
* Makefile.am seems error prone as we need to keep the visual studio project in sync (but maybe we don't have the choice?). so we'd end up with two "common places".
* Move those defines to a `asio-defines.h` file that is included in both our version of `asio.h` and in a new cpp file that includes that `asio-defines.h` and `asio.cpp` (so we would not compile directly the asio.cpp in `asio/src/`). We can also remove all asio related defines from the makefile/project files.

MonsieurNicolas:
Also, can you acknowledge the "Contributor License Agreement"? https://github.com/stellar/stellar-core/blob/master/CONTRIBUTING.md#submitting-changes

santegoeds:
Signed...

MonsieurNicolas:
thanks! no problem on timeline. I don't know if working upstream will work as asio seems to be very slow wrt release cadence. So I would say: keep the change to what we can control :)

santegoeds:
Sure.  I had a very quick look at the changes and afaics it's only to make sure some defines exist.

Ideally I would have thought that these would be build parameters but I understand it's a bit of a pain to keep two build systems in sync.

MonsieurNicolas:
yeah like I said the "easy" way might be to stop having those defines in header files like this and just move them into their respective build environment (Makefile and vsproj). The ideal way is that we only have one version of those defines. I don't know if those are automagically defined during the autogen step (when using the Makefile), if so the list of duplicates might be even smaller (like the ones that we already have in the Makefile)

santegoeds:
Apologies for the slow progress on this.

There is only a single object file that's included in `lib3rdparty.a`, so I've made a small change to build it from `lib/asio.cpp` which includes both the custom defines and the original `asio.cpp` from the submodule.

This should make sure that all object files are build using the same `#define`.  Please let me know if I've missed something or if you prefer another solution.

Still haven't tested it on Windows, unfortunately.

MonsieurNicolas:
thanks, I think this will work. I'll give it a try on Windows a little later if you don't get to it

MonsieurNicolas:
we've released 9.2.0, so this is good to go. Thanks again, I know it's not easy to contribute on the side

MonsieurNicolas:
r+ ba1bda40ce4ae14e64a7ea2dc0a946e8f29f5a39

PR #1455, opened 2017-12-13
Fix for inability for nodes to reach full connectivity

### LINKED ISSUE #1450: Nodes ignore nodes from their known nodes table
It looks like the logic for loading peers from the known peers table is wrong:
it only loads the first few entries, which may already be connected peers.

Solution is to probably stream peers from that table until the desired amount of new peers is reached.

Interestingly, this may cause nodes to fill up too efficiently and cause nodes that restart to be "locked out" of the network.

MonsieurNicolas:
fixes #1450 

This PR fixes a problem with the way nodes connect to other nodes.

In the current implementation, we may load from the peers table only peers that are currently connected, which causes the logic to connect to more peers to not do anything.
With this change, the node connects to the expected number of peers.

This PR also includes a change to the way `MAX_PEER_CONNECTIONS` is defined.

The setting is being replaced by a relative number `MAX_ADDITIONAL_PEER_CONNECTIONS` instead as the previous setting was quite error prone:
* people only setting `TARGET_PEER_CONNECTIONS` may not realize that `MAX_PEER_CONNECTIONS` would limit the number of connections the node could achieve
* it's a desirable property for the network to have a good margin between `TARGET_PEER_CONNECTIONS` and `MAX_PEER_CONNECTIONS` as this allows for nodes to connect to other nodes. Exposing it as `MAX_ADDITIONAL_PEER_CONNECTIONS` makes it clear that this is a buffer number.

In order to reach an appropriate buffer number, the default value for `MAX_ADDITIONAL_PEER_CONNECTIONS` may increase the effective value of `MAX_PEER_CONNECTIONS`.

### DIFF
--- a/docs/stellar-core_example.cfg
+++ b/docs/stellar-core_example.cfg
@@ -79,14 +79,25 @@ NETWORK_PASSPHRASE="Public Global Stellar Network ; September 2015"
 PEER_PORT=11625
 
 # TARGET_PEER_CONNECTIONS (Integer) default 8
-# This server will send outbound connection attempts until it is at this
+# This controls how aggressively the server will connect to other peers.
+# It will send outbound connection attempts until it is at this
 #   number of peer connections.
 TARGET_PEER_CONNECTIONS=8
 
+# MAX_ADDITIONAL_PEER_CONNECTIONS (Integer) default -1
+# Numbers of peers above the target that can connect to this instance
+# This gives room for other peers to connect to this instance.
+# Setting this too low will result in peers stranded out of the network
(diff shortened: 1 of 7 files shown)


MonsieurNicolas:
@vogel @jonjove can you guys review this?

vogel:
r+ 4887d82d337ed761928fb1a9be0b47d1c68a57b4

vogel:
@latobarita: retry

MonsieurNicolas:
@latobarita: retry

MonsieurNicolas:
@latobarita: retry

PR #1456, opened 2017-12-13
allow publishing to multiple history archives at once

### LINKED ISSUE #1457: Publishing to 2 archives does not work
When 2 history archives are configured with "put" command stellar-core crashes during doing its first publish.

vogel:
Allow publishing to multiple history archives at once.

Before that change creating second and next PutSnapshotFilesWork failed because PublishWork already had a subwork with name "put-snapshot-files".

Fixes #1457


MonsieurNicolas:
r+ 1b11594f44dc681057b9b31f377a5408f4a502a8

PR #1458, opened 2017-12-13
Clean up .xdr.gz bucket files after publish

vogel:
This is first non-perfect version of code that will save a lot of space in directories used by stellar-code.

What I like about that patch:
* extremely simple
* does its job

What I don't like about it:
* removing code does not correspond with creation code (different place, different scope)

Should be good to apply for now, when progress on the real patch is being made.

### DIFF
--- a/src/bucket/Bucket.cpp
+++ b/src/bucket/Bucket.cpp
@@ -68,6 +68,8 @@ Bucket::~Bucket()
     {
         CLOG(TRACE, "Bucket") << "Bucket::~Bucket removing file: " << mFilename;
         std::remove(mFilename.c_str());
+        CLOG(TRACE, "Bucket") << "Bucket::~Bucket removing file: " << mFilename << ".gz";
+        std::remove((mFilename + ".gz").c_str());
     }
 }
 


MonsieurNicolas:

MonsieurNicolas:
I think this change needs to be moved to https://github.com/stellar/stellar-core/pull/1459/

MonsieurNicolas:
mmm, actually I think it makes more sense to instead build a list of buckets based on the content of the buckets folder (both compressed and not compressed), and leverage the code in `forgetUnreferencedBuckets` to remove the ones that are not referenced.
Ideally we would just have added those buckets on startup to mSharedBuckets - but the current code doesn't seem to support having mSharedBucket reference compressed buckets

vogel:
It would be nice to have C++17 support for `std::filesystem`. I've not found nice portable way of listing directory contents before that.

MonsieurNicolas:
I think adding a simple function to return the list of files under a given path in fs.h, and have the implementation use `dirent.h` (POSIX header) should work. When you have a version, I can figure out how to make it work on Windows (on Windows I may be able to use `std::filesystem`, otherwise I'll use something like https://github.com/tronkko/dirent/tree/1.23.1 )

vogel:
#1537 is a much better solution for this

PR #1459, opened 2017-12-13
Invert set retain

vogel:
setRetain logic is now inverted - BucketManager checks if bucket should be retained at the only 2 places that did setRetain(true) before - in current HistoryArchiveState and in publish queue table in database.

This is much simpler solution.

Additionaly, I've moved bucket iterator classes to separate files.

### DIFF
--- a/src/bucket/BucketInputIterator.cpp
+++ b/src/bucket/BucketInputIterator.cpp
@@ -1,0 +1,66 @@
+// Copyright 2017 Stellar Development Foundation and contributors. Licensed
+// under the Apache License, Version 2.0. See the COPYING file at the root
+// of this distribution or at http://www.apache.org/licenses/LICENSE-2.0
+
+#include "bucket/BucketInputIterator.h"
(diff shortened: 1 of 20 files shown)


MonsieurNicolas:

vogel:
Rebased

MonsieurNicolas:
this needs to be rebased

vogel:
Rebased. #1458 delayed until new implementation is done (with listing all files).

src/bucket/BucketInputIterator.cpp
```diff
@@ -0,0 +1,66 @@
+// Copyright 2017 Stellar Development Foundation and contributors. Licensed
```
MonsieurNicolas: new files need to be added to Visual Studio project files

vogel:
Added files to VS project

MonsieurNicolas:
r+ 075b2892190d9db09d31796741a10358a563a3ba

PR #1462, opened 2017-12-14
publish queued history just after node is synced

### LINKED ISSUE #1268: Too long waiting time for publish
Initial conditions:
- node configured for publishing history
- big complete or recent catchup initiated

Results:
- queue for publishing is growing larger and larger
- first prepare-snapshot work is being reset multiple times, each time with increasing waiting time
- after catchup is done, prepare-snapshot can have about 3600 seconds of waiting - this is full hour

Solution
- do not start publish work items in other states than LM_SYNCED_STATE

Possible problems:
```
    mSnapshot->mLocalState.resolveAnyReadyFutures();
    mSnapshot->makeLive();
```
- from ResolveSnapshotWork::onRun will be run later, need to check if this changes anything

vogel:
Do not start publish work as soon as history entry has been added to
publishqueue table - do that after node is synced. Publish wont work
when node is not synced, it will just backoff PublishWork. It can cause
publish to be started more than hour later that it could be.

Fixes #1268

### DIFF
--- a/src/history/HistoryManagerImpl.cpp
+++ b/src/history/HistoryManagerImpl.cpp
@@ -510,7 +510,10 @@ HistoryManagerImpl::queueCurrentHistory()
 void
 HistoryManagerImpl::takeSnapshotAndPublish(HistoryArchiveState const& has)
 {
-    if (mPublishWork)
+    // we can only publish when ledger is synced
+    // if not, ResolveSnapshotWork will fail
+    if (mPublishWork ||
+        mApp.getLedgerManager().getState() != LedgerManager::LM_SYNCED_STATE)
     {
         mPublishDelay.Mark();
         return;
--- a/src/ledger/LedgerManagerImpl.cpp
+++ b/src/ledger/LedgerManagerImpl.cpp
@@ -148,6 +148,12 @@ LedgerManagerImpl::setState(State s)
         {
             mApp.getCatchupManager().logAndUpdateCatchupStatus(true);
         }
+
+        if (mState == LM_SYNCED_STATE)
+        {
+            // we can now publish history again
+            mApp.getHistoryManager().publishQueuedHistory();
+        }
     }
 }
 
--- a/src/main/ApplicationImpl.cpp
+++ b/src/main/ApplicationImpl.cpp
@@ -348,12 +348,6 @@ ApplicationImpl::start()
                 maintenance();
             }
             mOverlayManager->start();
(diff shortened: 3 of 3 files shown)


MonsieurNicolas:
The fix could have been done by getting rid of the reason it fails in the first place: why do we need to have `LedgerManager` in state `LM_SYNCED_STATE` at all?
If things got placed in the published queue - wouldn't that imply that they got handed off to the publish subsystem at some point? I would much prefer those semantics.

Only reason I can think of why we have this condition is that maybe this was added before there was a "ledger chain validation" (that allows to validate that the headers we got link back to the network ledger) - without that validation we can only be sure that the chain is valid after catching up.

With the chain validation done, the only post condition to a ledger close during history catchup should be that the ledger is indeed the one we expected (and only then put it in the publish queue).

MonsieurNicolas:

vogel:
I'm not sure why this was introduced.

But I think following scenerio is possible:
* node does not support current version of protocol
* node starts doing catchup
* at some point `lm.closeLedger()` is called from `ApplyLedgerChainWork::applyHistoryOfSingleLedger`
* it calls `HistoryManager::maybeQueueHistoryCheckpoint`, so a history entry is queued for publishing and maybe even published in next few seconds
* then `ApplyLedgerChainWork::applyHistoryOfSingleLedger` does `lm.getLastClosedLedgerHeader().hash != hHeader.hash` and it fails (because of version mismatch), so node is publishing invalid entry

Current code prevents that.

MonsieurNicolas:
I see - I think that you are bringing up a good point that is not just limited to catchup:
in general we do not want to publish the latest ledger, but wait for the next ledger to close just to be sure that we computed the same ledger than the rest of the network.

So, I think we can tackle this problem by:
* delaying publish by one ledger - the buckets can still be pushed to the publish queue but snapshot would just wait for the current ledger to be passed whatever it's trying to publish (can be done by changing the condition in `ResolveSnapshotWork::onRun()` - it does have the same issue of potentially causing too much backoff, but the risk is small that we don't close the next ledger)
* remove conditions (as per my suggestion earlier)

That way we will be not be failing things, and even better, we'll be publishing during catchup and thus avoid building a huge publish backlog (which, in catchup-complete mode can be orders of many GB).

In the event that we build a bad ledger, the only recovery available right now is to perform a `newdb`, so having a bad entry in the publish queue is fine.

PR #1463, opened 2017-12-14
remove manual checkpoint command

vogel:
This is now not needed. There is no way to use that checkpoint from any
other stellar-core instances, as manual catchups are now done using
normal checkpoints (every 64-ledgers ones).


MonsieurNicolas:
needs to be rebased

MonsieurNicolas:
we'll include this change in the upcoming release

vogel:
rebased

MonsieurNicolas:
r+ 7f3e1d07b85def62c5f713d0f99d4d2c3beef5da

PR #1466, opened 2017-12-18
Upgrade improvements

### LINKED ISSUE #1453: Improve upgrade experience
As of v9.0.0 we're pretty strict when it comes to validating upgrades but there are a few trade-offs:
current logic is fairly simple as a node looks at the value of `PREFERRED_UPGRADE_DATETIME` and if the current network time is before that time it will consider values invalid.

A few problems with the current solution:
* if the node gets upgraded to a new stellar-core binary, and the operator forgets to update `PREFERRED_UPGRADE_DATETIME`, the node will vote right away for the protocol upgrade which may cause issues
* if the network is reconfigured, the node will lose sync and vote for old values after rejoining the network, which is not the intention

A potential way to solve this is to add a new field in the ledger header to track the "version number" of the network wide settings `(protocol_version, fee, reserve, ...)`.

The configuration then would also contain the "version number" that the node is configured for, making it possible to adjust the behavior depending on what the current version number is (for example, only validate and vote if the version number matches, consider a version mismatch always invalid).

There might be a way to instead recycle the `ledgerVersion` field (currently 32 bit), turning it into a combination of a strictly monotonic field (16 bits for example) and use the other bits to track what ledgerVersion tracks today (ie: protocol version). 
16 bits seems plenty for network wide settings upgrade: doing such a change every week leaves room for over 1000 years of upgrades.

MonsieurNicolas:
resolves #1453 

This PR changes the way network settings are done:
1. It replaces the config driven upgrade mechanism by a "single shot" mechanism that gets cleared when the desired upgrade is seen from SCP
2. requires node operators to specify that they want an upgrade (as opposed to guessing based on config values). This makes voting for upgrades by accident much less likely.
3. changes how upgrades are validated by allowing nodes to accept the upgrades if a v-blocking set voted for it during the ballot protocol phase yet not voting for those during nomination

### DIFF
--- a/src/herder/HerderImpl.cpp
+++ b/src/herder/HerderImpl.cpp
@@ -764,22 +764,6 @@ HerderImpl::triggerNextLedger(uint32_t ledgerSeqToTrigger)
         }
     }
 
-    if (!upgrades.empty())
-    {
-        auto message = fmt::format(
-            "Proposing herder configuration upgrades: {0}; network "
-            "values for LCL are: {1}",
-            Upgrades::toString(upgrades), Upgrades::toString(lcl.header));
-        CLOG(INFO, "Herder") << message;
-        mApp.getStatusManager().setStatusMessage(
-            StatusCategory::REQUIRES_UPGRADES, message);
-    }
-    else
-    {
-        mApp.getStatusManager().removeStatusMessage(
-            StatusCategory::REQUIRES_UPGRADES);
-    }
-
     mHerderSCPDriver.nominate(slotIndex, newProposedValue, proposedSet,
                               lcl.header.scpValue);
 }
@@ -789,6 +773,21 @@ HerderImpl::setUpgrades(Upgrades::UpgradeParameters const& upgrades)
 {
     mUpgrades.setParameters(upgrades, mApp.getConfig());
     persistUpgrades();
+
+    auto desc = mUpgrades.toString();
+
+    if (!desc.empty())
+    {
+        auto message = fmt::format("Armed with network upgrades: {}", desc);
+        CLOG(INFO, "Herder") << message;
+        mApp.getStatusManager().setStatusMessage(
+            StatusCategory::REQUIRES_UPGRADES, message);
+    }
+    else
+    {
+        mApp.getStatusManager().removeStatusMessage(
+            StatusCategory::REQUIRES_UPGRADES);
+    }
 }
 
 std::string
--- a/src/herder/HerderImpl.cpp
+++ b/src/herder/HerderImpl.cpp
@@ -808,7 +808,7 @@ HerderImpl::setUpgrades(Upgrades::UpgradeParameters const& upgrades)
 }
 
 std::string
-HerderImpl::getUpgrades()
+HerderImpl::getUpgradesJson()
 {
     return mUpgrades.getParameters().toJson();
 }
--- a/src/herder/HerderImpl.cpp
+++ b/src/herder/HerderImpl.cpp
@@ -962,6 +962,33 @@ HerderImpl::restoreSCPState()
 }
 
 void
+HerderImpl::persistUpgrades()
+{
+    auto s = mUpgrades.getParameters().toJson();
+    mApp.getPersistentState().setState(PersistentState::kLedgerUpgrades, s);
+}
+
+void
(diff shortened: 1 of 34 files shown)


src/herder/HerderImpl.cpp
```diff
@@ -787,7 +787,14 @@ HerderImpl::triggerNextLedger(uint32_t ledgerSeqToTrigger)
 void
 HerderImpl::setUpgrades(Upgrades::UpgradeParameters const& upgrades)
 {
-    mUpgrades.setParameters(upgrades);
+    mUpgrades.setParameters(upgrades, mApp.getConfig());
+    persistUpgrades();
+}
+
+std::string
+HerderImpl::getUpgrades()
```
vogel: This does not match setUpgrades signature.

I would expect setUpgrades/getUpgrades to have matching arguments/return type.
Maybe rename that to getUpgradesJson or make it return mUpgrades.getParameters() and let the caller to worry about calling .toJson()?

src/herder/HerderImpl.cpp
```diff
@@ -977,7 +984,16 @@ HerderImpl::restoreUpgrades()
     {
         Upgrades::UpgradeParameters p;
         p.fromJson(s);
-        mUpgrades.setParameters(p);
+        try
+        {
+            // use common code to set status
```
vogel: I think this comment is misleading. What status?

src/herder/Upgrades.cpp
```diff
@@ -66,8 +66,15 @@ Upgrades::Upgrades(UpgradeParameters const& params) : mParams(params)
 }
 
 void
-Upgrades::setParameters(UpgradeParameters const& params)
```
vogel: I'm not sure about passing whole Config here as we only need one value of it. What about passing `uint32_t expectedProtocol` version instead?

src/main/CommandHandler.cpp
```diff
@@ -678,6 +712,79 @@ CommandHandler::unban(std::string const& params, std::string& retStr)
...
+            return;
+        }
+        tm.tm_isdst = false;
+        p.mUpgradeTime = VirtualClock::tmToPoint(tm);
+
+        auto addParam = [&](std::string const& name,
+                            stellar::optional<uint32>& f) {
+            uint32 v;
+            bool updated;
+            if (!parseNumParam(retMap, name, v, retStr,
```
vogel: I start to think we should use exceptions inside parseNumParam and just return optional<uint32> from it.
Now its API is very unnatural - it has 3 return values - bool, uint32 and another bool.

src/herder/HerderImpl.cpp
```diff
@@ -189,6 +189,17 @@ HerderImpl::valueExternalized(uint64 slotIndex, StellarValue const& value)
         static_cast<uint32>(slotIndex),
         getSCP().getExternalizingState(slotIndex));
 
+    // reflect upgrades with the ones included in this SCP round
+    {
+        bool updated;
+        auto newUpgrades = mUpgrades.removeUpgrades(
+            value.upgrades.begin(), value.upgrades.end(), updated);
+        if (updated)
```
vogel: I think it would be nicer to say here `if (newUpgrades != mUpgrades)`. It would require implementing `!=`.

MonsieurNicolas: yeah if we end up needing to compare we can do that at some point

src/herder/HerderImpl.cpp
MonsieurNicolas: it refers to `statusManager`, I had to put this comment to explain why it makes sense to do read followed by write. I thought that splitting status out of `setUpgrades` was not worth it as it would create code paths where people forget to update status

src/main/CommandHandler.cpp
MonsieurNicolas: Yeah that would make sense, especially that now I made the exception handler global. I created an issue for this

jedmccaleb:
r+ 448efc2

src/herder/HerderImpl.cpp
MonsieurNicolas: good point - I'll make that change

MonsieurNicolas:
r+ ab42d2fb55970fc49f23394d2866e57b63429fcb

PR #1470, opened 2017-12-20
Change maybe put signature

vogel:
Rename to maybePut and change one argument that is InputIterator but in reality it is BucketEntry.


MonsieurNicolas:
r+ 1b0d62f72abeaa7cf175a015cc8b1c30b105fd2d

PR #1471, opened 2017-12-21
Simplify ledger entry cmp

vogel:


MonsieurNicolas:
r+ 7314bd6fe8a1fa07a49270a50d5dbe0f4ea6841e

vogel:
@latobarita: retry

MonsieurNicolas:
it looks like the change is not compatible with the version of clang that we're using

PR #1475, opened 2017-12-30
Improve lockFile error output

mfrg:
Any failure in lockFile would simply log that the lock file was already locked, which is not true in some cases (i.e. permission denied or other open/flock failure cases).  This changes lockFile to not return a bool, and instead throw with a more descriptive message for the error.  Since this function would already throw sometimes (checked case for already locked internally, the insert could throw) it seemed sensible to standardize on that model.

**Example output for the permission denied case**

**New**
2017-12-30T12:07:42.230 GBNLJ [default FATAL] Got an exception: unable to open lock file: /tmp/buckets/stellar-core.lock (Permission denied) [main.cpp:652]

**Old**
2017-12-30T12:15:52.907 GC2GL [default FATAL] Got an exception: Found existing lockfile '/tmp/buckets/stellar-core.lock' that is already locked. [main.cpp:652]

### DIFF
--- a/src/bucket/BucketManagerImpl.cpp
+++ b/src/bucket/BucketManagerImpl.cpp
@@ -110,12 +110,11 @@ BucketManagerImpl::getBucketDir()
 
         std::string lock = d + "/" + kLockFilename;
 
-        if (!fs::lockFile(lock))
-        {
-            std::string msg("Found existing lockfile '" + lock +
-                            "' that is already locked.");
-            throw std::runtime_error(msg);
-        }
+        // there are many reasons the lock can fail so let lockFile throw
+        // directly for more clear error messages since we end up just raising
+        // a runtime exception anyway
+        fs::lockFile(lock);
+
         mLockedBucketDir = make_unique<std::string>(d);
     }
     return *(mLockedBucketDir);
--- a/src/util/Fs.cpp
+++ b/src/util/Fs.cpp
@@ -8,6 +8,7 @@
 #include "util/Logging.h"
 #include <map>
 #include <regex>
+#include <sstream>
 
 #ifdef _WIN32
 #include <direct.h>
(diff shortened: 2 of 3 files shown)


MonsieurNicolas:
Thanks for the PR! This makes sense.
As you're a first time contributor, can you go through https://github.com/stellar/stellar-core/blob/master/CONTRIBUTING.md#submitting-changes ?

mfrg:
No problem, done.

MonsieurNicolas:
r+ 642a9f7b59efcdf6f348070b41bd400982560ac1

MonsieurNicolas:
looks like you didn't run `clang-format` before submitting your patch.
Please do so and rebase
```
diff --git a/src/util/Fs.cpp b/src/util/Fs.cpp
index c2740c7..0ee843e 100644
--- a/src/util/Fs.cpp
+++ b/src/util/Fs.cpp
@@ -7,8 +7,8 @@
 #include "lib/util/format.h"
 #include "util/Logging.h"
 #include <map>
-#include <sstream>
 #include <regex>
+#include <sstream>
 
 #ifdef _WIN32
 #include <direct.h>
@@ -173,7 +173,8 @@ lockFile(std::string const& path)
 
     if (fd == -1)
     {
-        errmsg << "unable to open lock file: " << path << " (" << strerror(errno) << ")";
+        errmsg << "unable to open lock file: " << path << " ("
+               << strerror(errno) << ")";
         throw std::runtime_error(errmsg.str());
     }
 
@@ -181,7 +182,8 @@ lockFile(std::string const& path)
     if (r != 0)
     {
         close(fd);
-        errmsg << "unable to flock file: " << path << " (" << strerror(errno) << ")";
+        errmsg << "unable to flock file: " << path << " (" << strerror(errno)
+               << ")";
         throw std::runtime_error(errmsg.str());
     }
 ```

mfrg:
Should be all set now

MonsieurNicolas:
I prefer to avoid the clang-format churn (we used to before having it enforced by the bot and it's annoying when traversing history), can you squash those commits and I'll merge?

mfrg:
Yup, makes sense, how's it look now?

MonsieurNicolas:
r+ 96da5c50d587404ece391f965c533a5896acbb19

MonsieurNicolas:
@latobarita: retry

MonsieurNicolas:
@latobarita: retry

Issue #1473, opened 2017-12-28
Fail to build stellar-core in MacOS high Sierra


swwaden:
In file included from herder/Upgrades.cpp:9:
../lib/cereal/include/cereal/archives/json.hpp:831:8: error: no matching member function for call to 'loadValue'
    ar.loadValue( t );
    ~~~^~~~~~~~~

3 warnings and 1 error generated.
make[2]: *** [herder/Upgrades.o] Error 1
make[1]: *** [install] Error 2
make: *** [install-recursive] Error 1

MonsieurNicolas:
this was fixed in 9.1.0

PR #1477, opened 2018-01-02
Replace lib/cereal folder with a submodule

jonjove:
The fix for #1473 was introduced in https://github.com/USCiLab/cereal/commit/0022152a85958ca76ce4fe7a26dbdde93c692e22. That commit was included in cereal v1.1.0. Instead of update to v1.1.0, which is quite old at this point, I updated to v1.2.2 which is the most recent release. The change log can be found at https://github.com/USCiLab/cereal/releases.


vogel:
This one should be merged after history issue is resolved and tested with parallel catchup. We need to check if history serialization/deserialization does not change.

Note: add tests for serialization/deserialization with HAS->string and string->HAS.

MonsieurNicolas:
+1 let's add some simple round trip unit test for `has` and let's get this merged soon

MonsieurNicolas:
r+ e4349eaad2bff03d28952f1fd2f9eebc7c3c96ca

MonsieurNicolas:
r+ 2e0db44f01ae18882200ccb946662b746bb89503

MonsieurNicolas:
@latobarita: retry

PR #1481, opened 2018-01-08
buckets: do not keep keepDeadEntries in FutureBucket, instead pass it…

vogel:
… in makeLive

This fixes issue where bucket would store dead entries because if was
merged after deserialization.

### DIFF
--- a/src/history/StateSnapshot.cpp
+++ b/src/history/StateSnapshot.cpp
@@ -50,11 +50,12 @@ StateSnapshot::StateSnapshot(Application& app, HistoryArchiveState const& state)
 void
 StateSnapshot::makeLive()
 {
-    for (auto& hb : mLocalState.currentBuckets)
+    for (auto i = 0; i < mLocalState.currentBuckets.size(); i++)
     {
+        auto& hb = mLocalState.currentBuckets[i];
         if (hb.next.hasHashes() && !hb.next.isLive())
         {
-            hb.next.makeLive(mApp);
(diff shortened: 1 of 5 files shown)


src/history/StateSnapshot.cpp
```diff
@@ -50,12 +50,14 @@ StateSnapshot::StateSnapshot(Application& app, HistoryArchiveState const& state)
 void
 StateSnapshot::makeLive()
 {
+    auto i = 0;
     for (auto& hb : mLocalState.currentBuckets)
```
MonsieurNicolas: I think it would be cleaner to use `i` as the index in the loop and access `hb` using a direct lookup `mLocalState.currentBuckets[i]` instead of tracking the same thing in two different ways

jonjove: I recommend using `mLocalState.currentBuckets.at(i)` since `std::vector::at` has bounds checking whereas `std::vector::operator[]` does not.

vogel: I do not thing we need to check bounds when we do i = 0; i < mLocalState.currentBuckets.size();, so I'll just use []

vogel:
All comments addressed

MonsieurNicolas:
was there an issue for this we can link back to?

MonsieurNicolas:
r+ 3cca167201541881fb98e9309a6c5d7b41c85c0e

PR #1495, opened 2018-01-16
decouple SCP tracking from externalization

### LINKED ISSUE #1494: nodes sometimes fallback to "joining SCP" when machine is under stress
I've seen this happen on slow hardware (laptop), or when running many parallel jobs.

It looks like the way timers are implemented is suboptimal as "joining SCP" should only be a function of the performance of consensus

MonsieurNicolas:
there is no reason to not attempt nomination in this case.

resolves #1494


vogel:
r+ b660b398588b87fd6c235279743316c785b59f0b

PR #1502, opened 2018-01-19
Reduce range of operator less

vogel:
This time it works both with clang and g++ - David has fixed xdrpp.


MonsieurNicolas:
r+ 91a14dd589367f88f7f24cc713fb217febc273a1

PR #1511, opened 2018-01-24
Ledger state

jonjove:
These changes form the framework for the transition to the new data-layer. All of the objects that will make up the implementation of the data-layer are included. However, at this point the new data-layer is not yet used anywhere in stellar-core.

### DIFF
--- a/src/ledger/LedgerState.cpp
+++ b/src/ledger/LedgerState.cpp
@@ -660,6 +660,24 @@ LedgerState::getOffers(Asset const& selling, Asset const& buying,
     }
 }
 
+bool
+LedgerState::isLoadedInMemory(LedgerKey const& key)
+{
+    auto iter = mState.find(key);
+    if (iter != mState.end())
+    {
+        return true;
+    }
+    else if (mParent)
+    {
+        return mParent->isLoadedInMemory(key);
+    }
+    else
+    {
+        return false;
+    }
+}
+
 LedgerState::LoadBestOfferContext::LoadBestOfferContext(
     Asset const& selling, Asset const& buying, LedgerState& ledgerState)
     : mSelling(selling)
@@ -702,29 +720,36 @@ LedgerState::LoadBestOfferContext::loadBestOffer()
         mTop.reset();
     }
 
-    loadFromDatabaseIfNecessary();
-    if (mInMemory.empty())
+    while (!mTop)
     {
-        if (mFromDatabase.empty())
+        loadFromDatabaseIfNecessary();
+        if (mInMemory.empty())
         {
-            return nullptr;
+            if (mFromDatabase.empty())
+            {
+                return nullptr;
+            }
+            else
+            {
+                mTop = mFromDatabase.front();
+                mFromDatabase.erase(mFromDatabase.begin());
+            }
+        }
+        else if (mFromDatabase.empty() ||
+                 compareOffers(mInMemory.top(), mFromDatabase.front()))
+        {
+            mTop = mInMemory.top();
+            mInMemory.pop();
         }
         else
         {
             mTop = mFromDatabase.front();
             mFromDatabase.erase(mFromDatabase.begin());
+            if (mLedgerState.isLoadedInMemory(LedgerEntryKey(*mTop->entry())))
+            {
+                mTop.reset();
+            }
         }
-    }
-    else if (mFromDatabase.empty() ||
-             compareOffers(mInMemory.top(), mFromDatabase.front()))
-    {
-        mTop = mInMemory.top();
-        mInMemory.pop();
-    }
-    else
-    {
-        mTop = mFromDatabase.front();
-        mFromDatabase.erase(mFromDatabase.begin());
     }
 
     auto const& entry = mTop->ignoreInvalid().entry();
--- a/src/ledger/LedgerState.cpp
+++ b/src/ledger/LedgerState.cpp
@@ -25,6 +25,9 @@ operator==(InflationVotes const& lhs, InflationVotes const& rhs)
 
 LedgerStateRoot::LedgerStateRoot(Database& db) : mHasChild(false), mDb(db)
 {
+    // TODO(jonjove): Remove this once LedgerStateRoot is properly integrated
+    // with LedgerManager. This is required to make tests pass right now though
+    mPreviousHeader.ledgerSeq = 2;
 }
 
 void
@@ -74,6 +77,10 @@ LedgerState::LedgerState(LedgerState& parent)
     for (auto const& state : mParent->mState)
     {
         state.second->invalidate();
+    }
+    if (mParent->mHeader)
+    {
+        mParent->mHeader->invalidate();
     }
     mParent->mLoadBestOfferContext.clear();
 }
@@ -262,11 +269,12 @@ LedgerState::mergeStateIntoRoot()
 void
 LedgerState::mergeHeaderIntoRoot()
 {
-    if (mHeader)
+    if (!mHeader)
     {
-        storeHeaderInDatabase();
-        mRoot->setPreviousHeader(mHeader->ignoreInvalid().header());
+        loadHeader()->invalidate();
     }
+    storeHeaderInDatabase();
+    mRoot->setPreviousHeader(mHeader->ignoreInvalid().header());
 }
 
 void
@@ -417,10 +425,6 @@ LedgerState::loadHeader()
     assert(!mChild);
 
     mHeader = loadHeaderHelper();
-    auto& header = mHeader->header();
-    ++header.ledgerSeq;
-    header.previousLedgerHash = sha256(xdr::xdr_to_opaque(header));
-    mHeader->invalidate();
     return mHeader;
 }
 
@@ -453,8 +457,11 @@ LedgerState::loadHeaderHelper()
     }
     else
     {
-        auto const& header = mRoot->getPreviousHeader();
-        lhr = StateHeader(new LedgerHeaderReference(header, header));
+        auto const& previousHeader = mRoot->getPreviousHeader();
+        lhr = StateHeader(new LedgerHeaderReference(previousHeader, previousHeader));
+        auto& header = lhr->header();
+        ++header.ledgerSeq;
+        header.previousLedgerHash = sha256(xdr::xdr_to_opaque(previousHeader));
     }
     return lhr;
 }
--- a/src/ledger/LedgerState.cpp
+++ b/src/ledger/LedgerState.cpp
@@ -162,14 +162,14 @@ LedgerState::getDeadEntries()
 }
 
 void
-LedgerState::commit(std::function<void()> f)
+LedgerState::commit(std::function<void()> onCommitToDatabase)
 {
     assert(mParent || mRoot);
     assert(!mChild);
 
     if (mParent)
     {
-        assert(!f);
+        assert(!onCommitToDatabase);
         mergeStateIntoParent();
         mergeHeaderIntoParent();
         mParent->mChild = nullptr;
@@ -182,9 +182,9 @@ LedgerState::commit(std::function<void()> f)
         {
             mergeStateIntoRoot();
             mergeHeaderIntoRoot();
-            if (f)
+            if (onCommitToDatabase)
             {
-                f();
+                onCommitToDatabase();
             }
             sqlTx.commit();
         }
--- a/src/ledger/LedgerState.cpp
+++ b/src/ledger/LedgerState.cpp
@@ -1,0 +1,801 @@
+// Copyright 2017 Stellar Development Foundation and contributors. Licensed
+// under the Apache License, Version 2.0. See the COPYING file at the root
+// of this distribution or at http://www.apache.org/licenses/LICENSE-2.0
+
+#include "ledger/LedgerState.h"
+#include "crypto/Hex.h"
+#include "crypto/SHA.h"
+#include "ledger/LedgerEntryReference.h"
+#include "ledger/LedgerHeaderReference.h"
+#include "xdr/Stellar-ledger.h"
+#include "xdrpp/marshal.h"
+
+// TODO(jonjove): Remove this header (used for LedgerEntryKey)
+#include "ledger/EntryFrame.h"
+
+namespace stellar
+{
+using xdr::operator==;
+
+bool
+operator==(InflationVotes const& lhs, InflationVotes const& rhs)
+{
+    return (lhs.votes == rhs.votes) && (lhs.inflationDest == rhs.inflationDest);
+}
+
+LedgerStateRoot::LedgerStateRoot(Database& db) : mHasChild(false), mDb(db)
+{
+}
+
+void
+LedgerStateRoot::hasChild(bool child)
+{
+    assert(!child || !mHasChild);
+    mHasChild = child;
+}
+
+LedgerHeader const&
+LedgerStateRoot::getPreviousHeader()
+{
+    return mPreviousHeader;
+}
+
+void
+LedgerStateRoot::setPreviousHeader(LedgerHeader const& header)
+{
+    mPreviousHeader = header;
+}
+
+Database&
+LedgerStateRoot::getDatabase()
+{
+    return mDb;
+}
+
+Database::EntryCache&
+LedgerStateRoot::getCache()
+{
+    // TODO(jonjove): Do we want to continue using the cache from Database?
+    // We could refactor so the cache is a member of LedgerStateRoot
+    return mDb.getEntryCache();
+}
+
+LedgerState::LedgerState(LedgerStateRoot& root)
+    : mRoot(&root), mParent(nullptr), mChild(nullptr)
+{
+    mRoot->hasChild(true);
+}
+
+LedgerState::LedgerState(LedgerState& parent)
+    : mRoot(nullptr), mParent(&parent), mChild(nullptr)
+{
+    assert(!mParent->mChild);
+    mParent->mChild = this;
+    for (auto const& state : mParent->mState)
+    {
+        state.second->invalidate();
+    }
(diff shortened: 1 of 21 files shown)


src/ledger/LedgerState.cpp
```diff
@@ -0,0 +1,801 @@
...
+{
+    return (lhs.votes == rhs.votes) && (lhs.inflationDest == rhs.inflationDest);
+}
+
+LedgerStateRoot::LedgerStateRoot(Database& db) : mHasChild(false), mDb(db)
+{
+}
+
+void
+LedgerStateRoot::hasChild(bool child)
```
vogel: name of this method suggest that it returns bool, but it does not

src/ledger/LedgerState.cpp
```diff
@@ -0,0 +1,801 @@
...
+Database&
+LedgerStateRoot::getDatabase()
+{
+    return mDb;
+}
+
+Database::EntryCache&
+LedgerStateRoot::getCache()
+{
+    // TODO(jonjove): Do we want to continue using the cache from Database?
```
vogel: +1 for moving cache here

src/ledger/LedgerState.cpp
```diff
@@ -0,0 +1,801 @@
...
+    for (auto const& state : mParent->mState)
+    {
+        state.second->invalidate();
+    }
+    mParent->mLoadBestOfferContext.clear();
+}
+
+LedgerState::~LedgerState()
+{
+    if (mParent || mRoot)
```
vogel: is it possible to be true? I see only 2 constructors here and one always sets mParant and second one mRoot
maybe assert would be better?

src/ledger/LedgerState.cpp
```diff
@@ -0,0 +1,801 @@
...
+LedgerEntryChanges
+LedgerState::getChanges()
+{
+    assert(mParent || mRoot);
+    assert(!mChild);
+
+    LedgerEntryChanges changes;
+    for (auto const& state : mState)
+    {
+        // TODO(jonjove): Should output meta if *entry == *previous?
```
vogel: probably: https://github.com/stellar/stellar-core/commit/fe955a910a43ae1a7c3ac744b579502b7565f01c

src/ledger/LedgerState.cpp
```diff
@@ -0,0 +1,801 @@
...
+            }
+        }
+    }
+    return changes;
+}
+
+std::vector<LedgerEntry>
+LedgerState::getLiveEntries()
+{
+    assert(mRoot);
```
vogel: this looks like it is only allowed to be called by LedgerState just below root
what about moving that code to LedgerStateRoot?

src/ledger/LedgerState.cpp
jonjove: This comes into play if `LedgerState::rollback` or `LedgerState::commit` has already been called. When either of those occur, we effectively detach the nested `LedgerState` from the rest of the chain. In this case, `mParent == mRoot == nullptr`. This prevents `LedgerState` from rolling back if it has already been handled. An analogous assert appears in `rollback` and `commit`.

src/ledger/LedgerState.cpp
```diff
@@ -0,0 +1,801 @@
...
+
+void
+LedgerState::commit(std::function<void()> f)
+{
+    assert(mParent || mRoot);
+    assert(!mChild);
+
+    if (mParent)
+    {
+        assert(!f);
```
vogel: please use more descriptive name for that

src/ledger/LedgerState.cpp
```diff
@@ -0,0 +1,801 @@
...
+        mergeStateIntoParent();
```
vogel: like those, those are great!

src/ledger/LedgerState.cpp
```diff
@@ -0,0 +1,801 @@
...
+{
+    for (auto const& state : mState)
+    {
+        state.second->invalidate();
+        auto const& entry = state.second->ignoreInvalid().entry();
+
+        auto iter = mParent->mState.find(state.first);
+        if (iter == mParent->mState.end())
+        {
+            StateEntry ler(new LedgerEntryReference(entry, nullptr));
```
vogel: since C++ I don't like seeing naked new in code, what about make_shared?

src/ledger/LedgerState.cpp
```diff
@@ -0,0 +1,801 @@
...
+
+    mHeader.reset();
+    mState.clear();
+    mLoadBestOfferContext.clear();
+}
+
+LedgerState::StateEntry
+LedgerState::create(LedgerEntry const& entry)
+{
+    assert(mParent || mRoot);
```
vogel: this set of assets appears everywhere, maybe promote it to a method?

src/ledger/LedgerState.cpp
```diff
@@ -0,0 +1,801 @@
...
+LedgerState::StateHeader
+LedgerState::loadHeader()
+{
+    assert(mParent || mRoot);
+    assert(!mChild);
+
+    mHeader = loadHeaderHelper();
+    auto& header = mHeader->header();
+    ++header.ledgerSeq;
+    header.previousLedgerHash = sha256(xdr::xdr_to_opaque(header));
```
vogel: just wondering, will it use proper value of ledgerSeq as it was just incremented? seems strange to be that methods named "loadHeader" does modifications on header object

src/ledger/LedgerState.cpp
jonjove: Definitely a good point. Renamed to `onCommitToDatabase`. This was added to allow some other SQL changes to be bundled into the SQL transaction that writes `LedgerState` to database. This will be used during ledger close as some things are written, for example, to `PersistentState` within the same SQL transaction.

src/ledger/LedgerState.cpp
jonjove: `std::make_shared` can only be used when objects have public constructors (since otherwise `std::make_shared` does not have access to the constructor). But `LedgerEntryReference` and `LedgerHeaderReference` have private constructors as they can only be constructed by `LedgerState`. If you prefer it, we can promote this type of code into a method of its own. For example, something like `StateEntry makeSharedLedgerEntryReference(...)`. I think this is potentially a good idea, since it will mean that naked `new` will be in one place where we can have a clarifying comment explaining the necessity.

src/ledger/LedgerState.cpp
jonjove: I considered this also (it was actually that way in the first draft of the code) but I wasn't sure it would actually be easier to follow in the future if that was a method. If the consensus is that it would be easier that way, I'm not opposed to changing it back.

src/ledger/LedgerState.cpp
jonjove: Actually, this doesn't look quite right to me. I'm going to revisit this now. I hadn't written any tests for this function yet because I wasn't sure exactly what I wanted to test, but I'm going to write one now.

vogel:
What do you think about extracting LedgerState classes to new directory? Either ledger/state or just ledgerstate?

src/ledger/LedgerState.cpp
```diff
@@ -0,0 +1,801 @@
...
+            StateEntry ler(new LedgerEntryReference(entry, nullptr));
+            ler->invalidate();
+            mParent->mState[state.first] = ler;
+        }
+        else
+        {
+            auto const& previous =
+                iter->second->ignoreInvalid().previousEntry();
+            StateEntry ler(new LedgerEntryReference(entry, previous));
+            ler->invalidate();
```
vogel: This means that we don't want anyone to use that LedgerEntryReference anymore? Even if it was just created here and is 100% valid?

Do I understand this correctly? If so, please and comment and explain why is that needed. I think that this means we won't be able to use this entry in this LedgerState and one will have to create a new child LedgerState to use it?

src/ledger/LedgerState.cpp
```diff
@@ -0,0 +1,801 @@
...
+}
+
+LedgerState::StateEntry
+LedgerState::createHelper(LedgerEntry const& entry, LedgerKey const& key)
+{
+    StateEntry ler;
+
+    auto iter = mState.find(key);
+    if (iter != mState.end())
+    {
```
vogel: when can this happen? it is unclear to me

src/ledger/LedgerState.cpp
```diff
@@ -0,0 +1,801 @@
...
+}
+
+LedgerState::StateEntry
+LedgerState::loadHelper(LedgerKey const& key)
+{
+    StateEntry ler;
+
+    auto iter = mState.find(key);
+    if (iter != mState.end())
+    {
```
vogel: when can this happen? it is unclear to me

src/ledger/LedgerState.cpp
```diff
@@ -0,0 +1,801 @@
...
+        auto const& header = mRoot->getPreviousHeader();
+        lhr = StateHeader(new LedgerHeaderReference(header, header));
+    }
+    return lhr;
+}
+
+LedgerState::StateEntry
+LedgerState::loadFromDatabase(LedgerKey const& key)
+{
+    assert(mRoot);
```
vogel: looks like this belong to LedgerStateRoot

src/ledger/LedgerState.cpp
```diff
@@ -0,0 +1,801 @@
...
+    }
+    else
+    {
+        mRoot->getCache().put(cacheKey, nullptr);
+    }
+    return ler;
+}
+
+void
+LedgerState::storeInDatabase(StateEntry const& state)
```
vogel: looks like this belong to LedgerStateRoot

src/ledger/LedgerState.cpp
```diff
@@ -0,0 +1,801 @@
...
+        mOffersLoadedFromDatabase += mFromDatabase.size();
+    }
+}
+
+LedgerState::StateEntry
+LedgerState::LoadBestOfferContext::loadBestOffer()
+{
+    if (mTop)
+    {
+        assert(!mTop->valid());
```
vogel: I find that counter intuitive that we are adding on purpose non-valid item to mInMemory here.
Maybe this is a problem with name of "valid" method?

src/ledger/LedgerState.cpp
```diff
@@ -0,0 +1,801 @@
...
+        mFromDatabase = root->loadBestOffersFromDatabase(
+            5, mOffersLoadedFromDatabase, mSelling, mBuying);
+        mOffersLoadedFromDatabase += mFromDatabase.size();
+    }
+}
+
+LedgerState::StateEntry
+LedgerState::LoadBestOfferContext::loadBestOffer()
+{
+    if (mTop)
```
vogel: That inserts top value back into mInMemory. What is purpose of that? Won't it mean that each time the same value is returned?

src/ledger/LedgerState.cpp
```diff
@@ -0,0 +1,801 @@
...
+            root = root->mParent;
+        }
+        mFromDatabase = root->loadBestOffersFromDatabase(
+            5, mOffersLoadedFromDatabase, mSelling, mBuying);
+        mOffersLoadedFromDatabase += mFromDatabase.size();
+    }
+}
+
+LedgerState::StateEntry
+LedgerState::LoadBestOfferContext::loadBestOffer()
```
vogel: I think maybe "popNextBestOffer" would be a better name, as it will be called multiple times and each time it should return different value. load preffix suggest that it should return the same value each time.

src/ledger/LedgerState.cpp
```diff
@@ -0,0 +1,801 @@
...
+        return lhs.offerID < rhs.offerID;
+    }
+    else
+    {
+        return false;
+    }
+}
+
+std::vector<InflationVotes>
+LedgerState::loadInflationWinners(size_t maxWinners, int64_t minBalance)
```
vogel: haha, this one is clever ;)

src/ledger/LedgerState.cpp
jonjove: The `LedgerEntryReference` that is constructed here isn't exactly being created here. This is actually the process of copying the data from the child `LedgerState` to the parent `LedgerState`. So there was already a `LedgerEntryReference` that referenced this data, and it was already used as much as needed in the child `LedgerState` that is being committed.

I adopted the convention that any transaction-type operation (for example: commit, rollback, sub-transaction) would invalidate all existing `LedgerEntryReference`. Being invalid does not mean that you need to create a child `LedgerState` in order to use it, because it is valid to call `load` on the `LedgerKey` of an invalid `LedgerEntryReference` (see lines 372-397). Indeed, this is exactly what is done in the following example:
```
LedgerKey key = ...;
LedgerState lsParent(...);
LedgerState lsChild(lsParent);
someFunction(lsChild, key); // Some rarely accessed branches call lsChild.load(key)
lsChild.commit();
lsParent.load(key); 
```
Because the call to `lsChild.commit()` leaves every `LedgerEntryReference` in the invalid state, `lsParent.load(key)` will never fail with "Valid LedgerEntryReference already exists for this key" even if `someFunction` did not leave the `LedgerEntryReference` corresponding to `key` in the invalid state. 

It is worth noting that "re-loading" the invalidated `LedgerEntryReference` is a purely in-memory operation.

Let me know if this is still not clear, happy to explain more thoroughly or give more examples of different usage patterns (it might also be interesting to look at LedgerStateTests.cpp to see what kind of usage is possible).

src/ledger/LedgerState.cpp
jonjove: Here is an example where this occurs:
```
LedgerKey key = ...;
LedgerState lsParent(...);
LedgerState lsChild(lsParent);
lsChild.load(key)->erase();
lsChild.commit();
lsParent.create(key);
```
Let me know if you need further explanation or examples on what is happening here, I'm happy to explain further.

src/ledger/LedgerState.cpp
jonjove: The comment on line 226 gives an example of how this can happen.

src/ledger/LedgerState.cpp
jonjove: The semantics here are actually more like "loadBestOffer" then "popNextBestOffer". `loadBestOffer` has analogous functionality to `std::priority_queue<...>::top()`. For example, the following test case should succeed:
```
Asset a1, a2 = ...;
LedgerState ls(...);
auto ler1 = ls.loadBestOffer(a1, a2);
auto le1 = *ler1->entry();
ler1->invalidate();
auto ler2 = ls.loadBestOffer(a1, a2);
auto le2 = *ler2->entry();
REQUIRE(le1 == le2);
```
If you want it to behave like "popNextBestOffer" than you would have to use `LedgerEntryReference::erase` like:
```
Asset a1, a2 = ...;
LedgerState ls(...);
auto ler1 = ls.loadBestOffer(a1, a2);
auto le1 = *ler1->entry();
ler1->erase(); // Now we have popped
ler1->invalidate();
auto ler2 = ls.loadBestOffer(a1, a2);
auto le2 = *ler2->entry();
REQUIRE(le1 != le2);
```
This functionality is natural to me, since you do not want to remove the best offer from the database every time you load it.

src/ledger/LedgerState.cpp
jonjove: It does not mean that the same value is returned each time, although if the value is loaded and not modified or erased then it will be loaded again (as it should be, since it would still be the best offer at that point). The comment on line 686 gives an example where the offer loaded by `loadBestOffer` is erased, giving functionality similar to top followed by pop. Consider the case where the offer loaded by `loadBestOffer` is modified such that it is no longer the best offer. Then when it is pushed back into `mInMemory`, it will not necessarily be at the top of the heap.

src/ledger/LedgerState.cpp
jonjove: I agree that this could indicate a problem with the name.

`LoadBestOfferContext` maintains the invariant that there is at most one valid `LedgerEntryReference` referring to an offer (with the correct assets), and if there is one then it is the one that was just returned by `loadBestOffer`. This means it is incorrect to call `loadBestOffer` if any `LedgerEntryReference` referring to an offer (with the correct assets) is valid.

Why is this required though? I think it would be possible to implement this without this constraint, but it would make it impossible to do the performance optimization which is done in `LoadBestOfferContext`. Specifically, if there could be multiple valid `LedgerEntryReferences` referring to offers (with the correct assets) then `LoadBestOfferContext::loadBestOffer` would have to re-process the entire `mInMemory` on each invocation. It is easy to see that this is the case. Any valid `LedgerEntryReference` can be modified at any time (and there is no callback to notify `LedgerState` that a modification has occurred). Suppose that I invoke `LoadBestOfferContext::loadBestOffer`, then modify many offers (with the correct assets), then call `LoadBestOfferContext::loadBestOffer` again. At this point `mInMemory` could be invalid (in the heap sense) at arbitrarily many locations, so it would need to be re-formed into a heap. But that would be no better than just iterating to find the best offer at every invocation. By making the restriction that if a `LedgerEntryReference` referring to an offer (with the correct assets) is only valid if it was just returned by `loadBestOffer` then we know that modifications can only occur at the top of the heap.

The consequence of this is that a sequence of calls to `loadBestOffer` can be efficient if you do not make any intervening calls to `LedgerState::load` or `LedgerState::create` that constructed a `LedgerEntryReference` referring to an offer (with the correct assets). If such an intervening call is made, then the `LoadBestOfferContext` is destroyed so that it has to be reconstructed at the next call to `loadBestOffer`, thereby maintaining the constraint.

src/ledger/LedgerState.cpp
jonjove: Fixed.

MonsieurNicolas:
This got super-seeded by #1600

PR #1512, opened 2018-01-24
Periodic maintenance

### LINKED ISSUE #923: add automatic "maintenance" to stellar-core
stellar-core has a `maintenance` endpoint that should be periodically called to clean up the database.

`maintenance` right now only performs historical pruning of data that was successfully uploaded to history and/or imported into dependent services such as Horizon.

At a minimum, we should call it on startup (probably add a config to control this).

Maybe consider adding an automatic schedule within core (less important).

This can only be implemented after all dependencies such as Horizon properly update their cursor.

vogel:
Fixes #923

### DIFF
--- a/src/main/CommandHandler.cpp
+++ b/src/main/CommandHandler.cpp
@@ -1002,7 +1002,7 @@ CommandHandler::maintenance(std::string const& params, std::string& retStr)
     http::server::server::parseParams(params, map);
     if (map["queue"] == "true")
     {
-        mApp.maintenance();
+        mApp.maintenance(50000);
         retStr = "Done";
     }
     else
--- a/src/main/CommandHandler.cpp
+++ b/src/main/CommandHandler.cpp
@@ -326,10 +326,11 @@ CommandHandler::fileNotFound(std::string const& params, std::string& retStr)
         "endpoint."
         "</p><p><h1> /getcursor?[id=ID]</h1> gets the cursor identified by "
         "'ID'.  If ID is not defined then all cursors will be returned."
-        "</p><p><h1> /maintenance[?queue=true]</h1> Performs maintenance tasks "
-        "on the instance."
-        "<ul><li><i>queue</i> performs deletion of queue data.See setcursor "
-        "for more information</li></ul>"
+        "</p><p><h1> /maintenance[?queue=true[&count=N]]</h1> Performs "
+        "maintenance tasks on the instance."
+        "<ul><li><i>queue</i> performs deletion of queue data. Deletes at most "
+        "count entries from each table (defaults to 50000). See setcursor for "
+        "more information</li></ul>"
         "</p><p><h1> "
         "/unban?node=NODE_ID</h1>"
         "remove ban for PEER_ID"
@@ -1002,7 +1003,12 @@ CommandHandler::maintenance(std::string const& params, std::string& retStr)
     http::server::server::parseParams(params, map);
     if (map["queue"] == "true")
     {
-        mApp.maintenance(50000);
+        uint32_t count = 50000;
+        if (!parseNumParam(map, "count", count, retStr,
+                           Requirement::OPTIONAL_REQ))
+            return;
+
+        mApp.maintenance(count);
         retStr = "Done";
     }
     else
--- a/src/main/CommandHandler.cpp
+++ b/src/main/CommandHandler.cpp
@@ -12,6 +12,7 @@
 #include "lib/util/format.h"
 #include "main/Application.h"
 #include "main/Config.h"
+#include "main/Maintainer.h"
 #include "overlay/BanManager.h"
 #include "overlay/OverlayManager.h"
 #include "util/Logging.h"
@@ -1008,7 +1009,7 @@ CommandHandler::maintenance(std::string const& params, std::string& retStr)
                            Requirement::OPTIONAL_REQ))
             return;
 
-        mApp.maintenance(count);
+        mApp.getMaintainer().performMaintenance(count);
         retStr = "Done";
     }
     else
--- a/src/main/CommandHandler.cpp
+++ b/src/main/CommandHandler.cpp
@@ -68,59 +68,36 @@ CommandHandler::CommandHandler(Application& app) : mApp(app)
 
     mServer->add404(std::bind(&CommandHandler::fileNotFound, this, _1, _2));
 
-    mServer->addRoute("bans", std::bind(&CommandHandler::safeRouter, this,
-                                        &CommandHandler::bans, _1, _2));
-    mServer->addRoute("catchup", std::bind(&CommandHandler::safeRouter, this,
-                                           &CommandHandler::catchup, _1, _2));
-    mServer->addRoute("checkdb", std::bind(&CommandHandler::safeRouter, this,
-                                           &CommandHandler::checkdb, _1, _2));
-    mServer->addRoute("connect", std::bind(&CommandHandler::safeRouter, this,
-                                           &CommandHandler::connect, _1, _2));
-    mServer->addRoute("dropcursor",
-                      std::bind(&CommandHandler::safeRouter, this,
-                                &CommandHandler::dropcursor, _1, _2));
-    mServer->addRoute("droppeer", std::bind(&CommandHandler::safeRouter, this,
-                                            &CommandHandler::dropPeer, _1, _2));
-    mServer->addRoute("generateload",
-                      std::bind(&CommandHandler::safeRouter, this,
-                                &CommandHandler::generateLoad, _1, _2));
-    mServer->addRoute("info", std::bind(&CommandHandler::safeRouter, this,
-                                        &CommandHandler::info, _1, _2));
-    mServer->addRoute("ll", std::bind(&CommandHandler::safeRouter, this,
-                                      &CommandHandler::ll, _1, _2));
-    mServer->addRoute("logrotate",
-                      std::bind(&CommandHandler::safeRouter, this,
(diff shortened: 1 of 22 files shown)


src/main/CommandHandler.cpp
```diff
@@ -969,7 +970,12 @@ CommandHandler::maintenance(std::string const& params, std::string& retStr)
     http::server::server::parseParams(params, map);
     if (map["queue"] == "true")
     {
-        mApp.maintenance(50000);
+        auto count = 50000;
```
MonsieurNicolas: you should use `uint32` instead of auto, otherwise you're allowing people to pass in negative numbers

MonsieurNicolas: talking about that, it looks like `parseNumParam` has a bug: it doesn't check if the stream failed when parsing the input `str.fail()`

src/main/Maintainer.cpp
```diff
@@ -0,0 +1,23 @@
+// Copyright 2018 Stellar Development Foundation and contributors. Licensed
```
MonsieurNicolas: new files -> needs to be added to the Visual Studio project files

src/main/Config.cpp
```diff
@@ -321,14 +322,37 @@ Config::load(std::string const& filename)
                 }
                 ALLOW_LOCALHOST_FOR_TESTING = item.second->as<bool>()->value();
             }
-            else if (item.first == "MAINTENANCE_ON_STARTUP")
+            else if (item.first == "AUTOMATIC_MAINTENANCE_PERIOD")
```
MonsieurNicolas: you need to update `stellar-core_example.cfg`

MonsieurNicolas: in the comments in the config, include also which values to use to disable it

src/main/Config.cpp
```diff
@@ -321,14 +322,37 @@ Config::load(std::string const& filename)
...
             {
-                if (!item.second->as<bool>())
+                if (!item.second->as<int64_t>())
+                {
+                    throw std::invalid_argument(
+                        "invalid AUTOMATIC_MAINTENANCE_PERIOD");
+                }
+                int64_t parsedAutomaticMaintenancePeriod =
+                    item.second->as<int64_t>()->value();
+                if (parsedAutomaticMaintenancePeriod <= 0 ||
```
MonsieurNicolas: maybe allow 0 to disable

src/main/Maintainer.cpp
```diff
@@ -3,16 +3,37 @@
...
 
-Maintainer::Maintainer(Application& app) : mApp{app}
+Maintainer::Maintainer(Application& app) : mApp{app}, mTimer{mApp}
 {
 }
 
+void
+Maintainer::start()
+{
+    scheduleMaintenance();
```
MonsieurNicolas: we could put a condition that `AUTOMATIC_MAINTENANCE_PERIOD > 0` to disable this

src/main/Maintainer.cpp
```diff
@@ -3,16 +3,37 @@
...
+}
+
+void
+Maintainer::scheduleMaintenance()
+{
+    mTimer.expires_from_now(mApp.getConfig().AUTOMATIC_MAINTENANCE_PERIOD);
+    mTimer.async_wait([this]() { tick(); }, VirtualTimer::onFailureNoop);
```
MonsieurNicolas: I think this is safe (app being destructed vs timer events)

src/main/Config.cpp
MonsieurNicolas: I just realized that I actually specified 0 as the way to disable it in the issue

vogel:
Updated

src/main/Config.cpp
```diff
@@ -321,14 +322,37 @@ Config::load(std::string const& filename)
...
+                        "invalid AUTOMATIC_MAINTENANCE_COUNT");
                 }
-                MAINTENANCE_ON_STARTUP = item.second->as<bool>()->value();
+                int64_t parsedAutomaticMaintenanceCount =
+                    item.second->as<int64_t>()->value();
+                if (parsedAutomaticMaintenanceCount < 0 ||
+                    parsedAutomaticMaintenanceCount > UINT32_MAX)
+                    throw std::invalid_argument(
+                        "invalid AUTOMATIC_MAINTENANCE_COUNT");
+                AUTOMATIC_MAINTENANCE_COUNT = static_cast<unsigned short>(
```
MonsieurNicolas: cast should be `uint32_t` not `unsigned short`

src/main/Config.cpp
```diff
@@ -321,14 +322,37 @@ Config::load(std::string const& filename)
...
-                if (!item.second->as<bool>())
+                if (!item.second->as<int64_t>())
+                {
+                    throw std::invalid_argument(
+                        "invalid AUTOMATIC_MAINTENANCE_PERIOD");
+                }
+                int64_t parsedAutomaticMaintenancePeriod =
+                    item.second->as<int64_t>()->value();
+                if (parsedAutomaticMaintenancePeriod < 0 ||
+                    parsedAutomaticMaintenancePeriod > UINT32_MAX)
```
MonsieurNicolas: style: use braces around the `throw` (especially as the if is already spanning multiple lines)

src/main/Config.cpp
```diff
@@ -321,14 +322,37 @@ Config::load(std::string const& filename)
...
                 {
                     throw std::invalid_argument(
-                        "invalid MAINTENANCE_ON_STARTUP");
+                        "invalid AUTOMATIC_MAINTENANCE_COUNT");
                 }
-                MAINTENANCE_ON_STARTUP = item.second->as<bool>()->value();
+                int64_t parsedAutomaticMaintenanceCount =
+                    item.second->as<int64_t>()->value();
+                if (parsedAutomaticMaintenanceCount < 0 ||
+                    parsedAutomaticMaintenanceCount > UINT32_MAX)
```
MonsieurNicolas: same: braces around `throw`

src/main/Config.cpp
```diff
@@ -166,6 +167,49 @@ Config::loadQset(std::shared_ptr<cpptoml::toml_group> group, SCPQuorumSet& qset,
...
+readString(ConfigItem const& item)
+{
+    if (!item.second->as<std::string>())
+    {
+        throw std::invalid_argument(fmt::format("invalid {}", item.first));
+    }
+    return item.second->as<std::string>()->value();
+}
+
+template <typename T>
```
MonsieurNicolas: nice addition! long overdue

src/main/Config.cpp
```diff
@@ -189,196 +233,73 @@ Config::load(std::string const& filename)
...
-                {
-                    throw std::invalid_argument("invalid FAILURE_SAFETY");
-                }
-                int64_t f = item.second->as<int64_t>()->value();
-                if (f < -1 || f >= INT32_MAX)
-                {
-                    throw std::invalid_argument("invalid FAILURE_SAFETY");
-                }
-                FAILURE_SAFETY = (int32_t)f;
+                FAILURE_SAFETY = readInt<int32_t>(item, 0, INT32_MAX - 1);
```
MonsieurNicolas: -1 is valid

src/main/Config.cpp
```diff
@@ -414,116 +331,39 @@ Config::load(std::string const& filename)
...
-                    item.second->as<int64_t>()->value();
-                if (parsedTargetPeerConnections <= 0 ||
-                    parsedTargetPeerConnections > UINT16_MAX)
-                {
-                    throw std::invalid_argument(
-                        "invalid TARGET_PEER_CONNECTIONS");
-                }
-                TARGET_PEER_CONNECTIONS =
-                    static_cast<unsigned short>(parsedTargetPeerConnections);
+                TARGET_PEER_CONNECTIONS = readInt<unsigned short>(item);
```
MonsieurNicolas: 0 is not allowed

src/main/Config.cpp
```diff
@@ -414,116 +331,39 @@ Config::load(std::string const& filename)
...
-                int64_t parsedMaxPeerConnections =
-                    item.second->as<int64_t>()->value();
-                if (parsedMaxPeerConnections <= 0 ||
-                    parsedMaxPeerConnections > UINT16_MAX)
-                {
-                    throw std::invalid_argument("invalid MAX_PEER_CONNECTIONS");
-                }
-                MAX_PEER_CONNECTIONS =
-                    static_cast<unsigned short>(parsedMaxPeerConnections);
+                MAX_PEER_CONNECTIONS = readInt<unsigned short>(item);
```
MonsieurNicolas: 0 is not allowed

src/main/Config.cpp
```diff
@@ -592,24 +428,11 @@ Config::load(std::string const& filename)
...
             else if (item.first == "MAX_CONCURRENT_SUBPROCESSES")
             {
-                if (!item.second->as<int64_t>())
-                {
-                    throw std::invalid_argument(
-                        "invalid MAX_CONCURRENT_SUBPROCESSES");
-                }
-                MAX_CONCURRENT_SUBPROCESSES =
-                    (size_t)item.second->as<int64_t>()->value();
+                MAX_CONCURRENT_SUBPROCESSES = readInt<size_t>(item);
```
MonsieurNicolas: I imagine that 0 should not be allowed

vogel:
Updated

MonsieurNicolas:
r+ 864c7963cd22ed33e6029e473bf4228406dfccfb

MonsieurNicolas:
@latobarita: retry

PR #1514, opened 2018-01-26
fix typo in admin.m

nicokunz:
The word "to" is missing from the sentence. I had included a suggested replacement for the whole paragraph below.

It would be better to change the two sentences from:
Here, each player can decide to look wherever they want. Even look at only one
other player, which is a risk if this other player decides to quit the game.

To:
Here each player can decide to look wherever they want, even if they can only see one other player (or nobody). This is risky if that player quits the game or is not following the rules.

### DIFF
--- a/docs/software/admin.md
+++ b/docs/software/admin.md
@@ -280,8 +280,8 @@ The quorum here is the set of players that end up with the winning hat color.
 
 A traditional quorum has people see everybody else (placed on a circle, at
 the right distance).
-Here, each player can decide to look wherever they want. Even to look at only one
-other player, which is a risk if this other player decides to quit the game.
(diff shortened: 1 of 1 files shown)


MonsieurNicolas:
yeah I agree, you can make the change.
Note that in order to open a pull request you need to sign the form, see https://github.com/stellar/stellar-core/blob/master/CONTRIBUTING.md#submitting-changes

nicokunz:
I have signed the agreement and updated the text.

MonsieurNicolas:
cool thank you. please squash the commits so that we can merge it as one change

MonsieurNicolas:
this section of the doc was removed in the new version of the admin guide

PR #1519, opened 2018-01-30
Update sequence number processing

### LINKED ISSUE #1445: Decouple sequence number consumption out of fee processing
This is a pre-requisite to implementing `BumpSequenceOp` described in https://github.com/stellar/stellar-protocol/issues/53

MonsieurNicolas:
resolves #1445 

this PR changes the way we update the sequence number in the account in preparation for the `BumpSeqOperation`:
* it moves sequence number processing from the pre-apply stage (where fees are processed) to the actual transaction apply time (protocol v10)
* it also changes the meta format for all versions. generating a "tx level side effect" meta. this field only gets updated starting at protocol 10, but I figured it was better to move to the new meta in all cases to avoid carrying over the complexity in future versions of Horizon (old meta support can be removed later)

### DIFF
--- a/src/transactions/TransactionFrame.cpp
+++ b/src/transactions/TransactionFrame.cpp
@@ -209,11 +209,10 @@ TransactionFrame::resetResults()
 }
 
 bool
-TransactionFrame::commonValid(SignatureChecker& signatureChecker,
-                              Application& app, LedgerDelta* delta,
-                              SequenceNumber current)
+TransactionFrame::commonValidPreSeqNum(Application& app, LedgerDelta* delta)
 {
-    bool applying = (delta != nullptr);
+    // this function does validations that are independent of the account state
+    //    (stay true regardless of other side effects)
 
     if (mOperations.size() == 0)
     {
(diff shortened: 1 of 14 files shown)


src/transactions/TransactionFrame.cpp
```diff
@@ -285,6 +301,11 @@ TransactionFrame::commonValid(SignatureChecker& signatureChecker,
             getResult().result.code(txBAD_SEQ);
             return false;
         }
+        if (applying)
```
vogel: I don't like the idea of commonValid function changing anything. The name implies that it is pure function, but now it is not. I think we should drop idea of having `common` valid method with `applying` parameter and try to do something different here.

We call `commonValid` from two places:
* `checkValid` to check validity before accepting transaction and
* `apply` to check validity when applying transaction

And it the middle of it is fee and seq number processing (but only if applying is true!).
If we split that in 3 steps code could became clearer:

* `checkPreValid` (needs better name - it is `commonValidPreSeqNum` now)
* `checkSeqNumValid`
* `checkPostValid` (needs better name and additional parameter `balanceAfter`)

Then checkValid could do something like this:
```
if (!checkPreValid()) return false;
if (!checkSeqNumValid()) return false;
if (!checkPostValid(mSigningAccount->getAccount().balance - mEnvelope.tx.fee)) return false;
```

And apply could do something like this:
```
if (!checkPreValid()) return false;
if (lm.getCurrentLedgerVersion() >= 10)
{
  if (!checkSeqNumValid()) return false;
  setSeqNum() and storeChange();
}
if (!checkPostValid(lm.getCurrentLedgerVersion() > 8 ? mSigningAccount->getAccount().balance - mEnvelope.tx.fee : mSigningAccount->getAccount().balance)) return false;
```

I think it will be much cleared that all those conditions with `applying` and account change inside `validCommon` method.

MonsieurNicolas:
@vogel I updated based on your feeback. I ended up changing back "Check" to only do checks (and moved the side effect code outside), it avoids introducing new functions and semantics that make it hard to work across protocol versions

vogel:
I like the changes! Now we only need to do full catchup with that ;)

PR #1520, opened 2018-01-31
updated sqlite to 3.21.0

MonsieurNicolas:
just like the title says

sqlite was not upgraded for a while and there has been quite a few improvements

This version was released last October and seems stable


vogel:
r+ 239dda076a1c751000aafc208f6335ac80779ba3

PR #1525, opened 2018-02-05
fix issues with flood tests

vogel:
There were two issues with flood tests:

1. SCP envelope was injected into node before its txSet and qSet were available on that node, now the order is inversed (txSet and qSet are added before). To achieve that PendingEnvelopes from Herder were made visible to avoid checks for envelopes that need that data in Herder class

2. GetQSet messages are sent all at once for all nodes without any intervening crank. With TCP mode it could take 1.5 or more to prepare, inject and process all SCP envelopes (this includes creating new timers and sending TCP message). This means that in the first crank a lot of new GetQSet messages could be created because of timeout, usually to nodes that do not know anything about such QSet, which could result in a flood of those messages and further problems. This version uses virtual timer even for TCP mode, which solves this problem.

### DIFF
--- a/src/herder/Herder.h
+++ b/src/herder/Herder.h
@@ -111,6 +111,11 @@ class Herder
     // We are learning about a new envelope.
     virtual EnvelopeStatus recvSCPEnvelope(SCPEnvelope const& envelope) = 0;
 
+    // We are learning about a new fully-fetched envelope.
+    virtual EnvelopeStatus recvSCPEnvelope(SCPEnvelope const& envelope,
+                                           const SCPQuorumSet& qset,
+                                           TxSetFrame txset) = 0;
+
     // a peer needs our SCP state
     virtual void sendSCPStateToPeer(uint32 ledgerSeq, PeerPtr peer) = 0;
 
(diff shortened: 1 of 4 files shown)


MonsieurNicolas:
Ah the problem was so simple and it escaped us for so long! Thanks for the investigation.
I can live with the first part even though it's not super clean as it makes the test even more coupled with the herder implementation. Alternative would be to create a peer just to inject - but that sounds quite complicated as well.
Second one, introducing a new TCP mode in virtual clock mode does not sound like the right thing to do: in this case it may actually have uncovered a real problem.
I think the tests can simply be fixed by cranking between batches of injected work instead of cranking at the end, wouldn't that work?

MonsieurNicolas:
there is a better way for the first one that is simply add a new method to receive a fully specified SCP message (SCP envelope + quorum set + tx set), that way we don't expose the internals outside of herder

vogel:
Cranking does not work either. There is always a point where asio scheduling on one thread and 15 nodes cannot do its work.

So I'm leaving only first part of the fix with your suggestion about new method as it works fine on Linux.

It is possible that this will be more properly fixed with overlay redesign.

MonsieurNicolas:
r+ 03501558288cd328c44f3881b8fe913b6472e954

Issue #1532, opened 2018-02-06
Implement bumpSequence


MonsieurNicolas:
as per https://github.com/stellar/stellar-protocol/issues/53

(depends on #1445 )

vogel:
Can this one be closed?

PR #1533, opened 2018-02-07
BumpSequence

### LINKED ISSUE #1445: Decouple sequence number consumption out of fee processing

MonsieurNicolas:
Resolves #1445 #1532

This commit superseeds #1519 (that only contained the commits related to sequence number processing)

I didn't squash commits on purpose to preserve the timeline wrt various revs of the specification that led to the current implementation.

# The first batch of commits updates the way we process sequence numbers

note: I only made a minor adjustment to the last commit compared to #1519 

* it moves sequence number processing from the pre-apply stage (where fees are processed) to the actual transaction apply time (protocol v10)
* it also changes the meta format for all versions. generating a "tx level side effect" meta. this field only gets updated starting at protocol 10, but I figured it was better to move to the new meta in all cases to avoid carrying over the complexity in future versions of Horizon (old meta support can be removed later)
* notable change in test code is that we do more validations from `applyCheck` (previously we were skipping a bunch of validations if the failures were too early)

# After that we introduce `BumpSequence`
The first couple drafts had additional failure modes/conditions on the operation.

The latest one makes `BumpSequence` extremely simple and instead:
* allows the sequence number to jump to any arbitrary point in the future ; for this to work I implemented support for uint64 sequence numbers in SQL (which only stores *signed* integers so we "wrap around"). I don't see a better way to do this (other than just not use integers at all and store that column as an base32 string which may be even more confusing and requires a schema migration). The conversion from signed to unsigned may still cause problems but is in C++ a fully specified behavior (bits are the same). @nullstyle any thoughts on this?
* change `MergeAccount` to fail if recreating the account in the same ledger (for example in a subsequent operation in the same transaction) would lead to an account with a smaller sequence number. For this to work, I had to adjust some tests that were overlooking this fact.


src/transactions/MergeTests.cpp
```diff
@@ -127,7 +150,9 @@ TEST_CASE("merge", "[tx][merge]")
             REQUIRE(a1.getBalance() ==
                     a1Balance + b1Balance - txFrame->getFee());
             REQUIRE(!loadAccount(b1, *app, false));
-            REQUIRE(a1SeqNum == a1.loadSequenceNumber());
+            // a1 gets recreated with a sequence number based on the current
+            // ledger
+            REQUIRE(a1.loadSequenceNumber() == 0x300000000ull);
```
vogel: Won't it break catchup complete on current pubnet?

src/transactions/MergeTests.cpp
```diff
@@ -579,6 +616,30 @@ TEST_CASE("merge", "[tx][merge]")
...
+        closeLedgerOn(*app, 3, 1, 1, 2017);
         for_all_versions(*app, [&] { mergeFrom.merge(root); });
     }
+
+    SECTION("merge too far")
+    {
+        for_versions_from(10, *app, [&]() {
+            SequenceNumber maxSeqNum = app->getLedgerManager().getLedgerNum();
+            maxSeqNum = (maxSeqNum << 32) - 1;
+            maxSeqNum--; // a1.merge will increment at the transaction level
```
vogel: if we move that 2 lines down we could get rid of line 637

src/transactions/OperationFrame.cpp
```diff
@@ -153,6 +160,15 @@ OperationFrame::checkValid(SignatureChecker& signatureChecker, Application& app,
                            LedgerDelta* delta)
 {
     bool forApply = (delta != nullptr);
+    if (!isVersionSupported(app.getLedgerManager().getCurrentLedgerVersion()))
```
vogel: it is bad that we cannot move MANAGE_DATA_NOT_SUPPORTED_YET here as well

vogel:
What happened to this one?

```
Another thing that I also realized is that we should also probably make it fail instead of no-op:

in the context of bumpseq being the only operation in a transaction it doesn't matter
if bumpseq is one of several operations in a transaction. it allows to have safe assumptions on the workflow.
```

src/transactions/MergeTests.cpp
MonsieurNicolas: no it won't: what I did here is change the test to be explicit about what is going on by adding an extra `    closeLedgerOn(*app, 2, 1, 1, 2016); ` after setting up the original accounts.
the check `REQUIRE(a1SeqNum == a1.loadSequenceNumber())` was just misleading because the reason `a1SeqNum` was not changing was that the setup and the transaction being tested (with all its sub-operations) were occurring during the same ledger.

src/transactions/OperationFrame.cpp
MonsieurNicolas: yeah `MANAGE_DATA_NOT_SUPPORTED_YET` should not have been implemented like this in the first place and now we have to keep that behavior to replay old versions of the protocol

MonsieurNicolas:
Q on "Why `bumpSequence` doesn't fail when attempting to bump backwards": I am answering in https://github.com/stellar/stellar-protocol/issues/53 as it's not clear that it was dropped.

In the issue, I am also proposing moving `SequenceNumber`  at the protocol layer to `int64` across the board as the current approach is just too hacky and unsafe across languages.

@nullstyle any input especially on that last point?

nullstyle:
@MonsieurNicolas at first take, I think switching to `int64` for sequence numbers should be fine.  I'll do a little checking around just to further confirm.

src/test/TxTests.cpp
```diff
@@ -41,48 +42,72 @@ namespace txtest
...
+        auto deleted = delta.deleted();
+        REQUIRE(deleted.begin() == deleted.end());
+        auto modified = delta.modified();
+        REQUIRE(modified.begin() != modified.end());
+        int modifiedCount = 0;
+        for (auto m : modified)
         {
-            doApply = (code != txBAD_SEQ);
+            modifiedCount++;
+            REQUIRE(modifiedCount == 1);
```
vogel: cant we just check before loop if modified.size() == 1?

src/transactions/BumpSequenceOpFrame.cpp
```diff
@@ -0,0 +1,72 @@
...
+    app.getMetrics()
+        .NewMeter({"op-bump-sequence", "success", "apply"}, "operation")
+        .Mark();
+    return true;
+}
+
+bool
+BumpSequenceOpFrame::doCheckValid(Application& app)
+{
+    if (mBumpSequenceOp.bumpTo < 0)
```
vogel: bumping to 0 is OK (always no-op)?

src/transactions/BumpSequenceTests.cpp
```diff
@@ -0,0 +1,80 @@
...
+                auto newSeq = a.loadSequenceNumber() + 2;
+                a.bumpSequence(newSeq);
+                REQUIRE(a.loadSequenceNumber() == newSeq);
+            }
+            SECTION("large bump")
+            {
+                auto newSeq = INT64_MAX;
+                a.bumpSequence(newSeq);
+                REQUIRE(a.loadSequenceNumber() == newSeq);
+                SECTION("no more tx when INT64_MAX is reached")
```
vogel: so this is second way to lock account, next to setting master weight to 0, nice!

src/transactions/MergeOpFrame.cpp
```diff
@@ -92,6 +92,23 @@ MergeOpFrame::doApply(Application& app, LedgerDelta& delta,
         return false;
     }
 
+    if (ledgerManager.getCurrentLedgerVersion() >= 10)
+    {
+        SequenceNumber maxSeqNum =
+            ledgerManager.getCurrentLedgerHeader().ledgerSeq;
+        maxSeqNum = (maxSeqNum << 32) - 1;
```
vogel: I think a comment here explaining why this condition exists would be nice

src/transactions/MergeTests.cpp
vogel: In that case maybe remove line 637 and use maxSeqNum + 1 in 638 and 639?

src/transactions/BumpSequenceOpFrame.cpp
MonsieurNicolas: 0 is a valid sequence number, so yes

src/transactions/BumpSequenceTests.cpp
MonsieurNicolas: no it doesn't lock the account, it just cannot be used as source account for a transaction

src/transactions/MergeOpFrame.cpp
MonsieurNicolas: I forgot that I wanted to get rid of this magic across the code base

src/test/TxTests.cpp
MonsieurNicolas: it's a fake iterator, it only implements `begin` and `end`

MonsieurNicolas:
@vogel I updated that latest commit with your suggestions

MonsieurNicolas:
(rebased to current master)

vogel:
r+ f9f9187ed4721cec0f53bd208ba3451e61d7c0d9

Teri68 (approved):

Teri68 (approved):

PR #1535, opened 2018-02-08
Known cursors

### LINKED ISSUE #1508: add a way to populate cursors when setting up a node
# Description of the problem

As we're about to implement #923 a race condition gets created when setting up an environment with a consumer such as Horizon:
if it takes a while to start up Horizon after core is started, for example because it takes a long time for core to catchup to the current ledger (in catchup_complete mode), there is a chance that the automatic maintenance gets executed even though Horizon didn't run yet.

The problem here is that the cursor that Horizon uses is not yet registered with core as Horizon does that when it first runs, causing some data to be deleted before Horizon gets a chance to ingest it.

# Approach
The solution is to add a mechanism for core to have the known cursors for a given environment added from the get go.

A couple ways this can be done (to be discussed before moving on to implementation):
* add new options to `newdb`, core can then be invoked like this
```
stellar-core --newdb --cursors HORIZON,FOO
```
This version couples the functionality to `newdb` which may or may not be desirable. It also bloats the command line even more.

* add a new option to `setcursor` for setting a cursor only if it doesn't exist, and add the following to core's config file
```
COMMANDS=[
"setcursor?id=HORIZON&cursor=1&createonly=true" , 
"setcursor?id=FOO&cursor=1&createonly=true"
]
```
This version is very simple to implement but allows cursors to be created at any time.

@nullstyle if you think about use cases, which version works for you best? My preference is the second one, but I'd like to hear your opinion.

vogel:
Fixes #1508

### DIFF
--- a/docs/stellar-core_example.cfg
+++ b/docs/stellar-core_example.cfg
@@ -150,6 +150,10 @@ KNOWN_PEERS=[
 "core-testnet2.stellar.org",
 "core-testnet3.stellar.org"]
 
+# KNOWN_CURSORS (list of strings) default is empty
+# Set of cursors added at each startup with value '1'.
+KNOWN_CURSORS=["HORIZON"]
+
 #######################
 ##  SCP settings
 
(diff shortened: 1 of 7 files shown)


MonsieurNicolas:
r+ 715eee8823ac32006c456aca5cbb11bdeef3ec81

PR #1537, opened 2018-02-09
Cleanup stale buckets

### LINKED ISSUE #1295: Investigate large number of buckets in bucket directory

vogel:
Fixes #1295

### DIFF
--- a/src/bucket/BucketManagerImpl.cpp
+++ b/src/bucket/BucketManagerImpl.cpp
@@ -71,10 +71,22 @@ BucketManagerImpl::BucketManagerImpl(Application& app)
 
 const std::string BucketManagerImpl::kLockFilename = "stellar-core.lock";
 
-static std::string
+namespace
+{
+std::string
 bucketBasename(std::string const& bucketHexHash)
 {
     return "bucket-" + bucketHexHash + ".xdr";
+}
+
+bool isBucketFile(std::string const& name) {
+    static std::regex re("^bucket-[a-z0-9]{64}\\.xdr(\\.gz)?$");
+    return std::regex_match(name, re);
+};
+
+uint256 extractFromFilename(std::string const& name) {
+    return hexToBin256(name.substr(7, 64));
+};
 }
 
 std::string
@@ -253,14 +265,6 @@ BucketManagerImpl::cleanupStaleFiles()
                    [](std::pair<Hash, std::shared_ptr<Bucket>> const& p) {
                        return p.first;
                    });
-
-    auto isBucketFile = [](std::string const& name) {
-        static std::regex re("^bucket-[a-z0-9]{64}\\.xdr(\\.gz)?$");
-        return std::regex_match(name, re);
-    };
-    auto extractFromFilename = [](std::string const& name) {
-        return hexToBin256(name.substr(7, 64));
-    };
 
     for (auto f : fs::findfiles(getBucketDir(), isBucketFile))
     {
--- a/src/bucket/BucketManagerImpl.cpp
+++ b/src/bucket/BucketManagerImpl.cpp
@@ -16,6 +16,7 @@
 #include "util/types.h"
 #include <fstream>
 #include <map>
+#include <regex>
 #include <set>
 
 #include "medida/counter.h"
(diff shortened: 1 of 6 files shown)


src/bucket/BucketManagerImpl.cpp
```diff
@@ -240,6 +240,48 @@ BucketManagerImpl::forgetUnreferencedBuckets()
...
+    return referenced;
+}
+
+void
+BucketManagerImpl::cleanupStaleFiles()
+{
+    std::lock_guard<std::recursive_mutex> lock(mBucketMutex);
+    auto referenced = getReferencedBuckets();
+    std::transform(std::begin(mSharedBuckets), std::end(mSharedBuckets),
+                   std::inserter(referenced, std::end(referenced)),
```
MonsieurNicolas: `std::back_inserter` is what you want here

src/util/Fs.h
```diff
@@ -40,6 +42,11 @@ bool mkdir(std::string const& path);
 // Make a dir path like mkdir -p, i.e. recursive, uses '/' as dir separator
 bool mkpath(std::string const& path);
 
+// Get list of all files with names matching predicate
```
MonsieurNicolas: add comment that file names are relative to path

src/bucket/BucketManagerImpl.cpp
vogel: It does not work on std::set, so I'm using regular inserter

vogel:
Updated.

MonsieurNicolas:
I have a commit ready to add to this pr (Windows support) but I don't have rights to push to your branch @vogel

MonsieurNicolas:
I'm wondering if we should delete buckets that are older than let's say a day to avoid re-downloading buckets for no reason between aborted catchup. Or are those buckets in a temp folder anyways?

MonsieurNicolas:
@vogel see my question above

vogel:
I don't think I understand that question.

We are deleting buckets that we are not referencing to, so why should we delete all buckets older than one day? Some buckets change more rarely than once a day and we should keep them.

Which buckets do you have in mind?

MonsieurNicolas:
yes, what I am asking is if we should keep buckets that are more recent than a day in the event that we restart stellar core during catchup to avoid re-downloading a bunch of stuff from history. If we place those buckets in temp folders, then this won't change anything, but I am not sure without going through the code.

vogel:
It uses tmp dir, see `BucketDownloadWork::BucketDownloadWork`.

MonsieurNicolas:
yeah that's what I suspected. Thanks for looking

MonsieurNicolas:
r+ 2aecca8f95bf9c8441bc9b0e1851231d0fa81a3b

MonsieurNicolas:
@latobarita: retry

MonsieurNicolas:
@latobarita: retry

Issue #1488, opened 2018-01-12
flaky test: "Catchup manual"


MonsieurNicolas:
This test seems to take arbitrary long time to complete, which in some cases cause CI to timeout

MonsieurNicolas:
@vogel 
we're still hitting this quite often. I also suspect that some of the integration flakiness is related:
what I see there is that sometimes nodes take much longer than what I would expect to download a snapshot, as if they were trying to download the wrong ledger (like attempt to do download the next checkpoint)

PR #1538, opened 2018-02-09
cleanup parseNumParam

### LINKED ISSUE #1469: cleanup parseNumParam (CommandHandler)
function signature is bad - we should make it throw now that all handlers are dealing with exceptions properly.
New signature should probably return an `optional<>` and takes as parameter a string in case of error

marta-lokhova:
resolves #1469 

* Update the signature to return optional<T> and removed redundant arguments
* Throw an exception in case arg can't be loaded; Note that if a number param followed by garbage (e.g. non-numeric characters), an exception will be thrown also
* tested affected endpoints with various parameters my running stellar-core locally

### DIFF
--- a/src/main/CommandHandler.cpp
+++ b/src/main/CommandHandler.cpp
@@ -357,43 +357,43 @@ CommandHandler::manualClose(std::string const& params, std::string& retStr)
     }
 }
 
-enum class Requirement
-{
-    OPTIONAL_REQ,
-    REQUIRED_REQ
-};
-
 template <typename T>
-bool
-parseNumParam(std::map<std::string, std::string> const& map,
-              std::string const& key, T& val, std::string& retStr,
-              Requirement requirement, bool& valueUpdated)
+optional<T>
+maybeParseNumParam(std::map<std::string, std::string> const& map,
+                   std::string const& key, T& defaultVal)
 {
-    valueUpdated = false;
     auto i = map.find(key);
     if (i != map.end())
     {
         std::stringstream str(i->second);
-        str >> val;
-        if (val == 0)
+        str >> defaultVal;
+
+        // Throw an error if not all bytes were loaded into `val`
+        if (str.fail() || !str.eof())
         {
-            retStr = fmt::format("Failed to parse '{}' argument", key);
-            return false;
+            std::string errorMsg =
+                fmt::format("Failed to parse '{}' argument", key);
+            throw std::runtime_error(errorMsg);
         }
-        valueUpdated = true;
-        return true;
+        return make_optional<T>(defaultVal);
     }
-    return requirement == Requirement::OPTIONAL_REQ;
+
+    return nullopt<T>();
 }
 
 template <typename T>
-bool
+T
 parseNumParam(std::map<std::string, std::string> const& map,
-              std::string const& key, T& val, std::string& retStr,
-              Requirement requirement)
+              std::string const& key)
 {
-    bool valueUpdated;
-    return parseNumParam(map, key, val, retStr, requirement, valueUpdated);
+    T val;
+    auto res = maybeParseNumParam(map, key, val);
+    if (!res)
+    {
+        std::string errorMsg = fmt::format("'{}' argument is required!", key);
+        throw std::runtime_error(errorMsg);
+    }
+    return val;
 }
 
 void
@@ -412,12 +412,8 @@ CommandHandler::generateLoad(std::string const& params, std::string& retStr)
         std::map<std::string, std::string> map;
         http::server::server::parseParams(params, map);
 
-        if (!parseNumParam(map, "accounts", nAccounts, retStr,
-                           Requirement::OPTIONAL_REQ))
-            return;
-
-        if (!parseNumParam(map, "txs", nTxs, retStr, Requirement::OPTIONAL_REQ))
-            return;
+        maybeParseNumParam(map, "accounts", nAccounts);
+        maybeParseNumParam(map, "txs", nTxs);
 
         {
             auto i = map.find("txrate");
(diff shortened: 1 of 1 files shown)


src/main/CommandHandler.cpp
```diff
@@ -364,36 +364,27 @@ enum class Requirement
 };
 
 template <typename T>
-bool
+optional<T>
 parseNumParam(std::map<std::string, std::string> const& map,
-              std::string const& key, T& val, std::string& retStr,
-              Requirement requirement, bool& valueUpdated)
+              std::string const& key, T& val)
```
MonsieurNicolas: you don't need to pass in `val` anymore.
If people want to do default values, I would expect callers to do things like
```
int n = 1000; // default value
auto v = parseNumParam<int>(m, "n");
if (v)
{
    n = *v;
}
```
non optional case is simply:
```
int n = *parseNumParam<int>(m, "n");
```

src/main/CommandHandler.cpp
```diff
@@ -364,36 +364,27 @@ enum class Requirement
...
-    valueUpdated = false;
     auto i = map.find(key);
     if (i != map.end())
     {
         std::stringstream str(i->second);
         str >> val;
-        if (val == 0)
+
+        // Throw an error if not all bytes were loaded into `val`
+        if (str.fail() || str.rdbuf()->in_avail() > 0)
```
MonsieurNicolas: isn't the second condition just `!str.eof()`?

src/main/CommandHandler.cpp
```diff
@@ -725,17 +710,8 @@ CommandHandler::upgrades(std::string const& params, std::string& retStr)
...
-            {
-                retStr = (fmt::MemoryWriter()
-                          << fmt::format("could not parse {}: '{}'\n", name,
-                                         retMap[name])
-                          << retStr)
-                             .str();
-            }
-            else if (updated)
+            auto res = parseNumParam(retMap, name, v);
+            if (res)
```
MonsieurNicolas: isn't that code equivalent to `f = parseNumParam(...)` ?
If so, this lambda is now obsolete

src/main/CommandHandler.cpp
```diff
@@ -950,8 +926,7 @@ CommandHandler::setcursor(std::string const& params, std::string& retStr)
 
     uint32 cursor;
 
-    if (!parseNumParam(map, "cursor", cursor, retStr,
-                       Requirement::REQUIRED_REQ))
+    if (!parseNumParam(map, "cursor", cursor))
     {
```
MonsieurNicolas: this code can be deleted

marta-lokhova: Don't we need this check to gracefully handle the case when `cursor` is not passed in the request? With the updated logic based on your comments, it would be something like 
```
    uint32 cursor;
    auto res = parseNumParam(map, "cursor");

    if (!res)
    {
        retStr = "Invalid cursor";
        return;
    }
    else
    {
        cursor = *res;
    }
```

MonsieurNicolas: If it’s a required parameter this should be taken care of by `parseNumParam`

src/main/CommandHandler.cpp
marta-lokhova: I think you're right. `reset()` would result in the same behavior as assigning `nullopt` (which is `shared_ptr`() under the hood)

marta-lokhova:
Thanks for reviewing! I've updated the PR based on the comments.

src/main/CommandHandler.cpp
```diff
@@ -363,28 +363,53 @@ enum class Requirement
     REQUIRED_REQ
 };
 
-template <typename T>
-optional<T>
+optional<uint32>
```
MonsieurNicolas: why did you remove the template?

src/main/CommandHandler.cpp
```diff
@@ -363,28 +363,53 @@ enum class Requirement
...
-        return make_optional<T>(val);
+        return make_optional<uint32>(val);
     }
 
-    return nullopt<T>();
+    return nullopt<uint32>();
+}
+
+optional<uint32>
+parseNumParam(std::map<std::string, std::string> const& map,
```
MonsieurNicolas: After looking at the changes overall, I don't think this contract makes sense to me:
we already have a `parseNumParam` that returns an `optional<T>` (that can be used for anything).

From the use cases point of view, I think it would probably make more sense to rename the optional version to `maybeParseNumParam` and add a dedicated method to deal with the non-optional case:
in this case, the method's signature would not return an optional, and there is no default value.

Then it probably makes sense to update the `maybeParseNumParam` that returns an optional take care of the default value (and have the other one call it with a default value internally).

src/main/CommandHandler.cpp
```diff
@@ -363,28 +363,53 @@ enum class Requirement
...
+              std::string const& key, uint32& defaultVal,
+              Requirement requirement)
+{
+    auto val = parseNumParam(map, key);
+    if (val)
+    {
+        if (requirement == Requirement::OPTIONAL_REQ)
+        {
+            defaultVal = *val;
```
MonsieurNicolas: that logic doesn't make sense to me: it's updating `defaultVal`?

MonsieurNicolas:
cool, can you rebase and squash commit and we'll get this in

marta-lokhova:
Sure, it's been updated.

MonsieurNicolas:
r+ 4a9ead9d06f98e188742aaae9e58b8675bfb354c

MonsieurNicolas:
@latobarita: retry

MonsieurNicolas:
hit by #1488

MonsieurNicolas:
@latobarita: retry

MonsieurNicolas:
maybe running that test with a combination of the shell commands `taskset ...` and `cpulimit -i ...`  can help diagnose this outside of travis? @vogel

PR #1544, opened 2018-02-13
Remove scp dependency from driver

vogel:


MonsieurNicolas:
I don't think we'll take this PR until we understand better the changes from overlay as this moves more logic into Herder (and I'd like Herder to be smaller not bigger).

PR #1549, opened 2018-02-15
Scp timeout update

### LINKED ISSUE #1548: SCP rounds are timing out too aggressively
This looks similar to #1333 (not as bad - this bug is only causing delays of a few seconds) but analysis of SCP logs show that the root cause is different:
from what I could see, nodes tend to timeout and bump their ballot counter too aggressively.

Net effect is that rounds eventually complete but only after raising their timeouts long enough that all nodes get to participate at the same time.

After looking at the code, it looks like we're actually not following exactly what the whitepaper says:
we're supposed to only timeout when there is a quorum with a higher (or equal) counter, not before.

relevant section in the whitepaper is "6.2.2. Timeouts and ballot updates"

MonsieurNicolas:
Resolves #1548 #1550 

This PR updates the SCP timeout code in the Ballot Protocol implementation to wait for a quorum (and not have a timeout if there is no quorum)

It also addresses an issue where we potentially trigger a ledger close too early

### DIFF
--- a/src/herder/HerderImpl.cpp
+++ b/src/herder/HerderImpl.cpp
@@ -75,7 +75,6 @@ HerderImpl::HerderImpl(Application& app)
     , mHerderSCPDriver(app, *this, mUpgrades, mPendingEnvelopes)
     , mLastSlotSaved(0)
     , mTrackingTimer(app)
-    , mLastTrigger(app.getClock().now())
     , mTriggerTimer(app)
     , mRebroadcastTimer(app)
     , mApp(app)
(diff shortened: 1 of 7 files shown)


MonsieurNicolas:
I just added a commit fixes a bug that I am surprised that asan didn't catch. It's a bug currently on master that breaks SCP tests (temp variable lifetime issue)

MonsieurNicolas:
I moved the commits unrelated to this change into #1553 as there was more breakage in master (tests don't pass without the other PR merged)

jedmccaleb:
r+ 94e9e84

MonsieurNicolas:
github's confusing UI didn't have the right commit head

MonsieurNicolas:
r+ 94e9e840e92700cc90c95572682209e0fbcb8649

MonsieurNicolas:
@latobarita: retry

Issue #582, opened 2015-07-07
TxSetFrame::trimInvalid and checkValid need to be refactored


MonsieurNicolas:
they share a lot of code

evanlimanto:
I believe this issue can be closed now?

MonsieurNicolas:
yes it can be. github will auto close issues if you use "resolves #N" instead of "Addresses #N" like you did in the PR. See https://help.github.com/articles/closing-issues-using-keywords/ for a list of keywords

evanlimanto:
got it, thanks.

PR #1552, opened 2018-02-16
Refactor TxSetFrame to reuse common code

evanlimanto:
Addresses #582. Essentially moved code from TxSetFrame::trimInvalid and TxSetFrame::checkValid into a helper function checkOrTrim, and called this function instead.

### DIFF
--- a/src/herder/TxSetFrame.cpp
+++ b/src/herder/TxSetFrame.cpp
@@ -186,98 +186,19 @@ TxSetFrame::surgePricingFilter(LedgerManager const& lm)
     }
 }
 
-// TODO.3 this and checkValid share a lot of code
-void
-TxSetFrame::trimInvalid(Application& app,
-                        std::vector<TransactionFramePtr>& trimmed)
-{
-    soci::transaction sqltx(app.getDatabase().getSession());
-    app.getDatabase().setCurrentTransactionReadOnly();
-
-    sortForHash();
-
-    map<AccountID, vector<TransactionFramePtr>> accountTxMap;
-
-    for (auto tx : mTransactions)
-    {
-        accountTxMap[tx->getSourceID()].push_back(tx);
-    }
-
-    for (auto& item : accountTxMap)
-    {
-        // order by sequence number
-        std::sort(item.second.begin(), item.second.end(), SeqSorter);
-
-        TransactionFramePtr lastTx;
-        SequenceNumber lastSeq = 0;
-        int64_t totFee = 0;
-        for (auto& tx : item.second)
-        {
-            if (!tx->checkValid(app, lastSeq))
-            {
-                trimmed.push_back(tx);
-                removeTx(tx);
-                continue;
-            }
-            totFee += tx->getFee();
-
-            lastTx = tx;
-            lastSeq = tx->getSeqNum();
-        }
-        if (lastTx)
-        {
-            // make sure account can pay the fee for all these tx
-            int64_t newBalance =
-                lastTx->getSourceAccount().getBalance() - totFee;
-            if (newBalance < lastTx->getSourceAccount().getMinimumBalance(
-                                 app.getLedgerManager()))
-            {
-                for (auto& tx : item.second)
-                {
-                    trimmed.push_back(tx);
-                    removeTx(tx);
-                }
-            }
-        }
-    }
-}
-
-// need to make sure every account that is submitting a tx has enough to pay
-// the fees of all the tx it has submitted in this set
-// check seq num
 bool
-TxSetFrame::checkValid(Application& app) const
+TxSetFrame::checkOrTrim(
+    Application& app,
+    std::function<bool(TransactionFramePtr, SequenceNumber)>
+        processInvalidTxLambda,
+    std::function<bool(std::vector<TransactionFramePtr> const&)>
+        processInsufficientBalance)
 {
-    // Establish read-only transaction for duration of checkValid.
-    soci::transaction sqltx(app.getDatabase().getSession());
-    app.getDatabase().setCurrentTransactionReadOnly();
-
-    auto& lcl = app.getLedgerManager().getLastClosedLedgerHeader();
-    // Start by checking previousLedgerHash
-    if (lcl.hash != mPreviousLedgerHash)
-    {
-        CLOG(DEBUG, "Herder")
-            << "Got bad txSet: " << hexAbbrev(mPreviousLedgerHash)
-            << " ; expected: "
-            << hexAbbrev(
-                   app.getLedgerManager().getLastClosedLedgerHeader().hash);
-        return false;
-    }
-
-    if (mTransactions.size() > lcl.header.maxTxSetSize)
-    {
-        CLOG(DEBUG, "Herder")
-            << "Got bad txSet: too many txs " << mTransactions.size() << " > "
-            << lcl.header.maxTxSetSize;
-        return false;
-    }
-
     map<AccountID, vector<TransactionFramePtr>> accountTxMap;
 
     Hash lastHash;
-    for (auto tx : mTransactions)
+    for (auto& tx : mTransactions)
     {
-        // make sure the set is sorted correctly
         if (tx->getFullHash() < lastHash)
         {
             CLOG(DEBUG, "Herder")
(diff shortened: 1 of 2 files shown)


MonsieurNicolas:
cool, do you think we could avoid having the branching in this common function and instead move the common code into a helper function that builds and iterates over `accountTxMap` (maybe pass in a lambda from the `checkValid` and `trimInvalid` functions to specialize the behavior)? Equivalent solution could be to have a function to populate `accountTxMap` and return an object that allows to iterate over the elements.

src/herder/TxSetFrame.cpp
```diff
@@ -323,18 +295,31 @@ TxSetFrame::checkValid(Application& app) const
...
+                    return false;
+                }
             }
         }
     }
     return true;
 }
 
+
+
```
MonsieurNicolas: extra CR needs to be removed

evanlimanto:
updated code — created lambdas to pass into a helper function.

src/herder/TxSetFrame.cpp
```diff
@@ -186,21 +186,23 @@ TxSetFrame::surgePricingFilter(LedgerManager const& lm)
     }
 }
 
-// TODO.3 this and checkValid share a lot of code
-void
-TxSetFrame::trimInvalid(Application& app,
-                        std::vector<TransactionFramePtr>& trimmed)
+template <typename Lambda1, typename Lambda2, typename Lambda3>
```
MonsieurNicolas: you don't need the template here: lambdas have the same signature or is that to just to try to get better optimized code? Either way, I don't think it's necessary, the compiler will inline if it needs to

src/herder/TxSetFrame.cpp
```diff
@@ -186,21 +186,23 @@ TxSetFrame::surgePricingFilter(LedgerManager const& lm)
...
-
-    sortForHash();
-
     map<AccountID, vector<TransactionFramePtr>> accountTxMap;
 
-    for (auto tx : mTransactions)
+    Hash lastHash;
+    for (auto& tx : mTransactions)
     {
+        if (!checkTxLambda(tx, lastHash)) {
```
MonsieurNicolas: You can ditch the checkTxLambda and always check: it will add a very small overhead to the "Trim" path but the code ends up being simpler  and "Trim" is not called that often

src/herder/TxSetFrame.cpp
```diff
@@ -230,25 +233,55 @@ TxSetFrame::trimInvalid(Application& app,
             int64_t newBalance =
                 lastTx->getSourceAccount().getBalance() - totFee;
             if (newBalance < lastTx->getSourceAccount().getMinimumBalance(
-                                 app.getLedgerManager()))
+                  app.getLedgerManager()) && !processLastInvalidTxLambda(item.second, lastTx))
```
MonsieurNicolas: I think `processLastInvalidTxLambda` should be:
* renamed to `processInsufficientBalance`
* move the call inside the block (right before returning `false`), as Trim doesn't care about the return value
* ditch the second parameter (lastTx) as it's I think equal to item.second.back()

evanlimanto:
thanks for the comments! I've fixed the code.

src/herder/TxSetFrame.cpp
```diff
@@ -186,21 +186,25 @@ TxSetFrame::surgePricingFilter(LedgerManager const& lm)
...
 }
 
-// TODO.3 this and checkValid share a lot of code
-void
-TxSetFrame::trimInvalid(Application& app,
-                        std::vector<TransactionFramePtr>& trimmed)
+bool
+TxSetFrame::checkOrTrim(Application &app,
+                        std::function<bool(TransactionFramePtr, SequenceNumber)> processInvalidTxLambda,
+                        std::function<bool(const std::vector<TransactionFramePtr>&)> processInsufficientBalance)
```
MonsieurNicolas: in the code base we prefer `T const&` instead of `const T&`

src/herder/TxSetFrame.cpp
```diff
@@ -230,25 +235,52 @@ TxSetFrame::trimInvalid(Application& app,
...
+
+    sortForHash();
+
+    auto processInvalidTxLambda = [&](TransactionFramePtr tx,
+                                      SequenceNumber lastSeq) -> bool {
+        trimmed.push_back(tx);
+        removeTx(tx);
+        return true;
+    };
+    auto processInsufficientBalance = [&](const vector<TransactionFramePtr> &item) -> bool {
```
MonsieurNicolas: same thing (const and `-> bool`)

src/herder/TxSetFrame.cpp
```diff
@@ -272,67 +304,26 @@ TxSetFrame::checkValid(Application& app) const
...
-                    << " tx invalid"
-                    << " lastSeq:" << lastSeq
-                    << " tx: " << xdr::xdr_to_string(tx->getEnvelope())
-                    << " result: " << tx->getResultCode();
-
-                return false;
-            }
-            totFee += tx->getFee();
+    auto processInvalidTxLambda = [&](TransactionFramePtr tx,
+                                      const SequenceNumber &lastSeq) -> bool {
```
MonsieurNicolas: (const and `-> bool`)

src/herder/TxSetFrame.cpp
```diff
@@ -272,67 +304,26 @@ TxSetFrame::checkValid(Application& app) const
...
-            if (newBalance < lastTx->getSourceAccount().getMinimumBalance(
-                                 app.getLedgerManager()))
-            {
-                CLOG(DEBUG, "Herder")
-                    << "bad txSet: " << hexAbbrev(mPreviousLedgerHash)
-                    << " account can't pay fee"
-                    << " tx:" << xdr::xdr_to_string(lastTx->getEnvelope());
+        return false;
+    };
+    auto processInsufficientBalance = [&](const vector<TransactionFramePtr> &item) -> bool {
```
MonsieurNicolas: (const and `-> bool`)

src/herder/TxSetFrame.h
```diff
@@ -41,9 +41,12 @@ class TxSetFrame
 
     std::vector<TransactionFramePtr> sortForApply();
 
-    bool checkValid(Application& app) const;
+    bool checkValid(Application& app);
     void trimInvalid(Application& app,
                      std::vector<TransactionFramePtr>& trimmed);
+    bool checkOrTrim(Application& app,
+                     std::function<bool(TransactionFramePtr, SequenceNumber)> processInvalidTxLambda,
+                     std::function<bool(const std::vector<TransactionFramePtr>&)> processLastInvalidTxLambda);
```
MonsieurNicolas: (const)

src/herder/TxSetFrame.cpp
```diff
@@ -230,25 +235,52 @@ TxSetFrame::trimInvalid(Application& app,
...
+                        std::vector<TransactionFramePtr>& trimmed)
+{
+    // Establish read-only transaction for duration of trimInvalid
+    soci::transaction sqltx(app.getDatabase().getSession());
+    app.getDatabase().setCurrentTransactionReadOnly();
+
+    sortForHash();
+
+    auto processInvalidTxLambda = [&](TransactionFramePtr tx,
+                                      SequenceNumber lastSeq) -> bool {
```
MonsieurNicolas: you can remove the `-> bool` as this type inference works fine in c++ 11

evanlimanto:
pushed the fixes!

MonsieurNicolas:
Great thank you!

MonsieurNicolas:
r+ c42efcf7d7dbf7a0355e46980593f1162076d646

evanlimanto:
sorry, i accidentally committed an extra file src/main/dumpxdr.cpp. i've removed from the PR, so it should be good now.

MonsieurNicolas:
np I missed the extra file!

MonsieurNicolas:
r+ d4e2258

evanlimanto:
seems like these are clang-format errors. i've re-updated the code.

MonsieurNicolas:
sorry, I just noticed that the new method is public, it should be private as it's used internally by the two other methods. Can you make that change please?

evanlimanto:
i've made the change. sorry i missed that!

MonsieurNicolas:
r+ 6bc12c7

PR #1558, opened 2018-02-22
Overlay err load improvements

### LINKED ISSUE #1560: Hard to connect to the network without having bootstrapped a list of peers
Right now if a node tries to join the network from fresh it only knows a very few number of peers.
If those peers are at capacity, they will reject connections and the new node is "locked out" of the network.

We should send a list of alternate nodes that may have capacity before shutting down the connection.

vogel:
Fixes #1560 

Send list of peers before dropping

### DIFF
--- a/src/overlay/Peer.cpp
+++ b/src/overlay/Peer.cpp
@@ -216,6 +216,11 @@ Peer::getIOTimeoutSeconds() const
 void
 Peer::receivedBytes(size_t byteCount, bool gotFullMessage)
 {
+    if (shouldAbort())
+    {
+        return;
+    }
+
     LoadManager::PeerContext loadCtx(mApp, mPeerID);
     mLastRead = mApp.getClock().now();
     if (gotFullMessage)
@@ -284,8 +289,9 @@ Peer::drop(ErrorCode err, std::string const& msg)
     sendMessage(m);
     // note: this used to be a post which caused delays in stopping
     // to process read messages.
-    // this has no effect wrt the sending queue.
-    drop();
+    // this will try to send all data from send queue if the error is ERR_LOAD
+    // - it sends list of peers
+    drop(err != ERR_LOAD);
 }
 
 void
--- a/src/overlay/Peer.cpp
+++ b/src/overlay/Peer.cpp
@@ -1063,6 +1063,12 @@ Peer::recvAuth(StellarMessage const& msg)
 
     auto self = shared_from_this();
 
+    if (mRole == REMOTE_CALLED_US)
+    {
+        sendAuth();
+        sendPeers();
+    }
+
     if (!mApp.getOverlayManager().acceptAuthenticatedPeer(self))
     {
         CLOG(WARNING, "Overlay") << "New peer rejected, all slots taken";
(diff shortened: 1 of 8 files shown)


src/overlay/TCPPeer.cpp
```diff
@@ -382,48 +446,10 @@ TCPPeer::drop()
...
-            asio::error_code ec2;
-            self->mSocket->close(ec2);
-            if (ec2)
-            {
-                CLOG(ERROR, "Overlay")
-                    << "TCPPeer::drop close socket failed: " << ec2.message();
-            }
-        });
-    });
+    // if write queue is not empty, messageSender will take care of shutdown
```
MonsieurNicolas: > messageSender will take care of shutdown

this is not always the case:
* see my comment in `messageSender` on how to fix the "normal shutdown"
* you also need to keep the `mIdleTimer` in case writes "never" complete. Note that during shutdown we do not want to reset that timer when receiving data - I would suggest looking for all calls to `receivedBytes` and make sure that we stop processing reads in each of those functions with:
```
    if (shouldAbort())
    {
        return;
    }
```

src/overlay/Peer.cpp
```diff
@@ -1061,6 +1061,11 @@ Peer::recvAuth(StellarMessage const& msg)
     {
         CLOG(WARNING, "Overlay") << "New peer rejected, all slots taken";
         mDropInRecvAuthRejectMeter.Mark();
```
MonsieurNicolas: wouldn't it be better to just move this code (if condition etc) to right before `// send SCP State` ?
This would avoid duplicating the handshaking code (and in this case - the handshake was successful we're dropping the peer for another reason).

src/overlay/ConnectionTests.cpp
```diff
@@ -0,0 +1,73 @@
+// Copyright 2018 Stellar Development Foundation and contributors. Licensed
```
MonsieurNicolas: new file -> please update Visual C++ project

MonsieurNicolas: now that I think of it, I think we should just put this new test in the `OverlayTests.cpp` file as it's just another variation of the existing tests (and tests there are basically testing connections).

src/overlay/TCPPeer.cpp
MonsieurNicolas: Also as a condition, checking if `mWriting` is true is cleaner

MonsieurNicolas: Also, we may want to add a parameter to `drop` to shutdown immediately (probably the default) as opposed to do this delayed shutdown (in case of timeout like with `Peer::idleTimerExpired` or bad behavior we probably want to stop the connection asap and not wait for flushing).

vogel:
updated

src/overlay/Peer.h
```diff
@@ -267,7 +267,10 @@ class Peer : public std::enable_shared_from_this<Peer>,
     }
 
     void drop(ErrorCode err, std::string const& msg);
-    virtual void drop() = 0;
+
+    // If force is true, it will drop immiediately without waiting for all
```
MonsieurNicolas: typo `immiediately`

MonsieurNicolas:
r+ a0ec98877a7a0a981792968ae24d4c42a1d4f40e

PR #1562, opened 2018-02-23
 increase overlay version to 6

### LINKED ISSUE #1507: Bump olver
Even though we've not made changes to overlay, we should bump olver so that we can "shake out" nodes that have not been updated in a very long time in the future (we'll make olver the min version at a later date) by disconnecting nodes running very old versions of stellar-core:
we still see nodes running versions as old as 0.6.1 - that cannot possibly be used for anything useful as they do not understand the protocol used by the network (and as such cannot validate anything).

vogel:
Fixes #1507


MonsieurNicolas:
r+ 0020344e73c24bd3abce2afbc0b3adf012ff1fda

PR #1564, opened 2018-02-26
Small README fix for DB-schemas -- peer rank not used

robertDurst:
Ran into this while experimenting with Stellar-Core. Noticed this was removed a couple years ago ref #759.


MonsieurNicolas:
Thank you!

PR #1571, opened 2018-02-28
Update asio library to use std::atomic_thread_fence, when available

evanlimanto:
This issue is the same as https://github.com/otland/forgottenserver/issues/1972.

To fix this, I used changes from https://github.com/chriskohlhoff/asio/commit/ffc9dd443fc2ca1aece0bbda14f3c521ceeb7924


MonsieurNicolas:
I don't think we should cherry pick commits from asio: I'd rather stick to releases. Especially in our case we're using a very old version of asio (that have other problems) that needs to be upgraded anyways. There is https://github.com/stellar/stellar-core/pull/1451 (not completed, maybe you can help?) that attempts to move to a more sustainable way of consuming asio releases.

evanlimanto:
got it, thanks. i'll close this PR.

PR #1582, opened 2018-03-05
Update load generation logic

marta-lokhova:
-- Description
	* Refactor LoadGenerator to achieve the following:
		- Create accounts in a separate step
			- Be able to batch up to 100 ops/tx
		- Create single payments in a separate step
		- Deterministically generate destination account based on the source account and its sequence number 
		- Mark load generation complete after changes are propagated to the DB
		- Update tests to use the new load generator

-- Usage 
	* A user would hit `generateload` endpoint with the following params:
		- mode [create|pay] - determines accounts creation or payments
		- nAccounts - number of accounts to create or use for payments
		- nTxs - num payments to perform
		- batchsize - when creating accounts, batch multiple ops in a transaction
		- txrate - tx/s injestion rate, (num or "auto")
	* Note that accounts creations and payments are separate calls to the endpoint


src/test/TestAccount.cpp
```diff
@@ -24,9 +24,13 @@ TestAccount::loadSequenceNumber() const
 void
 TestAccount::updateSequenceNumber()
```
marta-lokhova: Note that current version of updateSequenceNumber on master breaks the code (because of usage of REQUIRE), so this change is copied from bump sequence PR.

src/simulation/LoadGenerator.cpp
```diff
@@ -40,156 +40,78 @@ namespace stellar
...
+    : mMinBalance(0), mLastSecond(0), mApp(app)
 {
-    // Root account gets enough XLM to create 10 million (10^7) accounts, which
-    // thereby uses up 7 + 3 + 7 = 17 decimal digits. Luckily we have 2^63 =
-    // 9.2*10^18, so there's room even in 62bits to do this.
-    auto root = make_shared<AccountInfo>(0, txtest::getRoot(networkID),
-                                         10000000ULL * LOADGEN_ACCOUNT_BALANCE,
-                                         0, 0, *this);
-    mAccounts.push_back(root);
+    getRootAccount();
```
vogel: Is that necessary here?

src/simulation/LoadGenerator.cpp
```diff
@@ -40,156 +40,78 @@ namespace stellar
...
 
 LoadGenerator::~LoadGenerator()
 {
     clear();
 }
 
-std::string
-LoadGenerator::pickRandomAsset()
+bool
+LoadGenerator::getRootAccount()
```
vogel: I think that name and return value of this method are misleading.

This method does not return root account, and its return value is never used.
I propose to either:
* rename it to void LoadGenerator::createRootAccount
* make it return `res` or `nullptr`, remove `mRoot` field and pass returned value to `creationTransaction` as a parameter

src/simulation/LoadGenerator.cpp
```diff
@@ -221,591 +143,459 @@ maybeAdjustRate(double target, double actual, uint32_t& rate, bool increaseOk)
...
 {
-    soci::transaction sqltx(app.getDatabase().getSession());
-    app.getDatabase().setCurrentTransactionReadOnly();
+    soci::transaction sqltx(mApp.getDatabase().getSession());
+    mApp.getDatabase().setCurrentTransactionReadOnly();
+    getRootAccount();
 
-    updateMinBalance(app);
+    // Finish if no more txs need to be created.
+    if ((isCreate && nAccounts == 0) || (!isCreate && nTxs == 0))
```
vogel: In what case is that true?

src/simulation/LoadGenerator.cpp
```diff
@@ -221,591 +143,459 @@ maybeAdjustRate(double target, double actual, uint32_t& rate, bool increaseOk)
...
-        CLOG(INFO, "LoadGen") << "Load generation complete.";
-        app.getMetrics().NewMeter({"loadgen", "run", "complete"}, "run").Mark();
-        clear();
-    }
-    else
+    uint32_t ledgerNum = mApp.getLedgerManager().getLedgerNum();
+    uint64_t numToProcess = 1;
+    uint64_t totalAccounts = mAccounts.size();
+
+    for (uint32_t i = 0; i < txPerStep; ++i)
```
vogel: It looks for me as if there are 2 different methods hiding inside this loop, one for isCreate == true and one for == false. Or maybe it is even small interface with 2 implementation. Following all those if (isCreate) branches one after another is quite hard.

src/simulation/LoadGenerator.cpp
vogel: Never mind, found that case:)

src/simulation/LoadGenerator.cpp
```diff
@@ -221,591 +143,459 @@ maybeAdjustRate(double target, double actual, uint32_t& rate, bool increaseOk)
...
-        }
+    uint64_t now =
+        static_cast<uint64_t>(VirtualClock::to_time_t(mApp.getClock().now()));
+    bool secondBoundary = now != mLastSecond;
 
-        // Emit a log message once per second.
-        if (secondBoundary)
-        {
-            using namespace std::chrono;
+    if (autoRate && secondBoundary)
```
vogel: I think it deserves separate method.

src/simulation/LoadGenerator.cpp
```diff
@@ -221,591 +143,459 @@ maybeAdjustRate(double target, double actual, uint32_t& rate, bool increaseOk)
...
         }
     }
 }
 
-LoadGenerator::AccountInfoPtr
-LoadGenerator::pickRandomPath(LoadGenerator::AccountInfoPtr from,
-                              uint32_t ledgerNum,
-                              std::vector<LoadGenerator::AccountInfoPtr>& path)
+bool
+LoadGenerator::checkAccountSynced(std::vector<TestAccountPtr>& result,
```
vogel: I think we could just make std::vector<TestAccountPtr> return type and result result.
RVO will take care of copying and having that gives us enough information to get the `synced` value as well.

src/simulation/LoadGenerator.cpp
```diff
@@ -843,265 +619,68 @@ LoadGenerator::TxMetrics::report()
...
-            for (auto& tl : mTo->mTrustLines)
-            {
-                if (tl.mIssuer == mPath.back())
-                {
-                    tl.mBalance += mAmount;
-                }
-            }
-        }
-        else
+        static const char* TX_STATUS_STRING[Herder::TX_STATUS_COUNT] = {
```
vogel: it is duplicated in CommandHandler, I think we need to extract that to somewhere

src/main/CommandHandler.cpp
```diff
@@ -427,11 +441,19 @@ CommandHandler::generateLoad(std::string const& params, std::string& retStr)
...
-        retStr = fmt::format(
-            "Generating load: {:d} accounts, {:d} txs, {:d} tx/s = {:f} hours",
-            nAccounts, nTxs, txRate, hours);
+        uint32_t numItems = isCreate ? nAccounts : nTxs;
+        std::string itemType = isCreate ? "accounts" : "txs";
+        double hours = (numItems / txRate) / 3600.0;
+
+        if (batchSize > 100)
+        {
+            batchSize = 100;
```
vogel: what about displaying message that such values of batchSize are illegal?

src/main/CommandHandler.cpp
```diff
@@ -407,12 +407,13 @@ CommandHandler::generateLoad(std::string const& params, std::string& retStr)
...
         uint32_t batchSize = 100; // Only for account creations
         bool autoRate = false;
+        std::string mode = "create";
 
         std::map<std::string, std::string> map;
         http::server::server::parseParams(params, map);
 
         bool isCreate;
-        std::string mode = parseNumParam<std::string>(map, "mode");
+        maybeParseNumParam<std::string>(map, "mode", mode);
```
vogel: looks like name of this function is misleading - mode is not Num

src/simulation/LoadGenerator.cpp
marta-lokhova: It ensures mRoot is set, in case an outside method calls `clear` before `generateLoad`.

marta-lokhova:
Thanks for reviewing! Will update the code based on the suggestions.

vogel:
Looks good to me!
Could you rebase it on current master and squash some commits?

MonsieurNicolas:
Marking this PR as dependent on #1533 (for the changes in test accounts)

src/main/ApplicationImpl.h
```diff
@@ -86,8 +86,9 @@ class ApplicationImpl : public Application
 
     virtual bool manualClose() override;
 
-    virtual void generateLoad(uint32_t nAccounts, uint32_t nTxs,
-                              uint32_t txRate, bool autoRate) override;
+    virtual void generateLoad(bool IsCreate, uint32_t nAccounts, uint32_t nTxs,
```
vogel: typo - function arguments should start with lower case

MonsieurNicolas:
r+ 662d653a936b52dbe237e67ae3e04f86d1466a6d

marta-lokhova:
@MonsieurNicolas looks like a clang-format issue. Fixed now.

MonsieurNicolas:
r+ 31a038da8ad6b90b2f196d16ae089bf03a504f9d

MonsieurNicolas:
@latobarita: retry

PR #1585, opened 2018-03-05
Always connect to preferred peers

### LINKED ISSUE #1584: preferred peers may end up being dropped
Current code gives some priority to preferred peers but there are a few situations where they are ignored:
* node receiving a lot of inbound connections
* node dropping preferred peers (timeout) may not reconnect

both of these are caused by the `TARGET_PEER_CONNECTIONS` setting that causes a node to stop attempting to connect to more peers if that number is reached.

MonsieurNicolas:
resolves #1584 

This PR ensures that a node always attempts to connect to preferred peers, both at startup and later when preferred peers are temporary disconnected.


src/overlay/PeerRecord.cpp
```diff
@@ -358,11 +362,10 @@ PeerRecord::backOff(VirtualClock& clock)
 std::chrono::seconds
 PeerRecord::computeBackoff(VirtualClock& clock)
 {
-    uint32 backoffCount = std::min<uint32>(MAX_BACKOFF_EXPONENT, mNumFailures);
+    int32 backoffCount = std::min<int32>(MAX_BACKOFF_EXPONENT, mNumFailures);
 
     auto nsecs = std::chrono::seconds(
-        std::rand() %
```
vogel: for paranoid extra safety: backoffCount could be negative here (db error?), then value after % will be 0 and this could crash, maybe we need a check here?

src/overlay/PeerRecord.cpp
```diff
@@ -24,8 +24,12 @@ using namespace std;
 using namespace soci;
 
 PeerRecord::PeerRecord(string const& ip, unsigned short port,
-                       VirtualClock::time_point nextAttempt, uint32_t fails)
-    : mIP(ip), mPort(port), mNextAttempt(nextAttempt), mNumFailures(fails)
+                       VirtualClock::time_point nextAttempt, int fails)
+    : mIP(ip)
+    , mPort(port)
+    , mIsPreferred(false)
```
vogel: that wont compile until next commit

src/overlay/PeerRecord.cpp
```diff
@@ -348,7 +353,7 @@ PeerRecord::storePeerRecord(Database& db)
         auto tm = VirtualClock::pointToTm(mNextAttempt);
         auto prep = db.getPreparedStatement("UPDATE peers SET "
                                             "nextattempt = :v1, "
-                                            "numfailures = :v2 "
+                                            "numfailures = :v2, "
```
vogel: this should go with previous commit

src/overlay/PeerRecord.cpp
vogel: never mind, I see it fixed in 8cbe34399973cf89e81bff5dc7e0d3158942b55f

src/overlay/OverlayManagerImpl.cpp
```diff
@@ -192,29 +192,51 @@ OverlayManagerImpl::storeConfigPeers()
 }
 
 void
-OverlayManagerImpl::connectToMorePeers(int max)
+OverlayManagerImpl::connectToMorePeers(int max, bool preferred)
```
vogel: I would prefer this to look more like:

```
connectToPeers(vector<peer>);
vector<peer> getPrefferedPeersToConnect(int max)
vector<peer> getPeersToConnect(int max);
```

With calls like:

```
connectToPeers(getPrefferedPeersToConnect(max));
connectToPeers(getPeersToConnect(max));
```

Instead of this bool flag

vogel:
Test case "broadcast() broadcasts" fails with following message:

```pm.mAuthenticatedPeers.size() == 5
8 == 5
```

MonsieurNicolas:
updated

src/overlay/OverlayManagerImpl.cpp
```diff
@@ -224,8 +243,9 @@ OverlayManagerImpl::connectToMorePeers(int max)
         {
             continue;
         }
-        if (getAuthenticatedPeersCount() >=
-            mApp.getConfig().TARGET_PEER_CONNECTIONS)
+        // we always try to connect to preferred peers
```
vogel: what will happen here if we have all available connections occupied with preferred peers and pr is a non-connected preferred?

vogel:
Looks good now. Please answer my question and I'll r+ this.

src/overlay/OverlayManagerImpl.cpp
MonsieurNicolas: in this case it will connect anyways and will take the number of connections above TARGET at this stage (as a pending connection).
The logic in `OverlayManagerImpl::acceptAuthenticatedPeer` will deal with the connection properly later (kicking out other nodes if it can - enforcing `MAX_PEER_CONNECTIONS`).
If `MAX_PEER_CONNECTIONS` is mis-configured (like cannot accommodate for all preferred peers, including by host and by key), it will try every few seconds to reconnect, but that seems fine.

vogel:
r+ 402e6e4e75fa13bae866901e402de3d9a4a0f342

vogel:
clang-format required

MonsieurNicolas:
I ran clang-format on all commits

MonsieurNicolas:
r+ 89b10a1ec23c2f0efae42fb26e90d10d6453c38d

PR #1586, opened 2018-03-06
SCP: properly trigger the startedBallotProtocol callback

MonsieurNicolas:
Old code would only trigger that callback in some but not all cases (in particular when the node is "forced" to prepare by a quorum).

This makes the metric `scp.ballot.started` updated properly.

### DIFF
--- a/src/scp/BallotProtocol.cpp
+++ b/src/scp/BallotProtocol.cpp
@@ -390,7 +390,6 @@ BallotProtocol::bumpState(Value const& value, uint32 n)
 
     if (updated)
     {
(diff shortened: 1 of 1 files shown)


vogel:
r+ 823e8ac035e843494b92d9343758d1f94f64c179

vogel:
make format required

MonsieurNicolas:
forgot to run clang-format

MonsieurNicolas:
r+ 994460ffa0d8d671fab5c5279744346a7b5629a8

PR #1587, opened 2018-03-06
Added clearmetrics endpoint

marta-lokhova:
* What
Endpoint to clear metrics; if `domain` param passed in, clears metrics for only that domain
* Why 
This is needed for load test. Since we don't care about account creation metrics, we'd need to refresh the metrics to record only payments data
* How 
MetricResetter class is a child class of MetricProcessor. It implements Process method for each type of metric

### DIFF
--- a/src/main/ApplicationImpl.cpp
+++ b/src/main/ApplicationImpl.cpp
@@ -612,6 +612,20 @@ ApplicationImpl::syncAllMetrics()
     syncOwnMetrics();
 }
 
+void
+ApplicationImpl::clearMetrics(std::string const& domain)
+{
+    MetricResetter resetter;
+    auto const& metrics = mMetrics->GetAllMetrics();
+    for (auto const& kv : metrics)
+    {
+        if (domain.empty() || kv.first.domain() == domain)
+        {
+            kv.second->Process(resetter);
+        }
+    }
+}
+
 TmpDirManager&
 ApplicationImpl::getTmpDirManager()
 {
--- a/src/main/CommandHandler.cpp
+++ b/src/main/CommandHandler.cpp
@@ -83,6 +83,7 @@ CommandHandler::CommandHandler(Application& app) : mApp(app)
     addRoute("maintenance", &CommandHandler::maintenance);
     addRoute("manualclose", &CommandHandler::manualClose);
     addRoute("metrics", &CommandHandler::metrics);
+    addRoute("clearmetrics", &CommandHandler::clearMetrics);
     addRoute("peers", &CommandHandler::peers);
     addRoute("quorum", &CommandHandler::quorum);
     addRoute("setcursor", &CommandHandler::setcursor);
@@ -278,6 +279,9 @@ CommandHandler::fileNotFound(std::string const& params, std::string& retStr)
         "</p><p><h1> /metrics</h1>"
         "returns a snapshot of the metrics registry (for monitoring and "
         "debugging purpose)"
+        "</p><p><h1> /clearmetrics?[domain=DOMAIN]</h1>"
+        "clear metrics for a specified domain. If no domain specified, "
+        "clear all metrics (for testing purposes)"
         "</p><p><h1> /peers</h1>"
         "returns the list of known peers in JSON format"
         "</p><p><h1> /quorum?[node=NODE_ID][&compact=true]</h1>"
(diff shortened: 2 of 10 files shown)


src/main/ApplicationImpl.cpp
```diff
@@ -609,6 +609,39 @@ ApplicationImpl::syncAllMetrics()
...
 
+template <typename T>
+std::shared_ptr<T>
+castMetric(std::shared_ptr<medida::MetricInterface> metric)
+{
+    return std::dynamic_pointer_cast<T>(metric);
+}
+
+void
+ApplicationImpl::clearMetrics(std::string domain)
```
vogel: domain can be const-ref here

src/main/ApplicationImpl.cpp
```diff
@@ -609,6 +609,39 @@ ApplicationImpl::syncAllMetrics()
...
+    auto const& metrics = mMetrics->GetAllMetrics();
+    for (auto const& kv : metrics)
+    {
+        auto metricName = kv.first;
+        if (domain.empty() || metricName.domain() == domain)
+        {
+            auto metric = kv.second;
+            if (auto counter = castMetric<medida::Counter>(metric))
+            {
+                counter->clear();
```
MonsieurNicolas: are the clear methods thread safe?

src/main/CommandHandler.cpp
```diff
@@ -277,6 +278,8 @@ CommandHandler::fileNotFound(std::string const& params, std::string& retStr)
         "</p><p><h1> /metrics</h1>"
         "returns a snapshot of the metrics registry (for monitoring and "
         "debugging purpose)"
+        "</p><p><h1> /clearmetrics?[domain=DOMAIN]</h1>"
+        "clear metrics for all/specified domains (for testing purposes)"
```
MonsieurNicolas: the same string is also in `docs/software/commands.md`

MonsieurNicolas: I opened https://github.com/stellar/stellar-core/issues/1589 to avoid having to deal with this problem in the future

MonsieurNicolas: you also need to document the behavior if domain is not specified

src/main/CommandHandler.cpp
```diff
@@ -965,4 +968,18 @@ CommandHandler::maintenance(std::string const& params, std::string& retStr)
...
 }
+
+void
+CommandHandler::clearMetrics(std::string const& params, std::string& retStr)
+{
+    std::map<std::string, std::string> map;
+    http::server::server::parseParams(params, map);
+
+    std::string domain;
+    maybeParseNumParam(map, "domain", domain);
```
MonsieurNicolas: same comment than in a different PR, `maybeParseNumParam` needs to be renamed if we're using it to parse types other than numbers.

src/main/ApplicationImpl.cpp
marta-lokhova: Looks like it. Timer and Histogram are using mutex, while counter only clears one variable and uses `std::atomic`.

src/main/CommandHandler.cpp
marta-lokhova: The other PR was updated with the name change. for this PR I was going off of master. Depending on what gets merged first, I'll be able to update it during rebase.

src/main/ApplicationImpl.cpp
```diff
@@ -609,6 +609,39 @@ ApplicationImpl::syncAllMetrics()
...
+ApplicationImpl::clearMetrics(const std::string& domain)
+{
+    auto const& metrics = mMetrics->GetAllMetrics();
+    for (auto const& kv : metrics)
+    {
+        auto metricName = kv.first;
+        if (domain.empty() || metricName.domain() == domain)
+        {
+            auto metric = kv.second;
+            if (auto counter = castMetric<medida::Counter>(metric))
```
jonjove: This should be replaced with `medida::MetricProcessor` and `medida::MetricInterface::Process` in order to avoid the use of dynamic cast.

jonjove:
r+ 05eb4db

fuzzyami:
Hey guys. I've got a question related to this endpoint:
is there a reason NOT to use this endpoint in production setups? I'd like to collect 5 minutes worth of metrics (ie call this endpoint every 5 minutes), would this hinder the normal operation of my cores? if so, perhaps it should be documented. Perhaps there's another way to window the amount of metrics that I'm not familiar with?
thx!

MonsieurNicolas:
yeah you should not do this, the metrics library that we use will actually not produce the right data and also doesn't reset all counters.

You're much better off using something like https://github.com/stellar/packages/blob/master/stellar-core-prometheus-exporter/stellar-core-prometheus-exporter.py

PR #1588, opened 2018-03-07
Build for mac os X seirra

djinn:
Fix atomic fencing


MonsieurNicolas:
Thanks for submitting this change. I don't think it makes sense to take this in, it was actually already submitted earlier https://github.com/stellar/stellar-core/pull/1571 . reason is that we only want to track asio releases instead of modified ones and also, *I think* (I do not have a Mac), this is actually not a problem with High Sierra (10.13), only Sierra (10.12).

MonsieurNicolas:
we're now using asio directly as a submodule

PR #1590, opened 2018-03-08
Adjustment to automatic maintenance

MonsieurNicolas:
This PR adjusts the following in automatic maintenance:
* change the default to run only once every 4 hours. This seems to be a good number (leaving room for buffer), as a default as it seems to perform OK on low end systems (with low bandwidth to the postgres database)
* clears prepared transactions when performing the actual clean up (there is a potential race otherwise that may cause delays)
* fix a potential integer overflow when invoking cleanup with large numbers


jonjove:
r+ deb728a

PR #1593, opened 2018-03-09
Scp quorum seen fix

MonsieurNicolas:
we were not resetting timers properly when receiving messages not causing a state transition (typically when the local node needs to be included to form a quorum)


jedmccaleb:
r+ f34020c

PR #1596, opened 2018-03-12
Performance fix for HerderSCPDriver::nominate method

EugeneChung:
- Calling sha256() is CPU-consuming task but the result of it is only for debug logging. Writing strings to debug logger stream is also useless. So they must be done when Logging::logDebug("Herder") is true.
- The valueHash for debug log must be sha256(mCurrentValue). Calling xdr::xdr_to_opaque(mCurrentValue) is wrong.

### DIFF
--- a/src/herder/HerderSCPDriver.cpp
+++ b/src/herder/HerderSCPDriver.cpp
@@ -653,14 +653,16 @@ HerderSCPDriver::nominate(uint64_t slotIndex, StellarValue const& value,
     mCurrentValue = xdr::xdr_to_opaque(value);
     mLedgerSeqNominating = static_cast<uint32_t>(slotIndex);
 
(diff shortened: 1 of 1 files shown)


MonsieurNicolas:
Hi
thanks for the patch. This method is only called once per consensus round so code simplicity is better (this particular change ends up having no real performance impact).

This made me realize that we do not have a good backlog for performance related improvements as we've been doing higher level stress tests so far (that are much harder to turn into something actionable for people new to the system)... I just opened https://github.com/stellar/stellar-core/issues/1597 to start driving this

EugeneChung:
It’s a common mistake of developers when adding debugging-purpose codes. And it’s also common approach to wrap those codes with log-level checking conditional statement. I know it’s minor gain for CPU but it should be done for back-end servers even if it’s just minor.

My patch is simple. Just adding if. I don’t why you address simplicity.

PR #1600, opened 2018-03-14
Data layer

jonjove:
This pull request refactors the data layer of stellar-core. The pull request conceptually consists of several distinct parts:
- The first part of the pull request introduces `LedgerState` and several associated classes, which constitute the center of the new data layer. These classes make a variety of guarantees which were not possible with the old data layer, notably that it is no longer possible to have the same data loaded in multiple places.
- The second part of the pull request transitions all of the operations and some supporting classes to the new data layer.
- The third part of the pull request removes the notion of current ledger from the `LedgerManager`.
- The fourth part of the pull request removes the old data layer. This includes some updates to invariants and bucket application.


src/transactions/ManageOfferOpFrame.cpp
```diff
@@ -164,6 +167,8 @@ ManageOfferOpFrame::doApply(Application& app, LedgerDelta& delta,
             return false;
         }
 
+        mSellSheepOffer->storeDelete(tempDelta, db);
```
MonsieurNicolas: I think you should also make the call to `addNumEntries` here to keep the account in a consistent state (and behavior is basically the same between modify and create except for result).

I think it just means that the condition later needs to be updated to
```
if (!creatingNewOffer ||
                 app.getLedgerManager().getCurrentLedgerVersion() > 8)
```

src/transactions/OfferExchange.cpp
```diff
@@ -110,6 +110,40 @@ exchangeV3(int64_t wheatReceived, Price price, int64_t maxWheatReceive,
     return result;
 }
 
+LoadBestOfferContext::LoadBestOfferContext(Database& db, Asset const& selling,
```
MonsieurNicolas: nice change!

MonsieurNicolas:
You should also cleanup warnings, I built this on Windows and got those:
```
1>c:\src\stellar-core\src\bucket\buckettests.cpp(1061): warning C4189: 'sess': local variable is initialized but not referenced
1>c:\src\stellar-core\src\history\historytestsutils.cpp(327): warning C4189: 'lm': local variable is initialized but not referenced
1>c:\src\stellar-core\src\ledger\ledgermanagerimpl.cpp(293): warning C4456: declaration of 'ls' hides previous local declaration
1>c:\src\stellar-core\src\ledger\ledgermanagerimpl.cpp(240): note: see declaration of 'ls'
1>LedgerStateTrustLineSQL.cpp
1>c:\src\stellar-core\src\ledger\ledgerstatetests.cpp(395): warning C4189: 'nAccountsChanged': local variable is initialized but not referenced
1>c:\src\stellar-core\src\simulation\coretests.cpp(560): warning C4189: 'complete': local variable is initialized but not referenced
1>c:\src\stellar-core\src\test\txtests.cpp(455): warning C4456: declaration of 'ls' hides previous local declaration
1>c:\src\stellar-core\src\test\txtests.cpp(436): note: see declaration of 'ls'
1>c:\src\stellar-core\src\test\txtests.cpp(474): warning C4456: declaration of 'ls' hides previous local declaration
1>c:\src\stellar-core\src\test\txtests.cpp(436): note: see declaration of 'ls'
1>c:\src\stellar-core\src\test\txtests.cpp(486): warning C4456: declaration of 'ls' hides previous local declaration
1>c:\src\stellar-core\src\test\txtests.cpp(436): note: see declaration of 'ls'
1>c:\src\stellar-core\src\test\txtests.cpp(533): warning C4456: declaration of 'ls' hides previous local declaration
1>c:\src\stellar-core\src\test\txtests.cpp(518): note: see declaration of 'ls'
1>c:\src\stellar-core\src\test\txtests.cpp(554): warning C4456: declaration of 'ls' hides previous local declaration
1>c:\src\stellar-core\src\test\txtests.cpp(518): note: see declaration of 'ls'
1>c:\src\stellar-core\src\test\txtests.cpp(567): warning C4456: declaration of 'ls' hides previous local declaration
1>c:\src\stellar-core\src\test\txtests.cpp(518): note: see declaration of 'ls'
1>c:\src\stellar-core\src\transactions\changetrusttests.cpp(28): warning C4189: 'db': local variable is initialized but not referenced
1>c:\src\stellar-core\src\transactions\managedataopframe.cpp(50): warning C4456: declaration of 'header' hides previous local declaration
1>c:\src\stellar-core\src\transactions\managedataopframe.cpp(36): note: see declaration of 'header'
1>c:\src\stellar-core\src\transactions\managedataopframe.cpp(87): warning C4456: declaration of 'header' hides previous local declaration
1>c:\src\stellar-core\src\transactions\managedataopframe.cpp(36): note: see declaration of 'header'
1>c:\src\stellar-core\src\transactions\offerexchange.cpp(216): warning C4456: declaration of 'header' hides previous local declaration
1>c:\src\stellar-core\src\transactions\offerexchange.cpp(178): note: see declaration of 'header'
1>c:\src\stellar-core\src\transactions\paymenttests.cpp(147): warning C4456: declaration of 'a1Balance' hides previous local declaration
1>c:\src\stellar-core\src\transactions\paymenttests.cpp(81): note: see declaration of 'a1Balance'
1>c:\src\stellar-core\src\transactions\paymenttests.cpp(173): warning C4456: declaration of 'a1Balance' hides previous local declaration
1>c:\src\stellar-core\src\transactions\paymenttests.cpp(81): note: see declaration of 'a1Balance'
1>c:\src\stellar-core\src\transactions\paymenttests.cpp(198): warning C4456: declaration of 'a1Balance' hides previous local declaration
1>c:\src\stellar-core\src\transactions\paymenttests.cpp(81): note: see declaration of 'a1Balance'
1>c:\src\stellar-core\src\transactions\paymenttests.cpp(298): warning C4456: declaration of 'rootBalance' hides previous local declaration
1>c:\src\stellar-core\src\transactions\paymenttests.cpp(80): note: see declaration of 'rootBalance'
1>c:\src\stellar-core\src\transactions\changetrustopframe.cpp(70): warning C4456: declaration of 'header' hides previous local declaration
1>c:\src\stellar-core\src\transactions\changetrustopframe.cpp(32): note: see declaration of 'header'
1>c:\src\stellar-core\src\transactions\changetrustopframe.cpp(121): warning C4456: declaration of 'header' hides previous local declaration
1>c:\src\stellar-core\src\transactions\changetrustopframe.cpp(32): note: see declaration of 'header'
```

src/ledger/LedgerState.cpp
```diff
@@ -0,0 +1,1252 @@
...
+        auto entry = state.second->ignoreInvalid().entry();
+        if (entry)
+        {
+            entry->lastModifiedLedgerSeq = ledgerSeq;
+        }
+    }
+}
+
+/*
+template <typename ValueType>
```
MonsieurNicolas: remove commented out code

src/ledger/LedgerStateOfferSQL.cpp
```diff
@@ -0,0 +1,344 @@
...
+
+LedgerState::StateEntry
+LedgerState::loadOfferFromDatabase(LedgerKey const& key)
+{
+    auto& db = mRoot->getDatabase();
+
+    uint64_t offerID = key.offer().offerID;
+    std::string actIDStrKey = KeyUtils::toStrKey(key.offer().sellerID);
+
+    std::string sql = "SELECT sellerid, offerid, "
```
MonsieurNicolas: undo the duplication of this sql string like it was before as all calls to `loadOffersFromDatabase` must exactly match when it comes to columns being returned.

src/main/ApplicationImpl.cpp
```diff
@@ -728,6 +731,12 @@ ApplicationImpl::getStatusManager()
     return *mStatusManager;
 }
 
+LedgerStateRoot&
+ApplicationImpl::getLedgerStateRoot()
+{
+    return *mLedgerStateRoot;
```
MonsieurNicolas: I would suggest asserting that this is only used in the main thread otherwise it will potentially corrupt things (code is not thread safe)

src/ledger/LedgerStateTests.cpp
```diff
@@ -0,0 +1,655 @@
...
+            lhr->header() = generateLedgerHeader();
+            lh2 = lhr->header();
+            lsChild.commit();
+        }
+        REQUIRE(ls.loadHeader()->header() == lh2);
+    }
+}
+
+// TODO(jonjove): Might be a few other cases worth testing
+TEST_CASE("LedgerState::forget and LedgerEntryReference::forgetFromLedgerState",
```
MonsieurNicolas: I think that the "forget" cases should be incorporated into the "commit" and "rollback" tests as it should be covering the same matrix

MonsieurNicolas:
I did a review of the ledger state related changes before any transitions ie I didn't review "Transition OfferExchange to LedgerState".
Overall looks pretty much in line with what we discussed before.

src/transactions/ManageDataOpFrame.cpp
```diff
@@ -55,26 +57,23 @@ ManageDataOpFrame::doApply(Application& app, LedgerDelta& delta,
...
-            dataFrame->storeAdd(delta, db);
-            mSourceAccount->storeChange(delta, db);
+            header->invalidate();
+
+            LedgerEntry newData;
+            newData.data.type(DATA);
+            newData.data.data().accountID = getSourceID();
+            newData.data.data().dataName = mManageData.dataName;
+            newData.data.data().dataValue = *mManageData.dataValue;
+            dataRef = ls.create(newData);
```
MonsieurNicolas: why the assignment to `dataRef`?

src/transactions/ManageOfferOpFrame.cpp
```diff
@@ -384,31 +388,35 @@ ManageOfferOpFrame::doApply(Application& app, LedgerDelta& delta,
                 innerResult().code(MANAGE_OFFER_LOW_RESERVE);
                 return false;
             }
-            mSellSheepOffer->mEntry.data.offer().offerID =
-                tempDelta.getHeaderFrame().generateID();
+
+            newOffer.data.offer().offerID = ++header->header().idPool;
```
MonsieurNicolas: why not keep this as a method `GenerateOfferID()` in `LedgerHeaderReference`?

src/transactions/PathPaymentOpFrame.cpp
```diff
@@ -282,24 +280,31 @@ PathPaymentOpFrame::doApply(Application& app, LedgerDelta& delta,
             return false;
         }
 
-        auto ok = sourceAccount->addBalance(-curBSent);
+        if (getCurrentLedgerVersion(header) < 8)
+        {
+            sourceAccount.forget(ls);
```
MonsieurNicolas: same than earlier: comment to explain a little bit what is going on here

src/transactions/PathPaymentOpFrame.cpp
```diff
@@ -282,24 +280,31 @@ PathPaymentOpFrame::doApply(Application& app, LedgerDelta& delta,
...
-                TrustFrame::loadTrustLine(getSourceID(), curB, db, &delta);
-        }
-        else
-        {
-            auto tlI =
-                TrustFrame::loadTrustLineIssuer(getSourceID(), curB, db, delta);
-
-            if (!tlI.second)
+            // We need a LedgerState here since it is possible that changes to
+            // issuer were already stored.
```
MonsieurNicolas: comment is misleading, I think what you mean is that issuer may be one of the other accounts used in this function (such as `sourceAccount`), in which case, isn't it easier to just invalidate after changing the balance of the `sourceAccount`? (RAII to the rescue)

src/transactions/PathPaymentOpFrame.cpp
```diff
@@ -309,10 +314,11 @@ PathPaymentOpFrame::doApply(Application& app, LedgerDelta& delta,
                 innerResult().noIssuer() = curB;
                 return false;
             }
-            sourceLineFrame = tlI.first;
+            lsIssuer.rollback();
```
MonsieurNicolas: I thought default was rollback? or you just want to be explicit?

src/ledger/LedgerManagerImpl.cpp
```diff
@@ -544,11 +562,17 @@ LedgerManagerImpl::historyCaughtup(asio::error_code const& ec,
         {
         case CatchupWork::ProgressState::APPLIED_BUCKETS:
         {
-            mCurrentLedger = make_shared<LedgerHeaderFrame>(lastClosed.header);
-            storeCurrentLedger();
+            LedgerState ls(mApp.getLedgerStateRoot());
```
MonsieurNicolas: Isn't this basically what `advanceLedgerPointers` does + store ?

src/ledger/LedgerManagerImpl.cpp
```diff
@@ -881,37 +909,43 @@ LedgerManagerImpl::checkDbState()
...
-            tx->storeTransactionFee(*this, thisTxDelta.getChanges(), ++index);
-            thisTxDelta.commit();
+            LedgerState lsInner(ls);
+            auto header = lsInner.loadHeader();
+            uint32_t ledgerSeq = header->header().ledgerSeq;
+            header->invalidate();
+
+            tx->processFeeSeqNum(lsInner);
+            tx->storeTransactionFee(mApp.getDatabase(), ledgerSeq, lsInner.getChanges(), ++index);
+            lsInner.updateLastModified();
```
MonsieurNicolas: why is this `updateLastModified` call here?
Did you add it because of the one missing in `processFeeSeqNum`?

I would expect it to be done *before* calling `lsInner.getChanges()` so that the meta data in the database contains this information?

MonsieurNicolas: this actually makes me think that maybe we should remove the explicit calls to `updateLastModified` from the codebase and instead have ledger state do it:
* every time "commit" is called
* every time one of the getChanges is called
* we should crash if `updateLastModified` would modify entries when calling getLive/dead entries (something horrible happened)

it's fairly cheap to do and that way we don't introduce yet another way for people to shoot themselves in the foot.

Also, we should guard against calling `updateLastModified` after commit - as we're modifying ledger entries

src/ledger/LedgerEntryTests.cpp
```diff
@@ -1,261 +0,0 @@
...
-TEST_CASE("Ledger Entry tests", "[ledgerentry]")
-{
-    Config cfg(getTestConfig(0));
-
-    VirtualClock clock;
-    Application::pointer app = createTestApplication(clock, cfg);
-    app->start();
-    Database& db = app->getDatabase();
-
-    SECTION("round trip with database")
```
MonsieurNicolas: I could not find equivalent tests (that would now be LedgerStateRoot tests that commit to the database)

src/transactions/PaymentOpFrame.cpp
```diff
@@ -68,8 +68,7 @@ PaymentOpFrame::doApply(Application& app, LedgerState& ls)
     opRes.code(opINNER);
     opRes.tr().type(PATH_PAYMENT);
     PathPaymentOpFrame ppayment(op, opRes, mParentTx);
-    // TODO(jonjove): This needs to be changed
-    ppayment.setSourceAccountPtr(mSourceAccount);
+    // TODO(jonjove): Used to set source account pointer here, why?
```
MonsieurNicolas: that was because it's a different object from what I remember - not relevant anymore :)

src/ledger/LedgerStateAccountSQL.cpp
```diff
@@ -229,8 +229,11 @@ LedgerState::storeSignersInDatabase(StateEntry const& state)
     std::vector<Signer> signers;
     if (previousEntry)
     {
-        auto key = LedgerEntryKey(*previousEntry);
-        signers = loadSignersFromDatabase(key);
+        signers = previousEntry->data.account().signers;
+        assert(std::adjacent_find(signers.begin(), signers.end(),
```
MonsieurNicolas: how do we know that `previousEntry` not set implies that we can skip loading from the database? (probably a question pertinent to all entries, not just signers)

src/invariant/AccountSubEntriesCountIsValid.cpp
```diff
@@ -34,7 +34,7 @@ AccountSubEntriesCountIsValid::getName() const
 std::string
 AccountSubEntriesCountIsValid::checkOnOperationApply(
     Operation const& operation, OperationResult const& result,
-    LedgerState& ls)
+    LedgerState const& ls, std::shared_ptr<LedgerHeaderReference const> header)
```
MonsieurNicolas: nice!

src/transactions/TransactionFrame.cpp
```diff
@@ -558,8 +558,9 @@ TransactionFrame::applyOperations(SignatureChecker& signatureChecker,
             }
             if (!errorEncountered)
             {
+                auto header = ls.loadHeader();
```
MonsieurNicolas: we're probably missing some checks somewhere as this is wrong:
it should be `lsOperation.loadHeader()`

src/invariant/BucketListIsConsistentWithDatabase.cpp
```diff
@@ -90,6 +88,8 @@ BucketListIsConsistentWithDatabase::checkOnBucketApply(
     std::shared_ptr<Bucket const> bucket, uint32_t oldestLedger,
     uint32_t newestLedger)
 {
+    mApp.getLedgerStateRoot().flushCache();
```
MonsieurNicolas: why do we need the call to `flushCache()` here? This function only reads from the database... I would expect writers being the only ones that need to do this (under very specific conditions)

src/ledger/LedgerState.cpp
```diff
@@ -0,0 +1,1228 @@
...
+            else
+            {
+                // In this case, there are offers (with the correct assets) in
+                // memory. It is possible, then, that the best offer in the
+                // database is already loaded in memory. If that is the case, we
+                // must skip the result from the database. Then we repeat the
+                // entire process (not just this branch) to find the next data.
+                mTop = mFromDatabase.front();
+                mFromDatabase.erase(mFromDatabase.begin());
+                if (mLedgerState.isInMemory(LedgerEntryKey(*mTop->entry())))
```
MonsieurNicolas: I am not sure I understand what is going on here:
isn't `mInMemory` initialized with all ledger states (recursively)?
If so, why would we need to call `isInMemory` like this later on?

src/ledger/LedgerState.cpp
```diff
@@ -0,0 +1,1228 @@
...
+    if (iter != mLedgerState.mState.end())
+    {
+        auto const& previous = iter->second->ignoreInvalid().previousEntry();
+        mTop = mLedgerState.makeStateEntry(entry, previous);
+    }
+    else
+    {
+        mTop = mLedgerState.makeStateEntry(entry, entry);
+    }
+    mLedgerState.mState[key] = mTop;
```
MonsieurNicolas: I would prefer not duplicating the code that lives in LegderState proper:
in this case, I *think* that you got `previous` wrong for example... (should be doing what `LedgerState::loadHelper` does : use whatever value `parent` says)

Easiest might be to just push values from `loadBestOffersFromDatabase` into the cache, and just call `LedgerState::loadHelper` (that will hit the cache).

MonsieurNicolas: we really need more meta data tests...

src/ledger/LedgerState.cpp
```diff
@@ -0,0 +1,1252 @@
...
+LedgerStateRoot::hasChild()
+{
+    return mHasChild;
+}
+
+void
+LedgerStateRoot::hasChild(bool child)
+{
+    assert(!child || !mHasChild);
+    mHasChild = child;
```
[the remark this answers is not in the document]
vogel: It allows seting mHasChild to false, if child is false, assertion will be true. So maybe setHasChild and removeChild?

src/ledger/LedgerState.cpp
```diff
@@ -0,0 +1,1252 @@
...
+LedgerStateRoot::getDatabase()
+{
+    return mDb;
+}
+
+Database::EntryCache&
+LedgerStateRoot::getCache()
+{
+    // TODO(jonjove): Do we want to continue using the cache from Database?
+    // We could refactor so the cache is a member of LedgerStateRoot
```
vogel: I agree that we should do that.

src/ledger/LedgerState.cpp
```diff
@@ -0,0 +1,1228 @@
...
+    mParent->mChild = this;
+    for (auto const& state : mParent->mState)
+    {
+        state.second->invalidate();
+    }
+    if (mParent->mHeader)
+    {
+        mParent->mHeader->invalidate();
+    }
+    mParent->mLoadBestOfferContext.reset();
```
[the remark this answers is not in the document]
vogel: I think that move only moves content of mParent->mLoadBestOfferContext pointer, still keeping that pointer not-null, which is fixed by this reset.

src/ledger/LedgerState.cpp
```diff
@@ -0,0 +1,1252 @@
...
+        //    lsChild.load(key)->erase();
+        //    lsChild.commit();
+        //    lsParent.create(key);
+        if (iter->second->ignoreInvalid().entry())
+        {
+            throw std::runtime_error("Key already exists in memory");
+        }
+
+        auto newEntry = std::make_shared<LedgerEntry>(entry);
+        if (!mChild)
```
vogel: Please add comment to explain why this condition is needed and why when there is mChild, previous value is nullptr.

src/ledger/LedgerState.cpp
```diff
@@ -0,0 +1,1252 @@
...
+        if (mInMemory.empty())
+        {
+            if (mFromDatabase.empty())
+            {
+                return nullptr;
+            }
+            else
+            {
+                // In this case, there are offers (with the correct assets) in
+                // memory. It is possible, then, that the best offer in the
```
vogel: isn't memory empty here as checked on line 923?

src/ledger/LedgerState.cpp
```diff
@@ -0,0 +1,1252 @@
...
+            if (mLedgerState.isInMemory(LedgerEntryKey(*mTop->entry())))
+            {
+                mTop.reset();
+            }
+        }
+    }
+
+    auto const& entry = mTop->ignoreInvalid().entry();
+    auto key = LedgerEntryKey(*entry);
+    auto iter = mLedgerState.mState.find(key);
```
vogel: I think this is a repeating pattern that may need a method of its own

src/ledger/LedgerState.cpp
```diff
@@ -0,0 +1,1252 @@
...
+
+    if (!mHeader)
+    {
+        loadHeader()->invalidate();
+    }
+    if (mHeader->ignoreInvalid().header().ledgerVersion >= 8)
+    {
+        assert(se->entry());
+        assert(se->previousEntry());
+        assert(*se->entry() == *se->previousEntry());
```
vogel: why is that needed?

src/ledger/LedgerStateTests.cpp
```diff
@@ -0,0 +1,655 @@
+// Copyright 2017 Stellar Development Foundation and contributors. Licensed
+// under the Apache License, Version 2.0. See the COPYING file at the root
+// of this distribution or at http://www.apache.org/licenses/LICENSE-2.0
+
+#include "ledger/LedgerEntryReference.h"
+#include "ledger/LedgerHeaderReference.h"
+#include "ledger/LedgerState.h"
+#include "ledger/LedgerTestUtils.h"
+#include "lib/catch.hpp"
```
vogel: please use <lib/catch.hpp> for consistency

src/ledger/LedgerStateTests.cpp
```diff
@@ -0,0 +1,655 @@
...
+            {
+                VirtualClock clock;
+                Config cfg = getTestConfig();
+                cfg.INVARIANT_CHECKS = {};
+                Application::pointer app = createTestApplication(clock, cfg);
+
+                auto leParent1 = generateLedgerEntry(let);
+                auto leParent2 = generateLedgerEntryWithSameKey(leParent1);
+                auto parentKey = LedgerEntryKey(leParent1);
+                REQUIRE(parentKey == LedgerEntryKey(leParent2));
```
vogel: is this check required? if so, maybe place it in generateLedgerEntryWithSameKey ?

src/ledger/LedgerStateTests.cpp
```diff
@@ -0,0 +1,655 @@
...
+
+TEST_CASE("LedgerState::loadBestOffer in database but modified in memory",
+          "[ledger][ledgerstate][loadbestoffer]")
+{
+    VirtualClock clock;
+    Config cfg = getTestConfig();
+    cfg.INVARIANT_CHECKS = {};
+    Application::pointer app = createTestApplication(clock, cfg);
+
+    auto generateOffer = []() {
```
vogel: I think those two can be moved outside of this function to be reused

src/ledger/LedgerStateTests.cpp
```diff
@@ -0,0 +1,655 @@
...
+                LedgerState ls(root);
+                ls.create(le);
+                ls.commit();
+            }
+            {
+                LedgerState ls(root);
+                ls.load(key)->erase();
+                REQUIRE_THROWS_AS(ls.commit([]() {
+                    throw std::runtime_error("Abort database commit");
+                }),
```
vogel: did you forget clang-format or is it clang-format issue? This few lines of code looks strange

src/transactions/TransactionUtils.cpp
```diff
@@ -0,0 +1,199 @@
...
+
+void
+generateOfferID(std::shared_ptr<LedgerEntryReference> offer,
+                std::shared_ptr<LedgerHeaderReference> header)
+{
+    offer->entry()->data.offer().offerID = ++header->header().idPool;
+}
+
+uint64_t
+getStartingSequenceNumber(std::shared_ptr<LedgerHeaderReference> header)
```
vogel: I think we are overusing shared_ptr all over the code.
For example here a normal (or const) reference would do better:

LedgerHeaderReference const& header

It would not require doing copy of shared_ptr and doing atomic inc/dec of ref-counter. It also suggest that LedgerHeaderReference will not be changes inside this method (which is different in generateOfferID, where non-const referece to offer and header should be passed).

src/transactions/TransactionUtils.cpp
```diff
@@ -0,0 +1,199 @@
...
+    catch (std::runtime_error& e)
+    {
+        return nullptr;
+    }
+}
+
+std::shared_ptr<TrustLineReference>
+loadTrustLine(LedgerState& ls, AccountID const& accountID, Asset const& asset)
+{
+    assert(asset.type() == ASSET_TYPE_NATIVE || asset.type() == ASSET_TYPE_CREDIT_ALPHANUM4 || asset.type() == ASSET_TYPE_CREDIT_ALPHANUM12);
```
vogel: those two asserts are almost the same, which one is valid?

src/transactions/OfferExchange.cpp
```diff
@@ -4,10 +4,16 @@
...
 #include "database/Database.h"
-#include "ledger/LedgerDelta.h"
-#include "ledger/LedgerManager.h"
-#include "ledger/TrustFrame.h"
+#include "ledger/LedgerEntryReference.h"
+#include "ledger/LedgerHeaderReference.h"
+#include "ledger/LedgerState.h"
 #include "util/Logging.h"
+#include "transactions/TransactionUtils.h"
+#include <memory>
```
vogel: please keep local an library includes in two separate batches

src/transactions/OfferExchange.cpp
```diff
@@ -110,98 +117,72 @@ exchangeV3(int64_t wheatReceived, Price price, int64_t maxWheatReceive,
...
+    int64_t maxWheatReceived, int64_t& numWheatReceived,
+    int64_t maxSheepSend, int64_t& numSheepSend)
 {
     assert(maxWheatReceived > 0);
     assert(maxSheepSend > 0);
 
-    // we're about to make changes to the offer
-    mDelta.recordEntry(sellingWheatOffer);
+    // Note: These must be copies, not references.
+    Asset sheep = sellingWheatOffer->entry()->data.offer().buying;
```
vogel: Looks very complicated to get into the "buing" and "selling" fields, any chance to make that simpler?

src/transactions/OfferExchange.cpp
```diff
@@ -222,86 +204,106 @@ OfferExchange::crossOffer(OfferFrame& sellingWheatOffer,
...
+        auto header = ls.loadHeader();
+        accountB.addNumEntries(header, -1);
+        header->invalidate();
+        accountB.invalidate();
     }
     else
     {
-        sellingWheatOffer.getOffer().amount -= numWheatReceived;
-        sellingWheatOffer.storeChange(mDelta, db);
+        sellingWheatOffer->entry()->data.offer().amount -= numWheatReceived;
```
vogel: what about stellar::addBalance here?

src/transactions/CreateAccountOpFrame.cpp
```diff
@@ -31,18 +33,13 @@ CreateAccountOpFrame::CreateAccountOpFrame(Operation const& op,
...
-                              LedgerManager& ledgerManager)
+CreateAccountOpFrame::doApply(Application& app, LedgerState& ls)
 {
-    AccountFrame::pointer destAccount;
-
-    Database& db = ledgerManager.getDatabase();
-
-    destAccount =
-        AccountFrame::loadAccount(delta, mCreateAccount.destination, db);
+    auto destAccount = stellar::loadAccount(ls, mCreateAccount.destination);
```
vogel: why not make it ls.loadAccount(....) ?

src/transactions/TransactionFrame.cpp
```diff
@@ -121,7 +124,20 @@ TransactionFrame::getMinFee(LedgerManager const& lm) const
...
     }
 
-    return lm.getTxFee() * count;
+    return getCurrentTxFee(lsr) * count;
+}
+
+int64_t
+TransactionFrame::getMinFee(std::shared_ptr<LedgerHeaderReference> header) const
+{
+    size_t count = mOperations.size();
```
vogel: use std::max here to select between 1 and mOperations.size

src/test/TxTests.cpp
```diff
@@ -180,7 +181,7 @@ void
 checkTransaction(TransactionFrame& txFrame, Application& app)
 {
     REQUIRE(txFrame.getResult().feeCharged ==
-            app.getLedgerManager().getTxFee());
+            getCurrentTxFee(app.getLedgerStateRoot()));
```
vogel: I think you can as well rename it to getCurrentOpFee, as in reality it is fee per operation, not per transaction

src/herder/HerderImpl.cpp
```diff
@@ -343,19 +344,23 @@ HerderImpl::recvTransaction(TransactionFramePtr tx)
         }
     }
 
-    if (!tx->checkValid(mApp, highSeq))
     {
-        return TX_STATUS_ERROR;
+        LedgerState ls(lsOuter);
```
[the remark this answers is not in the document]
vogel: or with an enum LedgerState ls(lsOuter, LedgerState::COMMIT_BY_DEFAULT);

vogel:
The first thing to mention is that some simplifications could be made to avoid repetition of code, for example:

```
    auto header = ls.loadHeader();
    uint32_t seq = getCurrentLedgerNum(header);
    header->invalidate();

    auto header = ls.loadHeader();
    sourceAccount.addNumEntries(header, -1);
    header->invalidate();
```

Could be:

```
    uint32_t seq = getCurrentLedgerNum(ls);
    sourceAccount.addNumEntries(ls, -1);
```

I also agree with Nicolas that we need a RAII class to get rid of all of those load/invalidate pairs, because only few of them can be removed as above with replacing some arguments with LedgerState.

The next thing would be to rename `invalidate` method. Its very misleading to me. It does not mean, that `header` value got invalid. For me it is more like `release` from smart pointers.

`header->invalidate` suggest that header value is now invalid, which is not true, only pointer/reference is now invalid
`header->release` suggest that header pointer/reference is now invalid, as its released its value, so other pointers to it can be created.

The problem is that it can cause different kind of confusion, as `header.release()` is also defined by shared_ptr, so it is probably still not good name. What about `disown` known from Unix/Linux?

I don't like that ::forget() method requires passing to it LedgerState. Is that always the same LedgerState given LedgerEntry was created with (it looks so). Why not save LedgerState in LedgerEntry and use it for this case? It can prevent some bugs in future.

PR #1601, opened 2018-03-14
Record key SCP metrics

marta-lokhova:
Record timings for the following parts of SCP:
* Time from nomination starting to first prepare 
* Time from first prepare to value externalizing

### DIFF
--- a/src/herder/Herder.h
+++ b/src/herder/Herder.h
@@ -55,6 +55,9 @@ class Herder
     // How many ledgers in the past we keep track of
     static uint32 const MAX_SLOTS_TO_REMEMBER;
 
+    // Threshold used to filter out irrelevant events.
+    static std::chrono::nanoseconds const TIMERS_THRESHOLD_NANOSEC;
+
     static std::unique_ptr<Herder> create(Application& app);
 
     enum State
--- a/src/herder/HerderSCPDriver.cpp
+++ b/src/herder/HerderSCPDriver.cpp
@@ -17,6 +17,7 @@
 #include "xdr/Stellar-SCP.h"
 #include "xdr/Stellar-ledger-entries.h"
 #include <medida/metrics_registry.h>
+#include <util/format.h>
 #include <xdrpp/marshal.h>
 
 namespace stellar
@@ -53,6 +54,10 @@ HerderSCPDriver::SCPMetrics::SCPMetrics(Application& app)
           app.getMetrics().NewCounter({"herder", "state", "current"}))
     , mHerderStateChanges(
           app.getMetrics().NewTimer({"herder", "state", "changes"}))
+    , mNominateToPrepare(
+          app.getMetrics().NewTimer({"scp", "timing", "nominated"}))
+    , mPrepareToExternalize(
+          app.getMetrics().NewTimer({"scp", "timing", "externalized"}))
 {
 }
 
@@ -701,6 +706,7 @@ void
 HerderSCPDriver::startedBallotProtocol(uint64_t slotIndex,
                                        SCPBallot const& ballot)
 {
+    recordSCPEvent(slotIndex, false);
     mSCPMetrics.mStartBallotProtocol.Mark();
 }
 void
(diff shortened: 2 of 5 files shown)


src/herder/HerderSCPDriver.cpp
```diff
@@ -722,4 +728,97 @@ HerderSCPDriver::acceptedCommit(uint64_t slotIndex, SCPBallot const& ballot)
 {
     mSCPMetrics.mAcceptedCommit.Mark();
 }
+
+void
+HerderSCPDriver::setMetricStartTime(uint64_t slotIndex,
+                                    VirtualClock::time_point start,
```
MonsieurNicolas: I would ditch the `start` parameter as it's always called with `now` (and rename the method to just be `recordSCPEvent`)

src/herder/HerderSCPDriver.cpp
```diff
@@ -722,4 +728,97 @@ HerderSCPDriver::acceptedCommit(uint64_t slotIndex, SCPBallot const& ballot)
...
+                                    bool isNomination)
+{
+    SCPTimingPtr timing;
```
MonsieurNicolas: don't use `shared_ptr` here, then you can simplify the code to just be:

`auto &timing = mSCPExecutionTimes[slotIndex];`

src/herder/HerderSCPDriver.cpp
```diff
@@ -722,4 +728,97 @@ HerderSCPDriver::acceptedCommit(uint64_t slotIndex, SCPBallot const& ballot)
...
+void
+HerderSCPDriver::recordSCPExecutionMetrics(uint64_t slotIndex)
+{
+    auto externalizeStart = mApp.getClock().now();
+
+    auto SCPTimingIt = mSCPExecutionTimes.find(slotIndex);
+    if (SCPTimingIt == mSCPExecutionTimes.end())
+    {
+        // No metric was recorded
+        CLOG(WARNING, "Herder")
```
MonsieurNicolas: remove logging as it's not useful

src/herder/HerderSCPDriver.cpp
```diff
@@ -722,4 +728,97 @@ HerderSCPDriver::acceptedCommit(uint64_t slotIndex, SCPBallot const& ballot)
...
+        {
+            // Update the metric
+            mSCPMetrics.mPrepareToExternalize.Update(prepareDelta);
+        }
+        CLOG(DEBUG, "Herder") << fmt::format("Prepare delta for slot {} is {}",
+                                             slotIndex, prepareDelta.count());
+    }
+
+    // Clean up timings map
+    mSCPExecutionTimes.erase(slotIndex);
```
MonsieurNicolas: you need to properly erase execution times, see how it's done with `mSCPTimers`

src/herder/HerderSCPDriver.cpp
```diff
@@ -722,4 +728,97 @@ HerderSCPDriver::acceptedCommit(uint64_t slotIndex, SCPBallot const& ballot)
...
+
+    auto checkThreshold = [&](std::chrono::nanoseconds delta) -> bool {
+        if (delta < Herder::NOMINATION_THRESHOLD_NANOSECS)
+        {
+            auto qset = mApp.getConfig().QUORUM_SET;
+
+            if (!(qset.innerSets.size() == 0 && qset.validators.size() == 1 &&
+                  qset.validators[0] == getSCP().getLocalNodeID()))
+            {
+                CLOG(WARNING, "Herder")
```
MonsieurNicolas: same here those logs are not useful

src/herder/HerderSCPDriver.cpp
```diff
@@ -722,4 +728,97 @@ HerderSCPDriver::acceptedCommit(uint64_t slotIndex, SCPBallot const& ballot)
...
+        // No metric was recorded
+        CLOG(WARNING, "Herder")
+            << "No SCP metrics recorded for slot, " << slotIndex
+            << " this should not have happened.";
+        return;
+    }
+
+    auto SCPTiming = SCPTimingIt->second;
+
+    auto checkThreshold = [&](std::chrono::nanoseconds delta) -> bool {
```
MonsieurNicolas: it seems that this lambda should instead take the parameters:
start, end, a reference to the timer to update and a string for the logging.

Then it's just a call like
`recordTiming(*SCPTiming->mNominationStart, *SCPTiming->mPrepareStart, mSCPMetrics.mNominateToPrepare, "nomination");`

src/herder/HerderSCPDriver.cpp
```diff
@@ -722,4 +728,97 @@ HerderSCPDriver::acceptedCommit(uint64_t slotIndex, SCPBallot const& ballot)
...
+        if (delta < Herder::NOMINATION_THRESHOLD_NANOSECS)
```
MonsieurNicolas: the code would be tighter if instead you were computing the threshold at the beginning of the function (outside of the lambda) as single node just means that the threshold is 0

src/herder/HerderSCPDriver.cpp
```diff
@@ -53,6 +54,10 @@ HerderSCPDriver::SCPMetrics::SCPMetrics(Application& app)
           app.getMetrics().NewCounter({"herder", "state", "current"}))
     , mHerderStateChanges(
           app.getMetrics().NewTimer({"herder", "state", "changes"}))
+    , mNominateToPrepare(
+          app.getMetrics().NewTimer({"scp", "value", "nominate"}))
```
MonsieurNicolas: better names: `{"scp", "timing", "nominated" }` and `{"scp", "timing", "externalized"}`

src/herder/Herder.h
```diff
@@ -55,6 +55,9 @@ class Herder
     // How many ledgers in the past we keep track of
     static uint32 const MAX_SLOTS_TO_REMEMBER;
 
+    // Min time between before first prepare (to avoid metric skewing)
+    static std::chrono::nanoseconds const NOMINATION_THRESHOLD_NANOSECS;
```
MonsieurNicolas: comment doesn't really correspond to reality (it's not just about nomination):
this threshold is used to filter outliers when timing events.
I would just call it `TIMERS_THRESHOLD_NANOSEC`

src/herder/HerderSCPDriver.h
```diff
@@ -164,11 +168,27 @@ class HerderSCPDriver : public SCPDriver
...
     };
 
     SCPMetrics mSCPMetrics;
 
+    struct SCPTiming
+    {
+        optional<VirtualClock::time_point> mNominationStart;
+        optional<VirtualClock::time_point> mPrepareStart;
+    };
+    using SCPTimingPtr = std::shared_ptr<SCPTiming>;
```
MonsieurNicolas: per my earlier comment: ditch the shared_ptr

marta-lokhova:
updated

src/herder/HerderSCPDriver.cpp
```diff
@@ -730,95 +730,76 @@ HerderSCPDriver::acceptedCommit(uint64_t slotIndex, SCPBallot const& ballot)
...
         return;
     }
 
     auto SCPTiming = SCPTimingIt->second;
 
-    auto checkThreshold = [&](std::chrono::nanoseconds delta) -> bool {
-        if (delta < Herder::NOMINATION_THRESHOLD_NANOSECS)
+    auto recordTiming = [&](VirtualClock::time_point start,
+                            VirtualClock::time_point end, medida::Timer& timer,
+                            std::string logStr) {
```
MonsieurNicolas: `logStr` should be const ref

src/herder/HerderSCPDriver.cpp
```diff
@@ -730,95 +730,76 @@ HerderSCPDriver::acceptedCommit(uint64_t slotIndex, SCPBallot const& ballot)
...
-    mSCPExecutionTimes[slotIndex] = timing;
 }
 
 void
 HerderSCPDriver::recordSCPExecutionMetrics(uint64_t slotIndex)
 {
     auto externalizeStart = mApp.getClock().now();
 
+    // Use threshold of 0 in case of a single node
+    auto qset = mApp.getConfig().QUORUM_SET;
```
MonsieurNicolas: use `auto& qset`

src/herder/HerderSCPDriver.cpp
```diff
@@ -730,95 +730,76 @@ HerderSCPDriver::acceptedCommit(uint64_t slotIndex, SCPBallot const& ballot)
...
-            << "No SCP metrics recorded for slot, " << slotIndex
-            << " this should not have happened.";
         return;
     }
 
     auto SCPTiming = SCPTimingIt->second;
 
-    auto checkThreshold = [&](std::chrono::nanoseconds delta) -> bool {
-        if (delta < Herder::NOMINATION_THRESHOLD_NANOSECS)
+    auto recordTiming = [&](VirtualClock::time_point start,
```
MonsieurNicolas: use `auto &SCPTiming`

marta-lokhova:
fixed

src/herder/HerderSCPDriver.cpp
```diff
@@ -722,4 +728,78 @@ HerderSCPDriver::acceptedCommit(uint64_t slotIndex, SCPBallot const& ballot)
...
+    if (SCPTimingIt == mSCPExecutionTimes.end())
+    {
+        return;
+    }
+
+    auto& SCPTiming = SCPTimingIt->second;
+
+    auto recordTiming = [&](VirtualClock::time_point start,
+                            VirtualClock::time_point end, medida::Timer& timer,
+                            const std::string& logStr) {
```
MonsieurNicolas: we follow the convention `std::string const& `

MonsieurNicolas:
r+ d27cb8ee995ff585829b750b2cf862b641510e5f

PR #1606, opened 2018-03-16
Outbound connectivity fixes

MonsieurNicolas:
this PR addresses a few problems that I found while running tests:
* too many connections being attempted would result in an error "too many open files"
* some minor validation bug introduced by 7cc86833b2813bc32443b18302c9547407ac4ab7
* SEND_PEERS was potentially returning more peers than needed
* on startup, node could max out its connections by attempting to connect to too many nodes all at once

Those problems combined (especially the last one) would cause the node to appear stuck for a few seconds while queuing up potentially thousands of outbound connections


src/overlay/OverlayManagerImpl.cpp
```diff
@@ -222,7 +231,11 @@ OverlayManagerImpl::getPreferredPeersFromConfig()
 std::vector<PeerRecord>
 OverlayManagerImpl::getPeersToConnectTo(int maxNum)
 {
-    const int batchSize = std::max(20, maxNum);
+    // don't connect to too many peers at once
+    maxNum = std::min(maxNum, 50);
+
+    // batch is how many peers to load from the database every time
+    const int batchSize = std::max(50, maxNum);
```
vogel: I think it will be 50 every time now as it is basically std::max(50, std::min(any, 50))

src/xdr/Stellar-overlay.x
```diff
@@ -112,7 +112,7 @@ case DONT_HAVE:
 case GET_PEERS:
     void;
 case PEERS:
-    PeerAddress peers<>;
```
vogel: that does not change the binary format?

src/overlay/OverlayManagerImpl.cpp
MonsieurNicolas: yes this is correct, but because they are different context I prefer to keep the formulas in code as they don't need to be both 50

src/xdr/Stellar-overlay.x
MonsieurNicolas: no change in the binary format, the `<N>` notation only changes the marshaller to enforce a max

MonsieurNicolas:
@latobarita: retry

PR #1610, opened 2018-03-20
improve caching shared keys

### LINKED ISSUE #1592: overlay: investigate key exchange caching
looking at `PeerAuth::getSharedKey`, it looks like caching *may* be wrong:

it looks like it's computing `EcdhDeriveSharedKey` based on the parameter `role`, yet the cache doesn't seem to remember which role was used to compute the key.

This may cause the key to be cached wrong if a peer decides to switch role later on (ie, decides to connect to a peer X after peer X got disconnected, with peer X being the initiator the previous time).

With the cache invalid, it may cause a bunch of errors such as:
`[Overlay WARNING] Received error (ERR_AUTH): unexpected MAC`

not sure if the solution is to just sort the values somehow (and get rid of that flag to compute the shared secret), keep both versions or to invalidate the cache

vogel:
Fixes #1592

### DIFF
--- a/src/overlay/PeerAuth.cpp
+++ b/src/overlay/PeerAuth.cpp
@@ -76,14 +76,16 @@ HmacSha256Key
 PeerAuth::getSharedKey(Curve25519Public const& remotePublic,
                        Peer::PeerRole role)
 {
-    if (mSharedKeyCache.exists(remotePublic))
+    auto key = PeerSharedKeyId{remotePublic, role};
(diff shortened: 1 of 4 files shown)


MonsieurNicolas:
r+ efc7c47b9f166acd17fc54a39cd85759dea69591

vogel:
Fixed compilation error

MonsieurNicolas:
r+ f3456f42cd628e83438b7032ad7227d97e04e9f6

PR #1612, opened 2018-03-20
properly fail RunCommandWork in case of exception in run()

### LINKED ISSUE #1577: spawn process failure not handled properly in history work
It looks like history gets borked when commands fail for some reason.

I've seen it fail with gzip:
```
[Process ERROR] Error starting process: posix_spawn() failed [ProcessManagerImpl.cpp:489]
[Process ERROR] When running: gzip -c /data/buckets/bucket-0a523ea9edf300595bf3a35392a2c92eb3b5db373db2bd3fc163fdfc437adb26.xdr [ProcessManagerImpl.cpp:490]
[Process ERROR] posix_spawn() failed: Cannot allocate memory [ProcessManagerImpl.cpp:432]
```

when this happens, it seems that it continues as if the process completed successfully, causing all sorts of problems downstream:
```
The user-provided path /data/buckets/bucket-0a523ea9edf300595bf3a35392a2c92eb3b5db373db2bd3fc163fdfc437adb26.xdr.gz does not exist.
...
[Work WARNING] Scheduling retry #1/5 in 1 sec, for put-remote-file bucket/0a/52/3e/bucket-0a523ea9edf300595bf3a35392a2c92eb3b5db373db2bd3fc163fdfc437adb26.xdr.gz
```

Unclear if this is a bug at the `ProcessManager` layer or at the `GunzipFileWork` layer

vogel:
Fixes #1577

### DIFF
--- a/src/process/ProcessManagerImpl.cpp
+++ b/src/process/ProcessManagerImpl.cpp
@@ -210,6 +210,10 @@ ProcessExitEvent::Impl::run()
                        &pi)     // Pointer to PROCESS_INFORMATION structure
     )
     {
+        if (si.hStdOutput != NULL)
+        {
+            CloseHandle(si.hStdOutput);
+        }
         CLOG(ERROR, "Process") << "CreateProcess() failed: " << GetLastError();
         throw std::runtime_error("CreateProcess() failed");
     }
--- a/src/process/ProcessManagerImpl.cpp
+++ b/src/process/ProcessManagerImpl.cpp
@@ -486,6 +486,7 @@ ProcessManagerImpl::maybeRunPendingProcesses()
         }
         catch (std::runtime_error& e)
         {
+            i->cancel(std::make_error_code(std::errc::io_error));
             CLOG(ERROR, "Process") << "Error starting process: " << e.what();
             CLOG(ERROR, "Process") << "When running: " << i->mCmdLine;
         }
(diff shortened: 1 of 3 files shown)


src/process/ProcessManagerImpl.cpp
```diff
@@ -486,6 +486,7 @@ ProcessManagerImpl::maybeRunPendingProcesses()
         }
         catch (std::runtime_error& e)
         {
+            i->cancel(std::make_error_code(std::errc::io_error));
```
MonsieurNicolas: it looks like both versions (posix and windows) of `ProcessExitEvent::Impl::run()` are buggy and don't clean up all their resources when throwing (posix leaks `fileActions` and windows leaks `hStdOutput` on the second throw), we should fix this as part of this PR

vogel:
I think it is fixed now

src/process/ProcessManagerImpl.cpp
```diff
@@ -210,6 +210,7 @@ ProcessExitEvent::Impl::run()
                        &pi)     // Pointer to PROCESS_INFORMATION structure
     )
     {
+        CloseHandle(si.hStdOutput);
```
MonsieurNicolas: I think you have to do
```
if (si.hStdOutput != NULL)
{
    CloseHandle(si.hStdOutput);
}
```

vogel:
Should be OK now.

MonsieurNicolas:
r+ f2fb7c4b667bdab6b1a640d89bbd977310e651c0

PR #1616, opened 2018-03-21
Remove busy loops

### LINKED ISSUE #1605: review and fix busy loops in code
We should review the code base for busy loops (loops that are waiting on external events without using proper eventing mechanism).

An example busy loop is:
https://github.com/stellar/stellar-core/blob/34d8265301d5e0689712aada55fbd8f3061f2f58/src/work/WorkManager.h#L37-L40

This one causes `--newhist` to use 100% CPU when a initializing a history archive (waiting for an external process to complete).

In some case, it may be overkill to await on the right event, at a minimum we should add a call to sleep in the loop, as we already do in other places in the code
`std::this_thread::sleep_for(chrono::milliseconds(50));`

jonjove:
Resolves #1605

### DIFF
--- a/src/history/HistoryManagerImpl.cpp
+++ b/src/history/HistoryManagerImpl.cpp
@@ -347,15 +347,8 @@ InferredQuorum
 HistoryManagerImpl::inferQuorum()
 {
     InferredQuorum iq;
-    bool done = false;
-    auto handler = [&done](asio::error_code const& ec) { done = true; };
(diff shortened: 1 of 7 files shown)


vogel:
Looks like there are no more calls to `executeWork` with bool set to false - it can be removed now.

MonsieurNicolas:
r+ 2588100d12e4dc88cf72f671a5298e4780e30bbe

MonsieurNicolas:
r+ 5cbf1f151026844afb2452c73be169689edf349c

PR #1617, opened 2018-03-22
Fix build warnings

mfontanini:
I found a few warnings when building this on gcc 5.4.0 (the default one on ubuntu xenial).

The main issue were some third party libraries, like asio and easylogging++, which had several warnings regarding "unused" variables/members and those were polluting the rest of the build. I used `-isystem` rather than `-I` to pull those include paths so now they're treated as an external directory and therefore they don't generate any warnings while including them on stellar-core.

Also, I noticed `common.mk` is using -I"$(top_srcdir)/src" and -I"$(top_builddir)/src" but they both map to the same directory. Maybe one of them should be removed?

### DIFF
--- a/common.mk
+++ b/common.mk
@@ -1,14 +1,14 @@
 # Hand-written file with variables common to all makefiles
 
 AM_CPPFLAGS = -DSQLITE_OMIT_LOAD_EXTENSION=1
-AM_CPPFLAGS += -I"$(top_srcdir)" -I"$(top_srcdir)/src" -I"$(top_builddir)/src"
+AM_CPPFLAGS += -isystem "$(top_srcdir)" -I"$(top_srcdir)/src" -I"$(top_builddir)/src"
 AM_CPPFLAGS += $(libsodium_CFLAGS) $(xdrpp_CFLAGS) $(libmedida_CFLAGS)	\
 	$(soci_CFLAGS) $(sqlite3_CFLAGS) $(libasio_CFLAGS)
-AM_CPPFLAGS += -I"$(top_srcdir)/lib"			\
-	-I"$(top_srcdir)/lib/autocheck/include"		\
-	-I"$(top_srcdir)/lib/cereal/include"		\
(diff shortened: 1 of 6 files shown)


MonsieurNicolas:
> Also, I noticed common.mk is using -I"$(top_srcdir)/src" and -I"$(top_builddir)/src" but they both map to the same directory. Maybe one of them should be removed?

They are not always the same, you can check out https://www.gnu.org/software/automake/manual/html_node/Program-Variables.html under `AM_CPPFLAGS` that gives the details.

MonsieurNicolas:
Thanks, I am marking this for merge

MonsieurNicolas:
r+ ebbd4f8286383d0e2dc456f3d21217ca59877b22

MonsieurNicolas:
Yes, it looks like you didn't run clang-format on the second commit @mfontanini

mfontanini:
Hmm okay. I imagine this either shouldn't have made it to the repo (this file was modified a couple of weeks ago) or this shouldn't fail for the parts of the code that weren't actually modified by the commit. Anyway, I can push a PR for this in a few hours.

MonsieurNicolas:
it's the [change you made](https://github.com/stellar/stellar-core/pull/1617/commits/25f3444d87c73abb7fb283d86a6acf00f71813e1#diff-41c5d7c517d159aef966bc24a50a7298R251) that causes that function to get reformatted

MonsieurNicolas:
Thanks! Yes, maybe we should cause "format" to be called when building to avoid this kind of problem. Let me open an issue about this

MonsieurNicolas:
r+ 39e97e60b223e13fc859cf5e5a08ebb3ea278213

mfontanini:
Looks like this hit travis' build time limit on Linux?

MonsieurNicolas:
yes, I am rekicking it

MonsieurNicolas:
@latobarita: retry

PR #1623, opened 2018-03-24
Move using xdr::operator statements into a separate header

### LINKED ISSUE #1529: Standardize on the use of `using xdr::operator<==>`
It looks like we have many places where we import xdr related methods such as the `xdr::operator` family of operators directly into the `stellar` namespace, things like
```
namespace stellar
{

using xdr::operator==;
```

The problem is that when this is done like this we end up with unreliable errors such as #1528 as we depend on the include order (which changes over time).

We need to standardize by doing one of:
* only import those operators in the scope where they are used
```
void
withshow()
{
  using xdr::operator<<;
```
* import those operators at a "root" level include (such as `types.h`)

it's unclear which solution actually works best given the rules for unqualified name lookup.

it *seems* that the scoped based solution does not really work as we use things like `std::map<LedgerKey, EntryFrame::pointer>` as parameter (or return value) which would cause the need for inclusion in all scopes where the parameter is used.

mfontanini:
Following up on what #1529 says, the use of `using xdr::operator...` statements within different scopes doesn't really fix this when using a type outside of the scope it's defined in that requires those operators to be available (e.g. `std::map`).

This PR creates a new header just for XDR operators. I thought this was a good way to approach this as there doesn't seem to be a "standard XDR" header that's included throughout the code and I didn't want to pollute something like `util/types.h` as this is very specific, but I can change it otherwise.

Fixes #1529

### DIFF
--- a/src/overlay/Peer.cpp
+++ b/src/overlay/Peer.cpp
@@ -20,6 +20,7 @@
 #include "overlay/StellarXDR.h"
 #include "util/Logging.h"
 #include "util/XDROperators.h"
+
 #include "medida/meter.h"
 #include "medida/metrics_registry.h"
 #include "medida/timer.h"
--- a/src/overlay/Peer.cpp
+++ b/src/overlay/Peer.cpp
@@ -19,7 +19,7 @@
 #include "overlay/PeerRecord.h"
 #include "overlay/StellarXDR.h"
 #include "util/Logging.h"
-
+#include "util/XDROperators.h"
 #include "medida/meter.h"
 #include "medida/metrics_registry.h"
 #include "medida/timer.h"
@@ -37,8 +37,6 @@ namespace stellar
 
 using namespace std;
 using namespace soci;
-
-using xdr::operator<;
 
 medida::Meter&
 Peer::getByteReadMeter(Application& app)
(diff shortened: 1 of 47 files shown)


MonsieurNicolas:
@vogel can you review this PR ?

vogel:
Looks good to me. Could you please rebase on current master?

mfontanini:
Rebased, sorry for the delay.

src/overlay/Peer.cpp
```diff
@@ -19,7 +19,7 @@
 #include "overlay/PeerRecord.h"
 #include "overlay/StellarXDR.h"
 #include "util/Logging.h"
-
+#include "util/XDROperators.h"
```
vogel: please leave empty line after this, then clang-format wont be neccessary and library headers will be separated from inner ones

vogel:
One small change required - an empty line between #includes

mfontanini:
Ah okay, I was confused as to why I had to run clang format after rebasing, that makes sense. Fixed!

vogel:
r+ e98f856dee1f05bf1e41b03cf077eff0fca4735d

PR #1624, opened 2018-03-24
Include soci.h directly rather than through SociNoWarnings.h

mfontanini:
I didn't see `SociNoWarnings.h` when I made PR https://github.com/stellar/stellar-core/pull/1617. This header is not needed anymore, as using `-isystem` will ignore all these warnings on third party libraries.

I'm not entirely sure what the include directive convention is for third party libs. I see paths like `"xdrpp/header.h"`, `<xdrpp/header.h>` and `"lib/xdrpp/xdrpp/"` in different include directives so I'm not sure if using `<soci.h>` is acceptable.


vogel:
Is -isystem awailable for both gcc and clang? I was not aware of it when I made this workaround header file.

I would like our includes to be consistent, so we need to come up with a convention. My preferred one it to use "those-includes" only for files in "src/" and <those> for system includes and library includes. Not sure if everybody will agree with that.

MonsieurNicolas:
yes, `-isystem` is supported on both gcc and clang.

I agree with using `#include "foo"` for basically "files that we own and manage" (everything under `src/`) vs using `#include <foo>` for system and external dependencies (things under `root/lib`)

mfontanini:
I agree with that. Would you want the clang format file to prioritize third party includes that are pulled from `lib` in a different way from other headers included via `<header>` so they show up before e.g. `#include <string>`? Not sure how that can work but I can submit a PR making all of these consistent.

MonsieurNicolas:
sounds good.

re: prioritization of `lib`, I think it's already the case (but uses quotes), but the rule will have to be updated to use `<`:
```
IncludeCategories:
  - Regex:           '^"util/asio\.h'
    Priority:        -1
  - Regex:           '^".*'
    Priority:        1
  - Regex:           '^"lib/.*'
    Priority:        2
  - Regex:           '^<.*'
    Priority:        3
  - Regex:           '.\*'
    Priority:        4
```

mfontanini:
Alright so I guess all includes for third party libraries should be changed from e.g. `xdrpp/types.h` to `lib/xdrpp/xdrpp/types.h`? That would actually be the case for xdrpp.
Otherwise the regex could proritize any paths with slashes withing an `#include <>` vs other includes with brackets so that you can do `#include <xdrpp/types.h>` and that is still sorted so it shows up before the C++ header includes.

MonsieurNicolas:
yes, xdrpp should not get special treatment in this context.

As includes get reordered, make sure that we don't end up reintroducing bugs related to how templates get instantiated as fixed in ecabfa6f19f02b9113fc6fe520ff786d9eabf2f7 (basically template specialization differ depending on include order). I don't know if there is a simple way to check in general, we found some of those before as visual studio links differently (doesn't allow duplicate symbols for different implementations, I think there is a Mac specific flag for clang but that's not super reliable).

mfontanini:
So I started to look at this and it really looks terrible if all third party libraries are included through `lib`. e.g. medida includes end up having the path `lib/libmedida/src/medida/medida.h`. You'll basically have to always copy some include from another file because there's no way you'll remember that full path. 

What about using symlinks to simplify all include paths? I mean having some specific directory within `lib` that contains symlinks to the include path for all third party libs, like:

```
stellar-core/lib/include $ ls -l
total 12
lrwxrwxrwx 1 matias matias 24 Mar 28 19:30 medida -> ../libmedida/src/medida/
lrwxrwxrwx 1 matias matias 27 Mar 28 19:30 sodium -> ../libsodium/src/libsodium/
lrwxrwxrwx 1 matias matias 15 Mar 28 19:30 xdrpp -> ../xdrpp/xdrpp/
...
```

This way you'll be able to do e.g. `#include <medida/timer.h>`.

mfontanini:
Actually there should be a `lib` in there so you can do `#include <lib/medida/timer.h>`.

MonsieurNicolas:
I don't know about the symlinks: maybe we keep this change as is and don't bother with the move to system includes

MonsieurNicolas:
r+ 74b79ed3d4ea1b7881f9546f51242e447a2c44dd

PR #1625, opened 2018-03-25
Catch exceptions when formatting command string templates

mfontanini:
If the configuration file being used contains any invalid format strings (e.g. `echo hello {0`) for commands, stellar-core blows up without any logging messages specifying what happened.

This PR catches exceptions when formatting strings in `HistoryArchive`, logs them and then throws again. As far as I can see, this is the only class that deals with parsing user provided format strings.

### DIFF
--- a/src/history/HistoryArchive.cpp
+++ b/src/history/HistoryArchive.cpp
@@ -35,6 +35,22 @@ namespace stellar
 
 unsigned const HistoryArchiveState::HISTORY_ARCHIVE_STATE_VERSION = 1;
 
+template <typename... Tokens>
+std::string
+formatString(std::string const& templateString, Tokens const&... tokens)
+{
+    try
+    {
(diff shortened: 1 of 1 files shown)


MonsieurNicolas:
cool, thank you.
For background the reason we do not catch exceptions at the top level (where we could log the same thing) is so that we can get good core dumps (when rethrowing exceptions, even with `throw;`, we lose a lot of context) - if you know of a better solution let me know.

MonsieurNicolas:
r+ b70b91ecf462e76406229535b3c96cccbcba4b6c

mfontanini:
This failure looks like a fluke. Probably just needs re-triggering.

MonsieurNicolas:
@latobarita: retry

MonsieurNicolas:
@latobarita: retry

PR #1627, opened 2018-03-26
Trigger catchup

### LINKED ISSUE #1613: Approach to downloading snapshot seems flawed
It seems there is a problem with the way we trigger history catchup:
from my observations, nodes try to download checkpoints way too early (pointing to a bug in how we compute the amount of time we should wait before even trying to download the snapshot).

This in turn creates delays as the downloads are failing for no good reason (and is confusing operators).

I think it would make a lot more sense to not do it based on time but on the ledgers that we are observing from consensus (we can still give an ETA based on how far currently observed ledger is from the next checkpoint):
as we are already queuing up those ledgers, we could simply wait to trigger catchup when a ledger observed meets the "checkpoint criteria" instead of blindly starting the catchup process.

Basically we would change `LedgerManagerImpl::startCatchUp` to not always set the state to `LM_CATCHING_UP_STATE` but instead move to that state if `catchupHistory` can start work (ie `configuration` is on a snapshot boundary).

vogel:
Fixes #1613

### DIFF
--- a/src/ledger/LedgerManagerImpl.cpp
+++ b/src/ledger/LedgerManagerImpl.cpp
@@ -437,7 +437,7 @@ LedgerManagerImpl::valueExternalized(LedgerCloseData const& ledgerData)
             // that way we will have a way to verify history consistency -
             // compare previousLedgerHash of buffered ledger with last one
             // downloaded from history
-            startCatchUp({ledgerData.getLedgerSeq() - 1, getCatchupCount(mApp)},
+            startCatchup({ledgerData.getLedgerSeq() - 1, getCatchupCount(mApp)},
                          false);
         }
         break;
@@ -479,7 +479,7 @@ LedgerManagerImpl::valueExternalized(LedgerCloseData const& ledgerData)
 }
 
 void
-LedgerManagerImpl::startCatchUp(CatchupConfiguration configuration,
+LedgerManagerImpl::startCatchup(CatchupConfiguration configuration,
                                 bool manualCatchup)
 {
     auto lastClosedLedger = getLastClosedLedgerNum();
--- a/src/ledger/LedgerManagerImpl.cpp
+++ b/src/ledger/LedgerManagerImpl.cpp
@@ -393,10 +393,7 @@ LedgerManagerImpl::valueExternalized(LedgerCloseData const& ledgerData)
             if (mLastClosedLedger.hash ==
                 ledgerData.getTxSet()->previousLedgerHash())
             {
-                if (st == LM_BOOTING_STATE)
-                {
-                    setState(LM_SYNCED_STATE);
-                }
+                setState(LM_SYNCED_STATE);
                 closeLedger(ledgerData);
                 CLOG(INFO, "Ledger")
                     << "Closed ledger: " << ledgerAbbrev(mLastClosedLedger);
--- a/src/ledger/LedgerManagerImpl.cpp
+++ b/src/ledger/LedgerManagerImpl.cpp
@@ -422,56 +422,115 @@ LedgerManagerImpl::valueExternalized(LedgerCloseData const& ledgerData)
                 << mLastClosedLedger.header.ledgerSeq
                 << ", network closed ledger " << ledgerData.getLedgerSeq();
 
-            assert(mSyncingLedgers.size() == 0);
-            auto addResult = mSyncingLedgers.add(ledgerData);
-            assert(addResult == SyncingLedgerChainAddResult::CONTIGUOUS);
-            mSyncingLedgersSize.set_count(mSyncingLedgers.size());
-            CLOG(INFO, "Ledger")
-                << "Close of ledger " << ledgerData.getLedgerSeq()
-                << " buffered, starting catchup";
-
-            // catchup just before first buffered ledger
-            // that way we will have a way to verify history consistency -
-            // compare previousLedgerHash of buffered ledger with last one
-            // downloaded from history
-            startCatchup({ledgerData.getLedgerSeq() - 1, getCatchupCount(mApp)},
-                         false);
+            initializeCatchup(ledgerData);
         }
         break;
 
     case LedgerManager::LM_CATCHING_UP_STATE:
     {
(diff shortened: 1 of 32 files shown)


src/ledger/LedgerManagerImpl.h
```diff
@@ -51,6 +51,14 @@ class LedgerManagerImpl : public LedgerManager
...
 
     SyncingLedgerChain mSyncingLedgers;
+    uint32_t mCatchUpTriggerLedger{0};
+    bool mWaitingForCatchUpTriggerLedger{false};
+
+    void initializeCatchup(LedgerCloseData const& ledgerData,
+                           uint32_t catchUpTriggerLedger);
+    void continueCatchup(LedgerCloseData const& ledgerData);
+    void addToSyncingLedgers(LedgerCloseData const& ledgerData);
+    void starCatchupIf(uint32_t lastReceivedLedgerSeq);
```
marta-lokhova: minor: looks like a typo, should be `startCatchupIf`

src/ledger/LedgerManagerImpl.cpp
```diff
@@ -638,66 +641,15 @@ LedgerManagerImpl::historyCaughtup(asio::error_code const& ec,
...
-                    << ", trying to apply buffered close " << lcd.getLedgerSeq()
-                    << " with txhash "
-                    << hexAbbrev(lcd.getTxSet()->getContentsHash());
-                mSyncingLedgers = {};
-                mSyncingLedgersSize.set_count(mSyncingLedgers.size());
-                CLOG(ERROR, "Ledger") << "Catchup will restart at next close.";
-                setState(LM_BOOTING_STATE);
-                return;
-            }
+            assert(lcd.getLedgerSeq() ==
```
MonsieurNicolas: I don't see how this really works, there was a code path here that leads to `LM_BOOTING_STATE` - which cannot happen with the new code (as far as I can tell)?
Meaning that if we "catchup" to the checkpoint but there is a gap in history it will set the state to synced even though it's not (and what it logged "this round of catchup will fail and restart." will only happen after some timeout?).

vogel: This is not possible because of how SyncingLedgerChain works - it can only contain contiguous chain of ledgers.

You can look at the version of LedgerManager before implementing SyncingLedgerChain - https://github.com/stellar/stellar-core/blob/9c29696964ad1268fe442bbf6ea5de17e3485d06/src/ledger/LedgerManagerImpl.cpp#L407

It was working the same way, so the only way it could happen even then was if the first ledger in mSyncingLedgers was overshooting. And it was impossible, because then verification would not pass at all. So this code was never used.

Instead when there is a gap, the catchup goes as far as it can, then it gets to SYNCED state until next ledger gets externalized and then it goes back to CATCHING_UP.

MonsieurNicolas: I see. As catchup can take hours, there is a non trivial chance of having gaps: we should make it that at the end of catchup it waits for a ledger to close (probably by adding a new state `LM_FINISH_CATCHUP_STATE`), that way we can formalize the transition to "Synced!" and avoid triggering publish, etc when not fully in sync yet.

src/ledger/LedgerManagerImpl.cpp
```diff
@@ -424,51 +424,16 @@ LedgerManagerImpl::valueExternalized(LedgerCloseData const& ledgerData)
...
-                << "Close of ledger " << ledgerData.getLedgerSeq()
-                << " buffered, starting catchup";
-
-            // catchup just before first buffered ledger
-            // that way we will have a way to verify history consistency -
-            // compare previousLedgerHash of buffered ledger with last one
-            // downloaded from history
-            startCatchUp({ledgerData.getLedgerSeq() - 1, getCatchupCount(mApp)},
-                         false);
+            auto catchUpTriggerLedger =
```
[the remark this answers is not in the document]
vogel: It is a leftover from version when all of that code was in CatchupManager. I'll remove that. It can always be re-added later, when we have more time to refactor Herder/Ledger/Catchup managers.

MonsieurNicolas:
@vogel you need to rebase this on top of master

vogel:
Not yet ready, some problems with flooding.

vogel:
Tests fixed. Ready for review.

src/ledger/LedgerManagerImpl.cpp
```diff
@@ -388,47 +408,42 @@ LedgerManagerImpl::valueExternalized(LedgerCloseData const& ledgerData)
...
-                << ", network closed ledger " << ledgerData.getLedgerSeq();
-
             initializeCatchup(ledgerData);
         }
         break;
+        }
+    }
+    break;
 
     case LedgerManager::LM_CATCHING_UP_STATE:
```
MonsieurNicolas: The code would be simpler and easier to understand if we only had one list of states for the state machine: right now you have nested switches and state transitions are nested too, this is not how people typically describe state machines and while it looks like name mangling, it forces to enumerate all possible states for the state machine and all transitions.

So, replace `LM_CATCHING_UP_STATE` by something like this:
```
LM_CATCHING_UP_WAITING_TRIGGER_STATE,
LM_CATCHING_UP_APPLYING_STATE,
LM_CATCHING_UP_FINALIZING_STATE
```

vogel: I did it for two reasons:
1. adding LM_CATCHING_UP... states complicated many other switch statements which only care if  any of LM_CATCHING_UP... is on
2. at some point I hope that whole catchup code will be in CatchupManager, then it will definitely require separate state machine, which is now nested in LedgerManager state machine

MonsieurNicolas: yes I agree it makes sense in the context of a refactor of that code into `CatchupManager`, I guess we have to prioritize this refactor. I created issue #1674 to track this.
I'll merge this PR now anyways then

MonsieurNicolas:
r+ 4702d28c951b4c81ec9e832e4901070a7fdf67a9

PR #1630, opened 2018-03-27
change wrong phrase.

cmcm2222:
I think this phrase is wrong. 
so I fixed it.


MonsieurNicolas:
iff stands for "if and only if", it's a standard abbreviation for it

PR #1631, opened 2018-03-27
Dont adjust txrate for account creations

marta-lokhova:
An issue I noticed when running benchmark tests: looks like batching 100 account creation ops per tx (or any other high number of ops) with auto adjustment of tx per step doesn't work very well. The adjustment is based on how often steps are submitted, and in case of batching multiple ops in a tx, it looks like less steps per s are submitted (possibly because the system is more loaded dealing with these large txs?), so `tx per step` keeps snowballing.

### DIFF
--- a/src/simulation/LoadGenerator.cpp
+++ b/src/simulation/LoadGenerator.cpp
@@ -74,7 +74,7 @@ LoadGenerator::createRootAccount()
 }
 
 uint32_t
-LoadGenerator::getTxPerStep(uint32_t txRate)
+LoadGenerator::getTxPerStep(uint32_t txRate, bool isCreate)
 {
     // txRate is "per second"; we're running one "step" worth which is a
     // fraction of txRate determined by STEP_MSECS. For example if txRate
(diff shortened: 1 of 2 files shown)


MonsieurNicolas:
Not sure I understand why we would limit this at this layer: if some test is being run on hardware somewhere that cannot deal with auto rate for account creation... why is it passing `true` for auto rate?

marta-lokhova:
This is different from auto rate. `getTxPerStep` still gets called even with autoRate=false. The function decides how many transactions to submit in one load generation step. There's a comment that explains the rationale behind this:
```
   // There is a wrinkle here though which is that the tx-apply phase might
    // well block timers for up to half the close-time; plus we'll be probably
    // not be scheduled quite as often as we want due to the time it takes to
    // run and the time the network is exchanging packets. So instead of a naive
    // calculation based _just_ on target rate and STEP_MSECS, we also adjust
    // based on how often we seem to be waking up and taking loadgen steps in
    // reality.
```

MonsieurNicolas:
Oh I see. This doesn't seem right either then. This problem will exist for other transactions (it's just a matter of cranking txrate up).

Assuming txrate is achieveable, it should never "runaway", so either:
* txrate is too high in your particular case
* or the computation of `txPerStep` based on `1 min avg` is too high
https://github.com/marta-lokhova/stellar-core/blob/cb14aceffbf585401585d9fffae809db7b8232a9/src/simulation/LoadGenerator.cpp#L99

PR #1641, opened 2018-04-01
Actually kill subprocesses (issue #1636)

### LINKED ISSUE #1636: Child processes not terminated properly when stopping applications
This is not a problem when running stellar-core normally, only in test code.

Right now child processes seem to die because their `stdin` handle gets closed when the main process exits.

This does not seem to happen in test code:
Enabling tracing while performing the scenario described in https://github.com/stellar/stellar-core/issues/1635 shows that applications are started while child processes are still running.

From what I can see:
`ProcessManagerImpl::shutdown()` seems to not work as expected as processes keep running after shutdown is called - maybe calling `kill` explicitely is in order?

Notes:
* even if `shutdown` was terminating processes properly, its implementation is wrong as it loops over all processes for all applications instead of just the application that is stopping
* it looks like we should distinguish `ApplicationImpl::gracefulStop` from `ApplicationImpl::~ApplicationImpl`, with `gracefulStop` being the place where we should signal processes to exit somehow (with a `SIGINT` during "clean" shutdown, followed by a `SIGKILL` in the destructor).

mfontanini:
This fixes #1636. I had to make a few changes for this to work, following @MonsieurNicolas's comments on the ticket. The list of changes implemented here are:

* Now instead of sharing the map of subprocesses globally, those are kept in a per instance basis. Since there's still a need for forwarding signals properly to all `ProcessManager` instances, the instances themselves are registered globally so that whoever gets the signal can forward it to all of them. I had to use raw pointers here as there's no easy way to register instances on their constructor otherwise. I'm not a fan of this but it works.
* On Linux, `ProcessManager::shutdown` will send `SIGINT` to all subprocesses. Once the destructor is hit, then `SIGKILL` will be sent, regardless of whether `SIGINT` was already sent or not.
* On Windows, `TerminateProcess` is called on the destructor. I'm not sure what to do on shutdown as, as far as I can see, there's no `SIGINT` equivalent on that OS. Maybe it's worth calling `TerminateProcess` on shutdown and only do that on the destructor for cases when shutdown isn't initially called?
* I fixed a circular `shared_ptr` dependency between the `ProcessManagerImpl` and its subprocesses. Basically they both had a `shared_ptr` to each other so if any subprocesses were kept alive, then the process manager would never be destructed.
* I moved `gNumProcessesActive` to be an `atomic`. This is only really needed when decrementing the value in the `async_wait` callback on the Windows implementation of `ProcessExitEvent::Impl::run` as the rest of the accesses are protected by the `mImplsMutex` mutex. This could be replaced by a lock and a counter but I didn't want to introduce another global lock just for that.
* As for tests, on Linux it's simple: run `sleep 10` and then kill it right away. I made sure the process is actually dead after `stellar-core` stops running. On Windows I had to use `waitfor /T 10 pause` instead. The other alternatives I tried (e.g. `timeout`) fail because the process can't read from stdin, hence it aborts right away. I also made sure the process actually exits while testing this.

### DIFF
--- a/src/process/ProcessManagerImpl.cpp
+++ b/src/process/ProcessManagerImpl.cpp
@@ -47,10 +47,6 @@ ProcessManager::create(Application& app)
     return std::make_shared<ProcessManagerImpl>(app);
 }
 
-std::recursive_mutex ProcessManagerImpl::gManagersMutex;
-
-std::vector<ProcessManagerImpl*> ProcessManagerImpl::gManagers;
-
 std::atomic<size_t> ProcessManagerImpl::gNumProcessesActive{0};
 
 class ProcessExitEvent::Impl
@@ -121,11 +117,6 @@ ProcessManagerImpl::~ProcessManagerImpl()
     {
         killProcess(*pair.second);
     }
-
-    std::lock_guard<std::recursive_mutex> guard(gManagersMutex);
-    // Unregister this manager globally
-    gManagers.erase(std::remove(gManagers.begin(), gManagers.end(), this),
-                    gManagers.end());
 }
 
 bool
@@ -176,7 +167,6 @@ ProcessManagerImpl::ProcessManagerImpl(Application& app)
     , mIOService(app.getClock().getIOService())
     , mSigChild(mIOService)
 {
-    registerManager();
 }
 
 void
@@ -340,7 +330,6 @@ ProcessManagerImpl::ProcessManagerImpl(Application& app)
 {
     std::lock_guard<std::recursive_mutex> guard(mImplsMutex);
     startSignalWait();
-    registerManager();
 }
 
 void
@@ -358,21 +347,31 @@ ProcessManagerImpl::handleSignalWait()
     {
         return;
     }
-    for (;;)
+    // Store tuples (pid, status)
+    std::vector<std::tuple<int, int>> signaledChildren;
+    std::lock_guard<std::recursive_mutex> guard(mImplsMutex);
+    for (auto const& implPair : mImpls)
     {
+        const int pid = implPair.first;
         int status = 0;
-        int pid = waitpid(-1, &status, WNOHANG);
-        if (pid > 0)
+        // If we find the child for which we received this SIGCHLD signal,
+        // store the pid and status
+        if (waitpid(pid, &status, WNOHANG) > 0)
         {
-            std::lock_guard<std::recursive_mutex> guard(gManagersMutex);
-            for (ProcessManagerImpl* manager : gManagers)
-            {
-                manager->handleProcessTermination(pid, status);
-            }
+            signaledChildren.push_back(std::make_tuple(pid, status));
         }
-        else
+    }
+
+    if (!signaledChildren.empty())
+    {
+        CLOG(DEBUG, "Process") << "found " << signaledChildren.size()
+                               << " child processes that terminated";
+        // Now go all over all (pid, status) and handle them
+        for (auto const& pidStatus : signaledChildren)
         {
-            break;
+            const int pid = std::get<0>(pidStatus);
+            const int status = std::get<1>(pidStatus);
+            handleProcessTermination(pid, status);
         }
     }
     startSignalWait();
@@ -385,7 +384,7 @@ ProcessManagerImpl::handleProcessTermination(int pid, int status)
     auto pair = mImpls.find(pid);
     if (pair == mImpls.end())
     {
-        // Possible this pid belonged to another ProcessManager
+        CLOG(DEBUG, "Process") << "failed to find process with pid " << pid;
         return;
     }
     auto impl = pair->second;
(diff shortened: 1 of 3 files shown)


src/process/ProcessManagerImpl.cpp
```diff
@@ -291,83 +306,21 @@ ProcessManagerImpl::handleSignalWait()
...
     for (;;)
     {
         int status = 0;
         int pid = waitpid(-1, &status, WNOHANG);
         if (pid > 0)
         {
-            auto pair = gImpls.find(pid);
-            if (pair == gImpls.end())
+            std::lock_guard<std::recursive_mutex> guard(gManagersMutex);
+            for (ProcessManagerImpl* manager : gManagers)
```
MonsieurNicolas: Are you sure about this loop?
I would imagine that because each `ProcessManagerImpl` registered `mSigChild` they should have the callback triggered.
Or is asio kinda broken and follows a "last handler wins" type of contract?

MonsieurNicolas: You can look at the "quick restart" test https://github.com/stellar/stellar-core/blob/master/src/herder/HerderTests.cpp#L903
that creates a simulation object (creates a bunch of apps with their own timer). This is the correct way to spin up multiple applications.

In some other places, 2 apps are created with the same clock.
There is a potential problem on shutdown as `ApplicationImpl::shutdownMainIOService()` kills the IOService (shared by multiple apps).
https://github.com/stellar/stellar-core/blob/master/src/overlay/OverlayTests.cpp#L67

src/process/ProcessManagerImpl.cpp
```diff
@@ -266,7 +247,7 @@ ProcessExitEvent::Impl::run()
     {
         if (si.hStdOutput != NULL)
         {
-            CloseHandle(si.hStdOutput);
+            C loseHandle(si.hStdOutput);
```
MonsieurNicolas: oops :)

MonsieurNicolas:
I looked at `void signal_set_service::deliver_signal(int signal_number)` in `include/asio/detail/impl/signal_set_service.ipp` and it looks like asio delivers the signal per IOService - so you should not need to loop over `gManagers`.

src/process/ProcessManagerImpl.cpp
mfontanini: wat, I actually built this on windows and it worked. my linux laptop must have added a space in revenge for using another OS. I'll fix this later today and I also want to add that "multi process manager" test.

mfontanini:
Sure, I didn't have enough time to add the multi process manager test. I want to try that out and see how it behaves.

mfontanini:
So I had another look at this and the reason why the loop has to be there is because of `waitpid`, not because of the SIGCHLD signal. The signal is delivered to all process managers but that triggers them to call `waitpid` and that is basically global as it waits for any pid. That will return only once for every pid, hence the loop.

I can change this to loop through all pids each process manager controls and call `waitpid` on each of them. Since this uses `WNOHANG`, it should return immediately. This way the loop could be removed. I'm not sure how many processes can be concurrently running here, at some point that solution would become a bit inefficient. If you think that's better, I can change it.

MonsieurNicolas:
I thought about this, and I think the right change is to have an array in each process manager, and have each process manager only look for the ones it cares about like you describe here: main reason for me is that it seems like we're setting up a trap for someone later (at best deadlocks) by potentially calling callbacks from a different IOService. With this proposed change, we keep the call tree "clean" from any assumption on how asio works internally.

mfontanini:
Alright, I implemented these changes. Note that in order to keep the signature of `handleProcessTermination` the same (as I think it works better that way), I collect all (pid, status) tuples that terminated and then process them all at once out of the loop as handling the termination will erase from `mImpls`.

Also note that `asio::signal_set` won't queue multiple signals for the same type so you can't try to handle just one child per signal.

src/process/ProcessTests.cpp
```diff
@@ -160,35 +160,44 @@ TEST_CASE("subprocess storm", "[process]")
...
+                CLOG(DEBUG, "Process") << "error code: " << ec.message();
+            }
+            exitedCount++;
+            errorCodes.push_back(ec);
+        });
+    }
 
     // Shutdown so we force the command execution to fail
-    app->getProcessManager().shutdown();
+    app1->getProcessManager().shutdown();
```
MonsieurNicolas: I think there is a small race condition here:
I just tried this on my Windows dev box and calling shutdown before the process is fully started causes a dialog box to pop up on the screen (and stays there).
Error message is:
`The application was unable to start correctly (0xc0000142). Click OK to close the application`

Workaround is to sleep for a little bit before calling shutdown to leave some time for the OS to fully spawn the process.

mfontanini: Okay, will fix. I don't think this is an issue in Linux and maybe the laptop I have Windows on is slow enough that this is not an issue, I didn't notice this even though I ran it multiple times. I'll add a sleep later today!

MonsieurNicolas:
r+ 5c571fecf9e1c84050d8fa6a4d93ea7d7478ae2e

MonsieurNicolas:
@latobarita: retry

MonsieurNicolas:
@latobarita: retry

PR #1651, opened 2018-04-11
Added tx per ledger metric

marta-lokhova:


src/ledger/LedgerManagerImpl.cpp
```diff
@@ -113,6 +113,8 @@ LedgerManagerImpl::LedgerManagerImpl(Application& app)
     : mApp(app)
     , mTransactionApply(
           app.getMetrics().NewTimer({"ledger", "transaction", "apply"}))
+    , mTransactionCount(
+          app.getMetrics().NewHistogram({"ledger", "transaction", "count"}))
```
MonsieurNicolas: any reason you're using `Histogram` here instead of a `Meter` ?

marta-lokhova: Not sure how `Meter` is helpful in this case. All we get from it are various rates (e.g. 1min, 15 min, etc). Plus it does not have `Clear` method that we should use between runs. With `Histogram`, we have access to a distribution of txs/ledger, so we can see how much it varies during a test run.

MonsieurNicolas:
r+ 35a11a744c8d4b60ad23bb9a8a970d59fc24e426

PR #1658, opened 2018-04-13
Upgraded to Catch2 (v2.2.2)

jonjove:


src/test/TestPrinter.h
```diff
@@ -1,36 +0,0 @@
...
-
-// Copyright 2017 Stellar Development Foundation and contributors. Licensed
-// under the Apache License, Version 2.0. See the COPYING file at the root
-// of this distribution or at http://www.apache.org/licenses/LICENSE-2.0
-
-#include "catchup/CatchupWork.h"
-#include "history/HistoryTestsUtils.h"
-#include "lib/catch.hpp"
-#include "xdr/Stellar-transaction.h"
-#include "xdrpp/printer.h"
```
MonsieurNicolas: Oh so you're now relying on `operator<<` defined in `xdrpp/printer.h` ? There is a comment in there that says that we have to do `using xdr::operator<<;`, otherwise the specialization doesn't end up being properly defined.

Because of that, I *think* that you still need `TestPrinter.h` so that we always define that specialization.

MonsieurNicolas:
r+ 7e146b1a616dffd78d3568357a071d7fc1786bf8

MonsieurNicolas:
r+ f0c285cdec86c86ad8841b1114222a11f2cc3354

MonsieurNicolas:
@latobarita: retry

PR #1660, opened 2018-04-16
add a mechanism to only test a specific version

MonsieurNicolas:
usage: --test --version <N>
 where N is the version number

### DIFF
--- a/docs/software/commands.md
+++ b/docs/software/commands.md
@@ -58,7 +58,16 @@ forcescp doesn't change the requirements for quorum so although this node will e
   network to sign for.  For example, the production stellar network is
   "`Public Global Stellar Network ; September 2015`" while the test
   network is "`Test SDF Network ; September 2015`".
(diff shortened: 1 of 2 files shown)


src/main/main.cpp
```diff
@@ -184,6 +184,9 @@ usage(int err = 1)
           "                           (Default is STELLAR_NETWORK_ID "
           "environment variable)\n"
           "      --test               Run self-tests\n"
+          "           --all-versions  Run with all possible protocol versions\n"
```
jonjove: I don't think we need to have `--all-versions` or `--version` documented here. They will appear in `--test -?` now. The documentation added here should be specified where they are added to the CLI.

MonsieurNicolas: ah yes the new version of Clara makes this irrelevant

MonsieurNicolas:
@jonjove : I removed the changes that were in main.cpp

jonjove:
r+ e82b9d8

PR #1663, opened 2018-04-16
allow incremental upgrades, not just upgrade to the latest

### LINKED ISSUE #1662: allow nodes to vote for different protocol versions than the latest one
Especially useful on test networks, but potentially useful in public network as it allows to progressively upgrade to different versions of the protocol.

This allows for example to have a binary supporting version 11 of the protocol yet have 11 only deployed to testnet while pubnet is still on version 9 (and upgrade it from 9 to 10, not to 11).

MonsieurNicolas:
Resolves #1662


docs/software/commands.md
```diff
@@ -167,8 +167,8 @@ debugging purpose).
         transactions are ordered by transaction fee(lower fee transactions
          are held for later).<br>
     * protocolversion (uint32) defines the protocol version to upgrade to.
-         When specified it must match the protocol version supported by the
-        node<br>
+         When specified it must match one of the protocol versions supported
+         by the node and should be greater than the current ledger<br>
```
vogel: I think it is a typo. `It should be greater that version of the current ledger.` seems better

src/herder/Upgrades.cpp
```diff
@@ -238,7 +238,7 @@ Upgrades::removeUpgrades(std::vector<UpgradeType>::const_iterator beginUpdates,
 bool
 Upgrades::isValid(uint64_t closeTime, UpgradeType const& upgrade,
                   LedgerUpgradeType& upgradeType, bool nomination,
-                  Config const& cfg) const
+                  Config const& cfg, LedgerHeader const& header) const
```
vogel: what about passing just ledgerVersion here? we are not using full LedgerHeader struct here

src/herder/Upgrades.cpp
```diff
@@ -267,9 +267,10 @@ Upgrades::isValid(uint64_t closeTime, UpgradeType const& upgrade,
             res = mParams.mProtocolVersion &&
                   (newVersion == *mParams.mProtocolVersion);
         }
-        // only upgrade to the latest supported version of the protocol
-        // is allowed
-        res = res && (newVersion == cfg.LEDGER_PROTOCOL_VERSION);
+        // only allow upgrades to a supported version of the protocol
+        res = res && (newVersion <= cfg.LEDGER_PROTOCOL_VERSION);
+        // and enforce versions to be strictly monotonic
+        res = res && (newVersion > header.ledgerVersion);
```
vogel: I like this very much!

src/main/CommandHandler.cpp
```diff
@@ -315,8 +315,8 @@ CommandHandler::fileNotFound(std::string const& params, std::string& retStr)
         "transactions are ordered by transaction fee(lower fee transactions"
         " are held for later).<br>"
         "protocolversion (uint32) defines the protocol version to upgrade to."
-        " When specified it must match the protocol version supported by the"
-        " node<br>"
+        " When specified it must match one of the protocol versions supported"
+        " by the node and should be greater than the current ledger <br>"
```
vogel: the same grammar issue as before

src/herder/Upgrades.cpp
MonsieurNicolas: Conceptually it should take a `ledgerHeader` as a parameter as upgrades perform arbitrary changes based on the ledger header (I can see having conditions only enforced after a certain version number, or re-introduce some of the max-rate change to fee/base reserve logic we removed as part of the transition but that can be re-added now that we formalized upgrades), I guess the bug in the interface is that it takes "closeTime" (and we'll be chasing all fields from the ledger header over time), let me fix this.

MonsieurNicolas: This also brings back the symmetry between `createUpgradesFor` and `isValid` (where the outside world should not have to know about the internals of what upgrades do)

MonsieurNicolas:
@vogel I updated per your comment

docs/software/commands.md
vogel: I mean that the wording suggest that protocol version should be greater than the `current ledger number`, not `current ledger version`.

MonsieurNicolas: ah yes, makes sense. fixed!

MonsieurNicolas:
@vogel updated requirement with protocol version number

vogel:
r+ 12f5c32b8184b7d66b6ac5e14d348822c777ee19

MonsieurNicolas:
@latobarita: retry

PR #1664, opened 2018-04-16
Load best offer context

jonjove:
Extracted from #1600. Separates the implementation of batch loading best offers from the process of iterating over them.

### DIFF
--- a/src/transactions/OfferTests.cpp
+++ b/src/transactions/OfferTests.cpp
@@ -1083,4 +1083,23 @@ TEST_CASE("create offer", "[tx][offers]")
             });
         }
     }
+
+    SECTION("updated offers respect reserve")
+    {
+        auto market = TestMarket{*app};
+        auto a1 = root.create("A", app->getLedgerManager().getMinBalance(2) + 4*txfee + 10);
+        a1.changeTrust(usd, trustLineLimit);
+        for_all_versions(*app, [&] {
+            auto offer = market.requireChangesWithOffer(
+                {},
+                [&] {
+                    return market.addOffer(a1, {xlm, usd, oneone, 9});
+                });
+            market.requireChangesWithOffer(
+                {},
+                [&] {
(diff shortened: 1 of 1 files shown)


src/transactions/ManageOfferOpFrame.cpp
```diff
@@ -164,6 +167,8 @@ ManageOfferOpFrame::doApply(Application& app, LedgerDelta& delta,
             return false;
         }
 
+        mSellSheepOffer->storeDelete(tempDelta, db);
```
MonsieurNicolas: I would prefer keeping calls to `storeDelete` with their `addNumEntries` counter part so that the account state is valid during execution

src/transactions/ManageOfferOpFrame.cpp
```diff
@@ -168,6 +168,9 @@ ManageOfferOpFrame::doApply(Application& app, LedgerDelta& delta,
         }
 
         mSellSheepOffer->storeDelete(tempDelta, db);
+        bool res = mSourceAccount->addNumEntries(-1, ledgerManager);
+        assert(res);
+        mSourceAccount->storeChange(tempDelta, db);
```
MonsieurNicolas: `maxAmountOfSheepCanSell` is not computed properly anymore as code would end up in the `else` branch https://github.com/jonjove/stellar-core/blob/cf0063534781b5ecdb829fb3243f7b18ae8b1134/src/transactions/ManageOfferOpFrame.cpp#L227

I am surprised that tests didn't catch this.

jonjove: Added a test to catch this.

MonsieurNicolas:
turns out that touching the account to update the number of subentries breaks things even though it's a no-op: the last modified field in the account entry ends up being the only field changed, but it is enough to break the hash of the ledger as the bucket list changes.

I suggest we go back to your original proposal for now, with some comments that warn about this for future us.

PR #1665, opened 2018-04-16
Parallelize tests

### LINKED ISSUE #1645: Split up test runs in travis
As we are running more tests, we need to figure out how to parallelize their execution.

The first step in this parallelization effort is to enable parallelization in separate environments (ie, when running in Travis for example).

This boils down to figuring out how to partition the test runs over N partitions (with each partition K with 1 <= K <= N):
* an approach could be to generate a list of test cases that needs to be run, then break down this list in n partitions. Some variation of:
   * `./stellar-core --test --list-test-names-only > all-tests.txt`
   * "some clever command that picks test number K+i*N for each K" > all-tests-K.txt
   * run `stellar-core --test -f all-tests-K.txt`
* an equivalent but maybe simpler version of this would be to make a slight modification to catch (maybe there is a way to do this already?) to support predicates that "replaces" some tests by a no-op, so running tests will look like `stellar-core --test --partition-K` (where `K` is the partition number - note the hack here of using `--partition-K` instead of `--partition K` as catch doesn't allow for parameters on custom command line)

With this done, as we have currently 2 jobs in Travis, we can partition the tests into 2:
* clang build -> run tests partition "1"
* gcc build -> run tests partition "2"

### LINKED ISSUE #1646: change stellar-core to allow for running tests in parallel on the same machine
When #1645 is done we can modify the test infrastructure in stellar-core to allow for parallel runs in the same environment (ie on the same machine).

We should be able to run a bunch of `stellar-core --test --partition-K` processes at the same time.

Example of changes that need to happen (with `K` being the partition number being invoked):
* change all test paths to use `K` somehow (I think we already name the main log file `stellar0.log`), the others are (should) be based on tmdir which is already random
* change database names to be offset by `(K-1)*NB_DATABASES` where `NB_DATABASES` is the number of databases that tests use (right now, I think it's 11 ie we use `test0... test10`) - see how `getTestConfig` generates database names

jonjove:
Resolves #1645. Resolves #1646.


src/test/selftest-parallel
```diff
@@ -0,0 +1,44 @@
...
+fi
+
+cleanup() {
+    rm -rf $TEST_PARTITIONS_DIR
+}
+TEST_PARTITIONS_DIR=$(mktemp -d ./test-partitions.XXXXXXXX)
+trap cleanup 0 2 15
+
+cd $TEST_PARTITIONS_DIR
+perl -e 'my @all_tests = `../stellar-core --ll FATAL --test --list-test-names-only '"$TEST_SPEC"'`;
```
MonsieurNicolas: I think it would be cleaner here to
* not `cd` into `TEST_PARTITION_DIR` and invoke stellar core with `./stellar-core` (not assuming that stellar-core can be run from a different folder - as it may load dependencies)
* pass in arguments to the perl script (such as the value of `TEST_PARTITIONS_DIR`, `NUM_PARTITIONS`)
   * this is done like this:
```sh
$ perl -e '$a=$ARGV[0]; $b=$ARGV[1]; print "a is $a; b is $b\n"' -- "A" "B's value"
a is A; b is B's value
```

src/test/selftest-parallel
```diff
@@ -0,0 +1,44 @@
...
+cd ..
+
+declare -a TEST_JOBS
+for i in $RUN_PARTITIONS; do
+    "$1" $((50*i)) $TEST_PARTITIONS_DIR/test-partition-$i.txt &
+    TEST_JOBS[i]=$!
+done
+
+EXIT_CODE=0
+for i in $RUN_PARTITIONS; do
```
MonsieurNicolas: `for p in "${TEST_JOBS[@]}" ; do` loops over the array

src/test/selftest-parallel
```diff
@@ -0,0 +1,44 @@
...
+    wait ${TEST_JOBS[i]}
+    if test "$?" != 0; then
```
MonsieurNicolas: this should be `if [[ $? -ne 0 ]] ; then` (`!=` is technically used to compare strings)

src/test/selftest-parallel
```diff
@@ -0,0 +1,44 @@
+#!/bin/sh
+
```
MonsieurNicolas: we may want to add
`set -e` here so that the script fails if any command fails

src/test/selftest-parallel
```diff
@@ -0,0 +1,44 @@
...
+             open(my $out, ">", "test-partition-$p.txt");
+             for (my $i = $p; $i < @all_tests; $i+='"$NUM_PARTITIONS"') {
+                 print $out $all_tests[$i];
+             }
+             close $out;
+         }'
+cd ..
+
+declare -a TEST_JOBS
+for i in $RUN_PARTITIONS; do
```
MonsieurNicolas: I suspect we need to update the trap handler to also kill processes in `TEST_JOBS` in case the script is interrupted - maybe just kill the group with something like:
```
kill -INT -$$
sleep 2
kill -KILL -$$
```

src/test/test.cpp
```diff
@@ -168,14 +170,17 @@ test(int argc, char* const* argv, el::Level ll,
...
-    session.cli(
-        session.cli() |
-        clara::Opt(gTestAllVersions)["--all-versions"]("Test all versions") |
-        clara::Opt(gVersionsToTest,
-                   "version")["--version"]("Test specific version(s)"));
+    auto cli = session.cli();
+    cli |= clara::Opt(gTestAllVersions)["--all-versions"]("Test all versions");
+    cli |= clara::Opt(gVersionsToTest,
+                      "version")["--version"]("Test specific version(s)");
+    cli |= clara::Opt([](int x) { gBaseInstance = x; },
```
MonsieurNicolas: no need for lambda, it should be
`clara::Opt(gBaseInstance, "base-instance")["--base-instance"] ...`

travis-build.sh
```diff
@@ -33,11 +33,18 @@ g++ -v
 llvm-symbolizer --version || true
 
 # Create postgres databases
+if test $CXX = 'clang++'; then
```
MonsieurNicolas: are we sure that those are the strings?
Maybe use something like this instead?
```
echo "$CXX" | grep -q 'clang++'
if [[ $? -eq 0 ]] ; then
```

jonjove: According to the Travis documentation:

> Testing against two compilers will create (at least) 2 rows in your build matrix. For each row, the Travis CI C++ builder will export the `CXX` env variable to point to either `g++` or `clang++` and `CC` to either `gcc` or `clang`.

so I expect this should work correctly as is.

src/test/selftest-parallel
MonsieurNicolas: I was thinking to use a handler for killing children, but not needed at this point I agree.

MonsieurNicolas:
r+ 71200d8f9b8b92918b8fdfe3e784347774331c37

MonsieurNicolas:
Need to run clang-format

MonsieurNicolas:
r+ dc97c31e

Issue #1635, opened 2018-03-28
descriptors do not have `CLOEXEC` set (posix)


MonsieurNicolas:
It looks like descriptors do not have the `FD_CLOEXEC` option set, which means that descriptors are inherited by child processes.

I found that for example, that running in a loop:
```
./stellar-core --ll ERROR --test "simulate upgrades"
```

yields random failures:
```
simulate upgrades
  2 of 3 vote early (v-blocking) - 3 upgrade anyways
-------------------------------------------------------------------------------
herder/UpgradesTests.cpp:619
...............................................................................

herder/UpgradesTests.cpp:619: FAILED:
due to unexpected exception with message:
  bind: Address already in use
```

I found that this particular one is caused by the http port not getting released somehow.

As a test, adding this in `lib/http/server.cpp` (after bind):
```c++
    int fd = acceptor_.native_handle();
    int flags;
    if ((flags = fcntl(fd, F_GETFD)) != -1)
        fcntl(fd, F_SETFD, flags | FD_CLOEXEC);
```
makes the test reliable.

I checked (Linux) the descriptors open in a child process, and could see something like this (forcing history to call a fake `curl` that does `sleep 1h`):

```
$ grep -r flags /proc/10006/fdinfo/
/proc/10006/fdinfo/0:flags:     0102002
/proc/10006/fdinfo/1:flags:     0102002
/proc/10006/fdinfo/2:flags:     0102002
/proc/10006/fdinfo/3:flags:     0102001
/proc/10006/fdinfo/4:flags:     02004002
/proc/10006/fdinfo/5:flags:     02000002
/proc/10006/fdinfo/6:flags:     02000002
/proc/10006/fdinfo/6:settime flags: 00
/proc/10006/fdinfo/7:flags:     02004000
/proc/10006/fdinfo/8:flags:     02004001
/proc/10006/fdinfo/9:flags:     02100002
/proc/10006/fdinfo/10:flags:    02100002
/proc/10006/fdinfo/11:flags:    02100002
/proc/10006/fdinfo/12:flags:    02004002
/proc/10006/fdinfo/13:flags:    0100002
/proc/10006/fdinfo/14:flags:    04002
/proc/10006/fdinfo/15:flags:    02004002
/proc/10006/fdinfo/16:flags:    02000002
/proc/10006/fdinfo/17:flags:    02000002
/proc/10006/fdinfo/17:settime flags: 00
/proc/10006/fdinfo/18:flags:    04002
[...]

PR #1666, opened 2018-04-17
Updates to how nodes decide to trigger the next ledger

### LINKED ISSUE #1489: flaky test: "SCP messages flooding"
The test
```
Flooding
  scp messages flooding
  outer nodes
  tcp
```

seems to timeout randomly in CI (I also observed that on Windows)

### LINKED ISSUE #1652: Nomination time has large standard deviation
While performing various load testing, we noticed that the time between nomination starting and the ballot protocol starting varies between runs and even between nodes even though there is little variance on the consensus time.

It seems that it's caused by each node having their own opinion (offset) on when rounds should start.

MonsieurNicolas:
resolve #1652 , resolve #1489

This PR has a couple changes that should help with nomination timing:
* leader weights were not always computed right during nomination, leading to some rounds with no leader
* we now align to when the ballot protocol starts for deciding when to trigger
   * while this may introduce a small delay between rounds, this implementation should in most cases allow nodes to automatically synchronize to each other (assuming a uniform time for messages to make it to various nodes)

### DIFF
--- a/src/crypto/SecretKey.cpp
+++ b/src/crypto/SecretKey.cpp
@@ -68,19 +68,10 @@ SecretKey::Seed::~Seed()
     std::memset(mSeed.data(), 0, mSeed.size());
 }
 
-PublicKey
+PublicKey const&
 SecretKey::getPublicKey() const
 {
-    PublicKey pk;
-
-    assert(mKeyType == PUBLIC_KEY_TYPE_ED25519);
-
-    if (crypto_sign_ed25519_sk_to_pk(pk.ed25519().data(), mSecretKey.data()) !=
-        0)
-    {
-        throw std::runtime_error("error extracting public key from secret key");
-    }
-    return pk;
+    return mPublicKey;
 }
 
 SecretKey::Seed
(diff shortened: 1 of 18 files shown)


jedmccaleb:
r+ 4243d5f

MonsieurNicolas:
Some tests were failing with those changes.

Turns out that the changes increased the chances of the "SCP flooding tests" to fail more often, which allowed me to identify some problems and fixes.

Updated PR includes in addition to the nomination changes:
* ability to change the quorum set when running simulations
* SCP flooding test fix
   * there was a race condition that would cause the whitelist of nodes on the network to kick in while trying to inject SCP messages from other nodes
   * fix was to include those other nodes in the quorum (and make sure that they won't cause externalize to be called)
* fixed a potential performance issue when receiving quorum sets (was causing tests to be much slower than they should be, but the change also makes sense in live environment)

vogel:
LGTM. Should we merge that now?

MonsieurNicolas:
@vogel yes, it can be branch is up to date

MonsieurNicolas:
@latobarita: retry

MonsieurNicolas:
kicking again, tests pass on a couple of dev boxes (except with the occasional bind issue caused by #1635 ), so I am not sure why they fail in travis yet.

PR #1670, opened 2018-04-21
Improve item fetcher

vogel:


vogel:
Main reason for this PR is to simplify item fetching. Old version used strange combination of PendingEnvelopes/ItemFetcher/Tracker classes that did a lot of things and had weird dependencies on each other.

This PR changes ItemFetcher to handle all of fetching, not only for one item. It uses queue to track what is needed and when fetching should be performed (after starting of item fetching a new entry is added to queue to support timeouts). This will allow to better separation of Overlay into Transport and Overlay and it will allow more callback-based communication between Transport, Overlay and Herder.

Also a lot of class extraction was done. Tracker was additionally doing what now ItemPeerQueue, EnvelopeItem does and some more of small classes do. Also PendingEnvelopes was split into Pending/Fetching nad ReadyEnvelopes. This allows for more fine-grained tests (which were also added).

And one of smallest, but for me most important, changes - PendingEnvelopes is no longer calling recvEnvelope on Herder. Now Herder is getting list of now-ready envelopes from PendingEnvelopes::handleQSet/TxSet and it can call recvEnvelope itself. It removes some cycles and recursion from code.

src/herder/EnvelopeItemMap.h
```diff
@@ -0,0 +1,63 @@
...
+    std::set<ItemKey> removeIf(std::function<bool(SCPEnvelope const&)> cond);
+
+    /**
+     * Return all envelopes that are mapped to itemKey.
+     */
+    std::set<SCPEnvelope> envelopes(ItemKey itemKey);
+
+  private:
+    std::map<SCPEnvelope, std::set<ItemKey>> mEnvelopeItems;
+    std::map<ItemKey, std::set<SCPEnvelope>> mItemEnvelopes;
```
marta-lokhova: This is kind of an open-ended question: it looks like this solution does a double-store, i.e. same envelope-itemkey info is stored twice (in addition to all envelopes also being stored in FetchingEnvelopes). I'm wondering if it's worth looking into making it more efficient, e.g. using pointers?

src/overlay/TxSetCache.h
```diff
@@ -0,0 +1,41 @@
...
+
+#include <lib/util/lrucache.hpp>
+
+namespace stellar
+{
+
+using TxSetFramePtr = std::shared_ptr<TxSetFrame>;
+
+// all the txsets we have learned about per ledger#
+class TxSetCache
```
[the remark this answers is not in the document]
vogel: copy-paste bug, I'll rename it when rebasing

src/herder/EnvelopeItemMap.cpp
```diff
@@ -0,0 +1,165 @@
...
+
+std::set<ItemKey>
+EnvelopeItemMap::removeIf(std::function<bool(SCPEnvelope const&)> cond)
+{
+    auto result = std::set<ItemKey>{};
+    for (auto it = std::begin(mEnvelopeItems); it != std::end(mEnvelopeItems);)
+    {
+        if (cond(it->first))
+        {
+            it = mEnvelopeItems.erase(it);
```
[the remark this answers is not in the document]
vogel: I already have iterator for this envelope, and using `EnvelopeItemMap::remove(SCPEnvelope const& envelope) ` would mean looking for it again, which would be inefficient.

It would also mean that each item for each of envelopes that passes `cond` is looked for multiple times, so a lot of unneccessary std::map::find calls.

src/herder/EnvelopeItemMap.h
vogel: It would require to have a global map that maps SCPEnvelope to pointer to SCPEnvelope (probably shared pointer) - so each time we receive given envelope, we can match it with its pointer. I think it may be worth consideration, but probably we should first gather data about how many envelopes are we keeping in memory and how many unique envelopes are among them.

This would also help in deciding if my solution for `removeIf` is necessary or if simpler one will suffice (i.e., we keep too small amount of envelopes in memory to even bother optimizing the code).

Builds/VisualStudio/stellar-core.vcxproj
```diff
@@ -278,14 +278,22 @@ exit /b 0
     <ClCompile Include="..\..\src\database\DatabaseTests.cpp" />
     <ClCompile Include="..\..\src\database\DatabaseUtils.cpp" />
     <ClCompile Include="..\..\src\herder\Herder.cpp" />
+    <ClCompile Include="..\..\src\herder\EnvelopeItemMap.cpp" />
```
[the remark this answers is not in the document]
MonsieurNicolas: in particular, there are simple (non functional) refactors in there, mixed with actual changes

src/herder/FetchingEnvelopes.h
```diff
@@ -0,0 +1,98 @@
...
+
+namespace stellar
+{
+
+/**
+ * Container for envelopes that are currently being fetched. Also keeps track of
+ * envelopes that are discarded and should be ignored. Only allows envelopes
+ * above certaing slot index that can be set by setMinimumSlotIndex.
+ */
+class FetchingEnvelopes
```
MonsieurNicolas: I don't think `FetchingEnvelopes` is useful as an abstraction: it's 90% of what `PendingEnvelopes` is

src/herder/FetchingEnvelopes.cpp
```diff
@@ -0,0 +1,237 @@
...
+    {
+        startFetching(peer, envelope, items);
+        return false;
+    }
+}
+
+std::set<SCPEnvelope>
+FetchingEnvelopes::handleQuorumSet(const SCPQuorumSet& q, bool force)
+{
+    auto itemKey = ItemKey{ItemType::QUORUM_SET, sha256(xdr::xdr_to_opaque(q))};
```
MonsieurNicolas: Per my other comment, the fact that `hash(x)` is used as a key is an implementation detail of overlay; there is no need to have this done in herder. NB: it doesn't mean that you can't have methods that lookup by hash exposed by overlay

src/herder/FetchingEnvelopes.cpp
```diff
@@ -0,0 +1,237 @@
...
+    }
+
+    if (!isQuorumSetSane(q, false))
+    {
+        discardEnvelopesWithItem(itemKey);
+        return std::set<SCPEnvelope>{};
+    }
+
+    mNodesInQuorum.clear();
+    mApp.getOverlayManager().getQSetCache().add(itemKey.getHash(), q);
```
MonsieurNicolas: I think this PR would look a lot better if we were not managing the cache directly from herder: this looks to me like something we can delegate to the item fetcher code

src/herder/FetchingEnvelopes.h
```diff
@@ -0,0 +1,98 @@
+#pragma once
+
+// Copyright 2018 Stellar Development Foundation and contributors. Licensed
+// under the Apache License, Version 2.0. See the COPYING file at the root
+// of this distribution or at http://www.apache.org/licenses/LICENSE-2.0
+
+#include "herder/EnvelopeItemMap.h"
```
MonsieurNicolas: It would be a bit less confusing if you were following a PImpl pattern as part of those refactors - this would allow to include implementation classes only from the cpp file (ie: nobody outside should care about EnvelopeItemMap)

src/herder/HerderImpl.cpp
```diff
@@ -73,19 +74,19 @@ HerderImpl::SCPMetrics::SCPMetrics(Application& app)
 
 HerderImpl::HerderImpl(Application& app)
     : mPendingTransactions(4)
-    , mPendingEnvelopes(app, *this)
-    , mHerderSCPDriver(app, *this, mUpgrades, mPendingEnvelopes)
+    , mHerderSCPDriver(app, *this, mUpgrades,
+                       [this](TxSetFramePtr txSet) {
```
MonsieurNicolas: do we need to use lambdas like this in the new implementation?

src/herder/Herder.h
```diff
@@ -104,7 +104,7 @@ class Herder
     virtual void restoreState() = 0;
 
     virtual bool recvSCPQuorumSet(SCPQuorumSet const& qset) = 0;
-    virtual bool recvTxSet(Hash const& hash, TxSetFrame const& txset) = 0;
+    virtual bool recvTxSet(TxSetFramePtr txset) = 0;
```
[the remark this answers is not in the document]
vogel: Currently this unique ID can be created from txset/quorumset, so passing this along that value is not necessary and can be source of error if wrong value is passed. Moving that down the call stack reduces chances of mistake.

Herder still can call overlay with didntNeedThis(makeKey(txset)) which is now possible thanks to new keyFor method in ItemFetcher

vogel:
Updated and rebased.

src/overlay/PendingEnvelopes.cpp
```diff
@@ -144,7 +142,7 @@ PendingEnvelopes::isNodeInQuorum(NodeID const& node)
     else
     {
         // search through the known slots
-        SCP::TriBool r = mHerder.getSCP().isNodeInQuorum(node);
+        SCP::TriBool r = mApp.getHerder().getSCP().isNodeInQuorum(node);
```
MonsieurNicolas: don't move quorum related functions to overlay: we should keep overlay agnostic to SCP as much as possible

src/overlay/NodesInQuorum.cpp
```diff
@@ -0,0 +1,50 @@
+// Copyright 2018 Stellar Development Foundation and contributors. Licensed
+// under the Apache License, Version 2.0. See the COPYING file at the root
+// of this distribution or at http://www.apache.org/licenses/LICENSE-2.0
+
+#include "overlay/NodesInQuorum.h"
```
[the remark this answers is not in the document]
MonsieurNicolas: also, I find the value of this new class questionable as it's just a method on top of lru

src/overlay/EnvelopeHandler.h
```diff
@@ -0,0 +1,71 @@
...
+#include "overlay/Peer.h"
+#include "transactions/TransactionFrame.h"
+#include "xdr/Stellar-overlay.h"
+
+namespace stellar
+{
+
+struct ItemKey;
+
+class EnvelopeHandler
```
MonsieurNicolas: Some of the changes look good. I think that there are some fundamental issues in this PR though:
* you kept `PendingEnvelope` and `ReadyEnvelope` as one, but this causes all sorts of problems: dependency loops, questionable data ownership.
   * I think it would be much cleaner to move the "pending" part into overlay (like you did) but keep the "ready" part in herder.
   * Concern of overlay will be to feed (and maybe cache in the future) full envelopes (with QSet and TxSet) to herder
   * that way we don't need new layers of abstraction to just do "processSCPQueue"
* we should start thinking about getting rid of some of the broken semantics of current's implementation (that you kept from the item fetcher design)
   * right now we have recursive calls that happen when for example we close a ledger: we craft a quorum set from the local state and by feeding it into FutureEnvelope we turn an envelope into "ready state". It makes the code hard to follow, ideally we would just have overlay (later) call "ReceiveEnvelope".

MonsieurNicolas: `EnvelopeHandler` seems to not be the right interface exposed by overlay: by taking `Peers` as parameters it keeps requiring its consumers (Herder in this case) to keep track of peers - and this doesn't seem to be the right direction

MonsieurNicolas: this may be caused by the point I raised above: we probably want to split PendingEnvelope and ReadyEnvelope, which should reduce the dependencies on overlay constructs

src/history/HistoryTests.cpp
```diff
@@ -424,6 +424,7 @@ TEST_CASE("persist publish queue", "[history]")
         // checkpoint still queued.
         CHECK(hm0.getPublishSuccessCount() == 0);
         CHECK(hm0.getMinLedgerQueuedToPublish() == 7);
+        clock.cancelAllEvents();
```
MonsieurNicolas: too many changes in the same commit - break it up

src/overlay/ItemPeerQueue.cpp
```diff
@@ -0,0 +1,60 @@
...
+// under the Apache License, Version 2.0. See the COPYING file at the root
+// of this distribution or at http://www.apache.org/licenses/LICENSE-2.0
+
+#include "overlay/ItemPeerQueue.h"
+
+namespace stellar
+{
+
+void
+ItemPeerQueue::addKnowing(Peer::pointer peer)
```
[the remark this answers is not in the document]
vogel: I think this class makes separate of concepts better. If we change strategy on how we query peer, we will have to change or remove this class, but merging it back into ItemFetchQueue does not seem as a good idea, as it would put again too much responsibility into one class.

src/test/TestUtils.cpp
```diff
@@ -2,7 +2,9 @@
 // under the Apache License, Version 2.0. See the COPYING file at the root
 // of this distribution or at http://www.apache.org/licenses/LICENSE-2.0
 
-#include "TestUtils.h"
+#include "test/TestUtils.h"
+#include "crypto/Hex.h"
```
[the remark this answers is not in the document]
vogel: clang-format properly handles this case without newline

src/overlay/EnvelopeHandler.h
vogel: EnvelopeHandler is supposed to be interface exposed by Transport layer and used by Overlay - Overlay knows about Peers, so it is OK. Herder should not know about EnvelopeHandler at all.

vogel:
Updated with asked changes

Note .1: I've not rebased this, so too large commits are still too large and not split, also there exists commits that remove changes from previous ones. Also I've kept some bad names (as push and seen in Herder.h), as I'll improve them at the time of rebase.

Note .2: It looks like removing recursion in different way that ioService.post may be very hard/ugly to do. Removing it from combineCandidates would require lots of changes in SCP layer, as it is called by processEnvelope which is recursive itself

Note .3: Moving NodesInQuorum back to Herder makes Overlay dependent on Herder again (as this must be checked before passing envelope from Peer to PendingEnvelopes, which is on Overlay layer).

Node .4 Moving ReadyEnvelopes to Herder added another dependency from Overlay, do you have an idea on how we could remove that?

src/overlay/PendingEnvelopes.cpp
```diff
@@ -0,0 +1,213 @@
...
+
+    auto& discardedSet =
+        mEnvelopes[envelope.statement.slotIndex].mDiscardedEnvelopes;
+    auto discarded =
+        std::find(std::begin(discardedSet), std::end(discardedSet), envelope);
+    return discarded != std::end(discardedSet);
+}
+
+void
+PendingEnvelopes::setMinimumSlotIndex(uint64_t slotIndex)
```
MonsieurNicolas: we should not be tracking the "minimum slot index" here: this is the responsibility of `OverlayEnvelopeHandler`

src/overlay/PendingEnvelopes.cpp
```diff
@@ -0,0 +1,213 @@
...
+PendingEnvelopes::~PendingEnvelopes()
+{
+}
+
+EnvelopeHandler::EnvelopeStatus
+PendingEnvelopes::handleEnvelope(Peer::pointer peer,
+                                 SCPEnvelope const& envelope)
+{
+    auto const& nodeID = envelope.statement.nodeID;
+    if (envelope.statement.slotIndex < mMinimumSlotIndex)
```
MonsieurNicolas: this filtering is redundant with `OverlayEnvelopeHandler::processEnvelope`

src/overlay/OverlayEnvelopeHandler.h
```diff
@@ -0,0 +1,46 @@
...
+  public:
+    explicit OverlayEnvelopeHandler(Application& app);
+    ~OverlayEnvelopeHandler() = default;
+
+    void setValidRange(uint32_t min, uint32_t max) override;
+    EnvelopeStatus envelope(Peer::pointer,
+                            SCPEnvelope const& envelope) override;
+
+    void getQuorumSet(Peer::pointer peer, Hash const& hash) override;
+    std::set<SCPEnvelope> quorumSet(Peer::pointer peer,
```
MonsieurNicolas: same thing wrt naming, `quorumSet` and `txSet` are not good names; maybe prefix with `recv`

src/overlay/OverlayEnvelopeHandler.cpp
```diff
@@ -0,0 +1,201 @@
...
+            << " from: "
+            << mApp.getConfig().toShortString(envelope.statement.nodeID)
+            << " s:" << envelope.statement.pledges.type()
+            << " i:" << envelope.statement.slotIndex
+            << " a:" << mApp.getStateHuman();
+
+    if (envelope.statement.nodeID == mLocalNodeID)
+    {
+        CLOG(DEBUG, "Overlay") << "recvSCPEnvelope: skipping own message";
+        return EnvelopeHandler::ENVELOPE_STATUS_DISCARDED;
```
MonsieurNicolas: maybe we should add a counter that gets incremented every time we drop messages

src/herder/HerderImpl.cpp
```diff
@@ -861,7 +806,8 @@ HerderImpl::getJsonInfo(size_t limit)
         mApp.getConfig().toStrKey(mApp.getConfig().NODE_SEED.getPublicKey());
 
     ret["scp"] = getSCP().getJsonInfo(limit);
-    ret["queue"] = mPendingEnvelopes.getJsonInfo(limit);
+    ret["queue"] = mApp.getOverlayManager().getJsonInfo(limit);
```
MonsieurNicolas: let's not use the same `queue` element as it's not managed by herder anymore.
this also means that we should not invoke overlay status from `herder` but instead have `CommandHandler::scpInfo` call the relevant overlay method

src/herder/ReadyEnvelopes.cpp
```diff
@@ -0,0 +1,141 @@
...
+        }
+        else
+            break;
+    }
+}
+
+void
+ReadyEnvelopes::dumpInfo(Json::Value& ret, size_t limit)
+{
+    auto& q = ret["queue"];
```
MonsieurNicolas: I don't think you want to reference `queue` here

src/herder/ReadyEnvelopes.cpp
```diff
@@ -0,0 +1,141 @@
...
+            result.push_back(entry.first);
+        }
+    }
+    return result;
+}
+
+void
+ReadyEnvelopes::setMinimumSlotIndex(uint64_t slotIndex)
+{
+    mMinimumSlotIndex = slotIndex;
```
MonsieurNicolas: not a big fan of `mMinimumSlotIndex` here either: the filtering should happen at herder (that knows which slot it's tracking). It's actually quite error prone to have to keep those mininum/maximum slots in sync.
For example, I think on startup you are not setting `mMinimumSlotIndex` so it stays as 0 even though that's not correct.

src/overlay/OverlayEnvelopeHandler.cpp
```diff
@@ -0,0 +1,201 @@
...
+std::set<SCPEnvelope>
+OverlayEnvelopeHandler::quorumSet(Peer::pointer peer, SCPQuorumSet const& qSet,
+                                  bool force)
+{
+    auto result =
+        mApp.getOverlayManager().getPendingEnvelopes().handleQuorumSet(qSet,
+                                                                       force);
+    if (result.first)
+    {
+        mApp.getHerder().clearNodesInQuorumCache();
```
MonsieurNicolas: not sure I like that we're clearing the cache here - maybe we don't have the choice yet (until we properly manage LRU data in overlay vs "real" data in herder)

src/test/TestUtils.cpp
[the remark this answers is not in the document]
MonsieurNicolas: ah yes forgot about this

src/overlay/OverlayEnvelopeHandler.cpp
vogel: Looks like this is not needed. I've removed that from this branch and it passes all the tests.
I've also removed corresponding line from master and it also passes all the tests.

vogel:
Another batch of changes. I think that now everything (except visual studio changes) is implemented.

This PR requires heavy rebasing and squashing (and adding visual studio changes).

MonsieurNicolas:
sweeping stale pull requests

PR #1672, opened 2018-04-26
Quorum filtering fix

MonsieurNicolas:
One of the changes introduced by PR #1666 broke the quorum filtering code

This PR:
* reverts the broken change (sometimes I wonder what I am doing to write something so broken)
* fixes the real problem that existed with filtering
* adds a proper test for SCP filtering (should have been done it in the first place)


vogel:
r+ 12adce2046e568d690f3e0f03da5311ef95abc8f

vogel:
@latobarita: retry

PR #1673, opened 2018-04-26
fix fallout from PR1666 part two

MonsieurNicolas:
Looks like PR #1666 had another broken commit: the change ce3f03f2512e6d64174a26e5b1f501cd632aa44f introduced a bug where the "public key" field was not properly set when creating a `SecretKey` from an `StrKey`. Turns out there were not tests for that, so I included one (tests rely on binary seeds instead).

I also found that logging would silently fail to log to files in some cases, so I did a pass at all Easylogging settings and disabled/enabled the ones that seem relevant.


jedmccaleb:
r+ d08f428

PR #1677, opened 2018-05-01
Adding error message when FAILURE_SAFETY = -1

alfiedotwtf:


MonsieurNicolas:
Thanks for this submission. Can you describe the before/after?

It's not clear to me why the message from https://github.com/stellar/stellar-core/pull/1677/files#diff-720cc1cfd6b4747004f45648fd62035dR602 is not enough?

alfiedotwtf:
I didn't have FAILURE_SAFETY set in my config. But when run, I got:

    "Can't have FAILURE_SAFETY=0 unless you also set"

I couldn't understand why it was telling me I had FAILURE_SAFETY set to 0 because when it's NOT set, it should default to -1. So I assumed a bug. Digging into the code, it's clear what happened.

So this change is to fix the error message for when FAILURE_SAFETY is set to -1 so that the next person doesn't get confused as I did :)

MonsieurNicolas:
Oh I see.

Instead of duplicating the code that checks, maybe it would be better to log that we're assigning a value to `FAILURE_SAFETY` (and which value)? That way there is also a log line that gives the resulting number, which can be useful to diagnose the other error condition "Not enough nodes / thresholds too strict...".

src/test/run-selftest-pg
```diff
@@ -30,6 +29,18 @@ PSQL=psql
 findpg pg_ctl
 PGCTL=pg_ctl
 
+if [ "$(which $PGCTL)" == "" ]; then
+    findpg pg_ctlcluster
+    PGCTL=pg_ctlcluster
+
+    if [ "$(which $PGCTL)" == "" ]; then
+        echo "Could not find pg_ctl or pg_ctlcluster"
+	exit 1
```
MonsieurNicolas: mixed use of space/tab: (`exit 1` has leading tab instead of spaces)

alfiedotwtf: Gah! Sorry about that

MonsieurNicolas:
r+ d9fc22c6a9e21084c02498aaf87159a06ba60774

src/test/run-selftest-pg
```diff
@@ -30,6 +29,18 @@ PSQL=psql
...
+    findpg pg_ctlcluster
+    PGCTL=pg_ctlcluster
+
+    if [ "$(which $PGCTL)" == "" ]; then
+        echo "Could not find pg_ctl or pg_ctlcluster"
+        exit 1
+    fi
+fi
+
+exit
```
MonsieurNicolas: I think this `exit` was added by accident

alfiedotwtf: :(

Testing the tests. Yep, removed. Thanks for the pickup.

MonsieurNicolas:
r+ 22d4245ebb2d8340258d74d4c9bc7027b1737148

PR #1679, opened 2018-05-03
curl needs to create directories leading up to the file

alfiedotwtf:


MonsieurNicolas:
mmm, this doesn't sound right. core is supposed to be creating those dirs as temp folders. Do you have an issue with repro steps?

alfiedotwtf:
It's weird... I straced stellar-core and saw what was happening, and so raised this PR. I've since blown away my docker instance (from my own build, not the Stellar-provided one) and ran it with the same config, and it seems to be working now. Has anything changed Stellar side?

alfiedotwtf:
I haven't managed to reproduce this despite my trying, so I'll close this... but it's spooky-action-at-a-distance :(

PR #1681, opened 2018-05-05
Use FD_CLOEXEC on listening sockets and lock files (issue #1635)

mfontanini:
This sets FD_CLOEXEC on:

* The HTTP server socket.
* The peer listening socket on `PeerDoor`.
* Accepted sockets through `PeerDoor`.
* `TCPPeer` outbound sockets.
* Lock files. This is done via O_CLOEXEC rather than `fcntl`.

I couldn't find an existing file where this would fit so I created a new one.

### DIFF
--- a/src/overlay/PeerDoor.cpp
+++ b/src/overlay/PeerDoor.cpp
@@ -75,6 +75,7 @@ PeerDoor::handleKnock(shared_ptr<TCPPeer::SocketType> socket)
 {
     CLOG(DEBUG, "Overlay") << "PeerDoor handleKnock() @"
                            << mApp.getConfig().PEER_PORT;
+
     Peer::pointer peer = TCPPeer::accept(mApp, socket);
     if (peer)
     {
--- a/src/overlay/PeerDoor.cpp
+++ b/src/overlay/PeerDoor.cpp
@@ -32,7 +32,12 @@ PeerDoor::start()
         CLOG(DEBUG, "Overlay") << "PeerDoor binding to endpoint " << endpoint;
         mAcceptor.open(endpoint.protocol());
         mAcceptor.set_option(asio::ip::tcp::acceptor::reuse_address(true));
-        fd::disableProcessInheritance(mAcceptor);
+        if (!fd::disableProcessInheritance(mAcceptor))
+        {
+            CLOG(DEBUG, "Overlay")
+                << "Failed to disable process inheritance for "
+                << "listening socket";
+        }
         mAcceptor.bind(endpoint);
         mAcceptor.listen();
         acceptNextPeer();
--- a/src/overlay/PeerDoor.cpp
+++ b/src/overlay/PeerDoor.cpp
@@ -8,6 +8,7 @@
 #include "main/Config.h"
 #include "overlay/OverlayManager.h"
 #include "overlay/TCPPeer.h"
+#include "util/Fd.h"
 #include "util/Logging.h"
 #include <memory>
 
@@ -31,6 +32,7 @@ PeerDoor::start()
         CLOG(DEBUG, "Overlay") << "PeerDoor binding to endpoint " << endpoint;
         mAcceptor.open(endpoint.protocol());
         mAcceptor.set_option(asio::ip::tcp::acceptor::reuse_address(true));
+        fd::disableProcessInheritance(mAcceptor);
         mAcceptor.bind(endpoint);
         mAcceptor.listen();
         acceptNextPeer();
--- a/src/util/Fd.cpp
+++ b/src/util/Fd.cpp
@@ -1,0 +1,48 @@
+// Copyright 2018 Stellar Development Foundation and contributors. Licensed
+// under the Apache License, Version 2.0. See the COPYING file at the root
+// of this distribution or at http://www.apache.org/licenses/LICENSE-2.0
+
+#include "Fd.h"
+
+namespace stellar
+{
+namespace fd
+{
+
+#ifdef _WIN32
+
+// No-op on Windows
+bool
+disableProcessInheritance(asio::ip::tcp::acceptor& /*acceptor*/)
+{
+    return true;
+}
+
+bool
+disableProcessInheritance(int /*fileDescriptor*/)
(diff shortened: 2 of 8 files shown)


src/util/Fd.cpp
```diff
@@ -0,0 +1,48 @@
+// Copyright 2018 Stellar Development Foundation and contributors. Licensed
+// under the Apache License, Version 2.0. See the COPYING file at the root
+// of this distribution or at http://www.apache.org/licenses/LICENSE-2.0
+
+#include "Fd.h"
+
```
MonsieurNicolas: wondering if you should explicitely have `#include <fcntl.h>` in order to call fcntl (portability)

src/overlay/PeerDoor.cpp
```diff
@@ -31,6 +32,7 @@ PeerDoor::start()
         CLOG(DEBUG, "Overlay") << "PeerDoor binding to endpoint " << endpoint;
         mAcceptor.open(endpoint.protocol());
         mAcceptor.set_option(asio::ip::tcp::acceptor::reuse_address(true));
+        fd::disableProcessInheritance(mAcceptor);
```
MonsieurNicolas: note that there is a very small race condition here: if fork is called from a different thread between the call to "open" and this call, it will be too late.
This is why `SOCK_CLOEXEC` can be passed in to `socket` ; problem it seems is that I don't think there is a good way to fix this with the way asio works (let me know if you find otherwise)

src/overlay/PeerDoor.cpp
```diff
@@ -75,6 +75,12 @@ PeerDoor::handleKnock(shared_ptr<TCPPeer::SocketType> socket)
 {
     CLOG(DEBUG, "Overlay") << "PeerDoor handleKnock() @"
                            << mApp.getConfig().PEER_PORT;
+    if (!fd::disableProcessInheritance(socket->next_layer()))
```
MonsieurNicolas: as you fixed this bug in a subsequent commit, I would squash the fix into this commit.

Also - you have a typo on the commit title "desscriptor"

src/overlay/PeerDoor.cpp
mfontanini: Yeah, I'm aware of the race condition but I'm not sure how well that plays with asio. This is mostly an issue in tests right? And I don't know how the code flow works but is the `PeerDoor` created in a non multi threaded scenario?

Anyway, I'll try to see if there's a way to do this properly on asio. There's the issue of accepted sockets as well, for which this should use that same flag on `accept` but there doesn't seem to be an easy way to do this with asio.

src/overlay/PeerDoor.cpp
mfontanini: Sure, I'll fix this.

mfontanini:
I'll push these fixes probably tomorrow or Thursday night, sorry for the delay.

src/overlay/PeerDoor.cpp
MonsieurNicolas: It's not just in tests: we have worker threads (look for usage of `getWorkerIOService()`) that run in parallel. Right now we do not fork processes from those, but I imagine we will as it's the kind of work that we can easily punt to the side. If you can't find a simple solution that's fine for now: what we may be able to do in the future is do "cooperative forking" and not call fork when we know that there is a possible chance of hitting this race condition

mfontanini: I see, that makes sense. This is such stupid problem (and funnily enough we hit the same issue at work last week) but there doesn't seem to be an easy way to mitigate it. I found [this glib thread](https://mail.gnome.org/archives/gtk-devel-list/2015-March/msg00038.html) which is interesting but probably not the [best way](https://mail.gnome.org/archives/gtk-devel-list/2015-March/msg00044.html) to solve it. I'll have another look sometime this week.

MonsieurNicolas:
Let me known when I should take a look at this again

mfontanini:
Just fixed the commit history and added the fcntl.h include. Sorry it took me a while, I haven't had much time recently.

MonsieurNicolas:
Thank you.

I thought about the race condition, and also found that we missed to call the function on the result of `accept` in the http server which in turn made me think that trying to set this flag on fds with the presence of external libraries is probably a futile exercise (like soci to connect to postgres for example); so I looked for ways to mitigate this, and stumbled upon the dumbest solution that seems to just work and is portable, closing all fds other than stdin/stdout/stderr after forking:
```c++
  int omax =sysconf(_SC_OPEN_MAX);
  for(int i=3; i < omax; i++)
  {
    close(i);
  }
```
limits for descriptors are never that high (ie, not in the millions) so the loop is fast.

Let me know what you think - I realize that you spent some time on this (and you're quite busy) - but this seems to be a much more reliable solution than the one that I originally proposed (I feel dumb now).

mfontanini:
I actually considered doing that, the only issue is right now the code isn't using `fork` but instead it's using `posix_spawnp`. The main issue with moving towards `fork` is that detecting a failure in the execution of the child is slightly harder (e.g. using a pipe) but it should work if you want to go that way. This is actually the way in which we were planning on fixing this at work, where we just use `fork` so there's not much to be modified.

I really don't mind throwing this away. I'd rather submit a proper fix than a partial one that doesn't fully fix the issue :). Also, I will have some free time tonight so unless I'm burnt from work I could give this a go.

mfontanini:
I don't think I understand what you mean by that. Do you mean iterating through all fds before calling `posix_spawnp` (from 3 to max) and setting FD_CLOEXEC on them? Doesn't this still have a race condition where a fd is created in between you run the loop and `posix_spawnp` is called? Unless `FD_CLOEXEC` would affect non open fds but that sounds odd.

MonsieurNicolas:
Yes, this is not to address the race condition (I don't think we have any proposal yet on how to mitigate the race), this is to address the issue that we don't necessarily control the places that open descriptors (or simply forget to set `FD_CLOEXEC` on some descriptors).

mfontanini:
Closing in favor of #1690

PR #1684, opened 2018-05-08
Bucket apply fix (small change)

MonsieurNicolas:
This PR fixes:
* bucket benchmark test were nested in a transaction (which is not what bucketapplicator does) and didn't work with sqlite
* bucketapplicator was processing some items twice

### DIFF
--- a/src/bucket/BucketTests.cpp
+++ b/src/bucket/BucketTests.cpp
@@ -1109,36 +1109,44 @@ TEST_CASE("bucket apply", "[bucket]")
     REQUIRE(count == 1);
 }
 
-#ifdef USE_POSTGRES
 TEST_CASE("bucket apply bench", "[bucketbench][!hide]")
 {
-    VirtualClock clock;
-    Config cfg(getTestConfig(0, Config::TESTDB_POSTGRESQL));
-    Application::pointer app = createTestApplication(clock, cfg);
-    app->start();
+    auto runtest = [](Config::TestDbMode mode) {
+        VirtualClock clock;
+        Config cfg(getTestConfig(0, mode));
+        Application::pointer app = createTestApplication(clock, cfg);
+        app->start();
 
-    std::vector<LedgerEntry> live(100000);
-    std::vector<LedgerKey> noDead;
+        std::vector<LedgerEntry> live(100000);
+        std::vector<LedgerKey> noDead;
 
-    for (auto& l : live)
-    {
-        l.data.type(ACCOUNT);
-        auto& a = l.data.account();
-        a = LedgerTestUtils::generateValidAccountEntry(5);
-    }
+        for (auto& l : live)
+        {
+            l.data.type(ACCOUNT);
+            auto& a = l.data.account();
+            a = LedgerTestUtils::generateValidAccountEntry(5);
+        }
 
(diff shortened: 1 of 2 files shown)


src/bucket/BucketTests.cpp
```diff
@@ -1109,36 +1109,60 @@ TEST_CASE("bucket apply", "[bucket]")
...
 
-    auto& db = app->getDatabase();
-    auto& sess = db.getSession();
+        LedgerEntry prev;
+        for (auto& l : live)
+        {
+            if (!LedgerEntryIdCmp{}(prev.data, l.data))
+            {
+                auto k = KeyUtils::toStrKey(l.data.account().accountID);
+                CLOG(INFO, "Bucket") << "With duplicate bucket: " << k;
```
jonjove: I think this should say "With duplicate account: " rather than "With duplicate bucket: ". But what is the advantage of logging anything in this case? If we want a test where there are/aren't duplicates, we should set that up explicitly rather than logging when it does occur.

src/bucket/BucketTests.cpp
```diff
@@ -1109,36 +1109,60 @@ TEST_CASE("bucket apply", "[bucket]")
...
+        std::vector<LedgerEntry> live(100000);
+        std::vector<LedgerKey> noDead;
 
-    for (auto& l : live)
-    {
-        l.data.type(ACCOUNT);
-        auto& a = l.data.account();
-        a = LedgerTestUtils::generateValidAccountEntry(5);
-    }
+        for (auto& l : live)
```
jonjove: Why loop over `LedgerTestUtils::generateValidAccountEntry` instead of directly using `LedgerTestUtils::generateValidAccountEntries`?

src/bucket/BucketTests.cpp
```diff
@@ -1109,36 +1109,60 @@ TEST_CASE("bucket apply", "[bucket]")
...
+            }
+            prev = l;
+        }
 
-    CLOG(INFO, "Bucket") << "Applying bucket with " << live.size()
-                         << " live entries";
-    {
-        soci::transaction sqltx(sess);
+        std::shared_ptr<Bucket> birth =
+            Bucket::fresh(app->getBucketManager(), live, noDead);
```
jonjove: If you don't loop to check for duplicate accounts (lines 1135-1144), then there will be no need to `std::sort` the entries in advance (lines 1130-1133) since that is handled as part of `Bucket::fresh`.

src/bucket/BucketTests.cpp
MonsieurNicolas: I'm deleting this logging: it was leftover from the debugging session that led me to the fix in the BucketApplicator code

src/bucket/BucketTests.cpp
MonsieurNicolas: yes, got rid of that

src/bucket/BucketTests.cpp
MonsieurNicolas: I didn't write this code but it seems that it would require generating an array of AccountEntry and then transform it into an array of LedgerEntry.

jonjove:
r+ d8175ea

PR #1687, opened 2018-05-11
don't consume the sequence number if threshold not met (protocol 10)

MonsieurNicolas:
this is a small adjustment that ensures that we're future proof when it comes to behavior:
in protocol 10 we're processing sequence numbers in the "apply transaction" phase.

One of the things that can happen is that signers get updated in one transaction, invalidating other transactions from the transaction set (the same way bumpseq can invalidate other transactions).

With this change we properly no-op (other than take the fee) when encountering those invalidated transactions - with the current behavior (not live yet) we would have consumed the transaction number even though the transaction fails with `txBAD_AUTH`


MonsieurNicolas:
I don’t think we want this behavior yet as it’s technically allowing the transaction to be replayed twice (succeeds once though)

PR #1688, opened 2018-05-14
Transport layer

vogel:
Merge after #1670


marta-lokhova:
Could you maybe add some description about what you did in this PR or point to a doc if there's one?

vogel:
This is a first part of overlay overhaul.

It creates additional layer, Transport, that sits below Overlay.

Transport is responsible for connecting to peers (where overlay is responsible to selecting which peers to connect to), handling banned peers and sending/receiving messages.

Connection/Message/Transaction/EnvelopeHandler interfaces are a way to reduce cycles in dependencies between layers. For example: Herder can use Overlay and Overlay can use Transport, but it should not be possible the other way. Those interfaces allow to pass messages from Transport to Overlay and from Overlay to Herder without Transport and Overlay having any knowledge of Herder layer (this exposes problem with Application class, as currently it is possible for any class to use any other by using accessors from go-object Application, but this is topic for another discussion).

MonsieurNicolas:
sweeping stale pull requests

PR #1693, opened 2018-05-17
Randomize tests in jenkins

MonsieurNicolas:
Now that we are partitioning, it looks like some partitions run much longer than others.
Proposal here is to simply randomize partitions (in a deterministic way).


jonjove:
r+ e999e60

PR #1694, opened 2018-05-18
fix broken build

stanford-scs:
Need to reference sources relative to $(srcdir), as otherwise out-of-directory builds fail.


MonsieurNicolas:
r+ 9bcc735555e2512577743912b8392d21678d0a0e

PR #1698, opened 2018-05-21
Document performance evaluation

### LINKED ISSUE #1597: write a document to help to define how to grade a particular build's resource utilization when running a single node
If anyone wants to contribute performance improvements, it's a bit of a black box and hard to actually contribute as a submitted patch cannot quantify the improvements made to the code base.

Performance is a very broad topic, but it should be simple to put together a document that focuses on "local performance" (ie not related to network topology).

The document's goal is to enumerate the list of recipes that should be run as well as the metrics that would be monitored as part of those runs.

The recipes would be a combination of:
* tests, tagged properly ; they can be invoked with something like `--test [CPU]` (for tests targeting CPU utilization for example). All tests should be tagged with a common tag to make it easier to run all tests at once
* specific scenarios (like replay range N..M from the public network)

Metrics should be:
* execution time
* RAM utilization (minimum and peak)

The document should make it extremely simple for anybody to evaluate and compare changes as optimization work is usually full of trade-offs, regardless of platform.

I imagine that the first iteration of the document would only focus on RAM and CPU (and can be extended later to include IO).

On the platform front, the document could contain in an appendix recipes on recommended tools and usage.

MonsieurNicolas:
resolves #1597 

Did a first pass at writing some guidelines on how to perform performance evaluation of a branch / pull request. It's more of a starting point, but should already allow people to understand what to do.

I also cleaned the up issue template and pull request template.


performance-eval.md
```diff
@@ -0,0 +1,182 @@
+# Purpose of this document
+
+This document highlights the various ways a particular branch or changeset can be evaluated when it comes to performance characteristics.
+
+When submiting changes that may impact the network's performance, we require to perform some level of comparison of the changes between that version and  latest master or release.
```
marta-lokhova: typo "submitting"

MonsieurNicolas:
thanks!

PR #1701, opened 2018-05-25
Improve bucket verification

marta-lokhova:


src/work/Work.h
```diff
@@ -97,6 +97,16 @@ class Work : public WorkParent
     void advance();
     void reset();
 
+    size_t getMaxRetries()
```
MonsieurNicolas: I don't think those accessors add much value: you're just testing that the objects got initialized with the right parameters

src/work/Work.cpp
```diff
@@ -256,14 +256,14 @@ Work::advance()
                             << getUniqueName() << " successful, scheduling run";
         scheduleRun();
     }
-    else if (anyChildFatalFailure())
+    else if (anyChildFatalFailure() && allChildrenDone())
```
MonsieurNicolas: this seems wrong: why wait for all children to be done? If you are having problems on retry it probably points to a bug in the `onReset` method (or in the destructor as the way children are reset is by just deleting them).
btw, I spotted an issue with `RunProcess` that doesn't seem to properly handle reset as `ProcessManager` doesn't even seem to have a way to shutdown processes selectively, which could be part of the problem (other problems may include things like not using temp folders properly for example).

marta-lokhova: As discussed offline, in order to fix the problem with children not being properly cleared, we'd need to implement some sort of aborting logic. I made a note about it in the design doc, but for this iteration am going to leave it as is (as in, without waiting for children).

src/history/HistoryTests.cpp
```diff
@@ -204,6 +207,195 @@ TEST_CASE("History publish queueing", "[history][historydelay][historycatchup]")
...
+            catchupSimulation.getHistoryConfigurator().getArchiveDirName();
+        REQUIRE(!dir.empty());
+        auto bucketFile =
+            dir + "/" +
+            fs::remoteName(HISTORY_FILE_TYPE_BUCKET, hashes[0], "xdr");
+
+        // Corrupt the content of a bucket file
+        std::remove((bucketFile + ".gz").c_str());
+        {
+            std::string s = "I am corrupted";
```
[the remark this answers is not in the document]
marta-lokhova: I don't think that's the case. I added an extra REQUIRE that shows that download (which is a child of verify) is successful.

src/historywork/BatchDownloadVerifyBucketsWork.cpp
```diff
@@ -0,0 +1,78 @@
+// Copyright 2018 Stellar Development Foundation and contributors. Licensed
```
MonsieurNicolas: this doesn't belong in this commit: split into refactor commits and changes to how we download and verify buckets

MonsieurNicolas: (also new files need to be added to the Visual Studio project)

src/historywork/BatchWork.cpp
```diff
@@ -0,0 +1,127 @@
...
+#include "catchup/CatchupManager.h"
+#include "lib/util/format.h"
+#include "main/Application.h"
+#include "util/Logging.h"
+#include <history/FileTransferInfo.h>
+
+namespace stellar
+{
+
+BatchWork::BatchWork(Application& app, WorkParent& parent,
```
MonsieurNicolas: `BatchWork` should introduce the building block for batching - this one does too much as it hardcodes "download" (so you cannot batch generic work with it).

src/historywork/BatchDownloadVerifyBucketsWork.cpp
```diff
@@ -0,0 +1,78 @@
...
+
+#include "BatchDownloadVerifyBucketsWork.h"
+#include "VerifyBucketWork.h"
+#include "historywork/GetAndUnzipRemoteFileWork.h"
+#include "lib/util/format.h"
+
+namespace stellar
+{
+
+BatchDownloadVerifyBucketsWork::BatchDownloadVerifyBucketsWork(
```
MonsieurNicolas: I am not sure this `BatchDownloadVerifyBucketsWork` implements the right contract:
I would expect it to be a batch of `<verify + download>` work (and therefore implement a generator function for verify+download)

src/historywork/BatchWork.h
```diff
@@ -0,0 +1,54 @@
...
+  public:
+    BatchWork(Application& app, WorkParent& parent, std::string const& type,
+              TmpDir const& downloadDir);
+    virtual ~BatchWork() override;
+    virtual void onReset() override;
+    virtual void notify(std::string const& child) override;
+
+    virtual bool hasNext() = 0;
+    virtual std::string getNextDownloadFilename() = 0;
+    virtual std::string assignChildWork(std::string ft) = 0;
```
MonsieurNicolas: `assignChildWork` does not make sense to me - I would expect a function that returns or modifies some `Work` instead of hijacking the assignment of a child from a filename

src/historywork/BatchDownloadVerifyBucketsWork.cpp
MonsieurNicolas: Also - why do we need a new class? I would expect `DownloadBucketsWork` to just be refactored (as it's already a class that "downloads and verifies buckets")

src/historywork/BatchWork.cpp
```diff
@@ -0,0 +1,72 @@
...
+        CLOG(WARNING, "Work")
+            << "BatchWork notified by unknown child " << child;
+        return;
+    }
+
+    assert(i->second);
+    markMetrics(*(i->second));
+
+    std::vector<std::string> done;
+    for (auto const& c : mChildren)
```
jonjove: `std::remove_if` would probably be a nicer way to implement this. If you do this, you might consider refactoring lines 29-33 into a function which is used both on line 29 and as part of this.

src/historywork/BatchWork.h
```diff
@@ -0,0 +1,40 @@
...
+       work if it has bandwidth.
+    **/
+  public:
+    BatchWork(Application& app, WorkParent& parent, std::string name);
+    ~BatchWork() override;
+    void onReset() override;
+    void notify(std::string const& child) override;
+
+    // Optionally mark metrics
+    virtual void markMetrics(Work& work){};
```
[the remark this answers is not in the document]
marta-lokhova: `notify` logically belongs to BatchWork and is the same for all children (with minor variation in metrics), so I did not want to duplicate same code in child classes. I also did not want for children to override `notify` with a dependency on parent's `notify`.

src/history/HistoryTests.cpp
```diff
@@ -204,6 +206,168 @@ TEST_CASE("History publish queueing", "[history][historydelay][historycatchup]")
           catchupSimulation.getApp().getLedgerManager().getLedgerNum());
 }
 
+TEST_CASE("History bucket verification",
+          "[history][bucketverification[batching]]")
+{
+    /* Tests bucket verification stage of catchup. Assumes ledger chain
+     * verification was successful. **/
+
+    CatchupSimulation catchupSimulation{};
```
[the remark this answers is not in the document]
marta-lokhova: None of these perform catchup end to end, "History bucket verification" specifically tests DownloadBucketsWork. `catchupSimulation` is needed to publish correct history and get proper hashes. The second app does not really start catchup, it spins up a test app with proper configs (which perhaps could be changed to simply using TestApplication).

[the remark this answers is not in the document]
MonsieurNicolas: right, my point is exactly that: because those tests don't perform end to end catchup, we don't need the test infrastructure that is used for end to end catchup (the overkill part), especially because here you are actually overriding the data produced anyways.

src/historywork/BatchWork.h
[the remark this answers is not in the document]
MonsieurNicolas: I am not sure I understand what you are saying: one of the typical patterns for overriding a method is to call the parent implementation to avoid duplication. I don't see why metrics should be treated differently in the context of `batchWork`, maybe there should be a `updateMetrics` member (that doesn't take parameters) exposed at the `Work` level, but that seems to be a different conversation.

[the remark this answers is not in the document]
marta-lokhova: Although common, I don't think it's necessarily right here, as it requires every subclass to know that it must call parent's notify. 
I was also thinking if it makes sense to move metrics into download work classes (like GetAndUnzipRemoteFileWork) instead of batching classes. That would require keeping track of file types though.

src/historywork/BatchWork.cpp
marta-lokhova: `std::remove_if` is not applicable for associative containers like `std::map`.

jonjove: I did not realize that `mChildren` is associative, and you are correct that `std::remove_if` does not work for associative containers. However, there is still a better solution than the two-stage delete approach used here. `std::map::erase` returns a valid iterator to the element following the last removed element. Then the following is possible:
```c++
for (auto it = mChildren.begin(); it != mChildren.end(); )
{
    if (it->second->getState() == WORK_SUCCESS)
    {
        it = mChildren.erase(it);
        // CLOG
        // yieldMoreWork if hasNext
    }
    else
    {
        ++it;
    }
}
```
There are a few instances of this pattern in existing code, compare against `SetOptionsOpFrame::doApply` (starting at line 164) and `SCP::purgeSlots` (starting at line 81) for example.

marta-lokhova:
Needs to be rebased once updates are reviewed.

src/historywork/DownloadBucketsWork.h
```diff
@@ -1,40 +1,42 @@
...
+    void resetIter() override;
     std::string getStatus() const override;
-    void onReset() override;
     void notify(std::string const& child) override;
+
+  private:
+    std::map<std::string, std::shared_ptr<Bucket>> mBuckets;
+    std::map<std::string, std::shared_ptr<Work>> mVerifyDownloadMap;
+    std::vector<std::string> mHashes;
+    std::vector<std::string>::const_iterator mNextBucketIndex;
```
jonjove: Maybe `mNextBucketIter` now?

src/history/HistoryTests.cpp
```diff
@@ -204,6 +206,119 @@ TEST_CASE("History publish queueing", "[history][historydelay][historycatchup]")
...
+TEST_CASE("History bucket verification",
+          "[history][bucketverification][batching]")
+{
+    /* Tests bucket verification stage of catchup. Assumes ledger chain
+     * verification was successful. **/
+
+    Config cfg(getTestConfig(1, Config::TESTDB_ON_DISK_SQLITE));
+    VirtualClock clock;
+    auto cg = std::make_shared<TmpDirHistoryConfigurator>();
+    cg->configure(cfg, true);
```
MonsieurNicolas: why are you passing `true` here? Test app is not publishing to this archive, is it?

src/history/HistoryTests.cpp
```diff
@@ -204,6 +206,119 @@ TEST_CASE("History publish queueing", "[history][historydelay][historycatchup]")
...
+    {
+        hashes.push_back(bucketGenerator.generateBucket(
+            ArchiveFileState::CONTENTS_AND_HASH_OK));
+
+        // Download and verify buckets
+        auto verify =
+            wm.executeWork<DownloadBucketsWork>(mBuckets, hashes, *tmpDir);
+        REQUIRE(verify->getState() == Work::WORK_SUCCESS);
+    }
+    SECTION("verify fail due to corrupted content")
```
MonsieurNicolas: missing test for download fails?

src/history/HistoryTests.cpp
```diff
@@ -204,6 +206,119 @@ TEST_CASE("History publish queueing", "[history][historydelay][historycatchup]")
...
+        *app, app->getHistoryArchiveManager().getHistoryArchive("test")};
+    std::vector<std::string> hashes;
+    auto& wm = app->getWorkManager();
+    std::map<std::string, std::shared_ptr<Bucket>> mBuckets;
+    auto tmpDir =
+        std::make_unique<TmpDir>(app->getTmpDirManager().tmpDir("bucket-test"));
+
+    SECTION("successful download and verify")
+    {
+        hashes.push_back(bucketGenerator.generateBucket(
```
MonsieurNicolas: I think tests should push more than 1 bucket as we expect that work to do its job with a vector. I'd suggest always adding a "good" bucket and then have the various sections add a second one that is bad.
No need to add tests for single bucket vs many at this time

src/history/HistoryTests.cpp
```diff
@@ -204,6 +206,119 @@ TEST_CASE("History publish queueing", "[history][historydelay][historycatchup]")
...
+    yieldMoreWork() override
+    {
+        return addWork<Work>(fmt::format("child-{:d}", mCount++), 0)
+            ->getUniqueName();
+    }
+};
+
+TEST_CASE("Work batching", "[batching]")
+{
+    Config cfg(getTestConfig(0, Config::TESTDB_ON_DISK_SQLITE));
```
MonsieurNicolas: you don't need a persistent database

src/history/HistoryTestsUtils.cpp
```diff
@@ -89,6 +90,86 @@ S3HistoryConfigurator::configure(Config& mCfg, bool writable) const
...
+        put(bucketEntry);
+    }
+
+    // Finish writing and close the bucket file
+    assert(mBuf);
+    mOut.writeOne(*mBuf, mHasher.get(), &mBytesPut);
+    mObjectsPut++;
+    mBuf.reset();
+
+    if (state == ArchiveFileState::CORRUPTED_CONTENTS)
```
MonsieurNicolas: There are 2 types of corruptions and bad data that we want to check:
* hash mismatch - which is what you ended up doing here but it would have been better to just change the hash by changing the last byte of the hash to a different value instead of changing the content
* good hash but the content is invalid (that will trip validation in different ways)
  * invalid because of duplicate content like what you did here ; really what we want is that the ledger entries are not sorted, so swapping two elements before writing to disk might be a better test than duplicate
  * invalid because of garbage (that doesn't represent a ledger entry and will trip marshaling code)

src/history/HistoryTestsUtils.cpp
```diff
@@ -89,6 +90,86 @@ S3HistoryConfigurator::configure(Config& mCfg, bool writable) const
...
+    {
+        REQUIRE(NUM_ITEMS_PER_BUCKET);
```
MonsieurNicolas: no need for this `REQUIRE` : `bucketEntries[0]` would throw

src/history/HistoryTestsUtils.cpp
```diff
@@ -89,6 +90,86 @@ S3HistoryConfigurator::configure(Config& mCfg, bool writable) const
...
+        mApp.getTmpDirManager().tmpDir("tmp-bucket-generator"));
+}
+
+std::string
+TestBucketGenerator::generateBucket(ArchiveFileState state)
+{
+    BucketOutputIteratorForTesting bucketOut{mTmpDir->getName()};
+    uint256 hash;
+
+    if (state == ArchiveFileState::CORRUPTED_HASH)
```
MonsieurNicolas: this looks more like a missing file test than corrupted hash

src/history/HistoryTestsUtils.cpp
```diff
@@ -89,6 +90,86 @@ S3HistoryConfigurator::configure(Config& mCfg, bool writable) const
...
+
+    {
+        auto& wm = mApp.getWorkManager();
+        auto archive =
+            mApp.getHistoryArchiveManager().getHistoryArchive("test");
+        auto put = wm.addWork<PutRemoteFileWork>(filename + ".gz",
+                                                 ft.remoteName(), archive);
+        auto mkdir = put->addWork<MakeRemoteDirWork>(ft.remoteDir(), archive);
+        auto gzip = mkdir->addWork<GzipFileWork>(filename, true);
+        gzip->advance();
```
MonsieurNicolas: do we also want a test that corrupts the zip file (to cause unzip to fail)?

src/history/HistoryTestsUtils.h
```diff
@@ -22,7 +29,16 @@ class HistoryManager;
 namespace historytestutils
 {
 
+enum class ArchiveFileState
```
MonsieurNicolas: `ArchiveFileState` is not a good name for this. Call it `TestBucketFileType` that gives a better idea of what it is

src/historywork/BatchDownloadWork.cpp
```diff
@@ -19,21 +17,19 @@ BatchDownloadWork::BatchDownloadWork(Application& app, WorkParent& parent,
                                      CheckpointRange range,
                                      std::string const& type,
                                      TmpDir const& downloadDir)
-    : Work(app, parent,
-           fmt::format("batch-download-{:s}-{:08x}-{:08x}", type, range.first(),
-                       range.last()))
+    : BatchWork{app, parent,
```
MonsieurNicolas: Stay consistent with the use of initializer list syntax. Also changing from one form to the other doesn't accomplish anything other than making rebasing more painful

src/history/HistoryTests.cpp
marta-lokhova: Test app uses `PutRemoteFileWork` to upload buckets, which requires put command to be configured.

marta-lokhova:
Regarding the actual bucket verification code, I noticed that it does not enforce XDR format when reading the file and hashes everything anyway (see https://github.com/stellar/stellar-core/blob/06c5bac64f9865a707b79038141db52ad1db4f33/src/historywork/VerifyBucketWork.cpp#L59). I wonder what's the rationale behind that. I think this could also be fixed by using `readOne` in `XDRInputFileStream`. Thoughts, @MonsieurNicolas?

src/historywork/DownloadBucketsWork.cpp
```diff
@@ -17,16 +15,23 @@ DownloadBucketsWork::DownloadBucketsWork(
     Application& app, WorkParent& parent,
     std::map<std::string, std::shared_ptr<Bucket>>& buckets,
     std::vector<std::string> hashes, TmpDir const& downloadDir)
-    : Work{app, parent, "download-and-verify-buckets"}
+    : BatchWork{app, parent, "download-verify-buckets"}
     , mBuckets{buckets}
-    , mHashes{std::move(hashes)}
+    , mHashes{hashes}
+    , mNextBucketIndex{hashes.begin()}
```
MonsieurNicolas: 1. you forgot to rename `mNextBucketIndex` into ``mNextBucketIterator``
2. this is wrong: `mNextBucketIterator` should be `mHashes.begin()` not `hashes.begin()`

src/historywork/DownloadBucketsWork.cpp
```diff
@@ -17,16 +15,23 @@ DownloadBucketsWork::DownloadBucketsWork(
...
     , mDownloadDir{downloadDir}
     , mDownloadBucketStart{app.getMetrics().NewMeter(
-          {"history", "download-bucket", "start"}, "event")}
+          {"history", fmt::format("download-{:s}", HISTORY_FILE_TYPE_BUCKET),
```
MonsieurNicolas: this change is unrelated to this PR

src/historywork/DownloadBucketsWork.cpp
```diff
@@ -40,41 +45,62 @@ DownloadBucketsWork::getStatus() const
...
     if (mState == WORK_RUNNING || mState == WORK_PENDING)
     {
-        return "downloading buckets";
+        if (!mHashes.empty())
+        {
+            auto numDone = std::distance(mHashes.begin(), mNextBucketIndex);
+            auto total = static_cast<uint32_t>(mHashes.size());
+            auto pct = (100 * numDone) / total;
+            return fmt::format(
+                "downloading and verifying buckets: {:d}/{:d} ({:d}%)", numDone,
```
MonsieurNicolas: nice

src/historywork/DownloadBucketsWork.cpp
```diff
@@ -40,41 +45,62 @@ DownloadBucketsWork::getStatus() const
...
 DownloadBucketsWork::notify(std::string const& child)
 {
-    auto i = mChildren.find(child);
-    if (i == mChildren.end())
-    {
-        CLOG(WARNING, "Work")
-            << "DownloadBucketsWork notified by unknown child " << child;
-        return;
-    }
+    auto downloadWork = mVerifyDownloadMap.find(child);
```
MonsieurNicolas: Looks like `mVerifyDownloadMap` is only used to ensure that we don't update the wrong metrics - I would ditch `mVerifyDownloadMap` entirely and keep the code simple.
If there is a bug in how notify is called we have bigger problems as in the code base we seem to ignore those. On that front, (and general), we should log `notified by unknown child` with an `ERROR` log level instead of `INFO`.

src/bucket/BucketOutputIterator.cpp
```diff
@@ -76,6 +70,18 @@ BucketOutputIterator::put(BucketEntry const& e)
     *mBuf = e;
 }
 
+bool
+BucketOutputIterator::enforceSorted(BucketEntry const& e)
```
MonsieurNicolas: per our chat, I don't think we need to mess with out of order entries, ie we don't need this method `enforceSorted`

src/history/HistoryTests.cpp
```diff
@@ -204,6 +225,145 @@ TEST_CASE("History publish queueing", "[history][historydelay][historycatchup]")
...
+            TestBucketState::CORRUPTED_ZIPPED_FILE));
+
+        auto verify =
+            wm.executeWork<DownloadBucketsWork>(mBuckets, hashes, *tmpDir);
+        REQUIRE(verify->getState() == Work::WORK_FAILURE_RAISE);
+        downloadStatusCheck(*verify, false);
+    }
+    SECTION("verify fails with duplicate buckets")
+    {
+        hashes.push_back(bucketGenerator.generateBucket(
```
MonsieurNicolas: per our chat, duplicate buckets and out of order buckets don't need to be tested here (we can rely 100% on hash and the fact that the overall network is well behaved from that point of view)

src/history/HistoryTestsUtils.cpp
```diff
@@ -287,10 +398,10 @@ CatchupSimulation::generateRandomLedger()
     mLedgerCloseDatas.emplace_back(ledgerSeq, txSet, sv);
     lm.closeLedger(mLedgerCloseDatas.back());
 
-    mLedgerSeqs.push_back(lm.getLastClosedLedgerHeader().header.ledgerSeq);
-    mLedgerHashes.push_back(lm.getLastClosedLedgerHeader().hash);
-    mBucketListHashes.push_back(
-        lm.getLastClosedLedgerHeader().header.bucketListHash);
+    auto lclh = lm.getLastClosedLedgerHeader();
```
MonsieurNicolas: thanks! Better would be a const reference

marta-lokhova:
Updated and rebased.

src/historywork/BatchWork.cpp
```diff
@@ -0,0 +1,68 @@
...
+
+    assert(i->second);
+
+    for (auto childIt = mChildren.begin(); childIt != mChildren.end();)
+    {
+        if (childIt->second->getState() == WORK_SUCCESS)
+        {
+            CLOG(DEBUG, "History") << "Finished child work " << childIt->first;
+            childIt = mChildren.erase(childIt);
+            if (hasNext())
```
vogel: I see that this is taken almost literary from BatchDownloadWork, but only now I have realized that this may not be proper way of adding more work in 100% cases.

Imagine MAX_CONCURRENT_SUBPROCESSES is 16 and at the onReset call there are only 8 work items to do and after first completed work there are additional 32 ones. With current code it would only be able to execute 8 work items in parallel.

I think it would be nice to extract last few lines from onReset to new method runAsManyJobsAsPossible (but with a better name) and call it from onReset and just after this for loop.

src/historywork/BatchWork.h
```diff
@@ -0,0 +1,34 @@
...
+    void onReset() override;
+
+    // Note: If a subclass chooses to override notify, it needs to make sure
+    // it calls notify of BatchWork.
+    void notify(std::string const& child) override;
+
+  protected:
+    virtual bool hasNext() = 0;
+    virtual std::string yieldMoreWork() = 0;
+    virtual void resetIter() = 0;
```
vogel: I dont like the name, maybe `resetQueue`?

src/historywork/BatchDownloadWork.cpp
```diff
@@ -53,94 +49,57 @@ BatchDownloadWork::getStatus() const
...
 }
 
-void
-BatchDownloadWork::notify(std::string const& child)
+std::string
+BatchDownloadWork::yieldMoreWork()
 {
-    auto i = mChildren.find(child);
-    if (i == mChildren.end())
+    if (!hasNext())
```
vogel: I think there is some problem with the API here. yieldMoreWork is called only in BatchWork and its result value is not used for anything, it is also always proceeded by hasNext call. I think this suggest that it would be better to make it return bool, true when new work has been yielded and false if otherwise.

Then hasNext could be removed from BatchWork and return value of yieldMoreWork would be useful.

src/historywork/BatchDownloadWork.cpp
```diff
@@ -53,94 +49,57 @@ BatchDownloadWork::getStatus() const
...
+bool
+BatchDownloadWork::hasNext()
 {
-    if (mNext > mRange.last())
-    {
-        return;
-    }
-
-    FileTransferInfo ft(mDownloadDir, mFileType, mNext);
-    if (fs::exists(ft.localPath_nogz()))
```
vogel: Are we not checking that anymore in new version?

src/historywork/BatchDownloadWork.cpp
marta-lokhova: I don't think we should this; this logic will be re-used as we add BatchWork's child classes, especially if an item in the batch depends on the previously yielded work. In fact, the new ledger verification code (not pushed yet) needs to keep track of yielded work, so having it as bool is not sufficient.

src/historywork/BatchWork.cpp
marta-lokhova: I'm not sure the scenario where `there are only 8 work items to do and after first completed work there are additional 32 ones` is possible, could you elaborate? I don't think any children of `BatchWork` should allow for range changes (that would cause additional work to be added)

vogel: This is more for a general-correctness, as I'm not aware of a current work item that allows for range changes.

It serves two, mostly cosmetical, purposes:
1. to split cleaning up completed work items from launching new ones, for more single-responsibility in the code
2. to have only one place in code that launches new work items, removing repetition

src/historywork/BatchDownloadWork.cpp
vogel: OK. Then another proposition is to make yieldMoreWork return empty string when no work has been executed. That could save us one possible throw and one call to hasNext per iteration.

marta-lokhova: The logic behind current implementation is to mimic a standard iterator, where the responsibility  of maintaining the iterator range is completely up to the caller. `hasNext` is pretty cheap, as it's a single comparison, so I'm not sure how your suggestion would make the code better.

src/historywork/BatchWork.cpp
marta-lokhova: Ok, I can add that in case there's any work in the future that would add more work while running.

src/historywork/BatchWork.h
marta-lokhova: Hmm, I don't think `resetQueue` is better. We don't really have a queue here, but we are iterating over a range, generating work on the go (not queuing it)

src/historywork/BatchWork.cpp
marta-lokhova: Updated.

MonsieurNicolas:
Looks like there was some problem with the visual studio file: I added a commit that you should squash into the corresponding commit @marta-lokhova

src/history/HistoryTests.cpp
```diff
@@ -204,6 +225,124 @@ TEST_CASE("History publish queueing", "[history][historydelay][historycatchup]")
...
+    SECTION("no hashes to verify")
+    {
+        // Ensure proper behavior when no hashes are passed in
+        auto verify = wm.executeWork<DownloadBucketsWork>(
+            mBuckets, std::vector<std::string>(), *tmpDir);
+        REQUIRE(verify->getState() == Work::WORK_SUCCESS);
+    }
+}
+
+class TestBatchWork : public BatchWork
```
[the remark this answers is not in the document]
marta-lokhova: Placed TestBatchWork in its own namespace (like in some tests in the codebase; i don't think we actually have any pattern)

marta-lokhova:
Thanks for reviewing! squashed VS fix commit, addressed the comments and rebased on top of master.

marta-lokhova:
this is now rebased

MonsieurNicolas:
and this needs to be rebased again @marta-lokhova

marta-lokhova:
rebased

src/history/HistoryTests.cpp
```diff
@@ -133,6 +172,25 @@ dbModeName(Config::TestDbMode mode)
     }
 }
 
+static void
+downloadStatusCheck(Work& parent, bool success)
+{
+    for (auto const& child : parent.getChildren())
+    {
+        if (child.second->isDone())
```
MonsieurNicolas: I think that in the context of tests (that should be deterministic):
* we should have `REQUIRE(child.second->isDone())` (otherwise, we just silently skip work)
* on the actual state of `child.second`,
   * we should not check what children of children are doing (I am not sure why we do this?)
   * instead we should check that `child.second->getState()` has the expected value

To make things clearer, we could have this function take a pointer to `DownloadBucketsWork` instead of just `Work`.

MonsieurNicolas:
r+ b2379edb5ee8ea981c5ddbdc2ae0cbfd8d7daf14

MonsieurNicolas:
r+ 8205862b571839e6c92b1e325b81cd818d840417

PR #1709, opened 2018-06-06
Signature verification 10

### LINKED ISSUE #1464: txBAD_AUTH_EXTRA Signature bug
I was trying out something I'd thought would be a simple thing yesterday, and I still can't figure it out.

I have two accounts, `ownerId` and `assetId`. When this starts, `assetId` has masterweight = 0, and is signed for by `ownerId`.

My transaction tries to change back master weight to 1, remove `ownerId` as a signer, before finally merging `assetId` back into `ownerId`. It fails.

```
return horizon.loadAccount(ownerId)
.then(function (account) {
    return new stellar.TransactionBuilder(account)
    .addOperation(stellar.Operation.setOptions({
        source: assetId,
        masterWeight: 1,
        signer: {
            ed25519PublicKey: ownerId,
            weight: 0
        }
    }))
    .addOperation(stellar.Operation.accountMerge({
        source: assetId,
        destination: ownerId
    }))
    .build();
});
```

`ownerId` is the envelope source account, and pays for the fees.
`assetId` is the source account for all the individual operations.

Before setOptions, `assetId` is signed for by `sourceId`
Before accountMerge, `assetId` is signed for by itself.

By now it should be clear that both accounts need to sign the transaction.
So why does this result in a failure of not all signatures being used? None of the signatures on their own are enough.

https://www.stellar.org/laboratory/#xdr-viewer?input=AAAAAIMQY6sfxBUJuz9KxDd7XTlymmYQ1Yvl41noMjNWqVrEAAAAyABbLc8AAAACAAAAAAAAAAAAAAACAAAAAQAAAABSZn6pw9muRSb60r3%2BVmVDYvEVAqy8TtUpdT6BZbpKFgAAAAUAAAAAAAAAAAAAAAAAAAABAAAAAQAAAAAAAAAAAAAAAAAAAAAAAAABAAAAAIMQY6sfxBUJuz9KxDd7XTlymmYQ1Yvl41noMjNWqVrEAAAAAAAAAAEAAAAAUmZ%2BqcPZrkUm%2BtK9%2FlZlQ2LxFQKsvE7VKXU%2BgWW6ShYAAAAIAAAAAIMQY6sfxBUJuz9KxDd7XTlymmYQ1Yvl41noMjNWqVrEAAAAAAAAAAJWqVrEAAAAQMpPaByRlskf5W%2BDSYgnpsf%2FKdgD1sq%2FTB6szPJOAnqRLEAPj45pMPiPYTb5uPlpMVBnsff7sJFgyZJqIhYnJAllukoWAAAAQI%2B4TDrs5jjGYdmXtiqiEz16Sl44%2Fnr4LP3HzmvLtP%2BXvj5gor%2BlQ7TDV8YUwvzseqDO9PggyQPGs0ufoG2lAwQ%3D&type=TransactionEnvelope&network=test

vogel:
# Description

Resolves #1464

Those commits changes way how signatures are checked during applying phase of transactions. In general, signatures for all operations are now checked in batch just before apply phase (after sequence number processing).

# Checklist
- [x] Reviewed the [contributing](../CONTRIBUTING.md) document
- [x] Ran `clang-format`
- [x] Compiles
- [x] Ran all tests
- [x] If change impacts performance, include supporting evidence per the [performance document](../performance-eval.md)

### DIFF
--- a/src/test/TestAccount.cpp
+++ b/src/test/TestAccount.cpp
@@ -2,13 +2,13 @@
 // under the Apache License, Version 2.0. See the COPYING file at the root
 // of this distribution or at http://www.apache.org/licenses/LICENSE-2.0
 
-#include "TestAccount.h"
-
+#include "test/TestAccount.h"
 #include "ledger/DataFrame.h"
-#include "lib/catch.hpp"
 #include "main/Application.h"
 #include "test/TestExceptions.h"
 #include "test/TxTests.h"
+
+#include <lib/catch.hpp>
 
 namespace stellar
 {
@@ -158,13 +158,9 @@ TestAccount::hasTrustLine(Asset const& asset) const
 }
 
 void
-TestAccount::setOptions(AccountID* inflationDest, uint32_t* setFlags,
-                        uint32_t* clearFlags, ThresholdSetter* thrs,
-                        Signer* signer, std::string* homeDomain)
+TestAccount::setOptions(SetOptionsArguments const& arguments)
 {
-    applyTx(tx({txtest::setOptions(inflationDest, setFlags, clearFlags, thrs,
-                                   signer, homeDomain)}),
-            mApp);
+    applyTx(tx({txtest::setOptions(arguments)}), mApp);
 }
 
 void
--- a/src/test/TxTests.cpp
+++ b/src/test/TxTests.cpp
@@ -188,8 +188,7 @@ applyCheck(TransactionFramePtr tx, Application& app, bool checkSeqNum)
         // checks that the failure is the same if pre checks failed
         if (!check)
         {
-            if (app.getLedgerManager().getCurrentLedgerVersion() >= 10 ||
-                tx->getResultCode() != txFAILED)
+            if (tx->getResultCode() != txFAILED)
             {
                 REQUIRE(checkResult == tx->getResult());
             }
@@ -320,7 +319,6 @@ TxSetResultMeta
 closeLedgerOn(Application& app, uint32 ledgerSeq, int day, int month, int year,
               std::vector<TransactionFramePtr> const& txs)
 {
-
     auto txSet = std::make_shared<TxSetFrame>(
         app.getLedgerManager().getLastClosedLedgerHeader().hash);
 
--- a/src/test/TxTests.cpp
+++ b/src/test/TxTests.cpp
@@ -188,7 +188,25 @@ applyCheck(TransactionFramePtr tx, Application& app, bool checkSeqNum)
         // checks that the failure is the same if pre checks failed
         if (!check)
         {
-            REQUIRE(checkResult == tx->getResult());
+            if (app.getLedgerManager().getCurrentLedgerVersion() >= 10 ||
+                tx->getResultCode() != txFAILED)
+            {
+                REQUIRE(checkResult == tx->getResult());
+            }
+            else
+            {
+                auto const& txResults = tx->getResult().result.results();
+                auto const& checkResults = checkResult.result.results();
+                for (auto i = 0u; i < txResults.size(); i++)
+                {
+                    REQUIRE(checkResults[i] == txResults[i]);
+                    if (checkResults[i].code() == opBAD_AUTH)
+                    {
+                        // results may not match after first opBAD_AUTH
+                        break;
+                    }
+                }
+            }
         }
 
         if (code != txNO_ACCOUNT)
--- a/src/test/TxTests.cpp
+++ b/src/test/TxTests.cpp
@@ -353,6 +353,12 @@ getAccount(const char* n)
     return SecretKey::fromSeed(seed);
 }
 
+Signer
+makeSigner(SecretKey key, int weight)
+{
+    return Signer{KeyUtils::convertKey<SignerKey>(key.getPublicKey()), weight};
+}
+
 AccountFrame::pointer
 loadAccount(PublicKey const& k, Application& app, bool mustExist)
 {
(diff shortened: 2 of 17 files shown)


src/transactions/TxResultsTests.cpp
```diff
@@ -755,17 +756,17 @@ TEST_CASE("txresults", "[tx][txresults]")
                                         {PAYMENT_SUCCESS, SET_OPTIONS_SUCCESS,
                                          opBAD_AUTH}));
             });
-            for_versions({7}, *app, [&] {
+            for_versions({8, 9}, *app, [&] {
```
MonsieurNicolas: don't reverse the order of tests here

src/transactions/TransactionFrame.cpp
```diff
@@ -606,6 +637,10 @@ TransactionFrame::apply(LedgerDelta& delta, TransactionMetaV1& meta,
         {
             processSeqNum(app.getLedgerManager(), txDelta);
         }
+        if (cv == ValidationType::kFullyValid)
```
MonsieurNicolas: I think it would be cleaner to have:
* a new `ValidationType::kInvalidPostAuth` state (if `checkSignature` returns true in `TransactionFrame::commonValid`)
* call `processSignatures` if `cv >= ValidationType::kInvalidPostAuth`
    * this removes the need for `processSignature` to "rollback" `cv` like it does now and basically follows the same pattern than `processSeqNum`

MonsieurNicolas: now that I think of it (spec issue?) maybe we should not `processSeqNum` if auth is not good?
I am thinking scenario of
* tx_n: setOptions(signers, ...)
* tx_n+1: oldSigners ...
there is also a different transaction `tx_n+1: newSigners` that can be executed
in this case we take the fee, but don't mess with the sequence number to avoid invalidating a potential other transaction

`txINSUFFICIENT_BALANCE` cannot happen when applying in v10 so this seems ok.

src/transactions/TxEnvelopeTests.cpp
```diff
@@ -1025,14 +1027,10 @@ TEST_CASE("txenvelope", "[tx][envelope]")
...
             for_all_versions(*app, [&] {
-                applyCheck(tx, *app);
-                REQUIRE(tx->getResultCode() == txSUCCESS);
-                REQUIRE(SetOptionsOpFrame::getInnerCode(
-                            tx->getOperations()[0]->getResult()) ==
-                        SET_OPTIONS_SUCCESS);
-                REQUIRE(SetOptionsOpFrame::getInnerCode(
-                            tx->getOperations()[1]->getResult()) ==
-                        SET_OPTIONS_SUCCESS);
+                validateTxResults(
```
MonsieurNicolas: nice!

src/transactions/TransactionFrame.cpp
MonsieurNicolas: scratch that last one: it would allow the transaction to be replayed twice (would just need to change signers or signatures)

src/test/TxTests.cpp
```diff
@@ -114,11 +114,12 @@ applyCheck(TransactionFramePtr tx, Application& app, bool checkSeqNum)
 
         REQUIRE((!res || tx->getResultCode() == txSUCCESS));
 
+        // this is not true for protocol versions < 10
```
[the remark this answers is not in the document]
vogel: This is not the case when we execute setSigner and setThresholds separately. It can be put in better commit after rebasing.

src/test/TxTests.cpp
```diff
@@ -2,14 +2,13 @@
 // under the Apache License, Version 2.0. See the COPYING file at the root
 // of this distribution or at http://www.apache.org/licenses/LICENSE-2.0
 
-#include "TxTests.h"
-
+#include "test/TxTests.h"
```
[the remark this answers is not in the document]
vogel: Yes, but in most cases we use full path, so I added that for consistency.

for example: HerderImpl.cpp:

```
#include "herder/HerderPersistence.h"
#include "herder/HerderUtils.h"
#include "herder/LedgerCloseData.h"
#include "herder/TxSetFrame.h"
```

src/test/TxTests.cpp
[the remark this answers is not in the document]
MonsieurNicolas: cool, instead of disabling the check like you did in a later commit, just add the proper condition in the code: most operations should fail the same way. I can see how certain early failures (transaction level, like auth) could fail differently - but those are the exceptions to the rule :)

src/test/TestAccount.cpp
```diff
@@ -158,13 +158,9 @@ TestAccount::hasTrustLine(Asset const& asset) const
 }
 
 void
-TestAccount::setOptions(AccountID* inflationDest, uint32_t* setFlags,
```
MonsieurNicolas: looks very nice now :)

src/test/TxTests.cpp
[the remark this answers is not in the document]
MonsieurNicolas: ok, that's fine

src/test/TxTests.cpp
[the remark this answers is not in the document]
MonsieurNicolas: FYI - this still needs to be changed (and not blanket exempt protocol versions less than 10)

src/transactions/TransactionFrame.cpp
```diff
@@ -286,6 +286,33 @@ TransactionFrame::processSeqNum(LedgerManager& lm, LedgerDelta& delta)
...
+            app.getMetrics()
+                .NewMeter({"transaction", "invalid", "invalid-op"},
+                          "transaction")
+                .Mark();
+            markResultFailed();
+            return false;
+        }
+    }
+
+    removeUsedOneTimeSignerKeys(signatureChecker, delta,
```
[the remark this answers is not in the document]
vogel: I've implemented this change. It changes one test case to behave as before - https://github.com/stellar/stellar-core/pull/1709/commits/1a9cbd7fbb336943021abd7209b4c9a85ce59a65#diff-a4dc9da0d27fe7b8c7fcbaa5ac1bf72cL455.

Do you think that this behavior is proper one? (one time signature is no longer removed then transaction results in txBAD_AUTH_EXTRA).

And this is similar to checkValid, but it calls checkSignature on operation instead of checkValid. Not sure if this can be easily extracted.

vogel:
I think that the current version is mergeable.

vogel:
All required changes applied.

I think that now it is possible to block pre-signed transaction very easy. If bad actor wants to disable any pre-signed transaction, he/she can just have one prepared with extra signature and post it as soon as proper sequence number is achieved in source account. Then sequence number will be incremented and pre-signed transaction will became inapplicable to anyone.

Do I understand it correctly?

src/transactions/TransactionFrame.cpp
```diff
@@ -304,12 +305,29 @@ TransactionFrame::processSignatures(SignatureChecker& signatureChecker,
                           "transaction")
                 .Mark();
             markResultFailed();
-            return false;
+            allValid = false;
```
MonsieurNicolas: We should not mark those metrics several times for a given transaction.
I think it's better to move the code contained here into the `if (!allValid)` block below (and only use this place to set `allValid`)

src/transactions/TransactionFrame.cpp
```diff
@@ -310,6 +310,17 @@ TransactionFrame::processSignatures(SignatureChecker& signatureChecker,
...
                                 app.getLedgerManager());
+
+    if (!signatureChecker.checkAllSignaturesUsed())
+    {
+        getResult().result.code(txBAD_AUTH_EXTRA);
+        app.getMetrics()
+            .NewMeter({"transaction", "invalid", "bad-auth-extra"},
+                      "transaction")
+            .Mark();
+        return false;
```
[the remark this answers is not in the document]
MonsieurNicolas: I didn't realize that `removeUsedOneTimeSignerKeys` relies on `SignatureChecker` - your implementation does what I wanted though

src/transactions/TransactionFrame.cpp
```diff
@@ -286,6 +286,46 @@ TransactionFrame::processSeqNum(LedgerManager& lm, LedgerDelta& delta)
...
+bool
+TransactionFrame::processSignatures(SignatureChecker& signatureChecker,
+                                    Application& app, LedgerDelta& delta)
+{
+    if (app.getLedgerManager().getCurrentLedgerVersion() < 10)
+    {
+        return true;
+    }
+
+    auto allOpsValid = std::all_of(
```
MonsieurNicolas: this implementation has a problem (compared with the version you had before that was a loop):
if the one time signature is used in an operation after a failed auth operation, it will not be marked as "used", and will stick around (making `removeUsedOneTimeSignerKeys` no-op).

In order to preserve that property, you could use instead `std::accumulate` but given what we're doing, it's actually shorter and more readable to just use a for loop like before.

MonsieurNicolas:
looks like you introduced a subtle bug with the latest change (we may want to add a test for that edge case) @vogel

src/transactions/TransactionFrame.cpp
```diff
@@ -286,6 +286,46 @@ TransactionFrame::processSeqNum(LedgerManager& lm, LedgerDelta& delta)
...
+        std::begin(mOperations), std::end(mOperations),
+        [&signatureChecker, &app](auto const& op) {
+            return op->checkSignature(signatureChecker, app, nullptr);
+        });
+
+    removeUsedOneTimeSignerKeys(signatureChecker, delta,
```
MonsieurNicolas: another problem that @jonjove found: `removeUsedOneTimeSignerKeys` must use the `delta` version of `LoadAccount` (master is buggy as well but bug has more potential implications with this change)

vogel: This should be changed for all versions or only for versions >= 10? I'm not sure if this has any impact on generated buckets?

MonsieurNicolas:
r+ 2a88584716bec4f7d648ea1eaf042be72fd6a06f

PR #1711, opened 2018-06-12
Reverse ledger chain verification

marta-lokhova:
In this PR, ledger chain verification is updated to verify checkpoint starting from the recent one and working its way back to the beginning of the range. Additionally, ledger chain verification work does not make calls to ledger manager anymore, as the trusted SCP hash is passed into CatchupWork directly.


src/catchup/CatchupWork.cpp
```diff
@@ -25,15 +25,15 @@ namespace stellar
 
 CatchupWork::CatchupWork(Application& app, WorkParent& parent,
                          CatchupConfiguration catchupConfiguration,
-                         bool manualCatchup, ProgressHandler progressHandler,
-                         size_t maxRetries)
+                         ProgressHandler progressHandler, size_t maxRetries,
+                         optional<Hash> trustedScpHash)
```
vogel: maxRetries is last argument for constructors of all Work items, could you please leave it last here as well for consistency?

src/history/HistoryTests.cpp
```diff
@@ -554,7 +555,7 @@ TEST_CASE("Catchup recent", "[history][catchuprecent]")
 
     for (auto a : apps)
     {
-        catchupSimulation.catchupApplication(initLedger, 80, false, a);
+        REQUIRE(catchupSimulation.catchupApplication(initLedger, 80, false, a));
```
vogel: nice catch!

src/history/HistoryTests.cpp
```diff
@@ -609,11 +610,14 @@ TEST_CASE("Catchup manual", "[history][catchupmanual]")
                 auto a = catchupSimulation.catchupNewApplication(
                     configuration.toLedger(), configuration.count(), true,
                     dbMode, name);
+                REQUIRE(a);
                 // manual catchup-complete to first checkpoint
-                catchupSimulation.catchupApplication(
-                    initLedger1, std::numeric_limits<uint32_t>::max(), true, a);
+                REQUIRE(catchupSimulation.catchupApplication(
+                    initLedger1, std::numeric_limits<uint32_t>::max(), true,
+                    a));
```
vogel: I think 'a' variable is not longer necessary here. You can just do REQUIRE on line 610 and use plain 'true' on 617 and 620.

src/catchup/VerifyLedgerChainWork.cpp
```diff
@@ -94,46 +95,41 @@ VerifyLedgerChainWork::getStatus() const
...
 VerifyLedgerChainWork::onReset()
 {
-    auto setLedger = mApp.getLedgerManager().getLastClosedLedgerHeader();
-    if (setLedger.header.ledgerSeq < 2)
-    {
-        setLedger = {};
-    }
-    if (mFirstVerified.header.ledgerSeq != 0)
-    {
-        mFirstVerified = setLedger;
```
vogel: Is it save to not reset mFirstVerified here? I think so, as mFirstVerified is verified at the end of this work, so the work cannot fail after that. If my analysis is correct, we should have a comment here explaining that.

src/catchup/VerifyLedgerChainWork.cpp
```diff
@@ -197,56 +207,91 @@ VerifyLedgerChainWork::verifyHistoryOfSingleCheckpoint()
...
     mApp.getCatchupManager().logAndUpdateCatchupStatus(true);
 
-    if (mCurrCheckpoint >
-        mApp.getHistoryManager().checkpointContainingLedger(mRange.last()))
+    if (mCurrCheckpoint <
+        mApp.getHistoryManager().checkpointContainingLedger(mRange.first()))
     {
-        throw std::runtime_error("Verification overshot target ledger");
+        throw std::runtime_error(
+            "Verification overshot first ledger in the range.");
```
vogel: overshot -> undershot

src/history/HistoryTests.cpp
vogel: ok, for some reason I was expecting this to be bool.

src/catchup/CatchupWork.cpp
```diff
@@ -172,8 +174,9 @@ CatchupWork::verifyLedgers(LedgerRange const& range)
     CLOG(INFO, "History")
         << "Catchup verifying ledger chain for checkpointRange ["
         << range.first() << ".." << range.last() << "]";
-    mVerifyLedgersWork = addWork<VerifyLedgerChainWork>(
-        *mDownloadDir, range, mManualCatchup, mFirstVerified, mLastVerified);
+    mVerifyLedgersWork =
+        addWork<VerifyLedgerChainWork>(*mDownloadDir, range, mFirstVerified,
```
MonsieurNicolas: we should probably use this occasion to get rid of this "pass by reference" that is quite ugly (`mFirstVerified`) and set it when `mVerifyLedgersWork` succeeds with something like `mFirstVerified  = mVerifyLedgersWork->getFirstVerified()`.
Also, maybe rename it to something a bit less cryptic - I don't know what "first verified" means especially now that we verify backwards. Maybe `mVerifiedRangeStart` ? (also rename the corresponding member in `VerifyLedgerChainWork`)

src/catchup/VerifyLedgerChainWork.cpp
```diff
@@ -94,46 +95,48 @@ VerifyLedgerChainWork::getStatus() const
...
-    if (setLedger.header.ledgerSeq < 2)
-    {
-        setLedger = {};
-    }
-    if (mFirstVerified.header.ledgerSeq != 0)
-    {
-        mFirstVerified = setLedger;
-    }
-    if (mLastVerified.header.ledgerSeq != 0)
+    // Note that mFirstVerified is not reset here, since it is set
```
MonsieurNicolas: Instead of this comment, I think it would be much cleaner to always reset `mFirstVerified` to the zero value.

src/catchup/VerifyLedgerChainWork.cpp
```diff
@@ -94,46 +95,48 @@ VerifyLedgerChainWork::getStatus() const
...
+    // in the end of verification, where work wouldn't fail
+
+    if (mVerifiedAhead.header.ledgerSeq != 0)
```
MonsieurNicolas: why the condition on resetting `mVerifiedAhead` ?

src/catchup/VerifyLedgerChainWork.cpp
```diff
@@ -197,56 +229,98 @@ VerifyLedgerChainWork::verifyHistoryOfSingleCheckpoint()
...
     }
 
-    auto status = HistoryManager::VERIFY_STATUS_OK;
     if (curr.header.ledgerSeq == mRange.last())
     {
-        CLOG(INFO, "History") << "Verifying catchup candidate "
-                              << curr.header.ledgerSeq << " with LedgerManager";
-        status = mApp.getLedgerManager().verifyCatchupCandidate(curr,
-                                                                mManualCatchup);
+        // Verifying the most recent checkpoint means no other checkpoints are
```
MonsieurNicolas: I think that we should refactor those two ifs: there is no reason to distinguish the "trusted ledger from SCP" and the "trusted ledger from the next checkpoint":
I would just set `nextCheckpointFirstLedger` to either the trusted value from SCP (if we're verifying the last checkpoint) or to the value of `mVerifiedAhead`

src/catchup/VerifyLedgerChainWork.h
```diff
@@ -24,28 +24,36 @@ class VerifyLedgerChainWork : public Work
...
+    // First ledger of last verified checkpoint. Needed for a checkpoint that
+    // is being verified: last ledger in current checkpoint must agree with
+    // mVerifiedAhead
+    LedgerHeaderHistoryEntry mVerifiedAhead{};
 
     medida::Meter& mVerifyLedgerSuccessOld;
     medida::Meter& mVerifyLedgerSuccess;
     medida::Meter& mVerifyLedgerFailureLedgerVersion;
     medida::Meter& mVerifyLedgerFailureOvershot;
+    medida::Meter& mVerifyLedgerFailureUndershot;
```
MonsieurNicolas: I don't think we care to have that kind of granularity for metrics

src/catchup/VerifyLedgerChainWork.cpp
```diff
@@ -197,56 +229,98 @@ VerifyLedgerChainWork::verifyHistoryOfSingleCheckpoint()
...
+    assert(ledger.header.ledgerSeq == mRange.last());
+    if (!mTrustedHash)
+    {
+        CLOG(INFO, "History")
+            << "No trusted hash provided to verify "
+            << LedgerManager::ledgerAbbrev(ledger) << " against.";
+    }
+    else
+    {
+        CLOG(INFO, "History")
```
MonsieurNicolas: this should be `DEBUG` not `INFO`

src/catchup/VerifyLedgerChainWork.cpp
```diff
@@ -197,56 +229,98 @@ VerifyLedgerChainWork::verifyHistoryOfSingleCheckpoint()
...
+}
+
+HistoryManager::LedgerVerificationStatus
+VerifyLedgerChainWork::verifyAgainstTrustedHash(
+    LedgerHeaderHistoryEntry& ledger)
+{
+    assert(ledger.header.ledgerSeq == mRange.last());
+    if (!mTrustedHash)
+    {
+        CLOG(INFO, "History")
```
MonsieurNicolas: this should be `DEBUG` not `INFO`

src/ledger/LedgerManager.h
```diff
@@ -160,7 +160,8 @@ class LedgerManager
     // This method is present in the public interface to permit testing and
     // command line catchups.
     virtual void startCatchup(CatchupConfiguration configuration,
-                              bool manualCatchup) = 0;
+                              bool manualCatchup,
+                              optional<Hash> trustedHash) = 0;
```
MonsieurNicolas: similar comment than for `catchupHistory` regarding `trustedHash`

src/catchup/CatchupManager.h
```diff
@@ -30,8 +30,8 @@ class CatchupManager
 
     // Run catchup with given configuration and verify mode.
     virtual void catchupHistory(CatchupConfiguration catchupConfiguration,
-                                bool manualCatchup,
-                                CatchupWork::ProgressHandler handler) = 0;
+                                CatchupWork::ProgressHandler handler,
+                                optional<Hash> trustedHash) = 0;
```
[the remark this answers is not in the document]
marta-lokhova: Good idea, I think it would definitely make more sense to have ledger hash in `CatchupConfiguration`. Will update.

src/catchup/CatchupWork.cpp
```diff
@@ -129,14 +130,15 @@ CatchupWork::hasAnyLedgersToCatchupTo() const
...
-    if (mLastClosedLedgerAtReset <= mRemoteState.currentLedger)
+    if (mLastClosedLedgerAtReset.header.ledgerSeq <= mRemoteState.currentLedger)
     {
         return true;
     }
 
     CLOG(INFO, "History")
         << "Last closed ledger is later than current checkpoint: "
-        << mLastClosedLedgerAtReset << " > " << mRemoteState.currentLedger;
+        << mLastClosedLedgerAtReset.header.ledgerSeq << " > "
```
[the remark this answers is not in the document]
marta-lokhova: Is there more information on the bug? (a github issue maybe?)

Regarding retrying with a different archive, yes, that would be good. In order for this to happen, we first need to ensure that a single archive is used during catchup (not a guarantee right now, archive is selected randomly each download).

MonsieurNicolas:
needs to be rebased

MonsieurNicolas:
I guess there were comments, maybe it's easier to first address comments before rebasing

src/catchup/VerifyLedgerChainWork.h
marta-lokhova: Updated this, and kept "ledger-verify" success and failure metrics only.

src/catchup/VerifyLedgerChainWork.cpp
marta-lokhova: Done.

src/catchup/VerifyLedgerChainWork.cpp
marta-lokhova: This was in the original code, looks like it's unnecessary, so I removed it.

marta-lokhova:
FYI, this is rebased and comments are addressed.

MonsieurNicolas:
needs to be rebased again @marta-lokhova

marta-lokhova:
Rebased and updated. A few things:
* I added better handling of last closed ledger to the verification logic
* I added some ledger header generation infrastructure for testing as well as a verification test.

src/catchup/VerifyLedgerChainWork.cpp
```diff
@@ -150,25 +151,35 @@ VerifyLedgerChainWork::verifyHistoryOfSingleCheckpoint()
...
                     << " does not agree with LCL "
                     << LedgerManager::ledgerAbbrev(mLastClosed.first,
                                                    *mLastClosed.second);
-                mVerifyLedgerChainFailure.Mark();
                 return HistoryManager::VERIFY_STATUS_ERR_BAD_HASH;
             }
         }
+        // Verify LCL that is just before the first ledger in range
+        else if (curr.header.ledgerSeq == mLastClosed.first + 1 &&
+                 curr.header.ledgerSeq ==
```
MonsieurNicolas: I don't think you need the second condition: if `curr.header.ledgerSeq == mLastClosed.first + 1` -> the hashes must agree.

src/catchup/VerifyLedgerChainWork.cpp
```diff
@@ -150,25 +151,35 @@ VerifyLedgerChainWork::verifyHistoryOfSingleCheckpoint()
...
+                     mApp.getHistoryManager().prevCheckpointLedger(
+                         mRange.first()))
+        {
+            auto lclResult = verifyLedgerHistoryLink(*mLastClosed.second, curr);
+            if (lclResult != HistoryManager::VERIFY_STATUS_OK)
+            {
+                return lclResult;
```
MonsieurNicolas: need logging for this code path

src/catchup/VerifyLedgerChainWork.cpp
marta-lokhova: Added the second condition to avoid potentially checking the same link twice (if LCL is inside the range, it'd be verified later anyway), but it can be removed since the overhead is likely negligible.

MonsieurNicolas: yeah I agree, it just makes the code harder to read to have this extra condition

marta-lokhova:
@MonsieurNicolas, addressed your comments (the PR is squashed and rebased now too)

MonsieurNicolas:
r+ 6274c089f7ec335274139250fff310ee4edc97aa

PR #1718, opened 2018-06-25
Asset backed offers

jonjove:
Implements cap-0003, which has been discussed in https://github.com/stellar/stellar-protocol/issues/36. Note that this is a preliminary implementation and does not yet contain all necessary tests, so it should not be merged.


src/transactions/OfferExchange.cpp
```diff
@@ -887,14 +891,53 @@ OfferExchange::crossOfferV10(OfferFrame& sellingWheatOffer,
             TrustFrame::loadTrustLine(accountBID, sheep, db, mDelta);
     }
 
+    Liabilities prevLiabilities;
```
MonsieurNicolas: you don't need to have `prevLiabilities` defined in this scope, in fact I don't think you need this variable at all.
Also a comment that explains what is going on would help, something in the line of "free up the liabilities of this offer while we cross it"

src/transactions/OfferExchange.cpp
```diff
@@ -967,6 +1011,38 @@ OfferExchange::crossOfferV10(OfferFrame& sellingWheatOffer,
         sellingWheatOffer.storeChange(mDelta, db);
     }
 
+    if (mLedgerManager.getCurrentLedgerVersion() >= 11)
+    {
+        int64_t deltaBuying = sellingWheatOffer.getBuyingLiabilities();
+        if (sellingWheatOffer.getBuying().type() == ASSET_TYPE_NATIVE)
+        {
+            bool res = accountB->addBuyingLiabilities(deltaBuying);
+            assert(res);
```
MonsieurNicolas: same thing with those `assert` calls - I know it should not happen, but better safe than sorry and throw instead.

src/transactions/OfferExchange.cpp
```diff
@@ -967,6 +1011,38 @@ OfferExchange::crossOfferV10(OfferFrame& sellingWheatOffer,
         sellingWheatOffer.storeChange(mDelta, db);
     }
 
+    if (mLedgerManager.getCurrentLedgerVersion() >= 11)
```
MonsieurNicolas: should that be in the `else` above?

src/transactions/TransactionFrame.cpp
```diff
@@ -341,8 +341,16 @@ TransactionFrame::commonValid(SignatureChecker& signatureChecker,
             ? mSigningAccount->getAccount().balance
             : mSigningAccount->getAccount().balance - mEnvelope.tx.fee;
 
-    // don't let the account go below the reserve
-    if (balanceAfter < mSigningAccount->getMinimumBalance(lm))
+    // Have to account for liabilities once they are available
+    int64_t minBalance = mSigningAccount->getMinimumBalance(lm);
```
MonsieurNicolas: maybe use `getBalanceAboveReserve` instead?

src/herder/Upgrades.cpp
```diff
@@ -318,4 +328,237 @@ Upgrades::timeForUpgrade(uint64_t time) const
...
+Upgrades::dropAll(Database& db)
+{
+    db.getSession() << "DROP TABLE IF EXISTS upgradehistory";
+    db.getSession() << "CREATE TABLE upgradehistory ("
+                       "ledgerseq    INT NOT NULL CHECK (ledgerseq >= 0), "
+                       "upgradeindex INT NOT NULL, "
+                       "upgrade      TEXT NOT NULL, "
+                       "changes      TEXT NOT NULL, "
+                       "PRIMARY KEY (ledgerseq, upgradeindex)"
+                       ")";
```
MonsieurNicolas: you're missing:
* the SQL schema upgrade code (it's in a different commit) and updates to `db-schema.md`
* maintenance implementation (see `LedgerManagerImpl::deleteOldEntries`)
* an index on `ledgerseq` (for maintenance to work)

src/herder/Upgrades.cpp
```diff
@@ -318,4 +328,237 @@ Upgrades::timeForUpgrade(uint64_t time) const
...
+static void
+prepareLiabilities(LedgerManager& ledgerManager, LedgerDelta& ld)
+{
+    auto& db = ledgerManager.getDatabase();
+    db.getEntryCache().clear();
+    auto offersByAccount = OfferFrame::loadAllOffers(db);
+
+    for (auto& accountOffers : offersByAccount)
+    {
+        std::map<Asset, std::unique_ptr<int64_t>> initialBuyingLiabilities;
```
MonsieurNicolas: add a comment on why it's a `unique_ptr`

src/herder/Upgrades.cpp
```diff
@@ -318,4 +328,237 @@ Upgrades::timeForUpgrade(uint64_t time) const
...
+                  int64_t balanceAboveReserve,
+                  std::map<Asset, std::unique_ptr<int64_t>> const& liabilities,
+                  std::function<int64_t(AccountID const&, Asset const&, int64_t,
+                                        LedgerDelta&, Database&)>
+                      getCap,
+                  LedgerDelta& ld, Database& db)
+{
+    auto iter = liabilities.find(asset);
+    assert(iter != liabilities.end());
+    return iter->second ? *iter->second > getCap(accountID, asset,
```
MonsieurNicolas: comment here that explains that overflow -> delete

src/herder/Upgrades.cpp
```diff
@@ -318,4 +328,237 @@ Upgrades::timeForUpgrade(uint64_t time) const
...
+            }
+        }
+
+        for (auto const& assetLiabilities : liabilities)
+        {
+            Asset const& asset = assetLiabilities.first;
+            Liabilities const& liab = assetLiabilities.second;
+            if (asset.type() == ASSET_TYPE_NATIVE)
+            {
+                account.ext.v(1);
```
MonsieurNicolas: call `addBuyingLiabilities` and `addSellingLiabilities` to avoid messing with the `ext.v1()` syntax outside of `AccountFrame`

MonsieurNicolas: same for trustline below

src/herder/Upgrades.cpp
```diff
@@ -318,4 +328,237 @@ Upgrades::timeForUpgrade(uint64_t time) const
...
+        auto& account = accountFrame->getAccount();
+        assert(accountFrame);
+        int64_t balanceAboveReserve =
+            accountFrame->getBalanceAboveReserve(ledgerManager);
+
+        std::map<Asset, Liabilities> liabilities;
+        for (auto const& offerFrame : accountOffers.second)
+        {
+            auto const& offer = offerFrame->getOffer();
+            bool erase = shouldDeleteOffer(
```
MonsieurNicolas: It would be cleaner to use `std::bind` with placeholders instead of having a bunch of parameters passed in to `shouldDeleteOffer` for the sole purpose of the `getBalance` function.
(better would have been to use lambdas and captures, but that function is already big enough)

src/database/Database.cpp
```diff
@@ -140,9 +141,23 @@ Database::applySchemaUpgrade(unsigned long vers)
             }
         }
         break;
+
     case 6:
         mSession << "ALTER TABLE peers ADD flags INT NOT NULL DEFAULT 0";
         break;
+
+    case 7:
+        Upgrades::dropAll(*this);
```
MonsieurNicolas: better is to do those changes in the commits that modify the schema (only exception is SCHEMA_VERSION that is modified in the first of those commits)

src/invariant/LiabilitiesMatchOffers.cpp
```diff
@@ -0,0 +1,249 @@
...
+LiabilitiesMatchOffers::checkBalanceAndLimit(IterType const& iter,
+                                             uint32_t ledgerVersion) const
+{
+    auto const& current = iter->current->mEntry;
+    if (current.data.type() == ACCOUNT)
+    {
+        if (shouldCheckAccount(*iter, ledgerVersion))
+        {
+            auto const& account = current.data.account();
+            Liabilities liabilities = (ledgerVersion >= 11)
```
MonsieurNicolas: instead of using `ledgerVersion`, you can simply use `account.ext.v() == 1`

src/invariant/LiabilitiesMatchOffers.cpp
```diff
@@ -0,0 +1,249 @@
...
+                    "Balance not compatible with liabilities for account {}",
+                    xdr::xdr_to_string(account));
+            }
+        }
+    }
+    else if (current.data.type() == TRUSTLINE)
+    {
+        auto const& trust = current.data.trustLine();
+        Liabilities liabilities =
+            (ledgerVersion >= 11) ? trust.ext.v1().liabilities : Liabilities{};
```
MonsieurNicolas: same here, better to use `trust.ext.v()`

src/invariant/LiabilitiesMatchOffers.cpp
```diff
@@ -0,0 +1,249 @@
+// Copyright 2018 Stellar Development Foundation and contributors. Licensed
```
MonsieurNicolas: new/deleted files -> need to update visual studio project file

src/test/test.cpp
```diff
@@ -88,7 +88,7 @@ getTestConfig(int instanceNumber, Config::TestDbMode mode)
                                        "CacheIsConsistentWithDatabase",
                                        "ConservationOfLumens",
                                        "LedgerEntryIsValid",
-                                       "MinimumAccountBalance"};
+                                       "LiabilitiesMatchOffers"};
```
MonsieurNicolas: forgot to update the example config (documentation for invariants)

src/invariant/CacheIsConsistentWithDatabase.cpp
```diff
@@ -16,11 +17,13 @@ std::shared_ptr<Invariant>
 CacheIsConsistentWithDatabase::registerInvariant(Application& app)
 {
     return app.getInvariantManager()
-        .registerInvariant<CacheIsConsistentWithDatabase>(app.getDatabase());
+        .registerInvariant<CacheIsConsistentWithDatabase>(
```
vogel: To think about: is it better to pass just the objects that we need or just give up and pass whole Application everywhere (as it seems that over time more and more objects are required in multiple places)?

src/invariant/CacheIsConsistentWithDatabase.cpp
```diff
@@ -16,11 +17,13 @@ std::shared_ptr<Invariant>
...
+            app.getDatabase(), app.getLedgerManager());
 }
 
-CacheIsConsistentWithDatabase::CacheIsConsistentWithDatabase(Database& db)
-    : Invariant(false), mDb{db}
+CacheIsConsistentWithDatabase::CacheIsConsistentWithDatabase(Database& db,
+                                                             LedgerManager& lm)
+    : Invariant(false), mDb{db}, mLedgerManager(lm)
```
vogel: Note: false here is cryptic, should be replaced with enum when that makes sense.

src/herder/Upgrades.cpp
```diff
@@ -318,4 +328,237 @@ Upgrades::timeForUpgrade(uint64_t time) const
...
+{
+    auto iter = liabilities.find(asset);
+    assert(iter != liabilities.end());
+    return iter->second ? *iter->second > getCap(accountID, asset,
+                                                 balanceAboveReserve, ld, db)
+                        : true;
+}
+
+static void
+prepareLiabilities(LedgerManager& ledgerManager, LedgerDelta& ld)
```
vogel: I'm not sure if I understand this method correctly, please let me know if this is correct:

For each account, remove all offers that are using assets, for which the liabilities are too big. So it does not keep any offers from that set, even if keeping only some of them would make all the liabilities proper?

src/herder/UpgradesTests.cpp
```diff
@@ -557,6 +558,464 @@ TEST_CASE("Ledger Manager applies upgrades properly", "[upgrades]")
...
+            REQUIRE(getAssetLiabilities(a1, cur2) == Liabilities{4000, 4000});
+            REQUIRE(getLiabilities(a2) == Liabilities{4000, 4000});
+            REQUIRE(getAssetLiabilities(a2, cur1) == Liabilities{4000, 4000});
+            REQUIRE(getAssetLiabilities(a2, cur2) == Liabilities{8000, 0});
+        }
+    }
+
+    SECTION("liabilities overflow")
+    {
+        auto createOfferSmall =
```
vogel: seems not to be used

src/transactions/OfferTests.cpp
```diff
@@ -621,6 +650,14 @@ TEST_CASE("create offer", "[tx][offers]")
                             }),
                         ex_MANAGE_OFFER_CROSS_SELF);
                 });
+
+                for_versions_to(11, *app, [&] {
+                    issuer.pay(a1, usd, 20000);
+
+                    // ensure we could receive proceeds from the offer
```
vogel: probably invalid comment

src/transactions/OfferTests.cpp
```diff
@@ -1039,6 +1170,12 @@ TEST_CASE("create offer", "[tx][offers]")
                          {b1, {{usd, 150}, {idr, trustLineBalance - 100}}},
                          {c1, {{usd, trustLineBalance - 225}, {idr, 150}}}});
                 });
+
+                for_versions_from(11, *app, [&] {
+                    // makes "A" only capable of holding 75 "USD"
```
vogel: invalid comment

src/transactions/OfferTests.cpp
```diff
@@ -572,6 +590,17 @@ TEST_CASE("create offer", "[tx][offers]")
                         auto actualPayment = a1IDrs - delta;
                         checkCrossed(b1, actualPayment, offerAmount);
                     });
+
+                    for_versions_from(10, *app, [&]() {
+                        REQUIRE_THROWS_AS(
```
MonsieurNicolas: this is not what this test is trying to test:
test ensures that we compute reserve as if creating the offer.
I think that, with liabilities, there should be both a check that verifies that creating an offer with exactly `actualPayment` works, but trying to sell more (by 1 stroop) fails.

src/transactions/OfferTests.cpp
```diff
@@ -1096,38 +1235,38 @@ TEST_CASE("create offer", "[tx][offers]")
...
                     auto offerB1 = market.requireChangesWithOffer(
                         {{offerC1.key, OfferState::DELETED}}, [&] {
                             return market.addOffer(b1, offerB1Params,
                                                    offerRemaining);
                         });
                 });
+
+                for_versions_from(10, *app, [&]() {
+                    // make it that c1 can only receive 9 USD (for 1 IDR)
+                    REQUIRE_THROWS_AS(c1.changeTrust(usd, 9 * assetMultiplier),
```
MonsieurNicolas: same thing here: the way the setup is done is now impossible, probably clearer to not test this anymore like that and instead rely on `changeTrust` tests

src/transactions/OfferTests.cpp
```diff
@@ -1292,6 +1431,15 @@ TEST_CASE("create offer", "[tx][offers]")
                                           {xlm, usd, oneone, 110});
             });
         });
+        for_versions_from(10, *app, [&] {
+            auto offer = market.requireChangesWithOffer({}, [&] {
+                return market.addOffer(a1, {xlm, usd, oneone, 9});
+            });
+            REQUIRE_THROWS_AS(market.updateOffer(a1, offer.key.offerID,
+                                                 {xlm, usd, oneone, 111},
+                                                 {xlm, usd, oneone, 110}),
```
MonsieurNicolas: it's a bit confusing to have an expected state different as it's a failure case

src/transactions/OfferTests.cpp
```diff
@@ -1292,6 +1431,15 @@ TEST_CASE("create offer", "[tx][offers]")
                                           {xlm, usd, oneone, 110});
             });
         });
+        for_versions_from(10, *app, [&] {
+            auto offer = market.requireChangesWithOffer({}, [&] {
+                return market.addOffer(a1, {xlm, usd, oneone, 9});
```
MonsieurNicolas: I think we should change the amounts to be the ones around the edge case:
so use "10" instead of "9", and use "11" instead of "111"

src/transactions/PathPaymentTests.cpp
```diff
@@ -2725,40 +2802,9 @@ TEST_CASE("pathpayment", "[tx][pathpayment]")
...
-            // clang-format off
-            market.requireBalances(
-                {{source, {{xlm, minBalance4 - 2 * txfee}, {cur1, 0}, {cur2, 0}, {cur3, 0}, {cur4, 0}}},
-                 {mm12a, {{xlm, minBalance3 - 4 * txfee}, {cur1, 0}, {cur2, 1}, {cur3, 0}, {cur4, 0}}},
-                 {mm12b, {{xlm, minBalance3 - 3 * txfee}, {cur1, 80}, {cur2, 0}, {cur3, 0}, {cur4, 0}}},
-                 {mm23, {{xlm, minBalance3 - 3 * txfee}, {cur1, 0}, {cur2, 40}, {cur3, 0}, {cur4, 0}}},
-                 {mm34, {{xlm, minBalance3 - 3 * txfee}, {cur1, 0}, {cur2, 0}, {cur3, 20}, {cur4, 0}}},
-                 {destination, {{xlm, minBalance1 - txfee}, {cur1, 0}, {cur2, 0}, {cur3, 0}, {cur4, 10}}}});
-            // clang-format on
+            REQUIRE_THROWS_AS(mm12a.pay(gateway, cur2, 39),
```
MonsieurNicolas: should those edge cases be tested differently?

src/transactions/TransactionFrame.cpp
```diff
@@ -352,6 +350,35 @@ TransactionFrame::commonValid(SignatureChecker& signatureChecker,
     }
 
     return ValidationType::kFullyValid;
+
+    //// if we are in applying mode fee was already deduced from signing account
```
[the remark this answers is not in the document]
jonjove: This was commented out as I was updating this to use `getBalanceAboveReserve`. Fixed.

src/transactions/OfferTests.cpp
jonjove: Agreed.

src/transactions/OfferTests.cpp
jonjove: Done.

src/transactions/OfferTests.cpp
jonjove: Agreed, fixed.

src/transactions/PathPaymentTests.cpp
jonjove: These edge cases aren't possible anymore, since it is not possible to have an offer with excess liabilities. Let's not test them starting in version 10.

src/ledger/AccountFrame.cpp
```diff
@@ -118,10 +118,82 @@ AccountFrame::getBalance() const
     return (mAccountEntry.balance);
 }
 
+int64_t
+getBuyingLiabilities(AccountEntry const& acc, LedgerManager const& lm)
+{
+    assert(lm.getCurrentLedgerVersion() >= 10);
```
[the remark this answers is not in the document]
jonjove: We haven't followed this convention with other version numbers, and I'm not sure it would add much clarity. For example, do you think `assert(lm.getCurrentLedgerVersion() >= PROTOCOL_VERSION_10);` would be considerably more clear?

src/herder/UpgradesTests.cpp
```diff
@@ -557,6 +558,464 @@ TEST_CASE("Ledger Manager applies upgrades properly", "[upgrades]")
     }
 }
 
+TEST_CASE("upgrade to version 11", "[upgrades]")
+{
+    VirtualClock clock;
+    auto cfg = getTestConfig(0);
+    cfg.LEDGER_PROTOCOL_VERSION = 9;
+    auto app = Application::create(clock, cfg);
```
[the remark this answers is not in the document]
jonjove: Good idea. All uses of `Application::create` in this file changed to `createTestApplication`.

src/ledger/AccountFrame.cpp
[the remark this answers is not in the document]
marta-lokhova: The reason I suggested it:
1. You can easily change it. Previously you had 11, so instead of changing it to 10 in every place, you could just change one variable. 
2. you could name it something like `PROTOCOL_VERSION_WITH_LIABILITIES` and it would be way less vague than just `10`

src/transactions/OfferTests.cpp
```diff
@@ -776,6 +814,27 @@ TEST_CASE("create offer", "[tx][offers]")
...
+                    c1.changeTrust(idr, trustLineLimit);
+                    c1.changeTrust(usd, trustLineLimit);
+                    issuer.pay(c1, idr, 20000);
+
+                    // matches the offer from A
+                    auto cOffer = market.requireChangesWithOffer({}, [&] {
+                        return market.addOffer(c1, {idr, usd, price, 100});
+                    });
+                    // drain account
+                    REQUIRE_THROWS(c1.pay(issuer, idr, 20000));
```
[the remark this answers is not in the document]
MonsieurNicolas: I think this test is just a test that consumes whatever is in the order book and a new offer is created: so it's really "cross all offers on the book, new offer remains".
Compared to other test where a new offer is not created as the order book "wins".

src/ledger/AccountFrame.cpp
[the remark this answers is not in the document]
MonsieurNicolas: let's not do that in this PR: v10 is broader than liabilities.

src/transactions/OfferTests.cpp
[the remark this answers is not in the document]
jonjove: Renamed this test to "offer crosses, removes all offers, and remains" which should improve clarity and avoid future confusion. The only reason this test doesn't work in version 10 is the invalid offer created by account `c1`. I have retained this in versions before 10 and removed it starting in version 10, leaving the rest of the test intact.

src/ledger/LedgerManagerImpl.cpp
```diff
@@ -854,15 +855,43 @@ LedgerManagerImpl::closeLedger(LedgerCloseData const& ledgerData)
...
+            CLOG(ERROR, "Ledger") << "Unknown exception during upgrade";
+            getCurrentLedgerHeader() = previousHeader;
+        }
     }
+    // WARNING: At this point, LedgerManager.getCurrentLedgerHeader() may have
+    // been modified directly as part of the upgrade process. Nothing should be
+    // added between here and the following call to
+    //     ledgerDelta.commit();
 
+    getCurrentLedgerHeader() = initialHeader;
```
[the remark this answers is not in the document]
MonsieurNicolas: The issue is that this is functionally equivalent to removing the check completely: if a bug is introduced while processing transactions, it will go unnoticed because of that code.

If you're worried about multiple upgrades doing the wrong thing (which with the current implementation can happen), then we should just
* add a new `LedgerDelta` to wrap all upgrades (and perform the "override" right after the loop, but scoped to upgrades only).
* restore `ledgerManager.getCurrentLedgerHeader()` to its value saved right before performing upgrades

Also, you should add a comment in `Upgrades::applyVersionUpgrade` that
`    ledgerManager.getCurrentLedgerHeader().ledgerVersion = newVersion;`
is basically violating the rule of only using `ledgerDelta.getHeader()` (for future you when you rebase the data layer changes).

src/ledger/AccountFrame.h
```diff
@@ -23,6 +23,9 @@ namespace stellar
 class LedgerManager;
 class LedgerRange;
 
+int64_t getBuyingLiabilities(AccountEntry const& acc, LedgerManager const& lm);
```
vogel: Looks like versions of those for AccountEntry and TrustLineEntry could be merged as a template.

src/ledger/OfferFrame.cpp
```diff
@@ -600,4 +628,112 @@ OfferFrame::dropAll(Database& db)
...
+        {
+            acc = AccountFrame::loadAccount(delta, getSellerID(), db);
+            assert(acc);
+        }
+        assert(acc->getID() == getSellerID());
+        return acc;
+    };
+
+    auto loadTrustIfNecessaryAndValidate = [this, &delta, &db](
+                                               TrustFrame::pointer const& trust,
```
vogel: missing clang-format?

src/transactions/InflationOpFrame.cpp
```diff
@@ -108,7 +108,7 @@ InflationOpFrame::doApply(Application& app, LedgerDelta& delta,
             {
                 lcl.totalCoins += toDoleThisWinner;
             }
-            if (!winner->addBalance(toDoleThisWinner))
+            if (!winner->addBalance(toDoleThisWinner, ledgerManager))
             {
                 throw std::runtime_error(
```
vogel: This throw was added here as it was impossible for above condition even to get true.

But is it impossible now? Imagine an issuer that creates asset that requires authorization. It sends INT64_MAX of that asset to one account that has enough XLM to be a winner in inflation. Then that account can make an offer to sell this asset 1:1 for XLM, making its buying liabilities so big, that its addBalance returns 0. Then at the next inflation round all nodes will drop this transaction with txINTERNAL_ERROR, making inflation operations no-op for as long as they want.

I think we need another solution for handling that case now.

src/transactions/ManageOfferOpFrame.cpp
```diff
@@ -223,10 +248,39 @@ ManageOfferOpFrame::doApply(Application& app, LedgerDelta& delta,
...
+                    .NewMeter({"op-manage-offer", "invalid", "line-full"},
+                              "operation")
+                    .Mark();
+                innerResult().code(MANAGE_OFFER_LINE_FULL);
+                return false;
+            }
+
+            int64_t availableBalance =
+                (sheep.type() == ASSET_TYPE_NATIVE)
+                    ? mSourceAccount->getBalanceAboveReserve(ledgerManager)
```
vogel: I think we could change getBalanceAboveReserver name to getAvailableBalance for symmetry?

bartekn:
While incorporating this in Horizon, I wanted to check how this works in action. I cloned @jonjove's repo and `git checkout` the branch in this PR. Then I created a new standalone network (with master account `M`) and:
1. Created a new account `A`.
2. Created a trustline `M.USD` on `A` (`INT64_MAX` limit).
3. Sent 100 `M.USD` from `M` to `A`.
4. Created an offer on `A`: sell 100 XLM, buy 100 `M.USD`.

I expected the `buyingliabilities` and `sellingliabilities` on `A` (both account and trust line) to change, however they are equal `NULL`. I think this may be happening because both fields are unitialized (`NULL`) - maybe because `prepareLiabilities` is executed only on protocol upgrade step (?).

src/transactions/OfferTests.cpp
```diff
@@ -1620,4 +1688,136 @@ TEST_CASE("create offer", "[tx][offers]")
...
+                    app->getLedgerManager().getMinBalance(2) + txfee;
+                auto acc1 = root.create("acc1", minBalance);
+                auto market = TestMarket{*app};
+
+                acc1.changeTrust(usd, 2000);
+
+                // Test when no existing offers
+                root.pay(acc1, xlm, 500 + txfee);
+                REQUIRE_THROWS_AS(
+                    market.addOffer(acc1, {xlm, usd, Price{1, 1}, 501},
```
vogel: I think that addOffer API allows you to leave last parameter if it will be the same as second one, as it defaults to OfferState::SAME

src/transactions/TxResultsTests.cpp
```diff
@@ -797,4 +798,40 @@ TEST_CASE("txresults", "[tx][txresults]")
             });
         }
     }
+
+    SECTION("fees with liabilities")
+    {
+        auto acc = root.create("acc", lm.getMinBalance(1) + baseFee + 1000);
```
vogel: I think that edge case here is that if you added +1 to that balance, it would pass even for version 1. Maybe it is worth adding such test case for contrast?

src/herder/Upgrades.cpp
```diff
@@ -318,4 +329,348 @@ Upgrades::timeForUpgrade(uint64_t time) const
...
+//   1. Calculate total liabilities for each asset
+//   2. For every asset with excess buying liabilities according to (1), erase
+//      all offers buying that asset. For every asset with excess selling
+//      liabilities according to (1), erase all offers selling that asset.
+//   3. Update liabilities to reflect offers remaining in the book.
+// It is essential to note that the excess liabilities are determined only
+// using the initial result of step (1), so it does not matter what order the
+// offers are processed.
+static void
+prepareLiabilities(LedgerManager& ledgerManager, LedgerDelta& ld)
```
bartekn: Shouldn't this function also remove offers from accounts with `trustline.isAuthorized() == false`? `AllowTrust` operation removes them when revoking trust so it seems that leaving them is an invalid state.

src/herder/Upgrades.cpp
```diff
@@ -318,4 +329,348 @@ Upgrades::timeForUpgrade(uint64_t time) const
...
+                                             "during upgrade");
+                }
+            }
+            else
+            {
+                auto trustFrame = TrustFrame::loadTrustLine(accountOffers.first,
+                                                            asset, db, &ld);
+                int64_t deltaSelling =
+                    liab.selling -
+                    trustFrame->getSellingLiabilities(ledgerManager);
```
bartekn: This errors when there are orphaned offers:
```
Exception during upgrade: trustline does not exist [LedgerManagerImpl.cpp:880]
```

Repro:
1. Run standalone testnet (testnet passphrase).
2. Submit:
  ```
  AAAAAGL8HQvQkbK2HA3WVjRrKmjX00fG8sLI7m0ERwJW/AX3AAABkAAAAAAAAAABAAAAAAAAAAAAAAAEAAAAAAAAAAAAAAAADJNEgW9b6gnIKGtYeSq1SiCyMJLujoFBoDtnljGf65oAAAAXSHboAAAAAAEAAAAADJNEgW9b6gnIKGtYeSq1SiCyMJLujoFBoDtnljGf65oAAAAGAAAAAVVTRAAAAAAAYvwdC9CRsrYcDdZWNGsqaNfTR8bywsjubQRHAlb8BfcAAAAAO5rKAAAAAAEAAAAADJNEgW9b6gnIKGtYeSq1SiCyMJLujoFBoDtnljGf65oAAAADAAAAAAAAAAFVU0QAAAAAAGL8HQvQkbK2HA3WVjRrKmjX00fG8sLI7m0ERwJW/AX3AAAAAACYloAAAAABAAAAAQAAAAAAAAAAAAAAAQAAAAAMk0SBb1vqCcgoa1h5KrVKILIwku6OgUGgO2eWMZ/rmgAAAAYAAAABVVNEAAAAAABi/B0L0JGythwN1lY0aypo19NHxvLCyO5tBEcCVvwF9wAAAAAAAAAAAAAAAAAAAAJW/AX3AAAAQK6ngzT5yLjXBKPibogU0NBm9M+W+ROt2JcPiIe1mf9eGNGaWhj0DYyG3GkTzrN4Tzg+LpXhiEO4cu6gylK6bQkxn+uaAAAAQAADEbTeCGHZ3RQVmUOX2pM8J+lFr5cq1PjOQZtoxSy1jYTnrvoRHoc7z9eSj9tv8aU/9ufEYBUYawc/l6hLMAI=
  ```
3. Upgrade to V10.

src/ledger/AccountFrame.cpp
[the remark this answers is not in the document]
MonsieurNicolas: yeah like I said, not for this PR; and this would make the code even more complicated to understand as features depend on each other on a per protocol basis.

src/herder/Upgrades.cpp
```diff
@@ -380,4 +387,276 @@ Upgrades::deleteOldEntries(Database& db, uint32_t ledgerSeq, uint32_t count)
...
+                              uint32_t newReserve)
+{
+    LedgerHeader& header = ld.getHeader();
+    bool didReserveIncrease = newReserve > header.baseReserve;
+
+    header.baseReserve = newReserve;
+    ledgerManager.getCurrentLedgerHeader().baseReserve = newReserve;
+    if (header.ledgerVersion >= 10 && didReserveIncrease)
+    {
+        prepareLiabilities(ledgerManager, ld);
```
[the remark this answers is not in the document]
MonsieurNicolas: no, this implementation is better as it associates upgrade side effects to the appropriate upgrade.

src/transactions/InflationOpFrame.cpp
MonsieurNicolas: Good catch @vogel : I think you are correct, `toDoleThisWinner` should be capped by the amount that the account can receive instead of just being `bigDivide(amountToDole, w.mVotes, totalVotes, ROUND_DOWN)`.

jonjove: Good catch indeed. Fixed as described above.

src/ledger/LiabilitiesTests.cpp
```diff
@@ -505,7 +513,7 @@ TEST_CASE("liabilities", "[ledger][liabilities]")
             REQUIRE(tf->getSellingLiabilities(lm) == 0);
             if (res)
             {
-                REQUIRE(tf->getTrustLine().ext.v() == 1);
+                REQUIRE(tf->getTrustLine().ext.v() == (deltaLiabilities != 0));
```
MonsieurNicolas: while this is correct, I find the implicit conversion from `true` to `1` quite obscure when used like this.

src/transactions/OfferTests.cpp
jonjove: Good point, fixed on many lines.

src/ledger/AccountFrame.cpp
```diff
@@ -118,10 +118,115 @@ AccountFrame::getBalance() const
...
+    }
+    return res;
+}
+
 bool
-AccountFrame::addBalance(int64_t delta)
+AccountFrame::addBalance(int64_t delta, LedgerManager const& lm)
 {
-    return stellar::addBalance(mAccountEntry.balance, delta);
+    if (delta == 0 && lm.getCurrentLedgerVersion() >= 10)
```
vogel: why is currentLedgerVersion checked here?

src/ledger/OfferFrame.cpp
```diff
@@ -600,4 +628,112 @@ OfferFrame::dropAll(Database& db)
...
+                                        LedgerManager& ledgerManager)
+{
+    auto loadAccountIfNecessaryAndValidate = [this, &account, &delta, &db]() {
+        AccountFrame::pointer acc = account;
+        if (!acc)
+        {
+            acc = AccountFrame::loadAccount(delta, getSellerID(), db);
+            assert(acc);
+        }
+        assert(acc->getID() == getSellerID());
```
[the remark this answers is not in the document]
vogel: Right, I have missed that case.

src/ledger/AccountFrame.cpp
jonjove: Good question, I see now that it doesn't need to be checked. Fixed.

vogel:
LGTM, I'll merge when acceptance tests are working

vogel:
r+ 3c9dfc0f57058586d7e4c638921bab7ad5dd6f1c

vogel:
@latobarita: retry

vogel:
@latobarita: retry

Issue #1706, opened 2018-06-04
Potential race in *Work workflows when retrying/stopping work


MonsieurNicolas:
It looks like on retry (or destruction), the pattern followed is to typically call `clearChildren`.

Issue is that this does not actually stop children from running (references may still be held in callbacks), and may cause multiple instances of the same "Work" to run (potentially conflicting with each other).

There are a couple of challenges that need to be overcome here:
* there should be new sematics to stop or abort `Work` - such transition may be asynchronous in nature and may require additional state transitions in the various `Work` implementations (so that parents properly wait for all of their children to be done aborting work).
   * most `Work` implementations should probably handle "success" from a child work differently if they are themselves in aborted state (to avoid run away work to trigger more work for example).
* synchronous stopping such as in destructors, should "just work" and be safe
   * this is especially challenging as `RunCommandWork` uses `ProcessManager` and process termination is likely to stay asynchronous (ie, relies on an event posted on the main IOService).
   * It may be possible to "force kill" processes in this case to at least ensure that resources are freed relatively quickly
   * `RunCommandWork` (and others) may have to behave a little bit differently in that notification to parents should not occur (as parents may be destructed or partially destructed) 
   * as "zombie" processes may still run for a bit we may have to add additional logic before spawning new instances of related work to avoid conflicts (using unique folders per instance may be enough - but might be too error prone). For example: a new "catchup" work may trigger a bunch of child work that conflicts with old child work from a previous (aborted) run.

marta-lokhova:
Here are some thoughts/ideas on asynchronous `abort` for Work classes:

High level: when parent is notified of child’s failure, a few things must happen:
* Parent goes into "aborting" state and “Abort” signal is issued to all other children, which in turn issue the signal to their children and so on until everybody knows that they are being aborted.
* Parent does not proceed (i.e. does not scheduleAbort/Failure) until all children move to aborted (or completed, see explanation below) state 
* Once the above happens, parent is scheduled to either abort itself or fail, propagating failure state from the child that triggered abort 

### Implementation Details
* Two new states
    * WORK_ABORTING 
    * WORK_FAILURE_ABORTED [work finished aborting]
* `abort` and `scheduleAbort` methods
    * `scheduleAbort`
        * Similarly to `scheduleRun` or `scheduleComplete`, schedules the work to complete aborting 
[...]

PR #1729, opened 2018-07-11
Clean work abort

marta-lokhova:
Graceful abort implementation as described here: https://github.com/stellar/stellar-core/issues/1706
(includes aborting Work in both PENDING and RUNNING states)
Note that after these changes, ProcessManager will need to be updated to abort work first prior to shutting down.

### DIFF
--- a/src/process/ProcessManager.h
+++ b/src/process/ProcessManager.h
@@ -49,11 +49,12 @@ class ProcessManager : public std::enable_shared_from_this<ProcessManager>,
 {
   public:
     static std::shared_ptr<ProcessManager> create(Application& app);
-    virtual ProcessExitEvent runProcess(std::string const& cmdLine,
-                                        std::string outputFile = "") = 0;
+    virtual std::shared_ptr<ProcessExitEvent>
+    runProcess(std::string const& cmdLine, std::string outputFile = "") = 0;
     virtual size_t getNumRunningProcesses() = 0;
     virtual bool isShutdown() const = 0;
     virtual void shutdown() = 0;
+    virtual void shutdownProcess(std::shared_ptr<ProcessExitEvent> pe) = 0;
     virtual ~ProcessManager()
     {
     }
--- a/src/process/ProcessManagerImpl.cpp
+++ b/src/process/ProcessManagerImpl.cpp
@@ -107,16 +107,16 @@ ProcessManagerImpl::~ProcessManagerImpl()
         forceShutdown(impl);
     };
     // Use SIGKILL on any processes we already used SIGINT on
-    while (!mKillableImpls.empty())
+    while (!mKillable.empty())
     {
-        auto impl = std::move(mKillableImpls.front());
-        mKillableImpls.pop_front();
-        killProcess(*impl);
+        auto pe = std::move(mKillable.front());
+        mKillable.pop_front();
+        killProcess(*pe->mImpl);
     }
     // Use SIGKILL on any processes we haven't politely asked to exit yet
     for (auto& pair : mImpls)
     {
-        killProcess(*pair.second);
+        killProcess(*pair.second->mImpl);
     }
 }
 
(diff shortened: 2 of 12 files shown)


src/work/Work.cpp
```diff
@@ -249,25 +266,45 @@ Work::advance()
...
-    else if (anyChildFatalFailure())
+    else if (anyChildRaiseFailure())
     {
         CLOG(DEBUG, "Work") << "some of " << mChildren.size() << " children of "
-                            << getUniqueName() << " fatally failed, scheduling "
-                            << "fatal failure";
-        scheduleFatalFailure();
+                            << getUniqueName() << " failed, propagating "
+                            << "abort";
+        abort(WORK_COMPLETE_FAILURE);
```
vogel: this is basically 'abort all children and then retry', I'm not sure if naming is good here

vogel:
I'm not sure if that is 100% correct implementation.

I think that we should be able to, for example, abort `ApplyBucketsWork` during bucket application. Imagine that we are in a middle of the work and `ApplyBucketsWork` is in `WORK_RUNNING` state. Then, for some reason, its parent work gets aborted. It does not call `abort` on `ApplyBucketsWork` because of the state.

But it should be 100% possible to do that, as `ApplyBucketsWork` schedules new success after each good application.

Also I don't think 'Work::run' should only abort when it is in `PENDING` state. It should be able to do that even in `RUNNING` state, as being in that place suggests, that it is between chunks of work and not in the middle of one.

marta-lokhova:
Ok, I think it should be possible. My initial concern was that aborting work in `running` state might leave us in some weird state. But analyzing it more, I think if we can guarantee that work is aborted while in consistent state (e.g. previous chunk is successful, next chunk hasn't started, all connections/file/etc are closed/released), then we should be ok. (`onAbort` can be used for such checks). It will also make the code a bit nicer, and I think we could avoid the need to implement a timeout.

marta-lokhova:
Also, regarding work being in a weird state, there's extra complexity specifically for `RunCommandWork`, because if we want to abort it while it's running, that means we need some interaction with `ProcessManager`, since the process might be either in the process queue or already running. So while it should be possible to abort running work, it will be quite complex.

src/work/WorkManagerImpl.cpp
```diff
@@ -61,6 +61,13 @@ WorkManagerImpl::notify(std::string const& child)
         mApp.getMetrics().NewMeter({"work", "root", "failure"}, "unit").Mark();
         mChildren.erase(child);
     }
+    else if (i->second->getState() == Work::WORK_FAILURE_ABORTED)
```
MonsieurNicolas: Before reviewing this PR, I opened https://github.com/stellar/stellar-core/issues/1755 as I thought semantics were already not super clean and error prone, now that we have abort(ing), we really need to formalize well what is going on, otherwise we're going to run into very strange bugs.

Also, the semantics implied here from `onAbort` don't really seem to follow what was described in https://github.com/stellar/stellar-core/issues/1706 (or at least it's unclear that it can work if `onAbort` only triggers some work).

I would recommend going back to basics: describe a state machine, its transitions and when certain callbacks (`onXYZ`) get called. Right now the `mAborting` flag makes it hard to tell which state transitions are valid vs invalid (and what is supposed to happen).

The two ways to abort are:
* somebody wants to abort some work, this causes transitions to `WORK_ABORTING` to `WORK_ABORTED` ; any work that was in flight is now complete (and was aborted if needed).
* abort in preparation for a retry, something like "decision to retry" -> `WORK_ABORTING_FOR_RETRY` (aborting work) -> `WORK_PENDING` (reset) -> `WORK_RUNNING` (run) ...

src/process/ProcessManagerImpl.cpp
```diff
@@ -159,6 +159,41 @@ ProcessManagerImpl::shutdown()
     }
 }
 
+void
+ProcessManagerImpl::shutdownProcess(std::shared_ptr<ProcessExitEvent> pev)
```
MonsieurNicolas: this implementation of `shutdownProcess` is not desirable in the typical case (it's doing `kill -9 ...`):
we want "clean" shutdown by default in case processes start sub processes. The process hierarchy is typically something like `stellar-core -> bash -> aws_cli-> aws_cli_children`, a "force kill" of `bash` may leave the `aws_cli` process (and children) running.

You can keep this implementation under a "force" parameter, but the MVP may not need this (ie: the only place where we would need it is if we're implementing timeout for abort)

src/process/ProcessManagerImpl.h
```diff
@@ -27,14 +27,14 @@ class ProcessManagerImpl : public ProcessManager
     // Subprocesses will be removed asynchronously, hence the lock on
     // just this member
     std::recursive_mutex mImplsMutex;
-    std::map<int, std::shared_ptr<ProcessExitEvent::Impl>> mImpls;
+    std::map<int, std::shared_ptr<ProcessExitEvent>> mImpls;
```
MonsieurNicolas: rename this: it's not `mImpls` anymore

src/process/ProcessManager.h
```diff
@@ -49,11 +49,12 @@ class ProcessManager : public std::enable_shared_from_this<ProcessManager>,
 {
   public:
     static std::shared_ptr<ProcessManager> create(Application& app);
-    virtual ProcessExitEvent runProcess(std::string const& cmdLine,
-                                        std::string outputFile = "") = 0;
+    virtual std::shared_ptr<ProcessExitEvent>
```
MonsieurNicolas: if we move to using to `std::shared_ptr<ProcessExitEvent>` across the board, should we make `ProcessExitEvent` non-copyable/movable?

marta-lokhova:
Abort is implemented via the new work interface https://github.com/stellar/stellar-core/pull/1819

PR #1731, opened 2018-07-12
Snapshot based ledger header + tx results verification

marta-lokhova:
This PR makes some initial changes to how ledger chain verification is done (shouldn't be merged yet as it depends on https://github.com/stellar/stellar-core/pull/1701 and https://github.com/stellar/stellar-core/pull/1711). It implements the following:
* Instead of catchup downloading all headers then verifying them, there's new `DownloadVerifyLedgersWork` class that takes care of downloading and verifying the whole chain, and reports back to `CatchupWork`.
* To coordinate checkpoints, `DownloadVerifyLedgersWork` uses batches of `LedgerChainVerificationSnapshotWork` - each snapshot implements download + verify and reports back to `DownloadVerifyLedgersWork`. This approach allows us to retry properly (e.g. if verify fails, re-download the file instead verifying it again)
* `LedgerChainVerificationSnapshotWork`  extends `BatchableWork`: a class that allows registering and notifying dependent work. This is useful during chain verification, where we need to wait for a previous checkpoint to finish before proceeding with verification of a current checkpoint (while waiting, we could at least download the file)

* Similarly, new classes for transaction result verification are introduced. These are a bit simpler because snapshots can run independently without being blocked on other snapshots.
* Currently tx results are verified per single ledger (https://github.com/stellar/stellar-core/blob/master/docs/history.md claims that these files can get pretty big so I wanted to make sure we won't monopolize the resources). If that is too granular, it should fairly simple to make a change to do it per checkpoint, or in chunks that are >1 ledger. 

* Tests require some re-factoring and will be updated once "improve bucket verification" PR is merged (so that they all share similar infrastructure and avoid duplication)

### DIFF
--- a/src/history/HistoryTests.cpp
+++ b/src/history/HistoryTests.cpp
@@ -23,6 +23,7 @@
 #include "work/WorkManager.h"
 
 #include "historywork/DownloadBucketsWork.h"
+#include "historywork/DownloadVerifyTxResultsWork.h"
 #include <lib/catch.hpp>
 #include <lib/util/format.h>
 
@@ -521,6 +522,41 @@ TEST_CASE("Ledger chain verification",
             HistoryManager::VERIFY_STATUS_OK);
         checkExpectedBehavior(Work::WORK_FAILURE_FATAL, lcl,
                               HashUtils::random());
+    }
+}
+
+TEST_CASE("Tx results verification", "[batching][resultsverification]")
+{
+    Config cfg(getTestConfig(0));
+    VirtualClock clock;
+    auto cg = std::make_shared<TmpDirHistoryConfigurator>();
+    cg->configure(cfg, true);
+    Application::pointer app = createTestApplication(clock, cfg);
+
+    CHECK(app->getHistoryArchiveManager().initializeHistoryArchive("test"));
+    CheckpointRange checkpointRange{
+        63, 703, app->getHistoryManager().getCheckpointFrequency()};
+
+    auto tmpDir = app->getTmpDirManager().tmpDir("ledger-chain-snap-test");
+    auto resultsGen = TestTxResultsGenerator{
+        *app, app->getHistoryArchiveManager().getHistoryArchive("test"),
+        checkpointRange, tmpDir};
+
+    auto& wm = app->getWorkManager();
+
+    SECTION("all valid")
+    {
+        resultsGen.makeTxResultFiles(true);
+        auto ledgerVerify = wm.executeWork<DownloadVerifyTxResultsWork>(
+            checkpointRange, tmpDir);
+        REQUIRE(ledgerVerify->getState() == Work::WORK_SUCCESS);
+    }
+    SECTION("invalid hash")
+    {
+        resultsGen.makeTxResultFiles(false);
+        auto ledgerVerify = wm.executeWork<DownloadVerifyTxResultsWork>(
+            checkpointRange, tmpDir);
+        REQUIRE(ledgerVerify->getState() == Work::WORK_FAILURE_FATAL);
     }
 }
 
--- a/src/history/HistoryTestsUtils.cpp
+++ b/src/history/HistoryTestsUtils.cpp
@@ -15,6 +15,7 @@
 #include "test/TestUtils.h"
 #include "test/TxTests.h"
 #include "test/test.h"
+#include "util/Math.h"
 #include "util/XDROperators.h"
 #include "work/WorkManager.h"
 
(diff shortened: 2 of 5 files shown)


src/history/HistoryTests.cpp
```diff
@@ -204,6 +225,124 @@ TEST_CASE("History publish queueing", "[history][historydelay][historycatchup]")
...
+    auto cg = std::make_shared<TmpDirHistoryConfigurator>();
+    cg->configure(cfg, true);
+    Application::pointer app = createTestApplication(clock, cfg);
+    CHECK(app->getHistoryArchiveManager().initializeHistoryArchive("test"));
+
+    auto bucketGenerator = TestBucketGenerator{
+        *app, app->getHistoryArchiveManager().getHistoryArchive("test")};
+    std::vector<std::string> hashes;
+    auto& wm = app->getWorkManager();
+    std::map<std::string, std::shared_ptr<Bucket>> mBuckets;
```
vogel: m prefix suggest member

src/work/BatchableWork.h
```diff
@@ -0,0 +1,43 @@
...
+
+namespace stellar
+{
+
+class BatchableWork : public Work
+{
+    // Class that provides skeleton for snapshot classes.
+    // Child classes can create dependencies by utilizing
+    // registerDependent and notifyCompleted and overriding unblockWork. This is
+    // useful if we need to enforce a DAG in the work structure. For example,
```
vogel: I wonder if (in future) we could use this DAG instead of child-parent relationship in Work.

src/historywork/BatchWork.h
```diff
@@ -17,6 +18,13 @@ class BatchWork : public Work
        they'd like to perform. This class only acts as a commander, adding more
        work if it has bandwidth.
     **/
+    void handleNewWork(bool isFirst);
+
+    // Keep track of last assigned work as any next work will depend on it
+    // (has to be strong reference in case it finished prior to new work
+    // assigned)
+    std::shared_ptr<BatchableWork> mLastAssignedWork;
```
MonsieurNicolas: I don't think this is the right change:
it seems to me that it would be much cleaner to have `yieldMoreWork` manage dependencies between work items as it's the iterator, and it knows what it's producing (so if we use lambdas for notifications, we have access to the right types).

Right now having `BatchWork` doing it hard codes for no good reason the dependency between work items.

MonsieurNicolas: This probably requires some changes to `BatchableWork` to enable those semantics, but we would end up with much cleaner contract and semantics:
* allow more complex synchronization by allowing any `BatchableWork` to tell its associated `BatchWork` that it's waiting on an existing `BatchableWork` after the fact (again, by having dependencies tracked at the generator level, it's easy to keep track of the right `BatchableWork`)

src/work/BatchableWork.h
```diff
@@ -0,0 +1,43 @@
...
+    };
+
+    std::shared_ptr<BatchableWork> mDependentSnapshot;
+    BatchableWorkResultData mSnapshotData{};
+
+  public:
+    BatchableWork(Application& app, WorkParent& parent, std::string name,
+                  size_t maxRetries = RETRY_A_FEW);
+    ~BatchableWork() override;
+    void registerDependent(std::shared_ptr<BatchableWork> blockedWork);
```
MonsieurNicolas: What do we expect the implementor of a `BatchableWork` to do ? The contract is not super clear to me (vs responsibilities of `BatchWork` implementors)

src/historywork/BatchWork.cpp
```diff
@@ -65,4 +69,27 @@ BatchWork::notify(std::string const& child)
     mApp.getCatchupManager().logAndUpdateCatchupStatus(true);
     advance();
 }
+
+void
+BatchWork::handleNewWork(bool isFirst)
+{
+    auto newWork = yieldMoreWork();
+
+    if (!isFirst)
```
MonsieurNicolas: not sure I understand the logic of `isFirst`: shouldn't that just be `if (mLastAssignedWork) {` ?

src/historywork/GetAndUnzipRemoteFileWork.h
```diff
@@ -6,14 +6,15 @@
 
 #include "work/Work.h"
 
+#include "work/BatchableWork.h"
```
MonsieurNicolas: we should instead replace the include of `work.h`

src/catchup/VerifyLedgerChainWork.cpp
```diff
@@ -103,15 +112,10 @@ VerifyLedgerChainWork::getStatus() const
...
 VerifyLedgerChainWork::onReset()
 {
-    // Note that mFirstVerified is not reset here, since it is set
-    // in the end of verification, where work wouldn't fail
-
-    if (mVerifiedAhead.header.ledgerSeq != 0)
+    if (mFirstVerified.header.ledgerSeq != 0)
     {
-        mVerifiedAhead = {};
+        mFirstVerified = {};
```
MonsieurNicolas: split this commit, there seems to be multiple unrelated changes in there

src/catchup/VerifyLedgerChainWork.cpp
```diff
@@ -284,29 +308,20 @@ VerifyLedgerChainWork::verifyAgainstTrustedHash(
...
 {
-    mApp.getCatchupManager().logAndUpdateCatchupStatus(true);
-
-    if (mCurrCheckpoint <
-        mApp.getHistoryManager().checkpointContainingLedger(mRange.first()))
+    if (!mReady)
     {
-        throw std::runtime_error(
-            "Verification undershot first ledger in the range.");
+        return WORK_PENDING;
```
MonsieurNicolas: I think there is a logical bug here:
if we're not ready, why did we schedule success? We should continue running instead

src/catchup/VerifyLedgerChainWork.h
```diff
@@ -43,14 +44,24 @@ class VerifyLedgerChainWork : public Work
     medida::Meter& mVerifyLedgerChainFailure;
     medida::Meter& mVerifyLedgerChainFailureEnd;
 
-    HistoryManager::LedgerVerificationStatus verifyHistoryOfSingleCheckpoint();
+    void
+    makeReady()
```
MonsieurNicolas: `setReady` would be more accurate (`makeX` is used for constructing objects)

src/historywork/DownloadVerifyLedgersWork.h
```diff
@@ -0,0 +1,52 @@
...
+    CheckpointRange mCheckpointRange;
+    uint32_t mNext;
+    TmpDir const& mDownloadDir;
+
+    LedgerHeaderHistoryEntry& mFirstVerified;
+    LedgerHeaderHistoryEntry const& mLastClosedLedger;
+    optional<Hash> mTrustedHash;
+
+    // Download Metrics
+    medida::Meter& mDownloadStart;
```
MonsieurNicolas: I would like this to show up as a refactor as opposed to new code

src/historywork/LedgerChainVerificationSnapshotWork.cpp
```diff
@@ -0,0 +1,144 @@
...
+            fmt::format("Blocking work for checkpoint {:s} passed invalid data",
+                        mCheckpoint));
+    }
+
+    mVerifiedAhead = verified;
+
+    if (mDownloadVerifyWork)
+    {
+        mDownloadVerifyWork->makeReady();
+        advanceChildren();
```
MonsieurNicolas: When you redo/rethink the Work state machine, I would like to avoid having direct calls to `advance` and `advanceChildren`: this is an internal to `Work` and seems error prone (why it's not just `run`, and done by some common code in `BatchableWork`)

PR #1732, opened 2018-07-24
Command line

### LINKED ISSUE #1418: Improve our command-line interface and code

vogel:
# Description

Resolves #1418

This PR implement new command-line syntax with Clara library. New command is more similar to git:

```
stellar-core new-db --conf testnet.cfg
stellar-core new-hist h1 h2 h3
stellar-core gen-seed --conf testnet.cfg
stellar-core run --conf testnet.cfg
```

# Checklist
- [x] Reviewed the [contributing](../CONTRIBUTING.md) document
- [x] Ran `clang-format`
- [x] Compiles
- [x] Ran all tests
- [x] If change impacts performance, include supporting evidence per the [performance document](../performance-eval.md)


src/main/Commands.cpp
```diff
@@ -0,0 +1,921 @@
...
+
+    fuzz(fileName, options.mLogLevel, options.mMetrics);
+    return 0;
+}
+
+int
+runGenFuzz(Args const& args)
+{
+    std::string fileName;
+    auto parser = Parser{} | Arg(fileName, "FILE").required();
```
MonsieurNicolas: `FILE` -> `FILE-NAME`

also, maybe you should have a helper function for the predicate: it's quite a bit of copy/paste

src/main/Commands.cpp
```diff
@@ -0,0 +1,921 @@
...
+
+    inferQuorumAndWrite(*options.mConfig);
+    return 0;
+}
+
+int
+runLoadXDR(Args const& args)
+{
+    std::string xdr;
+    auto parser = Parser{} | Arg(xdr, "XDR").required();
```
MonsieurNicolas: `XDR` -> `FILE-NAME`

src/main/Commands.cpp
```diff
@@ -0,0 +1,921 @@
...
+
+    initializeDatabase(*options.mConfig);
+    return 0;
+}
+
+int
+runNewHist(Args const& args)
+{
+    std::vector<std::string> newHistories;
+    auto parser = Parser{} | Arg(newHistories, "HISTORY").required();
```
MonsieurNicolas: `HISTORY` -> `HISTORY-LABEL`

src/main/Commands.cpp
```diff
@@ -0,0 +1,921 @@
...
+    showOfflineInfo(*options.mConfig);
+    return 0;
+}
+
+int
+runPrintTransaction(Args const& args)
+{
+    std::string transaction;
+    bool base64{false};
+    auto parser = Arg(transaction, "TRANSACTION").required() |
```
MonsieurNicolas: `TRANSACTION` -> `FILE-NAME`

src/main/Commands.cpp
```diff
@@ -0,0 +1,921 @@
...
+    printtxn(transaction, base64);
+    return 0;
+}
+
+int
+runReportLastHistoryCheckpoint(Args const& args)
+{
+    std::string outputFile;
+    auto parser =
+        Parser{} | Opt(outputFile, "FILE")["--output-file"]("output file");
```
MonsieurNicolas: `FILE` -> `FILE-NAME`

src/main/Commands.cpp
```diff
@@ -0,0 +1,921 @@
...
+    return 0;
+}
+
+int
+runSignTransaction(Args const& args)
+{
+    std::string transaction;
+    std::string netId;
+    bool base64;
+    auto parser = Arg(transaction, "TRANSACTION").required() |
```
MonsieurNicolas: `TRANSACTION` -> `FILE-NAME`

src/main/Commands.cpp
```diff
@@ -0,0 +1,921 @@
...
+                  Opt(netId, "STRING")["--netid"]("network ID used for signing")
```
MonsieurNicolas: `STRING` -> `NETWORK-PASSPHRASE`

src/main/Commands.cpp
```diff
@@ -0,0 +1,921 @@
...
+    std::cout << STELLAR_CORE_VERSION << std::endl;
+    return 0;
+}
+
+int
+runWriteQuorum(Args const& args)
+{
+    std::string outputFile;
+    auto parser =
+        Parser{} | Opt(outputFile, "FILE")["--output-file"]("output file");
```
MonsieurNicolas: `FILE` -> `FILE-NAME`

MonsieurNicolas:
Also note that `required()` is not implemented yet in Clara (https://github.com/catchorg/Clara/pull/28 ) ; but you already compensate for this from what I saw.

src/main/Config.cpp
```diff
@@ -237,20 +237,30 @@ Config::loadQset(std::shared_ptr<cpptoml::toml_group> group, SCPQuorumSet& qset,
 }
 
 void
-Config::load(std::string const& filename)
+Config::load(std::string fileName)
```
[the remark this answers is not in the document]
vogel: I was able to revert that change. Making this into static function will be possible after removing DeprecatedCommandLine - for now it would be more work that it deserves (because of Config wariable being used both inside and outside of catch block and me not wanting to do Config cfg; cfg = Config::load();).

src/main/Commands.cpp
```diff
@@ -0,0 +1,921 @@
...
+#include "test/test.h"
+#include "util/Logging.h"
+#include "util/optional.h"
+
+#include <iostream>
+#include <lib/catch.hpp>
+#include <lib/util/format.h>
+
+// must be included after catch.hpp
+#include "test/SimpleTestReporter.h"
```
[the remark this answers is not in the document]
vogel: The problem is that I have to do that.

CATCH_CONFIG_RUNNER does not only inject definitions for tests-related things, but also for the Clara library. That means that we need to keep test running code and command line parsing code in one place (as it is impossible to use CATCH_CONFIG_RUNNER in two separate cpp files).

There is second possibility - including separate Clara library. As Catch export all clara code as Catch::clara and standalone clara does is using just clara namespace it is possible for those two to co-exists. This allows us to have better separation between test and command parsing code, but having two clara libraries in one application seems wrong to me.

src/main/Commands.cpp
```diff
@@ -0,0 +1,921 @@
...
+    optional<int> execute(int argc, char* const* argv);
+    void writeToStream(std::ostream& os) const;
+
+  private:
+    std::string mExeName;
+    std::vector<Command> mCommands;
+};
+
+void
+writeWithTextFlow(std::ostream& os, std::string const& text)
```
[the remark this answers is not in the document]
vogel: Those are in anonymous namespace for now, which has similar effect (function name is not exported and is only available inside given .cpp file). Do you want me to change that to statics?

MonsieurNicolas:
@vogel : let me know when you want me to take a look again at this PR. Also, feel free to rebase/squash commits as I will probably just take a full pass anyways

vogel:
This branch is now rebased.

MonsieurNicolas:
had a question in the issue https://github.com/stellar/stellar-core/issues/1418#issuecomment-440092513

vogel:
Questions answered and updated commands.md

MonsieurNicolas:
r+ 6a31e84234e2a43ecc61a5349a49b66c1c5f2229

vogel:
Formatted

MonsieurNicolas:
r+ 8727d276c025577ed08030c71df73d62172ed9c1

PR #1733, opened 2018-07-24
remove try/catch block from tests due to a fix applied some time ago

vogel:
Remove some code that is no longer needed.


MonsieurNicolas:
needs to be rebased

PR #1738, opened 2018-07-30
Trivial fixes

vogel:
Few trivial fixes to compile warnings

### DIFF
--- a/src/crypto/SecretKey.cpp
+++ b/src/crypto/SecretKey.cpp
@@ -249,7 +249,7 @@ KeyFunctions<PublicKey>::getKeyValue(PublicKey& key)
 {
     switch (key.type())
     {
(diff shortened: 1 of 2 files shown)


MonsieurNicolas:
First one looks like the beginning of a wack-a-mole game.
Better would be to also enable a warning for the virtual destructor missing when there are virtual methods? I think the option is `-Wnon-virtual-dtor` (best would be to enable more warnings in general, including this one)

MonsieurNicolas:
It looks like those are fixes for existing warnings (you should have mentioned that), so I think those are good fixes as is. I'll merge post datalayer changes

MonsieurNicolas:
r+ 9d832c5f259d02c0b286f7a46b40ce413945de15

PR #1744, opened 2018-08-06
Windows fixes post pr1718

### LINKED ISSUE #848: Windows: intermittent crash while attempting to adoptFileAsBucket
Unclear what is causing this.

the rename in BucketManagerImpl::adoptFileAsBucket fails with "Access denied".
When this happens, the function throws, and the system collapses.

It's unclear to me what is causing this intermittent problem.
A few potential candidates:
- file still open in a different thread (there used to be a race with the checksum thread that I fixed earlier, maybe there is something else)
- file still open in a different process (gzip?)
- anti-virus

If it's AV the only way to work around this might be to either sleep a bit before attempting the rename or make the "adoptFileAsBucket" retry-able.

MonsieurNicolas:
# Description
This PR fixes a few problems with the windows build:
* there were some issues in the solution file causing certain files to not be in the right sub-folder
* the project didn't compile due to excessive number of symbols
* various warnings needed to be fixed
* a race condition was triggered when catching up ; Resolves #848

### DIFF
--- a/src/bucket/BucketManagerImpl.cpp
+++ b/src/bucket/BucketManagerImpl.cpp
@@ -184,7 +184,14 @@ BucketManagerImpl::adoptFileAsBucket(std::string const& filename,
         {
             std::string err("Failed to rename bucket :");
             err += strerror(errno);
-            throw std::runtime_error(err);
+            // it seems there is a race condition with external systems
+            // retry after sleeping for a second works around the problem
+            std::this_thread::sleep_for(std::chrono::seconds(1));
+            if (rename(filename.c_str(), canonicalName.c_str()) != 0)
+            {
+                // if rename fails again, surface the original error
+                throw std::runtime_error(err);
+            }
         }
 
         b = std::make_shared<Bucket>(canonicalName, hash);
--- a/Builds/VisualStudio/stellar-core.vcxproj
+++ b/Builds/VisualStudio/stellar-core.vcxproj
@@ -89,6 +89,7 @@
       </ControlFlowGuard>
       <SmallerTypeCheck>false</SmallerTypeCheck>
       <DebugInformationFormat>ProgramDatabase</DebugInformationFormat>
+      <AdditionalOptions>/bigobj %(AdditionalOptions)</AdditionalOptions>
     </ClCompile>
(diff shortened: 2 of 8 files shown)


src/bucket/BucketManagerImpl.cpp
```diff
@@ -184,7 +184,14 @@ BucketManagerImpl::adoptFileAsBucket(std::string const& filename,
...
             std::string err("Failed to rename bucket :");
             err += strerror(errno);
-            throw std::runtime_error(err);
+            // there is a race condition with external systems (windows)
+            // retry after sleeping for a second seems to work around the problem
+            std::this_thread::sleep_for(std::chrono::seconds(1));
+            if (rename(filename.c_str(), canonicalName.c_str()) != 0)
+            {
+                // if rename fails again, surface the original error
+                throw std::runtime_error(err);
```
jonjove: What message should the error contain? Should it be from the first failure, the second failure, or both? If it were to fail twice, I don't think it is necessarily true that both failures would be the same.

src/bucket/BucketManagerImpl.cpp
```diff
@@ -184,7 +184,14 @@ BucketManagerImpl::adoptFileAsBucket(std::string const& filename,
         {
             std::string err("Failed to rename bucket :");
             err += strerror(errno);
-            throw std::runtime_error(err);
+            // there is a race condition with external systems (windows)
```
jonjove: If this is only known to occur on Windows, then I think this should be a conditional compile for Windows.

src/bucket/BucketManagerImpl.cpp
MonsieurNicolas: I decided to use the original error as it's the reason we're failing in the first place

src/bucket/BucketManagerImpl.cpp
MonsieurNicolas: I am not sure it's a Windows only problem: I run into it all the time on Windows, that's all I know. I'd rather keep the code the same on both platforms unless this causes problems

jonjove:
r+ d483a7e

PR #1752, opened 2018-08-09
Added logging to prepareLiabilities

### LINKED ISSUE #1751: `prepareLiabilities` should be logging what it's doing
If an upgrade is causing `prepareLiabilities` to be called, there is no information at all in the logs about it, even though it's quite a major event...

I was looking at performance impact of an upgrade, but the logs didn't have any information about the upgrade that happened.

We need to add a few logging statements:
* at the `INFO` level, when we start `prepareLiabilities` and when it ends, with summary information on what happened (number of offers deleted and updated, number of accounts/trustlines that got updated)
* at the `DEBUG` level, to give a better idea of what is going on; maybe we should be logging which offers are being deleted/adjusted for example

jonjove:
Adds logging to `prepareLiabilities`. Resolves #1751

### DIFF
--- a/src/herder/Upgrades.cpp
+++ b/src/herder/Upgrades.cpp
@@ -479,6 +479,95 @@ shouldDeleteOffer(Asset const& asset, int64_t effectiveBalance,
                         : true;
 }
 
+enum class UpdateOfferResult
+{
+    Unchanged,
+    Adjusted,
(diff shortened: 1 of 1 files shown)


MonsieurNicolas:
r+ e908a9c1d59ebf9e65ee8c90a4c3418d3663feb6

Issue #1765, opened 2018-08-22
instrument code when resolving futures


MonsieurNicolas:
There is an assumption in the code base that processes that take a long time don't hang the main thread.

One of those long running processes is bucket merging that happens to be running on a separate thread, in the main thread we call [resolve](https://github.com/stellar/stellar-core/blob/6964a5ef991d27fbdc1dfab9e90d3f4f03f73df1/src/bucket/FutureBucket.cpp#L208) that normally should not hang.

Right now we don't have any insight if this is true.

We should add conditional logging to `FutureBucket::resolve` if waiting for the future takes too long (more than 1 second).
As this should "never" happen, a debug level of `INFO` should be used.

marta-lokhova:
Merged

PR #1766, opened 2018-08-22
Log time it takes to resolve a bucket

marta-lokhova:
Addresses #1765

### DIFF
--- a/src/bucket/FutureBucket.cpp
+++ b/src/bucket/FutureBucket.cpp
@@ -12,6 +12,7 @@
 #include "bucket/FutureBucket.h"
 #include "crypto/Hex.h"
 #include "main/Application.h"
+#include "util/LogSlowExecution.h"
 #include "util/Logging.h"
 
 #include <chrono>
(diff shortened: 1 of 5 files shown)


src/bucket/FutureBucket.cpp
```diff
@@ -210,7 +211,18 @@ FutureBucket::resolve()
     checkState();
     assert(isLive());
     clearInputs();
+    auto timer = ElapsedTime();
+
     std::shared_ptr<Bucket> bucket = mOutputBucket.get();
+    auto elapsed = timer.elapsed<>();
+    if (elapsed.count() > 1000)
```
MonsieurNicolas: as this will be used in more than one place can we instead have a scope object that performs the conditional logging?
Something like:
```c++
std::shared_ptr<Bucket> bucket;
auto SlowLOG(INFO, "Bucket") << "Resolving bucket " << hexAbbrev(bucket->getHash());
{
    bucket= mOutputBucket.get(); 
}
```

MonsieurNicolas:
r+ cf765d91c29151e74e8e385fa0130be1534f3d84

marta-lokhova:
Failure looks unrelated:
```
herder/UpgradesTests.cpp:1647: FAILED:
due to unexpected exception with message:
  bind: Address already in use
```

PR #1771, opened 2018-08-29
Cascade tmpdir during catchup

### LINKED ISSUE #1735: Handle catchup directory management better
Right now during catchup we use a single directory to store all ledger header and transaction files. If I try to do catchup complete, it's 159800 checkpoints as of now (and this number will only keep growing), meaning over 300,000 files in a single `buckets/tmp/catchup` directory!  This seems very inefficient from the file system point of view. 

I think we can improve it by using sub-directories, perhaps in the similar way we do it when resolving remote directories for archive publishing.

marta-lokhova:
Resolves #1735

### DIFF
--- a/src/history/FileTransferInfo.h
+++ b/src/history/FileTransferInfo.h
@@ -25,7 +25,7 @@ class FileTransferInfo
     std::string mType;
     std::string mHexDigits;
     std::string mLocalPath;
-    std::string mSuffix;
+    std::string getLocalDir(TmpDir const& localRoot) const;
 
   public:
     FileTransferInfo(Bucket const& bucket)
@@ -39,7 +39,7 @@ class FileTransferInfo
                      uint32_t checkpointLedger)
         : mType(snapType)
         , mHexDigits(fs::hexStr(checkpointLedger))
-        , mLocalPath(snapDir.getName() + "/" + baseName_nogz())
+        , mLocalPath(getLocalDir(snapDir) + "/" + baseName_nogz())
     {
     }
 
@@ -47,7 +47,7 @@ class FileTransferInfo
                      std::string const& hexDigits)
         : mType(snapType)
         , mHexDigits(hexDigits)
-        , mLocalPath(snapDir.getName() + "/" + baseName_nogz())
+        , mLocalPath(getLocalDir(snapDir) + "/" + baseName_nogz())
     {
     }
 
--- a/src/history/FileTransferInfo.cpp
+++ b/src/history/FileTransferInfo.cpp
@@ -11,4 +11,23 @@ char const* HISTORY_FILE_TYPE_LEDGER = "ledger";
 char const* HISTORY_FILE_TYPE_TRANSACTIONS = "transactions";
 char const* HISTORY_FILE_TYPE_RESULTS = "results";
 char const* HISTORY_FILE_TYPE_SCP = "scp";
+
+std::string
+FileTransferInfo::getLocalDir(TmpDir const& localRoot) const
+{
+    auto localDir = localRoot.getName();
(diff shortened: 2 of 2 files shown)


src/history/FileTransferInfo.h
```diff
@@ -36,21 +35,44 @@ class FileTransferInfo
...
     }
 
+    std::string
+    getLocalDir(TmpDir const& snapDir, std::string const& snapType,
+                bool cascade) const
+    {
+        auto localDir = snapDir.getName();
+        if (cascade)
+        {
+            localDir += "/" + fs::remoteDir(snapType, mHexDigits);
```
MonsieurNicolas: move this into the cpp file

src/history/FileTransferInfo.h
```diff
@@ -36,21 +35,44 @@ class FileTransferInfo
...
+            int retries = 5;
```
MonsieurNicolas: comment on why you need retries, I imagine it's because of race conditions? Why 5?
Should it sleep a little between retries?

src/history/FileTransferInfo.h
```diff
@@ -36,21 +35,44 @@ class FileTransferInfo
...
         : mType(snapType)
         , mHexDigits(hexDigits)
-        , mLocalPath(snapDir.getName() + "/" + baseName_nogz())
+        , mLocalPath(getLocalDir(snapDir, snapType, cascade) + "/" +
+                     baseName_nogz())
     {
     }
 
+    std::string
+    getLocalDir(TmpDir const& snapDir, std::string const& snapType,
```
MonsieurNicolas: this should be a private function, right?

src/history/FileTransferInfo.h
```diff
@@ -36,21 +35,44 @@ class FileTransferInfo
...
     {
     }
 
     FileTransferInfo(TmpDir const& snapDir, std::string const& snapType,
-                     std::string const& hexDigits)
+                     std::string const& hexDigits, bool cascade = false)
         : mType(snapType)
         , mHexDigits(hexDigits)
-        , mLocalPath(snapDir.getName() + "/" + baseName_nogz())
+        , mLocalPath(getLocalDir(snapDir, snapType, cascade) + "/" +
```
MonsieurNicolas: why is `getLocalDir` taking `snapType` as a parameter but not `hexDigits` ? I imagine it should take neither as they are member variables

marta-lokhova: good catch

src/history/FileTransferInfo.h
marta-lokhova: this is similar to what we do with TmpDir, i imagine there could be OS-related errors, hence the retries. Ok, sleeping in between makes sense.

src/history/FileTransferInfo.h
```diff
@@ -36,18 +36,18 @@ class FileTransferInfo
     }
 
     FileTransferInfo(TmpDir const& snapDir, std::string const& snapType,
-                     uint32_t checkpointLedger)
+                     uint32_t checkpointLedger, bool cascade = false)
```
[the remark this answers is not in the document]
marta-lokhova: I think it wasn't necessary for some use cases (StateSnapshot, FetchRecentQsetsWork etc). I can get rid of it for consistency, if we do not mind the overhead.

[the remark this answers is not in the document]
MonsieurNicolas: yeah it's just just error prone to have both

marta-lokhova:
updated and rebased

MonsieurNicolas:
r+ 4e02d3f6964f22481e1dd5d7a26b48f8c4213f33

PR #1772, opened 2018-08-30
Remove unused import to main/Config.h

Geod24:
# Description

This import was not used. It also showed up as an obvious breach of abstraction (pulling in tons of dependencies transitively) when I attempted to isolate the SCP code in its own `.so`

### DIFF
--- a/src/scp/NominationProtocol.cpp
+++ b/src/scp/NominationProtocol.cpp
@@ -8,7 +8,6 @@
 #include "crypto/Hex.h"
 #include "crypto/SHA.h"
 #include "lib/json/json.h"
-#include "main/Config.h"
 #include "scp/LocalNode.h"
 #include "scp/QuorumSetUtils.h"
 #include "util/GlobalChecks.h"


MonsieurNicolas:
Thank you! Nice to see that you're looking into splitting the SCP code into its own submodule, I've never got to it.

MonsieurNicolas:
r+ 10f9d49e80bbee75351c39b33aa6930c7b08258e

Geod24:
> Nice to see that you're looking into splitting the SCP code into its own submodule, I've never got to it.

While I didn't originally intend to PR all changes upstream, since I took shortcuts (e.g. flat out removing the JSON code to kill the JSON dependency) which of course wouldn't fly here, it does look like it would be easier than anticipated, so I am considering it.

One issue I'm not sure how to tackle best is dependency on other part of `stellar-core`. E.g. SCP depends on the XDR headers (`xdr/Stellar-SCP.h`, `xdr/Stellar-types.h`), some utils (`XDROperators`, `numeric`, `NonCopyable`...) and of course `crypto` (`ScretKey`, `SHA`, etc...)
Should they also be modularized ? Copied into SCP (for XDR that might be okay, but for crypto I doubt so) ? Copied over to the submodule for dependency, but not actually used when building stellar ?

Since this will have an impact on your workflow, and I'm not so familiar with Stellar's code, I would be glad if you could take the time to open an issue with your thoughts on this.

MonsieurNicolas:
yeah I think we'll probably look into this later, I agree that right now, what you did (copying the code and gutting what you don't need) is the easiest.

PR #1773, opened 2018-08-30
Remove SCP's dependency on utils/types.h

Geod24:
```
By removing those unused imports and splitting some utils from
util/types.h to util/num64.h, we remove any dependency existing
between the SCP package and utils/types.h

The reason to do so is that types.h import overlay/StellarXDR.h,
which itself imports the various xdr/Stellar-[...].h definitions,
which includes types the SCP package isn't supposed to know about.
```

No code is added - code is merely moved.


src/util/types.h
```diff
@@ -60,22 +61,6 @@ strToAssetCode(xdr::opaque_array<N>& ret, std::string const& str)
...
-// calculates A*B/C when A*B overflows 64bits
-int64_t bigDivide(int64_t A, int64_t B, int64_t C, Rounding rounding);
-// no throw version, returns true if result is valid
-bool bigDivide(int64_t& result, int64_t A, int64_t B, int64_t C,
-               Rounding rounding);
-
-// no throw version, returns true if result is valid
-bool bigDivide(uint64_t& result, uint64_t A, uint64_t B, uint64_t C,
-               Rounding rounding);
-
```
MonsieurNicolas: I think you should move the other related functions:
`bigDivide` and `bigMultiply`

src/util/num64.h
```diff
@@ -0,0 +1,26 @@
+#pragma once
+
+// Copyright 2014 Stellar Development Foundation and contributors. Licensed
+// under the Apache License, Version 2.0. See the COPYING file at the root
+// of this distribution or at http://www.apache.org/licenses/LICENSE-2.0
+
+#include <stdint.h>
```
MonsieurNicolas: I think the portable way is to use `#include <cstdint>` instead (according to https://en.cppreference.com/w/cpp/types/integer )

src/util/types.h
Geod24: Right.
What about `addBalance` ? While the name is quite explicit about how it's used, the implementation is simplify an `add` with a ceiling.

Geod24:
Review applied:
- Use `<cstdint>`
- Move other `bigDivide` overload

Additionally renamed `num64` to `numeric`.

src/util/types.h
MonsieurNicolas: I'd leave it here as it's clearly a "stellar-core" type of function

MonsieurNicolas:
looks good, thank you!

MonsieurNicolas:
r+ 4161eea53efb5b454b78e715a3e736a34337ad28

MonsieurNicolas:
r+ 11ae0813bb8185331722d4fe4008c2be765f15a8

MonsieurNicolas:
looks like it hit `  bind: Address already in use` for some reason.

MonsieurNicolas:
that means that #1635 was not enough

MonsieurNicolas:
@latobarita: retry

Geod24:
Thanks for the merge/fix!

PR #1781, opened 2018-09-04
Signer weight fix

### LINKED ISSUE #1780: missing check for signer weight
# Issue type
  bug

# Version found
9.2

# Issue description
Looks like we're not checking if a signer weight is within bounds. This can lead to failures later on where a signature might be considered invalid even though it has enough weights.

## Actual result
`SetOptions` accepts 256 as a valid weight.

## Expected result
`SetOptions` should fail when a weight greater than 255 is passed in.

MonsieurNicolas:
# Description

Resolves #1780 

This PR:
* adds checks to `SetOptions` so that weights like 256 are not allowed anymore
* ensures that we end up with a max value when computing the sum of weights


jonjove:
r+ 817c9eb

PR #1786, opened 2018-09-09
Fix mPreparedPrime ballot update bug

ult-bobonovski:
Fix a bug that forgets to check the incompatibility of `mPrepared` ballot and new ballot before updating `mPreparedPrime`  ballot.

# Checklist
- [ ] Ran `clang-format`
- [ ] Compiles
- [ ] Ran all tests
- [ ] If change impacts performance, include supporting evidence per the [performance document](https://github.com/stellar/stellar-core/blob/master/performance-eval.md)

### DIFF
--- a/src/scp/BallotProtocol.cpp
+++ b/src/scp/BallotProtocol.cpp
@@ -1619,8 +1619,11 @@ BallotProtocol::setPrepared(SCPBallot const& ballot)
         }
         else if (comp > 0)
         {
-            // check if we should update only p'
-            if (!mPreparedPrime || compareBallots(*mPreparedPrime, ballot) < 0)
+            // check if we should update only p' with two situations:
+            // 1. p' is NULL
+            // 2. p' is smaller than ballot and p is incompatible with ballot
+            if (!mPreparedPrime || 
+                (compareBallots(*mPreparedPrime, ballot) < 0 && !areBallotsCompatible(*mPrepared, ballot)))
             {
                 mPreparedPrime = std::make_unique<SCPBallot>(ballot);
                 didWork = true;


MonsieurNicolas:
A few questions about this change:
1. why do you think we need this test? Which part of the whitpaper do you think is not implemented properly? afaik `p` and `p_prime` are supposed to be the top two ballots accepted prepared, it's very possible that `p_prime` is the one compatible with `b` (especially when `b` gets updated).
2. add/update test cases. This is also a good way to illustrate why we need the change if it triggers some strangeness (it's possible that it's a bug in the whitepaper itself, it would not be the first time).

ult-bobonovski:
Hi, MonsieurNicolas

About the first question:
In page 23 of the white paper, the meaning of `p` and `p'` are defined as 
> the two highest ballots accepted as prepared such that `p` < `p'` while `p` and `p'` are *incompatible*

unless this requirement is wrong, we have to check these conditions each time when we update `p` and `p'`.

Back to the proposed change, I will post the original code and write some comments there to show my proposed check should be added (if the whitepaper is correct)

```bool
BallotProtocol::setPrepared(SCPBallot const& ballot)
{
    bool didWork = false;

    if (mPrepared)
    {
        int comp = compareBallots(*mPrepared, ballot);
        if (comp < 0) // (bobonovski): we know mPrepared < ballot
        {
            if (!areBallotsCompatible(*mPrepared, ballot)) 
            {
                // (bobonovski): we are here, so ballot is incompatible with mPrepared,
                // the two requirements of p and p' from whitepaper are satisfied, 
                // and the new p and p' pairs are mPrepared and ballot respectively,
               //  we can update mPreparedPrime now
                mPreparedPrime = std::make_unique<SCPBallot>(*mPrepared);
            }
            // (bobonovski) either ways, we found a larger p, we can update mPrepared
            mPrepared = std::make_unique<SCPBallot>(ballot);
            didWork = true;
        }
        else if (comp > 0) // (bobonovski): we know ballot < mPrepared
        {
            // check if we should update only p'
            if (!mPreparedPrime || compareBallots(*mPreparedPrime, ballot) < 0)
            {
                // (bobonovski) we are here, we know ballot is larger than p', so we have
                // a possible change to find a larger p'. However, we have to make sure
               // this new p' candidate is incompatible with p to satisfy the requirement from
               // the whitepaper
                mPreparedPrime = std::make_unique<SCPBallot>(ballot);
                didWork = true;
            }
        }
    }
    else
    {
        mPrepared = std::make_unique<SCPBallot>(ballot);
        didWork = true;
    }
    return didWork;
}

MonsieurNicolas:
Ah yes, this makes sense. p and p' should not be compatible.

You just need to add a commit with updated test cases and it will be good to merge

MonsieurNicolas:
Did you have a chance to look at tests? Alternatively I can look into adding some myself, let me know.

ult-bobonovski:
I've been trapped in other works, It would be great that you can help with the tests. 
Thanks

PR #1788, opened 2018-09-10
New ledger state

### LINKED ISSUE #1417: Data layer consolidation
Right now we have a few data structures used to keep track of ledger entry state in our code base:
* "backend" calls to soci (SQL), where we perform CRUD and transactions
  * this layer also has a simple cache
* LedgerDelta construct used to build the input to the BucketList, also supports CRUD and transactions

Currently, the code is written in such a way that it's the caller's responsibility to always perform both calls to each subsystem correctly; it's also the caller's responsibility to keep track of objects in memory. 

We should be able to consolidate those calls behind a single data layer API that would perform the proper calls to the various subsystems.

With this we should also be able to:
* simplify the client code that would not have to worry about transactions being rolled back in one system but not the other
* move towards "in memory" transactions (not use nested SQL transactions in the typical case)
* detect (or better, not allow by switching to better semantics) the use of stale data in business logic
   * reads from a cached copy object that was updated later on (ie shadowing effect)
   * writes from an old copy (ie rollback effect)

jonjove:
Resolves #1417, replaces #1600 

This is a redesigned implementation which resolves many issues discussed in #1600. Some notable design improvements include:
- Better separation of responsibilities among recursive instances of `LedgerState` and `LedgerStateRoot`
- Better use of RAII to minimize the need of `invalidate` (now known as `deactivate`)
- Removed the notion of `forget` and replaced it with the notion of a read-only load

More tests are still required and `clang-format` has not yet been run.


src/ledger/LedgerState.cpp
```diff
@@ -0,0 +1,1311 @@
...
+        auto key = LedgerEntryKey(offer);
+        if (exclude.find(key) == exclude.end())
+        {
+            return std::make_shared<LedgerEntry>(offer);
+        }
+    }
+
+    while (!cached.second)
+    {
+        auto newOffers = loadBestOffers(buying, selling, 5, offers.size());
```
MonsieurNicolas: probably good to define a constant here as 5 is supposed to match on both lines

src/ledger/LedgerState.cpp
```diff
@@ -0,0 +1,1311 @@
...
+#include "util/XDROperators.h"
+#include "xdr/Stellar-ledger-entries.h"
+#include "xdrpp/marshal.h"
+#include <soci.h>
+
+namespace stellar
+{
+
+bool
+isBetterOffer(LedgerEntry const& lhsEntry, LedgerEntry const& rhsEntry)
```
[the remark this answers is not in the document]
MonsieurNicolas: in fact I would argue that those functions should live next to their SQL counterpart as they have to match

src/ledger/LedgerState.cpp
```diff
@@ -0,0 +1,1311 @@
...
+{
+    auto const& lhs = lhsEntry.data.offer();
+    auto const& rhs = rhsEntry.data.offer();
+
+    assert(lhs.buying == rhs.buying);
+    assert(lhs.selling == rhs.selling);
+
+    double lhsPrice = double(lhs.price.n) / double(lhs.price.d);
```
MonsieurNicolas: can we confirm that SQL actually performs the math like this?

src/ledger/LedgerState.cpp
```diff
@@ -0,0 +1,1311 @@
...
+{
+    std::string cacheKey = binToHex(xdr::xdr_to_opaque(buying))
+                         + binToHex(xdr::xdr_to_opaque(selling));
+    if (!mBestOffersCache->exists(cacheKey))
+    {
+        mBestOffersCache->put(cacheKey, {{}, false});
+    }
+    auto& cached = mBestOffersCache->get(cacheKey);
+
+    auto& offers = cached.first;
```
MonsieurNicolas: a comment here would help: I didn't realize at first that the cache was sorted properly

src/ledger/LedgerState.cpp
```diff
@@ -0,0 +1,1311 @@
...
+    while (!cached.second)
+    {
+        auto newOffers = loadBestOffers(buying, selling, 5, offers.size());
+        if (newOffers.size() < 5)
+        {
+            cached.second = true;
+        }
+
+        offers.insert(offers.end(), newOffers.begin(), newOffers.end());
+        for (auto const& offer : newOffers)
```
MonsieurNicolas: we may want to do a pass at those as well: we copy for no reason (could iterate over the last `N` items from offers)

src/ledger/LedgerState.cpp
```diff
@@ -0,0 +1,1311 @@
...
+
+std::map<LedgerKey, LedgerEntry>
+LedgerStateRoot::Impl::getOffersByAccountAndAsset(AccountID const& account,
+                                                  Asset const& asset)
+{
+    std::map<LedgerKey, LedgerEntry> res;
+    auto offers = loadOffersByAccountAndAsset(account, asset);
+    for (auto const& offer : offers)
+    {
+        res[LedgerEntryKey(offer)] = offer;
```
MonsieurNicolas: use emplace when you know that you have only new items

src/ledger/LedgerState.cpp
```diff
@@ -0,0 +1,1311 @@
...
+    std::shared_ptr<LedgerEntry const> const& entry)
+{
+    auto const previous = getNewestVersion(self, key).entry;
+    if (entry)
+    {
+        insertOrUpdateData(*entry, !previous);
+    }
+    else if (previous) // Deleted
+    {
+        deleteData(key);
```
MonsieurNicolas: why do we need to use `previous` when we delete? We could avoid loading the old value

src/ledger/LedgerState.cpp
```diff
@@ -0,0 +1,1311 @@
+// Copyright 2018 Stellar Development Foundation and contributors. Licensed
+// under the Apache License, Version 2.0. See the COPYING file at the root
+// of this distribution or at http://www.apache.org/licenses/LICENSE-2.0
+
+#include "ledger/LedgerState.h"
```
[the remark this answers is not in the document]
MonsieurNicolas: and what I mean by that is that I pushed a commit to your branch

src/ledger/LedgerState.cpp
```diff
@@ -0,0 +1,1311 @@
...
+void
+LedgerState::Impl::commitChild(LedgerState& self)
+{
+    auto entries = mChild->getEntries();
+    for (auto const& kv : entries)
+    {
+        auto const& key = kv.first;
+        if (kv.second)
+        {
+            mEntry[key] = std::make_shared<LedgerEntry>(*kv.second);
```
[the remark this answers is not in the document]
MonsieurNicolas: maybe not exactly like this (it would require special casing exception handling), my point is more that it's OK to have destructive semantics with the child, so if you were "moving" entries one by one into the parent, we can avoid the copying

src/bucket/BucketApplicator.cpp
```diff
@@ -25,32 +28,38 @@ BucketApplicator::operator bool() const
...
-        delta.commit();
         if ((++mSize & 0xff) == 0xff)
         {
             break;
         }
     }
-    sqlTx.commit();
-    mDb.clearPreparedStatementCache();
+    ls.commit();
+    mApp.getDatabase().clearPreparedStatementCache();
```
MonsieurNicolas: btw, should we still do these, or those "clear" statements should be done from within `LedgerStateRoot` ?

src/herder/Upgrades.cpp
```diff
@@ -489,19 +488,21 @@ enum class UpdateOfferResult
...
+    AbstractLedgerState& ls)
 {
     using namespace std::placeholders;
-    auto& offer = offerFrame.getOffer();
+    auto& offer = offerEntry.current().data.offer();
 
-    auto availableBalanceBind = std::bind(getAvailableBalance, offer.sellerID,
-                                          _1, _2, std::ref(ld), std::ref(db));
+    auto availableBalanceBind = std::bind(
+        (int64_t(*)(AccountID const&, Asset const&, int64_t,
```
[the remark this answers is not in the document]
MonsieurNicolas: interestingly, by putting the cast it helps the compiler resolve the type for the template and will use the right function as a result - scary nevertheless :)

src/bucket/BucketTests.cpp
```diff
@@ -1098,15 +1059,19 @@ TEST_CASE("bucket apply", "[bucket]")
```
MonsieurNicolas: 'sess': local variable is initialized but not referenced

src/herder/HerderTests.cpp
```diff
@@ -382,8 +386,11 @@ TEST_CASE("surge", "[herder]")
 
     auto& lm = app->getLedgerManager();
```
MonsieurNicolas: 'lm': local variable is initialized but not referenced

src/herder/TxSetFrame.cpp
```diff
@@ -165,7 +171,10 @@ TxSetFrame::surgePricingFilter(LedgerManager const& lm)
         map<AccountID, double> accountFeeMap;
         for (auto& tx : mTransactions)
         {
-            double r = tx->getFeeRatio(lm);
+            double fee = tx->getFee();
+            double minFee = tx->getMinFee(header);
```
MonsieurNicolas: need cast:
conversion from 'int64_t' to 'double', possible loss of data

src/invariant/BucketListIsConsistentWithDatabaseTests.cpp
```diff
@@ -2,330 +2,291 @@
...
-                auto index = dist(gen);
-                auto iter = selected.cbegin();
-                std::advance(iter, index);
-                REQUIRE_THROWS_AS(
-                    applyBucketsAndCrankUntilDone<ABW>(appGenerate, appApply,
-                                                       ledgerSeq, *iter, let),
-                    InvariantDoesNotHold);
-                ++nTests;
+                std::uniform_int_distribution<uint32_t> dist(
+                    0, filteredKeys.size() - 1);
```
MonsieurNicolas: need cast to `uint32_t` (from `size_t`):
conversion from 'size_t' to '_Ty', possible loss of data

src/invariant/BucketListIsConsistentWithDatabaseTests.cpp
```diff
@@ -495,187 +442,193 @@ class ApplyBucketsWorkModifyEntry : public ApplyBucketsWork
...
-        applyBucketsAndCrankUntilDone(appGenerate, appApply, ledgerSeq));
+    struct TestRootBucketListGenerator : public BucketListGenerator
+    {
+        uint32_t const mTargetLedger;
+        bool mModifiedRoot;
+
+        TestRootBucketListGenerator()
+            : BucketListGenerator()
+            , mTargetLedger(
+                  std::uniform_int_distribution<size_t>(2, 100)(*mGen))
```
MonsieurNicolas: should be `uint32_t` not `size_t` (`mTargetLedger`):
conversion from 'size_t' to 'const uint32_t', possible loss of data

src/invariant/LiabilitiesMatchOffersTests.cpp
```diff
@@ -295,27 +300,22 @@ TEST_CASE("Invariant for liabilities", "[invariant][liabilitiesmatchoffers]")
     SECTION("create then modify then delete offer")
     {
         auto offer = generateOffer(cur1, cur2, 100, Price{1, 1});
-        auto selling = generateSellingLiabilities(lm, offer, false, true);
-        auto buying = generateBuyingLiabilities(lm, offer, false, true);
+        auto selling = generateSellingLiabilities(*app, offer, false, true);
```
MonsieurNicolas: line 288 `LedgerManager& lm = app->getLedgerManager();`
'lm': local variable is initialized but not referenced

src/ledger/LedgerManagerImpl.cpp
```diff
@@ -308,7 +314,8 @@ LedgerManagerImpl::loadLastKnownLedger(
         }
         else
         {
-            advanceLedgerPointers();
+            LedgerState ls(mApp.getLedgerStateRoot());
```
MonsieurNicolas: `ls` shadows previous declaration (probably better to refactor this function anyways as to not have the previous one in scope at all)

src/transactions/ChangeTrustTests.cpp
```diff
@@ -28,7 +33,7 @@ TEST_CASE("change trust", "[tx][changetrust]")
 
     // set up world
     auto root = TestAccount::createRoot(*app);
-    auto const minBalance2 = app->getLedgerManager().getMinBalance(2);
+    auto const minBalance2 = app->getLedgerManager().getLastMinBalance(2);
```
MonsieurNicolas: 'db': local variable is initialized but not referenced (line 30)

src/transactions/ManageOfferOpFrame.cpp
```diff
@@ -359,79 +363,94 @@ ManageOfferOpFrame::doApply(Application& app, LedgerDelta& delta,
...
+                {
+                    sheepLineA = stellar::loadTrustLineWithoutRecord(ls, getSourceID(), sheep);
+                }
+                if (wheat.type() != ASSET_TYPE_NATIVE)
+                {
+                    wheatLineA = stellar::loadTrustLineWithoutRecord(ls, getSourceID(), wheat);
+                }
+
+                OfferEntry& oe = newOffer.data.offer();
+                int64_t maxSheepSend =
```
MonsieurNicolas: `maxSheepSend` and `maxWheatReceive` shadow previous declarations (warning)

src/transactions/TransactionFrame.cpp
```diff
@@ -273,39 +296,42 @@ TransactionFrame::commonValidPreSeqNum(Application& app, LedgerDelta* delta)
...
-TransactionFrame::processSeqNum(LedgerManager& lm, LedgerDelta& delta)
+TransactionFrame::processSeqNum(AbstractLedgerState& ls)
 {
-    if (lm.getCurrentLedgerVersion() >= 10)
+    auto header = ls.loadHeader();
+    if (header.current().ledgerVersion >= 10)
     {
-        if (mSigningAccount->getSeqNum() > mEnvelope.tx.seqNum)
+        auto sourceAccount = loadSourceAccount(ls, header);
+        if (sourceAccount > mEnvelope.tx.seqNum)
```
MonsieurNicolas: bug: should be `sourceAccount.current().data.account().seqNum`
on that note: why do we need the `bool()` operator?
I tend to think that implicit conversions are evil - which includes non explicit constructors that take a single parameter (mmm, doing a sweep)

src/transactions/TransactionUtils.cpp
```diff
@@ -0,0 +1,634 @@
...
+int64_t getMaxAmountReceive(LedgerStateHeader const& header,
+                            LedgerEntry const& le)
+{
+    if (le.data.type() == ACCOUNT)
+    {
+        int64_t maxReceive = INT64_MAX;
+        if (header.current().ledgerVersion >= 10)
+        {
+            auto const& acc = le.data.account();
+            maxReceive -= acc.balance + getBuyingLiabilities(header, le);
```
vogel: is that correct?

src/herder/Upgrades.cpp
```diff
@@ -599,40 +595,40 @@ prepareLiabilities(LedgerManager& ledgerManager, LedgerDelta& ld)
...
             addLiabilities(initialSellingLiabilities, offer.sellerID,
-                           offer.selling, offerFrame->getSellingLiabilities());
+                           offer.selling, getOfferSellingLiabilities(ls.loadHeader(), offerEntry));
         }
 
-        auto accountFrame =
-            AccountFrame::loadAccount(ld, accountOffers.first, db);
-        if (!accountFrame)
+        auto accountEntry = stellar::loadAccount(ls, accountOffers.first);
+        auto const& acc = accountEntry.current().data.account();
```
vogel: I think that should be moved after the if

src/ledger/LedgerManagerImpl.cpp
```diff
@@ -944,99 +972,40 @@ LedgerManagerImpl::deleteOldEntries(Database& db, uint32_t ledgerSeq,
...
+        LedgerState ls(lsOuter);
         for (auto tx : txs)
         {
-            LedgerDelta thisTxDelta(delta);
-            tx->processFeeSeqNum(thisTxDelta, *this);
-            tx->storeTransactionFee(*this, thisTxDelta.getChanges(), ++index);
-            thisTxDelta.commit();
+            LedgerState lsTx(ls);
+            tx->processFeeSeqNum(lsTx);
+            auto ledgerSeq = lsTx.loadHeader().current().ledgerSeq;
```
vogel: do we need to fetch that in each iteration?

src/ledger/LedgerState.h
```diff
@@ -0,0 +1,442 @@
...
+    };
+
+    std::map<LedgerKey, EntryDelta> entry;
+    HeaderDelta header;
+};
+
+class AbstractLedgerStateParent
+{
+  public:
+    class Identifier
```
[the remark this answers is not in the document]
vogel: I think this is 100% not used in current version (only passed to functions that do not do anything with it).

src/bucket/Bucket.cpp
```diff
@@ -249,120 +243,4 @@ Bucket::merge(BucketManager& bucketManager,
...
-    }
-}
-
-// FIXME issue NNN: this should be refactored to run in a read-transaction on a
-// background thread and take a soci::session& rather than Database&. For now we
-// code it to run on main thread because there's a bunch of code to move around
-// in the ledger frames and db layer to make that work.
-
-void
-checkDBAgainstBuckets(medida::MetricsRegistry& metrics,
```
vogel: looks like you've left definition for that in Bucket.h

src/catchup/ApplyLedgerChainWork.cpp
```diff
@@ -99,7 +99,7 @@ TxSetFramePtr
 ApplyLedgerChainWork::getCurrentTxSet()
 {
     auto& lm = mApp.getLedgerManager();
-    auto seq = lm.getCurrentLedgerHeader().ledgerSeq;
+    auto seq = lm.getLastClosedLedgerNum() + 1;
```
vogel: I think it can be good moment to rename getLastClosedLedgerNum to getLastClosedLedgerSeq

src/herder/HerderTests.cpp
```diff
@@ -132,8 +134,11 @@ TEST_CASE("standalone", "[herder]")
                     feedTx(b);
                 }
 
-                bool hasC =
-                    app->getLedgerManager().getCurrentLedgerVersion() >= 10;
+                bool hasC = false;
+                {
```
vogel: I see there are some other uses of this pattern:

```
LedgerState ls(app->getLedgerStateRoot());
read something from ls.loadHeader().current()
```

also some of the form:

```
LedgerState ls(app->getLedgerStateRoot());
write something to ls.loadHeader().current()
ls.commit();
```

and multiple usages that do not seem to follow any of those patterns.

I'm wondering if you have considered extracting two first into some functions, as introducing new scope just for that seems cumbersome, for example:

```
auto hasC = readHeader(app).ledgerVersion;
writeHeader(app, [](Heder &h){ h.ledgerVersion = 11; });
``` 

I think first one could be useful, not sure if second one is worth it.

src/ledger/LedgerState.cpp
```diff
@@ -0,0 +1,1311 @@
...
+        throw std::runtime_error("LedgerStateHeader is active");
+    }
+
+    mActiveHeader = std::make_shared<HeaderImpl>(self, mHeader);
+    LedgerStateHeader header(mActiveHeader);
+    f(header.current());
+}
+
+void
+LedgerState::Impl::updateLastModified()
```
vogel: This name makes me uneasy.

This function seals current object and *maybe* updates last modified fields for each entry. What about renaming it to `seal`?

src/ledger/LedgerState.cpp
```diff
@@ -0,0 +1,1311 @@
...
+    for (auto const& kv : entries)
+    {
+        auto const& key = kv.first;
+        if (kv.second)
+        {
+            mEntry[key] = std::make_shared<LedgerEntry>(*kv.second);
+        }
+        else
+        {
+            if (getNewestVersion(self, key).ledgerState == &self &&
```
vogel: I see that ledgerState field is always used in context like this (check if ledgerState == first parameter of getNewestVersion). As we do not use that information for any other reason maybe we could simplify that by changing it into `bool createdHere;` (name TBD).

src/herder/Upgrades.cpp
```diff
@@ -734,32 +723,31 @@ prepareLiabilities(LedgerManager& ledgerManager, LedgerDelta& ld)
...
+    uint32_t prevVersion = header.current().ledgerVersion;
 
-    header.ledgerVersion = newVersion;
-    ledgerManager.getCurrentLedgerHeader().ledgerVersion = newVersion;
-    if (header.ledgerVersion >= 10 && prevVersion < 10)
+    header.current().ledgerVersion = newVersion;
+    if (header.current().ledgerVersion >= 10 && prevVersion < 10)
     {
-        prepareLiabilities(ledgerManager, ld);
+        header.deactivate();
```
vogel: If I understand it correctly, header needs to be deactivated here only because it does not go out of scope here? If so, could we extract decision if prepareLiabilities should be called to a lambda, avoiding manual deactivation of header?

Like

```
auto shouldPrepare = [&](){
  auto header = ls.loadHeader();
  uint32_t prevVersion = header.current().ledgerVersion;
  header.current().ledgerVersion = newVersion;
  return header.current().ledgerVersion >= 10 && prevVersion < 10;
}();
if (shouldPrepare) { prepareLiabilities(ls); }
```

Not sure if that is better, but answer to that question will improve my confidence in how things works now.

src/ledger/LedgerState.cpp
```diff
@@ -0,0 +1,1311 @@
...
+                                              Asset const& asset)
+{
+    auto offers = mParent.getOffersByAccountAndAsset(account, asset);
+    for (auto const& kv : mEntry)
+    {
+        auto const& key = kv.first;
+        auto const& entry = kv.second;
+        if (!entry)
+        {
+            offers.erase(key);
```
vogel: I would move that after checking if type is OFFER

src/ledger/LedgerState.cpp
```diff
@@ -0,0 +1,1311 @@
...
+std::vector<LedgerStateEntry>
+LedgerState::Impl::loadOffersByAccountAndAsset(LedgerState& self,
+                                               AccountID const& accountID,
+                                               Asset const& asset)
+{
+    checkNotSealed();
+    checkNoChild();
+
+    std::vector<LedgerStateEntry> res;
+    auto offers = getOffersByAccountAndAsset(accountID, asset);
```
vogel: It looks like:

LedgerState::loadOffersByAccountAndAsset calls
LedgerState::Impl::loadOffersByAccountAndAsset which calls
LedgerState::Impl::getOffersByAccountAndAsset

but:

LedgerStateRoot::getOffersByAccountAndAsset calls
LedgerStateRoot::Impl::getOffersByAccountAndAsset which callsLedgerStateRoot::Impl::loadOffersByAccountAndAsset

I think that this load/get relationship is not 100% clear for me.
Could you explain that?

src/ledger/LedgerState.cpp
```diff
@@ -0,0 +1,1311 @@
...
+{
+    if (mChild)
+    {
+        mChild->rollback();
+    }
+
+    mActive.clear();
+    mActiveHeader.reset();
+
+    mParent.rollbackChild(id);
```
vogel: this looks a bit weird
State in its own rollback initiates parent->rollbackChild, and its name suggest that it will do rollback on this State object, which parent does not do. What about renaming rollbackChild to clearChild or resetChild?

src/ledger/LedgerState.cpp
```diff
@@ -0,0 +1,1311 @@
...
+            return std::make_shared<LedgerEntry>(offer);
+        }
+    }
+
+    while (!cached.second)
+    {
+        auto newOffers = loadBestOffers(buying, selling, 5, offers.size());
+        if (newOffers.size() < 5)
+        {
+            cached.second = true;
```
vogel: probably worth implementing struct, as `allLoaded` is a betten name than `second`

src/ledger/LedgerState.cpp
```diff
@@ -0,0 +1,1311 @@
...
+    mImpl->commitChild(*this);
+}
+
+void
+LedgerStateRoot::Impl::commitChild(LedgerStateRoot const& self)
+{
+    auto header = mChild->getHeader();
+    auto entries = mChild->getEntries();
+
+    mBestOffersCache->clear();
```
vogel: should we also clear that on rollbackChild?

src/herder/Upgrades.cpp
vogel: I think option 2. is OK, especially as prepareLiabilities is only called when we already have header instance.

src/ledger/LedgerState.cpp
```diff
@@ -0,0 +1,1311 @@
...
+            }
+            else
+            { // Existed in a previous LedgerState
+                mEntry[key] = nullptr;
+            }
+        }
+    }
+    mHeader = mChild->getHeader();
+
+    mChild = nullptr;
```
marta-lokhova: When is the object mChild points to supposed to be destroyed? I see that it's a raw pointer, is there a particular reason for this choice?

src/bucket/BucketApplicator.cpp
jonjove: This should be part of `LedgerStateRoot`, fixed.

src/ledger/LedgerState.cpp
jonjove: I don't like the name `seal` because it doesn't indicate that it will update the actual data stored in the `LedgerState`. Updated to `sealAndMaybeUpdateLastModified` since it accurately describes what this function does.

src/herder/TxSetFrame.cpp
```diff
@@ -281,19 +290,14 @@ TxSetFrame::trimInvalid(Application& app,
 bool
 TxSetFrame::checkValid(Application& app)
 {
-    // Establish read-only transaction for duration of checkValid
-    soci::transaction sqltx(app.getDatabase().getSession());
```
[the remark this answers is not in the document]
marta-lokhova: Also,  I see the same code in `LoadGenerator.cpp:194` (not removed) - is it needed there?

src/ledger/LedgerState.cpp
```diff
@@ -135,24 +186,23 @@ LedgerState::Impl::commit(Identifier id)
...
+        if (iter.entryExists())
         {
-            mEntry[key] = std::make_shared<LedgerEntry>(*kv.second);
+            mEntry[key] = std::make_shared<LedgerEntry>(iter.entry());
         }
         else
         {
-            if (getNewestVersion(self, key).ledgerState == &self &&
-                !mParent.getNewestVersion(key).entry)
+            // TODO(jonjove): Check that this is correct
```
MonsieurNicolas: remove TODO; I think this is correct too

src/ledger/LedgerState.cpp
```diff
@@ -906,58 +980,67 @@ LedgerStateRoot::Impl::addChild(AbstractLedgerState& child)
...
 {
     auto header = mChild->getHeader();
-    auto entries = mChild->getEntries();
 
     mBestOffersCache->clear();
 
     try
     {
-        for (auto const& kv : entries)
+        for (auto iter = mChild->getEntryIterator(); (bool)iter; ++iter)
```
MonsieurNicolas: much better now (no copy)

src/ledger/LedgerState.cpp
```diff
@@ -1168,66 +1360,82 @@ LedgerStateRoot::getAllOffers()
...
-            loadBestOffers(buying, selling, BATCH_SIZE, offers.size());
-        if (newOffers.size() < BATCH_SIZE)
+        std::list<LedgerEntry>::const_iterator newOfferIter;
+        try
         {
-            cached.allLoaded = true;
+            newOfferIter =
+                loadBestOffers(offers, buying, selling, BATCH_SIZE, offers.size());
+        }
+        catch (...)
```
MonsieurNicolas: same thing than earlier: use proper `catch(std::exception)` instead of just `catch(...)`

src/ledger/LedgerStateImpl.h
```diff
@@ -29,156 +29,260 @@ class LedgerState::Impl
...
     LedgerStateEntry create(LedgerState& self, LedgerEntry const& entry);
 
+    // deactivate has the strong exception safety guarantee
     void deactivate(LedgerKey const& key);
 
+    // deactivateHeader has the strong exception safety guarantee
     void deactivateHeader();
 
+    // erase has the basic exception safety guarantee. If it throws an exception
+    // other than LedgerStateRootFatalError, then
```
MonsieurNicolas: comment applies to other methods:
as a caller, I am not sure what those guarantees really mean and which recovery steps I should (or can) take. `LedgerStateRootFatalError` at least is clear ("you're dead");
in particular: if a cache is out of sync - does that mean that I am hosed?

src/ledger/LedgerStateOfferSQL.cpp
```diff
@@ -278,6 +277,56 @@ LedgerStateRoot::Impl::loadOffers(StatementContext& prep) const
     return offers;
 }
 
+std::list<LedgerEntry>::const_iterator
+LedgerStateRoot::Impl::loadOffers(StatementContext& prep,
+                                  std::list<LedgerEntry>& offers) const
+{
+    auto iterNext = --offers.cend();
```
MonsieurNicolas: I think this is UB if `offers` is empty

src/ledger/LedgerManagerImpl.cpp
```diff
@@ -991,6 +991,10 @@ LedgerManagerImpl::applyTransactions(std::vector<TransactionFramePtr>& txs,
         {
             throw e;
         }
+        catch (LedgerStateRootFatalError& e)
+        {
+            throw e;
```
MonsieurNicolas: just use `throw;` syntax to preserve the stack in the core file

MonsieurNicolas: (should have done that for the invariant exception as well)

src/ledger/LedgerState.cpp
```diff
@@ -747,37 +819,60 @@ LedgerState::Impl::load(LedgerState& self, LedgerKey const& key)
...
+    // can throw and the LedgerStateEntry destructor requires that mActive
+    // contains key. LedgerStateEntry constructor does not throw so this is
+    // still exception safe.
     mActive.emplace(key, toEntryImplBase(impl));
-    return LedgerStateEntry(impl);
+    LedgerStateEntry lse(impl);
+
+    // TODO(jonjove): Is std::map::operator[] exception safe?
+    // std::shared_ptr assignment is noexcept
+    mEntry[key] = current;
```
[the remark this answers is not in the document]
jonjove: Response to first comment is the same as above.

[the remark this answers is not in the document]
jonjove: Handling issues like the one mentioned in your second comment is a big part of what was added in the most recent commits. This possibility is already accounted for: see the comment on lines 824-827. Specifically, the destructor of `LedgerStateEntry` deactivates the key. Since the `LedgerStateEntry` is created before `mEntry` is updated, it is guaranteed that either both `mEntry` and `mActive` are updated or neither is updated (assuming that `operator[]` has the strong exception safety guarantee, as questioned in the `TODO`).

src/ledger/LedgerStateImpl.h
jonjove: Good point, need to clarify the language here. A guarantee like "the entry cache may be, but is not guaranteed to be, cleared" is intended to mean that you will get the same results but the exact data access pattern may be different (cache miss instead of cache hit on some future query) assuming that the change in data access pattern does not lead to some spurious exception (which could happen if a key was dropped from the cache due to a thrown exception and then a database query failed after the subsequent cache miss). The overall idea here was that functions that offers exception safety guarantees should be safe to continue using after a thrown exception _unless_ the thrown exception was `LedgerStateRootFatalError` in which case there is no way to proceed.

src/ledger/LedgerStateOfferSQL.cpp
jonjove: You are indeed correct, simplified the implementation to avoid this.

src/ledger/LedgerState.cpp
```diff
@@ -236,17 +278,25 @@ LedgerState::Impl::create(LedgerState& self, LedgerEntry const& entry)
...
```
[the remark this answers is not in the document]
MonsieurNicolas: My read is that we have strong guarantees:

src/ledger/LedgerState.cpp
```diff
@@ -915,31 +1031,95 @@ LedgerState::Impl::unsealHeader(
...
-                entry->lastModifiedLedgerSeq = mHeader.ledgerSeq;
+                entry->lastModifiedLedgerSeq = mHeader->ledgerSeq;
             }
         }
     }
+    return entries;
+}
+
+void
+LedgerState::Impl::maybeUpdateLastModifiedThenInvokeThenSeal(
```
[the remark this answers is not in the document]
jonjove: The name is clunky, but I think it reflects what the function actually does: `maybeUpdateLastModified` then invoke `f` then `seal`.

src/ledger/LedgerState.cpp
```diff
@@ -1217,7 +1211,7 @@ LedgerStateRoot::Impl::commitChild(EntryIterator iter)
     }
     catch (...)
     {
-        throw LedgerStateRootFatalError("LedgerStateRoot could not commit to database");
```
MonsieurNicolas: consider some logging (otherwise we'll just see a crash with no context at all)

src/ledger/LedgerState.cpp
```diff
@@ -244,8 +243,7 @@ LedgerState::Impl::commitChild(EntryIterator iter)
     }
     catch (...)
     {
-        mEntry.swap(previousEntries);
```
MonsieurNicolas: same thing (logging)

src/transactions/AllowTrustOpFrame.cpp
```diff
@@ -133,11 +133,8 @@ AllowTrustOpFrame::doApply(Application& app, AbstractLedgerState& ls)
                 throw std::runtime_error("Offer not buying or selling expected asset");
             }
 
-            LedgerStateEntry trustAcc;
-            TrustLineWrapper buyingTrust;
-            TrustLineWrapper sellingTrust;
-            releaseLiabilities(ls, header, offer, trustAcc, buyingTrust,
-                               sellingTrust);
+            releaseLiabilities(ls, header, offer);
+            auto trustAcc = stellar::loadAccount(ls, mAllowTrust.trustor);
```
MonsieurNicolas: wouldn't that throw if `mAllowTrust.trustor` is activated by `releaseLiabilities` or from a previous offer in the loop?

src/transactions/ManageOfferOpFrame.cpp
```diff
@@ -165,16 +165,8 @@ ManageOfferOpFrame::computeOfferExchangeParameters(
...
-    TrustLineWrapper wheatLineA;
-    if (sheep.type() != ASSET_TYPE_NATIVE)
-    {
-        sheepLineA = stellar::loadTrustLine(ls, getSourceID(), sheep);
-    }
-    if (wheat.type() != ASSET_TYPE_NATIVE)
-    {
-        wheatLineA = stellar::loadTrustLine(ls, getSourceID(), wheat);
-    }
+    auto sheepLineA = loadTrustLineIfNotNative(ls, getSourceID(), sheep);
```
MonsieurNicolas: I was going to ask in the previous commit :)

src/ledger/LedgerState.cpp
```diff
@@ -1364,6 +1364,11 @@ std::shared_ptr<LedgerEntry const>
 LedgerStateRoot::Impl::getBestOffer(Asset const& buying, Asset const& selling,
                                     std::set<LedgerKey>& exclude)
 {
+    // Note: Elements of mBestOffersCache are properly sorted lists of the best
```
MonsieurNicolas: I know we don't have anything like that right now, but we could have post conditions enabled in debug builds: `dbgAssert( isSorted(...) )`

src/ledger/LedgerStateImpl.h
```diff
@@ -0,0 +1,259 @@
+// Copyright 2018 Stellar Development Foundation and contributors. Licensed
```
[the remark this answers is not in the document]
jonjove: Done.

src/transactions/OfferExchange.cpp
```diff
@@ -1037,25 +1029,25 @@ crossOfferV10(AbstractLedgerState& ls, LedgerStateEntry& sellingWheatOffer,
...
+        else
+        {
+            acquireLiabilities(lsInner, header, sellingWheatOffer);
+        }
+        lsInner.commit();
     }
 
     // Note: Cannot use sellingWheatOffer or offer (which is a reference) since
-    // it may have been erased at this point.
+    // it is not active (and may have been erased) at this point.
```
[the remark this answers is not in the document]
jonjove: Clarified this comment.

src/catchup/ApplyLedgerChainWork.cpp
jonjove: There are many uses of `getLastClosedLedgerNum` (I'm seeing 67 on my local branch) and I think changing this would just make this PR even messier. I support the change but let's do it in a separate clean-up PR.

src/herder/HerderTests.cpp
jonjove: I think I'd rather leave it as is for now since it makes it very explicit that a `LedgerState` is being used (this is important for reasoning about deactivation and nesting of ledger states).

src/ledger/LedgerState.cpp
jonjove: I agree the naming is a little strange, but I chose this name since `rollbackChild` is to `rollback` as `commitChild` is to `commit`. I'm not sure if `clearChild` or `resetChild` are much better since they also imply that they are doing something to the child, which they are not.

src/ledger/LedgerStateTests.cpp
```diff
@@ -0,0 +1,566 @@
...
+        }
+        validate(root, lh, {le});
+    }
+
+    SECTION("key exists but is not active")
+    {
+        {
+            LedgerState ls(root, false);
+            ls.create(le);
+            REQUIRE_THROWS_AS(ls.create(le), std::runtime_error);
```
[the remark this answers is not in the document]
jonjove: Done.

src/ledger/TrustLineWrapper.cpp
```diff
@@ -194,6 +194,16 @@ TrustLineWrapper::deactivate()
     mImpl.reset();
 }
 
+std::unique_ptr<TrustLineWrapper::AbstractImpl> const&
+TrustLineWrapper::getImpl() const
+{
+    if (!(*this))
+    {
+        throw std::runtime_error("TrustLineWrapper not active");
```
vogel: I think this is more a std::logic_error than runtime_error

src/ledger/LedgerState.cpp
```diff
@@ -1369,12 +1367,16 @@ LedgerStateRoot::Impl::getBestOffer(Asset const& buying, Asset const& selling,
...
-    if (!mBestOffersCache->exists(cacheKey))
+    if (!mBestOffersCache.exists(cacheKey))
     {
-        putInBestOffersCache(cacheKey, {{}, false});
+        putInBestOffersCache(cacheKey, emptyCacheEntry);
     }
-    auto& cached = getFromBestOffersCache(cacheKey);
+
+    // Have to check again in case the max cache size is 0 or the put throws
+    auto& cached = mBestOffersCache.exists(cacheKey)
```
vogel: That means that `getFromBestOffersCache` can not throw anymore (as it is only used here).
What about making it `getFromBestOffersCache(key, BestOffersCacheEntry &default)`
and using as `auto& cached = getFromBestOffersCache(cacheKey, emptyCacheEntry)`

src/ledger/TrustLineWrapper.cpp
jonjove: Agreed this does fall more into the category of `std::logic_error`. The same would apply to most of the other errors that can be thrown by the new data layer actually. But right now we never use `std::logic_error` in stellar-core, although I am not sure why. I think I'll leave this for now and maybe at some point we can try to figure out which exceptions should fall into which categories.

src/ledger/LedgerState.cpp
```diff
@@ -1199,8 +1205,14 @@ LedgerStateRoot::Impl::commitChild(EntryIterator iter)
         mTransaction->commit();
         mDatabase.clearPreparedStatementCache();
     }
+    catch (std::exception& e)
+    {
+        std::fprintf(stderr, "fatal error during commit to LedgerStateRoot: %s\n", e.what());
```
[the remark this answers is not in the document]
jonjove: Done, and also changed `abort` to `dbgAbort`.

src/ledger/LedgerState.cpp
```diff
@@ -1506,12 +1497,18 @@ LedgerStateRoot::Impl::getInflationWinners(size_t maxWinners, int64_t minVotes)
     }
     catch (std::exception& e)
     {
-        std::fprintf(stderr, "fatal error when getting inflation winners from LedgerStateRoot: %s\n", e.what());
+        printErrorAndAbort("fatal error when getting inflation winners from LedgerStateRoot: ", e.what());
```
MonsieurNicolas: the c++ 11 way to do this is to simply declare that function with `noreturn`:
```c++
[[noreturn]]
void
printErrorAndAbort(...)
```

MonsieurNicolas:
r+ 2406720e

MonsieurNicolas:
r+ 482a12fbae84c4d4e88fcd0047227bbf6c979226

MonsieurNicolas:
r+ c20a90ecd9d523b279a2fd2705dc76c07243dc74

MonsieurNicolas:
r+ a4638ca0

MonsieurNicolas:
r+ 3509cf1

PR #1790, opened 2018-09-12
add --disable-bucket-gc argument for bucket recovery

vogel:
This new option allows for bucket recovery. In case where non-checkpoint bucket is required to continue stopped stellar-core, that bucket can be recovered using command-line catchup with this new option.

Imagine a stellar-core node was stopped with last-closed ledger 123456 and it misses bucket `3e44d29cf75b2ceae1dcd58cbb18791c3ab998a89c848b231228db19bcf9c780` that is not a checkpoint-bucket (it does not gets saved to history archives).

To recover that bucket one can run second stellar-core instance and perform following commands on it:
`stellar-core --conf second-instance.conf --newdb`
`stellar-core --conf second-instance.conf --catchup-at 123400 --disable-bucket-gc`
`stellar-core --conf second-instance.conf --catchup-to 123500 --disable-bucket-gc`

Bucket directory of that second instance should now contain all missing buckets files.

### DIFF
--- a/src/bucket/BucketManagerImpl.cpp
+++ b/src/bucket/BucketManagerImpl.cpp
@@ -272,6 +272,11 @@ BucketManagerImpl::getReferencedBuckets() const
 void
 BucketManagerImpl::cleanupStaleFiles()
 {
+    if (mApp.getConfig().DISABLE_BUCKET_GC)
+    {
+        return;
+    }
+
     std::lock_guard<std::recursive_mutex> lock(mBucketMutex);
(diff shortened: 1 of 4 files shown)


MonsieurNicolas:
needs to be rebased @vogel

vogel:
Rebased (new option added only to new command-line syntax).

MonsieurNicolas:
r+ df1d91304a8eecdb524e9d1aded3664b1d3f83bf

Issue #1755, opened 2018-08-14
design cleanup: Work interface is a mess


MonsieurNicolas:
It's quite unclear for implementors what the various methods (onReset, onStart, etc) are really supposed to do, especially when combining Work (what `WorkParent` does for example).

For example, I would expect implementations to follow:
* `onReset` initializes the Work to a pristine state (does NOT create work)
   * this is a common method that is called before start, retry but also failure
   * right now there is no "reset" function that can safely be called regardless of the reason
* `onStart` should be the one creating the initial batch of work (also called on retry as retry triggers in sequence reset, advance, start)

This is illustrated by `WorkParent` that doesn't manage its children at all: it's up to all implementors to remember to call `clearChildren()` on reset for example (or face undefined behavior on retry).

MonsieurNicolas:
I think that some of the implications of this are to ensure that `advance` is the place that centralizes decisions on state transition and "run" would create more work if needed.

marta-lokhova:
I did a bit of investigation of this looking for patterns and how the code it trying to use Work, so here are some thoughts:
* First of all, just for clarity, I'm adding Work state machine represented as a diagram. Below are allowed state transitions we currently have:
![work design notes 1](https://user-images.githubusercontent.com/9428003/44545722-afedc500-a6ca-11e8-9efa-d00d8fb9f6a5.jpg)

* Currently, decision of state transition is in two places: in `advance` and in `onSuccess`. It seems like `onSuccess` was _designed_ to manipulate state. And implementors heavily utilize that: for example, [ApplyBucketsWork](https://github.com/stellar/stellar-core/blob/master/src/catchup/ApplyBucketsWork.cpp#L153) goes back and forth between "RUNNING" and "PENDING" states (depending if it's done applying a level or not). So I don't know if it makes sense to centralize decisions in `advance` only.
* The `WorkParent-Work` relationship is not very intuitive to me; specifically, its separation of responsibilities. I feel like there are two different notions here: 
  1. a state machine that transitions based on certain rules (e.g. what was returned in `onSuccess`, or `advance`; with no dependencies advance is trivial)
  2. a parent that has children and manages them; its state transition rules are therefore more complex, as it needs to check children before proceeding (e.g. now `advance` has some rules)
To illustrate this point, I'll go back to what Work is trying to accomplish, from the code:
```
  1. May depend on other Work before starting
  2. May have parts that can run in parallel, other parts in serial
[...]

Issue #1649, opened 2018-04-09
Core doesn't publish to secondary history archive (e.g. AWS S3) if first archive failed (cache)


oryband:
# Issue type
Bug

# Version found
- 9.1.0
- 9.2.0

# Your environment and setup
- Dockerized Ubuntu 16.04 (https://hub.docker.com/_/ubuntu/)

# Issue description
I've set up two history archive: cache and AWS S3:
```toml
[HISTORY.cache]
get="cp /tmp/stellar-core/history/vs/{0} {1}"
put="cp {0} /tmp/stellar-core/history/vs/{1}"
mkdir="mkdir -p /tmp/stellar-core/history/vs/{0}"

[HISTORY.s3]
get="aws s3 cp s3://my-bucket/{0} {1}"
put="aws s3 cp {0} s3://my-bucket/{1}"
# mkdir isn't necessary for aws cli, 'cp' command takes care of that
```

I've accidentally only initialized (`--newhist`) the S3 archive, and forgot `cache`. Core tried to published to `cache` and failed, and didn't try to publish to S3 afterwards, even though it should have.

## Supporting files

Logs where full of the following lines:
```
cp: cannot stat '/tmp/stellar-core/history/vs/history/00/05/99/history-0005993f.json': No such file or directory
GBT6G [Process WARNING] process 1829 exited 1: cp /tmp/stellar-core/history/vs/history/00/05/99/history-0005993f.json buckets/tmp/history-a05d40f6bb31c070/stellar-history.json
GBT6G [Work WARNING] Scheduling retry #1/5 in 2 sec, for get-remote-file history/00/05/99/history-0005993f.json
```
However, there's was nothing else in the logs indicating publishing to S3 was skipped.

edwardsun888:
I met the same issue, it maybe No permission to do mkdir="mkdir -p /tmp/stellar-core/history/vs/{0}".
but I do not know to solve this problem currently.

benxage:
@edwardsun888 did you find a solution to the problem? It seems unnecessary for there to be an `mkdir` command since s3 automatically creates those folders for you.

oryband:
> I met the same issue, it maybe No permission to do mkdir="mkdir -p /tmp/stellar-core/history/vs/{0}".
> but I do not know to solve this problem currently.

The topic of this issue is not the error that I accidently made, but the fact the Core doesn't publish to secondary history archive (e.g. AWS S3) if first archive failed for whatever reason.

[...]

PR #1801, opened 2018-09-26
New work interface 

marta-lokhova:
{should NOT be merged}

This is an initial prototype of the Work interface described [here](https://github.com/stellar/stellar-core/issues/1755). It includes implementations of the new interface (BasicWork & Work), the scheduler (WorkScheduler), as well as a few examples on how existing works can utilize it (including RunCommandWork, DownloadBucketsWork etc).

For the refactored works, it's better to look at the whole files (as opposed to diffs). I left some of the tests from before I made the breaking changes (like actually renamed the classes) in case it is useful.


src/work/Work.cpp
```diff
@@ -1,437 +1,161 @@
...
+            return true;
+        }
     }
+    return false;
 }
 
-void
-Work::notifyParent()
+std::shared_ptr<Work>
+Work::yieldNextRunningChild()
```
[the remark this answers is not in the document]
marta-lokhova: I totally agree with you, but for this prototype my goal is to exhibit how the interface would roughly work, without focusing on particular optimizations. 
To your point though, this will probably get addressed in the actual implementation.

src/work/Work.cpp
```diff
@@ -1,437 +1,161 @@
...
 
-std::string
-Work::getStatus() const
+BasicWork::State
+Work::onRun()
 {
-    switch (mState)
-    {
-    case WORK_PENDING:
+    if (mScheduleSelf || !anyChildRunning())
```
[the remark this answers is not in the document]
marta-lokhova: I thought about this as well: the original idea was to schedule self once per cycle, but this implies that we'd need to run all children (and we don't really know how many of them we would have unless we impose a limit) before parent can take any action. If we are okay with that, I could definitely update this logic. The disadvantage of "once per child" approach is that it does a lot of redundant work.

src/work/BasicWork.cpp
```diff
@@ -0,0 +1,235 @@
...
+namespace stellar
+{
+
+size_t const BasicWork::RETRY_NEVER = 0;
+size_t const BasicWork::RETRY_ONCE = 1;
+size_t const BasicWork::RETRY_A_FEW = 5;
+size_t const BasicWork::RETRY_A_LOT = 32;
+size_t const BasicWork::RETRY_FOREVER = 0xffffffff;
+
+BasicWork::BasicWork(Application& app, std::string name,
```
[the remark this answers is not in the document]
marta-lokhova: for this comment and many below: a lot of this is almost literally taken from current Work interface (e.g. `getUniqueName`, virtual functions that shouldn't be virtual, all*/any* functions, etc). I agree it needs fixing and was an oversight, and I'll address them with the next iteration of changes.

src/work/BasicWork.cpp
```diff
@@ -0,0 +1,235 @@
...
+        throw std::runtime_error("Unknown Work::State");
+    }
+}
+
+void
+BasicWork::reset()
+{
+    CLOG(DEBUG, "Work") << "restarting " << getName();
+    setState(WORK_RUNNING);
+    onReset();
```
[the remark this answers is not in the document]
marta-lokhova: I'm not sure I agree. Why would we set state to WORK_WAITING? I believe we defined this state as "waiting for an external event" (like adding a new child, or process completion, etc), so this seems like a misuse. (especially for outside classes, as it's not obvious why work transitioned states like this without looking at the implementation). And `setState` actually sets the state only if the new state is different, so I'd rather prefer that.

src/work/BasicWork.cpp
```diff
@@ -0,0 +1,235 @@
...
+    case WORK_FAILURE_FATAL:
+        return "WORK_FAILURE_FATAL";
+    default:
+        throw std::runtime_error("Unknown Work::State");
+    }
+}
+
+void
+BasicWork::reset()
+{
```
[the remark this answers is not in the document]
marta-lokhova: Yes, unfortunately, `reset` needs to be protected as derived classes like work need it when adding a child for example. I could add a comment with an explanation, but ultimately this is an implementer's responsibility to call it correctly. By your logic, `wakeUp` can be called at any time as well, producing a weird state.

src/work/Work.cpp
```diff
@@ -1,437 +1,161 @@
...
-Work::getRetryDelay() const
+void
+Work::onReset()
 {
-    // Cap to 4096sec == a little over an hour.
-    uint64_t m = 2 << std::min(uint64_t(12), uint64_t(mRetries));
-    return std::chrono::seconds(rand_uniform<uint64_t>(1ULL, m));
+    CLOG(DEBUG, "Work") << "resetting  " << getName();
+    assert(allChildrenDone());
+    mNextChild = mChildren.begin();
```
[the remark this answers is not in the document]
marta-lokhova: Good catch!

src/historywork/RunCommandWork.h
```diff
@@ -8,23 +8,28 @@
...
-// method; this way we only run a command _once_ (when it's first
-// scheduled) rather than repeatedly (racing with other copies of itself)
-// when rescheduled.
+/**
+ * This class helps run various commands, that require
+ * process spawning. This work is not scheduled while it's
+ * waiting for a process to exit, and wakes up when it's ready
+ * to be scheduled again.
+ */
 class RunCommandWork : public Work
```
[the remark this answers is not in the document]
marta-lokhova: Regarding comments on `Work` vs `BasicWork`: I like the idea of keeping the work tree consistent and implementers only deriving from Work.

src/work/BasicWork.cpp
```diff
@@ -0,0 +1,235 @@
...
+        CLOG(DEBUG, "Work") << "work " << getName() << " : "
+                            << stateName(mState) << " -> " << stateName(st);
+        mState = st;
+    }
+}
+
+void
+BasicWork::wakeUp()
+{
+    setState(WORK_RUNNING);
```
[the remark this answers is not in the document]
marta-lokhova: I think it should not be necessary, as the parent must be running as well (and hence, can get scheduled). Will update.

src/work/WorkSchedulerImpl.cpp
```diff
@@ -0,0 +1,91 @@
...
+                if (!innerSelf)
+                {
+                    return;
+                }
+                innerSelf->mScheduled = false;
+                innerSelf->crankWork();
+                innerSelf->mNotifyCallback();
+            });
+        }
+        else if (self->getState() == BasicWork::WORK_WAITING &&
```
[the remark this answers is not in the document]
marta-lokhova: Good catch!

src/work/Work.h
```diff
@@ -1,140 +1,73 @@
...
+    bool allChildrenDone() const;
+    bool anyChildRaiseFailure() const;
+    bool anyChildFatalFailure() const;
+    bool anyChildRunning() const;
+    bool hasChildren() const;
 
-  private:
-    VirtualClock::duration getRetryDelay() const;
+  protected:
+    Work(Application& app, std::function<void()> callback, std::string name,
```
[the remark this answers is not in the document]
marta-lokhova: Since parent class throws away `name` when creating a child, I think passing by value and using `std::move` is more appropriate. Updated.

src/work/Work.h
```diff
@@ -1,140 +1,73 @@
...
+         size_t retries = BasicWork::RETRY_A_FEW);
+    void clearChildren();
+    State onRun() final;
+    void onReset() override;
```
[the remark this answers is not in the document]
marta-lokhova: Done.

marta-lokhova:
Thanks for reviewing! I've made some updates, mostly addressing design-related comments. Like I mentioned before, I'm not focusing on particular optimizations/other nits for this iteration, as it will be addressed once we agree on the design.

src/historywork/VerifyBucketWork.cpp
```diff
@@ -16,37 +16,77 @@
...
+}
+
+BasicWork::State
+VerifyBucketWork::doWork()
+{
+    // FIXME (mlo) this a logical problem, as this check shouldn't be here
+    // VerifyBucketWork shouldn't be required to know that it depends on
+    // GetAndUnzipRemoteFileWork. This will be resolved with batching though,
+    // as DownloadBuckets work will take care of when to dispatch download
+    // and verify tasks.
```
marta-lokhova: Btw, this is an example of a more general consequence that we get with this Work design: if a work did not dispatch any children, if shouldn't care if it has any dependent children. In this example, `VerifyBucketWork` depends on `GetAndUnzipRemoteFileWork`, because higher level work `DownloadBucketsWork` created this dependency. I think as we move on from this, we should have any parent work (in this case `DownloadBucketsWork`, or maybe even a new class `DownloadAndVerifyBucketWork`) be in charge of adding and tracking status of its children.

src/work/BasicWork.cpp
```diff
@@ -0,0 +1,248 @@
...
+    onReset();
+}
+
+void
+BasicWork::waitForRetry()
+{
+    if (mRetryTimer)
+    {
+        std::runtime_error(
```
jonjove: Missing `throw`.

src/work/WorkScheduler.cpp
```diff
@@ -0,0 +1,67 @@
...
+        auto self = weak.lock();
+        if (!self)
+        {
+            return;
+        }
+        self->mScheduled = false;
+        self->crankWork();
+        if (self->getState() == WORK_RUNNING)
+        {
+            self->wakeUp();
```
jonjove: `wakeUp` calls `setState`, then `onWakeUp`, then `mNotifyCallback` if it wasn't already running (although as mentioned in `BasicWork::wakeUp` I think this should be changed slightly). For `WorkScheduler`, this is equivalent to `setState` then `onWakeUp` since `mNotifyCallback` is always empty. But here you are checking if it is already `WORK_RUNNING`, in which case `wakeUp` is equivalent to `onWakeUp`. May as well just call `onWakeUp` here directly.

src/work/Work.h
```diff
@@ -1,140 +1,73 @@
...
+    bool anyChildFatalFailure() const;
+    bool anyChildRunning() const;
+    bool hasChildren() const;
 
-  private:
-    VirtualClock::duration getRetryDelay() const;
+  protected:
+    Work(Application& app, std::function<void()> callback, std::string name,
+         size_t retries = BasicWork::RETRY_A_FEW);
+    void clearChildren();
```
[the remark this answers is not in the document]
marta-lokhova: Investigated this more and talked to @MonsieurNicolas, and it appears that the original reason behind calling `clearChildren` in `Work*` destructors in an interesting case of `shared_ptr` vs resource de-allocation (this commit might be helpful: https://github.com/stellar/stellar-core/pull/1355). Essentially, it seems like a child Work class was able to get a `shared_ptr` from `weak_ptr` to its parent during that parent's destruction and call its methods, causing a crash. Looks like the fix was to `clearChildren` first to prevent calls to parent. 
Now, with the new design, we'd like to avoid this altogether, and I think it is possible with proper shutdown of `WorkScheduler`. The idea is to provide a guarantee that no work would not be destroyed until it reached complete state and reset (which implies cleaning up references to its children). This is basically an abort functionality. While the goal is to have it implemented once the implementation for the new work design is in place, for this first iteration, we decided to try and add WORK_DESTRUCTING state, and perform a no-op in `wakeUp` (as any child work would have a weak_ptr to parent in its callback and could call `wakeUp` on its half-destroyed parent).

src/historywork/RunCommandWork.h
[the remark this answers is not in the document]
marta-lokhova: Thinking about this more, I think this is related to a `FIXME` in `VerifyBucketWork::doWork`. Let's say we follow your suggestion - this means that outside work won't be allowed to `addWork` to some other works (that inherit directly from `BasicWork`, for example). In the `VerifyBucketWork` example, `DownloadBucketsWork` won't be allowed to add chains of GetRemoteFileWork->VerifyBucketWork, because `GetRemoteFileWork` would inherit from `RunCommandWork` (and hence from `BasicWork`). I think this pushes for a different mindset regarding how we add work. We either make parent `DownloadBucketsWork` worry about when to properly dispatch download and verify works (creating a flatter structure, as both download and verify works would be children of DownloadBucketsWork), or create an abstraction for "download and verify a bucket", and it would be a child of `DownloadBucketsWork`.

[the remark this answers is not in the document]
jonjove: I have also given this topic more thought. Both of my earlier comments here were viewing this as an implementation issue and not a design issue. After considering your latest comment, it is clear to me that this is in fact a design issue. The root cause of this issue is that `addWork` is part of the public interface of `Work`. I do not believe that this is desirable, since it blurs the meaning of child work.

Suppose that C is a child work of W. Does this imply that W depends on the output of C, or that W is composed from C? The first would indicate that the relationship between W and C is purely one of ordering: C must complete before W can proceed. The second would indicate that W manages C. For me, the answer is the second. The FIXME in `VerifyBucketWork::doWork` is a reflection of the logical issues associated with using the child work relationship to model a dependence on output: the derived class is not aware that it has any children (since the call to `addWork` is not virtual and has no mechanism to inform the derived class of the new child), yet it still must wait for them to finish. Specifically, when `VerifyBucketWork` is used as part of `DownloadBucketsWork`, we observe that it is-implemented-in-terms-of `GetAndUnzipRemoteFileWork` yet it is unaware of this relationship. (For more information on is-implemented-in-terms-of, see items 38 and 39 of Scott Meyers “Effective C++” Third Edition.)

The relationship is-implemented-in-terms-of should in this case be modeled using explicit composition. `VerifyBucketWork` is already an implicit composition, since the base class `Work` is managing the relationship to `GetAndUnzipRemoteFileWork`. To fix this, `addWork` should be `protected` rather than `public`. Then the derived class will know about all of its children, since it will have added them. In this case, the `VerifyBucketWork` concept could be broken down into its components:
- Get (and unzip) the remote file
- Verify the bucket

The first step is already implemented as `GetAndUnzipRemoteFileWork` (itself a composite work, as the name indicates). The second part could be implemented as `VerifyBucketWork` which would now only verify a bucket that already exists on disk. The composite could be named `DownloadAndVerifyBucketWork` which would first execute `GetAndUnzipRemoteFileWork`, then execute `VerifyBucketWork`. In this approach, there is no need for any external calls to `addWork`. An added benefit of this approach is that it will encourage implementors to distill work down to largely independent components, which should lead to much simpler and more modular implementations. It should be noted that this will come at the cost of implementors of composite work classes needing to manually manage the execution order among their children (some composite work classes are already implemented in this manner).

Of course, there exist types of work where new work should be added externally. For example, it should always be possible to add arbitrary child work to the `WorkScheduler`, since it is a top-level object only responsible for the scheduling of whatever work is added as its child. This could be achieved by adding `WorkScheduler::addWork` to the public interface of `WorkScheduler` which simply forwards to the protected function `Work::addWork`. More sophisticated examples could be built in an analogous manner, by exposing functions in the public interface which only allow a limited forwarding to `Work::addWork` (for example, the work to add must be of a certain type chosen by the class but the caller can choose the parameters).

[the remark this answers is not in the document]
jonjove: @MonsieurNicolas any thoughts about the proposal in the previous comment?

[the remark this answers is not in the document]
marta-lokhova: To add to what Jon posted and my previous statement that moving to this design comes with a cost for implementers, I'd like to add a few more thoughts:
* This means that Work would need to check on status of each child, and react appropriately to each possible state of the child (before if parent got to `onRun`, all of its children were guaranteed to be in successful state). 
* A positive thing I found about this is that parent can control if it wants to keep going (currently if a single child fails, parent fails by default). This would resolve issues like this one https://github.com/stellar/stellar-core/issues/1649, for example.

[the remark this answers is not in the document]
MonsieurNicolas: yeah I agree: using parent/child this way is not desirable. The only reason for deriving from `RunCommand` for example, would be to make some new kind of `RunCommand`, but in our code base I don't think we ever need this (as all we do is sequencing, maybe `batchWork` falls under a legitimate case of derivation).
So in general, we should derive only from `Work` or `BasicWork`.

As for `addWork`: yeah we can make it protected to force proper composition.
In some cases though, instead of creating single use `BasicWork` (or `Work`) objects, it might be enough to just perform the sequencing directly in `doWork` based on booleans/state tracking variable.

As for the other questions...

> This means that Work would need to check on status of each child, and react appropriately to each possible state of the child (before if parent got to onRun, all of its children were guaranteed to be in successful state).

yes, this is exactly what we discussed before. The sequencing is explicit in every `doWork` implementation, which is why providing some helper functions like `allChildrenSuccessful()` can help avoid writing the same checks everywhere. The more complicated `Work` already follows this pattern anyways.

src/work/Work.h
```diff
@@ -21,67 +20,84 @@ namespace stellar
...
+
+    friend class WorkSequence;
+};
+
+    /*
+     * WorkSequence is a helper class, that implementers can use if they
+     * wish to enforce the order of work execution.
+     */
+class WorkSequence : public Work
```
marta-lokhova: It seems like the sequence pattern is used frequently in the code (specifically, don't proceed to next work until current work completed successfully), e.g. `PutSnapshotFilesWork`, `PutHistoryArchiveStateWork` or `CatchupWork`. It might be useful to have a mechanism like WorkSequence that implementers can use to enforce execution order. We have previously discussed with @MonsieurNicolas that one of the parameters for this class could be a vector of shared_ptrs to `BasicWork`, however upon creation of BasicWork objects, It wasn't possible to construct a proper notification callback, because WorkSequence instance did not exist at that point. That led to some weird manipulations with `mNotifyCallback`, which was intended to be private and only usable by `BasicWork`. Additionally, it required iteration over mChildren, which we already do in `Work::yieldNextRunningChild`. So instead, I've slightly modified `Work` class to allow for serial execution. With parallel execution, we iterate over children, then parent, whereas with serial execution we keep yielding same child while it's running, scheduling parent only after the child has reached a complete state, so that the parent can decide how to proceed. Some example of `WorkSequence` usage are `PutHistoryArchiveStateWork` and `GzipAndPutFilesWork`.

src/catchup/DownloadBucketsWork.cpp
```diff
@@ -45,14 +45,14 @@ DownloadBucketsWork::doWork()
...
-        {
-            return WORK_WAITING;
-        }
         else if (anyChildRaiseFailure())
         {
             return WORK_FAILURE_RETRY;
         }
+        else if (!anyChildRunning())
+        {
+            return WORK_WAITING;
```
marta-lokhova: Note that WORK_WAITING state is also utilized here: this is useful in cases where some children have completed and some are waiting; parent will also go into waiting state, and will be scheduled when any of its children wakes it up.

src/work/BasicWork.h
```diff
@@ -0,0 +1,114 @@
...
+    // all the references (including children) are properly cleaned up for
+    // safe destruction. `onRun` performs necessary work logic, and hints the
+    // next state transition. `onSuccess` is called when work transitions into
+    // WORK_SUCCESS state. Similarly, `onFailure*` methods are called upon
+    // transitioning into appropriate failure state. An example usage would be
+    // cleaning up side effects, like downloaded files in these methods. If you
+    // want to force failure (and reset / retry) you can return
+    // WORK_FAILURE_RETRY or WORK_FAILURE_RAISE. After a retry count is
+    // passed, WORK_FAILURE_RETRY means WORK_FAILURE_RAISE anyways.
+    // WORK_FAILURE_FATAL is equivalent to WORK_FAILURE_RAISE passed up in
```
vogel: Is WORK_FAILURE_FATAL removed now?

src/work/BasicWork.cpp
```diff
@@ -0,0 +1,263 @@
...
+        << " sec, for " << getName();
+    setState(WORK_WAITING);
+    mRetryTimer->async_wait(
+        [weak]() {
+            auto self = weak.lock();
+            if (!self)
+            {
+                return;
+            }
+            self->mRetries++;
```
vogel: may be worth to assert here that state is still WORK_WAITING, just in case

src/work/BasicWork.cpp
```diff
@@ -0,0 +1,263 @@
...
+    , mNotifyCallback(std::move(callback))
+    , mName(std::move(name))
+    , mMaxRetries(maxRetries)
+{
+    reset();
+}
+
+BasicWork::~BasicWork()
+{
+    mState = WORK_DESTRUCTING;
```
vogel: At this point, lifetime of BasicWork already ended, so no external entity should be able to access mState.
I see it is only used in wakeUp which is called mainly from exit.async_wait() calls.

I think we only execute asio jobs in main thread, so it looks like it may not be possible to call wakeUp after destructor is called (as we always use weak_ptrs to Work items in such cases).

src/work/BasicWork.cpp
```diff
@@ -0,0 +1,263 @@
...
+BasicWork::getState() const
+{
+    return mState;
+}
+
+void
+BasicWork::setState(BasicWork::State st)
+{
+    auto maxR = getMaxRetries();
+    if (st == WORK_FAILURE_RETRY && (mRetries >= maxR))
```
vogel: I think we should do if (st != mState) first here, as setting WORK_FAILURE_RETRY when the state is already WORK_FAILURE_RETRY should not increase mRetries, as no real jobs was done in between.

We could even add assert to check if someone is setting WORK_FAILURE_RETRY on Work that is already an WORK_FAILURE_RETRY as it is probably a design error to do so.

src/work/Work.h
```diff
@@ -1,140 +1,102 @@
...
-
   public:
-    static size_t const RETRY_NEVER;
-    static size_t const RETRY_ONCE;
-    static size_t const RETRY_A_FEW;
-    static size_t const RETRY_A_LOT;
-    static size_t const RETRY_FOREVER;
-
-    enum State
+    enum Execution
```
vogel: I like it very much!

src/work/Work.h
```diff
@@ -1,140 +1,102 @@
...
-        WORK_COMPLETE_FAILURE,
-        WORK_COMPLETE_FATAL
-    };
+    virtual ~Work();
 
-    Work(Application& app, WorkParent& parent, std::string uniqueName,
-         size_t maxRetries = RETRY_A_FEW);
+    // Children status helper methods
+    bool allChildrenSuccessful() const;
+    bool allChildrenDone() const;
```
vogel: I think those deserve separate implementation for SEQUENTIAL work

src/work/BasicWork.h
marta-lokhova: yeah, I should remove the old comment (Btw, the reason WORK_FAILURE_FATAL was removed is because now parent can decide what it wants to do based on the state of its children, so this state seemed redundant).

src/work/BasicWork.cpp
marta-lokhova: I think the particular case you mentioned is an example of a more general set of restrictions/checks that we should enforce. For example, the allowed transitions should be:
* RUNNING->SUCCESS and RUNNING->FAILURE_RAISE
* RUNNING->RUNNING and RUNNING<->WAITING
* RUNNING->RETRY and RETRY->WAITING.

I agree that asserts would be helpful to spot weird behaviors. Will update.

src/work/BasicWork.h
```diff
@@ -0,0 +1,114 @@
...
+ *  BasicWork manages all state transitions via `crankWork` as well as retries.
+ *  While customers can trigger cranking and check on BasicWork's status,
+ *  it is implementers' responsibility to implement `onRun`, that
+ *  hints the next desired state.
+ */
+class BasicWork : public std::enable_shared_from_this<BasicWork>,
+                  private NonMovableOrCopyable
+{
+  public:
+    static size_t const RETRY_NEVER;
```
MonsieurNicolas: can we use an enum for those instead?

src/work/BasicWork.h
```diff
@@ -0,0 +1,114 @@
...
+    static size_t const RETRY_ONCE;
+    static size_t const RETRY_A_FEW;
+    static size_t const RETRY_A_LOT;
+    static size_t const RETRY_FOREVER;
+
+    enum State
```
MonsieurNicolas: use c++11 syntax for enum to make it "type safe".

need a comment that explains what those states are, probably split into:
* as a user of a BasicWork, what should I expect from those states (when I query the object with `getState`)
* as an implementor, when should I return any particular value?

After writing this, it then might be more clear that we need 2 enums and not one:
* possible state a BasicWork can be, there might be only one `WORK_FAILURE`
* transitional states (used internally, such as `WORK_FAILURE_RETRY`)

MonsieurNicolas: Also, wrt "terminal" states versus temporary ones, we should be consistent naming them (all the temporary states end with `-ING`, terminal states should be `DONE_XYZ`).

src/work/BasicWork.h
```diff
@@ -0,0 +1,114 @@
...
+    {
+        WORK_RUNNING,
+        WORK_SUCCESS,
+        WORK_WAITING,
+        WORK_FAILURE_RETRY,
+        WORK_FAILURE_RAISE,
+        WORK_DESTRUCTING
+    };
+
+    BasicWork(Application& app, std::function<void()> callback,
```
MonsieurNicolas: `callback` is not a good name, as it's related to `wakeUp`, maybe `wakeUpCall` gives a bit more of a clue of what it is?

src/work/BasicWork.h
```diff
@@ -0,0 +1,114 @@
...
+
+    // Main method for state transition, mostly dictated by `onRun`
+    void crankWork();
+
+    // TODO (mlo) look into uses of reset in current interface
+    // Proper cleanup needs to happen at some point, and `reset`
+    // might not be the best place. For instance, `GetRemoteFileWork`
+    // calls std::remove is `onReset`, so an attempt to remove a non-existent
+    // file happens every time a child is added, which is inefficient.
+    void reset();
```
MonsieurNicolas: why is `reset` in the public interface? Should be private

src/work/BasicWork.h
```diff
@@ -0,0 +1,114 @@
...
+    // want to force failure (and reset / retry) you can return
+    // WORK_FAILURE_RETRY or WORK_FAILURE_RAISE. After a retry count is
+    // passed, WORK_FAILURE_RETRY means WORK_FAILURE_RAISE anyways.
+    // WORK_FAILURE_FATAL is equivalent to WORK_FAILURE_RAISE passed up in
+    // the work chain - when WORK_FAILURE_FATAL is raised in one work item,
+    // all work items that leaded to this one will also fail without retrying.
+
+    virtual void onReset();
+    virtual State onRun() = 0;
+    virtual void onFailureRetry();
```
MonsieurNicolas: do we really need the `onFailureXXX` methods? it seems that this creates an arbitrary split of the `onRun` logic.

src/work/BasicWork.h
```diff
@@ -0,0 +1,114 @@
...
+    // A helper method that implementers can use if they plan to
+    // utilize WAITING state. This tells the work to return to RUNNING
+    // state, and propagate the notification up to the scheduler
+    // An example use of this would be RunCommandWork:
+    // a timer is used to async_wait for a process to exit,
+    // with a call to `wakeUp` upon completion.
+    void wakeUp();
+
+    // Default wakeUp callback that implementers can use
+    std::function<void()> wakeUpCallback();
```
MonsieurNicolas: there is no need for this default call back like this: `std::function` can take a `nullptr` value

src/work/BasicWork.cpp
MonsieurNicolas: @vogel is right: setting to `WORK_DESTRUCTING` only works when doing it from a derived class, so you probably need a special method to do just that (typically people call it `shutdown`)

src/work/BasicWork.cpp
```diff
@@ -0,0 +1,262 @@
...
+    default:
+        throw std::runtime_error("Unknown Work::State");
+    }
+}
+
+void
+BasicWork::reset()
+{
+    CLOG(DEBUG, "Work") << "resetting " << getName();
+    setState(WORK_RUNNING);
```
MonsieurNicolas: why do we call `setState` from reset? It should be the other way:
on certain state transitions (handled by `setState`), we should be calling `reset`.

MonsieurNicolas: you also need to cancel `mRetryTimer` in reset

src/work/BasicWork.cpp
```diff
@@ -0,0 +1,262 @@
...
+    std::weak_ptr<BasicWork> weak = shared_from_this();
+    auto t = getRetryDelay();
+    mRetryTimer->expires_from_now(t);
+    CLOG(WARNING, "Work")
+        << "Scheduling retry #" << (mRetries + 1) << "/" << mMaxRetries
+        << " in " << std::chrono::duration_cast<std::chrono::seconds>(t).count()
+        << " sec, for " << getName();
+    setState(WORK_WAITING);
+    mRetryTimer->async_wait(
+        [weak]() {
```
MonsieurNicolas: you need to setup the callback with error code `asio::error_code const& e`, that way you can properly handle cancellation: right now if for some reason the timer gets cancelled, you will not properly "wake up"

src/work/BasicWork.cpp
```diff
@@ -0,0 +1,262 @@
...
+            auto self = weak.lock();
+            if (!self)
+            {
+                return;
+            }
+            self->mRetries++;
+            self->mRetryTimer = nullptr;
+            self->wakeUp();
```
MonsieurNicolas: this is wrong:
* first you should do whatever you need to do, which, in this case is "transition to RUNNING", and then wakeup.
* wakeup should be called by `setState` as the expectation is that if you get out of `WAITING` state, you call it

src/work/BasicWork.cpp
```diff
@@ -0,0 +1,262 @@
...
+BasicWork::State
+BasicWork::getState() const
+{
+    return mState;
+}
+
+void
+BasicWork::setState(BasicWork::State st)
+{
+    auto maxR = getMaxRetries();
```
MonsieurNicolas: as per my previous comments, `setState` should be the place that deals with state transitions

src/work/BasicWork.h
MonsieurNicolas: do we also need `onSuccess` I am not sure what problem this is solving?

src/work/BasicWork.cpp
```diff
@@ -0,0 +1,262 @@
...
+
+void
+BasicWork::crankWork()
+{
+    assert(!isDone() && mState != WORK_WAITING);
+
+    auto nextState = onRun();
+    setState(nextState);
+
+    switch (mState)
```
MonsieurNicolas: this is really part of `setState`

src/work/BasicWork.cpp
```diff
@@ -0,0 +1,262 @@
...
+    {
+    case WORK_SUCCESS:
+        onSuccess();
+        onReset();
```
MonsieurNicolas: what we should call is `reset` not `onReset`

src/work/BasicWork.cpp
```diff
@@ -0,0 +1,262 @@
...
+        break;
+    case WORK_FAILURE_RAISE:
+        onFailureRaise();
+        onReset();
+        break;
+    case WORK_FAILURE_RETRY:
+        onFailureRetry();
+        waitForRetry();
```
MonsieurNicolas: why not reset before retrying?

src/work/BasicWork.h
marta-lokhova: It's convenient for things like cleaning up side effects, marking metrics, logging, reporting progress (these are some of the use cases today). I don't see a problem having these -- in practice, `onRun` already seems to be doing a lot (keeping track of its work and in some cases of its children).

src/work/BasicWork.h
marta-lokhova: This is a helper function that implementers can use if they want a basic wakeUp callback (e.g., in `Work::addWork`)

src/work/BasicWork.h
marta-lokhova: We can, also this would force implementers to use a retry option specified by the enum (as opposed to any number of retries we currently allow), which I think should be okay. I am adding this in the actual implementation (next PR).

src/work/BasicWork.h
marta-lokhova: Added a TODO for this, will be added in the actual implementation (since it requires changing every Work)

marta-lokhova:
This prototype has served its purpose - closing it in favor of https://github.com/stellar/stellar-core/pull/1819

PR #1817, opened 2018-10-25
Run clang-format, improve instructions

Geod24:
# Description

```
Also add missing newlines at the end of files
This was run on clang-format 8.0.0, but the option to end all namespaces
with a comment was disabled for the reviewer's sanity.
```


vogel:
Which version of clang-format do you have?

Geod24:
```
clang-format --version
clang-format version 8.0.0 (tags/google/stable/2018-08-24)
```

I saw that you guys are using v5.0.0 but the changes I added seemed in line with the rest of the code.

vogel:
I'm afraid that clang-format-5.0 will get those changes back...

MonsieurNicolas:
yeah we actually enforce clang-format 5.0 with our CI, so this points to clang-format 8.0 not producing the same result as 5.0
This PR is thus invalid as it is: if you're interested in looking into this further (we tried, but failed), it could be that we don't have a well formed `.clang-format` file, which would explain that things change release after release of clang.

Geod24:
I built `v5.0.{0,2}` from source and indeed some changes have been reverted.

The changes in question do look like a bug in clang-format / Libformat more than in your config file, I however did not find anything relevant in their Phabricator and did not investigate further.

So I removed most of the changes and left only 2 changes:
- `FixNamespaceComment` leads to a huge diff on my machine (over 400 lines) where `// namespace stellar` (and other) gets added at the end of each namespace. I am surprised that you are not seeing this on your CIs, since it was [implemented in v5.0.0](https://releases.llvm.org/5.0.0/tools/clang/docs/ReleaseNotes.html#clang-format).
- I cleaned up some blank lines

`make format` with v5.0.0 does not result in any diff anymore AFAICS.

MonsieurNicolas:
We use an official llvm 5.0 build so I am not sure what is going on. That same build seemed fine on various Linux distros and Windows ; for MacOS the version provided by Apple is not from the official 5.0 tag and was producing differences from what I remember.

The setup is fairly straightforward: look at our Travis configuration file https://github.com/stellar/stellar-core/blob/master/.travis.yml

Geod24:
Ah, [found the bug](https://reviews.llvm.org/D37904). Triggers when Allman is used.
This landed 2017-09-26 and v5.0.0 was released 2017-09-07, which is why `FixNamespaceComments` doesn't have any effect on your source.

@MonsieurNicolas : Can we merge this ? clang-format v5.0.0 won't revert any of those changes.

MonsieurNicolas:
great! yes, thank you.

MonsieurNicolas:
r+ 4752c26a156670c254a9f356cdaf2a960074a64a

PR #1821, opened 2018-11-01
crank returns that it's busy when dequeuing pending work

MonsieurNicolas:
# Description

#1818 introduced a change in the contract of `crank` by accident:
if for some reason crank is called without anything in IOService but something in the pending queue, it would consider that "there is no more work to do", even though work will happen the next turn of the crank.
The problem here is that this breaks:
* virtual clocks move the clock forward very aggressively (potentially timing out work that was in the pending queue)
* loops that use the return value of `crank` to know if there is no more work

The fix is very simple: just count moving items off the pending queue as actual work performed during crank.

I tested this with command line catchup that got broken by the previous commit.


jonjove:
r+ e7e8769

jonjove:
r+ 8ce9f67

PR #1823, opened 2018-11-05
proper fix of VirtualClock bug

vogel:

# Description

This is proper fix to bug introduced in #1818, instead of #1821.
Now it is impossible that work is put to pending queue without executing any work in current crank, as only "run_one" is called outside of `mDelayExecution`.

Also `mDelayExecution` is properly set to false at the end of crank, which was missed in original PR (I must have missed that during squash/rebase phase).

I'll try to add tests to catch issues like that in the future.

### DIFF
--- a/src/util/Timer.cpp
+++ b/src/util/Timer.cpp
@@ -257,52 +257,64 @@ VirtualClock::crank(bool block)
     {
         return 0;
     }
-    nRealTimerCancelEvents = 0;
+
     size_t nWorkDone = 0;
 
-    mDelayExecution = true;
-    if (mMode == REAL_TIME)
     {
-        // Fire all pending timers.
-        nWorkDone += advanceToNow();
-    }
+        std::lock_guard<std::recursive_mutex> lock(mDelayExecutionMutex);
+        // Adding to mDelayedExecutionQueue is now restricted to main thread.
 
-    // pick up some work off the IO queue
-    // calling mIOService.poll() here may introduce unbounded delays
-    // to trigger timers
-    const size_t WORK_BATCH_SIZE = 100;
-    size_t lastPoll;
-    size_t i = 0;
-    do
-    {
-        lastPoll = mIOService.poll_one();
-        nWorkDone += lastPoll;
-    } while (lastPoll != 0 && ++i < WORK_BATCH_SIZE);
+        mDelayExecution = true;
 
-    nWorkDone -= nRealTimerCancelEvents;
+        nRealTimerCancelEvents = 0;
 
-    {
-        std::lock_guard<std::mutex> lock(mDelayedExecutionQueueMutex);
+        if (mMode == REAL_TIME)
+        {
+            // Fire all pending timers.
+            // May add work to mDelayedExecutionQueue
+            nWorkDone += advanceToNow();
+        }
+
+        // Pick up some work off the IO queue.
+        // Calling mIOService.poll() here may introduce unbounded delays
+        // to trigger timers.
+        const size_t WORK_BATCH_SIZE = 100;
+        size_t lastPoll;
+        size_t i = 0;
+        do
+        {
+            // May add work to mDelayedExecutionQueue.
+            lastPoll = mIOService.poll_one();
+            nWorkDone += lastPoll;
+        } while (lastPoll != 0 && ++i < WORK_BATCH_SIZE);
+
+        nWorkDone -= nRealTimerCancelEvents;
+
         if (!mDelayedExecutionQueue.empty())
         {
-            block = false;
-            // count this as work
-            nWorkDone += 1;
+            // If any work is added here, we don't want to advance VIRTUAL_TIME
+            // and also we don't need to block, as next crank will have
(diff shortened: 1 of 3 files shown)


src/util/Timer.cpp
```diff
@@ -281,27 +281,26 @@ VirtualClock::crank(bool block)
...
         mDelayedExecutionQueue.clear();
     }
-
-    if (mMode == VIRTUAL_TIME && nWorkDone == 0)
-    {
-        // If we did nothing and we're in virtual mode,
-        // we're idle and can skip time forward.
-        nWorkDone += advanceToNext();
-    }
+    mDelayExecution = false;
```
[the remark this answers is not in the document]
vogel: I didn't want to use mutexes/atomic for mDelayExecution, because I don't think it is worth it.

In some cases of race conditions around the mDelayExecution member it is possible that job will be posted directly to io_service instead of to the delayed queue and in some cases it will be the opposite, but that should not matter much.

I do not understand what are you saying about queuing 2 events in a row. Which execution will make that happen?

vogel:
This implementation is much simpler than all previous ones.

But I think we can simplify this further if we only allows calls to postToNextCrank from the main thread (this would get rid of mutex and check in postToNextCrank and require adding just one more check in crank).

I don't think it makes sense to think in terms of "crank" in case of background threads.

src/util/Timer.cpp
```diff
@@ -257,52 +257,72 @@ VirtualClock::crank(bool block)
...
-    }
 
-    if (mMode == VIRTUAL_TIME && nWorkDone == 0)
-    {
-        // If we did nothing and we're in virtual mode,
-        // we're idle and can skip time forward.
-        nWorkDone += advanceToNext();
+        if (block && nWorkDone == 0)
+        {
+            assert(mDelayedExecutionQueue.empty());
```
MonsieurNicolas: this assert is confusing, a few lines above there is this:
```c++
            // May add work to mDelayedExecutionQueue for next crank.
            nWorkDone += advanceToNext();
```

also, why do we need the conditions to reset `mDelayExecution` ?

vogel: If nWorkDone is zero, that means that call advanceToNext() was not able to add anything to mDelayedExecutionQueue, so it must be empty (as it was cleared a few lines above).

If those conditions are `false`, `run_one` won't be called and `crank` wont block. In that case we don't need to force next `postToNextCrank` to post directly on io_service queue.

MonsieurNicolas: OK, yeah that makes sense (forgot that we now have a giant mutex).

I am still not sure about that condition on when to reset `mDelayExecution` to false: right now you need to both have `block` set and `nWorkDone == 0`.
* I don't understand why we would not set `mDelayExecution = false` regardless of the value of `block` if we're done with the backlog of work. The way this code is written seems to imply that the only way to call crank is in a loop like `while (crank(true)) {...}` and that people would not alternate between blocking and non blocking (also it will never reset if `block=false`).
* the condition on `nWorkDone == 0` seems arbitrary - I think that by having this condition all we're saying is that we'll delay as long as `IOService` has some work to do, so a comment would help as it's actually not obvious that this is the case (as we allow work to span multiple cranks). Ideally what we would do is really control the amount of work done per crank that is coming from the delayed queue, but this is out of scope of this PR.

vogel: Is there any other reason to set `mDelayExecution = false` than to unblock `run_one`? We still assume that the only way to execute any work that was posted to `VirtualClock` should be executed in `crank` (I'm assuming that using mTimer.getIOService().poll() is not allowed in our codebase).

If so, all of the work will get executed regardless of using `crank` in `while` loop or alternating between blocking and non blocking.

MonsieurNicolas: If you don’t reset `mDelayedExecution` and `crank` is called non blocking, then it will take two turns of the crank for delayed work to be executed instead of just one as there is no event in the `IOService` queue during the first run.

I don’t see a reason to distinguish execution in `poll_one` from a call to `run_one` during the next invocation of `crank`.

MonsieurNicolas:
r+ b146ef5205e14c064b8a2c4d101c9fe7539f1cc2

MonsieurNicolas:
you need to clang format this @vogel

vogel:
rebased

src/util/Timer.cpp
```diff
@@ -316,20 +328,37 @@ VirtualClock::crank(bool block)
 void
 VirtualClock::postToCurrentCrank(std::function<void()>&& f)
 {
+    std::lock_guard<std::recursive_mutex> lock(mDelayExecutionMutex);
+    // One immediate post is enough.
```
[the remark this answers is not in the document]
MonsieurNicolas: @vogel : did you see this question?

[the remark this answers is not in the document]
vogel: Fixed in new version.

MonsieurNicolas:
r+ 1d2c775e36a0df5f3900fb36130ffe16eda0f05c

PR #1824, opened 2018-11-07
Log rotation fix

marta-lokhova:
Updates to how we perform log rotation. This change should allow for proper usage of `logrotate` command on Linux systems. The fix locks all loggers and reconfigures them which causes files to be closed and re-opened (truncated). 

I tested it by running stellar-core with a default/templated log file, then calling `logrotate`. Additionally, I tried logging from the worker thread while performing `logrotate` from the main thread.

### DIFF
--- a/src/util/Logging.h
+++ b/src/util/Logging.h
@@ -13,8 +13,11 @@
 #define ELPP_WINSOCK2
 #define ELPP_DEBUG_ERRORS
 
-// NOTE: Nothing else should include easylogging directly
-//  include this file instead
+// NOTE: Nothing else should include easylogging directly include this file
+// instead Please think carefully modifying this file, and potentially using
+// synchronization primitives. It is easy to introduce data races and deadlocks,
+// so it is recommended to use valgrind --tool=helgrind to detect potential
+// problems.
 #include "lib/util/easylogging++.h"
 
 namespace stellar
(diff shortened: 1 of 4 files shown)


marta-lokhova:
Note that I added a lock to `registeredLoggers`; Valgrind detected that acquiring lock on Logger object and calling `getLogger` (or any function that dispatches a logger) can be potentially dangerous, as one thread could hold a lock to the logger object, while the other thread had a lock on registeredLoggers. Looks like with this change, valgrind does not produce anymore deadlock warnings when calling `logrotate`.

src/util/Logging.h
```diff
@@ -13,22 +13,39 @@
...
+        T* const mItem;
+        LockElObject(T* elObj);
+        ~LockElObject();
+
+        // Disable copy construction and copy assignment
+        LockElObject(LockElObject const&) = delete;
+        LockElObject& operator=(LockElObject const&) = delete;
+    };
+
     struct LockHelper
```
jonjove: `LockElObject` and `LockHelper` are only used in the `.cpp` file so they should be moved there. Cleaner and will make compilation faster if future maintainers ever change them.

marta-lokhova:
@jonjove updated (also squashed some commits)

jonjove:
r+ 6cd956b

PR #1826, opened 2018-11-08
Publish command

### LINKED ISSUE #1814: add an option to manual catchup to wait for publish to complete
Command line catchup can be used to setup full validators, when used this way we need to also wait for publish to complete.

note: there might be an edge case if the target ledger happens to be a snapshot, when working with SCP we wait for the next ledger to close, when this command line option is specified we should probably fail and require people to catchup to a non snapshot ledger (as the reason we wait is to make sure we don't have bad data)

vogel:
# Description

Resolves #1814

Add new command `publish` which publishes all (or almost all) remaining checkpoints to history archives. I may leave one unpublished in case LCL is equal to last ledger in last checkpoint, in which case we must wait for next ledger from network before progressing. Publishing of that last checkpoint is postponed to normal stellar-core run.

This PR requires #1732 to be merged first

# Checklist
- [x] Compiles
- [x] Ran all tests

### DIFF
--- a/src/main/ApplicationUtils.cpp
+++ b/src/main/ApplicationUtils.cpp
@@ -331,25 +331,7 @@ catchup(Application::pointer app, CatchupConfiguration cc,
         return 1;
     }
 
-    // set known cursors before starting maintenance job
-    ExternalQueue ps(*app);
-    ps.setInitialCursors(app->getConfig().KNOWN_CURSORS);
-    app->getMaintainer().start();
-
-    auto done = false;
-    app->getLedgerManager().loadLastKnownLedger(
-        [&done](asio::error_code const& ec) {
-            if (ec)
-            {
-                throw std::runtime_error(
-                    "Unable to restore last-known ledger state");
-            }
-
-            done = true;
-        });
-    auto& clock = app->getClock();
-    while (!done && clock.crank(true))
-        ;
+    app->start();
 
     try
     {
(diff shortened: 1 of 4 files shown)


MonsieurNicolas:
This needs to be rebased @vogel

vogel:
Rebased

src/main/ApplicationUtils.cpp
```diff
@@ -425,4 +425,54 @@ catchup(Application::pointer app, CatchupConfiguration cc,
...
+publish(Application::pointer app)
+{
+    if (!checkInitialized(app))
+    {
+        return 1;
+    }
+
+    // set known cursors before starting maintenance job
+    ExternalQueue ps(*app);
+    ps.setInitialCursors(app->getConfig().KNOWN_CURSORS);
```
MonsieurNicolas: it would be cleaner to just set the right config elements so that `app->start()` does what you want (in this case, I think you just need to set `STANDALONE` to true)

vogel: `STANDALONE` is set before - in CommandLine.cpp:runPublish function, by calling to `Config::setNoListen`.

Maintenance job is normally started in `ApplicationImpl::start`, which we don't use in command-line `catchup` and `publish`. I think we could move calling maintenance to the end of catchup/publish process, or just ignore that and postpone the maintenance itself to normal stellar-core runs (but we can't be sure there will be no instances of stellar-core used only for command-line catchup and publish).

MonsieurNicolas: You didn't understand my question: why not calling `start` if standalone is already set?

vogel:
Updated

MonsieurNicolas:
r+ ce473a1eef95308aaefa1b0b4dbe03fb3a98796d

PR #1830, opened 2018-11-14
New ledger state fixes

jonjove:
Fixes a few issues that were discovered after #1788 was merged

### DIFF
--- a/src/ledger/LedgerState.cpp
+++ b/src/ledger/LedgerState.cpp
@@ -663,13 +663,14 @@ LedgerState::Impl::getInflationWinners(size_t maxWinners, int64_t minVotes)
 
     // Have to load accounts that could be winners after accounting for the
     // change in their vote totals
-    int64_t maxIncrease =
+    auto maxIncreaseIter =
         std::max_element(deltaVotes.cbegin(), deltaVotes.cend(),
                          [](auto const& lhs, auto const& rhs) {
                              return lhs.second < rhs.second;
-                         })
-            ->second;
-    maxIncrease = std::max(int64_t(0), maxIncrease);
+                         });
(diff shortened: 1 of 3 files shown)


src/ledger/LedgerState.cpp
```diff
@@ -1026,17 +1026,22 @@ LedgerState::Impl::maybeUpdateLastModified() const
     throwIfSealed();
     throwIfChild();
 
-    auto entries = mEntry;
-    if (mShouldUpdateLastModified)
+    // Note: We do a deep copy here since a shallow copy would not be exception
```
MonsieurNicolas: can you clarify? exception safe in what respect? Why is it a problem to modify entries that were already modified in the current ledger state?

jonjove: `maybeUpdateLastModified` has the strong exception safety guarantee according to the comments in LedgerStateImpl.h. Since iterator arithmetic and potentially the lambda `f` that is used in `maybeUpdateLastModifiedThenInvokeThenSeal` can both throw, it is possible that only some or all of the `LedgerEntry` pointed to by `entries` have updated `lastModifiedLedgerSeq` but the `LedgerState` does not enter the `sealed` state. The idea was to do copy-and-swap to make this all-or-nothing (eg. achieve strong exception safety). But the original implementation used a shallow copy which did not achieve this since both `entries` and `mEntry` were pointing to the same underlying data, meaning that if `entries` was partially updated then `mEntry` was also partially updated.

MonsieurNicolas: I see. I am merging this but we may have to revisit this if this ends up causing too much overhead: I am not 100% sure we need that strong guarantees for those use cases

MonsieurNicolas:
r+ 628e76805708da1789a8def92087179af79be15a

PR #1834, opened 2018-11-21
fix include required for abort

MonsieurNicolas:
trivial change: abort is defined in `<cstdlib>` per https://en.cppreference.com/w/cpp/utility/program/abort

(this causes problems on minimal builds on Ubuntu 16.04)


MonsieurNicolas:
as this is trivial, I am r+ this myself as this blocks some deployments

MonsieurNicolas:
r+ cba1a614c60e5816c8b25f4923086cc9828ffb67

PR #1837, opened 2018-11-26
add support for more checkers in configure.ac

MonsieurNicolas:
# Description

This pull request adds a few configure options to enable various runtime checkers, as well as better documentation on how to build stellar-core with those options.

This PR also fixes a few problems found while running tests with those checkers (see individual commits for details).


src/crypto/CryptoTests.cpp
```diff
@@ -256,7 +256,7 @@ TEST_CASE("StrKey tests", "[crypto]")
 
         for (size_t p = 0u; p < encoded.size(); p++)
         {
-            for (int st = 0; st < 4; st++)
+            for (int st = 0; st < 9; st++)
```
jonjove: Why 9 here? Was the intention for `3 <= st < 9` to all have the same outcome?

src/herder/HerderSCPDriver.cpp
```diff
@@ -338,13 +339,16 @@ HerderSCPDriver::validateValue(uint64_t slotIndex, Value const& value,
                     << i;
                 res = SCPDriver::kInvalidValue;
             }
-            if (i != 0 && (lastUpgradeType >= thisUpgradeType))
+            else
```
jonjove: Why not just write this as `else if`?

MonsieurNicolas:
@jonjove updated to address your comments

src/crypto/CryptoTests.cpp
MonsieurNicolas: I ended up rewriting what I had, new version is easier to understand

src/crypto/CryptoTests.cpp
```diff
@@ -248,63 +262,71 @@ TEST_CASE("StrKey tests", "[crypto]")
...
+                default:
+                    abort();
                 }
                 uint8_t ver;
                 std::vector<uint8_t> dt;
                 if (corrupted != encoded)
                 {
-                    n_corrupted++;
+                    bool sameSize = (corrupted.size() == encoded.size());
+                    if (sameSize)
```
jonjove: Why is this check needed now when it wasn't needed before? Same for line 314.

MonsieurNicolas: because now the test computes the stats for crc related failures (same size, different content), and enforces (REQUIRE) the other types of corruption

jonjove:
r+ b014532

MonsieurNicolas:
r+ e2d3454ae8bbdcd31b39772e9867c29f035f05c8

MonsieurNicolas:
r+ dc728dabe5cf1db72f186319f6b0308166d99c74

MonsieurNicolas:
r+ 917545927b707ce1ae4e4e624885796f2a50b2cf

MonsieurNicolas:
r+ 5d863adf7f25a986e5b748ee2dd3ae9d301b6edb

MonsieurNicolas:
Failures were caused by old "libasan" included in ubuntu 14.04 (that cannot be upgraded), I had to update travis to ubuntu 16.04 (that still has some issues, but there was an easy workaround as it only impacts C code).

I kicked off a new job that passed the configure stage this time around

MonsieurNicolas:
turns out the issue was that containers in travis don't allow `ptrace` which is required by asan, which in turn causes strange "cross compilation" issues

MonsieurNicolas:
r+ 08e5bfa90f4ee9bd8c29870903914634a73fce99

MonsieurNicolas:
@latobarita: retry

MonsieurNicolas:
r+ be1272a41615b8fe0ef3184b430e434b060f5728

MonsieurNicolas:
r+ 43dbd6b291b29a845b364eeafb1aa32c1d83a0de

MonsieurNicolas:
trying a few things... sorry for the spam.
It looks like travis runs are heavily dependent on caches, and we're right over the timeout when building with no cache (as everything needs to be recompiled in this PR)

MonsieurNicolas:
previous run had optimizations disabled entirely, which cause some tests to timeout (lack of output...)

MonsieurNicolas:
r+ 0edc000bb85d82d26e67d016c90e7301057a3af0

MonsieurNicolas:
r+ b1abb7a3763968cbdd65eaf159f3a9495179d314

MonsieurNicolas:
r+ e5f70faca4b32a46006f3c5935067c159456152b

configure.ac
```diff
@@ -117,7 +175,7 @@ AC_ARG_ENABLE([afl],
               AS_HELP_STRING([--enable-afl],
                              [build with AFL (fuzzer) instrumentation]))
 AS_IF([test "x$enable_afl" = "xyes"], [
-  AS_IF([test "x$enable_asan" = "xyes"], [
+  AS_IF([test "x$sanitizeopts" != "x"], [
     AC_MSG_ERROR([AFL is presently incompatible with Address Sanitizer])
```
vogel: Now it can be any sanitizer, not only address.

MonsieurNicolas:
r+ 248c28e218aede0680873aa5170849a70379dcef

MonsieurNicolas:
r+ 31d7c376b6e6f2ba2176d7f9453a5e3d43ffd3fa

MonsieurNicolas:
r+ 9a84a2ab90b1bb84287af8d9c515178d109c9808

MonsieurNicolas:
r+ d1338965

MonsieurNicolas:
r+ 006a4f9f

PR #1838, opened 2018-11-26
bump min overlay version to 7

MonsieurNicolas:
# Description

This PR bumps the minimum version used by overlay to 7: on the production network, nodes running versions older than 10.0.0 may not forward all transactions (as they are not compatible with protocol version 10)


vogel:
r+ c99f3343255e114c69a3bdfcf7f3ed420e7fb8df

PR #1839, opened 2018-11-27
Use batch db operations in BucketApplicator::advance [WIP]

bobg:
This change improves performance during ingestion by batching database operations together.

Previously, `BucketApplicator::advance` operated by inserting, updating, or deleting one row at a time (and in the insert/update case, doing a separate `SELECT` to determine which was needed). This change removes the need for that `SELECT` by using the `INSERT ... ON CONFLICT (...) DO UPDATE SET ...` SQL syntax, which does an insert-or-update in a single server round-trip.

More importantly, this change introduces the `EntryFrame::Accumulator` type, which gathers inserts, updates, and deletes in memory during its lifetime, then commits them to the db when it goes out of scope. In each call to `BucketApplicator::advance`, a new Accumulator is created for each of the `accounts`, `accountdata`, `offers`, and `trustlines` tables.

The commit-to-db step is meant to happen in bulk for all the changes recorded in the accumulator. Unfortunately it turns out that, though [the soci database-independence layer](http://soci.sourceforge.net/) used by stellar-core does provide an API for bulk changes (via `use(vector<T>)`), the postgresql backend converts that to row-at-a-time logic. So this change also includes code to sidestep soci in the postgresql case when destructing accumulators.

With the changes in this PR, a benchmark run of `stellar-core --catchup-at 20854700 --catchup-to 20854788` on pubnet is about 2x faster than before.

### DIFF
--- a/src/bucket/BucketApplicator.cpp
+++ b/src/bucket/BucketApplicator.cpp
@@ -39,6 +39,8 @@ BucketApplicator::advance()
     soci::transaction sqlTx(mDb.getSession());
 
     {
+        // AccumulatorGroup gathers table-specific inserts, updates, and deletes in memory,
+        // applying them to the db in bulk when going out of scope.
         EntryFrame::AccumulatorGroup accums(mDb);
 
         while (mBucketIter)
--- a/src/bucket/BucketApplicator.cpp
+++ b/src/bucket/BucketApplicator.cpp
@@ -8,6 +8,8 @@
 #include "ledger/LedgerDelta.h"
 #include "util/Logging.h"
 
+#include <chrono>
+
 namespace stellar
 {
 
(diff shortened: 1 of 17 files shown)


bobg:
At the moment, `stellar-core --catchup-at 20854700 --catchup-to 20854788` does not quite finish. Near the very end it encounters this error:

```
2018-11-26T22:06:06.028 GBCVX [History INFO] Catchup downloading transactions for range [20854655..20854719]
2018-11-26T22:06:11.885 GBCVX [History INFO] Catching up: downloading transactions files 2/2 (100%)
2018-11-26T22:06:12.413 GBCVX [History INFO] Catchup applying transactions for range [20854655..20854700]
2018-11-26T22:06:12.413 GBCVX [History INFO] Replaying contents of 2 transaction-history files from LCL [seq=20854655, hash=ea00a9]
2018-11-26T22:06:12.413 GBCVX [History INFO] Catching up: applying checkpoint 1/1 (100%)
stellar-core: ledger/LedgerDelta.cpp:117: void stellar::LedgerDelta::addEntry(stellar::EntryFrame::pointer): Assertion `mNew.find(k) == mNew.end()' failed.
```

MonsieurNicolas:
closing this PR in favor of an issue that links to your branch. Thank you!

https://github.com/stellar/stellar-core/issues/1840

PR #1843, opened 2018-11-30
output DEPRECATED message in stderr to avoid confusing scripts

MonsieurNicolas:
#1732 emits warning messages when using old style command lines, but it emits those in `stdout` which breaks scripts that process what is output by stellar-core.

This PR emits those messages in stderr, which should mitigate this


MonsieurNicolas:
as our CI is broken, I am r+ this directly

MonsieurNicolas:
r+ e07443e414478789d6c782f61a144960251c8dd4

PR #1844, opened 2018-12-03
fix broken ECDH

MonsieurNicolas:
was initializing a buffer instead of reserving a max size
(issued introduced in 1a3039e6a9ad492f703984927c44db0e3b6f4103 )


marta-lokhova:
r+ f07269a99f048fba82707c02e6ef1350b2c75ad1

PR #1845, opened 2018-12-03
Improve close time check

MonsieurNicolas:
This PR adds an extra check when not tracking SCP (when starting up/out of sync), that causes nodes to reject values that have invalid time stamps.

Currently we rely 100% on the fact that there is only one main branch of history (versioned with the ledger sequence number), but on test networks, this is not always the case: when resetting a network (without changing the keys), it's possible that old (before the reset) messages are still available on the network, even though they are not valid anymore.

What this PR does to deal with this kind of situation is ensure that we always compare the timestamp of the last known closed ledger (ie, compare the "network time" with the one stored locally).

The situation can be illustrated by this sequence:
1. (...) validator `A` votes for `<10000, 2018-01-01 00:01:00>`
2. reset network (ledger sequence numbers reset to genesis)
3. node `B` stopped at `<10, 2018-01-02 00:00:00>`
3. (...) validator `B` sees a vote from A `<10000, 2018-01-01 00:01:00>`
   a. currently, this would cause `B` to potentially jump to ledger 10000 from the old fork
   b. with this fix, `B` will ignore those messages as `B` already knows that the network time is at least `2018-01-02 00:00:00`

This does not help when joining the network for the first time (as the last known ledger is "genesis"), but helps nodes properly reject messages after a restart.

### DIFF
--- a/src/herder/HerderSCPDriver.cpp
+++ b/src/herder/HerderSCPDriver.cpp
@@ -179,6 +179,31 @@ HerderSCPDriver::isSlotCompatibleWithCurrentState(uint64_t slotIndex) const
     return res;
 }
 
+bool
+HerderSCPDriver::checkCloseTime(uint64_t slotIndex, uint64_t lastCloseTime,
+                                StellarValue const& b) const
+{
+    // Check closeTime (not too old)
+    if (b.closeTime <= lastCloseTime)
+    {
+        CLOG(TRACE, "Herder")
+            << "Close time too old for slot " << slotIndex << ", got "
+            << b.closeTime << " vs " << lastCloseTime;
+        return false;
+    }
+
+    // Check closeTime (not too far in future)
+    uint64_t timeNow = mApp.timeNow();
+    if (b.closeTime > timeNow + Herder::MAX_TIME_SLIP_SECONDS.count())
+    {
+        CLOG(TRACE, "Herder")
+            << "Close time too far in future for slot " << slotIndex << ", got "
+            << b.closeTime << " vs " << timeNow;
+        return false;
+    }
+    return true;
+}
+
 SCPDriver::ValidationLevel
 HerderSCPDriver::validateValueHelper(uint64_t slotIndex,
                                      StellarValue const& b) const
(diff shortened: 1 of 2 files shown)


jonjove:
r+ 455f435

MonsieurNicolas:
@latobarita: retry

src/herder/HerderSCPDriver.cpp
```diff
@@ -187,13 +218,43 @@ HerderSCPDriver::validateValueHelper(uint64_t slotIndex,
...
+        }
     }
     else
     {
+        if (slotIndex == lcl.ledgerSeq)
+        {
+            // previous ledger
+            if (b.closeTime != lastCloseTime)
+            {
+                return SCPDriver::kInvalidValue;
```
jonjove: Perhaps we should log in each case that we return without invoking `checkCloseTime`?

src/herder/HerderSCPDriver.cpp
```diff
@@ -187,13 +218,43 @@ HerderSCPDriver::validateValueHelper(uint64_t slotIndex,
...
+            // basic sanity check on older value
+            if (b.closeTime >= lastCloseTime)
+            {
+                return SCPDriver::kInvalidValue;
+            }
+        }
+        else if (!checkCloseTime(slotIndex, lastCloseTime, b))
+        {
+            // new messages must meet the time criteria from lcl
+            // as lcl's close time is a lower bound
```
jonjove: This condition is triggered more generally than what is mentioned in the comment, since `checkCloseTime` also returns false if the time is too far in the future. This should be harmonized: if the comment is right then don't call `checkCloseTime` and just check the lower bound, if the comment is wrong then it should be corrected. In the previous implementation, if we weren't tracking then it would never check if the time is too far in the future.

src/herder/HerderSCPDriver.cpp
MonsieurNicolas: fixed

src/herder/HerderSCPDriver.cpp
MonsieurNicolas: agree, changed the comment to not be misleading

MonsieurNicolas:
@jonjove updated

jonjove:
r+ 8a18397

Issue #1518, opened 2018-01-30
Some buckets are missing


khelle:
I have another problem with Stellar Core. After resolving the problems with joining the public network, I was running the node for few days without problems. The start/restart worked flawlessly, the node connected each time to network, but now I encountered another problem. Despite node claiming to be fully synced, it does not see any transaction made. It did for some time as I tested it via Horizon API connected to the node, but now it does not see anything. When I look at the logs I can see:

```
Jan 30 12:52:06 localhost stellar-core[11412]: 2018-01-30T12:52:06.334 GBA2J [default INFO] Starting stellar-core v9.0.1-6-g226ba55
Jan 30 12:52:06 localhost stellar-core[11412]: 2018-01-30T12:52:06.334 GBA2J [default INFO] Config from /home/xlm/.stellar-core/stellar-core.cfg
Jan 30 12:52:06 localhost stellar-core[11412]: 2018-01-30T12:52:06.335 GBA2J [Database INFO] Connecting to: postgresql://dbname=dbname user=user password=******** host=127.0.0.1
Jan 30 12:52:06 localhost stellar-core[11412]: 2018-01-30T12:52:06.350 GBA2J [SCP INFO] LocalNode::LocalNode@GBA2J qSet: b3a703
Jan 30 12:52:06 localhost stellar-core[11412]: 2018-01-30T12:52:06.350 GBA2J [default INFO] Listening on 127.0.0.1:12059 for HTTP requests
Jan 30 12:52:06 localhost stellar-core[11412]: 2018-01-30T12:52:06.352 GBA2J [History INFO] Archive 'local' has 'put' and 'get' commands, will be read and written
Jan 30 12:52:06 localhost stellar-core[11412]: 2018-01-30T12:52:06.352 GBA2J [History INFO] Archive 'sdf1' has 'get' command only, will not be written
Jan 30 12:52:06 localhost stellar-core[11412]: 2018-01-30T12:52:06.352 GBA2J [History INFO] Archive 'sdf2' has 'get' command only, will not be written
Jan 30 12:52:06 localhost stellar-core[11412]: 2018-01-30T12:52:06.352 GBA2J [History INFO] Archive 'sdf3' has 'get' command only, will not be written
Jan 30 12:52:06 localhost stellar-core[11412]: 2018-01-30T12:52:06.353 GBA2J [default INFO] Loading last known ledger
[...]

PR #1846, opened 2018-12-04
enable --disable-bucket-gc for publish command

vogel:

# Description

This one allows `publish` command to have `--disable-bucket-gc` option.

### DIFF
--- a/src/main/CommandLine.cpp
+++ b/src/main/CommandLine.cpp
@@ -428,15 +428,20 @@ int
 runPublish(CommandLineArgs const& args)
 {
     CommandLine::ConfigOption configOption;
+    auto disableBucketGC = false;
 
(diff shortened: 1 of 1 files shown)


MonsieurNicolas:
why do we need this option for publish?

vogel:
If we don't have that for publish, then stale files can be removed at start.

If someone want's to keep buckets, and calls `catchup` command line with this option, then calling `publish` after that without it will erase all buckets that are not referenced in database, which mostly means buckets for non-checkpoint ledgers.

MonsieurNicolas:
that option is supposed to be used for recovery situations as a way to recover from https://github.com/stellar/stellar-core/issues/1518 . publishing sounds more like people would be running everything with that option, which doesn't sound desirable as disk utilization grows too fast

vogel:
OK, lets remove this.

Issue #1018, opened 2016-04-02
Make Bucket really immutable


vinniefalco:
`std::shared_ptr<Bucket const>`, just sayin'

graydon:
Fair number of cases now work with mutable Bucket objects, due to the lazily constructed indexes. There are cases that can be or should be const but those are already const.

PR #1847, opened 2018-12-05
Small bucket improvements

vogel:
# Description

This code adds asserts and changes Bucket to Bucket const to ensure that buckets are properly managed.


src/bucket/FutureBucket.cpp
```diff
@@ -297,12 +329,15 @@ FutureBucket::startMerge(Application& app, bool keepDeadEntries)
...
 {
+    assertThreadIsMain();
+
     checkState();
     assert(!isLive());
     assert(hasHashes());
     auto& bm = app.getBucketManager();
     if (hasOutputHash())
     {
+        // THIS
```
marta-lokhova: debugging leftover?

vogel: oops

MonsieurNicolas:
I would rather not refactor code like this. A good change started but not completed in this PR would be to move to using `Bucket const` in most places as per #1018 
Best to make this change would not be to add `const` everywhere though, but instead rename `Bucket` to something like `RawBucket` and define `Bucket` to be the `const` version of that type, that way, only a few places in the code base will have to be changed (only places are going to be where we actually create buckets).

vogel:
using Bucket = RawBucket const; requires also multiple class RawBucket; before, so I don't think it is optimal solution.

I've also tried to go one step further, and make using Bucket = std::shared_ptr<RawBucket const>;, which I think is better solution, but still require multiple statements per file. As we have multiple ways of defining shared pointers now:

typedef std::shared_ptr<TxSetFrame> TxSetFramePtr;
class Config
{
  typedef std::shared_ptr<Config> pointer;
}

I think I'll do a step back and clean then up a bit (and make consistent). Maybe even some *_fwd.h includes will be needed, that will contain only typedefs?

MonsieurNicolas:
sweeping stale pull requests

PR #1849, opened 2018-12-05
make application "start" idempotent

MonsieurNicolas:
# Description

Commit ce473a1eef95308aaefa1b0b4dbe03fb3a98796d introduced a bug where "start" can be called several times (when calling "startup" several times by using both `catchup-at` and `catchup-to` on the same command line for example).

Fix I picked is to just make the application `start` idempotent: alternative would be to not call start if it's not necessary, but that doesn't seem useful.


MonsieurNicolas:
@stellar/core-committers ready for review

jonjove:
r+ 4cfbeab

Issue #1853, opened 2018-12-06
investigate database schema updates


MonsieurNicolas:
There are a few changes that we should look into that should improve how the database behaves.
Of course we'll have to validate those with performance evaluation.

Here are a few ideas to get started:
* get rid of the "signers" table, store the data inline with the account (xdr, base64 encoded?). We always load signers, so this seems like a no-brainer.
* simplify some of data stored in columns such as AccountIDs, replace with a unique ID (int64 or int32) across the board. This requires an extra lookup (to find the actual accountID) but as many tables have indexes on accountID, this should consolidate a lot of the indexing/footprint required to deal with those keys

MonsieurNicolas:
first one is merged into 10.3.0 (with large performance implication) - more analysis is needed for potential changes

PR #1866, opened 2018-12-11
Move signers into account table

jonjove:
Updates the SQL schema so that signers are stored in the main accounts table rather than in a separate table. Related to #1853


vogel:
Looks good to me

MonsieurNicolas:
I would be interested to see some performance numbers to support this change, maybe write a mini-benchmark to get an idea of worst case performance when accounts are indeed used with signers (both read and write impact):
* there are not that many signers in the ledger today, so I would expect the impact of this change to be negligible on a catchup complete.
* for write, worst case here might be an account with max number of signers where only the sequence number gets updated (today signers are not updated when this happens).

graydon:
Further: because this is not included in an entry, we have to do entry-at-a-time analysis of the relation, which defeats batching in #1906 quite badly. It is the difference (performance-wise) between 8000 entries/sec and 12000, with nothing else changing except precision of managing this auxiliary relation.

graydon:
> @MonsieurNicolas I rebased on top of master, then added a simple micro-benchmark and an assert. I also ran clang-format. The results of the micro-benchmark over 3 samples (each containing 100,000 accounts) using SQLite are in the following table (of the form Avg +/- StDev in milliseconds)

@jonjove looks promising! If I'm reading correctly, "in-account is faster than separate table", right? I don't have the meanings reversed? :)

If so: would you be willing/able to run against pg too? (I guess for thoroughness; though I imagine _all_ these numbers are only so meaningful given we're about to start batching..)

jonjove:
@graydon Yes you are reading this correctly "in account table" (meaning this PR) is faster than "in separate table" (meaning master), and this is especially true when there are many signers. I will try this against postgres as well and share the results.

MonsieurNicolas:
Are you sure you are not rewriting signers in the rewrite case? I was expecting the create from scratch and read to be slower (as per your table), but account updates to be somewhat faster as signers should not be touched.

jonjove:
@MonsieurNicolas that's a good point. I have revisited how the signers table is modified in `master` (it's quite complicated). We can proceed through the loop (ledger/LedgerTxnAccountSQL.cpp:248) as follows: 
1. If there are no more signers, done.
2. Both iterators are pointing to the same signer (which is the first signer in both lists)
3. We reach ledger/LedgerTxnAccountSQL.cpp:258 which yields `update = true` since both signers are the same
4. We reach ledger/LedgerTxnAccountSQL.cpp:271 which evaluate to false since both signers are the same, so no SQL is executed
5. Both iterators are advanced (ledger/LedgerTxnAccountSQL.cpp:289-290)

After 5, we are at the same state as 1 except both iterators are pointing to the second signer (which is the same signer in both lists, if there is a second signer). Proceeding inductively, we see that no SQL is executed to update the list.

However, I now realize that in both cases the accounts must be re-read as part of load to determine whether it is an UPDATE or an INSERT (since the cache is disabled). This will no longer occur when switching to UPSERT as part of the batching that @graydon is working on. As a consequence, `master` must execute an additional query to load from the signers table which is not required in the "in account table" version. This load would be counted during the "Rewrite" phase. I think explanation is sufficient to explain the unexpected difference in "Rewrite" times, do you agree?

src/ledger/LedgerTxnTests.cpp
```diff
@@ -2018,3 +2020,163 @@ TEST_CASE("LedgerTxnEntry and LedgerTxnHeader move assignment", "[ledgerstate]")
...
+        {
+            runTest(Config::TESTDB_ON_DISK_SQLITE, 100000, 10);
+        }
+        SECTION("20 signers")
+        {
+            runTest(Config::TESTDB_ON_DISK_SQLITE, 100000, 20);
+        }
+    }
+
+#ifdef USE_POSTGRES
```
MonsieurNicolas: you don't have to copy the section, you can just do something like
```c++
    SECTION("sqlite")
    {
        runtest(Config::TESTDB_ON_DISK_SQLITE);
    }
#ifdef USE_POSTGRES
    SECTION("postgresql")
    {
        runtests(Config::TESTDB_POSTGRESQL);
    }
#endif
```

MonsieurNicolas:
yeah I don't think disabling the cache gives a good picture of the overhead as we're artificially forcing a second read of all signers.

Maybe we'd get a more realistic result by doing something like
```c++
            CLOG(WARNING, "Ledger") << "Reading accounts";
            {
                auto timer = getTimeScope(app, numSigners, "read");
                for (auto const& key : accounts)
                {
                    ltx.load(key);
                }
            }

            LedgerTxn modsltx(ltx);
            {
                for (auto const& key : accounts)
                {
                    ++modsltx.load(key).current().lastModifiedLedgerSeq;
                }
            }

            CLOG(WARNING, "Ledger")
                << "Writing accounts with unchanged signers";
            {
                auto timer = getTimeScope(app, numSigners, "rewrite");
                modsltx.commit();
                ltx.commit();
            }
```

src/ledger/LedgerTxnTests.cpp
jonjove: Good idea, done.

jonjove:
I don't think your suggestion for improving the benchmark would help at all. The fundamental issue is that as `LedgerTxnRoot` commits its child, it writes each entry to SQL. But it needs to determine whether to do an INSERT or an UPDATE. This proceeds in the following manner, for each entry in child:
1. If the entry has been deleted (meaning it is `nullptr`), just issue a SQL DELETE
2. Otherwise, invoke `LedgerTxnRoot::getNewestVersion` which first checks the cache (always empty in this case) and then issues a SQL SELECT if it is not in the cache
3. If `getNewestVersion` did not return nullptr, do SQL UPDATE otherwise do SQL INSERT

No rearranging of the nested `LedgerTxn` can improve this situation since only the outermost still exists at the time that the relevant `commit` occurs, and absolutely nothing is cached at the root regardless of what loads occur.

If you want to get data for the "Rewrite" phase that avoids the extra load, I can enable the cache with size 100,000. This way everything is cached, and we will only see the effect of the writes. Effectively, this will measure the incremental overhead of writing the signers column in the accounts table. Does this appeal to you?

MonsieurNicolas:
ah yes, I got confused by the line
```c++
        auto const previous = getNewestVersion(iter.key());
```
not realizing that it is implemented from root.

I guess enabling the cache for that test seems like a solution then.

jonjove:
With cache size equal to number of accounts (100,000):

```
        |                           | 0 Signers              | 10 Signers              | 20 Signers
--------|---------------------------|------------------------|-------------------------|------------------------
Create  | Signers in account table  | 1061.0740 +/-  28.1965 |  1103.8806 +/-   3.1076 |  1168.4226 +/-  13.0427
        | Signers in separate table | 1037.3534 +/-  21.3192 |  1095.2838 +/-   8.7505 |  1141.4513 +/-  17.0585
Write   | Signers in account table  | 2348.6627 +/-  59.0491 |  3474.2169 +/-  28.5275 |  4552.5901 +/-  30.6778
        | Signers in separate table | 2363.8784 +/-  49.9721 | 15540.7527 +/- 362.9378 | 34730.4546 +/- 875.2707
Read    | Signers in account table  | 3361.9636 +/- 137.1752 |  4356.2434 +/- 109.6522 |  5135.2969 +/-  40.4947
        | Signers in separate table | 3896.8229 +/-  30.2408 |  7178.9823 +/- 100.7750 | 12701.0539 +/-  59.0302
Rewrite | Signers in account table  | 3770.0806 +/-  29.3549 |  5202.7517 +/- 122.0125 |  6153.8035 +/-  54.9974
        | Signers in separate table | 3856.0687 +/-  71.5435 |  4055.9880 +/-  21.5564 |  4281.8317 +/-  59.2365
```

As expected, rewriting was slower in this PR than in master. I don't know why reading in a separate table with 20 signers was so much slower with caching than without. Either way, during transaction processing there is at least one read per write so it looks to me like the read improvement should compensate for the additional rewrite cost. How does this look to you @MonsieurNicolas?

src/ledger/LedgerTxnAccountSQL.cpp
```diff
@@ -68,12 +72,16 @@ LedgerTxnRoot::Impl::loadAccount(LedgerKey const& key) const
...
+    if (signersInd == soci::i_ok)
     {
-        auto signers = loadSigners(key);
-        account.signers.insert(account.signers.begin(), signers.begin(),
-                               signers.end());
+        assert(std::adjacent_find(account.signers.begin(),
+                                  account.signers.end(),
+                                  [](Signer const& lhs, Signer const& rhs) {
+                                      return !(lhs.key < rhs.key);
+                                  }) == account.signers.end());
```
MonsieurNicolas: I think this assert should be after setting account.signers

src/ledger/LedgerTxnAccountSQL.cpp
```diff
@@ -179,22 +217,25 @@ LedgerTxnRoot::Impl::insertOrUpdateAccount(LedgerEntry const& entry,
     std::string thresholds(decoder::encode_b64(account.thresholds));
     std::string homeDomain(account.homeDomain);
 
+    std::string signers(
```
MonsieurNicolas: you should only set `signers` if there are signers

src/ledger/LedgerTxnAccountSQL.cpp
```diff
@@ -209,6 +250,7 @@ LedgerTxnRoot::Impl::insertOrUpdateAccount(LedgerEntry const& entry,
     st.exchange(soci::use(entry.lastModifiedLedgerSeq, "v8"));
     st.exchange(soci::use(liabilities.buying, liabilitiesInd, "v9"));
     st.exchange(soci::use(liabilities.selling, liabilitiesInd, "v10"));
+    st.exchange(soci::use(signers, "v11"));
```
MonsieurNicolas: you should set this to `NULL` when `signers` is empty.
Look how it's done with `st.exchange(soci::into(inflationDest, inflationDestInd));`

src/util/Decoder.h
```diff
@@ -28,6 +28,7 @@ inline std::string
 encode_b32(T const& v)
 {
     std::string res;
+    // TODO(jonjove): Shouldn't this be encoded_size32?
```
MonsieurNicolas: spurious change not related to this commit (answer is yes though)

src/ledger/LedgerTxnAccountSQL.cpp
```diff
@@ -153,6 +127,70 @@ LedgerTxnRoot::Impl::loadInflationWinners(size_t maxWinners,
...
+        std::string signers(decoder::encode_b64(xdr::xdr_to_opaque(kv.second)));
+
+        auto prep = mDatabase.getPreparedStatement(
+            "UPDATE accounts SET signers = :v1 WHERE accountID = :id");
+        auto& st = prep.statement();
+        st.exchange(soci::use(signers, "v1"));
+        st.exchange(soci::use(kv.first, "id"));
+        st.define_and_bind();
+        st.execute(true);
+
```
MonsieurNicolas: in other places we do
```c++
        if (st.get_affected_rows() != 1)
        {
            throw std::runtime_error("Could not update data in SQL");
        }
```
after an update, is this right?

jonjove:
I'm actually not that surprised that the penalty for writing the extra data is so large. 500 bytes is more data than all the other columns of the table contain combined, so this really is a non-negligible change in data written. That being said, I did a benchmark with the column added when the table was created instead of during schema upgrade (as before) and with the cache enabled. I only did this analysis for this PR since it does not apply for master. The rows labeled "Column added during schema upgrade" were copied from my previous table ("With cache size equal to number of accounts") for comparison purposes only and were not re-run.

```
        |                                    | 0 Signers              | 10 Signers             | 20 Signers
--------|------------------------------------|------------------------|------------------------|-----------------------
Create  | Column added during schema upgrade | 1061.0740 +/-  28.1965 | 1103.8806 +/-   3.1076 | 1168.4226 +/- 13.0427
        | Column added on table creation     | 1037.7416 +/-  11.4633 | 1096.3039 +/-   5.1122 | 1176.4397 +/- 12.1811
Write   | Column added during schema upgrade | 2348.6627 +/-  59.0491 | 3474.2169 +/-  28.5275 | 4552.5901 +/- 30.6778
        | Column added on table creation     | 2269.8802 +/-  11.9027 | 3465.1589 +/-  28.7089 | 4546.8274 +/- 12.4169
Read    | Column added during schema upgrade | 3361.9636 +/- 137.1752 | 4356.2434 +/- 109.6522 | 5135.2969 +/- 40.4947
        | Column added on table creation     | 3288.4396 +/-   9.3378 | 4341.6159 +/-  42.2556 | 5122.7249 +/- 16.2210
Rewrite | Column added during schema upgrade | 3770.0806 +/-  29.3549 | 5202.7517 +/- 122.0125 | 6153.8035 +/- 54.9974
        | Column added on table creation     | 3717.2039 +/-  64.3404 | 5084.9034 +/-  29.6357 | 6110.2335 +/- 54.0024
```

jonjove:
@graydon @MonsieurNicolas Here are the results with cache enabled for postgres:

```
        |                           | 0 Signers               | 10 Signers               | 20 Signers
--------|---------------------------|-------------------------|--------------------------|---------------------------
Create  | Signers in account table  |  6392.7515 +/- 405.8140 |   5906.7670 +/-  78.5239 |   6037.1224 +/-  169.1585
        | Signers in separate table |  5798.9706 +/-  66.3260 |   5954.4027 +/- 143.3184 |   6096.7565 +/-   96.2425
Write   | Signers in account table  | 11670.0304 +/- 523.8610 |  12558.1102 +/- 260.6823 |  13725.3469 +/-   29.1450
        | Signers in separate table | 11796.7324 +/- 314.2727 | 123107.6749 +/- 688.9159 | 244030.2258 +/- 5357.4647
Read    | Signers in account table  |  9309.3851 +/- 300.3677 |   9868.7983 +/-  80.3476 |  11437.7853 +/-  175.1112
        | Signers in separate table | 12600.3958 +/- 271.7832 |  18546.5544 +/- 398.7538 |  20972.6937 +/-  718.8980
Rewrite | Signers in account table  | 15334.4626 +/- 676.9425 |  16188.1299 +/- 176.0429 |  18150.8801 +/-  335.2149
        | Signers in separate table | 15018.9724 +/-  90.0764 |  15413.2491 +/-  83.2891 |  15671.9864 +/   555.7293
```

MonsieurNicolas:
Looks good to me

MonsieurNicolas:
r+ 9cf9c410

MonsieurNicolas:
this was reverted by https://github.com/stellar/stellar-core/pull/1933

PR #1868, opened 2018-12-12
raport percent of completed bucket apply

### LINKED ISSUE #1633: Log percentage completed in `Bucket-apply` stage
# Issue type
feature request

# Issue description
I don't know what's the underlying data structure so not sure if it's even possible but at the `Bucket-apply` stage can we display progress in percents? Current log entries like:
```
Bucket-apply: committed 151551 entries
```
do not help much in estimating percentage completed.

vogel:

# Description

Resolves #1633

Display number of bytes applied, number of buckets applied and percentage instead of only number of entries applied in apply-bucket phase of catchup.

# Checklist
- [x] Compiles
- [x] Ran all tests

### DIFF
--- a/src/bucket/BucketApplicator.cpp
+++ b/src/bucket/BucketApplicator.cpp
@@ -25,9 +25,23 @@ BucketApplicator::operator bool() const
     return (bool)mBucketIter;
 }
 
-void
+size_t
+BucketApplicator::pos()
+{
+    return mBucketIter.pos();
+}
+
+size_t
+BucketApplicator::size() const
+{
+    return mBucketIter.size();
+}
+
+size_t
 BucketApplicator::advance()
 {
+    size_t count = 0;
+
     LedgerState ls(mApp.getLedgerStateRoot(), false);
     for (; mBucketIter; ++mBucketIter)
     {
@@ -53,17 +67,15 @@ BucketApplicator::advance()
                 entry.erase();
             }
         }
-        if ((++mSize & 0xff) == 0xff)
+
+        if ((++count & 0xff) == 0xff)
         {
             break;
         }
     }
     ls.commit();
 
-    if (!mBucketIter || (mSize & 0xfff) == 0xfff)
-    {
-        CLOG(INFO, "Bucket")
-            << "Bucket-apply: committed " << mSize << " entries";
-    }
+    mCount += count;
+    return count;
 }
 }
--- a/src/util/Fs.cpp
+++ b/src/util/Fs.cpp
@@ -442,6 +442,33 @@ checkNoGzipSuffix(std::string const& filename)
     }
 }
 
+size_t
+size(std::ifstream& ifs)
+{
+    assert(ifs.is_open());
+
+    ifs.seekg(0, ifs.end);
+    auto result = ifs.tellg();
+    ifs.seekg(0, ifs.beg);
+
+    return std::max(decltype(result){0}, result);
+}
+
+size_t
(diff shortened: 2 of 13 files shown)


src/util/Fs.cpp
```diff
@@ -442,6 +442,33 @@ checkNoGzipSuffix(std::string const& filename)
     }
 }
 
+std::streamsize
+size(std::ifstream& ifs)
+{
+    assert(ifs.is_open());
+
+    ifs.ignore(std::numeric_limits<std::streamsize>::max());
```
MonsieurNicolas: why not something like
```
    // get length of file:
    is.seekg (0, is.end);
    auto length = is.tellg();
    is.seekg (0, is.beg);
```

src/util/Fs.cpp
```diff
@@ -442,6 +442,33 @@ checkNoGzipSuffix(std::string const& filename)
...
+    ifs.seekg(0, std::ios_base::beg);
+
+    return result;
+}
+
+std::streamsize
+size(std::string const& filename)
+{
+    struct stat st;
+    if (stat(filename.c_str(), &st) == 0)
```
MonsieurNicolas: why the duplicate implementation? I am not sure this is portable. you can just open a stream and call the other one.

src/util/XDRStream.h
```diff
@@ -23,7 +25,8 @@ class XDRInputFileStream
 {
     std::ifstream mIn;
     std::vector<char> mBuf;
-    unsigned int mSizeLimit;
+    std::streamsize mSizeLimit;
```
MonsieurNicolas: I'd rather be consistent: let's use `size_t` everywhere instead of stream specific types

src/util/types.h
```diff
@@ -36,6 +36,8 @@ AccountID getIssuer(Asset const& asset);
 // returns true if the currencies are the same
 bool compareAsset(Asset const& first, Asset const& second);
 
+std::string formatSize(long size);
```
MonsieurNicolas: should take a `size_t`

src/util/Fs.cpp
vogel: According to https://stackoverflow.com/questions/22984956/tellg-function-give-wrong-size-of-file/22986486#22986486 it may not always tell us the number of bytes.

src/util/Fs.cpp
vogel: I didn't want to open a file just to get its size, but I can do it.

src/util/Fs.cpp
MonsieurNicolas: Couple things:
 * As we’re using `tellg` to calculate the current position it’s fine (we get a proper percentage)
 * The reason it’s bigger in that article is that it’s binary vs translated content. Doesn’t matter anyways because of the previous point

src/bucket/BucketApplicator.cpp
```diff
@@ -53,17 +67,15 @@ BucketApplicator::advance()
                 entry.erase();
             }
         }
-        if ((++mSize & 0xff) == 0xff)
+
+        ++count;
+        if ((++mCount & 0xff) == 0xff)
```
MonsieurNicolas: actually you can just use `count` here and add `count` to `mCount` outside the loop (after the commit)

vogel:
Fixed and rebased

MonsieurNicolas:
r+ 6a75f6248fb2095d27b8c43b7b940f7aea168798

PR #1869, opened 2018-12-12
Lock TmpDir root on startup

### LINKED ISSUE #1863: stellar-core crashes when a different stellar-core is invoked
It seems that a node in sync crashes if a different stellar-core is started with the same configuration file.

marta-lokhova:
Resolves #1863 

Looks like tmp dir on startup is never locked, causing racing core instances to delete and recreate same directories.

### DIFF
--- a/src/bucket/BucketManager.h
+++ b/src/bucket/BucketManager.h
@@ -16,6 +16,7 @@ namespace stellar
 
 class Application;
 class BucketList;
+class TmpDirManager;
 struct LedgerHeader;
 struct HistoryArchiveState;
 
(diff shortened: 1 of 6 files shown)


MonsieurNicolas:
I don't think this fix makes sense: there is already locking in existing code, per my comment in #1863, it's probably not done at the right time, we should not add a new lock type.
https://github.com/stellar/stellar-core/blob/1fe2e8a768ecc4db2d53a4a67fc733bb1e99ecd1/src/bucket/BucketManagerImpl.cpp#L132

MonsieurNicolas:
looks good now!

MonsieurNicolas:
r+ 8c82a7261968a8d9b816d527b82424fc1c4034cb

PR #1870, opened 2018-12-13
call cleanupStaleFiles after syncing to network

vogel:

# Description

Keep all buckets until catchup is done. It can save some time as buckets needed for catchup may be already downloaded but are not required by current contents of the database.

# Checklist
- [x] Compiles
- [x] Ran all tests

### DIFF
--- a/src/bucket/BucketManager.h
+++ b/src/bucket/BucketManager.h
@@ -98,6 +98,10 @@ class BucketManager : NonMovableOrCopyable
     // current BL.
     virtual void assumeState(HistoryArchiveState const& has) = 0;
 
+    // Remove staled bucket files that are not used currently. Call after
+    // syncing to the network.
+    virtual void cleanupStaleFiles() = 0;
+
     // Ensure all needed buckets are retained
     virtual void shutdown() = 0;
 };
--- a/src/bucket/BucketManagerImpl.cpp
+++ b/src/bucket/BucketManagerImpl.cpp
@@ -443,7 +443,6 @@ BucketManagerImpl::assumeState(HistoryArchiveState const& has)
     }
 
     mBucketList.restartMerges(mApp);
-    cleanupStaleFiles();
 }
 
 void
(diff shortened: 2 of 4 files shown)


MonsieurNicolas:
This is extremely (sha256 collision) unlikely to happen as a node catching up implies that it’s “in the past” compared to what it’s catching up to.

Am I missing something here?

vogel:
There is one case:
1. start catchup
2. stop stellar-core in the middle of downloading and verifying buckets
3. start catchup again
4. all downloaded buckets will be removed by cleanup and re-downloaded

This fix removes redundant downloading.

This was found during testing of apply-buckets reporting progress PR.

MonsieurNicolas:
Oh I see, this is because we're keeping buckets that have been verified but not applied yet.

Issue I have seen with this code is that it can take a while to run (not sure why it takes so long to enumerate files in a folder, assuming this is the reason), if we move it to run "inline" when the node is synced, it may cause more problems than it solves. Not sure how we can get more information about this other than adding better logging at the INFO level.

That said, it could be that with the changes to catchup (as part of the Work refactor), we can instead pipeline downloads & apply ledger (and avoid waiting to download all before applying). This would not require to change when we GC, just reduce the amount of data needed to download before applying (and we would remember where we stopped).

If we decide to merge this, I would prefer to keep the logic "close to where the effects happen":
* an issue I can see with this version is that it's potentially downloading (and adding) buckets on retry, which could lead to running out of disk space - should we also run cleanup when `Catchup` starts? That way on restart it will purge buckets that are obsolete for the current catchup.
* I don't know if clearing files when we reach the "synced" state is the right event as we startup in that state. I think that it should probably happen the first time we close a ledger as it deals with all other possible scenarios as far as I can tell.
* I imagine that we should run this code by its own (ie, posting to ioservice), as it may take a relatively long time to run.

vogel:
The other option would be to make this a command that needs to be executed manually, like

`stellar-core gc-buckets`

With all new bucket-retain-clenaup code it is unlikely that non-needed bucket will be kept in `buckets` directory. The `cleanupStaleFiles` was added to deal with leftover buckets that were not properly removed in previous versions of `stellar-core`.

MonsieurNicolas:
sweeping stale pull requests

PR #1871, opened 2018-12-14
Use unordered map/sets in ledger state

MonsieurNicolas:
# Description

This PR changes the way we track ledger state in memory from maps and sets to their hash table equivalent.
To do this, it introduces a hash function for `LedgerEntry` (based on sipHash found in libsodium).
The changes yield a 20% speed improvement on some of the heavy ledgers I tested, and on a large range of normal ledgers (1000+ ledgers) didn't see any regressions in memory or CPU utilization.

This PR also updates the documentation on how to use "perf" in Linux.


src/ledger/LedgerHashUtils.h
```diff
@@ -0,0 +1,45 @@
...
+        switch (lk.type())
+        {
+        case stellar::ACCOUNT:
+            res = stellar::shortHash::computeHash(
+                stellar::ByteSlice(lk.account().accountID.ed25519().data(), 8));
+            break;
+        case stellar::TRUSTLINE:
+            res = stellar::shortHash::computeHash(stellar::ByteSlice(
+                lk.trustLine().accountID.ed25519().data(), 8));
+            res ^= lk.trustLine().asset.type();
```
jonjove: This doesn't seem desirable. Every trust line owned by the same account with the same asset type will hash to the same value. So creating an offer between two `ASSET_TYPE_CREDIT_ALPHANUM4` assets will lead to a guaranteed collision, for example, since both trust lines will have the same hash. This probably needs to include at least some bytes from both the asset code and the issuer.

src/ledger/LedgerHashUtils.h
```diff
@@ -0,0 +1,45 @@
+#pragma once
```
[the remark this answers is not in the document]
jonjove: Actually you already have it in this order, but GitHub isn't displaying it in the right order because of the commit date. Please update the commit dates.

src/ledger/LedgerHashUtils.h
MonsieurNicolas: good point asset.type is not what I wanted to use here

MonsieurNicolas:
@jonjove I ended up not using magic values and instead used sizes derived from datasets. Performance results are about the same than with hard coded number.

src/ledger/LedgerStateTests.cpp
```diff
@@ -1208,22 +1208,20 @@ testAllOffers(
...
-                                        oe.amount));
-                ++expectedInnerIter;
-                ++innerIter;
+                auto const& oe = innerCur.current().data.offer();
+                auto d = std::make_tuple(oe.offerID, oe.buying, oe.selling,
+                                         oe.amount);
+                auto expectedInnerIter =
+                    std::find(expectedInner.begin(), expectedInner.end(), d);
+                REQUIRE(expectedInnerIter != expectedInner.end());
+                expectedInner.erase(expectedInnerIter);
```
jonjove: Why call `erase` here? Also, if you didn't erase here then there would be no need to copy-assign `expectedInner`.

MonsieurNicolas: yes, first version I had didn't do this, but I thought the test code could have bad data and we would not catch it. This guarantees that we don't do stupid things like having the same data specified several times in the expected set as it can only be used once.

MonsieurNicolas:
@latobarita: retry

PR #1874, opened 2018-12-18
allow global options with new command line syntax

### LINKED ISSUE #1864: command line parsing should be improved
The new command line mode has a few problems:
* it should be possible to put global options (like `--conf`) before or after the sub command.
   * Something like `./stellar-core --conf my-config.cfg new-db` and not just `./stellar-core new-db --conf my-config.cfg `
   * using the wrong syntax defaults to the legacy command line
* unknown sub-commands trigger the legacy command line, which combined with garbage on the command line defaulting to "run" is dangerous
   * `./stellar-core --conf my-config.cfg invalid-cmd` should just print the "help" message instead of starting up core like normal

We can't yet deprecate the old command line, so we should address all those issues.

vogel:

# Description

Resolves #1864

Now 3 options can be used before command:
* --conf
* --metrics
* --ll

Also unknown commands will trigger new command line syntax instead of default deprecated run command.

# Checklist
- [x] Compiles
- [x] Ran all tests


MonsieurNicolas:
I don't think you need to make that many changes to your previous code.

Basically, I think that all we need is extract whatever is before the command, and call Clara on it to pick new starting values for `configOption` using `configurationParser`, and call the code for each command as it is now (with the reduced parameters).

I put together a proof of concept (hacked around), to illustrate and it seems practical to just do https://github.com/MonsieurNicolas/stellar-core/commit/24582379f00446d005a562c2255ae0a5ad90c2b5
I did a simple loop to look for the command, might be better to use clara for this, I don't know.

src/main/CommandLine.cpp
```diff
@@ -763,23 +817,26 @@ handleCommandLine(int argc, char* const* argv)
...
-    auto command = commandLine.selectCommand(argc, argv);
+    auto ajustedCommandLine = commandLine.adjustCommandLine({argc, argv});
+    auto command = commandLine.selectCommand(ajustedCommandLine.first);
     if (!command)
     {
         return nullopt<int>();
     }
 
+    auto exeName = argv[0];
+    auto commandName = fmt::format("{0} {1}", argv[0], command->name());
```
MonsieurNicolas: I would not bother with `argv[0]` in case it has spaces

src/test/test.cpp
```diff
@@ -216,8 +218,9 @@ runTest(CommandLineArgs const& args)
     session.cli(parser);
 
     auto result = session.cli().parse(
-        args.mCommandName, Catch::clara::detail::TokenStream{
-                               std::begin(args.mArgs), std::end(args.mArgs)});
+        fmt::format("{0} {1}", args.mExeName, args.mCommandName),
```
MonsieurNicolas: same here (spaces in "exeName") - not as critical as it's test code though

src/main/CommandLine.cpp
MonsieurNicolas: I don't think we ever extract the binary path, so we're probably good to just ditch this entirely

vogel: what do you mean exactly? to make commandName = command->name() ?

MonsieurNicolas: I'd say put a hard-coded string such as "stellar-core" instead of `argv[0]`

vogel: updated

src/test/test.cpp
vogel: I switched this one to only use mCommandName, as it contains "stellar-core" inside

MonsieurNicolas:
r+ 2223822c172c800247c16fb5be2e1a99b805bce9

PR #1875, opened 2018-12-18
Fix test

marta-lokhova:


MonsieurNicolas:
This doesn't seem to be the right fix:
the reason to have a good bucket mixed with a bad one was to make sure that the logic fails properly the top level `DownloadBucketsWork`.

The issue is that `downloadStatusCheck` should be checking that
* only one child fails (in the same way it's checking things now)
* other children succeed

MonsieurNicolas:
Also in the future, put a description of what the problem was

marta-lokhova:
@MonsieurNicolas we cannot guarantee that other children completed with the current work interface, because Work does not wait for all to complete, but instead just proceeds to a failure state (see `Work::advance`)

PR #1881, opened 2018-12-20
Upgrade DB schema inside a transaction

### LINKED ISSUE #1791: schema upgrade sometimes doesn't upgrade the schema cleanly
# Issue type
  bug

# Version found
 10.0.0

# Issue description
It looks like any database failure occurring during a schema upgrade leaves the database in a bad state.

We should either
* perform the upgrade in a transaction (but this may cause other problems such as running out of space in the SQL working set)
* have all upgrade steps be idempotent

note: as all nodes are now running 10.0.0 (at schema version 7), it may make sense to consolidate all schema changes 0-6 into a single "6" (and avoid upgrading the schema for older versions).

## Actual result
After updating core from 9.2.0 to 10.0.0 core failed with this error:
```
2018-09-12T02:21:29.627 GAUTY [Database INFO] Applying DB schema upgrade to version 7
2018-09-12T02:21:29.636 GAUTY [default FATAL] Got an exception: Cannot execute query. ERROR:  column "buyingliabilities" of relation "accounts" already exists
 [main.cpp:671]
```

## Expected result
Schema upgrade should proceed.

## Workaround

Applying the upgrades manually to the database should allow the system to proceed.

Steps are:
1. connect to the core database
2. verify the schema version is `6` by running `SELECT state FROM storestate WHERE statename = 'databaseschema';`
3. apply the upgrades for version 7 as per https://github.com/stellar/stellar-core/blob/1fc018b4f52e8c7e716b023ccf30600af5b4f66d/src/database/Database.cpp#L151
4. update the schema version to `7` by running `UPDATE storestate SET state = 7 WHERE statename = 'databaseschema';`

marta-lokhova:
Resolves #1791 

Note that as part of this, anything prior to version 7 is consolidated

### DIFF
--- a/src/database/Database.cpp
+++ b/src/database/Database.cpp
@@ -108,40 +108,9 @@ Database::applySchemaUpgrade(unsigned long vers)
 {
     clearPreparedStatementCache();
 
+    soci::transaction tx(mSession);
     switch (vers)
     {
-    case 2:
-        HerderPersistence::dropAll(*this);
-        break;
-
-    case 3:
-        mApp.getLedgerStateRoot().dropData();
-        break;
-
-    case 4:
-        BanManager::dropAll(*this);
-        mSession << "CREATE INDEX scpquorumsbyseq ON scpquorums(lastledgerseq)";
-        break;
-
-    case 5:
-        try
-        {
-            mSession << "ALTER TABLE accountdata ADD lastmodified INT NOT NULL "
-                        "DEFAULT 0;";
-        }
-        catch (soci::soci_error& e)
-        {
-            if (std::string(e.what()).find("lastmodified") == std::string::npos)
-            {
-                throw;
-            }
-        }
-        break;
-
-    case 6:
-        mSession << "ALTER TABLE peers ADD flags INT NOT NULL DEFAULT 0";
-        break;
-
     case 7:
         Upgrades::dropAll(*this);
         mSession << "ALTER TABLE accounts ADD buyingliabilities BIGINT "
@@ -155,9 +124,17 @@ Database::applySchemaUpgrade(unsigned long vers)
         break;
 
     default:
-        throw std::runtime_error("Unknown DB schema version");
-        break;
+        if (vers <= 6)
+        {
+            throw std::runtime_error(
+                "Database version is too old, must use at least 7");
+        }
+        else
+        {
+            throw std::runtime_error("Unknown DB schema version");
+        }
     }
+    tx.commit();
 }
 
 void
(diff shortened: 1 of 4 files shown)


src/ledger/LedgerStateDataSQL.cpp
```diff
@@ -124,6 +124,7 @@ LedgerStateRoot::Impl::dropData()
                               "accountid    VARCHAR(56)  NOT NULL,"
                               "dataname     VARCHAR(64)  NOT NULL,"
                               "datavalue    VARCHAR(112) NOT NULL,"
+                              "lastmodified INT          NOT NULL DEFAULT 0,"
```
MonsieurNicolas: no need for "default 0" here as we always provide its value

src/overlay/PeerRecord.cpp
```diff
@@ -287,6 +287,7 @@ const char* PeerRecord::kSQLCreateStatement =
     "port          INT DEFAULT 0 CHECK (port > 0 AND port <= 65535) NOT NULL,"
     "nextattempt   TIMESTAMP NOT NULL,"
     "numfailures   INT DEFAULT 0 CHECK (numfailures >= 0) NOT NULL,"
+    "flags         INT NOT NULL DEFAULT 0,"
```
MonsieurNicolas: no need for default 0 as we always pass in a value

src/database/Database.cpp
```diff
@@ -108,41 +108,14 @@ Database::applySchemaUpgrade(unsigned long vers)
 {
     clearPreparedStatementCache();
 
-    switch (vers)
+    if (vers <= 6)
```
MonsieurNicolas: keep the switch as it makes the code easier to modify in the future (we'll continue to follow the same pattern)

src/database/Database.cpp
```diff
@@ -288,7 +288,7 @@ Database::initialize()
     HerderPersistence::dropAll(*this);
     mApp.getLedgerStateRoot().dropData();
     BanManager::dropAll(*this);
-    putSchemaVersion(1);
+    putSchemaVersion(6);
```
MonsieurNicolas: you should keep this change in the previous commit so that you don't even need to special case ` if (vers <= 6)` (and error in that case, which is also missing from this changeset)

marta-lokhova:
@MonsieurNicolas addressed your comments and squashed some commits.

MonsieurNicolas:
looking good

MonsieurNicolas:
r+ 27c1651883b8fb672896d8414f0bb88b5c5c5325

PR #1887, opened 2018-12-21
rebuild Visual Studio project files

MonsieurNicolas:
VS project had a bug introduced in dc0167c159678353dc54f0a755346692b2f19612 that would cause builds to not link properly

issue was the use of `CLCompile` instead of `CLInclude`
```xml
<ClCompile Include="..\..\src\historywork\BatchWork.h" />
```


marta-lokhova:
r+ e01b14cbcbcc2c60f9a8e450da9a2e7cc96e2ca0

Issue #1878, opened 2018-12-18
metrics: rethink metrics for invariants and `txINTERNAL` type of errors


MonsieurNicolas:
Right now we have
* too many metrics for invariants - as we're already logging those, we should just have one metric that tracks the total number (`Counter`) of invariant failures (making it easier to have monitors for those)
* no metric for critical failures such as `txINTERNAL` errors (`Counter`)

PR #1894, opened 2019-01-03
Improve logging and metrics for transaction subsystem

### LINKED ISSUE #1879: metrics: rework `closeledger` related metrics and logging
The use of metrics in the transaction subsystem is not very good:
* we're missing some key metrics
   * number of operations in a ledger (we have number of transactions per ledger)
* we're missing some logs
   * right now, even in `TRACE` level, we don't get any information on which operation is executed, etc.
      * We should have at least something like (`TRACE` level probably) `operation #1 start (PATH_PAYMENT)` and `operation #1: SUCCESS` (when done)
* we have too many metrics with no use case
  * `op-<NAME>.success.apply`, `op-<NAME>.failure.<ERR> ` should be replaced by logging (`TRACE` level probably) 
  * `operation.failure.<ERR>`  and `transaction.failure.<ERR>` should be replaced by logging (`TRACE` level probably) and an aggregate counter for failures

MonsieurNicolas:
# Description

Resolves #1879 #1878 

This PR replaces the use of metrics by better logging in the transaction subsystem and makes it easier to stop "abnormal" failures with metrics.

### DIFF
--- a/src/invariant/InvariantManagerImpl.cpp
+++ b/src/invariant/InvariantManagerImpl.cpp
@@ -32,7 +32,8 @@ InvariantManager::create(Application& app)
 }
 
 InvariantManagerImpl::InvariantManagerImpl(medida::MetricsRegistry& registry)
-    : mMetricsRegistry(registry)
+    : mInvariantFailureCount(
+          registry.NewCounter({"ledger", "invariant", "failure"}))
 {
 }
 
@@ -40,19 +41,17 @@ Json::Value
 InvariantManagerImpl::getJsonInfo()
 {
     Json::Value failures;
-    for (auto const& invariant : mInvariants)
-    {
-        auto& counter = mMetricsRegistry.NewCounter(
-            {"invariant", "does-not-hold", "count", invariant.first});
-        if (counter.count() > 0)
-        {
-            auto const& info = mFailureInformation.at(invariant.first);
 
-            auto& fail = failures[invariant.first];
-            fail["count"] = (Json::Int64)counter.count();
-            fail["last_failed_on_ledger"] = info.lastFailedOnLedger;
-            fail["last_failed_with_message"] = info.lastFailedWithMessage;
-        }
+    for (auto const& fi : mFailureInformation)
+    {
+        auto& fail = failures[fi.first];
+        auto& info = fi.second;
+        fail["last_failed_on_ledger"] = info.lastFailedOnLedger;
+        fail["last_failed_with_message"] = info.lastFailedWithMessage;
+    }
+    if (!failures.empty())
+    {
+        failures["count"] = (Json::Int64)mInvariantFailureCount.count();
     }
     return failures;
 }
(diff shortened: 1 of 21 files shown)


src/invariant/InvariantManagerImpl.cpp
```diff
@@ -208,12 +210,10 @@ InvariantManagerImpl::onInvariantFailure(std::shared_ptr<Invariant> invariant,
...
                                          uint32_t ledger)
 {
-    mMetricsRegistry
-        .NewCounter(
-            {"invariant", "does-not-hold", "count", invariant->getName()})
-        .inc();
-    mFailureInformation[invariant->getName()].lastFailedOnLedger = ledger;
-    mFailureInformation[invariant->getName()].lastFailedWithMessage = message;
+    mInvariantFailureCount.inc();
+    mFailureInformation.emplace(
```
jonjove: This changes the semantics of `mFailureInformation`. This captures the oldest failure rather than the newest since
> Inserts a new element into the container constructed in-place with the given args if there is no element with the key in the container.

according to https://en.cppreference.com/w/cpp/container/map/emplace.

src/invariant/InvariantManagerImpl.cpp
```diff
@@ -32,28 +32,32 @@ InvariantManager::create(Application& app)
...
 InvariantManagerImpl::getJsonInfo()
 {
     Json::Value failures;
+
     for (auto const& invariant : mInvariants)
     {
-        auto& counter = mMetricsRegistry.NewCounter(
-            {"invariant", "does-not-hold", "count", invariant.first});
-        if (counter.count() > 0)
+        auto it = mFailureInformation.find(invariant.first);
```
jonjove: I think this loop could be rewritten more simply as
```c++
for (auto const& failure : mFailureInformation)
{
    auto& fail = failures[failure.first];
    auto const& info = failure.second;
    fail["last_failed_on_ledger"] = info.lastFailedOnLedger;
    fail["last_failed_with_message"] = info.lastFailedWithMessage;
}
```

src/ledger/LedgerManagerImpl.cpp
```diff
@@ -1007,6 +1007,9 @@ LedgerManagerImpl::applyTransactions(std::vector<TransactionFramePtr>& txs,
         }
         catch (InvariantDoesNotHold&)
         {
+            CLOG(ERROR, "Ledger")
```
jonjove: Good idea to log in this case!

src/invariant/InvariantManagerImpl.cpp
MonsieurNicolas: I thought we didn't care to have the latest, just an example as we now log, but maybe it's better to keep the latest failure. Let me revert that

src/invariant/InvariantManagerImpl.cpp
MonsieurNicolas: good suggestion

MonsieurNicolas:
@jonjove updated with your suggestions

jonjove:
r+ 22c2c09

MonsieurNicolas:
@latobarita: retry

PR #1897, opened 2019-01-04
ledger: Add high-level diagram and comment about new data access layer.

graydon:
# Description

This just adds an overview diagram and comment explaining the high level structure and motivation of the new data access layer. I am planning on doing some renaming in a followup PR later today but I figured this would be a good checkpoint to start with.

# Checklist
- [x] Compiles
- [x] Ran all tests
- [x] (NA) If change impacts performance, include supporting evidence per the [performance document](https://github.com/stellar/stellar-core/blob/master/performance-eval/performance-eval.md)

### DIFF
--- a/src/ledger/LedgerState.h
+++ b/src/ledger/LedgerState.h
@@ -14,6 +14,156 @@
 #include <unordered_map>
 #include <unordered_set>
 
+/////////////////////////////////////////////////////////////////////////////
+//  Overview
+/////////////////////////////////////////////////////////////////////////////
+//
+// The LedgerState subsystem consists of a number of classes (made a bit
+// more numerous through the use of inner ::Impl "compiler firewall"
+// classes and abstract base classes), of which the essential members and
+// relationships are diagrammed here.
+//
+//
+//  +-----------------------------------+
+//  |LedgerStateRoot                    |
+//  |(will commit child entries to DB)  |
+//  |                                   |
+//  |Database &mDatabase                |
+//  |AbstractLedgerState *mChild -----------+
+//  +-----------------------------------+   |
+//      ^                                   v
+//      |   +-----------------------------------+
+//      |   |LedgerState                        |
+//      |   |(will commit child entries to self)|
+//      |   |                                   |
+//      +----AbstractLedgerStateParent &mParent |
+//          |AbstracLedgerState *mChild ----------+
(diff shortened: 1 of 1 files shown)


src/ledger/LedgerState.h
```diff
@@ -14,6 +14,157 @@
...
+//    sub-LedgerState is indicated by a non-null mChild pointer.
+//
+//  - Once a child is closed and the mChild pointer is reset to null,
+//    a new child may be opened. Attempting to open two children at once
+//    will throw an exception.
+//
+//  - The entries to be committed in each transaction are stored in the
+//    mEntry map, keyed by LedgerKey. This much is straightforward!
+//
+//  - Committing any LedgerState / transaction except the root will merge
```
jonjove: This note and the next are almost, but not exactly, correct. The distinction is that the root cannot be committed explicitly, it writes (and subsequently commits) to the underlying database automatically when its child is committed. So commit always merges the entries of the child into its parent's entries, which specifically means writing to the database if the parent is the root.

src/ledger/LedgerState.h
```diff
@@ -14,6 +14,157 @@
...
+//  Overview
+/////////////////////////////////////////////////////////////////////////////
+//
+// The LedgerState subsystem consists of a number of classes (made a bit
+// more numerous through the use of inner ::Impl "compiler firewall"
+// classes and abstract base classes), of which the essential members and
+// relationships are diagrammed here.
+//
+//
+//  +-----------------------------------+
```
jonjove: Nice ascii diagrams!

graydon:
> looks like you didn't branch off master, so you have a dupe commit with #1896

That was intentional: the tests don't pass on my machine without #1896. I figured it would have merged to master before we finish iterating on this one anyway :)

src/ledger/LedgerState.h
graydon: Ah, ok. WIll push fix in a moment, thanks for clarifying!

graydon:
@jonjove Pushed fix to that note. Let me know if anything else is off.

jonjove:
Looks good to me now! Rebase it (since the first commit is merged in a previous PR) and squash, then I'll merge it.

graydon:
@jonjove done

jonjove:
r+ 4830c5e

PR #1902, opened 2019-01-08
buckets: add bucket applicator rate meter.

graydon:
# Description

Just adds a metric for the rate of bucket entities applied per second (and includes it in the INFO progress lines during catchup). Helpful as a lightweight / easily compared "local DB performance" benchmark value.

# Checklist
- [x] Compiles
- [x] Ran all tests
- [x] (NA) If change impacts performance, include supporting evidence per the [performance document](https://github.com/stellar/stellar-core/blob/master/performance-eval/performance-eval.md)

### DIFF
--- a/src/catchup/ApplyBucketsWork.cpp
+++ b/src/catchup/ApplyBucketsWork.cpp
@@ -163,8 +163,12 @@ ApplyBucketsWork::advance(BucketApplicator& applicator)
         return;
     }
 
+    auto& meter =
+        mApp.getMetrics().NewMeter({"bucket", "apply", "entries"}, "entry");
     assert(mTotalSize != 0);
-    mAppliedEntries += applicator.advance();
+    auto sz = applicator.advance();
+    mAppliedEntries += sz;
+    meter.Mark(sz);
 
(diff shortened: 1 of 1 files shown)


src/catchup/ApplyBucketsWork.cpp
```diff
@@ -163,8 +163,12 @@ ApplyBucketsWork::advance(BucketApplicator& applicator)
         return;
     }
 
+    auto& meter =
```
marta-lokhova: minor: it's probably a bit more consistent and efficient to just have this meter as a member variable

graydon: fixed (this is carried on over in the bulk-operations PR; I'll close this one).

graydon:
Continued in the bulk-operations PR #1906 . @MonsieurNicolas I felt it was helpful to get performance characteristics and since it's surfaced to field users as an INFO log message during catchup it might be an easily-compared signal about the difference in performance between our dev setups and the sort of IO subsystems our production users have. But if you feel strongly opposed I can drop it from that PR.

PR #1904, opened 2019-01-08
Install new operator handler

### LINKED ISSUE #1757: audit code for accidental out of memory exception handling
We should ensure that our code crashes when running out of memory.

Right now there are places that don't rethrow on exception (rethrowing with `throw;` is fine), resulting in some exceptions being "eaten".

In particular, `std::bad_alloc` : we should instead crash the application.

The only proper exception handling for those is to either crash directly (call `abort()`) or rethrow with `throw;`;

I think the two ways we do this are:
```c++

catch( ... )
{
}
```
and
```c++

catch (std::exception)
{
}
```

marta-lokhova:
resolves #1757; I ended up taking a slightly different route by handling failed allocation prior to `bad_alloc` being thrown. This seems nicer as we don't need to track the code for accidental out-of-memory exception handling.

### DIFF
--- a/src/main/main.cpp
+++ b/src/main/main.cpp
@@ -7,15 +7,30 @@
 #include "util/Logging.h"
 
 #include "crypto/ByteSliceHasher.h"
+#include <cstdlib>
 #include <sodium/core.h>
 #include <xdrpp/marshal.h>
 
 INITIALIZE_EASYLOGGINGPP
 
+namespace stellar
+{
+static void
+outOfMemory()
+{
+    std::fprintf(stderr, "Unable to allocate memory\n");
+    std::fflush(stderr);
+    std::abort();
+}
+}
+
 int
 main(int argc, char* const* argv)
 {
     using namespace stellar;
+
+    // Abort when out of memory
+    std::set_new_handler(outOfMemory);
 
     Logging::init();
     if (sodium_init() != 0)


src/main/main.cpp
```diff
@@ -12,11 +12,24 @@
 
 INITIALIZE_EASYLOGGINGPP
 
+namespace stellar
+{
+static void
+outOfMemory()
+{
+    std::cerr << "Unable to allocate memory \n";
```
MonsieurNicolas: you have to use `std::endl` in order to flush

also, I think you need to include `<cstdlib>` for the call to `std::abort` to be portable

marta-lokhova: done

src/main/main.cpp
```diff
@@ -7,16 +7,30 @@
...
 #include <xdrpp/marshal.h>
 
 INITIALIZE_EASYLOGGINGPP
 
+namespace stellar
+{
+static void
+outOfMemory()
+{
+    std::cerr << "Unable to allocate memory" << std::endl;
```
jonjove: This doesn't look as safe as I would like. Are you certain that memory allocation via `operator new` never occurs on this line? If not, this will trigger `outOfMemory` recursively and `std::abort()` will not be reached.

MonsieurNicolas: we can just do fprintf/fflush over stderr as we don't do anything special

MonsieurNicolas:
r+ 8c1419fe484ca3c8e164fd1bae405a650547d212

PR #1909, opened 2019-01-11
Small improvement in encode_b32 to reserve the correct size

jonjove:
Reserve the correct size for the encoded string in `encode_b32`.


graydon:
r+ 40945dd

PR #1915, opened 2019-01-16
log slow executions of posted tasks

### LINKED ISSUE #1880: metrics: add metrics that tracks asio queues
Right now we use 2 queues: the main IOService queue and the worker queue.
We should track the delay in a metric between when a `post` call happens and the actual execution of the callback.

This is especially important to detect starvation or creation of a backlog in the respective subsystem. One of the theories behind #1764 is that too many merges are happening with too few threads for example.

vogel:

# Fixes #1880 

# Description

Allow writing to log file information about tasks created with `Application::postOnMainThread`, `Application::postOnMainThreadWithDelay` and `Application::postOnBackgroundThread` that are taking more than one second to start execution.

# Checklist
- [x] Compiles
- [x] Ran all tests

### DIFF
--- a/src/main/ApplicationImpl.cpp
+++ b/src/main/ApplicationImpl.cpp
@@ -43,6 +43,7 @@
 #include "scp/QuorumSetUtils.h"
 #include "simulation/LoadGenerator.h"
 #include "util/GlobalChecks.h"
+#include "util/LogSlowExecution.h"
 #include "util/StatusManager.h"
 #include "work/WorkManager.h"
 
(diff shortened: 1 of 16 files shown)


src/main/ApplicationImpl.cpp
```diff
@@ -69,6 +70,12 @@ ApplicationImpl::ApplicationImpl(VirtualClock& clock, Config const& cfg)
     , mStoppingTimer(*this)
     , mMetrics(std::make_unique<medida::MetricsRegistry>())
     , mAppStateCurrent(mMetrics->NewCounter({"app", "state", "current"}))
+    , mPostOnMainThreadDelay(
```
MonsieurNicolas: metrics.md need to be updated

vogel:
metrics updated

src/util/LogSlowExecution.h
```diff
@@ -9,26 +9,46 @@
...
-    std::chrono::system_clock::time_point mStart;
-    std::string mName;
-    int mThresholdInMs;
-
   public:
-    LogSlowExecution(std::string eventName, int ms = 1000)
+    enum class Mode
+    {
+        EXECUTION,
+        START
```
marta-lokhova: I think this enum is confusing: `LogSlowExecution` does not know that time before start will be measured, and `checkExecuted` can be called anywhere. I think it should be more generic, something like this:
```
enum class Mode 
{
    AUTOMATIC_RAII,
    MANUAL // In this mode, it is the caller's responsibility to check elapsed time
}
```

vogel:
Updated and rebased

marta-lokhova:
r+ a15de86d70f284964f22fb4d46422d94fa7088ff

PR #1918, opened 2019-01-17
Prefetcher API [WIP]

marta-lokhova:
Wanted to open a preliminary PR on this new prefetcher functionality. I did a sketch of the API (though I anticipate I'll be adding more to it once we know the exact needs of ledgerTxnRoot/ Batching API). I also added prefetch calls to some operations. I did some very basic testing on the prefetch queue and cache handling (not included in this PR). I still need to add the following items:
* Handle prefetching for load best offers 
* Handle data prefetch 
* Rest of the operations (e.g. set options inflation destination)
* Consider and document exception safety of the new functions
* Consider failure cases/ missing data in the DB. 

This is still very much WIP,  as I'll have a better idea once I see preliminary Batch API implementation.

### DIFF
--- a/src/ledger/LedgerManagerImpl.cpp
+++ b/src/ledger/LedgerManagerImpl.cpp
@@ -25,6 +25,8 @@
 #include "main/Application.h"
 #include "main/Config.h"
 #include "overlay/OverlayManager.h"
+#include "transactions/OperationFrame.h"
+#include "transactions/TransactionUtils.h"
 #include "util/Logging.h"
 #include "util/XDROperators.h"
 #include "util/format.h"
@@ -803,6 +805,16 @@ LedgerManagerImpl::closeLedger(LedgerCloseData const& ledgerData)
     // sorted such that sequence numbers are respected
     vector<TransactionFramePtr> txs = ledgerData.getTxSet()->sortForApply();
 
+    // Prefetch all tx source accounts
+    if (mApp.getConfig().PREFETCH_BATCH_SIZE > 0)
+    {
+        auto& root = mApp.getLedgerTxnRoot();
+        for (auto const& tx : txs)
+        {
+            stellar::prefetchAccount(root, tx->getSourceID());
+        }
+    }
+
     // first, charge fees
     processFeesSeqNums(txs, ltx);
 
@@ -942,6 +954,30 @@ LedgerManagerImpl::processFeesSeqNums(std::vector<TransactionFramePtr>& txs,
 }
 
 void
+LedgerManagerImpl::prefetchTransactionData(
+    std::vector<TransactionFramePtr>& txs)
+{
+    if (mApp.getConfig().PREFETCH_BATCH_SIZE > 0)
+    {
+        auto& root = mApp.getLedgerTxnRoot();
+        for (auto const& tx : txs)
+        {
+            for (auto const& op : tx->getOperations())
+            {
+                if (!(tx->getSourceID() == op->getSourceID()))
+                {
+                    stellar::prefetchAccount(root, op->getSourceID());
+                }
+                for (auto const& key : op->getLedgerKeysToPrefetch(mApp))
+                {
+                    root.prefetch(key);
+                }
+            }
+        }
+    }
+}
+
+void
 LedgerManagerImpl::applyTransactions(std::vector<TransactionFramePtr>& txs,
                                      AbstractLedgerTxn& ltx,
                                      TransactionResultSet& txResultSet)
@@ -950,19 +986,22 @@ LedgerManagerImpl::applyTransactions(std::vector<TransactionFramePtr>& txs,
 
     // Record counts
     auto numTxs = txs.size();
+    size_t numOps = 0;
     if (numTxs > 0)
     {
         mTransactionCount.Update(static_cast<int64_t>(numTxs));
-        size_t numOps =
-            std::accumulate(txs.begin(), txs.end(), size_t(0),
-                            [](size_t s, TransactionFramePtr const& v) {
-                                return s + v->getOperations().size();
-                            });
+        numOps = std::accumulate(txs.begin(), txs.end(), size_t(0),
+                                 [](size_t s, TransactionFramePtr const& v) {
+                                     return s + v->getOperations().size();
+                                 });
         mOperationCount.Update(static_cast<int64_t>(numOps));
         CLOG(INFO, "Tx") << fmt::format("applying ledger {} (txs:{}, ops:{})",
                                         ltx.loadHeader().current().ledgerSeq,
                                         numTxs, numOps);
     }
+
+    // Request some prefetches
+    prefetchTransactionData(txs);
 
     for (auto tx : txs)
     {
@@ -1003,6 +1042,27 @@ LedgerManagerImpl::applyTransactions(std::vector<TransactionFramePtr>& txs,
         tx->storeTransaction(mApp.getDatabase(), ledgerSeq, tm, ++index,
                              txResultSet);
     }
+
+    logTxApplyMetrics(ltx, numTxs, numOps);
+}
+
+void
+LedgerManagerImpl::logTxApplyMetrics(AbstractLedgerTxn& ltx, size_t numTxs,
+                                     size_t numOps)
+{
+    auto ledgerSeq = ltx.loadHeader().current().ledgerSeq;
+
+    CLOG(DEBUG, "Ledger") << "Ledger: " << ledgerSeq << " txs: " << numTxs
+                          << ", ops: " << numOps
+                          << ", AVG application time per tx: "
+                          << mTransactionApply.mean() << " ± "
+                          << mTransactionApply.std_dev();
+
+    CLOG(DEBUG, "Ledger")
+        << "Ledger: " << ledgerSeq << " txs: " << numTxs << ", ops: " << numOps
+        << ", MIN application time per tx: " << mTransactionApply.min()
+        << ", MAX " << mTransactionApply.max() << ", 75th percentile: "
+        << mTransactionApply.GetSnapshot().get75thPercentile();
 }
 
 void
--- a/src/ledger/LedgerTxn.cpp
+++ b/src/ledger/LedgerTxn.cpp
@@ -12,6 +12,7 @@
 #include "ledger/LedgerTxnHeader.h"
 #include "ledger/LedgerTxnImpl.h"
 #include "util/GlobalChecks.h"
+#include "util/Logging.h"
 #include "util/XDROperators.h"
 #include "util/types.h"
 #include "xdr/Stellar-ledger-entries.h"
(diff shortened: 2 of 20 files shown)


src/ledger/LedgerManagerImpl.cpp
```diff
@@ -969,6 +971,26 @@ LedgerManagerImpl::processFeesSeqNums(std::vector<TransactionFramePtr>& txs,
...
+void
+LedgerManagerImpl::prefetchTransactionData(
+    std::vector<TransactionFramePtr>& txs)
+{
+    auto& root = mApp.getLedgerTxnRoot();
+    for (auto const& tx : txs)
+    {
+        stellar::prefetchAccount(root, tx->getSourceID());
+        auto const& ops = tx->getOperations();
+        for (auto const& op : ops)
```
vogel: I think we could extract method in TransactionFrame to get all of used source ID where?

for (auto const& account : tx->getSourceAccounts()) ....

src/ledger/LedgerTxn.cpp
```diff
@@ -21,11 +21,176 @@
...
+            lst.push_front(key);
+        }
+        else
+        {
+            lst.push_back(key);
+        }
+
+    };
+
+    switch (key.type())
```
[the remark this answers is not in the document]
marta-lokhova: Good idea, done.

src/ledger/LedgerTxn.cpp
```diff
@@ -1233,6 +1399,8 @@ LedgerTxnRoot::Impl::commitChild(EntryIterator iter)
     // Clearing the cache does not throw
     mBestOffersCache.clear();
     mEntryCache.clear();
+    // TODO Prefetch queue must be emptied prior to root's commit.
```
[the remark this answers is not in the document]
marta-lokhova: That's the idea, wasn't sure if this function was the right place to do it, but seems like it is.

src/transactions/PathPaymentOpFrame.cpp
```diff
@@ -286,4 +286,47 @@ PathPaymentOpFrame::doCheckValid(Application& app, uint32_t ledgerVersion)
     }
     return true;
 }
+
+void
+PathPaymentOpFrame::doPrefetch(Application& app)
```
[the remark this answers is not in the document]
marta-lokhova: Good idea, centralizing it in `LedgerManagerImpl::prefetchTransactionData` for now.

src/ledger/LedgerManagerImpl.cpp
marta-lokhova: I'm not sure this would be useful. We have this loop used only once, and it seems sufficient to simply utilize the public interface of `TransactionFrame` to get accounts.

src/transactions/TransactionUtils.cpp
```diff
@@ -15,39 +15,97 @@
 namespace stellar
 {
 
-LedgerTxnEntry
-loadAccount(AbstractLedgerTxn& ltx, AccountID const& accountID)
+LedgerKey
+getLedgerKey(AccountID const& accountID)
```
vogel: I like those utils! I think those could even have shorter names, like just ledgerKey, it would make them into some kind of constructors (and I generally don't like 'get' prefixes)

src/transactions/ManageDataOpFrame.cpp
```diff
@@ -96,4 +96,12 @@ ManageDataOpFrame::doCheckValid(Application& app, uint32_t ledgerVersion)
 
     return true;
 }
+
+std::unordered_set<LedgerKey>
+ManageDataOpFrame::getLedgerKeysToPrefetch(Application& app)
+{
+    std::unordered_set<LedgerKey> keys;
+    keys.insert(stellar::getLedgerKey(getSourceID(), mManageData.dataName));
```
vogel: why is namespace necessary in getLedgerKey call?

src/transactions/PathPaymentOpFrame.cpp
```diff
@@ -286,4 +286,37 @@ PathPaymentOpFrame::doCheckValid(Application& app, uint32_t ledgerVersion)
...
+    {
+        if (asset.type() == ASSET_TYPE_NATIVE)
+        {
+            continue;
+        }
+
+        auto issuer = getIssuer(asset);
+        keys.insert(stellar::getLedgerKey(issuer));
+
+        if (asset == mPathPayment.destAsset)
```
vogel: That can be outside of loop.

src/transactions/PathPaymentOpFrame.cpp
```diff
@@ -286,4 +286,37 @@ PathPaymentOpFrame::doCheckValid(Application& app, uint32_t ledgerVersion)
...
+        {
+            keys.insert(stellar::getLedgerKey(mPathPayment.destination,
+                                              mPathPayment.destAsset));
+            keys.insert(
```
vogel: For source it should be trustline with srcAsset

src/ledger/LedgerManagerImpl.cpp
```diff
@@ -1003,6 +1042,27 @@ LedgerManagerImpl::applyTransactions(std::vector<TransactionFramePtr>& txs,
...
+    logTxApplyMetrics(ltx, numTxs, numOps);
+}
+
+void
+LedgerManagerImpl::logTxApplyMetrics(AbstractLedgerTxn& ltx, size_t numTxs,
+                                     size_t numOps)
+{
+    auto ledgerSeq = ltx.loadHeader().current().ledgerSeq;
+
+    CLOG(DEBUG, "Ledger") << "Ledger: " << ledgerSeq << " txs: " << numTxs
```
vogel: Those two can probably be merged

src/ledger/LedgerTxnImpl.h
```diff
@@ -554,6 +564,60 @@ class LedgerTxnRoot::Impl
...
+
+    std::unordered_set<LedgerKey> getBatchToPrefetch(LedgerEntryType type);
+
+    void insertPrefetched(LedgerKey const& key,
+                          std::shared_ptr<LedgerEntry const> entry);
+
+    void clearAllPrefetched();
+
+  private:
+    std::list<LedgerKey const> mAccountsQueue;
```
vogel: my compiler complains about using const types in std::list (clang++)

src/ledger/LedgerTxn.h
```diff
@@ -598,5 +618,10 @@ class LedgerTxnRoot : public AbstractLedgerTxnParent
     void writeSignersTableIntoAccountsTable();
     void encodeDataNamesBase64();
     void encodeHomeDomainsBase64();
+
+    void prefetch(LedgerKey const& key);
+    void prefetchNow(LedgerKey const& key);
```
vogel: what is use case for that?

src/ledger/LedgerTxn.cpp
```diff
@@ -1749,22 +1948,64 @@ LedgerTxnRoot::Impl::getNewestVersion(LedgerKey const& key) const
...
+    keysToPrefetch.insert(key);
+
     std::shared_ptr<LedgerEntry const> entry;
     try
     {
         switch (key.type())
         {
         case ACCOUNT:
-            entry = loadAccount(key);
+            if (!doPrefetch)
```
vogel: I would prefer to split that method into several smaller ones:
- if (doPrefetch) condition could be make top one
- loadEntry and bulkLoadEntries methods could be extracted, first one would do what that try block was doing before the PR, and bulkLoadEntries only do the prefetching branch:


```
if (doPrefetch)
{
  bulkLoadEntries(keysToPrefetch);
  entry = updatePrefetcherCache(res, key);
}
else
{
  entry = loadAccount(key);
}
```

src/ledger/LedgerTxn.cpp
```diff
@@ -57,6 +58,156 @@ AbstractLedgerTxnParent::~AbstractLedgerTxnParent()
...
+LedgerTxnRoot::Impl::Prefetcher::hasNextToPrefetch(LedgerEntryType type)
+{
+    auto const& queue = getQueueByType(type);
+    return !queue.empty();
+}
+
+LedgerKey
+LedgerTxnRoot::Impl::Prefetcher::getNextToPrefetch(LedgerEntryType type)
+{
+    auto& queue = getQueueByType(type);
```
vogel: then calling from `getBatchToPrefetch`, this is calling `getQueueByType` mBatchSize times even when result will always be the same

src/ledger/LedgerTxn.cpp
```diff
@@ -57,6 +58,156 @@ AbstractLedgerTxnParent::~AbstractLedgerTxnParent()
...
+    mAccountsQueue.clear();
+    mDataQueue.clear();
+    mOffersQueue.clear();
+    mTrustlinesQueue.clear();
+}
+
+bool
+LedgerTxnRoot::Impl::Prefetcher::hasNextToPrefetch(LedgerEntryType type)
+{
+    auto const& queue = getQueueByType(type);
```
vogel: then calling from `getBatchToPrefetch`, this is calling `getQueueByType` mBatchSize times even when result will always be the same

src/ledger/LedgerTxn.h
```diff
@@ -276,6 +276,25 @@ class EntryIterator
     LedgerKey const& key() const;
 };
 
+class AbstractPrefetcher
```
vogel: This name is misleading. It's more like AbstractPrefetchQueue

src/ledger/LedgerTxn.h
```diff
@@ -276,6 +276,25 @@ class EntryIterator
...
+{
+  public:
+    virtual ~AbstractPrefetcher();
+
+    virtual void queueForPrefetch(LedgerKey const& key, bool immediate) = 0;
```
vogel: Probably can drop "prefetch" suffix from all of the methods

src/ledger/LedgerTxn.cpp
```diff
@@ -1749,22 +1948,64 @@ LedgerTxnRoot::Impl::getNewestVersion(LedgerKey const& key) const
         return getFromEntryCache(cacheKey);
```
jonjove: I'm really not sure about having two layers of caches, I know @MonsieurNicolas was in favor of this originally but I don't remember why. The way I see it, only `LedgerTxnRoot` can add things to either cache. Since both caches are in complete control of `LedgerTxnRoot`, they should probably be consolidated into one cache. Whether the cache needs to be stored in the `Prefetcher` may be an overreach of responsibilities. As it stands now it is possible that something in the prefetch queue is also in the entry cache, leading to unnecessary loads. Perhaps the `Prefetcher` should just have read-only access to a cache stored somewhere else? Thoughts?

graydon:
First: I should apologize for being so late to this party, I realize y'all
have been discussing the design of this system for a while and I haven't
been participating when I really should have. I recognize that the resulting
"why don't we just do something much simpler" design feedback here is .. a
bit awkward, and if there's a strong reason to keep it as-is I am certainly
willing to listen!

However, I chatted a bit with @marta-lokhova today to get some backstory on
the design because I was having those "why don't we just" feelings: like maybe
"just" populating the primary entry cache, and "just" loading everything we
might need as a `set<LedgerKey>` rather than a queue with cancellation, and
so forth.

In that discussion most of what came up in the design considerations for
the existing approach had to do with memory pressure: concern about
fetching things that could potentially evict other more-important things
from the cache. That's of course a valid concern, but my recollection was
that we're really sizing the entry cache too small anyways, and that
entries are on the order of like 200-ish bytes anyways.

Rather than speculate further, I wrote a small test that opens a bucket and
reads every live entry in it into memory. I used a reasonably large recent
bucket from the live network with 853,971 live entries, and tracked heap
allocation.

Heaptrack describes this run as:

~~~
bytes allocated in total (ignoring deallocations): 217.99MB (60.17MB/s)
calls to allocation functions: 1181865 (326211/s)
temporary memory allocations: 2901 (800/s)
peak heap memory consumption: 212.26MB
~~~

and massif's largest survey says:

~~~
--------------------------------------------------------------------------------
  n        time(i)         total(B)   useful-heap(B) extra-heap(B)    stacks(B)
--------------------------------------------------------------------------------
 76  6,257,595,480      222,558,656      211,990,212    10,568,444            0
 77  6,268,771,741      222,621,120      212,050,397    10,570,723            0
~~~

So I think 212MB is about accurate for loading that bucket, which gives us
an average 248 bytes per ledger entry of memory pressure.

So, looking at that I think, yeah, that we can probably "just" size the
primary entry cache big enough not to worry about evictions at all (except
in some sort of bug/DoS scenario -- it's fine to give it _some_ cap just "for
safety"). Our current worst-case txn * ops rate might need 5000 entries per
ledger, so let's give ourselves a 20x safety margin and set the cache at
100k entries, that's still only going to cost 24mb under quite heavy load (20k
ops/sec).

If we're not worried about evictions, I _think_ we can use a simpler
set-based preload-candidate-collection and bulk-loading strategy, yes? Or
am I missing some other dimension of the problem?

src/transactions/PathPaymentOpFrame.cpp
marta-lokhova: Good catch, thank you!

src/transactions/PathPaymentOpFrame.cpp
marta-lokhova: Good idea, done.

src/ledger/LedgerTxn.cpp
marta-lokhova: This is now resolved in favor of having a single cache of large enough size.

src/transactions/TransactionUtils.cpp
marta-lokhova: I would prefer to keep these as is and avoid confusion between `ledgerKey` and `LedgerKey`.

PR #1921, opened 2019-01-22
Use std::locale::classic more places.

graydon:
Thoroughly uninteresting change. First hunk switches us from constructing (and seemingly not even using?) a `locale("C")` object in the middle of a relatively hot loop -- hot enough to show up significantly in some testing profiles -- to using `locale::classic()` explicitly, which appears to be an identical (but static) object. Second set of hunks just uses `locale::classic()` in places where we _were_ using a static `locale("C")` object, but inexplicably using our own rather than the system-provided one.

# Checklist
- [x] Compiles
- [x] Ran all tests

### DIFF
--- a/src/ledger/LedgerTestUtils.cpp
+++ b/src/ledger/LedgerTestUtils.cpp
@@ -45,7 +45,7 @@ template <typename T>
 void
 replaceControlCharacters(T& s, int minSize)
 {
-    std::locale loc("C");
+    auto& loc = std::locale::classic();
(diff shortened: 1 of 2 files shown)


MonsieurNicolas:
r+ 1a63f2c229f4f236bdf54e6ac36b65d6effa8a62

MonsieurNicolas:
Thank you.
ah we were supposed to change this when we've run out of ideas on what to do for performance improvements :)
Pretty sure we've also got some hidden `sleep` statements in the code

PR #1922, opened 2019-01-22
Database Schema: Base-64 Encode Data Names and Home Domains

jonjove:
Database schema upgrade to base-64 encode data names and home domains.

### DIFF
--- a/src/database/Database.cpp
+++ b/src/database/Database.cpp
@@ -124,9 +124,14 @@ Database::applySchemaUpgrade(unsigned long vers)
         break;
 
     case 8:
+        // Update schema for signers
         mSession << "ALTER TABLE accounts ADD signers TEXT";
         mApp.getLedgerTxnRoot().writeSignersTableIntoAccountsTable();
         mSession << "DROP TABLE IF EXISTS signers";
+
+        // Update schema for base-64 encoding
+        mApp.getLedgerTxnRoot().encodeDataNamesBase64();
+        mApp.getLedgerTxnRoot().encodeHomeDomainsBase64();
         break;
 
     default:
--- a/src/ledger/LedgerTxnAccountSQL.cpp
+++ b/src/ledger/LedgerTxnAccountSQL.cpp
@@ -160,12 +160,12 @@ LedgerTxnRoot::Impl::writeSignersTableIntoAccountsTable()
     }
 
     size_t numAccountsUpdated = 0;
-    for (auto const& kv : signersByAccount)
+    for (auto& kv : signersByAccount)
     {
-        assert(std::adjacent_find(kv.second.begin(), kv.second.end(),
-                                  [](Signer const& lhs, Signer const& rhs) {
-                                      return !(lhs.key < rhs.key);
-                                  }) == kv.second.end());
+        std::sort(kv.second.begin(), kv.second.end(),
+                  [](Signer const& lhs, Signer const& rhs) {
+                      return lhs.key < rhs.key;
+                  });
         std::string signers(decoder::encode_b64(xdr::xdr_to_opaque(kv.second)));
 
         auto prep = mDatabase.getPreparedStatement(
--- a/src/ledger/LedgerTxnAccountSQL.cpp
+++ b/src/ledger/LedgerTxnAccountSQL.cpp
@@ -61,7 +61,7 @@ LedgerTxnRoot::Impl::loadAccount(LedgerKey const& key) const
     }
 
     account.accountID = key.account().accountID;
-    account.homeDomain = homeDomain;
+    decoder::decode_b64(homeDomain, account.homeDomain);
 
     bn::decode_b64(thresholds.begin(), thresholds.end(),
                    account.thresholds.begin());
@@ -219,7 +219,7 @@ LedgerTxnRoot::Impl::insertOrUpdateAccount(LedgerEntry const& entry,
     }
 
     std::string thresholds(decoder::encode_b64(account.thresholds));
-    std::string homeDomain(account.homeDomain);
+    std::string homeDomain(decoder::encode_b64(account.homeDomain));
 
     soci::indicator signersInd = soci::i_null;
     std::string signers;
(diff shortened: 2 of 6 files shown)


src/ledger/LedgerTxnDataSQL.cpp
```diff
@@ -127,4 +127,71 @@ LedgerTxnRoot::Impl::dropData()
...
+    mEntryCache.clear();
+    mBestOffersCache.clear();
+
+    mDatabase.getSession()
+        << "ALTER TABLE accountdata RENAME TO accountdataunencoded;";
+    dropData();
+
+    size_t const LIMIT = 4096;
+    size_t offset = 0;
+    auto dataToEncode = loadDataToEncode(mDatabase, LIMIT, offset);
```
MonsieurNicolas: I would ditch the LIMIT offset (and loop) given that we have order of 100k data entries total

src/database/Database.cpp
```diff
@@ -51,7 +51,7 @@ using namespace std;
 
 bool Database::gDriversRegistered = false;
 
-static unsigned long const SCHEMA_VERSION = 8;
+static unsigned long const SCHEMA_VERSION = 9;
```
MonsieurNicolas: no need to bump the version: I imagine that we're going to merge this into the upcoming release

src/database/Database.cpp
```diff
@@ -129,6 +129,11 @@ Database::applySchemaUpgrade(unsigned long vers)
         mSession << "DROP TABLE IF EXISTS signers";
         break;
 
+    case 9:
```
MonsieurNicolas: this commit should be after the one that introduces `encodeHomeDomainsBase64`

src/ledger/LedgerTxnAccountSQL.cpp
```diff
@@ -330,4 +330,70 @@ LedgerTxnRoot::Impl::dropAccounts()
         << "CREATE INDEX accountbalances ON accounts (balance) WHERE "
            "balance >= 1000000000";
 }
+
+static std::unordered_map<std::string, std::string>
+loadHomeDomainsToEncode(Database& db, size_t limit, size_t offset)
+{
+    std::string accountID, homeDomain;
+
+    std::string sql = "SELECT accountid, homedomain FROM accounts "
```
MonsieurNicolas: you can make this a lot faster by adding `WHERE length(homedomain) > 0`

src/ledger/LedgerTxnAccountSQL.cpp
```diff
@@ -330,4 +330,70 @@ LedgerTxnRoot::Impl::dropAccounts()
...
+void
+LedgerTxnRoot::Impl::encodeHomeDomainsBase64()
+{
+    throwIfChild();
+    mEntryCache.clear();
+    mBestOffersCache.clear();
+
+    size_t const LIMIT = 4096;
+    size_t offset = 0;
+    auto homeDomainsToEncode =
```
MonsieurNicolas: same thing, you can safely ditch the LIMIT/offset and related logic, with the filtering we're talking less than 100k accounts which is safe to keep in memory

src/ledger/LedgerTxnAccountSQL.cpp
```diff
@@ -330,4 +330,70 @@ LedgerTxnRoot::Impl::dropAccounts()
...
+
+    std::string sql =
+        "UPDATE accounts SET homedomain = :v1 WHERE accountid = :v2";
+
+    auto prep = db.getPreparedStatement(sql);
+    auto& st = prep.statement();
+    st.exchange(soci::use(encodedHomeDomain));
+    st.exchange(soci::use(accountID));
+    st.define_and_bind();
+    st.execute(true);
```
MonsieurNicolas: should we check that nb effected row is 1?

src/ledger/LedgerTxnAccountSQL.cpp
```diff
@@ -330,4 +330,70 @@ LedgerTxnRoot::Impl::dropAccounts()
...
+    auto prep = db.getPreparedStatement(sql);
+    auto& st = prep.statement();
+    st.exchange(soci::into(accountID));
+    st.exchange(soci::into(homeDomain));
+    st.exchange(soci::use(limit));
+    st.exchange(soci::use(offset));
+    st.define_and_bind();
+    st.execute(true);
+
+    std::unordered_map<std::string, std::string> res;
```
MonsieurNicolas: nit: you don't really need a k:v type of structure for this (and return type of the function), a `std::vector<std::pair<>>` is enough

src/ledger/LedgerTxnDataSQL.cpp
MonsieurNicolas: technically if we do that we don't even need to keep the data around in `accountdataunencoded` (ie: we load all data, recreate the table, write the data), so much simpler logic ; one of the side effects of having 2 tables might be unnecessary overhead that will have to be vacuumed.

graydon: Yeah, I agree there's unlikely to be a problem doing it all in memory.

src/ledger/LedgerTxnImpl.h
```diff
@@ -421,5 +421,17 @@ class LedgerTxnRoot::Impl
     // - the prepared statement cache may be, but is not guaranteed to be,
     //   modified
     void writeSignersTableIntoAccountsTable();
+
+    // encodedDataNamesBase64 has the basic exception safety guarantee. If it
+    // throws an exception, then
+    // - the prepared statement cache may be, but is not guaranteed to be,
+    //   modified
+    void encodeDataNamesBase64();
```
vogel: Do we need those in that class? Woudn't static methods/free function be better here? Those are one-time calls (even one-time per instance, not per launch), so why pollute that interface?

vogel: OK, but I don't agree with that solution as I would like to see a layer below LedgerTxnRoot to handle the queries.

For example dropOffers does a lot of unnecessary things, like "throwIfChild" (it is only called from Database::initialize!). I think we could make it as safe as it is now with a different solution (but this is very low priority).

And if you really want all SQL statements in LedgerTxnRoot, why not move there also all of `ALTER TABLE` from liabilities and signers changes?

MonsieurNicolas: yeah I would like to see some separation too (and/or limit public methods) but that is outside of the scope of this PR and seems like lower priority. Open an issue to discuss that if you feel like we should push for it.

MonsieurNicolas:
r+ 9361012241017781fe537eff627800e9419ad39c

MonsieurNicolas:
I opened a couple issues related to the bug that you caught in the account signers change that you added to this PR

MonsieurNicolas:
this was reverted by https://github.com/stellar/stellar-core/pull/1933

PR #1923, opened 2019-01-22
Upgrade sqlite and add carray

graydon:
This is a common prefix of some sqlite surgery needed by work in PRs both myself and @jonjove are working on. Sqlite gets bumped up to a version that has upsert support, and the `carray` extension gets added. Shouldn't really be noticeable outside of the forthcoming changes that depend on it, but I figured I'd split it out for landing on its own.

# Checklist
- [x] Compiles
- [x] Ran all tests

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -143,7 +143,6 @@ AS_IF([test x != "x$sanitizeopts"], [
 
   # compile our own libraries when sanitizers are enabled
   libsodium_INTERNAL=yes
-  sqlite3_INTERNAL=yes
   xdrpp_INTERNAL=yes
 
 ])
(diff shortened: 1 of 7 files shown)


configure.ac
```diff
@@ -215,7 +215,7 @@ fi
 # extension, which is compiled-out of most platform sqlites. We therefore
 # always use our own bundled copy, version 3.26.0 at the time of this
 # writing.
-sqlite3_CFLAGS='-isystem $(top_srcdir)/lib/sqlite -DSQLITE_CORE'
+sqlite3_CFLAGS='-isystem $(top_srcdir)/lib/sqlite -DSQLITE_CORE -DSQLITE_OMIT_LOAD_EXTENSION=1'
```
jonjove: Nitpick: I think it would be better if this change (re-adding this define) was squashed into the first commit in this PR.

graydon: ok

graydon: fixed

jonjove:
r+ f7e11f0

Issue #1885, opened 2018-12-20
crash in command line catchup


MonsieurNicolas:
We're seeing intermittent crashes in acceptance tests.

Seen in d51f6222

Couple examples

```
/start: line 70:    16 Aborted                 (core dumped) stellar-core $CONFIG_OPTION --catchup-at $FROM --catchup-to $TO --output-file "${OUTPUT}"
parallel: This job failed:
./scripts/partial_catchup 23 24 1282879 testnet
```

```
/start: line 70:    18 Aborted                 (core dumped) stellar-core $CONFIG_OPTION --catchup-at $FROM --catchup-to $TO --output-file "${OUTPUT}"
parallel: This job failed:
./scripts/partial_catchup 19 24 1195327 testnet
```

MonsieurNicolas:
are you still looking at this @marta-lokhova ?

marta-lokhova:
@MonsieurNicolas I could not reproduce it locally, also it looks like there have been no failures since Dec 17. We could maybe switch log level to DEBUG and see if it fails again?

marta-lokhova:
Actually it happened again last night, seems like the same place every time.

marta-lokhova:
Update on this: we have not seen any crashes since #1929 was merged. My current theory is that the data race on `NtpClient::mCallbackCalled` caused `NtpWork` to be scheduled for completion twice (when `NtpWork` hits a timeout, and when `NtpClient` finishes on the background thread). This could cause strange things happen to the Work object. 

Closing this issue (will re-open if it comes back).

Issue #1815, opened 2018-10-24
Rethink how we check whether node's time is synchronized


marta-lokhova:
Looks like currently we check whether local time is synchronized with an NTP server (via NtpSynchronizationChecker/NtpWork/NtpClient). There are a few things:
* The implementation looks complex and I don't think it's actually using Work/WorkParent interfaces correctly, so it should be refactored/simplified. 
*  It also looks like there might be a potential race condition, since the work modifies `mIsSynchronized` from the worker thread (without using mutex or std::atomic)
* Additionally, @MonsieurNicolas proposed we might not even need to check against an NTP server, but against other nodes instead.

MonsieurNicolas:
A few more ideas here: ideally we can report both on how the local node is doing compared to the network *and* how other nodes are (not) nominating properly.

marta-lokhova:
@MonsieurNicolas I think trying to report how the other nodes are not nominating correctly may lead to false positives. Specifically, the ct picked for nomination depends on node externalizing latest ledger. So if the node is lagging for whatever reason, it may report other nodes as drifting (it will have ct higher than what it sees externalized) 

Reporting the other way around seems good enough: a node with drifted clock will see that its `ct` is behind whatever was externalized. Note that if the node is lagging, we may not get a warning, but it's probably ok, as long as the node recovers and start reporting the drift correctly (I'd prefer an occasional false negative rather than a false positive that may cause confusion among the operators). Let me know what you think.

MonsieurNicolas:
I think you're mixing "a node not nominating in time" with "a node has a bad clock".

The first kind of problem leads to things like:
* node starts nomination late, causing consensus to stall and timeout during nomination as a different leader is selected

The second kind of problem leads to:
* "jitter" in close time, that gives the impression that some ledgers were slow/fast.
     * If a node has a fast clock, ct will be set too high, which will appear to be a slow ledger in the dashboard. The next leader will rectify the drift back (which will show up as a "fast ledger" in the dashboard).
     * The other way is true as well. If a node has a slow clock, it will nominate with ct too small which looks like a fast ledger.
* if the delta is too high, may lead to multiple "1 second" ledgers (as nodes are forced to vote with `ct=lcl.ct+1`)
* if the delta is "way too high", the value will be considered invalid by other nodes until it gets back into the validity range or may cause the node to throw away messages from other nodes (and fall out of sync).

This issue deals with the second kind of problem: "bad clock".

[...]

PR #1929, opened 2019-01-25
Remove NtpSynchronizationChecker and related classes

marta-lokhova:
Extracted this from the Work PR (#1819), as this change is actually unrelated to that PR, and it *might* fix crashes in command line catchup (#1885). Specifically, I don't think this code is properly synchronized, causing worker thread to race with main.


MonsieurNicolas:
you forgot to remove the deleted files from the visual studio config files

marta-lokhova:
@MonsieurNicolas done

Geod24:
As a casual observer, I was wondering what was the reasoning for removing this?

marta-lokhova:
@Geod24 Mainly because we'd like to change how we check node's synchronization, and possibly make the checks against other nodes (additionally, we found a few synchronization issues in the current code). I opened an issue to track this also #1815

PR #1936, opened 2019-01-31
Periodically refresh known/preferred peer IPs

### LINKED ISSUE #1908: properly refresh peers configured by hostname in configuration file
Right now peers specified in the configuration file (`KNOWN_PEERS` and `PREFERRED_PEERS`) have their IP resolved on startup.

Problem is that if the IP address of one of those peers changes, stellar-core won't notice until restarted.

A way to avoid this could be to periodically (every X minutes) resolve the hostnames and add the new IPs to the list of peers.

marta-lokhova:
Resolves #1908.
note that as part of this we now avoid resolving preferred peer IPs twice in `storeConfigPeers`

### DIFF
--- a/src/overlay/OverlayManagerImpl.cpp
+++ b/src/overlay/OverlayManagerImpl.cpp
@@ -52,6 +52,8 @@ namespace stellar
 using namespace soci;
 using namespace std;
 
+constexpr std::chrono::seconds PEER_IP_RESOLVE_DELAY(600);
+
 OverlayManagerImpl::PeersList::PeersList(
     OverlayManagerImpl& overlayManager,
     medida::MetricsRegistry& metricsRegistry, std::string directionString,
@@ -229,6 +231,7 @@ OverlayManagerImpl::OverlayManagerImpl(Application& app)
     , mAuthenticatedPeersSize(app.getMetrics().NewCounter(
           {"overlay", "connection", "authenticated"}))
     , mTimer(app)
+    , mPeerIPTimer(app)
     , mFloodGate(app)
 {
     mPeerSources[PeerType::INBOUND] = std::make_unique<RandomPeerSource>(
@@ -253,7 +256,8 @@ OverlayManagerImpl::start()
     {
         mTimer.async_wait(
             [this]() {
-                storeConfigPeers();
+                this->storeConfigPeers();
+                this->triggerPeerResolution();
                 this->tick();
             },
             VirtualTimer::onFailureNoop);
@@ -301,24 +305,35 @@ OverlayManagerImpl::getPeersList(Peer* peer)
 }
 
 void
-OverlayManagerImpl::storePeerList(std::vector<std::string> const& list,
-                                  bool setPreferred)
+OverlayManagerImpl::storePeerList(std::vector<PeerBareAddress> const& addresses,
+                                  bool setPreferred, bool startup)
 {
     auto typeUpgrade = setPreferred
                            ? PeerManager::TypeUpdate::SET_PREFERRED
                            : PeerManager::TypeUpdate::UPDATE_TO_OUTBOUND;
-
-    for (auto const& peerStr : list)
+    if (setPreferred)
     {
-        try
+        mConfigurationPreferredPeers.clear();
+    }
+
+    for (auto const& peer : addresses)
+    {
+        if (setPreferred)
         {
-            auto address = PeerBareAddress::resolve(peerStr, mApp);
-            getPeerManager().update(address, typeUpgrade,
+            mConfigurationPreferredPeers.insert(peer);
+        }
+
+        if (startup)
+        {
+            getPeerManager().update(peer, typeUpgrade,
                                     PeerManager::BackOffUpdate::HARD_RESET);
         }
-        catch (std::runtime_error&)
+        else
         {
-            CLOG(ERROR, "Overlay") << "Unable to add peer '" << peerStr << "'";
+            // If address is present in the DB, `update` will ensure
+            // type is correctly updated. Otherwise, a new entry is created.
+            // Note that this won't downgrade preferred peers back to outbound.
+            getPeerManager().update(peer, typeUpgrade);
         }
     }
 }
@@ -326,28 +341,51 @@ OverlayManagerImpl::storePeerList(std::vector<std::string> const& list,
 void
 OverlayManagerImpl::storeConfigPeers()
 {
-    // compute normalized mConfigurationPreferredPeers
-    std::vector<std::string> ppeers;
-    for (auto const& s : mApp.getConfig().PREFERRED_PEERS)
+    // Synchronously resolve and store peers from the config
+    storePeerList(resolvePeers(mApp.getConfig().KNOWN_PEERS), false, true);
+    storePeerList(resolvePeers(mApp.getConfig().PREFERRED_PEERS), true, true);
+}
+
+void
+OverlayManagerImpl::triggerPeerResolution()
+{
+    assert(!mResolvedPeers.valid());
+
+    // Trigger DNS resolution on the background thread
+    using task_t = std::packaged_task<ResolvedPeers()>;
+    std::shared_ptr<task_t> task = std::make_shared<task_t>([this]() {
+        if (!this->mShuttingDown)
+        {
+            auto known = resolvePeers(this->mApp.getConfig().KNOWN_PEERS);
+            auto preferred =
+                resolvePeers(this->mApp.getConfig().PREFERRED_PEERS);
+            return ResolvedPeers{known, preferred};
+        }
+        return ResolvedPeers{};
+    });
+
+    mResolvedPeers = task->get_future();
+    mApp.postOnBackgroundThread(bind(&task_t::operator(), task),
+                                "OverlayManager: resolve peer IPs");
+}
+
+std::vector<PeerBareAddress>
+OverlayManagerImpl::resolvePeers(std::vector<string> const& peers)
+{
+    std::vector<PeerBareAddress> addresses;
+    for (auto const& peer : peers)
     {
         try
         {
-            auto pr = PeerBareAddress::resolve(s, mApp);
-            auto r = mConfigurationPreferredPeers.insert(pr);
-            if (r.second)
-            {
-                ppeers.push_back(r.first->toString());
-            }
+            addresses.push_back(PeerBareAddress::resolve(peer, mApp));
         }
-        catch (std::runtime_error&)
+        catch (std::runtime_error& e)
         {
             CLOG(ERROR, "Overlay")
-                << "Unable to add preferred peer '" << s << "'";
+                << "Unable to resolve peer '" << peer << "': " << e.what();
         }
     }
-
-    storePeerList(mApp.getConfig().KNOWN_PEERS, false);
-    storePeerList(ppeers, true);
+    return addresses;
 }
 
 std::vector<PeerBareAddress>
@@ -399,6 +437,19 @@ OverlayManagerImpl::tick()
     CLOG(TRACE, "Overlay") << "OverlayManagerImpl tick";
 
     mLoad.maybeShedExcessLoad(mApp);
+
+    if (mResolvedPeers.valid() &&
+        mResolvedPeers.wait_for(std::chrono::nanoseconds(1)) ==
+            std::future_status::ready)
+    {
+        CLOG(TRACE, "Overlay") << "Resolved peers are ready";
+        auto res = mResolvedPeers.get();
+        storePeerList(res.known, false, false);
+        storePeerList(res.preferred, true, false);
+        mPeerIPTimer.expires_from_now(PEER_IP_RESOLVE_DELAY);
+        mPeerIPTimer.async_wait([this]() { this->triggerPeerResolution(); },
+                                VirtualTimer::onFailureNoop);
+    }
 
     auto availablePendingSlots = availableOutboundPendingSlots();
     auto availableAuthenticatedSlots = availableOutboundAuthenticatedSlots();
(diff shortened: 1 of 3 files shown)


src/overlay/OverlayManagerImpl.cpp
```diff
@@ -257,9 +261,43 @@ OverlayManagerImpl::start()
...
             },
             VirtualTimer::onFailureNoop);
+
+        mPeerIPTimer.async_wait([this]() { this->maybeRefreshPeerIPs(); },
+                                VirtualTimer::onFailureNoop);
     }
 }
 
+PeerBareAddress
+OverlayManagerImpl::getLatestPeerAddress(std::string const& peerStr,
```
vogel: That name is misleading, even if this method return the lastest address.
Problem is that is also updates peer cache, updates peer table and possibly updates mConfigurationPreferredPeers, so it is much more than getting that address.

My proposition is to make it responsible for less things.

1. isPreferred parameter could be removed, and updating mConfigurationPreferredPeers could be done in maybeRefreshPeerIPs
2. name could be changed to something like updatePeerAddressCache
3. call to addAddressToExistingPeer could also be moved to maybeRefreshPeerIPs, and condition if (storedAddress != address) should be added to PeerManager::addAddressToExistingPeer

src/overlay/OverlayManagerImpl.cpp
```diff
@@ -391,6 +426,34 @@ OverlayManagerImpl::connectTo(std::vector<PeerBareAddress> const& peers,
...
+            getLatestPeerAddress(s, true);
+        }
+        for (auto const& s : mApp.getConfig().KNOWN_PEERS)
+        {
+            getLatestPeerAddress(s, false);
+        }
+    }
+    catch (std::runtime_error& e)
+    {
+        CLOG(ERROR, "Overlay")
```
vogel: This fails after just the first error and does not let other peers get resolved even if the error only occurs on first one. Moving this try/catch inside getLatestPeerAddress would fix that issue.

src/overlay/OverlayManagerImpl.cpp
marta-lokhova: What do you think about instead renaming this function to something like `resolveIPAndMaybeUpdateCaches`? The reason I don't like splitting it is because all cache/db updates are consolidated in one place, so it's less overhead to track these. Also, `storedAddress != address` condition prevents from redundant cache inserts.

vogel: I think that with method's name so long it may be better to extract small utility class just for cache handling (if the cache will be still there, after resolving Nicolas comments).

vogel:
Idea with setting peer type to OUTBOUND in storeConfigPeers is a good one.

src/overlay/OverlayManagerImpl.cpp
```diff
@@ -301,53 +308,92 @@ OverlayManagerImpl::getPeersList(Peer* peer)
...
-                ppeers.push_back(r.first->toString());
-            }
+            addresses.push_back(PeerBareAddress::resolve(peer, mApp));
         }
-        catch (std::runtime_error&)
+        catch (std::runtime_error& e)
         {
-            CLOG(ERROR, "Overlay")
-                << "Unable to add preferred peer '" << s << "'";
+            CLOG(ERROR, "Overlay") << "Unable to resolve peer '" << peer << "'";
```
MonsieurNicolas: you should probably include the error message from `e` in the error

src/overlay/OverlayManagerImpl.cpp
```diff
@@ -253,7 +256,11 @@ OverlayManagerImpl::start()
     {
         mTimer.async_wait(
             [this]() {
-                storeConfigPeers();
+                storeConfigPeers(true);
+                mPeerIPTimer.expires_from_now(PEER_IP_RESOLVE_DELAY);
```
MonsieurNicolas: I wonder if we should just call this code from `tick` instead of using a dedicated timer? Less things to keep track of

MonsieurNicolas: Alternatively, you could just call
```
storeConfigPeers(true);
storeConfigPeers(false);
```
to avoid duplicating the scheduling.

src/overlay/OverlayManagerImpl.cpp
```diff
@@ -301,53 +308,92 @@ OverlayManagerImpl::getPeersList(Peer* peer)
 }
 
 void
-OverlayManagerImpl::storePeerList(std::vector<std::string> const& list,
-                                  bool setPreferred)
+OverlayManagerImpl::storePeerList(std::vector<PeerBareAddress> addresses,
```
MonsieurNicolas: pass `std::vector<PeerBareAddress> const& addresses` instead of by value

src/overlay/OverlayManagerImpl.cpp
```diff
@@ -301,53 +308,92 @@ OverlayManagerImpl::getPeersList(Peer* peer)
...
+                                  bool setPreferred, bool startup)
 {
     auto typeUpgrade = setPreferred
                            ? PeerManager::TypeUpdate::SET_PREFERRED
                            : PeerManager::TypeUpdate::UPDATE_TO_OUTBOUND;
-
-    for (auto const& peerStr : list)
+    for (auto const& peer : addresses)
```
MonsieurNicolas: you should probably clear `mConfigurationPreferredPeers` before the loop if we're setting preferred peers as it's supposed to represent the peers from config

src/overlay/OverlayManagerImpl.cpp
```diff
@@ -301,53 +308,92 @@ OverlayManagerImpl::getPeersList(Peer* peer)
...
+            // Note that this won't downgrade promoted peers back to outbound.
+            getPeerManager().update(peer, typeUpgrade);
         }
     }
 }
 
 void
-OverlayManagerImpl::storeConfigPeers()
+OverlayManagerImpl::storeKnownAndPreferred(
+    std::vector<PeerBareAddress> known, std::vector<PeerBareAddress> preferred,
```
MonsieurNicolas: pass by const reference -> don't need to `std::move`

src/overlay/OverlayManagerImpl.cpp
```diff
@@ -301,53 +308,92 @@ OverlayManagerImpl::getPeersList(Peer* peer)
...
+    {
+        // On startup, perform IP resolution synchronously on the main thread
+        auto known = resolvePeers(mApp.getConfig().KNOWN_PEERS);
+        auto preferred = resolvePeers(mApp.getConfig().PREFERRED_PEERS);
+        storeKnownAndPreferred(known, preferred, startup);
+    }
+    else
+    {
+        mApp.postOnBackgroundThread(
+            [this, startup]() {
```
MonsieurNicolas: don't need `startup` as it's `false`

src/overlay/OverlayManagerImpl.cpp
```diff
@@ -301,53 +308,92 @@ OverlayManagerImpl::getPeersList(Peer* peer)
...
+                auto known = resolvePeers(this->mApp.getConfig().KNOWN_PEERS);
```
MonsieurNicolas: as we're on a different thread...
1. how do we guarantee that `this` is valid? 
2. how do we ensure that things don't blow up on shutdown? In particular, is it safe to have this run for a long time while the main thread is tearing down overlay?

src/overlay/OverlayManagerImpl.cpp
```diff
@@ -301,53 +308,92 @@ OverlayManagerImpl::getPeersList(Peer* peer)
...
+                auto preferred =
+                    resolvePeers(this->mApp.getConfig().PREFERRED_PEERS);
+
+                this->mApp.postOnMainThread(
```
MonsieurNicolas: related to shutdown: can we actually post on main thread if overlay is shutting down?

src/overlay/OverlayManagerImpl.cpp
```diff
@@ -301,53 +308,92 @@ OverlayManagerImpl::getPeersList(Peer* peer)
...
+                            VirtualTimer::onFailureNoop);
+                    },
+                    "Overlay: storeConfigPeers");
+            },
+            "Overlay: resolve Peer IPs and trigger store");
+    }
+}
+
+std::vector<PeerBareAddress>
+OverlayManagerImpl::resolvePeers(std::vector<std::string> peers)
```
MonsieurNicolas: pass by const reference

src/overlay/OverlayManagerImpl.cpp
MonsieurNicolas: looks to me that a pattern that uses `std::future` would be cleaner and easier to reason about, you could `peek` into this future from the `tick` cb

src/overlay/OverlayManagerTests.cpp
```diff
@@ -118,24 +118,89 @@ class OverlayManagerTests
...
         auto cfg = getTestConfig();
         cfg.TARGET_PEER_CONNECTIONS = 5;
+        cfg.KNOWN_PEERS = threePeers;
+        cfg.PREFERRED_PEERS = fourPeers;
         app = createTestApplication<ApplicationStub>(clock, cfg);
     }
 
     void
-    test_addPeerList()
+    test_addPeerList(bool async = false)
```
MonsieurNicolas: as you're touching this, can you rename the `test_*` functions to use something closer to our naming convention?

src/overlay/OverlayManagerImpl.cpp
```diff
@@ -301,53 +308,92 @@ OverlayManagerImpl::getPeersList(Peer* peer)
...
+OverlayManagerImpl::storeKnownAndPreferred(
+    std::vector<PeerBareAddress> known, std::vector<PeerBareAddress> preferred,
+    bool startup)
+{
+    storePeerList(std::move(known), false, startup);
+    storePeerList(std::move(preferred), true, startup);
+};
+
+void
+OverlayManagerImpl::storeConfigPeers(bool startup)
```
vogel: This function has two branches that have nothing in common, so it will be better to just have two functions here (storeConfigPeersSync and storeConfigPeersAsync?).

vogel:
When we look at the whole picture, there are some members and methods tightly coupled with each other:

- storePeerList
- storeKnownAndPreferred
- storeConfigPeers
- resolvePeers
- mConfigurationPreferredPeers

Exctracting them to separate utility class would make OverlayManagerImpl smaller and more mainteinable, also creating new entity with clear name and purpose (ConfigPeerResolver?).

src/overlay/OverlayManagerImpl.cpp
```diff
@@ -301,53 +308,92 @@ OverlayManagerImpl::getPeersList(Peer* peer)
...
+            getPeerManager().update(peer, typeUpgrade,
                                     PeerManager::BackOffUpdate::HARD_RESET);
         }
-        catch (std::runtime_error&)
+        else
         {
-            CLOG(ERROR, "Overlay") << "Unable to add peer '" << peerStr << "'";
+            // If address is present in the DB, `update` will ensure
+            // type is correctly updated. Otherwise, a new entry is created.
+            // Note that this won't downgrade promoted peers back to outbound.
```
vogel: typo, should be:

Note that this won't downgrade *preferred* peers back to outbound.

src/overlay/OverlayManagerImpl.cpp
marta-lokhova: Agreed that `std::future` seems to be safer in this case. 
To address your questions:
1. From what I see, when application is destructing, it first joins all threads (while blocking), before destroying its member variables. This means that the background thread will finish its work while the app (hence `mOverlayManager`, and hence captured `this`) is still alive. 
2. I think the right course of action here would be to stop any activity as soon as OverlayManager starts shutting down, i.e. stop ticking and resolving any peers.

src/overlay/OverlayManagerImpl.cpp
marta-lokhova: (Marking this as resolved as this is not an issue anymore with `std::future` approach)

src/overlay/OverlayManagerImpl.cpp
marta-lokhova: I ended up going with your latter suggestion. With the former I think we'd end up resolving peers too often (although we can check timestamps in tick to resolve only after some time, but that would use a timer anyway)

src/overlay/OverlayManagerImpl.cpp
marta-lokhova: Sure, updated to `storeConfigPeers` and `triggerPeerResolution`.

marta-lokhova:
Updated this, main change being the use of `std::future` instead of posting back to main thread.

@vogel wanted to follow up on your comment -- some code got moved around, and I removed `storeKnownAndPreferred`. With the updated code, do you think it's still worth separating the functions/members you mentioned? I also realize that overlay tests depend a lot on the privates of OverlayManager, as OverlayManagerTests is a friend of OverlayManager :(

vogel:
In my opinion it is almost always worth to separate members - it allows for smaller responsibilities, better testability and readability and also enforces encapsulation. I also think that the need of OverlayManagerTests is a result of OverlayManager being too big in general (how to unit test a class that has 30 methods and that depends on Application class that is king of [god object](https://en.wikipedia.org/wiki/God_object) itself?).

I would vote for separation, but I'm not sure this idea would get majority of votes.

marta-lokhova:
The problem is that I think OverlayManager itself is kind of tightly coupled with the functionality you mentioned (like it decides whether resolution should be sync or async, it relies on mConfigurationPreferredPeers, prevents resolving when shutting down, etc)

MonsieurNicolas:
I think the cleanup idea is not in scope for this PR. I am r+ this as it is

MonsieurNicolas:
r+ 54affb902158338874f2ee1e90bf702b3671dac5

PR #1944, opened 2019-02-04
set peers from config as OUTBOUND by default

vogel:

# Description

Set type of peers stored in configuration to be at least 'outbound'.
# Checklist
- [x] Compiles
- [x] Ran all tests

### DIFF
--- a/src/overlay/OverlayManagerImpl.cpp
+++ b/src/overlay/OverlayManagerImpl.cpp
@@ -304,8 +304,9 @@ void
 OverlayManagerImpl::storePeerList(std::vector<std::string> const& list,
                                   bool setPreferred)
 {
-    auto typeUpgrade = setPreferred ? PeerManager::TypeUpdate::SET_PREFERRED
-                                    : PeerManager::TypeUpdate::KEEP;
+    auto typeUpgrade = setPreferred
+                           ? PeerManager::TypeUpdate::SET_PREFERRED
+                           : PeerManager::TypeUpdate::UPDATE_TO_OUTBOUND;
 
     for (auto const& peerStr : list)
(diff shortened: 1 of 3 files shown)


MonsieurNicolas:
r+ d4ff1e6d

vogel:
Reformatted.

src/overlay/OverlayManagerImpl.cpp
```diff
@@ -304,8 +304,9 @@ void
 OverlayManagerImpl::storePeerList(std::vector<std::string> const& list,
                                   bool setPreferred)
 {
-    auto typeUpgrade = setPreferred ? PeerManager::TypeUpdate::SET_PREFERRED
-                                    : PeerManager::TypeUpdate::KEEP;
+    auto typeUpgrade = setPreferred
+                           ? PeerManager::TypeUpdate::SET_PREFERRED
+                           : PeerManager::TypeUpdate::UPDATE_TO_OUTBOUND;
```
marta-lokhova: Why do we need a new type here (and not just use `PeerManager::TypeUpdate::SET_OUTBOUND`)? @vogel @MonsieurNicolas

vogel: SET_OUTBOUND is unconditional, UPDATE_TO_OUTBOUND keeps preferred flag - that is for case when peer that is in KNOWN_PEERS is also in PREFERRED_PEER_KEYS, so we don't know at this moment that it is preferred, but we learned about it during some previous connection after authentication phase.

marta-lokhova: Thanks for explaining!

PR #1945, opened 2019-02-04
[trivial] change sqlite3 to postgresql

Varunram:
Trivial change in the sample testnet config file which is linked by other docs.

### DIFF
--- a/docs/stellar-core_testnet.cfg
+++ b/docs/stellar-core_testnet.cfg
@@ -8,7 +8,7 @@ KNOWN_PEERS=[
 "core-testnet2.stellar.org",
 "core-testnet3.stellar.org"]
 
-DATABASE="sqlite3://stellar.db"
+DATABASE="postgresql://dbname=stellar user=stellar"
 UNSAFE_QUORUM=true
 FAILURE_SAFETY=1
 


MonsieurNicolas:
Sorry can you expand a bit? This example config works fine and requires the least amount of setup.

Varunram:
I think stellar primarily uses postgresql? (do correct me if I'm wrong here). I am aware you can use sqlite3 instead of postgresql, but if a person wants to transition from testnet to mainnet using the same config but different validators, sqlite3 would not be a wise choice, hence the PR

MonsieurNicolas:
It depends on what people want to do. For full validators, right now sqlite3 happens to be the most efficient (both in terms of performance and setup) way to run stellar-core ; there are plenty of other files with examples that use postgresql for more complicated setup (such as using it with Horizon), but this doesn't belong into the main core source code.

Varunram:
I see, thanks! I think it would be nice to have a comment mentioning this similar to the one at https://github.com/stellar/stellar-core/blob/d5821103fbb625bc0a5b3cda0e9a9faf54f96d5d/docs/stellar-core_example.cfg. But this PR's concern has been addressed, so I shall close this one.

PR #1946, opened 2019-02-04
fix: crash in PeerManager::getPeersToSend

MonsieurNicolas:
function was returning more peers than asked for, resulting in crash:

```
2019-02-04T21:41:10.804 GDKXE [Overlay INFO] New connected peer (empty)
stellar-core: overlay/Peer.cpp:353: void stellar::Peer::sendPeers(): Assertion `peers.size() <= maxPeerCount' failed.
exited 23
```


marta-lokhova:
r+ 54ae9f2eb18506c201f4082e4a82866d7bc2e850

PR #1948, opened 2019-02-05
Delayed scp status

MonsieurNicolas:
# Description

This PR adds extra information to the quorum status, by providing hints on nodes that are participating to consensus but that are behind (ie, we only receive a PREPARE message instead of a CONFIRM/EXTERNALIZE)


graydon:
r+ 5af207e

PR #1955, opened 2019-02-09
better use of floodmap (lower memory footprint)

MonsieurNicolas:
This PR prioritizes keeping in the floodmap messages for the current SCP validity bracket

### DIFF
--- a/src/herder/HerderImpl.cpp
+++ b/src/herder/HerderImpl.cpp
@@ -377,12 +377,6 @@ HerderImpl::recvSCPEnvelope(SCPEnvelope const& envelope)
             << " i:" << envelope.statement.slotIndex
             << " a:" << mApp.getStateHuman();
 
-    if (envelope.statement.nodeID == getSCP().getLocalNode()->getNodeID())
-    {
-        CLOG(DEBUG, "Herder") << "recvSCPEnvelope: skipping own message";
-        return Herder::ENVELOPE_STATUS_DISCARDED;
-    }
-
     mSCPMetrics.mEnvelopeReceive.Mark();
 
     uint32_t minLedgerSeq = getCurrentLedgerSeq();
@@ -414,6 +408,12 @@ HerderImpl::recvSCPEnvelope(SCPEnvelope const& envelope)
                               << envelope.statement.slotIndex << "( "
                               << minLedgerSeq << "," << maxLedgerSeq << ")";
         return Herder::ENVELOPE_STATUS_DISCARDED;
+    }
+
+    if (envelope.statement.nodeID == getSCP().getLocalNode()->getNodeID())
+    {
+        CLOG(DEBUG, "Herder") << "recvSCPEnvelope: skipping own message";
+        return Herder::ENVELOPE_STATUS_SKIPPED_SELF;
     }
 
     auto status = mPendingEnvelopes.recvSCPEnvelope(envelope);
--- a/src/overlay/Floodgate.h
+++ b/src/overlay/Floodgate.h
@@ -38,7 +38,7 @@ class Floodgate
 
         uint32_t mLedgerSeq;
         StellarMessage mMessage;
-        std::set<Peer::pointer> mPeersTold;
+        std::set<std::string> mPeersTold;
 
         FloodRecord(StellarMessage const& msg, uint32_t ledger,
                     Peer::pointer peer);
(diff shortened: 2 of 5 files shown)


MonsieurNicolas:
added a few extra changes to this:
* we don't actually need to always send the entire local SCP state, and just rely on the selective pull. this avoids pushing a lot of SCP messages when new peers connect
* reduce the lifetime of the TCPPeer connection in floodmap
* update to latest release of asio

src/overlay/Peer.cpp
```diff
@@ -778,7 +776,11 @@ Peer::recvSCPMessage(StellarMessage const& msg)
                                 ? mRecvSCPExternalizeTimer.TimeScope()
                                 : (mRecvSCPNominateTimer.TimeScope()))));
 
-    mApp.getHerder().recvSCPEnvelope(envelope);
+    auto res = mApp.getHerder().recvSCPEnvelope(envelope);
+    if (res != Herder::ENVELOPE_STATUS_DISCARDED)
```
graydon: I'm a bit unclear on this. You changed the recvSCPEnvelope path to return ENVELOPE_STATUS_SKIPPED_SELF in the previous change, and I would have thought that SKIPPED_SELF should also imply _not_ flooding the message further, but in this condition it looks like a SKIPPED_SELF message _will_ get flooded further. Can you explain a bit what is being inhibited here, and why it's correct?

src/overlay/Peer.cpp
```diff
@@ -1020,10 +1020,8 @@ Peer::recvAuth(StellarMessage const& msg)
         return;
     }
 
-    // send SCP State
-    // remove when all known peers implements the next line
```
graydon: Is it currently the case that this comment's "when" condition has been achieved? Which protocol / release version included that change?

src/overlay/Floodgate.h
```diff
@@ -38,7 +38,7 @@ class Floodgate
 
         uint32_t mLedgerSeq;
         StellarMessage mMessage;
-        std::set<Peer::pointer> mPeersTold;
+        std::set<std::string> mPeersTold;
```
graydon: If this is about restricting unnecessary retention, I think a weak_ptr ought to do. Bonus: you can clear the expired entries out of each flood record as you go.

src/herder/HerderImpl.cpp
```diff
@@ -416,6 +410,12 @@ HerderImpl::recvSCPEnvelope(SCPEnvelope const& envelope)
         return Herder::ENVELOPE_STATUS_DISCARDED;
     }
 
+    if (envelope.statement.nodeID == getSCP().getLocalNode()->getNodeID())
+    {
+        CLOG(DEBUG, "Herder") << "recvSCPEnvelope: skipping own message";
+        return Herder::ENVELOPE_STATUS_SKIPPED_SELF;
```
graydon: Why did this condition _move_ from higher up to down here, in addition to changing return code?

src/overlay/Peer.cpp
MonsieurNicolas: SKIPPED_SELF just means that it was an SCP message created by the local node, so we do want to flood it; the meaning of ENVELOPE_STATUS_DISCARDED now is that it's a message that is discarded because it's not in the right range of ledgers regardless of where it came from.

src/overlay/Peer.cpp
MonsieurNicolas: this was added in 5f395e7bfd577659496c5a4a10dcfbb1e971879f (October 2017), so a long time ago

src/overlay/Floodgate.h
MonsieurNicolas: I thought about this but picked a unique ID instead:
* unique ID persist between re-connections, so we "remember" to not flood specific messages even if peers disconnect/reconnect (when using ptrs, we would not be able to do that)
* floodmap entries expire after a few ledgers, so I didn't think that we would clean those up more than what we do now

src/herder/HerderImpl.cpp
MonsieurNicolas: Like I mentioned in a previous comment (and in the commit message), this allows to never flood SCP messages that are not in the right range - without this change, we can't tell if the message is old or not, which is the only criteria we care about for flooding.

graydon: Ok, I think I understand now: previously both own-message and out-of-range cases resulted in early-return and continue-to-flood. The change here is to _keep_ own-message in that case (early-return and continue-to-flood) but to sharpen the out-of-range case to result in early-return and inhibit-flooding, which means that range test has to happen earlier as well as return a different code than the own-message case.

Phew! Thanks for the explanation. Approved!

graydon:
r+ c5f4ce2

MonsieurNicolas:
@latobarita: retry

PR #1956, opened 2019-02-11
allow VirtualClock::crank to process more of asio queue at a time

vogel:

# Description

Try to fix big-quorum issues with changes VirtualClock::crank semantics.

This version allows to process as many items from io_service queue as possible as long as there is no timer event that should be processed or a delayed work item. As we control all-but one timers in code (the only one that is outside of our control is mRealTimer that is used only for executing timers controlled by us by using advanceToNow() calls), we can safely do that without worrying that timers will be starved.

This code also has safeguard to ensure that at least 100 queue items will be executed if queue contains more that 100 items, so from io_service perspective it is at least as good as old code.

This code can have problems when:
- io_service queue is big (which can happen now)
- there are a lot of timers firing quickly one after another (I have not noticed anything like that, we use very small amount of timers)
- there are a lot of delayed (or even non-delayed) work items posted to main thread (as before, it does not seem we do that now)

# Checklist
- [x] Compiles
- [x] Ran all tests

### DIFF
--- a/src/util/Timer.cpp
+++ b/src/util/Timer.cpp
@@ -278,7 +278,7 @@ VirtualClock::crank(bool block)
         // Pick up some work off the IO queue.
         // Calling mIOService.poll() here may introduce unbounded delays
         // to trigger timers.
-        const size_t WORK_BATCH_SIZE = 100;
+        const size_t WORK_CHECK_INTERVAL = 100;
         size_t lastPoll;
         size_t i = 0;
         do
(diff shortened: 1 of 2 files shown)


MonsieurNicolas:
Which issue is this fixing? Can you expand a bit? Is it that right now we process too many items from the delayed queue at once, something else?

vogel:
This fixed issue with non processing enough network traffic when we are doing another heavy work (like applying buckets when we have too big quorum set). It seems more like a band-aid than a real fix, but it makes things more fair for the network side of code without sacrificing timers.

MonsieurNicolas:
How is this really helping processing more network traffic?

Functionality wise, is it really different? Looking at how `crank` works,

master does:
1. in a loop, call `crank`:
    1. process timers
    2. process 100 events from IOService queue
    3. post delayed queue into main IOService

this PR does:
1. in a loop, call `crank`:
    1. process timers
    2. if no expired timers and nothing in delayed queue
        a. process 100 events from IOService
    3. post delayed queue into main IOService

Because we have this outer loop calling `crank`, it doesn't seem that they are actually behaving that differently (or at all).
The only difference seems to be that
   * `crank` blocks longer as it can process more than 100 items in one go
   * when there are a lot of things to pick up from IOService, we will first queue delayed work before processing timers, where master may process timers before queuing delayed work

PR #1957, opened 2019-02-11
Optimize load best offers

jonjove:
This PR simplifies the schema for the `offers` table and adds the index appropriate for the SQL queries used to load best offers. It also fixes an issue causing unnecessary cache misses when loading best offers. I also added a performance benchmark, which produced results:

Benchmark Type | create (ms) | write (ms) | load (ms)
----------------|-------------|------------|-----------
postgres - old | 1494.47 +/- 4.33 | 2733.39 +/- 20.05 | 14367.67 +/- 491.10
postgres - new | 1429.79 +/- 20.99 | 2546.09 +/- 23.84 | 2458.40 +/- 3.69
sqlite - old | 306.87 +/- 5.28 | 789.33 +/- 4.30 | 32460.91 +/- 283.26
sqlite - new | 275.67 +/- 5.89 | 823.34 +/- 6.42 | 1624.85 +/- 9.12

### DIFF
--- a/src/ledger/LedgerTxnOfferSQL.cpp
+++ b/src/ledger/LedgerTxnOfferSQL.cpp
@@ -20,6 +20,11 @@ std::shared_ptr<LedgerEntry const>
 LedgerTxnRoot::Impl::loadOffer(LedgerKey const& key) const
 {
     uint64_t offerID = key.offer().offerID;
+    if (offerID > INT64_MAX)
+    {
+        return nullptr;
+    }
+
     std::string actIDStrKey = KeyUtils::toStrKey(key.offer().sellerID);
 
     std::string sql = "SELECT sellerid, offerid, sellingasset, buyingasset, "
@@ -645,8 +650,11 @@ class BulkLoadOffersOperation
         for (auto const& k : keys)
         {
             assert(k.type() == OFFER);
-            mSellerIDs.emplace_back(k.offer().sellerID);
-            mOfferIDs.emplace_back(unsignedToSigned(k.offer().offerID));
+            if (k.offer().offerID <= INT64_MAX)
+            {
+                mSellerIDs.emplace_back(k.offer().sellerID);
+                mOfferIDs.emplace_back(unsignedToSigned(k.offer().offerID));
+            }
         }
 
         for (size_t i = 0; i < mSellerIDs.size(); ++i)
--- a/src/ledger/LedgerTxnOfferSQL.cpp
+++ b/src/ledger/LedgerTxnOfferSQL.cpp
@@ -7,74 +7,14 @@
 #include "database/Database.h"
 #include "database/DatabaseTypeSpecificOperation.h"
 #include "ledger/LedgerTxnImpl.h"
+#include "util/Decoder.h"
+#include "util/Logging.h"
 #include "util/XDROperators.h"
 #include "util/types.h"
+#include "xdrpp/marshal.h"
 
 namespace stellar
 {
-
-void
-getAssetStrings(Asset const& asset, std::string& assetCodeStr,
-                std::string& issuerStr, soci::indicator& assetCodeIndicator,
-                soci::indicator& issuerIndicator)
-{
-    if (asset.type() == ASSET_TYPE_CREDIT_ALPHANUM4)
-    {
-        assetCodeToStr(asset.alphaNum4().assetCode, assetCodeStr);
-        issuerStr = KeyUtils::toStrKey(asset.alphaNum4().issuer);
-        assetCodeIndicator = soci::i_ok;
-        issuerIndicator = soci::i_ok;
-    }
-    else if (asset.type() == ASSET_TYPE_CREDIT_ALPHANUM12)
-    {
-        assetCodeToStr(asset.alphaNum12().assetCode, assetCodeStr);
-        issuerStr = KeyUtils::toStrKey(asset.alphaNum12().issuer);
-        assetCodeIndicator = soci::i_ok;
-        issuerIndicator = soci::i_ok;
-    }
-    else
-    {
-        assert(asset.type() == ASSET_TYPE_NATIVE);
-        assetCodeStr = "";
-        issuerStr = "";
-        assetCodeIndicator = soci::i_null;
-        issuerIndicator = soci::i_null;
-    }
-}
-
-void
-processAsset(Asset& asset, AssetType assetType, std::string const& issuerStr,
-             soci::indicator const& issuerIndicator,
-             std::string const& assetCode,
-             soci::indicator const& assetCodeIndicator)
-{
-    asset.type(assetType);
-    if (assetType != ASSET_TYPE_NATIVE)
-    {
-        if ((assetCodeIndicator != soci::i_ok) ||
-            (issuerIndicator != soci::i_ok))
-        {
-            throw std::runtime_error("bad database state");
-        }
-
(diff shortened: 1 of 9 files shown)


vogel:
Could we do similar trick with accounts? Store accounts as xdr, accountid, balance, inflationdest? Would that save us some cpu time?

vogel:
Edit: it may be even worth trying splitting this into two tables: {accountid, xdr} for one and {accountid, balance, inflactiondest} for inflation calculations.

MonsieurNicolas:
Sadly, I don't think we can make this change yet: Horizon (and I imagine others), are querying the core database directly in the context of path finding.

The only thing they can do without rewriting their code would be to load all offers in memory (which may not be a bad idea given the number of offers). @nikhilsaraf any opinion on this one, I think I heard that you guys were thinking about doing it in memory.

src/ledger/LedgerTxnOfferSQL.cpp
```diff
@@ -895,4 +675,131 @@ LedgerTxnRoot::Impl::bulkLoadOffers(
     return populateLoadedEntries(keys,
                                  mDatabase.doDatabaseTypeSpecificOperation(op));
 }
+
+static void
+processAsset(Asset& asset, AssetType assetType, std::string const& issuerStr,
```
graydon: It might be nice to name this and the loadAllOffers helper below in a way that indicates they only exist for the sake of migration.

src/ledger/LedgerTxnOfferSQL.cpp
```diff
@@ -895,4 +675,131 @@ LedgerTxnRoot::Impl::bulkLoadOffers(
...
+           "buyingasset      TEXT             NOT NULL,"
+           "amount           BIGINT           NOT NULL CHECK (amount >= 0),"
+           "pricen           INT              NOT NULL,"
+           "priced           INT              NOT NULL,"
+           "price            DOUBLE PRECISION NOT NULL,"
+           "flags            INT              NOT NULL,"
+           "lastmodified     INT              NOT NULL,"
+           "PRIMARY KEY      (offerid)"
+           ");";
+    mDatabase.getSession() << "CREATE INDEX bestofferindex ON offers "
```
graydon: This is a different set of indexes than the previous table and I'm slightly concerned that it will either fail to support the current or future set of queries; in particular AllowTrustOp seems like it does point queries that will want an index on each of buyingasset and sellingasset (it had them before on the issuers-component of each); and I was unclear on whether horizon wanted to do point queries on the table by asset or account+asset too.

(Also: for indexes that are supporting point queries, we should probably use hash indexes rather than the default btrees)

MonsieurNicolas:
hold onto merging this until we hear from the platform team's timeline

src/ledger/LedgerTxnOfferSQL.cpp
MonsieurNicolas: Agree with this, I scratched my head a few times before realizing that this was migration code

src/ledger/LedgerTxnOfferSQL.cpp
MonsieurNicolas: Why include `offerID` in that index? `offerID` is just a tie breaker, and the assumption is that prices tend to be different. I would add this one later if we need to.

What about the index on `buyingasset`? without it, `loadOffersByAccountAndAsset` will perform a full scan on the entire order book.

As for future uses, I am not too worried about that: I'd rather have a schema that reflects are current patterns.

For Horizon, this preserves indexes for path finding (that looks at selling or buying assets). Other queries are already bad in Horizon as they query on `sellerID` afaik.

jonjove: I'm not sure that assumption is right. A quick query of the database (as of ledger approximately 20 million) suggests there are

- 1312 offers where `offerID` would be needed to differentiate between 2
- 459 offers where `offerID` would be needed to differentiate between 3
- 220 offers where `offerID` would be needed to differentiate between 4
- 1667 offers where `offerID` would be needed to differentiate between 5 or more

and a specific example is 880 offers with the same `(sellingasset, buyingasset, price)`. All together, 31.7% of offers have the same `(sellingasset, buyingasset, price)` as at least one other offer.

With regard to `loadOffersByAccountAndAsset`, I'm not even sure if an index on `buyingasset` would be sufficient. We'd probably want to have an index on `(sellerid, sellingasset)` and an index on `(sellerid, buyingasset)` if we actually wanted the query to be well-optimized. I will look tomorrow into how that query performs with different kinds of index and reply with the results.

MonsieurNicolas: any update on this?

I would drop `offerID` for now
* we don't have this index right now
* 99% of the database has less than 5 offers at any given price, so it doesn't seem that speeding up the 1% is worth it at this time (while slowing down the 99%), unless you have data that shows that it makes any difference (like, replay ranges of legers instead of synthetic tests)

re: `loadOffersByAccountAndAsset` is probably a similar situation where 99% of the users have very few offers per trading pair so a small scan is probably fine

jonjove: I have run some artificial benchmarks, and also a small benchmark that catches up at ledger 22.5 million then replays 5000 ledgers on sqlite.

With regard to whether the index should contain `offerID`, it was faster in the benchmarks. But the actual history showed it was about 25% slower for bucket apply and effectively unchanged for replay. I think we should exclude `offerID` until the data in the ledger necessitates inclusion.

With regard to `loadOffersByAccountAndAsset` I tried adding various potentially useful indexes. In artificial benchmarks they do make this query a bit faster but at the cost of making writes much slower. In actual history, all indexes I tried made both bucket apply and replay slower. It's possible, perhaps even likely, that this range of ledgers did not include that query even once. But given the observations, I also think we should not add additional indexes for this purpose until we have concrete data supporting the effectiveness.

In summary, my proposal is remove `offerid` from `bestofferindex` and add no other indexes. @graydon @MonsieurNicolas if you are okay with this, I will update the PR accordingly.

bartekn:
I think we have a PR in Horizon ready: https://github.com/stellar/go/pull/972 but I couldn't test it with core instance because the DB wasn't migrated automatically (it worked before for signers change). Am I missing something or the code is not there yet?

jonjove:
@bartekn Did you try to do the migration from a database produced by v10.2.0 or from something more recent? Because all of this is going into the same stellar-core release, the schema change is part of the same upgrade as for signers. So if you had already upgraded to schema version 9 using stellar-core built from an earlier commit then it won't do this upgrade since it thinks it has already done it.

bartekn:
I'm running a code from this branch.
```
core=# select * from storestate where statename = 'databaseschema';
            statename             | state 
----------------------------------+-------
 databaseschema                   | 8

```

bartekn:
Ah, sorry, I realized I was connecting to a DB copy. 🤦‍♂️

marta-lokhova:
I think `BulkLoadOffersOperation::doSqliteSpecificOperation` and `BulkLoadOffersOperation:: doPostgresSpecificOperation` need to reflect the schema update also.

MonsieurNicolas:
r+ 038273877a90c943881b168e0a5f04e8a1938370

MonsieurNicolas:
@latobarita: retry

MonsieurNicolas:
r+ d5f89967f0f559f6c4b96af9d0fbe42ba346266c

PR #1958, opened 2019-02-13
Improve bucket error handling

### LINKED ISSUE #1952: crash writing buckets when hitting I/O error
It looks like we don't have proper error handing when there is an error writing buckets:
right now we're triggering an `assert` even though this is a fatal (but not a bug) condition.

We should:
* check for error conditions when performing writes, so that we fail closer to the actual failure (right now fails right before the `close` call)
* check for other potential errors, like are `open` or `close` throwing?
* throw with detail on why it's failing "out of disk space"

marta-lokhova:
Resolves #1952

### DIFF
--- a/src/bucket/BucketOutputIterator.cpp
+++ b/src/bucket/BucketOutputIterator.cpp
@@ -41,6 +41,7 @@ BucketOutputIterator::BucketOutputIterator(std::string const& tmpDir,
 {
     CLOG(TRACE, "Bucket") << "BucketOutputIterator opening file to write: "
                           << mFilename;
+    // Will throw if unable to open the file
     mOut.open(mFilename);
 }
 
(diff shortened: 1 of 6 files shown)


src/bucket/BucketOutputIterator.cpp
```diff
@@ -79,10 +80,9 @@ BucketOutputIterator::put(BucketEntry const& e)
 std::shared_ptr<Bucket>
 BucketOutputIterator::getBucket(BucketManager& bucketManager)
 {
-    assert(mOut);
```
marta-lokhova: Btw, I removed this assert, because I don't think we can actually guarantee this. The stream might not be flushed causing the call to `writeOne` succeed, while the actual disk write later fails. We should still be able to detect problems, since `writeAndThrowOnFail` and `close` will throw.

MonsieurNicolas:
actually, there might be a better way @marta-lokhova :
it looks like you can cause all failed operations to throw (by default, nothing throws, including `close`)
see http://www.cplusplus.com/reference/ios/ios/exceptions/

We may want to do something similar for reading instead of manually checking for failures (which nobody does).

src/bucket/BucketOutputIterator.cpp
MonsieurNicolas: `close` does not throw by default

marta-lokhova:
This is updated, btw

graydon:
r+ 3c89002

(nice!)

PR #1959, opened 2019-02-13
[HerderTests] Restart app after removal from simulation

marta-lokhova:
Fixing the test that is currently failing on Macs. Turns out the node removed from test simulation was not properly restarted, causing it to lose track of SCP.

I also double checked similar cases where we remove and re-add the node, and they all seem to be handled correctly.

### DIFF
--- a/src/herder/HerderTests.cpp
+++ b/src/herder/HerderTests.cpp
@@ -1073,7 +1073,8 @@ TEST_CASE("In quorum filtering", "[herder]")
         {
             qSetK[i].validators.emplace_back(extraK[1].getPublicKey());
         }
-        sim->addNode(extraK[i], qSetK[i]);
+        auto node = sim->addNode(extraK[i], qSetK[i]);
+        node->start();
         sim->addConnection(extraK[i].getPublicKey(), nodeIDs[0]);
     }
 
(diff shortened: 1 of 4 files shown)


src/herder/HerderTests.cpp
```diff
@@ -1123,7 +1123,9 @@ TEST_CASE("In quorum filtering", "[herder]")
     node3Config.QUORUM_SET.validators.emplace_back(extraK[2].getPublicKey());
     node3Config.QUORUM_SET.validators.emplace_back(extraK[3].getPublicKey());
 
-    sim->addNode(node3Config.NODE_SEED, node3Config.QUORUM_SET, &node3Config);
+    auto node3 = sim->addNode(node3Config.NODE_SEED, node3Config.QUORUM_SET,
+                              &node3Config);
+    node3->start();
```
MonsieurNicolas: mmm, looks like we're missing the call to `start` in many other places?
I imagine that after calling `addNode` we should either call `startAllNodes` or call `start` manually?

Maybe add an assert in the simulation class to catch non started applications (not sure how to check for that though)? I think tests kinda work here by accident.

marta-lokhova: From what I see, other test cases call either `start` or `startAllNodes`, except this one, "In quorum filtering" (Not counting the nodes generated in `Topologies.cpp`, since it doesn't really make sense to start them there)

Also, just noticed that some nodes added in earlier in this test are also not started, will update. Regarding the asserts, maybe we should throw in `Simulation::crankNode` (which is used by other cranking functions) if the app is not started.

MonsieurNicolas: `crankNode` sounds like a perfect place to check for that yes, should be easy to find the bad ones with the assert :)

marta-lokhova: ok, updated. We were not even using initial application state, so I updated `Application::getState`

graydon:
r+ 9216aa7

PR #1961, opened 2019-02-14
Drop slow peers

vogel:
# Description

This PR allows dropping peer that are to slow to accept messages. If sending a message to peer occurs after more than io_timeout seconds from pushing it to send queue, peer is dropped.

# Checklist
- [x] Compiles
- [x] Ran all tests


MonsieurNicolas:
keeping the simpler change for now - we may have to use this approach if it's not enough

PR #1963, opened 2019-02-14
overlay: drop peers that are not keeping up.

graydon:
# Description

Relative of / simpler approach to #1961 

This adds a third timestamp to each peer indicating the last time the peer's write queue was empty. This is checked at the same boundary (~30 seconds) as the idle timer. Peers that are keeping up will regularly bounce off empty-queue state: asio issues async_writes to the TCP layer and immediately fires their completion handlers until the local TCP buffers fill up, and those are typically hundreds of KB. Effectively this means that any peer that's not able to accept the traffic we're sending them gets disconnected rather than using up more memory on the sending side's write queue.

# Checklist
- [x] Compiles
- [x] Ran all tests
- [x] N/A (aside from memory usage, which it reduces) If change impacts performance, include supporting evidence per the [performance document](https://github.com/stellar/stellar-core/blob/master/performance-eval/performance-eval.md)

### DIFF
--- a/src/overlay/Peer.cpp
+++ b/src/overlay/Peer.cpp
@@ -60,6 +60,7 @@ Peer::Peer(Application& app, PeerRole role)
     , mIdleTimer(app)
     , mLastRead(app.getClock().now())
     , mLastWrite(app.getClock().now())
+    , mLastEmpty(app.getClock().now())
 
     , mMessageRead(
           app.getMetrics().NewMeter({"overlay", "message", "read"}, "message"))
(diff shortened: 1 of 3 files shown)


MonsieurNicolas:
r+ f4250142fce723876baf3a95963b727c7b0eb202

MonsieurNicolas:
@latobarita: retry

MonsieurNicolas:
flaky SCP flooding test over TCP

MonsieurNicolas:
@latobarita: retry

PR #1966, opened 2019-02-16
Scp messages flooding fix

### LINKED ISSUE #1489: flaky test: "SCP messages flooding"

MonsieurNicolas:
# Description

Resolves #1489 

This PR fixes a couple issues:
* the change to only request current and future SCP messages during handshake is a bit too aggressive, while it's sufficient for the local node, it may not be enough for other nodes. This can be a problem when many connections die at the same time.
* SCP flooding tests were mixing two cases, that I separated:
   * what happens when flooding messages and some validators externalize (we need to continue flooding messages with values matching consensus) - I will open a different PR for this
   * what happens when flooding many values in a given ledger - what this PR does

# Checklist
- [ ] Compiles
- [ ] Ran all tests

### DIFF
--- a/src/overlay/FloodTests.cpp
+++ b/src/overlay/FloodTests.cpp
@@ -207,9 +207,12 @@ TEST_CASE("Flooding", "[flood][overlay]")
         // a valid transaction set
 
         std::vector<SecretKey> keys;
+        std::unordered_map<PublicKey, SecretKey> keysMap;
         for (int i = 0; i < nbTx; i++)
         {
             keys.emplace_back(SecretKey::random());
+            auto& k = keys.back();
+            keysMap.insert(std::make_pair(k.getPublicKey(), k));
         }
 
         auto injectSCP = [&](int i) {
@@ -256,7 +259,6 @@ TEST_CASE("Flooding", "[flood][overlay]")
             prep.ballot.counter = 1;
             prep.quorumSetHash = qSetHash;
 
-            // use the sources to sign the message
             st.nodeID = keys[i].getPublicKey();
             envelope.signature = keys[i].sign(xdr::xdr_to_opaque(
                 inApp->getNetworkID(), ENVELOPE_TYPE_SCP, st));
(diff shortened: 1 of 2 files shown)


graydon:
Looks .. plausible to me, unsure.

@MonsieurNicolas concerning the Peer change, I'm a bit worried that is going to cause a 12x bigger temporary traffic spike on initial connection; are we sure this is safe / desirable?

Concerning the flood change, it looks to me like the only logical difference is the change from .666 to 1.0 in the proportion of peers required to form a quorum; the other edits look like you're doing the same thing as before just using slightly different phrasing. Am I missing something? If it is just .666 -> 1.0, why is that a relevant / important change?

MonsieurNicolas:
@graydon : the change 0.666 -> 1.0 is for clarity, the real change is in quorum adjuster that sets the threshold to 100% , reason for this is that the "outer nodes" tests (that were failing) has the default for the "inner core" hardcoded inside simulation. By setting the threshold to 100% we ensure that the network doesn't close a ledger. When closing, validators stop forwarding messages on old slots that don't agree with the consensus value (and this test has a different value for each node).

Regarding the peer change (feel free to split up this PR): a smaller fix would be to use a smaller number instead of `Herder::MAX_SLOTS_TO_REMEMBER` to at least ensure that nodes can track towards "current slot" with higher probability, so we need to properly propagate those.
I suspect that the change that we did in https://github.com/stellar/stellar-core/pull/1955/commits/083c54c377e891372936bf892c4c0c15c3ebe662 is too aggressive given the way overlay works right now and the expectations that SCP messages eventually get flooded to all peers on the network. I guess we could bundle that change with the "force disconnect" change as it increases the changes of bad things to happen even more (especially in situations where the network hits a global snag where various nodes are going to process certain ledgers for much longer than expected).
Real fix would be to have a way for peers to ask more granular questions to the network as a whole (like "send me SCP messages for ledger X relevant to me" (and have a way to find a peer with this data), but that's a deeper change to overlay.

graydon:
ok, makes sense.

graydon:
r+ cf3627d

PR #1970, opened 2019-02-20
PostgreSQL 9.5 and database version checking in general

graydon:
# Description

Recent changes for bulk-writing have required UPSERT functionality in postgresql, which is only present in 9.5 and later. As it happens we moved travis to test against 9.5 a while ago, we just haven't updated everything to reflect that requirement yet. This PR does more of that sort of change: switching the MSVC path to use 9.5 and also wiring in a version check when we open a connection, so we get a nice error message.

I also added a SQLite version check, which is a little bit overkill since we're currently bundling our own in all cases, but it seems like a nice enough thing to double check / trap with a friendly error message if we can.

(I believe that being on version 9.4 was what caused the recent acceptance tests to fail; I've updated the docker image in the meantime)

# Checklist
- [x] Compiles
- [x] Ran all tests
- [x] (N/A) If change impacts performance, include supporting evidence per the [performance document](https://github.com/stellar/stellar-core/blob/master/performance-eval/performance-eval.md)

### DIFF
--- a/Builds/VisualStudio/libsoci_postgresql/libsoci_postgresql.vcxproj
+++ b/Builds/VisualStudio/libsoci_postgresql/libsoci_postgresql.vcxproj
@@ -119,7 +119,7 @@
       <WarningLevel>Level3</WarningLevel>
       <Optimization>Disabled</Optimization>
       <SDLCheck>true</SDLCheck>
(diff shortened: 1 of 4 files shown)


jonjove:
r+ 45fe5b6

PR #1315, opened 2017-09-19
Catchup at

vogel:
Changes way that catchup is done (from recent/minimal/complete to catchup-count-ledgers-to-ledger, with auto detection if applying buckets or transaction is needed), then adds few simple functions to main.cpp.

PR #1972, opened 2019-02-21
Fix crash in apply buffered ledgers

### LINKED ISSUE #1965: crash while replaying buffered ledgers
# Issue type
bug

# Version found
10.2.0

# Issue description
watcher node following the network

## Actual result
Validator crashes while catching up

## Expected result
No crash

## Supporting data

Crash on `assert`:
https://github.com/stellar/stellar-core/blob/54504c714ab6e696283e0bd0fdf1c3a029b7c88b/src/ledger/LedgerManagerImpl.cpp#L710

```
2019-02-15T16:42:15.444 GB34S [History INFO] Catchup downloading history archive state for applying buckets at checkpoint 22497663 not needed
2019-02-15T16:42:15.444 GB34S [Ledger INFO] Caught up to LCL from history: [seq=22497668, hash=437419]
2019-02-15T16:42:15.444 GB34S [Work INFO] WorkManager got SUCCESS from catchup
2019-02-15T16:42:15.499 GB34S [Ledger INFO] Replaying buffered ledger-close: [seq=22497669, prev=437419, tx_count=11, sv: [  txH: 81ed73, ct: 1550248575, upgrades: [ ] ]]
2019-02-15T16:42:15.519 GB34S [Tx INFO] applying ledger 22497669 (txs:11, ops:31)
2019-02-15T16:42:15.705 GB34S [Herder WARNING] Lost track of consensus
...
2019-02-15T16:42:15.711 GB34S [Herder INFO] Quorum information for 22497731 : {"agree":3,"delayed":0,"disagree":0,"fail_at":2,"hash":"d60582","missing":0,"phase":"EXTERNALIZE"}

2019-02-15T16:42:15.721 GB34S [Ledger INFO] Got consensus: [seq=22497733, prev=768cfe, tx_count=10, sv: [  txH: e64642, ct: 1550248901, upgrades: [ ] ]]
2019-02-15T16:42:15.721 GB34S [Ledger INFO] Close of ledger 22497733 buffered
2019-02-15T16:42:15.772 GB34S [Ledger INFO] Got consensus: [seq=22497734, prev=396ffe, tx_count=9, sv: [  txH: 699c50, ct: 1550248906, upgrades: [ ] ]]
2019-02-15T16:42:15.772 GB34S [Ledger INFO] Close of ledger 22497734 buffered
```

vogel:
# Description

Resolves #1965

Fixes race condition in applyBufferedLedgers. With new switching to WAITING_FOR_CLOSING_LEDGER is done before posting applying code to next crank, so it is no longer possible for mSyncingLedger to become empty and then non-empty again without entering that state.

Also a new condition and explicit reset method is added to SyncingLedgerChain itself to prevent such bugs in future.

# Checklist
- [x] Compiles
- [x] Ran all tests

### DIFF
--- a/src/ledger/LedgerManagerImpl.cpp
+++ b/src/ledger/LedgerManagerImpl.cpp
@@ -694,19 +694,18 @@ void
 LedgerManagerImpl::applyBufferedLedgers()
 {
     assert(mCatchupState == CatchupState::APPLYING_BUFFERED_LEDGERS);
+    if (mSyncingLedgers.empty())
+    {
+        CLOG(INFO, "Ledger")
+            << "Caught up to LCL including recent network activity: "
+            << ledgerAbbrev(mLastClosedLedger)
+            << "; waiting for closing ledger";
+        setCatchupState(CatchupState::WAITING_FOR_CLOSING_LEDGER);
+        return;
+    }
 
     mApp.postOnMainThreadWithDelay(
         [&] {
-            if (mSyncingLedgers.empty())
-            {
-                CLOG(INFO, "Ledger")
-                    << "Caught up to LCL including recent network activity: "
-                    << ledgerAbbrev(mLastClosedLedger)
-                    << "; waiting for closing ledger";
-                setCatchupState(CatchupState::WAITING_FOR_CLOSING_LEDGER);
-                return;
-            }
-
             auto lcd = mSyncingLedgers.front();
             mSyncingLedgers.pop();
             mSyncingLedgersSize.set_count(mSyncingLedgers.size());
--- a/src/ledger/LedgerManagerImpl.cpp
+++ b/src/ledger/LedgerManagerImpl.cpp
@@ -511,6 +511,7 @@ LedgerManagerImpl::finalizeCatchup(LedgerCloseData const& ledgerData)
     assert(mState == LM_CATCHING_UP_STATE);
     assert(mCatchupState == CatchupState::WAITING_FOR_CLOSING_LEDGER);
     assert(mSyncingLedgers.empty());
+    mSyncingLedgers.reset();
 
     switch (closeLedgerIf(ledgerData))
     {
@@ -623,9 +624,23 @@ LedgerManagerImpl::startCatchup(CatchupConfiguration configuration,
         throw std::invalid_argument("Target ledger is not newer than LCL");
     }
 
-    setCatchupState(CatchupState::APPLYING_HISTORY);
-    assert(manualCatchup == mSyncingLedgers.empty());
+    if (manualCatchup)
+    {
+        if (getState() == LM_CATCHING_UP_STATE &&
+            mCatchupState != CatchupState::WAITING_FOR_CLOSING_LEDGER)
+        {
+            throw std::runtime_error(
+                "Unable to perform second catchup before first one finishes");
+        }
+        assert(mSyncingLedgers.empty());
+        mSyncingLedgers.reset();
+    }
+    else
+    {
+        assert(!mSyncingLedgers.empty());
+    }
 
+    setCatchupState(CatchupState::APPLYING_HISTORY);
     mApp.getCatchupManager().catchupHistory(
         configuration,
         std::bind(&LedgerManagerImpl::historyCaughtup, this, _1, _2, _3));
--- a/src/ledger/LedgerManagerImpl.cpp
+++ b/src/ledger/LedgerManagerImpl.cpp
@@ -536,6 +536,10 @@ LedgerManagerImpl::finalizeCatchup(LedgerCloseData const& ledgerData)
 void
 LedgerManagerImpl::addToSyncingLedgers(LedgerCloseData const& ledgerData)
 {
+    assert(mState == LM_CATCHING_UP_STATE);
+    assert(mCatchupState != CatchupState::NONE);
+    assert(mCatchupState != CatchupState::WAITING_FOR_CLOSING_LEDGER);
+
     switch (mSyncingLedgers.push(ledgerData))
     {
     case SyncingLedgerChainAddResult::CONTIGUOUS:
--- a/src/ledger/LedgerTests.cpp
+++ b/src/ledger/LedgerTests.cpp
@@ -2,22 +2,12 @@
 // under the Apache License, Version 2.0. See the COPYING file at the root
 // of this distribution or at http://www.apache.org/licenses/LICENSE-2.0
 
-#include "LedgerTestUtils.h"
-#include "database/Database.h"
-#include "herder/LedgerCloseData.h"
 #include "ledger/LedgerManager.h"
 #include "ledger/LedgerTxn.h"
-#include "ledger/LedgerTxnEntry.h"
-#include "ledger/LedgerTxnHeader.h"
-#include "lib/catch.hpp"
 #include "main/Application.h"
-#include "main/Config.h"
-#include "test/TestUtils.h"
 #include "test/test.h"
-#include "util/Logging.h"
-#include "util/Timer.h"
-#include "util/types.h"
-#include <xdrpp/autocheck.h>
+
+#include <lib/catch.hpp>
 
 using namespace stellar;
 
--- a/src/ledger/SyncingLedgerChain.cpp
+++ b/src/ledger/SyncingLedgerChain.cpp
@@ -33,19 +33,26 @@ SyncingLedgerChain::pop()
 SyncingLedgerChainAddResult
 SyncingLedgerChain::push(LedgerCloseData lcd)
 {
-    if (mChain.empty() ||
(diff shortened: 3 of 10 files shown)


graydon:
This looks great! I'm out of time for review today but will try to look at it tomorrow.

graydon:
Notes for review: this fixed an issue introduced in https://github.com/stellar/stellar-core/pull/1818

Also requires reading some backstory for those of us who have not been following the evolution of the catchup state machine in recent times: 

  - https://github.com/stellar/stellar-core/pull/1315
  - https://github.com/stellar/stellar-core/pull/1627

src/ledger/LedgerManagerImpl.cpp
```diff
@@ -619,9 +624,23 @@ LedgerManagerImpl::startCatchup(CatchupConfiguration configuration,
...
+    if (manualCatchup)
+    {
+        if (getState() == LM_CATCHING_UP_STATE &&
+            mCatchupState != CatchupState::WAITING_FOR_CLOSING_LEDGER)
+        {
+            throw std::runtime_error(
+                "Unable to perform second catchup before first one finishes");
+        }
+        assert(mSyncingLedgers.empty());
+        mSyncingLedgers.reset();
```
marta-lokhova: Is this the right place to reset `mSyncingLedgers`? I think it should be done when catchup is finished (which already happens in `finalizeCatchup`).

src/ledger/LedgerManagerImpl.cpp
```diff
@@ -619,9 +624,23 @@ LedgerManagerImpl::startCatchup(CatchupConfiguration configuration,
         throw std::invalid_argument("Target ledger is not newer than LCL");
     }
 
-    setCatchupState(CatchupState::APPLYING_HISTORY);
-    assert(manualCatchup == mSyncingLedgers.empty());
+    if (manualCatchup)
+    {
+        if (getState() == LM_CATCHING_UP_STATE &&
```
marta-lokhova: Will this if statement ever evaluate to `true`? It looks like command line already handles any catchup in progress: https://github.com/stellar/stellar-core/blob/master/src/main/CommandHandler.cpp#L554

src/ledger/SyncingLedgerChain.h
```diff
@@ -31,11 +32,13 @@ class SyncingLedgerChain final
     LedgerCloseData const& back() const;
     void pop();
     SyncingLedgerChainAddResult push(LedgerCloseData lcd);
+    void reset();
```
marta-lokhova: I'd love to see a comment here explaining when `reset` should be called, and why it should be explicitly called instead of just relying on the emptiness of `SyncingLedgerChain`.

src/ledger/LedgerTests.cpp
```diff
@@ -2,22 +2,12 @@
...
 #include "main/Application.h"
-#include "main/Config.h"
-#include "test/TestUtils.h"
 #include "test/test.h"
-#include "util/Logging.h"
-#include "util/Timer.h"
-#include "util/types.h"
-#include <xdrpp/autocheck.h>
+
+#include <lib/catch.hpp>
```
MonsieurNicolas: wow so much cruft! thanks

src/ledger/LedgerManagerImpl.cpp
marta-lokhova: Also, wrt asserts, with the new explicit reset, I don't think checking `mSyncingLedgers.empty()` is enough anymore , so you might want to make it stricter and check `mLastLedgerSeq` as well (maybe by introducing a new method `bool SyncingLedgerChain::isReset()`)

src/ledger/LedgerManagerImpl.cpp
```diff
@@ -536,6 +536,10 @@ LedgerManagerImpl::finalizeCatchup(LedgerCloseData const& ledgerData)
 void
 LedgerManagerImpl::addToSyncingLedgers(LedgerCloseData const& ledgerData)
 {
+    assert(mState == LM_CATCHING_UP_STATE);
```
MonsieurNicolas: not a fan of `asserts` like this:
* as mentioned in the issue, asserts without logs are a pain to debug
* `addToSyncingLedgers` is a helper function, ideally it should not have a dependency on the state machine as it makes the code harder to reason about (if there is a dependency - the `SyncingLedgerChain` abstraction should be improved)
* can't we just push responsibility of performing the right checks upstream in the call stack?
* if we want it to care about state, it should whitelist instead of blacklist

src/ledger/SyncingLedgerChain.cpp
```diff
@@ -33,21 +33,28 @@ SyncingLedgerChain::pop()
...
         return SyncingLedgerChainAddResult::TOO_OLD;
     }
 
     return SyncingLedgerChainAddResult::TOO_NEW;
 }
 
+void
+SyncingLedgerChain::reset()
+{
+    assert(empty());
```
MonsieurNicolas: I don't understand the contract: why do we require `empty()`?
Also, this is probably more of a `runtime_error` type of situation as it's not a bug in the `SyncingLedgerChain` implementation, but a bug in what callers are doing.

src/ledger/LedgerManagerImpl.cpp
```diff
@@ -623,9 +624,23 @@ LedgerManagerImpl::startCatchup(CatchupConfiguration configuration,
...
+        {
+            throw std::runtime_error(
+                "Unable to perform second catchup before first one finishes");
+        }
+        assert(mSyncingLedgers.empty());
+        mSyncingLedgers.reset();
+    }
+    else
+    {
+        assert(!mSyncingLedgers.empty());
```
MonsieurNicolas: again, I don't like this coupling with `mSyncingLedgers` and all those asserts.
Why do we need a non empty `mSyncingLedgers` at this stage? Expectations wrt `mSyncingLedgers` should be very simple. If you need an extra state, just add it.

Isn't it possible to have `addToSyncingLedgers` not actually add anything to `mSyncingLedgers` (like if `reset` was not called, which would potentially trip those asserts?). Right now it's hard to tell if this would just work.

src/ledger/SyncingLedgerChain.h
```diff
@@ -5,6 +5,7 @@
 // of this distribution or at http://www.apache.org/licenses/LICENSE-2.0
 
 #include <cstddef>
+#include <cstdint>
```
MonsieurNicolas: out of curiosity, why do you need this include?

src/ledger/LedgerManagerImpl.cpp
vogel: This place for reset is required by "manual" catchup. I don't like it. I don't like manual catchup from CommandHandler at all (it causes a lot of trouble, and is probably not used anywhere).

I would like to replace manual notion with "Online" and "Offline", where Online requires going throught WAITING_FOR_CLOSING_LEDGER state and Offline does not.

But anyway, this one line is for when you are doing manual catchup from CommandHandler and then another one later. And mSyncingLedgers must be empty here, because we are in WAITING_FOR_CLOSING_LEDGER, which is set when mSyncingLedgers is empty, and adding anything to mSyncingLedgers requires being not in WAITING_FOR_CLOSING_LEDGER state.

src/ledger/LedgerManagerImpl.cpp
vogel: Good point, this probably means command handler catchup probably does not work (please do not confuse it with command line catchup).

src/ledger/LedgerManagerImpl.cpp
vogel: I use such asserts as preconditions - adding logging here would serve no purpose, also moving asserts up the chan would remove the precondtions aspects from these asserts.

I would love to improve abstractions here, I thought I made a issue for that, but I cannot find it, so I probably didn't.

I think we should rework catchup/ledger abstractions a bit to make 100% separation between catchup code and applying current ledger code. That would probably require making another class that would decide whether to send given ledger to catchup or ledger class. But this is too big to implement here.

src/ledger/SyncingLedgerChain.cpp
vogel: If this is not empty, it mean that we have not used all of the buffered ledgers, which mean that the catchup has not finished.

src/ledger/LedgerManagerImpl.cpp
vogel: That would mean that startup is starting in automatic mode without any externalized ledger received by stellar-core, that is clearly wrong (only manual catchup can start that way). Also at that point `
addToSyncingLedgers` must add anything, because `reset` being not called before is not possible (that would some previous catchup is ongoing, which mean startCatchup is not possible to execute).

src/ledger/SyncingLedgerChain.h
vogel: I think `uint32_t` requires that.

PR #1974, opened 2019-02-26
Fix error in bulk loads for offers

jonjove:
Attempting to bulk load an offer keyed by (sid, oid) should not assert if an offer keyed by (sid', oid) exists. Instead it should just not return that offer.


graydon:
r+ 7115b87

Issue #1510, opened 2018-01-24
perform a memory usage analysis


MonsieurNicolas:
I think it has been a long time since we last did this.
Idea is to break down memory consumption by modules and see if some areas are getting out of control.

It looks like VM size on some nodes is close to 1GB and this is way too high given the current traffic.

A simple way to do it is to perform this analysis using:
* command line catchup (transaction subsystem)
* after joining testnet/public network (overlay kicks in)

MonsieurNicolas:
I think that tools like LSAN (leak sanitizer, enabled with asan), might be used for this as I think it's possible to start tracking for a range of time instead of the entire life time of the app

MonsieurNicolas:
snoozing: memory usage is very good right now

PR #1976, opened 2019-02-27
Quorum sets are sane, take advantage of it

MonsieurNicolas:
# Description

As part of #1510, I noticed that a function that we use all the time uses sets even though we don't need to:

historically quorum sets were more complicated, but since late 2015 we're enforcing that nodes in a quorum set only appear once.


graydon:
r+ 9d3b4fa

PR #1977, opened 2019-02-27
use stellar::shortHash for computing hashes of Hash

MonsieurNicolas:
# Description

Minor cleanup, when I implemented https://github.com/stellar/stellar-core/pull/1871 I didn't notice that we're using a custom and not so good hash function for the "Hash" type.

This PR fixes that.


graydon:
r+ aeeef92

PR #1978, opened 2019-02-27
util: replace LRU cache with two-random-choices-eviction cache.

graydon:
# Description

This replaces the LRU cache with a new one that does slightly better in a few ways:

  - Two-random-choice eviction avoids pathological thrashing due to LRU-hostile patterns.
  - Counts hits/misses/inserts/updates/evictions so we can push those to metrics.
  - Includes a single-lookup API rather than the two-lookup exists/get-with-exception API in the existing LRU cache (though supports the latter for backward compatibility).
  - Less memory-hungry: only keeps a single copy of each key; uses a vector of pointers for the auxiliary structure rather than a linked list with separate copies of keys.
  - Allocates its maximum size on construction (so _appears_ a bit more memory hungry? but easier to measure, fewer surprises, also no rehashing pauses; I _think_ this is the right thing to do.)

Perf shows LRU very slightly faster (as one would expect on non-pathological workloads) but it's close enough to machine-noise-level difference that while measuring the difference inverted a few times and LRU was slower. LRU does fewer insns but has more branch mispredicts, so time-wise it looks like a wash.

Since I threw in both a faster (single-lookup) API and counters to help us tune sizes (neither of which are at all exploited in this patch) I'm pretty sure we can win that back any difference if we're concerned about it. Personally I am not, it's more important to avoid thrashing cases.

# Checklist
- [x] Compiles
- [x] Ran all tests


src/util/CacheTable.h
```diff
@@ -0,0 +1,197 @@
...
+        mValueMap.erase(victim->first);
+        std::swap(victim, mValuePtrs.back());
+        mValuePtrs.pop_back();
+        ++mCounters.mEvicts;
+    }
+
+  public:
+    CacheTable(size_t maxSize) : mMaxSize(maxSize)
+    {
+        mValueMap.reserve(maxSize);
```
vogel: could use maxSize + 1, as both containers can grow over maxSize by one in `put`

src/util/CacheTable.h
```diff
@@ -0,0 +1,197 @@
...
+        mValueMap.clear();
+    }
+
+    template <typename F>
+    void
+    erase_if(F const& f)
+    {
+        for (size_t i = 0; i < mValuePtrs.size(); ++i)
+        {
+            map_value_type*& vp = mValuePtrs.at(i);
```
vogel: is that even legal?

src/util/CacheTests.cpp
```diff
@@ -0,0 +1,233 @@
+// Copyright 2019 Stellar Development Foundation and contributors. Licensed
```
vogel: need to be moved to src/util/test

graydon:
Closing out as @marta-lokhova is picking this up in #2013

PR #1980, opened 2019-02-27
bump overlay to 8

MonsieurNicolas:
# Description

This PR bumps the overlay version, no impact on the network as MIN is still 7 (we just need to regularly bump the version as to weed out old peers from the network)


marta-lokhova:
r+ 026fe7a5ac2992f2ff4f084e752aee45d14742f7

PR #1982, opened 2019-02-28
do not connect to peers that have more than 120 failures

### LINKED ISSUE #1962: exclude "really dead" peers from peers that we connect to
We should rethink a bit what to do with peers that are "probably dead".

Right now we keep trying to connect to peers forever.

What we should probably do is introduce a cuttoff for the number of failures, something like 120 (which would in average take about a week to reach given how we compute backoff).

When this cutoff is reached we should consider the peer "dead":
* do not try to connect to this peer (regardless of reason)
* do not forward this peer to other peers

Reason we need to do it like this (instead of deleting), is that this avoids resurrecting peers that are bad. We may want to add a "purge dead peers" command at some point, but it doesn't seem necessary yet.

vogel:

# Description

Resolves #1962

This one disables connecting to peers that have numfailures bigger than 120. Threshold for sending peers to other nodes is still kept at 10.

# Checklist
- [x] Compiles
- [x] Ran all tests

### DIFF
--- a/src/overlay/PeerManagerTests.cpp
+++ b/src/overlay/PeerManagerTests.cpp
@@ -8,10 +8,10 @@
 #include "main/Application.h"
 #include "main/Config.h"
 #include "overlay/OverlayManager.h"
+#include "overlay/RandomPeerSource.h"
 #include "overlay/StellarXDR.h"
 #include "test/TestUtils.h"
 #include "test/test.h"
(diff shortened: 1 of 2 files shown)


MonsieurNicolas:
r+ 7e8cc6ae34fefab08609b75341ffbce368113a10

PR #1986, opened 2019-03-01
Fix bulk loads for offers

marta-lokhova:
Looks like latest changes to bulk loads for offers introduced a few issues, which this PR attempts to fix:
* we should proceed to add the offer if `mSellerIDsByOfferID[offerID] == pubKey`
* using `continue` is dangerous because we have to remember to call `st.fetch()`, otherwise it's an infinite loop, so I removed it completely

### DIFF
--- a/src/ledger/LedgerTxnOfferSQL.cpp
+++ b/src/ledger/LedgerTxnOfferSQL.cpp
@@ -800,33 +800,32 @@ class BulkLoadOffersOperation
         std::vector<LedgerEntry> res;
         while (st.got_data())
         {
+            auto pubKey = KeyUtils::fromStrKey<PublicKey>(sellerID);
+
             // Exclude offers where sellerID in LedgerKey doesn't match sellerID
             // in LedgerEntry
-            auto pubKey = KeyUtils::fromStrKey<PublicKey>(sellerID);
             if (mSellerIDsByOfferID[offerID] == pubKey)
             {
-                continue;
+                res.emplace_back();
+                auto& le = res.back();
+                le.data.type(OFFER);
+                auto& oe = le.data.offer();
+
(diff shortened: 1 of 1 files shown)


src/ledger/LedgerTxnOfferSQL.cpp
```diff
@@ -800,34 +800,33 @@ class BulkLoadOffersOperation
         std::vector<LedgerEntry> res;
         while (st.got_data())
         {
+            auto pubKey = KeyUtils::fromStrKey<PublicKey>(sellerID);
```
vogel: sellerID does not change, so maybe it can be extracted before loop begings?

src/ledger/LedgerTxnOfferSQL.cpp
```diff
@@ -800,34 +800,33 @@ class BulkLoadOffersOperation
...
+
             // Exclude offers where sellerID in LedgerKey doesn't match sellerID
             // in LedgerEntry
-            auto pubKey = KeyUtils::fromStrKey<PublicKey>(sellerID);
             if (mSellerIDsByOfferID[offerID] == pubKey)
             {
```
vogel: this looks like condition has been inverted, so either: previous version is wrong, current version is wrong or I'm missing something

jonjove: Yes previous version was wrong. #1974 fixed something wrong into something wrong in a different way.... code matches comment now.

jonjove:
r+ 2175713

src/ledger/LedgerTxnOfferSQL.cpp
vogel: Right, sorry for that

PR #1987, opened 2019-03-01
Windows: fix race condition that causes spurious failures while adopting buckets

MonsieurNicolas:
# Description

This PR fixes a race condition where file descriptors are inherited by child processes (such as "gzip"), which causes some operations, such as renaming buckets to fail with a "file in use" error.

The fix here is to move to Windows 7+ APIs, that allow to whitelist which descriptors to pass to the child process, in our case we only need the descriptor for standard output.


graydon:
How fun.

graydon:
r+ 484eee7

MonsieurNicolas:
didn't see the first `#ifdef` was an `#ifndef`

MonsieurNicolas:
r+ 13e9306

PR #1924, opened 2019-01-24
remove unnecessary Application parameters

vogel:
Signed-off-by: Rafał Malinowski <rafal.przemyslaw.malinowski@gmail.com>

# Description

Remove unnecessary Application parameters from many OperationFrame classes

# Checklist
- [x] Reviewed the [contributing](https://github.com/stellar/stellar-core/blob/master/CONTRIBUTING.md#submitting-changes) document
- [x] Rebased on top of master (no merge commits)
- [x] Ran `clang-format` v5.0.0 (via `make format` or the Visual Studio extension)
- [x] Compiles
- [x] Ran all tests
- [x] If change impacts performance, include supporting evidence per the [performance document](https://github.com/stellar/stellar-core/blob/master/performance-eval/performance-eval.md)

PR #1988, opened 2019-03-02
Simplified prefetcher api

### LINKED ISSUE #1859: data access: pre-fetching of data
This issue tracks investigating pre-fetching changes for various data sets

marta-lokhova:
Resolves #1859 

Take 2 at the prefetcher implementation; this one is a much simpler version that avoids any queueing, and simply prefetches a set of unique keys in batches. 

I did some preliminary performance testing, and it looks like on postgres, catchup and applying ledger chain for 2000 ledgers took 175 s with prefetching and 230 without. With 5000 ledgers, same test took 548.59 with prefetching, and 650 s without.

### DIFF
--- a/src/ledger/LedgerManagerImpl.cpp
+++ b/src/ledger/LedgerManagerImpl.cpp
@@ -26,6 +26,8 @@
 #include "main/Config.h"
 #include "main/ErrorMessages.h"
 #include "overlay/OverlayManager.h"
+#include "transactions/OperationFrame.h"
+#include "transactions/TransactionUtils.h"
 #include "util/Logging.h"
 #include "util/XDROperators.h"
 #include "util/format.h"
@@ -119,6 +121,8 @@ LedgerManagerImpl::LedgerManagerImpl(Application& app)
     , mLedgerAgeClosed(app.getMetrics().NewTimer({"ledger", "age", "closed"}))
     , mLedgerAge(
           app.getMetrics().NewCounter({"ledger", "age", "current-seconds"}))
+    , mPrefetchHitRate(
+          app.getMetrics().NewCounter({"ledger", "prefetch", "hit-rate"}))
     , mLastClose(mApp.getClock().now())
     , mSyncingLedgersSize(
           app.getMetrics().NewCounter({"ledger", "memory", "queued-ledgers"}))
@@ -739,6 +743,8 @@ LedgerManagerImpl::syncMetrics()
 {
     mSyncingLedgersSize.set_count(mSyncingLedgers.size());
     mLedgerAge.set_count(secondsSinceLastLedgerClose());
+    mPrefetchHitRate.set_count(
+        std::round(mApp.getLedgerTxnRoot().getPrefetchHitRate() * 100));
     mApp.syncOwnMetrics();
 }
 
@@ -815,7 +821,8 @@ LedgerManagerImpl::closeLedger(LedgerCloseData const& ledgerData)
     // sorted such that sequence numbers are respected
     vector<TransactionFramePtr> txs = ledgerData.getTxSet()->sortForApply();
 
-    // first, charge fees
+    // first, prefetch source accounts fot txset, then charge fees
+    prefetchTxSourceIds(txs);
     processFeesSeqNums(txs, ltx,
                        ledgerData.getTxSet()->getBaseFee(header.current()));
 
@@ -959,6 +966,44 @@ LedgerManagerImpl::processFeesSeqNums(std::vector<TransactionFramePtr>& txs,
 }
 
 void
+LedgerManagerImpl::prefetchTxSourceIds(std::vector<TransactionFramePtr>& txs)
+{
+    if (mApp.getConfig().PREFETCH_BATCH_SIZE > 0)
+    {
+        std::unordered_set<LedgerKey> keys;
+        for (auto const& tx : txs)
+        {
+            keys.emplace(accountKey(tx->getSourceID()));
+        }
+        auto& root = mApp.getLedgerTxnRoot();
+        root.prefetch(keys);
+    }
+}
+
+void
+LedgerManagerImpl::prefetchTransactionData(
+    std::vector<TransactionFramePtr>& txs)
+{
+    if (mApp.getConfig().PREFETCH_BATCH_SIZE > 0)
+    {
+        auto& root = mApp.getLedgerTxnRoot();
+        std::unordered_set<LedgerKey> keysToPrefetch;
+        for (auto const& tx : txs)
+        {
+            for (auto const& op : tx->getOperations())
+            {
+                if (!(tx->getSourceID() == op->getSourceID()))
+                {
+                    keysToPrefetch.emplace(accountKey(op->getSourceID()));
+                }
+                op->insertLedgerKeysToPrefetch(keysToPrefetch);
+            }
+        }
+        root.prefetch(keysToPrefetch);
+    }
+}
+
+void
 LedgerManagerImpl::applyTransactions(std::vector<TransactionFramePtr>& txs,
                                      AbstractLedgerTxn& ltx,
                                      TransactionResultSet& txResultSet)
@@ -967,19 +1012,21 @@ LedgerManagerImpl::applyTransactions(std::vector<TransactionFramePtr>& txs,
 
     // Record counts
     auto numTxs = txs.size();
+    size_t numOps = 0;
     if (numTxs > 0)
     {
         mTransactionCount.Update(static_cast<int64_t>(numTxs));
-        size_t numOps =
-            std::accumulate(txs.begin(), txs.end(), size_t(0),
-                            [](size_t s, TransactionFramePtr const& v) {
-                                return s + v->getOperations().size();
-                            });
+        numOps = std::accumulate(txs.begin(), txs.end(), size_t(0),
+                                 [](size_t s, TransactionFramePtr const& v) {
+                                     return s + v->getOperations().size();
+                                 });
         mOperationCount.Update(static_cast<int64_t>(numOps));
         CLOG(INFO, "Tx") << fmt::format("applying ledger {} (txs:{}, ops:{})",
                                         ltx.loadHeader().current().ledgerSeq,
                                         numTxs, numOps);
     }
+
+    prefetchTransactionData(txs);
 
     for (auto tx : txs)
     {
@@ -1020,6 +1067,23 @@ LedgerManagerImpl::applyTransactions(std::vector<TransactionFramePtr>& txs,
         tx->storeTransaction(mApp.getDatabase(), ledgerSeq, tm, ++index,
                              txResultSet);
     }
+
+    logTxApplyMetrics(ltx, numTxs, numOps);
+}
+
+void
+LedgerManagerImpl::logTxApplyMetrics(AbstractLedgerTxn& ltx, size_t numTxs,
+                                     size_t numOps)
+{
+    auto ledgerSeq = ltx.loadHeader().current().ledgerSeq;
+    auto hitRate = mApp.getLedgerTxnRoot().getPrefetchHitRate() * 100;
+
+    CLOG(DEBUG, "Ledger") << "Ledger: " << ledgerSeq << " txs: " << numTxs
+                          << ", ops: " << numOps
+                          << ", prefetch hit rate (%): " << hitRate;
+
+    // We lose a bit of precision here, as medida only accepts int64_t
+    mPrefetchHitRate.set_count(std::round(hitRate));
 }
 
 void
--- a/src/ledger/LedgerTxn.cpp
+++ b/src/ledger/LedgerTxn.cpp
@@ -1257,17 +1257,21 @@ LedgerTxn::Impl::EntryIteratorImpl::clone() const
 
 // Implementation of LedgerTxnRoot ------------------------------------------
 LedgerTxnRoot::LedgerTxnRoot(Database& db, size_t entryCacheSize,
-                             size_t bestOfferCacheSize)
-    : mImpl(std::make_unique<Impl>(db, entryCacheSize, bestOfferCacheSize))
+                             size_t bestOfferCacheSize,
+                             size_t prefetchBatchSize)
+    : mImpl(std::make_unique<Impl>(db, entryCacheSize, bestOfferCacheSize,
+                                   prefetchBatchSize))
 {
 }
 
 LedgerTxnRoot::Impl::Impl(Database& db, size_t entryCacheSize,
-                          size_t bestOfferCacheSize)
+                          size_t bestOfferCacheSize, size_t prefetchBatchSize)
     : mDatabase(db)
     , mHeader(std::make_unique<LedgerHeader>())
     , mEntryCache(entryCacheSize)
     , mBestOffersCache(bestOfferCacheSize)
+    , mMaxCacheSize(entryCacheSize)
+    , mBulkLoadBatchSize(prefetchBatchSize)
     , mChild(nullptr)
 {
 }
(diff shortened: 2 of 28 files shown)


src/ledger/LedgerTxn.cpp
```diff
@@ -1891,4 +1893,75 @@ LedgerTxnRoot::encodeHomeDomainsBase64()
...
+    std::unordered_set<LedgerKey> offers;
+    std::unordered_set<LedgerKey> trustlines;
+    std::unordered_set<LedgerKey> data;
+
+    auto cacheResult =
+        [&](std::unordered_map<LedgerKey, std::shared_ptr<LedgerEntry const>>
+                res) {
+            for (auto const& item : res)
+            {
+                putInEntryCache(item.first, item.second, LoadType::PREFETCH);
```
graydon: Will this overwrite a LoadType::IMMEDIATE with a LoadType::PREFETCH?

(I doubt it's critical either way, just checking if it's intentional either way)

src/transactions/TransactionUtils.h
```diff
@@ -15,6 +15,12 @@ class AbstractLedgerTxn;
 class LedgerTxnEntry;
 class LedgerTxnHeader;
 class TrustLineWrapper;
+struct LedgerKey;
+
+LedgerKey getLedgerKey(AccountID const& accountID);
```
graydon: The fact that these 4 overloads make keys for accounts, trustlines, offers, and data seems a little non-obvious from their signatures & names. Maybe name them differently (eg. `getLedgerKeyForAccount`, `getLedgerKeyForTrustLine`, ...)?

graydon:
Just a couple nits, otherwise looks good! Thanks for taking a more-minimized go at this, it feels much easier to me to understand/predict how it's going to behave (and glad to hear it still gives some perf wins in this form!)

src/ledger/LedgerTxn.cpp
marta-lokhova: I think this wouldn't happen since `insertIfNotLoaded` checks if a key already exists in the cache.

src/main/Config.h
```diff
@@ -218,6 +218,11 @@ class Config : public std::enable_shared_from_this<Config>
     size_t ENTRY_CACHE_SIZE;
     size_t BEST_OFFERS_CACHE_SIZE;
 
+    // Data layer prefetcher configuration
+    // - PREFETCH_BATCH_SIZE determines how many records we'll prefetch per
+    // SQL load.
+    size_t PREFETCH_BATCH_SIZE;
```
jonjove: This should also be documented in docs/stellar-core_example.cfg.

src/ledger/LedgerTxn.cpp
```diff
@@ -1814,18 +1822,12 @@ LedgerTxnRoot::Impl::rollbackChild()
     mChild = nullptr;
 }
 
-LedgerTxnRoot::Impl::EntryCacheKey
-LedgerTxnRoot::Impl::getEntryCacheKey(LedgerKey const& key) const
```
jonjove: Nice, thanks for removing this!

src/ledger/LedgerTxnImpl.h
```diff
@@ -376,9 +376,19 @@ class LedgerTxn::Impl::EntryIteratorImpl : public EntryIterator::AbstractImpl
 // been lost.
 class LedgerTxnRoot::Impl
 {
-    typedef std::string EntryCacheKey;
-    typedef cache::lru_cache<EntryCacheKey, std::shared_ptr<LedgerEntry const>>
-        EntryCache;
+    enum class LoadType
```
jonjove: I don't think we're currently using these for anything but I think they are intended to facilitate the calculation of some metrics. Can we use these to design some metrics about how successful the prefetching strategy is?

src/ledger/LedgerManagerImpl.cpp
```diff
@@ -940,6 +943,45 @@ LedgerManagerImpl::processFeesSeqNums(std::vector<TransactionFramePtr>& txs,
...
 
+void
+LedgerManagerImpl::prefetchTxSourceIds(std::vector<TransactionFramePtr>& txs)
+{
+    if (mApp.getConfig().PREFETCH_BATCH_SIZE > 0)
+    {
+        std::unordered_set<LedgerKey> keys;
+        for (auto const& tx : txs)
+        {
+            keys.insert(getLedgerKey(tx->getSourceID()));
```
jonjove: Better to use `emplace` to avoid copies.

src/ledger/LedgerManagerImpl.cpp
```diff
@@ -940,6 +943,45 @@ LedgerManagerImpl::processFeesSeqNums(std::vector<TransactionFramePtr>& txs,
...
+    {
+        auto& root = mApp.getLedgerTxnRoot();
+        std::unordered_set<LedgerKey> keysToPrefetch;
+        for (auto const& tx : txs)
+        {
+            for (auto const& op : tx->getOperations())
+            {
+                if (!(tx->getSourceID() == op->getSourceID()))
+                {
+                    keysToPrefetch.insert(getLedgerKey(op->getSourceID()));
```
jonjove: Same as above regarding `emplace`.

src/transactions/ManageOfferOpFrame.cpp
```diff
@@ -472,4 +472,36 @@ ManageOfferOpFrame::buildOffer(AccountID const& account,
...
+    std::unordered_set<LedgerKey> keys;
+    if (mManageOffer.amount == 0)
+    {
+        return keys;
+    }
+
+    // Prefetch existing offer
+    if (mManageOffer.offerID)
+    {
+        keys.insert(getLedgerKey(getSourceID(), mManageOffer.offerID));
```
jonjove: Also better to use `emplace` for all of these (not going to write this comment again, but it applies to other analogous code as well).

src/transactions/OperationFrame.h
```diff
@@ -84,5 +84,8 @@ class OperationFrame
     {
         return mOperation;
     }
+
+    virtual std::unordered_set<LedgerKey>
+    getLedgerKeysToPrefetch(Application& app);
```
jonjove: I don't think we should include `app` here. @vogel is already working to clean this up in other places, see #1924.

Also one concern that I have about this interface is that it might lead to the construction of many intermediate containers. I'm not sure if it's worth changing this right now (I think this interface is much cleaner than one which would take `std::unordered_set<LedgerKey>` by reference for example), but it's something we should keep an eye on for performance reasons.

src/transactions/ManageDataOpFrame.cpp
```diff
@@ -96,4 +96,12 @@ ManageDataOpFrame::doCheckValid(Application& app, uint32_t ledgerVersion)
 
     return true;
 }
+
+std::unordered_set<LedgerKey>
+ManageDataOpFrame::getLedgerKeysToPrefetch(Application& app)
+{
+    std::unordered_set<LedgerKey> keys;
+    keys.insert(stellar::getLedgerKey(getSourceID(), mManageData.dataName));
+    return keys;
```
jonjove: Nitpick: For brevity could just `return {getLedgerKey(getSourceID(), mManageData.dataName)};`

src/transactions/CreateAccountOpFrame.cpp
```diff
@@ -105,4 +105,12 @@ CreateAccountOpFrame::doCheckValid(Application& app, uint32_t ledgerVersion)
 
     return true;
 }
+
+std::unordered_set<LedgerKey>
+CreateAccountOpFrame::getLedgerKeysToPrefetch(Application& app)
+{
+    std::unordered_set<LedgerKey> keys{
+        getLedgerKey(mCreateAccount.destination)};
+    return keys;
```
jonjove: Nitpick: For brevity could just `return {getLedgerKey(mCreateAccount.destination)};`

src/transactions/TransactionUtils.h
marta-lokhova: Sure. I think @vogel didn't like these names very much also. Maybe we can remove `get` and add entity names, like `ledgerKeyForAccount`, `ledgerKeyForTrustline`, etc.

src/transactions/OperationFrame.h
marta-lokhova: Good ideas, agreed.

src/transactions/TransactionUtils.h
vogel: I would vote for either just ledgerKey(..) for everything or accountKey, trustlineKey, offerKey...

src/ledger/LedgerManagerImpl.cpp
```diff
@@ -802,7 +804,8 @@ LedgerManagerImpl::closeLedger(LedgerCloseData const& ledgerData)
     // sorted such that sequence numbers are respected
     vector<TransactionFramePtr> txs = ledgerData.getTxSet()->sortForApply();
 
-    // first, charge fees
+    // first, prefetch source accounts fot txset, then charge fees
+    prefetchTxSourceIds(txs);
```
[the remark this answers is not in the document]
marta-lokhova: There is no benefit, but it's mostly to accommodate how `mOperations` in `TransactionFrame` is populated. Before fees are processed, `mOperations` is empty (and is populated via private a `resetResults` method)

src/ledger/LedgerTxnImpl.h
marta-lokhova: Thanks for the review! I'm out of time today, but I pushed a few commits that address your comments, except this one (which I'll re-visit when I get back).

src/transactions/TransactionUtils.h
marta-lokhova: ok, I went with Rafal's suggestion, hope that makes everyone happy :)

src/transactions/ManageOfferOpFrame.cpp
```diff
@@ -470,4 +470,36 @@ ManageOfferOpFrame::buildOffer(AccountID const& account,
     o.flags = flags;
     return o;
 }
+
+std::unordered_set<LedgerKey>
+ManageOfferOpFrame::getLedgerKeysToPrefetch()
+{
+    // TODO (mlo) will need to deal with best offers as well
```
MonsieurNicolas: do not keep TODO in code, open an issue instead

src/transactions/PathPaymentOpFrame.cpp
```diff
@@ -286,4 +286,39 @@ PathPaymentOpFrame::doCheckValid(uint32_t ledgerVersion)
...
     return true;
 }
+
+std::unordered_set<LedgerKey>
+PathPaymentOpFrame::getLedgerKeysToPrefetch()
+{
+    std::unordered_set<LedgerKey> keys;
+    keys.emplace(stellar::accountKey(mPathPayment.destination));
+
+    std::vector<Asset> allAssets{mPathPayment.sendAsset,
```
MonsieurNicolas: this is a bit wasteful to allocate a vector of `Asset`: you can probably just refactor this to use a function that processes an Asset and performs whatever is in the loop, and call the function on `sendAsset`, `destAsset` and using `std::for_each` on the path itself

src/transactions/OperationFrame.h
```diff
@@ -82,5 +82,7 @@ class OperationFrame
     {
         return mOperation;
     }
+
+    virtual std::unordered_set<LedgerKey> getLedgerKeysToPrefetch();
```
MonsieurNicolas: should be `const`?

src/ledger/LedgerManagerImpl.cpp
```diff
@@ -1002,6 +1046,27 @@ LedgerManagerImpl::applyTransactions(std::vector<TransactionFramePtr>& txs,
...
+    logTxApplyMetrics(ltx, numTxs, numOps);
+}
+
+void
+LedgerManagerImpl::logTxApplyMetrics(AbstractLedgerTxn& ltx, size_t numTxs,
+                                     size_t numOps)
+{
+    auto ledgerSeq = ltx.loadHeader().current().ledgerSeq;
+
+    CLOG(DEBUG, "Ledger")
```
MonsieurNicolas: not sure about logging this first line: we're already recording those via the metrics endpoint

src/ledger/LedgerTxn.cpp
```diff
@@ -1841,12 +1955,17 @@ LedgerTxnRoot::Impl::getFromEntryCache(EntryCacheKey const& cacheKey) const
...
+    LedgerKey const& key, std::shared_ptr<LedgerEntry const> const& entry,
+    LoadType type) const
 {
     try
     {
-        mEntryCache.put(cacheKey, entry);
+        mEntryCache.put(key, {entry, type});
+        if (type == LoadType::PREFETCH)
+        {
+            assert(mPrefetchMetrics.find(key) == mPrefetchMetrics.end());
```
MonsieurNicolas: avoid using `assert` if you can

src/ledger/LedgerManagerImpl.cpp
```diff
@@ -1002,6 +1046,27 @@ LedgerManagerImpl::applyTransactions(std::vector<TransactionFramePtr>& txs,
...
+        << "Ledger: " << ledgerSeq << " txs: " << numTxs << ", ops: " << numOps
+        << ", AVG application time per tx: " << mTransactionApply.mean()
+        << " ± " << mTransactionApply.std_dev()
+        << ", MIN application time per tx: " << mTransactionApply.min()
+        << ", MAX " << mTransactionApply.max() << ", 75th percentile: "
+        << mTransactionApply.GetSnapshot().get75thPercentile();
+
+    CLOG(DEBUG, "Ledger") << "Ledger: " << ledgerSeq << " txs: " << numTxs
```
MonsieurNicolas: Should we add the rate as a metric?

src/transactions/ManageOfferOpFrame.cpp
marta-lokhova: Done
https://github.com/stellar/stellar-core/issues/2031

src/ledger/LedgerManagerImpl.cpp
marta-lokhova: Sure, added a counter to track the hit rate.

marta-lokhova:
Updated.

src/ledger/LedgerManagerImpl.cpp
```diff
@@ -940,6 +943,45 @@ LedgerManagerImpl::processFeesSeqNums(std::vector<TransactionFramePtr>& txs,
...
+        std::unordered_set<LedgerKey> keysToPrefetch;
+        for (auto const& tx : txs)
+        {
+            for (auto const& op : tx->getOperations())
+            {
+                if (!(tx->getSourceID() == op->getSourceID()))
+                {
+                    keysToPrefetch.emplace(accountKey(op->getSourceID()));
+                }
+                auto knownKeys = op->getLedgerKeysToPrefetch();
```
MonsieurNicolas: I wonder if it would be better to change `getLedgerKeysToPrefetch` and turn it into a `insertLedgerKeysToPrefetch` that takes a reference to a set and avoid the intermediate set (causes memory churn and processing)

MonsieurNicolas:
you also need to rebase

src/ledger/LedgerManagerImpl.cpp
marta-lokhova: Updated.

marta-lokhova:
Updated and rebased (most notable change from the rebase is `ManageOfferOpFrameBase::insertLedgerKeysToPrefetch` to accommodate CAP006 changes)

MonsieurNicolas:
r+ 9b6f3a9c8a00c873f01bc43d2e2d312e45808287

marta-lokhova:
need to fix `clang-format`

marta-lokhova:
This is now updated.

MonsieurNicolas:
r+ 72db3e6f590528c7ecf5b354a30562a46ab29a69

PR #2000, opened 2019-03-08
add new field "failure" to tx endpoint

### LINKED ISSUE #1999: Remove ambiguous result from "tx" endpoint
# Issue description

Right now when submitting transactions to the "tx" endpoint, the result is

The transaction was "fast failed" and rejected by the node even before it got a chance to reach consensus.

The problem is that client code, if not careful, may think that this is a "normal" failure, which may cause clients to resubmit their transaction.

Proposal is to soft deprecate the `error` field (it will be removed in a future release), and replace it by a `failure` field that cannot be confused with a "in consensus" failure.

bad sequence number (`TransactionResultCode=-5=txBAD_SEQ`)
```json
{
  "status": "ERROR" ,
  "error": "AAAAAAAAA+j////7AAAAAA==",
  "failure": {
    "TransactionResultCode" : "-5"
  }
}
```

first operation hitting `opBAD_AUTH` (`TransactionResultCode=-1=txFAILED`)
```json
{
  "status": "ERROR" ,
  "error": "AAAAAAAAA+j/////AAAAAf////8AAAAA",
  "failure": {
    "TransactionResultCode" : "-1" ,
    "FirstFailedOp": "0",
    "OperationResult": "/////w=="
  }
}
```

second operation hitting `opINNER::ACCOUNT_MERGE::ACCOUNT_MERGE_MALFORMED`

```json
{
  "status": "ERROR" ,
  "error": "AAAAAAAAB9D/////AAAAAgAAAAAAAAABAAAAAAAAAAAAAAAI/////wAAAAA=",
  "failure": {
    "TransactionResultCode" : "-1" ,
    "FirstFailedOp": "1",
    "OperationResult": "AAAAAAAAAAj/////"
  }
}
```

MonsieurNicolas:
# Description

Resolves #1999

Do not merge yet - this adds less error prone failure information on the `tx` endpoint

### DIFF
--- a/src/main/CommandHandler.cpp
+++ b/src/main/CommandHandler.cpp
@@ -680,6 +680,43 @@ CommandHandler::ll(std::string const& params, std::string& retStr)
     retStr = root.toStyledString();
 }
 
+static bool
+isFailedOperationResult(OperationResult const& or)
+{
+    switch (or.code())
+    {
+    case opINNER:
+    {
+        auto& op = or.tr();
+#define checkOp(t, field) \
+    case t: \
+        return op.##field().code() < 0
+        switch (op.type())
+        {
+            checkOp(CREATE_ACCOUNT, createAccountResult);
+            checkOp(PAYMENT, paymentResult);
+            checkOp(PATH_PAYMENT, pathPaymentResult);
+            checkOp(MANAGE_OFFER, manageOfferResult);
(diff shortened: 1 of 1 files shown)


vogel:
FYI: there is new "Draft PR" feature in Github, those are unmergeable by default.

PR #2001, opened 2019-03-08
overlay: add config param PEER_STRAGGLER_TIMEOUT to control straggler drops

graydon:
# Description

This just adds a config param to control slow-peer disconnection behaviour that was introduced (in a hard-coded way) back in f4250142fce723876baf3a95963b727c7b0eb202.

It also increases the default to 120s, which is 4x the idle timeout and ought to be plenty to deal with transient conditions.

The idea here is just to give users a knob they can turn if we find too much disconnection is happening in the field, without having to run an emergency release just to change the default. _Likely_ the default value is fine, but you know: networks.

# Checklist
- [x] Compiles
- [x] Ran all tests

### DIFF
--- a/docs/stellar-core_example.cfg
+++ b/docs/stellar-core_example.cfg
@@ -117,6 +117,11 @@ PEER_AUTHENTICATION_TIMEOUT=2
 # time when authenticated.
 PEER_TIMEOUT=30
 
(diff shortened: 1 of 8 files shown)


MonsieurNicolas:
r+ f7c3eaba49d0074db146bd72b2c8f6710c62ea07

PR #2002, opened 2019-03-13
bug 1099 - consolidate PRNG use in tests and make seedable.

### LINKED ISSUE #1099: Make tests use randomized NodeIDs and AccountIDs
As shown in Force SCP, the result of tests sometimes depends on generated NodeID (for some values of NodeIDs in Force SCP a third node is chosen as round leader over and over again which timeouts the tests).

It is possible that similar issues would be found in other tests if ID for nodes (and possibly accounts) were not hardcoded but chosen at random.

graydon:
# Description

This switches all uses of PRNGs in testing to the single global PRNG and adds support for manual seeding on the command-line. Additionally it reports the seed at the end of a run that ends in failure, so that you can run tests in a shell while-loop passing random seeds and it will conveniently break on the first nondeterministic failure. Should help shake out tests that are only passing due to lucky configuration of the (currently default-seeded) PRNGs.

Resolves #1099 

# Checklist
- [x] Compiles
- [x] Ran all tests

### DIFF
--- a/src/crypto/SecretKey.cpp
+++ b/src/crypto/SecretKey.cpp
@@ -10,6 +10,7 @@
 #include "main/Config.h"
 #include "transactions/SignatureUtils.h"
 #include "util/HashOfHash.h"
+#include "util/Math.h"
 #include "util/RandomEvictionCache.h"
 #include <memory>
 #include <mutex>
@@ -148,6 +149,38 @@ SecretKey::random()
 #endif
     return sk;
 }
+
+#ifdef BUILD_TESTS
+static SecretKey
+pseudoRandomForTestingFromPRNG(std::default_random_engine& engine)
+{
+    std::vector<uint8_t> bytes;
+    for (size_t i = 0; i < crypto_sign_SEEDBYTES; ++i)
+    {
+        bytes.push_back(static_cast<uint8_t>(engine()));
+    }
+    return SecretKey::fromSeed(bytes);
+}
+
+SecretKey
+SecretKey::pseudoRandomForTesting()
+{
+    // Reminder: this is not cryptographic randomness or even particularly hard
+    // to guess PRNG-ness. It's intended for _deterministic_ use, when you want
+    // "slightly random-ish" keys, for test-data generation.
+    return pseudoRandomForTestingFromPRNG(gRandomEngine);
+}
+
+SecretKey
+SecretKey::pseudoRandomForTestingFromSeed(unsigned int seed)
+{
+    // Reminder: this is not cryptographic randomness or even particularly hard
+    // to guess PRNG-ness. It's intended for _deterministic_ use, when you want
+    // "slightly random-ish" keys, for test-data generation.
+    std::default_random_engine tmpEngine(seed);
+    return pseudoRandomForTestingFromPRNG(tmpEngine);
+}
+#endif
 
 SecretKey
 SecretKey::fromSeed(ByteSlice const& seed)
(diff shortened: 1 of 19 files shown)


src/invariant/test/ConservationOfLumensTests.cpp
```diff
@@ -161,13 +159,13 @@ TEST_CASE("Account balances changed without inflation",
         std::vector<LedgerEntry> entries1;
         std::generate_n(std::back_inserter(entries1), N,
                         std::bind(generateRandomAccount, 2));
-        entries1 = updateBalances(entries1, *app, gen);
+        entries1 = updateBalances(entries1, *app, gRandomEngine);
```
vogel: Is there a reason to pass gRandomEngine in here?
If we must always use the global one, let just use it inside updateBalances as well instead of parameter gen (applies also to other functions and methods).

vogel:
There are also a few calls of plain std::rand() in the code - do these need to migrate to gRandomEngine as well?

src/invariant/test/ConservationOfLumensTests.cpp
graydon: Fair, fixed.

graydon:
Concerning the use of the global system PRNG via `std::rand()`, Catch already seeds it (via `std::srand()`) when given the `--rng-seed` option. This is just extending that seeding process to our own `gRandomEngine`.

graydon:
(Pushed update with changes requested. Anything else or is this good to go?)

src/crypto/SecretKey.cpp
```diff
@@ -149,6 +150,20 @@ SecretKey::random()
     return sk;
 }
 
+SecretKey
```
MonsieurNicolas: should we ifdef this stuff with `BUILD_TESTS` or move to a test only file?

src/test/test.cpp
```diff
@@ -251,10 +261,18 @@ runTest(CommandLineArgs const& args)
     {
         gVersionsToTest.emplace_back(Config::CURRENT_LEDGER_PROTOCOL_VERSION);
     }
+    if (seed != 0)
```
MonsieurNicolas: I *think* a better way (or in addition to) is to use https://github.com/catchorg/Catch2/blob/master/docs/event-listeners.md#top
and set the seed every time we run a test or do something clever when sections start.

Without this, we still have non deterministic behavior of some tests (that depend on real clock) that bleed into other tests

MonsieurNicolas: IF it's not already the case, we can do the same thing to `std::rand()`

src/crypto/SecretKey.cpp
graydon: Oh, perhaps. Will look at that.

src/test/test.cpp
graydon: Oh, interesting! I will take a look, thanks!

MonsieurNicolas:
We should probably revive and merge this as it helps with tests stability @graydon

src/test/test.cpp
graydon: Fixed.

src/crypto/SecretKey.cpp
graydon: Fixed.

graydon:
@MonsieurNicolas Sorry this took so long, there were a lot of fussy details concerning the persistent test configs. Looks like it's all worked out now though.

MonsieurNicolas:
r+ 7e8d137eacff9230fac38505096213692c6fac73

PR #2009, opened 2019-03-15
improve error messages

### LINKED ISSUE #1796: ensure that messages logged with level FATAL, ERROR, INFO and WARNING are user friendly
# Issue type
  feature request

# Version found
  * 10.0.0

# Issue description
 Many users that haven't upgraded their core to 10.0.0 are seeing some strange errors about catchup but instead core should check if the latest ledger supports the current protocol version and if not exit with the simple, user-friendly message like: "Your stellar-core does not support protocol version X. Please upgrade to the latest version."

vogel:

# Description

Resolves #1796

Better error messages for some cases. Also remove one copy of messages from InvariantDoesNotHold exception (one still remain from terminate call).

# Checklist
- [x] Compiles
- [x] Ran all tests

### DIFF
--- a/src/catchup/ApplyLedgerChainWork.cpp
+++ b/src/catchup/ApplyLedgerChainWork.cpp
@@ -7,11 +7,14 @@
 #include "history/FileTransferInfo.h"
 #include "history/HistoryManager.h"
 #include "historywork/Progress.h"
+#include "invariant/InvariantDoesNotHold.h"
 #include "ledger/CheckpointRange.h"
 #include "ledger/LedgerManager.h"
 #include "lib/xdrpp/xdrpp/printer.h"
 #include "main/Application.h"
-#include "util/format.h"
+#include "main/ErrorMessages.h"
+
+#include <lib/util/format.h>
 #include <medida/meter.h>
 #include <medida/metrics_registry.h>
 
@@ -21,7 +24,7 @@ namespace stellar
 ApplyLedgerChainWork::ApplyLedgerChainWork(
     Application& app, WorkParent& parent, TmpDir const& downloadDir,
     LedgerRange range, LedgerHeaderHistoryEntry& lastApplied)
-    : Work(app, parent, std::string("apply-ledger-chain"))
+    : Work(app, parent, std::string("apply-ledger-chain"), RETRY_NEVER)
     , mDownloadDir(downloadDir)
     , mRange(range)
     , mCurrSeq(
@@ -247,7 +250,13 @@ ApplyLedgerChainWork::onRun()
         }
         scheduleSuccess();
     }
-    catch (std::runtime_error& e)
+    catch (InvariantDoesNotHold&)
+    {
+        // already displayed e.what()
+        CLOG(ERROR, "History") << "Replay failed";
+        throw;
+    }
+    catch (std::exception& e)
     {
         CLOG(ERROR, "History") << "Replay failed: " << e.what();
         scheduleFailure();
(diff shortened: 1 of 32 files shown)


src/ledger/LedgerManagerImpl.cpp
```diff
@@ -782,13 +785,20 @@ LedgerManagerImpl::closeLedger(LedgerCloseData const& ledgerData)
 
         CLOG(ERROR, "Ledger")
             << "Full LCL: " << xdr::xdr_to_string(getLastClosedLedgerHeader());
+        CLOG(ERROR, "Ledger") << POSSIBLY_CORRUPTED_QUORUM_SET;
```
MonsieurNicolas: this is wrong: bad hash is likely because local database or buckets were corrupted

src/ledger/LedgerManagerImpl.cpp
```diff
@@ -826,6 +836,7 @@ LedgerManagerImpl::closeLedger(LedgerCloseData const& ledgerData)
         catch (xdr::xdr_runtime_error)
         {
             CLOG(FATAL, "Ledger") << "Unknown upgrade step at index " << i;
+            CLOG(FATAL, "Ledger") << POSSIBLY_CORRUPTED_QUORUM_SET;
```
MonsieurNicolas: probably an internal bug

src/scp/Slot.cpp
```diff
@@ -127,13 +128,14 @@ Slot::processEnvelope(SCPEnvelope const& envelope, bool self)
...
     }
-    catch (...)
+    catch (std::exception& e)
     {
-        auto info = getJsonInfo();
-        CLOG(ERROR, "SCP") << "Exception in processEnvelope "
-                           << "state: " << info.toStyledString()
-                           << " processing envelope: "
-                           << mSCP.envToStr(envelope);
+        CLOG(FATAL, "SCP") << "SCP context:";
```
MonsieurNicolas: I don't think you need this change: as we rethrow, the exception string should already be logged upstream

src/catchup/ApplyLedgerChainWork.cpp
```diff
@@ -21,7 +24,7 @@ namespace stellar
 ApplyLedgerChainWork::ApplyLedgerChainWork(
     Application& app, WorkParent& parent, TmpDir const& downloadDir,
     LedgerRange range, LedgerHeaderHistoryEntry& lastApplied)
-    : Work(app, parent, std::string("apply-ledger-chain"))
+    : Work(app, parent, std::string("apply-ledger-chain"), RETRY_NEVER)
```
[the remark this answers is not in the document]
vogel: That should be done long time ago.

src/scp/Slot.cpp
vogel: Do you mean whole change or just adding e.what() to the message?

MonsieurNicolas: this particular change moving `catch(...)` to `catch(std::exception)` in order to log `e.what()`

MonsieurNicolas:
better, I think we need to make more changes to the logging code in overlay as it's very verbose right now even in the normal case

vogel:
overlay logging should now be much more user-friendly

MonsieurNicolas:
looks good, should be rebased though

vogel:
rebased

MonsieurNicolas:
r+ 7344644f2eb28af7bae0e3a43dd9302b21b582f6

Issue #1854, opened 2018-12-06
Implement non protocol changes as described in CAP0005 (fee changes)


MonsieurNicolas:
This issue tracks the implementation of changes required to implement [CAP0005](https://github.com/stellar/stellar-protocol/blob/master/core/cap-0005.md) that do not change the current protocol.

For example, "replace by fee" in the pending transaction queue.

PR #2011, opened 2019-03-16
Cap005 implementation

### LINKED ISSUE #1855: Implement CAP0005
This issue tracks the actual implementation of [CAP0005](https://github.com/stellar/stellar-protocol/blob/master/core/cap-0005.md)

MonsieurNicolas:
# Description

This PR implements [CAP0005](https://github.com/stellar/stellar-protocol/blob/master/core/cap-0005.md)

Resolves #1855 #1854 

There are a few differences from the original CAP005, documented in https://github.com/stellar/stellar-protocol/pull/272


src/herder/TxSetFrame.cpp
```diff
@@ -461,6 +461,32 @@ TxSetFrame::sizeOp() const
...
+    if (lh.ledgerVersion >= 11)
+    {
+        size_t ops = 0;
+        int64_t lowBaseFee = std::numeric_limits<int64_t>::max();
+        for (auto& txPtr : mTransactions)
+        {
+            auto txOps = txPtr->getEnvelope().tx.operations.size();
+            ops += txOps;
+            int64_t txBaseFee =
+                bigDivide(txPtr->getFeeBid(), 1, txOps, Rounding::ROUND_UP);
```
vogel: do we expect fee bids so big that bigDivide would be required to handle them?

MonsieurNicolas: ah yes, we just need a round_up int64 divide here.

src/herder/TxSetFrame.cpp
```diff
@@ -96,38 +98,44 @@ SeqSorter(TransactionFramePtr const& tx1, TransactionFramePtr const& tx2)
...
 
-    vector<vector<TransactionFramePtr>> txBatches(4);
-    map<AccountID, size_t> accountTxCountMap;
-    retList = mTransactions;
-    // sort all the txs by seqnum
-    std::sort(retList.begin(), retList.end(), SeqSorter);
+    // build txBatches
+    // txBatches i-th element contains each i-th transaction for accounts with a
+    // transaction in the transaction set
+    std::list<std::deque<TransactionFramePtr>> txBatches;
```
vogel: my compiler complains about not having list member in std
possibly missing include?

src/herder/TxSetFrame.cpp
```diff
@@ -136,61 +144,142 @@ TxSetFrame::sortForApply()
...
+    for (auto& am : actTxQueueMap)
+    {
+        // sort each in sequence number order
+        std::sort(am.second.mTxs.begin(), am.second.mTxs.end(), SeqSorter);
+    }
+    return actTxQueueMap;
+}
+
 void
 TxSetFrame::surgePricingFilter(Application& app)
```
vogel: It would be very nice if you could make independent tests just for it, with interface similar to this:

std::vector<TransactionFramePtr> surgePricingFilter(std::vector<TransactionFramePtr> const& originalTxs, Header &header);

It could be just an overload called by this one.

src/herder/TxSetFrame.cpp
```diff
@@ -136,61 +144,142 @@ TxSetFrame::sortForApply()
...
-        // determine the fee ratio for each account
-        map<AccountID, double> accountFeeMap;
-        for (auto& tx : mTransactions)
+
+    bool maxIsOps = header.current().ledgerVersion >= 11;
+
+    size_t opsLeft;
+    {
+        auto maxTxSetSize = header.current().maxTxSetSize;
+        opsLeft = maxIsOps ? maxTxSetSize : (maxTxSetSize * 100);
```
vogel: 100 is magic constant that appears all over here. Please give it name.

src/herder/TxSetFrame.cpp
```diff
@@ -136,61 +144,142 @@ TxSetFrame::sortForApply()
...
+    }
+
+    auto curSizeOps = maxIsOps ? sizeOp() : (sizeTx() * 100);
+    size_t surgeOpsCutoff = 0;
+    if (opsLeft >= 100)
```
vogel: I don't think this is the right condition.

If validators set maxTxSetSize to 200 before upgrading version and current txset has exactly 200 transactions, we have:

```
maxIsOps = false;
opsLeft = 20000;
surgeOpsCutoff = 19900;
curSizeOps = 20000;
```

Which triggers surge pricing in effect, even if number of transaction is not too big.

I thing it should check maxIsOps and dont accept optLeft < 100 in that case.

src/herder/TxSetFrame.cpp
```diff
@@ -136,61 +144,142 @@ TxSetFrame::sortForApply()
...
+    {
+        surgeOpsCutoff = opsLeft - 100;
```
vogel: why are we doing that?
If max number of operations is set to 10000 and this txset has 10000 why do we remove at least 100 of them?

src/herder/TxSetFrame.cpp
```diff
@@ -136,61 +144,142 @@ TxSetFrame::sortForApply()
...
-    if (mTransactions.size() > max)
-    { // surge pricing in effect!
-        CLOG(WARNING, "Herder")
-            << "surge pricing in effect! " << mTransactions.size();
-
-        // determine the fee ratio for each account
-        map<AccountID, double> accountFeeMap;
-        for (auto& tx : mTransactions)
+
+    bool maxIsOps = header.current().ledgerVersion >= 11;
```
vogel: In general I would like to see the limit code itself extracted to some minimal utility struct, named i.e. OperationsLimit. It could look like that:
```
auto ol = OperationsLimit{headerVersion, headerMaxTxSize, sizeTx(), sizeOp());
if (ol.requireSurgePricing())
{
 ....
  if (ol.addOperations(numberOfOps))
  {
     include ops into txset
  }
  else
  {
     drop ops for this account
  }
}
```

That would be nicely testable.

src/herder/HerderSCPDriver.cpp
```diff
@@ -489,7 +489,45 @@ HerderSCPDriver::setupTimer(uint64_t slotIndex, int timerID,
...
+    auto rSize = r->size(header);
+    if (lSize < rSize)
+    {
+        return true;
+    }
+    else if (lSize > rSize)
+    {
+        return false;
+    }
+    if (header.ledgerVersion >= 11)
```
vogel: For me it is a bit confusing that there are two condition on ledger version here, one hidden (inside l->size and r->size), and one visible here. I think it would be clearer here to explicitly use sizeTx or sizeOp depending on ledgerVersion

src/herder/TxSetFrame.h
```diff
@@ -22,12 +25,21 @@ class TxSetFrame
...
 
-    bool
-    checkOrTrim(Application& app,
-                std::function<bool(TransactionFramePtr, SequenceNumber)>
-                    processInvalidTxLambda,
-                std::function<bool(std::vector<TransactionFramePtr> const&)>
-                    processLastInvalidTxLambda);
+    struct AccountTransactionQueue
+    {
+        AccountID mID;
```
vogel: this member is not used, to it could be just `using AccountTransactionQueue = deque`

src/herder/TxSetFrame.cpp
```diff
@@ -136,23 +137,51 @@ TxSetFrame::sortForApply()
...
+        if (v1 < v2)
+        {
+            return true;
+        }
+        else if (v1 > v2)
+        {
+            return false;
+        }
+        // use hash of transaction as a tie breaker
+        return lessThanXored(top1->getFullHash(), top2->getFullHash(), mSeed);
```
graydon: I couldn't entirely follow where the conversation with Jeremy wound up on this but I _think_ using a variant of `shortHash::computeHash` might offer a little more strength.

src/herder/TxSetFrame.cpp
```diff
@@ -96,38 +98,44 @@ SeqSorter(TransactionFramePtr const& tx1, TransactionFramePtr const& tx2)
...
+    vector<TransactionFramePtr> retList;
+    retList.reserve(mTransactions.size());
     for (auto& batch : txBatches)
     {
         // randomize each batch using the hash of the transaction set
         // as a way to randomize even more
         ApplyTxSorter s(getContentsHash());
         std::sort(batch.begin(), batch.end(), s);
-        for (auto tx : batch)
+        for (auto& tx : batch)
```
graydon: auto const&

src/herder/TxSetFrame.cpp
```diff
@@ -369,6 +454,64 @@ TxSetFrame::previousLedgerHash() const
     return mPreviousLedgerHash;
 }
 
+size_t
+TxSetFrame::size(LedgerHeader const& lh) const
```
graydon: I think this should take the _version_ as a parameter, not the ledgerHeader. At the call sites above, it reads as though it's somehow measuring the size of the ledgerHeader. I would also probably rename it, so like  `sizeInOpsOrTxs(uint32_t protocolVersion)`

src/herder/TxSetFrame.cpp
MonsieurNicolas: fixed

src/herder/TxSetFrame.cpp
MonsieurNicolas: we already test this method and there are no use cases for a different interface

src/herder/TxSetFrame.cpp
MonsieurNicolas: fixed

src/herder/TxSetFrame.cpp
MonsieurNicolas: it was the right logic but just ended up doing nothing as we have enough capacity anyways, so I removed this code

src/herder/TxSetFrame.cpp
MonsieurNicolas: it doesn't remove anything, limit is `opsLeft`, I removed this code

src/herder/TxSetFrame.cpp
MonsieurNicolas: We already have enough abstraction: limits are implemented with `surgeQueue` and tests would not really be improved.

src/herder/HerderSCPDriver.cpp
MonsieurNicolas: At this layer what we compare are "transaction set sizes", ops or txs is an implementation detail
That said, I am renaming `size` for clarity

src/herder/TxSetFrame.h
MonsieurNicolas: good catch

src/herder/TxSetFrame.cpp
```diff
@@ -177,26 +205,61 @@ TxSetFrame::surgePricingFilter(Application& app)
...
+                it = r.first;
+            }
+            it->second.mTxs.emplace_back(tx);
         }
 
-        // sort tx by amount of fee they have paid
-        // remove the bottom that aren't paying enough
-        std::vector<TransactionFramePtr> tempList = mTransactions;
-        std::sort(tempList.begin(), tempList.end(), SurgeSorter(accountFeeMap));
+        Hash seed;
```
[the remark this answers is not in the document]
MonsieurNicolas: fixed

src/herder/TxSetFrame.cpp
```diff
@@ -136,23 +137,51 @@ TxSetFrame::sortForApply()
...
-        if (tx1->getSourceID() == tx2->getSourceID())
-            return tx1->getSeqNum() < tx2->getSeqNum();
-        double fee1 = mAccountFeeMap[tx1->getSourceID()];
-        double fee2 = mAccountFeeMap[tx2->getSourceID()];
-        if (fee1 == fee2)
-            return tx1->getSourceID() < tx2->getSourceID();
-        return fee1 > fee2;
+        if (tx1 == nullptr || tx1->mTxs.empty())
+        {
+            return !tx2->mTxs.empty();
```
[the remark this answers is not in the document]
MonsieurNicolas: fixed

src/herder/TxSetFrame.cpp
MonsieurNicolas: we're sorting `batch` so can't do that

MonsieurNicolas: done

src/herder/TxSetFrame.cpp
MonsieurNicolas: re: name, sound good.

For passing around an uint32, I normally try to avoid this as we lose type safety. I'll pick a better name.

MonsieurNicolas: Mmmm, actually I can't find a better name. I guess we're dealing with one of the hardest problems in software engineering: "naming".

Goal when I used `size` was to keep this `size` function as the generic "size" that should be used for protocol version agnostic code (and the standard library uses that name).

I don't think I want to name it by listing all possible ways it can be implemented now or in the future `sizeInOpsOrTxsOrPoneys`.

The reason this question first came up was that the method signature implies that it's computing the size of the `LedgerHeader`, but I am not sure this is the case:
* if I wanted such a function it should be called `sizeOf` or something like that
* if I want to be explicit, I could call it something like `getTxSetSize`, but this is redundant (class name is `txSet` already)
* `size` is the method name used to return the size of containers in the standard library (you don't do `std::size(mylist)` but `mylist.size()`)

What do you think @graydon ?

src/herder/TxSetFrame.cpp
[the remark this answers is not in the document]
MonsieurNicolas: ah yes good point

src/herder/TxSetFrame.cpp
```diff
@@ -136,23 +137,51 @@ TxSetFrame::sortForApply()
     return retList;
 }
 
-struct SurgeSorter
+struct AccountTransactionQueue
 {
-    map<AccountID, double>& mAccountFeeMap;
-    SurgeSorter(map<AccountID, double>& afm) : mAccountFeeMap(afm)
+    AccountID mID;
```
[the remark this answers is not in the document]
MonsieurNicolas: yeah I removed that

src/herder/TxSetFrame.cpp
```diff
@@ -161,14 +190,13 @@ TxSetFrame::surgePricingFilter(Application& app)
 {
     LedgerTxn ltx(app.getLedgerTxnRoot());
     auto header = ltx.loadHeader();
-    size_t max = header.current().maxTxSetSize;
-    if (mTransactions.size() > max)
-    { // surge pricing in effect!
+
+    size_t maxSize = header.current().maxTxSetSize;
```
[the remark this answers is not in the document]
MonsieurNicolas: I standardized on `size_t` as it's returned by calls like `mTransactions.size()`.
While it's implementation defined, as we require 64 bit (or better) arch, it will fit 32 bit values easily and we avoid casting all over the place.

src/herder/TxSetFrame.cpp
```diff
@@ -177,26 +205,61 @@ TxSetFrame::surgePricingFilter(Application& app)
...
+        std::vector<TransactionFramePtr> updatedSet;
+        updatedSet.reserve(mTransactions.size());
+        while (opsLeft > 0 && !surgeQueue.empty())
         {
-            removeTx(*iter);
+            auto cur = surgeQueue.top();
+            surgeQueue.pop();
+            // inspect the top candidate queue
+            auto& curTopTx = cur->mTxs.front();
+            if (curTopTx->getOperations().size() <= opsLeft)
```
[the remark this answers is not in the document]
MonsieurNicolas: fixed

src/herder/TxSetFrame.cpp
```diff
@@ -231,7 +235,7 @@ TxSetFrame::surgePricingFilter(Application& app)
             surgeQueue.push(&am.second);
         }
 
-        size_t opsLeft = maxSize * 100;
+        size_t opsLeft = maxIsOps ? maxSize : (maxSize * 100);
```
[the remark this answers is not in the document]
MonsieurNicolas: fixed

src/herder/HerderSCPDriver.cpp
```diff
@@ -569,7 +594,7 @@ HerderSCPDriver::combineCandidates(uint64_t slotIndex,
         }
     }
 
-    // take the txSet with the highest number of transactions,
+    // take the txSet with the highest number of transactions/operations,
```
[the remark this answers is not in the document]
MonsieurNicolas: fixed

src/herder/TxSetFrame.cpp
```diff
@@ -147,8 +147,8 @@ TxSetFrame::sortForApply()
 struct SurgeCompare
 {
     Hash mSeed;
-    LedgerTxnHeader const& mHeader;
-    SurgeCompare(Hash const& seed, LedgerTxnHeader const& header)
+    LedgerHeader const& mHeader;
```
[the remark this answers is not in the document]
MonsieurNicolas: fixed

src/util/numeric.cpp
```diff
@@ -8,6 +8,13 @@
 
 namespace stellar
 {
+
+int64_t
+roundedDivide(int64_t A, int64_t B, Rounding rounding)
```
jonjove: This function isn't safe. First, it doesn't protect against overflow or division-by-zero. Second, it isn't correct for negative `B` because `B < 0` implies
```
(A + B - 1) / B <= (A - 1) / B <= A/B
```
so it rounds down when it should round up in this case. As an explicit example, suppose `A = 1` and `B = -1`. Then
```
(A + B - 1) / B = (1 + (-1) - 1) / (-1) = (-1) / (-1) = 1
```
doesn't even have the correct sign. Given all these issues, let's just remove this function and use an appropriate `bigDivide` function instead.

src/herder/TxSetFrame.cpp
```diff
@@ -237,13 +248,13 @@ TxSetFrame::surgePricingFilter(Application& app)
             surgeQueue.pop();
             // inspect the top candidate queue
             auto& curTopTx = cur->front();
-            if (curTopTx->getOperations().size() <= opsLeft)
+            size_t opsCount = curTopTx->getEnvelope().tx.operations.size();
```
jonjove: Should this be `size_t opsCount = maxIsOps ? curTopTx->getEnvelope().tx.operations.size() : MAX_OPS_PER_TX`?

src/herder/TxSetFrame.cpp
```diff
@@ -196,25 +226,8 @@ TxSetFrame::surgePricingFilter(Application& app)
...
-            if (it == actTxQueueMap.end())
-            {
-                auto d = std::make_pair(id, AccountTransactionQueue{});
-                auto r = actTxQueueMap.insert(d);
-                it = r.first;
-            }
-            it->second.emplace_back(tx);
-        }
+
+        auto actTxQueueMap = buildAccountTxQueues();
```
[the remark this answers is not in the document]
MonsieurNicolas: not needed as I switched to using `getEnvelope().tx.operations` instead of `getOperations()`

src/herder/TxSetFrame.cpp
```diff
@@ -229,7 +229,8 @@ TxSetFrame::surgePricingFilter(Application& app)
 
         auto actTxQueueMap = buildAccountTxQueues();
 
-        SurgeCompare const surge(header);
+        auto headerCopy = header.current();
+        SurgeCompare const surge(headerCopy);
```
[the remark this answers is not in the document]
MonsieurNicolas: Because in general making copies in leaf functions is bad as at best it can cause performance issues (like if I need to call a different function from `surgePricingFilter`, I would share the same `headerCopy`)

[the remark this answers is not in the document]
jonjove: This argument sounds like a premature optimization to me. Let's start by writing code that's easy to use correctly (indeed, this interface was already used incorrectly in an earlier version of this PR, so it is not easy to use correctly today) and if we see a performance problem later then we can correct it.

src/util/numeric.cpp
MonsieurNicolas: yeah I was using the `bigDivide` function before and Rafal made the observation that it's not needed (so we don't need to deal with overflow/negative numbers).
I am reverting the change: it's not worth having this function that is too limited in the code base

src/herder/TxSetFrame.cpp
MonsieurNicolas: it ends up being equivalent as opsLeft is a multiple of `MAX_OPS_PER_TX`, but I agree the code is easier to understand by just computing the adjusted `opsCount` right away

jonjove:
r+ ba68033

MonsieurNicolas:
rekicking: there was an infra failure in travis and also a small change I forgot I had in there (fee amount set during validation to 0 instead of minfee)

MonsieurNicolas:
r+ afcbc958ae2422be49301bfad6a9bb266339f8cf

MonsieurNicolas:
@latobarita: retry

PR #2012, opened 2019-03-18
docs: un-indent table that old markdown renderer gets confused by

graydon:
Our website currently renders with markdown-tt/markdown-tt version 7.0 from 2016. This is before they fixed https://github.com/markdown-it/markdown-it/pull/224 which allowed indented tables. Meanwhile I added a doc that uses indented tables. Oops! I will un-indent the table in the meantime.

cc @bartekn in case you feel like bumping the markdown-tt version as well.


graydon:
r+ 59aa0c9

graydon:
@latobarita: retry

PR #2013, opened 2019-03-19
Random eviction cache

marta-lokhova:
An update to cache implementation (#1978 ) done almost entirely by @graydon (opening a new PR since the original commit in the PR got deconstructed and rebased). I basically added small tweaks, addressed all the comments and updated the tests (hence the Catch2 bump, so that we can use their new TEMPLATE_TEST_CASE)


lib/catch.hpp
```diff
@@ -1,6 +1,6 @@
 /*
- *  Catch v2.2.2
- *  Generated: 2018-04-06 12:05:03.186665
+ *  Catch v2.5.0
```
MonsieurNicolas: we should pick up 2.7.0 as it's the latest stable

src/main/test/LruCacheTests.cpp
```diff
@@ -1,132 +0,0 @@
...
-// under the Apache License, Version 2.0. See the COPYING file at the root
-// of this distribution or at http://www.apache.org/licenses/LICENSE-2.0
-
-#include "lib/catch.hpp"
-#include "lib/util/lrucache.hpp"
-
-namespace stellar
-{
-
-using IntCache = cache::lru_cache<int, int>;
```
MonsieurNicolas: why are those tests gone?

Builds/VisualStudio/stellar-core.vcxproj
```diff
@@ -745,6 +745,7 @@ exit /b 0
     <ClInclude Include="..\..\src\util\types.h" />
     <ClInclude Include="..\..\src\util\MetricResetter.h" />
     <ClInclude Include="..\..\src\util\XDRStream.h" />
+    <ClInclude Include="..\..\src\util\RandomEvictionCache.h" />
     <ClInclude Include="..\..\src\work\Work.h" />
```
MonsieurNicolas: also need to add `CacheTests.cpp` (next to `LRUCacheTests.cpp`)

src/main/test/LruCacheTests.cpp
MonsieurNicolas: should be moved to the util/test folder instead

marta-lokhova: Both Lru Cache and random eviction tests are covered by template tests in `CacheTests.cpp`. That file is already in `util/test`.

marta-lokhova:
Pushed some updates (Catch2 version bump, and VS files fix)  and rebased @MonsieurNicolas

src/main/test/LruCacheTests.cpp
MonsieurNicolas: fancy!

MonsieurNicolas:
r+ f8f236455de4944404e02a874ac7339e0bd7c7c9

marta-lokhova:
Looks like an actual failure when building with `clang`, going to investigate

src/util/RandomEvictionCache.h
```diff
@@ -0,0 +1,201 @@
...
+    }
+
+    template <typename F>
+    void
+    erase_if(F const& f)
+    {
+        for (size_t i = 0; i < mValuePtrs.size(); ++i)
+        {
+            map_value_type*& vp = mValuePtrs.at(i);
+            while (mValuePtrs.size() != 0 && f(vp->second.mValue))
```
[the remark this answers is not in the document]
jonjove: Also, this function would make it possible to transform the cache like `r.erase_if([](int& i) { ++i; return true; }`. Do we want to allow this?

src/util/RandomEvictionCache.h
```diff
@@ -0,0 +1,201 @@
...
+            return *p;
+        }
+        else
+        {
+            throw std::range_error("There is no such key in cache");
+        }
+    }
+
+    V*
+    maybeGet(K const& k)
```
jonjove: I'm not sure about this interface to the cache, I would probably feel more comfortable with something like `bool maybeGet(K const& k, V& v)`. The reason for this is that we often use caches of smart pointers where exists-in-cache-as-nullptr means it doesn't exist in the underlying data store (this is in `LedgerTxn` for example). This means that we'll have to write and review code like
```
auto v = maybeGet(k); // This should really say vptr here, but it's easy to forget
if (v && *v) // Easy to forget the second condition
{
// Do something that depends on *v being non-null 
}
```
where-as the other interface would be easier to use correctly with
```
V v;
if (maybeGet(k, v) && v) // Clear what each condition does, and that you really need both
{
// Do something that depends on v being non-null 
}
```

I'm not extremely determined about this change, but I do think it would be an improvement. What do @marta-lokhova and @graydon think?

marta-lokhova: It does seem like it'd be easy to make mistakes with the current state of things. In addition to what you said, we probably shouldn't give a raw pointer to the client in general, as we don't know when they will dereference it, and if the data it points to is still valid at that time.  

I'm wondering if we should remove `maybeGet` entirely. While `exists` + `get` is a little less efficient (two map lookups instead of one), it is also less error-prone. And hits/misses will still be updated properly, as `exists` will count misses if configured correctly.

src/util/RandomEvictionCache.h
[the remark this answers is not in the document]
marta-lokhova: Thanks for catching this, you are correct. The edge case you described actually happens in `"cache erase_if removes some nodes"` as well, which caused Travis to fail. 

Regarding the lambda argument, I think we can make it a bit more restrictive, and only allow pass-by-const-ref, something like `std::function<bool(V const&)>`

src/util/RandomEvictionCache.h
graydon: Agreed that the `maybeGet` with a bool and out-param is probably a nicer interface. Don't especially care if you want to eliminate it altogether, I just have a long history of people yelling at me when designing container APIs that make you do two lookups, but I agree it's probably not a big deal.

src/util/RandomEvictionCache.h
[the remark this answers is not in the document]
graydon: Nice catch.

src/util/RandomEvictionCache.h
```diff
@@ -0,0 +1,201 @@
...
+        map_value_type*& victim =
+            (vp1->second.mLastAccess < vp2->second.mLastAccess ? vp1 : vp2);
+        mValueMap.erase(victim->first);
+        std::swap(victim, mValuePtrs.back());
+        mValuePtrs.pop_back();
+        ++mCounters.mEvicts;
+    }
+
+  public:
+    explicit RandomEvictionCache(size_t maxSize) : mMaxSize(maxSize)
```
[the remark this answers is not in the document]
graydon: Hm, I think it's movable? The pointers point into hashtable buckets that will (I think) not be reallocated if the owning unordered_map moves..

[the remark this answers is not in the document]
marta-lokhova: Graydon, I *think* you are correct (references, pointers and iterators are still valid after the move). The standard is pretty vague on this, and there's currently a proposal to make the language clearer: http://www.open-std.org/JTC1/SC22/WG21/docs/lwg-active.html#2321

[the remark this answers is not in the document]
graydon: (I mean I also 100% don't mind making it non-movable just on principle. I think pretty much none of the stateful stuff in the application ever needs to be moved. Certainly not copied!)

marta-lokhova:
Thanks everyone for the reviews! Along with some logic/style fixes, I ended up making RandomEvictionCache non-copyable and non-movable, and simplifying the API by removing `maybeGet` (if we find it necessary in the future, I'm happy to re-visit this). For the ease of reviewing, I haven't squashed the last commit yet.

src/util/RandomEvictionCache.h
```diff
@@ -145,56 +150,51 @@ template <typename K, typename V> class RandomEvictionCache
...
+                // - `std::swap` does not throw
+                // - `std::unordered_map::pop_back` does not throw
                 mValueMap.erase(vp->first);
                 std::swap(vp, mValuePtrs.back());
                 mValuePtrs.pop_back();
             }
         }
     }
 
+    // `get` offers no exception safety guarantee. If it throws an exception,
```
jonjove: I think `get` has the strong exception safety guarantee. What led you to believe otherwise?

marta-lokhova: Originally, I was concerned with the assignment of `mLastAccess`. Looking closer, although the standard does not provide any guarantees on that front, since the type is `uint64_t`, and there's no type conversion, I _think_ it actually does not throw. This is now updated (and squashed)

MonsieurNicolas:
r+ ec29c565a06f6a4371a7d00cf5f1ebd7bf70037b

Issue #1764, opened 2018-08-22
hang when adopting/deleting buckets on ledger close


MonsieurNicolas:
# Issue type
bug

# Version found
9.2

# Your environment and setup
ubuntu 14.04 LTS

# Issue description
Looks like core's main thread sometimes hangs for a long time when adopting/deleting buckets (when closing a ledger)

## Expected result
It should take a few milliseconds to perform those operations.

## Supporting files

```
2018-08-22T16:20:34.812 GCM6Q [Bucket DEBUG] Adopting bucket file /data/buckets/tmp/bucket-31e56dc31107243b/tmp-bucket-75bf9a1ca6e0a3d3.xdr as /data/buckets/bucket-730cffe9873ef504b0d3f646e390b0d1465e413e8fa871729785edb0dca7b169.xdr
2018-08-22T16:20:34.813 GCM6Q [Bucket DEBUG] Adopting bucket file /data/buckets/tmp/bucket-31e56dc31107243b/tmp-bucket-8a103cbb60613895.xdr as /data/buckets/bucket-b3a0bbffb820ead7544806beb49f0c9c98e818538da070e34610fbcffc2a4a82.xdr
2018-08-22T16:20:34.813 GCM6Q [Bucket DEBUG] Adopting bucket file /data/buckets/tmp/bucket-31e56dc31107243b/tmp-bucket-e55f8059c8bb7097.xdr as /data/buckets/bucket-ffa4c2341f968bd08d64c5a192b12557d35a527675fb4b93838ed736ba52db63.xdr
2018-08-22T16:20:35.579 GCM6Q [Bucket DEBUG] Deleting bucket file /data/buckets/tmp/bucket-31e56dc31107243b/tmp-bucket-d9db4332ce684177.xdr that is redundant with existing bucket
2018-08-22T16:20:41.709 GCM6Q [Bucket DEBUG] Deleting bucket file /data/buckets/tmp/bucket-31e56dc31107243b/tmp-bucket-ded156d60d6f3f03.xdr that is redundant with existing bucket
2018-08-22T16:20:50.130 GCM6Q [Bucket DEBUG] Deleting bucket file /data/buckets/tmp/bucket-31e56dc31107243b/tmp-bucket-ef54cd7f993616f9.xdr that is redundant with existing bucket
[...]

PR #2014, opened 2019-03-19
Configurable count of worker threads

vogel:
# Description

Should help with #1764

Increase number of worker threads, make them configurable and lower priority. Should help with starting multiple merges at the same time.

Also replace deprecated asio API with new one.

# Checklist
- [x] Compiles
- [x] Ran all tests

### DIFF
--- a/src/main/ApplicationImpl.cpp
+++ b/src/main/ApplicationImpl.cpp
@@ -44,6 +44,7 @@
 #include "util/LogSlowExecution.h"
 #include "util/Logging.h"
 #include "util/StatusManager.h"
+#include "util/Thread.h"
 #include "util/TmpDir.h"
 #include "work/WorkManager.h"
 
@@ -91,9 +92,6 @@ ApplicationImpl::ApplicationImpl(VirtualClock& clock, Config const& cfg)
 
     mNetworkID = sha256(mConfig.NETWORK_PASSPHRASE);
 
-    unsigned t = std::thread::hardware_concurrency();
-    LOG(DEBUG) << "Application constructing "
-               << "(worker threads: " << t << ")";
     mStopSignals.async_wait([this](asio::error_code const& ec, int sig) {
         if (!ec)
         {
@@ -102,9 +100,16 @@ ApplicationImpl::ApplicationImpl(VirtualClock& clock, Config const& cfg)
         }
     });
 
+    unsigned t = std::thread::hardware_concurrency();
+    LOG(DEBUG) << "Application constructing "
+               << "(worker threads: " << t << ")";
     while (t--)
     {
-        mWorkerThreads.emplace_back([this, t]() { this->runWorkerThread(t); });
+        auto thread = std::thread{[this]() {
+            runCurrentThreadWithLowPriority();
+            mWorkerIOContext.run();
+        }};
+        mWorkerThreads.emplace_back(std::move(thread));
     }
 }
 
@@ -391,12 +396,6 @@ ApplicationImpl::start()
 
     while (!done && mVirtualClock.crank(true))
         ;
-}
-
-void
-ApplicationImpl::runWorkerThread(unsigned i)
-{
-    mWorkerIOContext.run();
 }
 
 void
--- a/src/main/ApplicationImpl.cpp
+++ b/src/main/ApplicationImpl.cpp
@@ -63,10 +63,10 @@ namespace stellar
 ApplicationImpl::ApplicationImpl(VirtualClock& clock, Config const& cfg)
     : mVirtualClock(clock)
     , mConfig(cfg)
-    , mWorkerIOService(std::thread::hardware_concurrency())
-    , mWork(std::make_unique<asio::io_service::work>(mWorkerIOService))
+    , mWorkerIOContext(std::thread::hardware_concurrency())
+    , mWork(std::make_unique<asio::io_context::work>(mWorkerIOContext))
     , mWorkerThreads()
-    , mStopSignals(clock.getIOService(), SIGINT)
+    , mStopSignals(clock.getIOContext(), SIGINT)
     , mStarted(false)
     , mStopping(false)
     , mStoppingTimer(*this)
@@ -290,7 +290,7 @@ ApplicationImpl::~ApplicationImpl()
         mProcessManager->shutdown();
     }
     reportCfgMetrics();
-    shutdownMainIOService();
+    shutdownMainIOContext();
     joinAllThreads();
     LOG(INFO) << "Application destroyed";
 }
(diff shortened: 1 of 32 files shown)


src/util/Thread.cpp
```diff
@@ -0,0 +1,57 @@
...
+#else
+
+void
+runWithLowPriority(std::thread& thread)
+{
+    auto native = thread.native_handle();
+    int policy;
+    sched_param param;
+
+    auto ret = pthread_getschedparam(native, &policy, &param);
```
MonsieurNicolas: I am not sure I understand why this is "low priority": according to https://linux.die.net/man/2/sched_setscheduler `SCHED_BATCH` doesn't change priority only the policy when switching tasks.

Wasn't `pthread_setschedprio` enough (and maybe more portable)?

src/main/ApplicationImpl.cpp
```diff
@@ -100,7 +100,7 @@ ApplicationImpl::ApplicationImpl(VirtualClock& clock, Config const& cfg)
         }
     });
 
-    unsigned t = std::thread::hardware_concurrency();
+    unsigned t = mConfig.WORKER_THREADS;
```
MonsieurNicolas: make this an `int` - that way we get rid of just `unsigned` that looks so "C"

src/main/Config.cpp
```diff
@@ -447,6 +448,11 @@ Config::load(std::string const& filename)
             {
                 COMMANDS = readStringArray(item);
             }
+            else if (item.first == "WORKER_THREADS")
+            {
+                WORKER_THREADS =
+                    static_cast<size_t>(readInt<int>(item, 1, 1000));
```
MonsieurNicolas: make `WORKER_THREADS` an `int`, that way we don't need the cast

src/main/Config.h
```diff
@@ -188,6 +188,9 @@ class Config : public std::enable_shared_from_this<Config>
     // totally insensitive to overloading.
     uint32_t MINIMUM_IDLE_PERCENT;
 
+    // thread-management config
+    size_t WORKER_THREADS;
```
MonsieurNicolas: forgot to document this in the example config

src/util/Thread.cpp
MonsieurNicolas: (portability here in context of other POSIX systems like BSD/MacOS)

vogel: According to http://man7.org/linux/man-pages/man7/sched.7.html:
- it is not possible to set priority != 0 for threads with policy of `SCHED_OTHER`, `SCHED_IDLE`, `SCHED_BATCH`
- `SCHED_OTHER` is default for threads on linux
- all other policies (`SCHED_DEADLINE`, `SCHED_RR`, `SCHED_FIFO`) are real-time policies, which have higher priority to all other threads, so priority value can only be used to further increase that priority, not to decrease it

So for changing thread priority we need to change both policy and priorty, it can be done by `pthread_setchedparam`.

Other way to reduce thread priority would be to `nice` it, but `nice` is per-process only according to POSIX. In Linux is can be made per-thread, but that is non portable.

So it only leaves us with two choices: `SCHED_BATCH` and `SCHED_IDLE`. `SCHED_IDLE` is described as being extremely low-priority, and `SCHED_BATCH` as suitable for long-running tasks. As our worker thread is mostly used for merging buckets, I've chosen `SCHED_BATCH` as appropriate value.

MonsieurNicolas: read http://www.icir.org/gregor/tools/pthread-scheduling.html (and https://developer.apple.com/library/archive/documentation/System/Conceptual/ManPages_iPhoneOS/man3/pthread_setschedparam.3.html for MacOS ).

From what I get, `SCHED_FIFO` or `SCHED_RR` are the only "portable" ones. Even though they are global scheduling policies (so maybe need root?), to use them in this context would be to set a lower priority (ie, something between `sched_get_priority_min` and whatever `sched_getparam` returns).

If you want to use Linux specific stuff, we can just use `nice` ( http://man7.org/linux/man-pages/man2/nice.2.html ) that works on threads like you noticed.

vogel: I've decided to use nice (this requires a bit of API changes, as it works on current thread).

Using FIFO or RR is not an option, because:
- it requires special permissions (by default root is needed)
- it only allows to increase priority (sched_getparam for main thread returns 0, sched_get_priority_min/max are 1-99 for FIFO and RR)

src/util/Thread.cpp
```diff
@@ -0,0 +1,46 @@
...
+}
+
+#else
+
+void
+runCurrentThreadWithLowPriority()
+{
+    constexpr auto const LOW_PRIORITY_NICE = 5;
+
+    auto newNice = nice(LOW_PRIORITY_NICE);
```
MonsieurNicolas: Check with somebody else, but I think this will not work for BSD (MacOS) and just renice the entire process.

Maybe MacOS doesn't allow to do "low priority threads" - there is some conflicting information on [their site](https://developer.apple.com/library/archive/documentation/Darwin/Conceptual/KernelProgramming/scheduler/scheduler.html#//apple_ref/doc/uid/TP30000905-CH211-BABFGDHA) where it seems threads at at least "normal priority" (but later, there is mention of tasks that can have `TASK_BACKGROUND_APPLICATION` (but I don't know if a task maps to a process, thread or something else).

Maybe we don't do anything on MacOS (but then we may be breaking FreeBSD as well in general)

vogel: I've moved nice to be only used on linux. Also added comment that maybe thread_policy_set with THREAD_PRECEDENCE_POLICY can be used to set low priority on MacOS (someone with MacOS must verify that) - it allows setting to IDLE_PRI, so probably some intermediate lower priority setting are also available.

And from the documentation I think that TASK_BACKGROUND_APPLICATION sets priority for whole process, not a thread.

MonsieurNicolas:
Next steps on this:
* squash commits
* open an issue to address FreeBSD/MacOS

and we'll merge

graydon:
So .. putting the utility threads in a lower priority class is totally possible, but the testsuite slows _way_ down; I'm not comfortable merging this without more performance investigation.

graydon:
Mm, there are a couple build failures on FreeBSD, will continue tomorrow. Meanwhile feel free to cherry-pick https://github.com/graydon/stellar-core/commit/efe54c54d401c4edaf48c5549bcd73ed4b6a8e3d from https://github.com/graydon/stellar-core/tree/configurable-count-of-worker-threads for the macOS part.

vogel:
updated, also merged Graydon's commit for MacOS

MonsieurNicolas:
r+ 5c3a9621283e934eab8db1baab998798a1e725ba

PR #2015, opened 2019-03-20
Remove help from CommandHandler, refer to docs instead.

### LINKED ISSUE #1589: dedupe documentation of commands between commands.md and CommandHandler
Right now the documentation for the various commands for the command handler are duplicated in both `CommandHandler.cpp` and `commands.md`

We should instead only maintain the one in `commands.md` and have CommandHandler return the content of that file, this can be done either:
* dynamically (ie open the file)
* statically (including the content of that file in a string)

Either way is fine, the dynamic way may require changes in other projects that package stellar-core, so it may be trickier to do right.
On the other hand, doing it statically will require more changes in the Makefile and Visual Studio projects.

sui77:
# Description

Resolves #1589 

May I suggest to simply remove the http command docs from CommandHandler and just refer to commands.md instead? I doubt that many people look that up in a browser and html console output is pointless anyways.

Also merged missing contents into the markdown file. I can do a separate PR for that if this radical solution is not appropriate.

# Checklist
- [X] Compiles
- [X] Ran all tests
- [X] If change impacts performance, include supporting evidence per the [performance document](https://github.com/stellar/stellar-core/blob/master/performance-eval/performance-eval.md)


vogel:
I like that!

MonsieurNicolas:
Thank you, good idea!

MonsieurNicolas:
r+ 063545412e983855326fb5d186c2406cd3fbc4a5

PR #2018, opened 2019-03-22
database: scrub passwords from connection strings with std::regex.

graydon:
# Description

This replaces the parser in DatabaseConnectionString.cpp with a call to `std::regex_replace`. The main purpose here is actually to avoid having it throw on startup everytime we start with `sqlite3://:memory:` but in the meantime I figured it looked like a regular grammar so why not simplify while I"m at it.

# Checklist
- [x] Compiles
- [x] Ran all tests

### DIFF
--- a/src/database/DatabaseConnectionString.cpp
+++ b/src/database/DatabaseConnectionString.cpp
@@ -4,160 +4,25 @@
 
 #include "DatabaseConnectionString.h"
 
-#include <stdexcept>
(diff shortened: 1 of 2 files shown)


graydon:
(cc @jonjove this is the startup-exception fix we were discussing)

MonsieurNicolas:
r+ f2069efa41168372ee47747f8b7350c90becb37a

PR #2020, opened 2019-03-22
Avoid empty bulk load operations

jonjove:
Discovered while reviewing #1988. The implementation of the bulk load operations (#1930) didn't optimize the case where we are attempting to bulk load no keys. This is fixed such that no SQL queries are executed in this case.


marta-lokhova:
I think this needs clang-format

jonjove:
Thanks for noticing @marta-lokhova, fixed.

marta-lokhova:
r+ 565dff0efe4b4484f274e33d49885fce3a5988bc

PR #2027, opened 2019-03-27
transactions: add big comment about offers and the order book.

graydon:
This adds a big explanatory comment about how the order book works and what the general ideas behind the limit and exact offers in the system are all about. I wrote it because the order book code is pretty opaque to me still, and I really couldn't make heads or tail of CAP-0006 without this level of pedantic self-explanation. Figured it might be useful to others.

I would _love_ a thorough read-through and correction of it if possible from @jonjove and/or @MonsieurNicolas because I'm only very very moderately convinced I got it right. I think the commentary hangs together in the sense that I think it's painting a logically-consistent explanation for things, but I am much less sure that it's _correct_ with respect to the code.

Apologies for the level of repetitiveness / redundancy of phrasing / pictures / etc. I know it's probably overkill for people familiar with the subsystem and ideas, but for me I really get completely lost every time I dig into it. Wanted a very explicit guide.

### DIFF
--- a/src/transactions/OfferExchange.h
+++ b/src/transactions/OfferExchange.h
@@ -8,6 +8,193 @@
 #include <functional>
 #include <vector>
 
+// This is a reference-and-orientation comment for people who, like myself, get
+// constantly mixed up when dealing with order-related code and concepts. If you
+// find reasoning about orders and markets and spreads totally straightforward,
+// skip this, nothing to see here!
+//
+// For background and diagram context on what the orderbook even _means_, take
+// a look at https://www.stellar.org/developers/guides/concepts/exchange.html
+//
+// We'll use an ASCII-art reproduction of the diagram in that document to
+// explain some aspects of the code. Here's a schematic of the order book
+// for a particular wheat/sheep pair:
+//
+//
+//                      Willing to sell more sheep at higher sale price
+//                --------------------------------------------------------->
+//
+//            ^  +----------------------------------------------------------+
+//     Higher |  |sell 60 sheep                                             |
+//     prices |  |@ 100 wheat:1                                             |
+//      for   |  +--------------------------------------+-------------------+
+//     sheep  |  |sell 40 sheep                         |
+//       =    |  |@ 75 wheat:1                          |
+//     better |  +-----------------------+--------------+
+//      for   |  |sell 25 sheep          |
+//     sheep  |  |@ 50 wheat:1           |
+//     seller |  +--------------+--------+
+//            |  |sell 16 sheep |                           "ask" for sheep,
+//            |  |@ 25 wheat:1  |                 <-------- "bid" for wheat
+//               +--------------+
+//                 ^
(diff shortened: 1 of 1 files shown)


MonsieurNicolas:
I'll take a look at this a bit later. Glancing at it, most should probably go in the "distributed exchange" docs https://github.com/stellar/docs/blob/master/guides/concepts/exchange.md as *any* user (not just stellar-core devs) should understand it.

You can leave the technical bits about where in the code things live in comments.

src/transactions/OfferExchange.h
```diff
@@ -8,6 +8,228 @@
...
+// So far so good? Ok, now let's talk about the more-detailed semantics of the
+// orders we support!
+//
+//
+// Limit orders, exact orders, and price selection
+// -----------------------------------------------
+//
+// Orders/Offers (see terminology note above) submitted to the system are
+// actually _limit_ offers, meaning that a new order will match against any
+// stored offer in he book offering the seller a _higher_ price for their merch
```
vogel: typo: in the book

src/transactions/OfferExchange.h
```diff
@@ -8,6 +8,228 @@
...
+// then the cheapest wheat at the top of the wheat-selling stack (that priced at
+// 20 wheat per sheep, i.e. paying the sheep-seller _more_) still matches the
+// new offer and sells, even though the sheep-seller was willing to sell sheep
+// for less. An offer for a higher price was already present, and the limit
+// order on the sheep sale only bounds the price below (and amount above). It
+// will take the higher price too, better for the seller!
+//
+// What price is the sale made at? The price of the offer stored in the book. In
+// fact, offers _in the book_ are not limit offers, they are _exact_ offers that
+// strictly cross _at the price written on them_, never anything else. In the
```
MonsieurNicolas: "never anything else" is actually not true, cap0003 describes this more, but rounding causes bias that can cause a better price (when the standing offer wins) or worst price (when the standing offer is taken entirely), it's also possible that the order book decides to just throw away parts of the order to avoid violating threshold (to avoid making more than 1% error on the price)

oleganza:
this patch is very illuminating, thanks.

src/transactions/OfferExchange.h
graydon: fixed.

src/transactions/OfferExchange.h
graydon: I think you mean CAP-004. Noted and added expanded section discussing.

graydon:
I split the first half (orientation) off to https://github.com/stellar/docs/pull/453 against the docs repo, and fixed the associated suggestions here. Anything else?

MonsieurNicolas:
just noticed that you have a libmedida change in there, probably not what you wanted to do

graydon:
@MonsieurNicolas whoops! nice catch thanks, reverted the medida change.

PR #2029, opened 2019-03-27
bump overlay to v9

MonsieurNicolas:
bumping overlay version ahead of protocol 11 bump


MonsieurNicolas:
r+ 26c98e0d1db68974283cc95f6176c841e4028516

PR #2033, opened 2019-04-04
Fix FreeBSD build.

graydon:
Absorbs recent medida update to fix FreeBSD build and change one ifdef of our own for the same purpose.


MonsieurNicolas:
r+ a4c3841ebcbb4409c2419094df20ed5a542da81c

PR #2035, opened 2019-04-05
Command line database schema upgrade

### LINKED ISSUE #1912: migrating the db without starting stellar-core
# Issue type
  feature

# Version found
  all

# Your environment and setup
  n/a

# Issue description
There's no mechanism to migrate the database without starting stellar-core.  I need to do the migrations with a different user from the user that stellar-core uses normally. Database upgrades happen in only 2 places: when you wipe the database and on startup:
* https://github.com/stellar/stellar-core/blob/51295ad8029f05e47ab0d879d7acd65e0293c2da/src/main/ApplicationImpl.cpp#L146
* https://github.com/stellar/stellar-core/blob/51295ad8029f05e47ab0d879d7acd65e0293c2da/src/main/ApplicationImpl.cpp#L321

For now the best workaround seems to be to set THRESHOLD_PERCENT in QUORUM_SET to a very large number and make stellar core crash on startup after doing the migration. Another strategy is to somehow read stellar's logs, waiting for it to finish starting and then kill it. I would really prefer to just have a command that lets me run the migrations by themselves.

## Actual result
Stellar core has to start in order to migrate the database.

## Expected result
I should be able to migrate the database independently of starting stellar core

vogel:
# Description

Resolves #1912 

New command is added: `stellar-core upgrade-db`. It upgraded database to current schema without running application.

# Checklist
- [x] Compiles
- [x] Ran all tests

### DIFF
--- a/docs/software/commands.md
+++ b/docs/software/commands.md
@@ -100,6 +100,8 @@ Command options can only by placed after command.
   * For example this will run just the tests tagged with `[tx]` using protocol
     versions 9 and 10 and stop after the first failure:
     `stellar-core --test -a --version 9 --version 10 "[tx]"`
+* **upgrade-db**: Upgrades local database to current schema version. This is
+  usually done automatically during stellar-core run or other command.
 * **version**: Print version info and then exit.
 * **write-quorum**: Print a quorum set graph from history.
 
--- a/src/database/Database.cpp
+++ b/src/database/Database.cpp
@@ -255,6 +255,7 @@ Database::upgradeToCurrentSchema()
         applySchemaUpgrade(vers);
         putSchemaVersion(vers);
     }
+    CLOG(INFO, "Database") << "DB schema is in current version";
     assert(vers == SCHEMA_VERSION);
 }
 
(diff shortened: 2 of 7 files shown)


MonsieurNicolas:
needs rebase @vogel

vogel:
Rebased

MonsieurNicolas:
Actually, after reviewing this change again, I don't think it makes sense to keep the separation between `initializeDatabase` and `application::create`:

with the change done as proposed here, we now need to perform a sequence that looks like `create -> initializeDB -> start`, this is error prone (to do it right we would need to keep track of a `mDbWasInitialized` bool) and increases the overall complexity in the code base as an application can be in more states.

The only reason is that you replaced `checkInitialized` by `initializeDB` (that now has side effects), yet `checkInitialized` was kinda bogus from the start anyways: there are many ways things can fail, so from a contract point of view it would have been better to have that function throw in the first place. Now, it's even more true.

My suggestion is to not expose `initializeDB` at all:
this should be done inside `initialize` and we should just throw if there is a problem.

That way, create gets simpler: something like
```c++
create(VirtualClock& clock, Config const& cfg, bool newDB = true)
    {
        auto ret = std::make_shared<T>(clock, cfg);
        ret->initialize(newDB);
        return ret;
    }
```

If we do it like this, we end up with not having to think much about which state the application is in (it's either created - including database setup), or started.

I do not think that all places that "handle" cases when `checkInitialized(app)` is `false` are correct anyways, so this should make the code base more consistent as well.
For example, looking at https://github.com/stellar/stellar-core/blob/a4e9f1e320529cb9cfc2cecf4c1e3f2fd6437043/src/main/ApplicationUtils.cpp#L73 , I don't see why `0` makes any sense as nothing happened.

vogel:
PR upgraded

MonsieurNicolas:
nice! much simpler

MonsieurNicolas:
r+ a8d3379dede470bbf5036f53ecf6d9fab5acb737

PR #2038, opened 2019-04-08
Fix txINTERNAL_ERROR in PaymentOp

### LINKED ISSUE #1949: PaymentOpFrame cannot handle PATH_PAYMENT_MALFORMED result
`PathPaymentOpFrame::doApply` can produce the operation result `PATH_PAYMENT_MALFORMED` but `PaymentOpFrame::doApply` does not handle this result. This leads to `txINTERNAL_ERROR`, instead of producing the corresponding operation result `PAYMENT_MALFORMED`. Two options to fix this:
- Don't return `PATH_PAYMENT_MALFORMED` during apply, returning more specific results instead that `PaymentOpFrame` can handle
- Handle `PATH_PAYMENT_MALFORMED` in `PaymentOpFrame`

jonjove:
Resolves #1949


MonsieurNicolas:
r+ ac50b94bea9dd7be9b77332eed3852d0d05de484

PR #2040, opened 2019-04-08
ledger: fix SQLite WAL checkpoint starvation.

graydon:
# Description

Previously (likely ever since the LedgerTxn code went in) the SQLite WAL file was being starved of passive auto-checkpoint opportunities (see https://sqlite.org/wal.html#ckpt for explanation on how it works) because the prepared statements were being held-open during commit. This is bad: the WAL file just grows endlessly. While it doesn't effect performance (there's an index of the live pages held in the shared-memory file), it _does_ accumulate unbounded dead pages, which eventually uses up all the space in the working directory and crashes your node.

This change fixes that. It'd be nice to detect this automatically in the future, will open a followup bug on it.

# Checklist
- [x] Compiles
- [x] Ran all tests

Re: performance, perf stat says `[createbench]` takes about 3% more time and instructions with this change, i.e. it's a little bit slower. When the WAL gets big, SQLite claims that failure-to-compact will actually cause it to go _slower_ (i.e. this change would make things _faster_), but either way I think we have to do this.

### DIFF
--- a/src/ledger/LedgerTxn.cpp
+++ b/src/ledger/LedgerTxn.cpp
@@ -1421,8 +1421,12 @@ LedgerTxnRoot::Impl::commitChild(EntryIterator iter, LedgerTxnConsistency cons)
                 (bool)iter ? LEDGER_ENTRY_BATCH_COMMIT_SIZE : 0;
             bulkApply(bleca, bufferThreshold, cons);
         }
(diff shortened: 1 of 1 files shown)


MonsieurNicolas:
r+ b1b87184cff76f90e5472955f9e0b7c516e74f33

PR #2041, opened 2019-04-08
configure: use -fno-omit-frame-pointer by default, as in production builds.

graydon:
# Description

This enables frame pointers by default. I like having them and I bet so do you! They make debugging and profiling substantially more reliable at lower cost: the tools don't have to go grovelling through hundreds of megs of DWARF to figure out where the frame is.

Perf-wise: adds 0.1% of instructions and sub-noise of time to benchmarks I run locally. Also it's already what we're building with for production .debs.

# Checklist
- [x] Compiles
- [x] Ran all tests

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -31,7 +31,7 @@ if test -z "${WFLAGS+set}"; then
    WFLAGS="$WFLAGS -Werror=unused-result"
 fi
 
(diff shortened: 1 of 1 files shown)


MonsieurNicolas:
r+ 7401e0b6ae61efcef71292732a43f541e43c84ea

PR #2043, opened 2019-04-09
Load generator improvements

marta-lokhova:
An attempt to reduce noise in tx submission in load generation -- instead of relying on submit step rate, check how many transactions need to be submitted per ledger, and adjust number of transactions per step based on the expected close time. Tested with sample load of 500 tx/s locally, and got 2365 txs/ledger mean and 199 std dev, which seems more stable than what we currently have.


graydon:
Perfect! Glad this works better!

graydon:
r+ 8f99821

marta-lokhova:
Thanks for the tip @graydon! I think, It is still possible to not submit enough txs per ledger (if lg isn't scheduled often enough), but at least this gets rid of huge tx bursts that we've observed previously.

PR #2044, opened 2019-04-10
Change offerID to int64_t

### LINKED ISSUE #1960: audit `unsigned` types in xdr
We've made those changes in the past, but we've missed a few places:
in a nutshell, we should not use unsigned types where we can as this causes problems in both SQL and languages that don't support unsigned integers.

Example:
`offerID` is `uint64`

Just like sequence number in the past, the change should be safe:
it may require some minor SDK changes, and the "on wire" format is compatible (as we don't have values that would overflow the signed range).

marta-lokhova:
Update offerID in XDR and the codebase to be signed 64-bit ints. Resolves #1960

### DIFF
--- a/src/invariant/LedgerEntryIsValid.cpp
+++ b/src/invariant/LedgerEntryIsValid.cpp
@@ -184,10 +184,9 @@ LedgerEntryIsValid::checkIsValid(TrustLineEntry const& tl, uint32 version) const
 std::string
 LedgerEntryIsValid::checkIsValid(OfferEntry const& oe, uint32 version) const
 {
-    if (oe.offerID > INT64_MAX)
+    if (oe.offerID <= 0)
     {
-        return fmt::format("Offer offerID ({}) exceeds limit ({})", oe.offerID,
-                           INT64_MAX);
+        return fmt::format("Offer offerID ({}) must be positive", oe.offerID);
     }
     if (!isAssetValid(oe.selling))
     {
--- a/src/ledger/LedgerTxnOfferSQL.cpp
+++ b/src/ledger/LedgerTxnOfferSQL.cpp
@@ -19,8 +19,8 @@ namespace stellar
 std::shared_ptr<LedgerEntry const>
 LedgerTxnRoot::Impl::loadOffer(LedgerKey const& key) const
 {
-    uint64_t offerID = key.offer().offerID;
-    if (offerID > INT64_MAX)
+    int64_t offerID = key.offer().offerID;
+    if (offerID < 0)
     {
         return nullptr;
     }
@@ -34,8 +34,7 @@ LedgerTxnRoot::Impl::loadOffer(LedgerKey const& key) const
     auto prep = mDatabase.getPreparedStatement(sql);
     auto& st = prep.statement();
     st.exchange(soci::use(actIDStrKey));
-    int64_t signedOfferID = unsignedToSigned(offerID);
-    st.exchange(soci::use(signedOfferID));
+    st.exchange(soci::use(offerID));
 
     std::vector<LedgerEntry> offers;
     {
@@ -43,9 +42,8 @@ LedgerTxnRoot::Impl::loadOffer(LedgerKey const& key) const
         offers = loadOffers(prep);
     }
 
-    return offers.size() == 0
-               ? nullptr
-               : std::make_shared<LedgerEntry const>(offers.front());
+    return offers.empty() ? nullptr
+                          : std::make_shared<LedgerEntry const>(offers.front());
 }
 
 std::vector<LedgerEntry>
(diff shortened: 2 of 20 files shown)


src/ledger/LedgerTxnOfferSQL.cpp
```diff
@@ -19,8 +19,8 @@ namespace stellar
 std::shared_ptr<LedgerEntry const>
 LedgerTxnRoot::Impl::loadOffer(LedgerKey const& key) const
 {
-    uint64_t offerID = key.offer().offerID;
-    if (offerID > INT64_MAX)
+    int64_t offerID = key.offer().offerID;
+    if (offerID < 0)
```
vogel: what about zero?

src/transactions/test/PathPaymentTests.cpp
```diff
@@ -3881,9 +3881,8 @@ TEST_CASE("pathpayment", "[tx][pathpayment]")
             // offer required to pass
             // - needs 8x of payment
             // for anti-arbitrage case
-            auto initialBalance =
-                2 * offerAmount; // we need twice as much money
-                                 // as in the
+            auto initialBalance = 2 * offerAmount; // we need twice as much
+                                                   // money as in the
             // offer because of Price{2, 1} that is
```
vogel: looks like clang-format cannot keep up here

src/invariant/LedgerEntryIsValid.cpp
```diff
@@ -184,10 +184,10 @@ LedgerEntryIsValid::checkIsValid(TrustLineEntry const& tl, uint32 version) const
 std::string
 LedgerEntryIsValid::checkIsValid(OfferEntry const& oe, uint32 version) const
 {
-    if (oe.offerID > INT64_MAX)
+    if (oe.offerID < 0)
```
jonjove: I think this should be `>= 0` as `offerID == 0` is not permitted in the ledger (in operations it indicates create a new offer).

src/ledger/LedgerTxnOfferSQL.cpp
```diff
@@ -648,9 +646,9 @@ class BulkLoadOffersOperation
         for (auto const& k : keys)
         {
             assert(k.type() == OFFER);
-            if (k.offer().offerID <= INT64_MAX)
+            if (k.offer().offerID > 0)
```
jonjove: This isn't consistent with line 23, which includes queries with `offerID == 0`.

src/xdr/Stellar-transaction.x
```diff
@@ -308,7 +308,7 @@ case MEMO_NONE:
 case MEMO_TEXT:
     string text<28>;
 case MEMO_ID:
-    uint64 id;
+    int64 id;
```
jonjove: Why is this safe? The memo is user-specified, so this could have been `UINT64_MAX`.

src/ledger/LedgerTxnOfferSQL.cpp
marta-lokhova: I'm mimicking whatever logic was here before.

src/ledger/LedgerTxnOfferSQL.cpp
marta-lokhova: Updated for consistency.

src/xdr/Stellar-transaction.x
marta-lokhova: Reverted this (i think you are right, though I saw some sdks using signed ints)

src/test/TestMarket.h
```diff
@@ -96,7 +96,7 @@ class TestMarket
              OfferState const& finishedState = OfferState::SAME);
 
     TestMarketOffer
-    updateOffer(TestAccount& account, uint64_t id, OfferState const& state,
+    updateOffer(TestAccount& account, int64_t id, OfferState const& state,
```
MonsieurNicolas: forgot to update `mLastAddedID` as well

src/transactions/test/OfferTests.cpp
```diff
@@ -2829,8 +2829,7 @@ TEST_CASE("create offer", "[tx][offers]")
                                        {usd, idr, Price{1, 1}, 1}),
                     ex_MANAGE_SELL_OFFER_NOT_FOUND);
                 REQUIRE_THROWS_AS(
-                    market.updateOffer(acc1,
-                                       static_cast<uint64_t>(INT64_MAX) + 1,
+                    market.updateOffer(acc1, INT64_MAX + 1,
```
MonsieurNicolas: `INT64_MAX + 1` overflows and `UINT16_MAX` to signed is UB

marta-lokhova:
r+ 25b0559c891075a2efdabb5243fe7f7a48387f13

PR #2045, opened 2019-04-10
Improve transitive quorum set logic, persist it into the database

MonsieurNicolas:
# Description

This PR makes a couple changes to how we deal with the transitive quorum:
* it simplified the quorum management code by maintaining a quorum set, attempting to expand it over time as quorum sets are discovered
    * this avoids recomputing the transitive set all the time
    * this also avoids merging the transitive set between slots
* it saves the mapping between nodes and quorum information in the database
   * this in turn allows to recover information about nodes, even if those nodes are down, or if the local node restarts


src/herder/test/QuorumTrackerTests.cpp
```diff
@@ -0,0 +1,138 @@
...
+                               lcl.header.scpValue.closeTime + i,
+                               emptyUpgradeSteps, 0};
+        auto v = xdr::xdr_to_opaque(sv);
+
+        return TxPair{v, txSet};
+    };
+
+    auto vv = makeValue(1);
+
+    auto checkInQuorum = [&](bool checkSelf, std::set<int> ids) {
```
vogel: checkSelf is always true

src/herder/HerderPersistenceImpl.cpp
```diff
@@ -92,6 +93,50 @@ HerderPersistenceImpl::saveSCPHistory(uint32_t seq,
...
+            continue;
+        }
+        auto qSetH = sha256(xdr::xdr_to_opaque(*p.second));
+        usedQSets.insert(std::make_pair(qSetH, p.second));
+
+        std::string nodeIDStrKey = KeyUtils::toStrKey(nodeID);
+        std::string qSetHHex(binToHex(qSetH));
+
+        auto prep = db.getPreparedStatement(
+            "UPDATE quoruminfo SET qsethash = :h WHERE nodeid = :id");
```
graydon: We're on a version of sql that supports upserts (INSERT ... ON CONFLICT SET ...) now, if you want to simplify this a bit.

src/herder/HerderPersistenceImpl.cpp
```diff
@@ -271,6 +271,60 @@ HerderPersistence::copySCPHistoryToStream(Database& db, soci::session& sess,
...
+    if (st.got_data())
+    {
+        auto h = hexToBin256(qsethHex);
+        res = make_optional<Hash>(std::move(h));
+    }
+    return res;
+}
+
+SCPQuorumSetPtr
+HerderPersistence::getQuorumSet(Database& db, soci::session& sess,
```
graydon: From a separate-commits-compile-on-their-own perspective, I think you are calling this function in the commit before this one that defines it. I don't care _too_ much but if you want to clean up / rebase-edit, might want to fix that.

src/herder/HerderPersistenceImpl.cpp
```diff
@@ -271,6 +271,60 @@ HerderPersistence::copySCPHistoryToStream(Database& db, soci::session& sess,
...
+                          soci::into(qset64), soci::use(qSetHashHex));
+
+    st.execute(true);
+
+    if (st.got_data())
+    {
+        std::vector<uint8_t> qSetBytes;
+        decoder::decode_b64(qset64, qSetBytes);
+
+        xdr::xdr_get g1(&qSetBytes.front(), &qSetBytes.back() + 1);
```
graydon: I thought xdr_{to,from}_opaque should handle this a little less-clunkily?

src/herder/PendingEnvelopes.cpp
```diff
@@ -179,6 +180,17 @@ PendingEnvelopes::isNodeInQuorum(NodeID const& node)
                         m->statement);
                     res = getQSet(h);
                 }
+                if (res == nullptr)
+                {
+                    // see if we had some information for that node
+                    auto& db = mApp.getDatabase();
```
graydon: This seems to me like it risks doing a SQL query on every SCP message (including those from strangers), even when they're returning null. I think we probably need a little more caching of the results here, both positive and negative.

src/herder/PendingEnvelopes.cpp
```diff
@@ -162,26 +161,30 @@ PendingEnvelopes::recvTxSet(Hash hash, TxSetFramePtr txset)
...
     {
-        // search through the known slots
-        SCP::TriBool r = mHerder.getSCP().isNodeInQuorum(node);
-
-        // consider a node in quorum if it's either in quorum
-        // or we don't know if it is (until we get further evidence)
-        res = (r != SCP::TB_FALSE);
-
-        mNodesInQuorum.put(node, res);
+        mQuorumTracker.rebuild([&](NodeID const& id) {
```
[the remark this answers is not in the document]
MonsieurNicolas: done

src/herder/QuorumTracker.h
```diff
@@ -0,0 +1,33 @@
...
+#include <unordered_map>
+
+namespace stellar
+{
+
+// helper class to help track the overall quorum
+class QuorumTracker : public NonMovableOrCopyable
+{
+    SCP& mSCP;
+    std::unordered_map<NodeID, SCPQuorumSetPtr> mQuorum;
```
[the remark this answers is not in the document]
MonsieurNicolas: done

src/herder/QuorumTracker.h
```diff
@@ -0,0 +1,33 @@
...
+
+  public:
+    QuorumTracker(SCP& scp);
+
+    // returns true if id is in quorum
+    bool isNodeInQuorum(NodeID const& id);
+
+    // attempts to expand quorum at node `id`
+    // returns true if expansion succeeded
```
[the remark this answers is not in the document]
MonsieurNicolas: done

src/herder/QuorumTracker.cpp
```diff
@@ -0,0 +1,71 @@
...
+
+bool
+QuorumTracker::isNodeInQuorum(NodeID const& id)
+{
+    auto it = mQuorum.find(id);
+    return it != mQuorum.end();
+}
+
+bool
+QuorumTracker::expand(NodeID const& id, SCPQuorumSetPtr qSet)
```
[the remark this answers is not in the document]
MonsieurNicolas: I am letting this up to the caller (added clarification to the contract of expand)

src/herder/QuorumTracker.cpp
```diff
@@ -0,0 +1,71 @@
...
+        if (it == mQuorum.end() || it->second == nullptr)
+        {
+            auto qSet = lookup(n);
+            if (qSet != nullptr)
+            {
+                LocalNode::forAllNodes(
+                    *qSet, [&](NodeID const& id) { backlog.insert(id); });
+            }
+            mQuorum[n] = qSet;
+        }
```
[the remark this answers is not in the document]
MonsieurNicolas: I want to return `false` in this case (which is already the case):
at this layer, it's possible that we would go from "this node has a quorum X" but we're told that from now on "node has quorum Y" (Y !=X and Y null or not, doesn't matter), I already deal with the case where X == Y explicitely.

Resolution for caller is to call `rebuild` when this happens

src/herder/test/QuorumTrackerTests.cpp
```diff
@@ -0,0 +1,138 @@
...
+
+TEST_CASE("quorum tracker", "[quorum][herder]")
+{
+    Config cfg(getTestConfig());
+
+    std::vector<SecretKey> otherKeys;
+    int const kKeysCount = 7;
+    for (int i = 0; i < kKeysCount; i++)
+    {
+        otherKeys.emplace_back(SecretKey::random());
```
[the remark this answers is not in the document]
MonsieurNicolas: ah yes, this didn't exist. fixing that

src/herder/PendingEnvelopes.h
```diff
@@ -137,5 +135,11 @@ class PendingEnvelopes
 
     TxSetFramePtr getTxSet(Hash const& hash);
     SCPQuorumSetPtr getQSet(Hash const& hash);
+
+    // returns true if we think that the node is in quorum
+    bool isNodeInQuorum(NodeID const& node);
```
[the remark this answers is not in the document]
MonsieurNicolas: done

src/herder/HerderPersistenceImpl.cpp
MonsieurNicolas: I am not too worried about this one yet, and sqlite doesn't support it

src/herder/HerderPersistenceImpl.cpp
MonsieurNicolas: fixed

src/herder/HerderPersistenceImpl.cpp
MonsieurNicolas: yes, this was from the old code
fixed

src/herder/test/QuorumTrackerTests.cpp
MonsieurNicolas: ah good catch, fixed

src/herder/PendingEnvelopes.cpp
MonsieurNicolas: no it won't:
rebuild is only happening when we update quorum information which only happens when we actually process an SCP message, and we only process SCP messages for nodes in the known transitive quorum.

graydon: ok, so only known people -- but is it still potentially an awful lot of messages?

src/herder/PendingEnvelopes.h
```diff
@@ -139,6 +139,8 @@ class PendingEnvelopes
     // returns true if we think that the node is in quorum
     bool isNodeInQuorum(NodeID const& node);
 
+    std::unordered_map<NodeID, SCPQuorumSetPtr> const& getCurrentQuorum() const;
```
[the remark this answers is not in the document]
MonsieurNicolas: done

MonsieurNicolas:
r+ 1dba743ada250e0fc7205ce4c64a56663fb6bd1b

MonsieurNicolas:
had to fix some tests that have some hacky setup: worked around it for now by adding a safeguard when we don't know the local node's quorum set

MonsieurNicolas:
r+ 24f14e154befc7d0bd7f0165e9a6012c79031657

MonsieurNicolas:
missed a few more tests that had bogus setup

MonsieurNicolas:
missed a few more tests that had bogus setup

MonsieurNicolas:
missed a few more tests that had bogus setup

MonsieurNicolas:
r+ 55d9b1838701fbb606dcdc1f05bc696fb10f63cd

MonsieurNicolas:
r+ 55d9b1838701fbb606dcdc1f05bc696fb10f63cd

MonsieurNicolas:
r+ 55d9b1838701fbb606dcdc1f05bc696fb10f63cd

MonsieurNicolas:
r+ 55d9b1838701fbb606dcdc1f05bc696fb10f63cd

PR #2046, opened 2019-04-10
Ban transactions for few ledgers

### LINKED ISSUE #2026: automatically "ban" transactions for a few ledgers
When a valid transaction is queued, it waits to be incorporated up to 3 ledgers.

At that point it's getting dropped from the "pending queue".

As this is done locally, it's possible that a different node forwards that same transaction back to the validator (repeating the cycle).

When surge pricing is in effect, this causes those transactions to stick around for some non deterministic time, which stops somebody from submitting a transaction with higher fee.

A simple mitigation is to add an automatic "ban" of hashes that we can use in this context.

This "ban" component would be configured to ban a particular hash for a few ledgers (like 10), and have "fast lookup" by hash (to tell if a hash is banned or not).

The tx subsystem would then:
* when a transaction is removed after 3 ledgers, its hash would be added to the ban list
* when a tx hash is banned
     * we would not add it to the pending list (and would not broadcast it)
     * we would return an error code to the "tx" endpoint that the transaction is banned (retry again later)

For now, I imagine that a simple set is good enough, we may look into more exotic ways of banning (bloom filters) later.

vogel:
# Description

Resolves #2026

This PR extracts new class - TransactionQueue - from Herder and adds capability to ban transaction for few ledgers. It means that any transaction that does not get into ledger during 4 closes (i.e. because of surge pricing and not sufficient fee) will not be able to enter the queue again for next 10 closes.

# Checklist
- [x] Compiles
- [x] Ran all tests

### DIFF
--- a/src/herder/HerderImpl.cpp
+++ b/src/herder/HerderImpl.cpp
@@ -744,8 +744,7 @@ HerderImpl::triggerNextLedger(uint32_t ledgerSeqToTrigger)
         }
     }
 
-    std::vector<TransactionFramePtr> removed;
-    proposedSet->trimInvalid(mApp, removed);
+    auto removed = proposedSet->trimInvalid(mApp);
     removeReceivedTxs(removed);
 
     proposedSet->surgePricingFilter(mApp);
--- a/src/herder/HerderImpl.cpp
+++ b/src/herder/HerderImpl.cpp
@@ -42,6 +42,8 @@ using namespace std;
 namespace stellar
 {
 
+constexpr auto const TRANSACTION_QUEUE_SIZE = 4;
+
 std::unique_ptr<Herder>
 Herder::create(Application& app)
 {
@@ -56,20 +58,11 @@ HerderImpl::SCPMetrics::SCPMetrics(Application& app)
           app.getMetrics().NewMeter({"scp", "envelope", "receive"}, "envelope"))
     , mCumulativeStatements(app.getMetrics().NewCounter(
           {"scp", "memory", "cumulative-statements"}))
-
-    , mHerderPendingTxs0(
-          app.getMetrics().NewCounter({"herder", "pending-txs", "age0"}))
-    , mHerderPendingTxs1(
-          app.getMetrics().NewCounter({"herder", "pending-txs", "age1"}))
-    , mHerderPendingTxs2(
-          app.getMetrics().NewCounter({"herder", "pending-txs", "age2"}))
-    , mHerderPendingTxs3(
-          app.getMetrics().NewCounter({"herder", "pending-txs", "age3"}))
 {
 }
 
 HerderImpl::HerderImpl(Application& app)
-    : mPendingTransactions(4)
+    : mTransactionQueue(app, TRANSACTION_QUEUE_SIZE)
     , mPendingEnvelopes(app, *this)
     , mHerderSCPDriver(app, *this, mUpgrades, mPendingEnvelopes)
     , mLastSlotSaved(0)
@@ -129,34 +122,6 @@ HerderImpl::bootstrap()
     ledgerClosed();
 }
 
-static uint64_t
-countTxs(HerderImpl::AccountTxMap const& acc)
-{
-    uint64_t sz = 0;
-    for (auto const& a : acc)
-    {
-        sz += a.second->mTransactions.size();
-    }
-    return sz;
-}
-
-static std::shared_ptr<HerderImpl::TxMap>
-findOrAdd(HerderImpl::AccountTxMap& acc, AccountID const& aid)
-{
-    std::shared_ptr<HerderImpl::TxMap> txmap = nullptr;
-    auto i = acc.find(aid);
-    if (i == acc.end())
-    {
-        txmap = std::make_shared<HerderImpl::TxMap>();
-        acc.insert(std::make_pair(aid, txmap));
-    }
-    else
-    {
-        txmap = i->second;
-    }
-    return txmap;
-}
-
 void
 HerderImpl::valueExternalized(uint64 slotIndex, StellarValue const& value)
 {
@@ -213,7 +178,7 @@ HerderImpl::valueExternalized(uint64 slotIndex, StellarValue const& value)
     mLedgerManager.valueExternalized(ledgerData);
 
     // perform cleanups
-    updatePendingTransactions(externalizedSet->mTransactions);
+    updateTransactionQueue(externalizedSet->mTransactions);
 
     // Evict slots that are outside of our ledger validity bracket
     if (slotIndex > MAX_SLOTS_TO_REMEMBER)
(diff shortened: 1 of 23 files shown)


src/herder/TransactionQueue.h
```diff
@@ -0,0 +1,76 @@
...
+
+    friend bool operator==(AccountTransactionsState const& x,
+                           AccountTransactionsState const& y);
+};
+
+enum class TransactionSubmitStatus
+{
+    PENDING = 0,
+    DUPLICATE,
+    ERROR,
```
MonsieurNicolas: those names are too generic and conflict with other defines (for example, `ERROR` is defined as `0` on my machine)

src/herder/TransactionQueue.cpp
```diff
@@ -0,0 +1,228 @@
...
+}
+
+void
+TransactionQueue::shift()
+{
+    mPendingTransactions.erase(--mPendingTransactions.end());
+    mPendingTransactions.emplace_front();
+}
+
+int
```
MonsieurNicolas: inconsistent type: int vs uint64

src/herder/TransactionQueue.cpp
```diff
@@ -0,0 +1,228 @@
...
+        }
+    }
+
+    return result;
+}
+
+bool
+operator==(AccountTransactionsState const& x, AccountTransactionsState const& y)
+{
+    if (x.mMaxSeq != y.mMaxSeq)
```
MonsieurNicolas: why not write this as `return !(x.mMaxSeq == x.mMaxSeq && x.mTotalFees == y.mTotalFees)` ?

src/herder/TransactionQueue.h
```diff
@@ -0,0 +1,76 @@
...
+    typedef std::unordered_map<AccountID, std::shared_ptr<TxMap>> AccountTxMap;
+
+    explicit TransactionQueue(Application& app, int pendingDepth);
+
+    TransactionSubmitStatus add(TransactionFramePtr tx);
+    // it is responsibility of the caller to always remove such sets of
+    // transactions that remaining ones have theis sequence numbers increasing
+    // by one
+    void remove(std::vector<TransactionFramePtr> const& txs);
+    void shift();
```
MonsieurNicolas: document what those methods do, `shift` is not super obvious

src/herder/TransactionQueue.h
```diff
@@ -0,0 +1,76 @@
...
+#include <memory>
+#include <unordered_map>
+#include <vector>
+
+namespace stellar
+{
+
+class Application;
+
+struct AccountTransactionsState
```
MonsieurNicolas: this type probably belongs under the `TransactionQueue` namespace as it doesn't make sense in the stellar namespace

Also, `AccountTransactionsState` doesn't really mean anything: it's not a state, I think it's more like `AccountQueueInformation` or something like that

src/herder/TransactionQueue.cpp
```diff
@@ -0,0 +1,228 @@
...
+        }
+    }
+
+    return state;
+}
+
+void
+TransactionQueue::shift()
+{
+    mPendingTransactions.erase(--mPendingTransactions.end());
```
MonsieurNicolas: use `pop_back` instead of `erase`

src/herder/TransactionQueue.cpp
```diff
@@ -187,6 +205,17 @@ TransactionQueue::getState(AccountID const& accountID) const
 void
 TransactionQueue::shift()
 {
+    mBannedTransactions.erase(--mBannedTransactions.end());
```
MonsieurNicolas: use pop_back here as well

src/herder/TransactionQueue.cpp
```diff
@@ -187,6 +205,17 @@ TransactionQueue::getState(AccountID const& accountID) const
...
+    mBannedTransactions.emplace_front();
+
+    for (auto const& map : mPendingTransactions.back())
+    {
+        for (auto const& toBan : map.second->mTransactions)
+        {
+            mBannedTransactions[0].insert(toBan.first);
```
MonsieurNicolas: move `mBannedTransactions[0]` outside the loops, probably just a reference to `mBannedTransactions.front()`

src/herder/TransactionQueue.cpp
```diff
@@ -0,0 +1,228 @@
...
+    {
+        sz += a.second->mTransactions.size();
+    }
+    return sz;
+}
+
+static std::shared_ptr<TransactionQueue::TxMap>
+findOrAdd(TransactionQueue::AccountTxMap& acc, AccountID const& aid)
+{
+    std::shared_ptr<TransactionQueue::TxMap> txmap = nullptr;
```
MonsieurNicolas: remove ` = nullptr`

src/herder/TransactionQueue.cpp
```diff
@@ -0,0 +1,228 @@
...
+
+void
+TransactionQueue::TxMap::addTx(TransactionFramePtr tx)
+{
+    auto const& h = tx->getFullHash();
+    if (mTransactions.find(h) != mTransactions.end())
+    {
+        return;
+    }
+    mTransactions.insert(std::make_pair(h, tx));
```
MonsieurNicolas: use `emplace` instead of `insert`

src/herder/TransactionQueue.cpp
```diff
@@ -0,0 +1,228 @@
...
+
+    auto state = getState(tx->getSourceID());
+    LedgerTxn ltx(mApp.getLedgerTxnRoot());
+    if (!tx->checkValid(ltx, state.mMaxSeq))
+    {
+        return TransactionSubmitStatus::ERROR;
+    }
+
+    auto sourceAccount = stellar::loadAccount(ltx, tx->getSourceID());
+    if (getAvailableBalance(ltx.loadHeader(), sourceAccount) <
```
MonsieurNicolas: This may overflow, replace with 
```
if (getAvailableBalance(ltx.loadHeader(), sourceAccount)-tx->getFeeBid() <
        state.mTotalFees)
```

MonsieurNicolas: (I think with this change, the rest of the code is fine as this is the only place where we add transactions)

src/herder/TransactionQueue.cpp
```diff
@@ -0,0 +1,228 @@
...
+                           }
+
+                           auto& transactions = txMap->second->mTransactions;
+                           return transactions.find(tx->getFullHash()) !=
+                                  std::end(transactions);
+                       });
+}
+
+AccountTransactionsState
+TransactionQueue::getState(AccountID const& accountID) const
```
MonsieurNicolas: same here: "State" doesn't mean anything

src/herder/TransactionQueue.h
```diff
@@ -0,0 +1,76 @@
...
+        AccountTransactionsState mCurrentState;
+        std::unordered_map<Hash, TransactionFramePtr> mTransactions;
+        void addTx(TransactionFramePtr);
+        void recalculate();
+    };
+    typedef std::unordered_map<AccountID, std::shared_ptr<TxMap>> AccountTxMap;
+
+    explicit TransactionQueue(Application& app, int pendingDepth);
+
+    TransactionSubmitStatus add(TransactionFramePtr tx);
```
MonsieurNicolas: this should be called `tryAdd` instead

src/herder/test/TransactionQueueTests.cpp
```diff
@@ -0,0 +1,389 @@
...
+
+        REQUIRE(txSet->sortForApply() == expectedTxSet.sortForApply());
+    }
+
+  private:
+    TransactionQueue mTransactionQueue;
+};
+}
+
+TEST_CASE("TransactionQueue", "[herder][TransactionQueue]")
```
MonsieurNicolas: thanks! We didn't have tests for this at all

src/herder/TransactionQueue.h
MonsieurNicolas: You can just name them `STATUS_PENDING`, `STATUS_DUPLICATE`, etc

src/herder/TransactionQueue.cpp
```diff
@@ -0,0 +1,228 @@
...
+
+static std::shared_ptr<TransactionQueue::TxMap>
+findOrAdd(TransactionQueue::AccountTxMap& acc, AccountID const& aid)
+{
+    std::shared_ptr<TransactionQueue::TxMap> txmap = nullptr;
+    auto i = acc.find(aid);
+    if (i == acc.end())
+    {
+        txmap = std::make_shared<TransactionQueue::TxMap>();
+        acc.insert(std::make_pair(aid, txmap));
```
MonsieurNicolas: use `emplace` not `insert`

src/herder/HerderImpl.cpp
```diff
@@ -69,7 +69,7 @@ HerderImpl::SCPMetrics::SCPMetrics(Application& app)
 }
 
 HerderImpl::HerderImpl(Application& app)
-    : mTransactionQueue(app, 4)
+    : mTransactionQueue(app, 4, 10)
```
MonsieurNicolas: time to use constants instead of magic numbers

src/herder/TransactionQueue.h
```diff
@@ -65,11 +67,14 @@ class TransactionQueue
...
 
     int countPending(int index) const;
+    int countBanned(int index) const;
+    bool isBanned(Hash hash) const;
     std::shared_ptr<TxSetFrame> toTxSet(Hash lclHash) const;
 
   private:
     Application& mApp;
     std::deque<AccountTxMap> mPendingTransactions;
+    std::deque<std::set<Hash>> mBannedTransactions;
```
MonsieurNicolas: I think you want to use `unordered_set` here

src/herder/TransactionQueue.h
```diff
@@ -65,11 +67,14 @@ class TransactionQueue
     AccountTransactionsState getState(AccountID const& accountID) const;
 
     int countPending(int index) const;
+    int countBanned(int index) const;
+    bool isBanned(Hash hash) const;
```
MonsieurNicolas: use `Hash const&` instead of passing by value

src/herder/test/TransactionQueueTests.cpp
```diff
@@ -17,6 +17,13 @@ using namespace stellar::txtest;
 
 namespace
 {
+enum class TransactionState
+{
+    UNKNOWN,
```
MonsieurNicolas: same here, names are too short and have a high risk of conflicting with C style defines

src/herder/TransactionQueue.h
```diff
@@ -0,0 +1,76 @@
...
+    // it is responsibility of the caller to always remove such sets of
+    // transactions that remaining ones have theis sequence numbers increasing
+    // by one
+    void remove(std::vector<TransactionFramePtr> const& txs);
+    void shift();
+
+    AccountTransactionsState getState(AccountID const& accountID) const;
+
+    int countPending(int index) const;
+    std::shared_ptr<TxSetFrame> toTxSet(Hash lclHash) const;
```
MonsieurNicolas: use `Hash const&` instead of passing by value

vogel:
rebased and applied all of the comments

src/herder/TransactionQueue.cpp
MonsieurNicolas: you didn't fix this one: this emits a warning as size_t -> int is unsafe, just use static_cast

src/herder/TransactionQueue.h
```diff
@@ -0,0 +1,90 @@
...
+{
+class Counter;
+}
+
+namespace stellar
+{
+
+class Application;
+
+struct AccountTransactionsQueueState
```
MonsieurNicolas: Forgot to address those couple points:

this type probably belongs under the TransactionQueue namespace as it doesn't make sense in the stellar namespace

Also, AccountTransactionsState doesn't really mean anything: it's not a state, I think it's more like AccountQueueInformation or something like that

src/herder/TransactionQueue.cpp
```diff
@@ -0,0 +1,269 @@
...
+TransactionQueue::remove(std::vector<TransactionFramePtr> const& dropTxs)
+{
+    for (auto& m : mPendingTransactions)
+    {
+        if (m.empty())
+        {
+            continue;
+        }
+
+        std::set<std::shared_ptr<TxMap>> toRecalculate;
```
MonsieurNicolas: forgot to address the use of `std::set` - should be `std::unordered_set`

src/herder/TransactionQueue.cpp
```diff
@@ -0,0 +1,269 @@
...
+                           }
+
+                           auto& transactions = txMap->second->mTransactions;
+                           return transactions.find(tx->getFullHash()) !=
+                                  std::end(transactions);
+                       });
+}
+
+AccountTransactionsQueueState
+TransactionQueue::getAccountTransactionQueueState(
```
MonsieurNicolas: `AccountTransactionQueueState` is still implying a state, I was suggesting earlier `AccountTransactionQueueInfo` or `AccountTransactionQueueSummary`

MonsieurNicolas:
changes were merged

PR #2047, opened 2019-04-10
Remove autorate and stabilize tx per step calculation

marta-lokhova:


MonsieurNicolas:
r+ 9a2119606a2a0cfd43cb676d49401b43e69c279c

PR #2048, opened 2019-04-12
Remove crypto dependency from SCP

MonsieurNicolas:
This PR moves crypto responsibilities from SCP into the application.

This change:
* ~~allows to not sign messages that don't get sent (performance)~~ (turns out we cannot do this, so I removed this commit)
* makes it easier to use different crypto schemes (or no crypto at all for simulation)


MonsieurNicolas:
r+ 842024e822eed1333d04dd70c5af89ebf98471be

MonsieurNicolas:
r+ aaac85c01de6d800c240ad8f31c97c236dcbdb14

MonsieurNicolas:
r+ cab5be00a647beeac3d4644f5079598f6e921c5a

PR #2051, opened 2019-04-14
Scp signed values during nomination

MonsieurNicolas:
# Description

This PR adds support for additional data attached to values during the nomination round, specifically values during nomination now include the validator identity of the validator that introduced the value to the network.

This makes it easier to understand the flow of SCP messages over the network, can be used for auditing purpose in the future (like see if specific validators are not including certain transactions on purpose for example).

This PR also adds extra logging to makes it easier to diagnose problems


marta-lokhova:
r+ 8ee5418b5418dc617fd6df42ef60681682efe23d

MonsieurNicolas:
forgot to run clang-format on one commit

MonsieurNicolas:
r+ b0aa76706b8dccb0cded4369611202c8bc147cac

MonsieurNicolas:
@latobarita: retry

PR #2054, opened 2019-04-15
Ban transactions for few ledgers

### LINKED ISSUE #2026: automatically "ban" transactions for a few ledgers

MonsieurNicolas:
# Description

Resolves #2026 

fixes build issues with #2046

### DIFF
--- a/Builds/VisualStudio/stellar-core.vcxproj
+++ b/Builds/VisualStudio/stellar-core.vcxproj
@@ -287,9 +287,11 @@ exit /b 0
     <ClCompile Include="..\..\src\herder\HerderUtils.cpp" />
     <ClCompile Include="..\..\src\herder\LedgerCloseData.cpp" />
     <ClCompile Include="..\..\src\herder\PendingEnvelopes.cpp" />
+    <ClCompile Include="..\..\src\herder\TransactionQueue.cpp" />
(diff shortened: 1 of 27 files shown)


MonsieurNicolas:
r+ 015d8591c6fa993d1bc6d75d86bd80b6a53cac08

MonsieurNicolas:
r+ 87eaa78fe7843ab2fb3080d707d2452e660a977b

PR #2057, opened 2019-04-16
Add missing bucket files to Visual Studio project file

MonsieurNicolas:
looks like we've missed adding some files with the bucket PR


vogel:
r+ 9217e95861507d9c9b871326a196f60be3005aed

PR #2063, opened 2019-04-22
Work subsystem overhaul

### LINKED ISSUE #1755: design cleanup: Work interface is a mess

marta-lokhova:
Re-opening the work PR that resolves #1755 

It is now properly rebased; additionally, the work test suite has been updated to simulate more failure modes. 

I remaining TODO is to add VS files. 
Also, https://github.com/graydon/stellar-core/commit/a0ac8352f49ff489379e22fb411a765631eb2d5b needs to be added as well.


marta-lokhova:
@graydon took another pass and pushed a small update, so this is up-to-date now

graydon:
r+ f3c9ef4

PR #2077, opened 2019-04-27
Filter old scp messages

MonsieurNicolas:
This PR performs a few improvements on how SCP messages are ignored based on the network time (close time in the ledger header).

This is especially useful on testnet where the network gets reset often.

The way we make this better is by:
1. adding a new setting `MAXIMUM_LEDGER_CLOSETIME_DRIFT` that improves the detection of what the "latest ledger" is, based on consensus data. This setting is only used when the node is started fresh (after  `new-db` is called). The heuristic here is that the latest ledger has to be relatively recent to the local's node clock.
2. when a node receives SCP messages that have a bad "close time", we fast fail (instead of failing after feeding the message to the SCP subsystem). This avoids doing work that we don't need to do (downloading transaction sets, etc), if the value is going to be considered invalid anyways.
3. when starting fresh (genesis), we don't restore the "SCP state".
   * Restoring state is only useful when a node rejoins an existing network (where there is a good chance that peers have SCP messages that can help the node move on from its current ledger).
   * on the other hand, when starting from genesis, it's better to just consume SCP messages as soon as possible (there is no need to wait 30 seconds to determine that the network is not closing the genesis ledger)

### DIFF
--- a/src/herder/HerderImpl.cpp
+++ b/src/herder/HerderImpl.cpp
@@ -294,6 +294,72 @@ HerderImpl::recvTransaction(TransactionFramePtr tx)
     return result;
 }
 
+bool
+HerderImpl::checkMinCloseTime(SCPEnvelope const& envelope)
+{
+    using std::placeholders::_1;
+    auto const& st = envelope.statement;
+
+    uint64_t ctCutoff = VirtualClock::to_time_t(mApp.getClock().now());
+    if (ctCutoff >= mApp.getConfig().MAXIMUM_LEDGER_CLOSETIME_DRIFT)
+    {
+        ctCutoff -= mApp.getConfig().MAXIMUM_LEDGER_CLOSETIME_DRIFT;
+    }
+    else
+    {
+        ctCutoff = 0;
+    }
+
+    StellarValue sv;
+    // performs the most conservative check:
+    // returns true if one of the values is valid
+    auto checkCTHelper = [&](std::vector<Value> const& values) {
+        return std::any_of(values.begin(), values.end(), [&](Value const& e) {
+            auto r = getHerderSCPDriver().toStellarValue(e, sv);
+            return r && sv.closeTime >= ctCutoff;
+        });
+    };
+
+    bool b;
+
+    switch (st.pledges.type())
+    {
+    case SCP_ST_NOMINATE:
+        b = checkCTHelper(st.pledges.nominate().accepted) ||
+            checkCTHelper(st.pledges.nominate().votes);
+        break;
+    case SCP_ST_PREPARE:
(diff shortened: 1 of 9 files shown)


src/herder/HerderSCPDriver.h
```diff
@@ -138,6 +138,10 @@ class HerderSCPDriver : public SCPDriver
     // returns false on error
     bool toStellarValue(Value const& v, StellarValue& sv);
 
+    // validate close time as much as possible
+    bool checkCloseTime(uint64_t slotIndex, uint64_t lastCloseTime,
```
marta-lokhova: Commit order is a bit strange here: the renaming of this function happens in the next commit, although the updated name is used here. Also, this change is reverted in the next commit anyway, and the function remains private.

src/herder/HerderImpl.cpp
```diff
@@ -937,6 +937,12 @@ HerderImpl::restoreSCPState()
 {
     // setup a sufficient state that we can participate in consensus
     auto const& lcl = mLedgerManager.getLastClosedLedgerHeader();
+    if (lcl.header.ledgerSeq == LedgerManager::GENESIS_LEDGER_SEQ)
+    {
+        // if we're on genesis ledger, there is no point in claiming we're "in sync"
```
marta-lokhova: need to run clang-format

src/herder/HerderSCPDriver.cpp
```diff
@@ -675,6 +675,21 @@ HerderSCPDriver::combineCandidates(uint64_t slotIndex,
     return xdr::xdr_to_opaque(comp);
 }
 
+bool
+HerderSCPDriver::toStellarValue(Value const& v, StellarValue& sv)
+{
+    try
+    {
+
```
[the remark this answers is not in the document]
MonsieurNicolas: sorry what do you mean by extra newline?

src/herder/HerderSCPDriver.h
MonsieurNicolas: no: there is `checkCloseTime` in `HerderSCPDriver` (that I made public), the one that got renamed is in `HerderImpl`

src/herder/HerderSCPDriver.cpp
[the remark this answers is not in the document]
marta-lokhova: line 683 is an empty line, is it there by accident?

marta-lokhova:
r+ 1b64b2831031e5cadee20063115fc88abaecb6ae

marta-lokhova:
r+ b71b53cddd4fe98aaaa9950e929be8f5e7b92334

PR #2078, opened 2019-04-27
Fail work on bad files

### LINKED ISSUE #1676: stellar-core crashes after downloading bad ledger file from history
As reported in #1675 

version: v9.2.0

```
2018-04-27T08:48:27.385 GBT6G [History INFO] Catching up: verifying checkpoint 1/10385 (0%)
2018-04-27T08:48:27.385 GBT6G [Fs ERROR] failed to open XDR file: buckets/tmp/catchup-52d5d909f4b62698/ledger-0000003f.xdr, reason: 2 [XDRStream.h:49]
terminate called after throwing an instance of 'std::runtime_error'
  what():  failed to open XDR file: buckets/tmp/catchup-52d5d909f4b62698/ledger-0000003f.xdr, reason: 2
2018-04-27T08:48:27.971 GBT6G [default INFO] Starting stellar-core v9.2.0
```

While crashing may be appropriate in some cases, it doesn't seem to be one of those here.

It seems we're missing some basic checks in the code for when things go wrong.

We should harden the code to ensure that (not sure which one we're already doing):
* after downloading a file, we actually have a file on disk matching the name that we expect
* after unzipping a file, that we actually have a file on disk with the content
* do proper exception handling when opening/reading/writing files

marta-lokhova:
Resolves #1676


src/util/FileSystemException.cpp
```diff
@@ -0,0 +1,16 @@
+// Copyright 2019 Stellar Development Foundation and contributors. Licensed
+// under the Apache License, Version 2.0. See the COPYING file at the root
+// of this distribution or at http://www.apache.org/licenses/LICENSE-2.0
+
+#include "FileSystemException.h"
+
+namespace stellar
+{
+
+FileSystemException::FileSystemException(std::string const& msg)
```
MonsieurNicolas: why not just having all of this inline and avoid adding a new cpp that doesn't really bring anything

marta-lokhova: good idea, updated

src/catchup/ApplyLedgerChainWork.cpp
```diff
@@ -264,6 +265,11 @@ ApplyLedgerChainWork::onRun()
         CLOG(ERROR, "History") << "Replay failed";
         throw;
     }
+    catch (FileSystemException&)
+    {
+        CLOG(ERROR, "History") << POSSIBLY_CORRUPTED_LOCAL_FS;
```
MonsieurNicolas: I think you need to also log the reason `e.what()` when you catch those exceptions?

marta-lokhova: It looks like `XDRInputFileStream` already logs the message.

src/historywork/FetchRecentQsetsWork.cpp
```diff
@@ -74,7 +76,16 @@ FetchRecentQsetsWork::doWork()
...
         XDRInputFileStream in;
         FileTransferInfo fi(*mDownloadDir, HISTORY_FILE_TYPE_SCP, i);
-        in.open(fi.localPath_nogz());
+        try
+        {
+            in.open(fi.localPath_nogz());
+        }
+        catch (FileSystemException&)
+        {
+            CLOG(ERROR, "History") << POSSIBLY_CORRUPTED_LOCAL_FS;
```
MonsieurNicolas: this one doesn't log the reason

MonsieurNicolas:
@marta-lokhova you can squash & rebase, looks good

marta-lokhova:
@MonsieurNicolas updated

MonsieurNicolas:
r+ cc37e8698c8c3128cdf260d34e05128611fc6a37

PR #2083, opened 2019-04-30
Proper handling of logging partitions

### LINKED ISSUE #2082: Silent failure of unknown partitions in "ll" endpoint
Calling `ll?scp=DEBUG` doesn't actually do anything, and instead returns as-if the log level for "SCP" was set properly.

We should:
* make partitions case insensitive
* return a proper error when a partition is not found

MonsieurNicolas:
resolves #2082 

This PR makes setting logging partitions case insensitive (ie, 'll?partition=scp&level=DEBUG` will work), as well as returning a proper error when a bad partition name is passed in.


graydon:
r+ da97cfb

PR #2085, opened 2019-04-30
Fixes issues found by code analysis (code cleanup)

MonsieurNicolas:
This PR fixes a few warnings that static analyzers complain about:
* uninitialized variables
* passing large objects by value

### DIFF
--- a/src/history/InferredQuorumUtils.cpp
+++ b/src/history/InferredQuorumUtils.cpp
@@ -18,44 +18,47 @@ namespace stellar
 {
 
 void
-checkQuorumIntersection(Config cfg)
+checkQuorumIntersection(Config const& cfg)
 {
     VirtualClock clock;
-    cfg.setNoListen();
-    Application::pointer app = Application::create(clock, cfg, false);
+    Config cfg2(cfg);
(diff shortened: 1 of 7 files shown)


src/history/InferredQuorumUtils.cpp
```diff
@@ -18,52 +18,55 @@ namespace stellar
 {
 
 void
-checkQuorumIntersection(Config cfg)
+checkQuorumIntersection(Config const& cfg)
```
jonjove: I'm not sure I see the point of this change. Instead of doing pass-by-value we just are doing pass-by-const-reference then copy. I don't see how this is better, we haven't removed the copy.

MonsieurNicolas: yes: it's functionally equivalent as there is still a copy.
The difference is that we don't have to keep reviewing large pass-by-value calls in the code base as we have explicit copies (either during code review or like this one when we have a warning from static analyzers).

graydon:
r+ c4aa05e

MonsieurNicolas:
@latobarita: retry

PR #2086, opened 2019-05-02
Fix flaky connectivity test

MonsieurNicolas:
One test was randomly failing due to a race condition between peers connecting to a central peer (in a nutshell: they would not discover other peers reliably as the list of alternate peers was non deterministic)


marta-lokhova:
r+ b70e0e28d7f70d2788d077500e1c803906dd4239

PR #2089, opened 2019-05-03
Optimize cross offers

jonjove:
Preliminary attempt to optimize `convertWithOffers` by changing the interface of `LedgerTxn::loadBestOffer`.

### DIFF
--- a/src/database/Database.cpp
+++ b/src/database/Database.cpp
@@ -55,7 +55,7 @@ using namespace std;
 
 bool Database::gDriversRegistered = false;
 
-static unsigned long const SCHEMA_VERSION = 10;
+static unsigned long const SCHEMA_VERSION = 11;
 
 // These should always match our compiled version precisely, since we are
 // using a bundled version to get access to carray(). But in case someone
@@ -221,6 +221,11 @@ Database::applySchemaUpgrade(unsigned long vers)
     case 10:
         // add tracking table information
         mApp.getHerderPersistence().createQuorumTrackingTable(mSession);
+        break;
+    case 11:
+        mSession << "DROP INDEX IF EXISTS bestofferindex;";
+        mSession << "CREATE INDEX bestofferindex ON offers "
+                    "(sellingasset,buyingasset,price,offerid);";
         break;
     default:
         if (vers <= 6)
--- a/src/ledger/LedgerTxn.cpp
+++ b/src/ledger/LedgerTxn.cpp
@@ -1085,26 +1085,18 @@ LedgerTxn::Impl::loadAllOffers(LedgerTxn& self)
 LedgerTxnEntry
 LedgerTxn::loadBestOffer(Asset const& buying, Asset const& selling)
 {
-    return getImpl()->loadBestOffer(*this, buying, selling, nullptr);
-}
-
-LedgerTxnEntry
-LedgerTxn::loadBestOffer(Asset const& buying, Asset const& selling,
-                         LedgerKey const& previousBest)
-{
-    return getImpl()->loadBestOffer(*this, buying, selling, &previousBest);
+    return getImpl()->loadBestOffer(*this, buying, selling);
 }
 
 LedgerTxnEntry
 LedgerTxn::Impl::loadBestOffer(LedgerTxn& self, Asset const& buying,
-                               Asset const& selling,
-                               LedgerKey const* previousBest)
+                               Asset const& selling)
 {
     throwIfSealed();
     throwIfChild();
 
     std::unordered_set<LedgerKey> exclude;
-    auto le = getBestOffer(buying, selling, previousBest);
+    auto le = getBestOffer(buying, selling, nullptr);
     return le ? load(self, LedgerEntryKey(*le)) : LedgerTxnEntry();
 }
 
--- a/src/ledger/LedgerTxn.cpp
+++ b/src/ledger/LedgerTxn.cpp
@@ -1812,36 +1812,10 @@ operator==(BestOffersCacheKey const& other) const
 }
 
 size_t
-LedgerTxnRoot::Impl::BestOffersCacheKeyHash::hashAsset(Asset const& asset)
-{
-    size_t res = asset.type();
-    switch (asset.type())
-    {
-    case ASSET_TYPE_NATIVE:
-        break;
-    case ASSET_TYPE_CREDIT_ALPHANUM4:
-    {
-        auto& a4 = asset.alphaNum4();
-        res ^= shortHash::computeHash(ByteSlice(a4.issuer.ed25519().data(), 8));
-        res ^= a4.assetCode[0];
-        break;
-    }
-    case ASSET_TYPE_CREDIT_ALPHANUM12:
-    {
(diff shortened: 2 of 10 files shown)


src/transactions/OfferExchange.cpp
```diff
@@ -1106,10 +1106,24 @@ convertWithOffers(
...
+        if (offerTrail.empty())
+        {
+            wheatOffer = ltx.loadBestOffer(sheep, wheat);
+        }
+        else
+        {
+            LedgerKey key(OFFER);
+            key.offer().sellerID = offerTrail.back().sellerID;
+            key.offer().offerID = offerTrail.back().offerID;
+            wheatOffer = ltx.loadBestOffer(sheep, wheat, key);
```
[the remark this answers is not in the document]
MonsieurNicolas: Actually, I think the way the code was written before is a lot easier to understand.

I don't think we ever enumerate the order book without actually consuming offers.
If there is a way to do this (in older versions of the protocol), we should instead just deal with it in a different way (as in: if loadBestOffer returns the same offer than the last iteration, we have to stop iterating somehow); that way we can drop the support for this "exclude" parameter that is confusing.

src/ledger/LedgerTxn.cpp
```diff
@@ -991,18 +1080,26 @@ LedgerTxn::Impl::loadAllOffers(LedgerTxn& self)
...
-                               Asset const& selling)
+                               Asset const& selling,
+                               LedgerKey const* previousBest)
 {
     throwIfSealed();
     throwIfChild();
 
     std::unordered_set<LedgerKey> exclude;
-    auto le = getBestOffer(buying, selling, exclude);
+    auto le = getBestOffer(buying, selling, previousBest);
```
MonsieurNicolas: `exclude` is not used anymore

src/ledger/LedgerTxn.cpp
```diff
@@ -991,18 +1080,26 @@ LedgerTxn::Impl::loadAllOffers(LedgerTxn& self)
 LedgerTxnEntry
 LedgerTxn::loadBestOffer(Asset const& buying, Asset const& selling)
 {
-    return getImpl()->loadBestOffer(*this, buying, selling);
+    return getImpl()->loadBestOffer(*this, buying, selling, nullptr);
+}
+
+LedgerTxnEntry
+LedgerTxn::loadBestOffer(Asset const& buying, Asset const& selling,
+                         LedgerKey const& previousBest)
```
MonsieurNicolas: `previousBest` should just be a `LedgerKey const* previousBest` as you can deal with null

src/database/Database.cpp
```diff
@@ -222,6 +222,11 @@ Database::applySchemaUpgrade(unsigned long vers)
         // add tracking table information
         mApp.getHerderPersistence().createQuorumTrackingTable(mSession);
         break;
+    case 11:
+        mSession << "DROP INDEX IF EXISTS bestofferindex;";
```
MonsieurNicolas: why do we need this index? I thought previous benchmarks demonstrated that they were not needed

src/ledger/LedgerTxn.cpp
```diff
@@ -1543,7 +1543,6 @@ LedgerTxnRoot::Impl::commitChild(EntryIterator iter, LedgerTxnConsistency cons)
     // Clearing the cache does not throw
     mBestOffersCache.clear();
     mEntryCache.clear();
-    mPrefetchMetrics.clear();
```
MonsieurNicolas: this seems to belong to a different PR

src/ledger/LedgerTxn.cpp
```diff
@@ -1670,7 +1669,15 @@ LedgerTxnRoot::prefetch(std::unordered_set<LedgerKey> const& keys)
 uint32_t
 LedgerTxnRoot::Impl::prefetch(std::unordered_set<LedgerKey> const& keys)
 {
-    uint32_t total = 0;
+    size_t const maxSize = ENTRY_CACHE_FILL_RATIO * mEntryCache.maxSize();
```
MonsieurNicolas: this also belongs to a different PR (prefetch updates)

src/ledger/LedgerTxn.cpp
```diff
@@ -496,68 +550,189 @@ LedgerTxn::Impl::getAllOffers()
...
+{
+    return getImpl()->getBestOffer(buying, selling, exclude);
+}
+
+std::shared_ptr<LedgerEntry const>
+LedgerTxn::Impl::getBestOffer(Asset const& buying, Asset const& selling,
+                              LedgerKey const* exclude)
+{
+    auto const& gbos = mGetBestOffersState;
+    if (!gbos && (mChild || !mEntry.empty()))
```
MonsieurNicolas: Isn't this just `if (!gbos)` ? I don't think it's valid to call `getBestOffer` without calling `prepareGetBestOffer` ?

jonjove: Actually, the most efficient way to use `prepareGetBestOffer` and `getBestOffer` is to:
1. Call `prepareGetBestOffer` on a `LedgerTxn`
2. Open a child `LedgerTxn`
3. Call `loadBestOffer` on the child (_without calling_ `prepareGetBestOffer` on the child)
4. Do things other than `loadBestOffer`
4. Commit the child
5. Repeat steps 2-4 as necessary

Indeed, this is what we do in `convertWithOffers`.

Why should we not call `prepareGetBestOffer` on the child? Because `prepareGetBestOffer` always recursively resets the streams of all parents to the beginning, so this will lead to `O(n^2)` iterations when making `O(n)` calls to `prepareGetBestOffer` followed by `getBestOffer`.

Why is it safe to not call `prepareGetBestOffer` when `!mChild && mEntry.empty()`? Because in this case there are no offers in the `LedgerTxn` so the `GetBestOfferState` won't contain anything. Note this argument only works if `prepareGetBestOffer` was already called for the parent. Technically I don't even think that the `!mChild` part is required.

Despite the fact that this works, it is neither beautiful nor intuitive. Any ideas for making this part of the interface cleaner would be much appreciated.

MonsieurNicolas: Ah yes, mmm, this is annoying. At a minimum it would be nice to enforce that `prepareGetBestOffer` was called on one of the parents. Right now I don't see how we can do this though

src/transactions/test/CrossOffersBenchmark.cpp
```diff
@@ -0,0 +1,319 @@
...
+    std::vector<TransactionFramePtr> txs;
+    size_t nOps = 0;
+
+    uint64_t seqNum = getNextSeqNum(app, src);
+    for (auto iter = dest.cbegin(); iter != dest.cend();)
+    {
+        txs.emplace_back(
+            createAccountsTx(app, src, seqNum++, balance, iter, dest.cend()));
+        nOps += txs.back()->getEnvelope().tx.operations.size();
+        if (nOps + 100 > 5000)
```
vogel: two magic numbers here, could be replaced by constant and variable

jonjove:
Replaced by #2307.

PR #2090, opened 2019-05-03
Build then run tests in travis

MonsieurNicolas:
This PR separates the build stage from the stage that runs tests in travis.

While this takes a little longer to run the same tests, it allows to build and update `ccache` independently of running tests.

In master, if a changeset modifies too many things (or the cache gets deleted), we're likely to timeout as the total time compilation + running tests exceeds 50 minutes.

Information on "build stages" can be found https://docs.travis-ci.com/user/build-stages/

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -46,3 +46,16 @@ branches:
  only:
  - auto
 
+stages:
+  - compile
+  - test
+
+matrix:
+  fast_finish: true
+  include:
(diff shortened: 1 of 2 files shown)


graydon:
r+ eec402a

MonsieurNicolas:
tested directly against travis, this latest iteration performs as expected. I am merging it.

Issue #1721, opened 2018-06-28
Option to show full public keys in quorum output


06chaynes:
Adding a flag to show entire length of public keys in output of 'stellar-core -c quorum' command would be handy during troubleshooting.

PR #2091, opened 2019-05-03
bugfix: show fullkeys for delayed, disagree and missing validators

sui77:
Fixing full keys output that I missed out for [delayed, disagree, missing] validators in /quorum?fullkeys=true.

Relates to #1721 

# Checklist
- [x] Compiles
- [x] Ran all tests
- [X] If change impacts performance, include supporting evidence per the [performance document](https://github.com/stellar/stellar-core/blob/master/performance-eval/performance-eval.md)

### DIFF
--- a/src/scp/BallotProtocol.cpp
+++ b/src/scp/BallotProtocol.cpp
@@ -2036,7 +2036,7 @@ BallotProtocol::getJsonQuorumInfo(NodeID const& id, bool summary, bool fullKeys)
         {
             if (!summary)
             {
-                missing.append(mSlot.getSCPDriver().toShortString(n));
+                missing.append(mSlot.getSCPDriver().toStrKey(n, fullKeys));
             }
             n_missing++;
         }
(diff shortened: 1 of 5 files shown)


src/scp/BallotProtocol.cpp
```diff
@@ -2036,7 +2036,10 @@ BallotProtocol::getJsonQuorumInfo(NodeID const& id, bool summary, bool fullKeys)
         {
             if (!summary)
             {
-                missing.append(mSlot.getSCPDriver().toShortString(n));
+                std::string nodeID =
```
MonsieurNicolas: maybe it would be cleaner to just make it that the conversion function takes a bool:
```
nodeID = mSlot.getSCPDriver().toStrKey(n, fullKeys)
```

that way we avoid having the checks all over the place

sui77: done

graydon:
r+ 9356a10

PR #2092, opened 2019-05-04
Add transitive quorum option to `quorum` endpoint

MonsieurNicolas:
This PR adds a new option to the `quorum` endpoint that provides an easy to use summary of the transitive quorum that should make it easier to write tools that analyze the health of the overall quorum (the data is an interpretation of what can be found in the `scp` endpoint for the most part).

For a preview, see the [Overall quorum analysis](https://github.com/MonsieurNicolas/stellar-core/blob/quorumTransitive/docs/software/admin.md#overall-quorum-analysis) section in the admin guide (included in this pull request).

### DIFF
--- a/src/herder/HerderImpl.cpp
+++ b/src/herder/HerderImpl.cpp
@@ -892,6 +892,125 @@ HerderImpl::getJsonQuorumInfo(NodeID const& id, bool summary, bool fullKeys,
     return ret;
 }
 
+Json::Value
+HerderImpl::getJsonTransitiveQuorumInfo(NodeID const& rootID, bool summary,
+                                        bool fullKeys)
+{
+    Json::Value ret;
+    Json::Value& nodes = ret["nodes"];
+
+    auto& q = mPendingEnvelopes.getCurrentlyTrackedQuorum();
+
+    auto rootLatest = getSCP().getLatestMessage(rootID);
+    std::map<Value, int> knownValues;
+
+    // walk the quorum graph, starting at id
+    std::unordered_set<NodeID> visited;
+    std::vector<NodeID> next;
+    next.push_back(rootID);
+    visited.emplace(rootID);
+    int distance = 0;
+    int valGenID = 0;
+    while (!next.empty())
+    {
+        std::vector<NodeID> frontier(std::move(next));
+        next.clear();
+        std::sort(frontier.begin(), frontier.end());
+        for (auto const& id : frontier)
+        {
+            Json::Value cur;
+            valGenID++;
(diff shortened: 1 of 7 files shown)


src/herder/HerderImpl.cpp
```diff
@@ -892,6 +892,116 @@ HerderImpl::getJsonQuorumInfo(NodeID const& id, bool summary, bool fullKeys,
...
+    // walk the quorum graph, starting at id
+    std::unordered_set<NodeID> visited;
+    std::deque<NodeID> next;
+    next.push_back(rootID);
+    visited.emplace(rootID);
+    int distance = 0;
+    int valGenID = 0;
+    while (!next.empty())
+    {
+        std::deque<NodeID> backlog(std::move(next));
```
graydon: I would probably call this "frontier" rather than "backlog" and there's not much use in making it a full deque rather than a vector; you build it in one pass and consume it in another after sorting, with no interleaving of push and pop.

src/herder/HerderImpl.cpp
```diff
@@ -892,6 +892,116 @@ HerderImpl::getJsonQuorumInfo(NodeID const& id, bool summary, bool fullKeys,
...
+                        }
+                    });
+                }
+                auto latest = getSCP().getLatestMessage(id);
+                if (latest)
+                {
+                    auto vals = Slot::getStatementValues(latest->statement);
+                    // updates the `knownValues` map, and generate a unique ID
+                    // for the value (heuristic to group votes)
+                    int trackingValID = -1;
```
graydon: I feel like this use of a numeric tracking-value-ID is more confusing than it's worth: the user will have to remember that this isn't _really_ the value voted on, but a proxy for it. I think it would be better to emit the hash here. If it really helps you eyeball problems, maybe emit both.

src/herder/HerderImpl.cpp
MonsieurNicolas: good points, fixed

src/herder/HerderImpl.cpp
MonsieurNicolas: done

MonsieurNicolas:
@graydon addressed your comments (and rebased on top of latest master)

graydon:
r+ af45e75

MonsieurNicolas:
`Json` library doesn't understand `uint64`, had to cast it

MonsieurNicolas:
r+ 1690090e133d96e92b68fc1d0afb77a2dcad3b65

PR #2095, opened 2019-05-07
Couple download and application of checkpoints in a single Work

### LINKED ISSUE #1951: catchup: applying transactions should be done in the same work unit than downloading
I noticed that we download a *lot* of checkpoint files when catching up, and those add up to GBs of data.

We should change the "unit" of work to follow a similar pattern than what we're doing to apply buckets, unit should be something like:
{ download tx file, verify tx file, {wait for the right ledger}, apply tx file, delete tx file }

That way we throttle how many checkpoints are in flight (limited by number of concurrent jobs that can exist at the same time).

#1747 would benefit also from that.

marta-lokhova:
This PR changes how we replay ledgers. Instead of downloading then applying everything, this change replays history per checkpoint, cleaning up files after checkpoint application succeeded. I ended up letting the parent work (in this case, `DownloadApplyTxsWork`) control its children fully (i.e. impose conditions, and wake up children when they are ready to continue running).

I'm not married to this design, but I could not justify implementing a full DAG infrastructure, given that the only use case for "waiting" work currently (and anticipated in the future) is "sequence A-B-C is running, but some other work D needs to complete before B". I'd appreciate any feedback/ideas though! 

Also, still need to add comments on how to properly use updated WorkSequence, and add more tests.

This will also resolve #1951

### DIFF
--- a/src/catchup/ApplyCheckpointWork.cpp
+++ b/src/catchup/ApplyCheckpointWork.cpp
@@ -2,7 +2,7 @@
 // under the Apache License, Version 2.0. See the COPYING file at the root
 // of this distribution or at http://www.apache.org/licenses/LICENSE-2.0
 
-#include "catchup/ApplyLedgerChainWork.h"
+#include "catchup/ApplyCheckpointWork.h"
 #include "herder/LedgerCloseData.h"
 #include "history/FileTransferInfo.h"
 #include "history/HistoryManager.h"
@@ -22,13 +22,14 @@
 namespace stellar
 {
 
-ApplyLedgerChainWork::ApplyLedgerChainWork(
-    Application& app, TmpDir const& downloadDir, LedgerRange range,
-    LedgerHeaderHistoryEntry& lastApplied)
-    : BasicWork(app, "apply-ledger-chain", RETRY_NEVER)
+ApplyCheckpointWork::ApplyCheckpointWork(Application& app,
+                                         TmpDir const& downloadDir,
+                                         uint32_t checkpoint, uint32_t boundary)
+    : BasicWork(app, "apply-checkpoint-" + std::to_string(checkpoint),
+                RETRY_NEVER)
     , mDownloadDir(downloadDir)
-    , mRange(range)
-    , mLastApplied(lastApplied)
+    , mCheckpoint(checkpoint)
+    , mApplyUntil(boundary == 0 ? mCheckpoint : boundary)
     , mApplyLedgerSuccess(app.getMetrics().NewMeter(
           {"history", "apply-ledger-chain", "success"}, "event"))
     , mApplyLedgerFailure(app.getMetrics().NewMeter(
@@ -37,41 +38,32 @@ ApplyLedgerChainWork::ApplyLedgerChainWork(
 }
 
 std::string
-ApplyLedgerChainWork::getStatus() const
+ApplyCheckpointWork::getStatus() const
 {
-    if (getState() == State::WORK_RUNNING)
+    if (getState() == BasicWork::State::WORK_RUNNING)
     {
-        std::string task = "applying checkpoint";
-        return fmtProgress(mApp, task, mRange.mFirst, mRange.mLast, mCurrSeq);
+        auto lcl = mApp.getLedgerManager().getLastClosedLedgerNum();
+        return fmt::format("Last applied ledger: {}", lcl);
     }
     return BasicWork::getStatus();
 }
 
 void
-ApplyLedgerChainWork::onReset()
+ApplyCheckpointWork::onReset()
 {
-    auto& lm = mApp.getLedgerManager();
-    auto& hm = mApp.getHistoryManager();
-
-    CLOG(INFO, "History") << fmt::format(
-        "Applying transactions for ledgers {}, LCL is {}", mRange.toString(),
-        LedgerManager::ledgerAbbrev(lm.getLastClosedLedgerHeader()));
-
-    mLastApplied = lm.getLastClosedLedgerHeader();
-
-    mCurrSeq = hm.checkpointContainingLedger(mRange.mFirst);
     mHdrIn.close();
     mTxIn.close();
     mFilesOpen = false;
 }
 
 void
-ApplyLedgerChainWork::openCurrentInputFiles()
+ApplyCheckpointWork::openCurrentInputFiles()
 {
     mHdrIn.close();
     mTxIn.close();
-    FileTransferInfo hi(mDownloadDir, HISTORY_FILE_TYPE_LEDGER, mCurrSeq);
-    FileTransferInfo ti(mDownloadDir, HISTORY_FILE_TYPE_TRANSACTIONS, mCurrSeq);
+    FileTransferInfo hi(mDownloadDir, HISTORY_FILE_TYPE_LEDGER, mCheckpoint);
+    FileTransferInfo ti(mDownloadDir, HISTORY_FILE_TYPE_TRANSACTIONS,
+                        mCheckpoint);
     CLOG(DEBUG, "History") << "Replaying ledger headers from "
                            << hi.localPath_nogz();
     CLOG(DEBUG, "History") << "Replaying transactions from "
@@ -83,7 +75,7 @@ ApplyLedgerChainWork::openCurrentInputFiles()
 }
 
 TxSetFramePtr
-ApplyLedgerChainWork::getCurrentTxSet()
+ApplyCheckpointWork::getCurrentTxSet()
 {
     auto& lm = mApp.getLedgerManager();
     auto seq = lm.getLastClosedLedgerNum() + 1;
@@ -117,7 +109,7 @@ ApplyLedgerChainWork::getCurrentTxSet()
 }
 
 bool
-ApplyLedgerChainWork::applyHistoryOfSingleLedger()
+ApplyCheckpointWork::applyHistoryOfSingleLedger()
 {
     LedgerHeaderHistoryEntry hHeader;
     LedgerHeader& header = hHeader.header;
@@ -228,12 +220,11 @@ ApplyLedgerChainWork::applyHistoryOfSingleLedger()
     }
 
     mApplyLedgerSuccess.Mark();
-    mLastApplied = hHeader;
     return true;
 }
 
 BasicWork::State
-ApplyLedgerChainWork::onRun()
+ApplyCheckpointWork::onRun()
 {
     try
     {
@@ -242,21 +233,18 @@ ApplyLedgerChainWork::onRun()
             openCurrentInputFiles();
         }
 
-        if (!applyHistoryOfSingleLedger())
-        {
-            mCurrSeq += mApp.getHistoryManager().getCheckpointFrequency();
-            mFilesOpen = false;
-        }
-
-        mApp.getCatchupManager().logAndUpdateCatchupStatus(true);
-
+        auto result = applyHistoryOfSingleLedger();
         auto& lm = mApp.getLedgerManager();
-        auto const& lclHeader = lm.getLastClosedLedgerHeader();
-        if (lclHeader.header.ledgerSeq == mRange.mLast)
+        auto const& lcl = lm.getLastClosedLedgerNum();
+
+        if (result)
         {
-            return State::WORK_SUCCESS;
+            mApp.getCatchupManager().logAndUpdateCatchupStatus(true);
+            auto done = lcl == mApplyUntil || lcl == mCheckpoint;
+            return done ? State::WORK_SUCCESS : State::WORK_RUNNING;
         }
-        return State::WORK_RUNNING;
+
+        return lcl == mCheckpoint ? State::WORK_SUCCESS : State::WORK_FAILURE;
     }
     catch (InvariantDoesNotHold&)
     {
@@ -274,5 +262,16 @@ ApplyLedgerChainWork::onRun()
         CLOG(ERROR, "History") << "Replay failed: " << e.what();
         return State::WORK_FAILURE;
     }
+}
+
+void
+ApplyCheckpointWork::onSuccess()
+{
+    // Application completed successfully, safe to delete files
+    FileTransferInfo hi(mDownloadDir, HISTORY_FILE_TYPE_LEDGER, mCheckpoint);
+    FileTransferInfo ti(mDownloadDir, HISTORY_FILE_TYPE_TRANSACTIONS,
+                        mCheckpoint);
+    std::remove(hi.localPath_nogz().c_str());
+    std::remove(ti.localPath_nogz().c_str());
 }
 }
--- a/src/historywork/DownloadApplyTxsWork.cpp
+++ b/src/historywork/DownloadApplyTxsWork.cpp
@@ -1,0 +1,123 @@
+// Copyright 2019 Stellar Development Foundation and contributors. Licensed
+// under the Apache License, Version 2.0. See the COPYING file at the root
+// of this distribution or at http://www.apache.org/licenses/LICENSE-2.0
+
+#include "historywork/DownloadApplyTxsWork.h"
+#include "catchup/ApplyCheckpointWork.h"
+#include "history/FileTransferInfo.h"
+#include "history/HistoryManager.h"
+#include "historywork/GetAndUnzipRemoteFileWork.h"
+#include "ledger/LedgerManager.h"
+#include "work/WorkSequence.h"
+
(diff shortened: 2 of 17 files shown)


src/catchup/ApplyCheckpointWork.cpp
```diff
@@ -243,21 +222,18 @@ ApplyLedgerChainWork::onRun()
...
 
-        if (!applyHistoryOfSingleLedger())
-        {
-            mCurrSeq += mApp.getHistoryManager().getCheckpointFrequency();
-            mFilesOpen = false;
-        }
-
-        mApp.getCatchupManager().logAndUpdateCatchupStatus(true);
-
+        auto result = applyHistoryOfSingleLedger();
```
MonsieurNicolas: I do not understand the change:

* either we change `applyHistoryOfSingleLedger` to always throw on error. The only time it returns `false` is when it has a bad input file (EOF) and it looks like it would go on an infinite loop with the updated logic
* or we handle the `false` case like before, and force the file to be reopen, but it looks to me that this is useless (what do we expect to be different with the retry?)

src/catchup/ApplyCheckpointWork.cpp
```diff
@@ -276,4 +252,15 @@ ApplyLedgerChainWork::onRun()
         return State::WORK_FAILURE;
     }
 }
+
+void
+ApplyCheckpointWork::onSuccess()
+{
+    // Application completed successfully, safe to delete files
```
MonsieurNicolas: How are those files deleted on error?

src/work/WorkSequence.h
```diff
@@ -9,6 +9,8 @@
 namespace stellar
 {
 
+using ConditionFn = std::function<bool()>;
```
MonsieurNicolas: this belongs to the `WorkSequence` namespace

src/work/WorkSequence.cpp
```diff
@@ -29,8 +31,33 @@ WorkSequence::onRun()
...
+            {
+                assert(mConditions.size() == mSequenceOfWork.size());
+                auto i =
+                    std::distance(mSequenceOfWork.cbegin(), mNextInSequence);
+                conditionSatisfied = !mConditions.at(i) || mConditions.at(i)();
+            }
+        }
+        catch (std::exception&)
+        {
+            return State::WORK_FAILURE;
```
MonsieurNicolas: should be logging?

src/work/WorkSequence.cpp
```diff
@@ -29,8 +31,33 @@ WorkSequence::onRun()
...
     {
-        w->startWork(wakeSelfUpCallback());
-        mCurrentExecuting = w;
+        bool conditionSatisfied = true;
+        try
+        {
+            // No pre-condition, or condition satisfied
+            if (!mConditions.empty())
+            {
+                assert(mConditions.size() == mSequenceOfWork.size());
```
MonsieurNicolas: don't use assert here: the check should be done in the constructor and throw

src/work/WorkSequence.h
```diff
@@ -17,13 +19,16 @@ namespace stellar
...
     std::vector<std::shared_ptr<BasicWork>>::const_iterator mNextInSequence;
     std::shared_ptr<BasicWork> mCurrentExecuting;
 
   public:
-    WorkSequence(Application& app, std::string name,
-                 std::vector<std::shared_ptr<BasicWork>> sequence,
-                 size_t maxRetries = RETRY_A_FEW);
+    WorkSequence(
+        Application& app, std::string name,
+        std::vector<std::shared_ptr<BasicWork>> sequence,
```
MonsieurNicolas: it looks like you expect each sequence to have an optional precondition, the contract should instead take a vector of pairs and a comment should explain what they are for

src/historywork/DownloadApplyTxsWork.cpp
```diff
@@ -0,0 +1,132 @@
...
+
+BasicWork::State
+DownloadApplyTxsWork::doWork()
+{
+    // This will properly abort the remaining children
+    if (anyChildRaiseFailure())
+    {
+        for (auto const& w : mRunning)
+        {
+            w->wakeUp();
```
MonsieurNicolas: not sure I understand why you have to manage children directly: this should be handled by `Work` already?

src/work/WorkSequence.h
MonsieurNicolas: actually there is a better way: you should just create a `GatedWork` (that derives from `BasicWork`) that takes a condition and `BasicWork` as a parameter, and when you need a specific work to be gated by some arbitrary condition, you just "wrap it" using that `GatedWork`.

It waits for the condition to be met (sleeps for 1 second if not), and when met, it "unblocks" itself and invokes the `BasicWork`.

src/historywork/DownloadApplyTxsWork.cpp
```diff
@@ -0,0 +1,132 @@
...
+        }
+        return State::WORK_FAILURE;
+    }
+
+    auto prev = mRunning.begin();
+    for (auto childIt = mRunning.begin(); childIt != mRunning.end();)
```
MonsieurNicolas: I think this is more complicated than it should be.

I am pretty sure that you can use `BatchWork` to perform what you're trying to do:
you can yield `download-apply` sequences (gated by the the last yielded value if there was one)

src/work/BasicWork.h
```diff
@@ -143,6 +143,13 @@ class BasicWork : public std::enable_shared_from_this<BasicWork>,
     // up once finished.
     void startWork(std::function<void()> notificationCallback);
 
+    // A helper method that implementers can use if they plan to utilize
+    // WAITING state. This tells the work to return to RUNNING state, and
+    // propagate the notification up to the scheduler. An example use of
+    // this would be RunCommandWork: a timer is used to async_wait for a
+    // process to exit, with a call to `wakeUp` upon completion.
+    void wakeUp(std::function<void()> innerCallback = nullptr);
```
MonsieurNicolas: I don't think you need this change anymore

src/historywork/DownloadApplyTxsWork.cpp
marta-lokhova: Yes, this is redundant (and was removed in the last commit)

src/catchup/ApplyCheckpointWork.cpp
marta-lokhova: Catchup fails, and the whole tmp dir is removed via RAII (this is how these files are currently handled on both error and success)

src/catchup/ApplyCheckpointWork.cpp
marta-lokhova: Not sure what you are referring to: returning `false` does mean `EOF`, in which case work decides whether it succeeded or failed based on the expected LCL, see line 247 (otherwise, how would it pass any catchup test that applies multiple checkpoints?)

Also, why would the files need to be re-open in `onRun`? Any failure will invoke `onReset`, which already takes care of it.

There's a distinction here between throwing and returning false: the former means that something got unexpectedly corrupted, while the latter means that we've read everything in this file.

src/historywork/DownloadApplyTxsWork.cpp
marta-lokhova: I think what you are suggesting is to have children asynchronously wait, while my implementation makes parent responsible for waking up an appropriate child. I thought about the approach you suggested when first implementing it (in fact, initially DownloadApplyTxsWork was inheriting from BatchWork). But this approach seemed less efficient and safe. Since we're doing it in batches, there will be a bunch of works waiting for each other, and potentially waking up and going back to sleep many times before the condition is satisfied (for example, applying 33 checkpoints with Postgres takes ~1.5 min on my machine, meaning last work in the batch will need to wake up almost 100 times). On the other hand, if some checkpoints apply really fast, dependent work still needs to wait for a second, whereas it could have proceeded earlier.
As to safety, parent work should keep references to whichever works are running and clean it up when needed (note that the condition to when it is okay to delete a child differs from ones of Work and BatchWork - those just delete a child once it succeeds or fails. In our case, we can only delete a child when it succeeds _and_ its dependent work succeeds (as the condition function only captures a weak pointer, so work must be kept alive).
The considerations above made me think that having parent centrally take care of its gated works is simpler and safer. Let me know what you think!

src/catchup/ApplyCheckpointWork.cpp
```diff
@@ -22,57 +22,37 @@
 namespace stellar
 {
 
-ApplyLedgerChainWork::ApplyLedgerChainWork(
-    Application& app, TmpDir const& downloadDir, LedgerRange range,
-    LedgerHeaderHistoryEntry& lastApplied)
-    : BasicWork(app, "apply-ledger-chain", RETRY_NEVER)
+ApplyCheckpointWork::ApplyCheckpointWork(Application& app,
+                                         TmpDir const& downloadDir,
+                                         uint32_t checkpoint, LedgerRange range)
```
vogel: This look strange to me. Why `ApplyCheckpointWork` need to know the whole range of applied checkpoint? I'm assuming that this is the responsibility of `DownloadApplyTxsWork`.

It looks like there is new condition with mRange added: `auto done = lcl == mRange.last() || lcl == mCheckpoint;`. This is probably for applying the last checkpoint, which may not be applied in full in case of command line catchup with `catchup-to` not pointing to a checkpoint end, as each `ApplyCheckpointWork` gets exactly the same `mRange` from `DownloadApplyTxsWork`.

It seem like a little bit better API could be used here. I see two possibilities:
1. replace `LegerRange range` with `uint32_t boundary` that is equal to `range.last`
2. replace pair of `uint32_t checkpoint, LedgerRange range` with a new `LedgerRange range` that has both ends (`first` and `last`) in the same checkpoint (which is obviously the just removed `uint32_t checkpoint`).

src/historywork/DownloadApplyTxsWork.cpp
vogel: It looks more and more that true DAG will be needed at some time in a future. What do you think about implementing it now?

src/historywork/DownloadApplyTxsWork.cpp
```diff
@@ -0,0 +1,132 @@
...
+            return self->getState() == State::WORK_SUCCESS;
+        };
+        cond = {nullptr, predicate};
+    }
+
+    auto result = addWork<WorkSequence>("download-apply-" +
+                                            std::to_string(mCurrCheckpoint),
+                                        seq, cond, RETRY_NEVER);
+    mRunning.push_back(result);
+    mCurrCheckpoint += mApp.getHistoryManager().getCheckpointFrequency();
```
vogel: I think it could be renamed to `mCheckpointToQueue`

src/historywork/DownloadApplyTxsWork.cpp
marta-lokhova: I thought about it as well: as mentioned in the PR description, I did not find a good reason to introduce DAG at this time: we only have 1 use case for this, and I anticipate the new validation logic will have the same use case.

marta-lokhova:
Friendly ping on this one: I did make some changes locally, but I'd like to resolve any outstanding discussions first (which will possibly result in more changes)

src/historywork/DownloadApplyTxsWork.cpp
MonsieurNicolas: I was not suggesting changing dependencies, just reusing code:
right now `DownloadApplyTxsWork` is duplicating the logic in `BatchWork` for no particular good reason (throttling, cleaning up of "completed" work). I don't see why using `BatchWork` would cause delays/sleeps.

Reason maybe you could not make it work was because of the conditions?
I think you can use regular `shared_ptr` to the previous work instead of weak pointers as we're not introducing cycles. This allows to let `BatchWork` manage the list of work as it wants and leave your own logic do what it wants as well.

src/work/WorkSequence.h
marta-lokhova: (notes from online and offline discussions)
I think the GatedWork approach is cleaner, my only concern is that we might have a lot of work waiting and waking up, especially for checkpoints that take a while to apply. This is not the case for current download-apply work, which wakes up its waiting children. What I don't like about the "parent wakes its child" approach is that the child can't recover (succeed or fail) on its own - it has to be woken up by its parent.

src/historywork/DownloadApplyTxsWork.cpp
```diff
@@ -0,0 +1,123 @@
...
+    mCheckpointToQueue =
+        mApp.getHistoryManager().checkpointContainingLedger(mRange.mFirst);
+    mLastApplied = mApp.getLedgerManager().getLastClosedLedgerHeader();
+    mRunning.clear();
+}
+
+bool
+DownloadApplyTxsWork::hasNext()
+{
+    wakeWaitingWork();
```
MonsieurNicolas: it's kinda weird/ugly to have `hasNext` with side effects

src/historywork/DownloadApplyTxsWork.cpp
```diff
@@ -0,0 +1,123 @@
...
+        if (work->getState() != State::WORK_FAILURE)
+        {
+            // At this point, sequence download->apply is either:
+            // - Still downloading files, in which case `wakeUp` will have
+            // no effect, and it will simply proceed to applying when done
+            // since the condition has been met OR
+            // - Finished downloading, but is waiting for previous checkpoint
+            // to finish applying; in this case, wakeUp will put it back into
+            // RUNNING state and the sequence can proceed with applying
+            work->wakeUp();
```
MonsieurNicolas: Calling `wakeUp` directly seems to lead to unspecified behavior: for example, if you call `wakeUp` on `RunCommand`, it will respawn a process from what I see (as it expects to only be awaken when the process completes).

Reason is that I think it's actually wrong to call `wakeUp` on arbitrary `BasicWork`: the only mechanism we have for callbacks is to pass a callback to `startWork`, callback that gets invoked when the work wakes up as defined by itself.

I think that if we want to be able to wake up other work, we need to create a new type of work where the contract allows it to be awoken like that.

I think the closest to this would be what I mentioned before, `ConditionalWork` (that waits on some condition to become true before triggering some other work), we can then safely call `doWork` even if the condition is not met.
If we want to avoid the "busy wait" (that sleeps between checks against the condition), we have many options. For example, we can register all instances of `conditionalWork` with some object (like the scheduler), and wake up all conditional work when "something interesting" happens (could be a combination of timeout and explicit calls to `wakeupConditionalWork` when an instance of `DownloadApplyTxsWork` completes for example).
This would also avoid calling the `wake up all` from inside `DownloadApplyTxWork` that is a bit hacky.

src/historywork/DownloadApplyTxsWork.h
```diff
@@ -0,0 +1,48 @@
...
+    uint32_t mCheckpointToQueue;
+    mutable std::list<std::shared_ptr<BasicWork>> mRunning;
+
+    void wakeWaitingWork();
+
+  public:
+    DownloadApplyTxsWork(Application& app, TmpDir const& downloadDir,
+                         LedgerRange range,
+                         LedgerHeaderHistoryEntry& lastApplied);
+    virtual ~DownloadApplyTxsWork() = default;
```
MonsieurNicolas: I don't think you need to define the destructor like this: the base class already has a virtual destructor

src/historywork/DownloadApplyTxsWork.h
```diff
@@ -0,0 +1,48 @@
...
+    TmpDir const& mDownloadDir;
+    LedgerHeaderHistoryEntry& mLastApplied;
+    uint32_t mCheckpointToQueue;
+    mutable std::list<std::shared_ptr<BasicWork>> mRunning;
+
+    void wakeWaitingWork();
+
+  public:
+    DownloadApplyTxsWork(Application& app, TmpDir const& downloadDir,
+                         LedgerRange range,
```
MonsieurNicolas: `LedgerRange` should be passed by cont ref

src/work/BatchWork.h
```diff
@@ -38,7 +38,7 @@ class BatchWork : public Work
 
     // Implementers aren't allowed to edit batching mechanism,
     // but instead provide iteration methods.
-    virtual bool hasNext() const = 0;
+    virtual bool hasNext() = 0;
```
MonsieurNicolas: oh this is really ugly and error prone (`hasNext` with side effects)

marta-lokhova: Yes, that’s why I didn’t want this work to inherit from BatchWork: there is no right place to check children works.

PR #2096, opened 2019-05-08
herder: make non validating nodes behave more like regular nodes

MonsieurNicolas:
This PR allows certain cleanup operations to occur, and also to measure nomination metrics even from non validating nodes.

This gives (some) visibility on what SCP is actually doing during nomination even from a non validating node (the `scp.timing.nominated` metric is updated) .


graydon:
r+ 7121f15

PR #2098, opened 2019-05-08
Simplify catchup test, avoid redundant regeneration of history

marta-lokhova:
"Catchup manual" was testing the same set of configurations every time, instead of randomly selecting them.

### DIFF
--- a/src/history/test/HistoryTests.cpp
+++ b/src/history/test/HistoryTests.cpp
@@ -874,38 +874,24 @@ TEST_CASE("Catchup recent", "[history][catchuprecent][!hide]")
 TEST_CASE("Catchup manual", "[history][catchupmanual]")
 {
     CatchupSimulation catchupSimulation{};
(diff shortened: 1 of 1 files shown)


src/catchup/test/CatchupWorkTests.cpp
```diff
@@ -17,6 +17,11 @@ using namespace stellar;
 auto max = std::numeric_limits<uint32_t>::max();
 namespace stellar
 {
+
+std::random_device gRandomDevice;
```
graydon: We recently consolidated random_device / random_engine usage so that all tests would be deterministically seedable. Please use extern std::default_random_engine gRandomEngine defined in util/Math.h rather than establishing your own elsewhere.

graydon:
r+  3a5252f

marta-lokhova:
@graydon looks like the commit hash is wrong

graydon:
whoops, sorry!

graydon:
r+ 6012314

PR #2101, opened 2019-05-10
display known keys when displaying short keys

MonsieurNicolas:
9356a10fe7a535156323f51a314935d1b60984d3 introduced a bug where it forces the use of the default implementation from `SCPDriver` instead of using the proper override.


graydon:
r+ 6ce3c35

PR #2102, opened 2019-05-10
Introduce transaction result verification logic

### LINKED ISSUE #2064: Archive validation: implement tx result verification
NOTE: the issue depends on #2063

We should add a new work, that verifies transaction result files for a range of ledgers. This will come in handy later, as we ramp up archive monitoring.

marta-lokhova:
Resolves #2064


src/historywork/DownloadVerifyTxResultsWork.cpp
```diff
@@ -0,0 +1,92 @@
...
+          {"history", "verify-tx-results", "success"}, "event"))
+    , mVerifyTxResultsFailure(app.getMetrics().NewMeter(
+          {"history", "verify-tx-results", "failure"}, "event"))
+{
+}
+
+std::string
+DownloadVerifyTxResultsWork::getStatus() const
+{
+    if (!isDone() && !isAborting())
```
vogel: I see there are two kind of this condition used for exactly the same thing in code:

1. `if (getState() == BasicWork::State::WORK_RUNNING)`
2. `if (!isDone() && !isAborting())`

Those give two results that are slightly different.
First one is true for internal states of `RUNNING`, `PENDING` and `ABORTING`, and second one for 
`PENDING`, `RUNNING`, `WAITING`, `RETRYING`. I think we should decide on one behavior there, or even change `BasicWork::getStatus` to call new virtual method `getRunningStatus` when `InternalState::RUNNING`.

marta-lokhova: Good idea, I think it should be the second, to correctly handle cases where work is aborting.

MonsieurNicolas:
moving this out for now: bucket list changes have higher priority.

MonsieurNicolas:
Code doesn't seem hooked up to anything - should this be used in the normal catchup flow anyways?

src/history/test/HistoryTests.cpp
```diff
@@ -284,6 +286,39 @@ TEST_CASE("Ledger chain verification", "[ledgerheaderverification]")
...
+    LOG(INFO) << "Testing missing ledger header fle scenario...";
+    {
+        FileTransferInfo ft(tmpDir, HISTORY_FILE_TYPE_LEDGER, range.last());
+        std::remove(ft.localPath_nogz().c_str());
+        auto verify =
+            wm.executeWork<DownloadVerifyTxResultsWork>(range, tmpDir);
+        REQUIRE(verify->getState() == BasicWork::State::WORK_FAILURE);
+    }
+
+    // TODO add extra test for mismatched hashes
```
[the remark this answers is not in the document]
marta-lokhova: Addressed in the later commit

marta-lokhova:
@MonsieurNicolas wasn’t sure if we want to introduce a new monitoring work for this. For catchup, we don’t use tx results, unless we want to verify results with whatever ApplyCheckpointWork produces. We could add a flag to CatchupWork to do that optionally. What do you think?

marta-lokhova:
To follow up on this: I'm working on an update to this PR to introduce a flag in catchup to verify tx results (already implemented here)

src/catchup/VerifyLedgerChainWork.cpp
```diff
@@ -102,7 +102,7 @@ VerifyLedgerChainWork::VerifyLedgerChainWork(
 std::string
 VerifyLedgerChainWork::getStatus() const
 {
-    if (getState() == State::WORK_RUNNING)
+    if (!isDone() && isAborting())
```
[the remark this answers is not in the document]
marta-lokhova: Good catch, it should have been `!isAborting()`

src/historywork/VerifyTxResultsWork.cpp
```diff
@@ -0,0 +1,157 @@
...
+        else
+        {
+            CLOG(DEBUG, "History")
+                << "Loaded tx result set for ledger " << ledger;
+            trs.txResultSet = mTxResultEntry.txResultSet;
+            return trs;
+        }
+    } while (mResIn && mResIn.readOne(mTxResultEntry));
+
+    return trs;
```
MonsieurNicolas: This seems weird: we return `trs` as if it was a valid result for that ledger, even though it's not.
In particular: if the file is corrupt and has missing results, it cannot be distinguished from the file containing an empty result set.

src/historywork/VerifyTxResultsWork.cpp
```diff
@@ -0,0 +1,157 @@
...
+    do
+    {
+        if (mTxResultEntry.ledgerSeq < ledger)
+        {
+            CLOG(DEBUG, "History") << "Skipping tx results for ledger "
+                                   << mTxResultEntry.ledgerSeq;
+        }
+        else if (mTxResultEntry.ledgerSeq > ledger)
+        {
+            // No tx results in this ledger
```
MonsieurNicolas: that should be an error (indicates truncated file)

src/historywork/VerifyTxResultsWork.cpp
```diff
@@ -0,0 +1,157 @@
...
+VerifyTxResultsWork::getCurrentTxResultSet(uint32_t ledger)
+{
+    TransactionHistoryResultEntry trs;
+    trs.ledgerSeq = ledger;
+
+    do
+    {
+        if (mTxResultEntry.ledgerSeq < ledger)
+        {
+            CLOG(DEBUG, "History") << "Skipping tx results for ledger "
```
MonsieurNicolas: I think this would be a bogus file (with extra results) ; for example: a file for ledgers [a..a+63] should not contain results for ledgers smaller than a, greater than a+64, and should not contain duplicates

src/historywork/VerifyTxResultsWork.cpp
```diff
@@ -0,0 +1,157 @@
...
+    catch (FileSystemException&)
+    {
+        CLOG(ERROR, "History") << "Transaction results failed verification: "
+                                  "are .xdr files downloaded?";
+        CLOG(ERROR, "History") << POSSIBLY_CORRUPTED_LOCAL_FS;
+        return false;
+    }
+
+    LedgerHeaderHistoryEntry curr;
+    while (mHdrIn && mHdrIn.readOne(curr))
```
MonsieurNicolas: not sure about the way the code is organized:
we're supposed to iterate over `mResIn` and `mHdrIn` in lockstep, but the code is written as if this was not the case.
I think that `getCurrentTxResultSet` should instead be a function that ensures that we have the right result (right now the code also reads, which causes the code to be strange)

MonsieurNicolas: I think that if `readOne` throws, this will cause the app to crash?

src/history/test/HistoryTests.cpp
```diff
@@ -383,6 +383,18 @@ TEST_CASE("History publish to multiple archives", "[history]")
...
+TEST_CASE("History catchup with extra validation", "[history][publish]")
+{
+    CatchupSimulation catchupSimulation{};
+    auto checkpointLedger = catchupSimulation.getLastCheckpointLedger(3);
+    catchupSimulation.ensureOfflineCatchupPossible(checkpointLedger);
+
+    auto app = catchupSimulation.createCatchupApplication(
+        std::numeric_limits<uint32_t>::max(), Config::TESTDB_ON_DISK_SQLITE,
+        "app");
+    REQUIRE(catchupSimulation.catchupOffline(app, checkpointLedger, true));
```
MonsieurNicolas: can you add at least one test that checks for a bad result file?

src/historywork/VerifyTxResultsWork.h
```diff
@@ -0,0 +1,45 @@
...
+                        uint32_t checkpoint);
+
+  protected:
+    BasicWork::State onRun() override;
+    void onReset() override;
+
+    bool
+    onAbort() override
+    {
+        return true;
```
MonsieurNicolas: I don't know if you can implement `abort` like this: this will cause callers to unwind even though verifications are happening in the background which may impact cleanup code, or retries in some weird way. I would imagine that you need to cause the worker to fail and wait for the failure to be posted back onto the main thread.

src/historywork/VerifyTxResultsWork.cpp
marta-lokhova: yeah, that's not desired. Fixed.

src/historywork/VerifyTxResultsWork.cpp
marta-lokhova: As discussed before, result files with missing entries are valid if a ledger did not contain any transactions, so we need to accommodate the "gaps". I added additional checks to ensure that we read valid entries though.

MonsieurNicolas:
r+ cd7e0d7e10a4ea3b2d1f660aff3274ade777b548

MonsieurNicolas:
r+ 57d10fae462d1f05aed45ca59a7c44011b33379b

PR #2105, opened 2019-05-16
Test improvements

marta-lokhova:
Few updates to the test suite, which makes it run a few minutes faster (mostly due to not using timeout mechanism when it doesn't need to)

Also, repair missing buckets work shouldn't be retrying - individual work sequences it manages already handle retries correctly.

### DIFF
--- a/src/history/test/HistoryTestsUtils.cpp
+++ b/src/history/test/HistoryTestsUtils.cpp
@@ -14,6 +14,7 @@
 #include "ledger/LedgerTxn.h"
 #include "ledger/LedgerTxnHeader.h"
 #include "lib/catch.hpp"
+#include "main/ApplicationUtils.h"
 #include "test/TestAccount.h"
 #include "test/TestUtils.h"
 #include "test/TxTests.h"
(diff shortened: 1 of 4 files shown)


src/history/test/HistoryTestsUtils.cpp
```diff
@@ -615,14 +616,9 @@ CatchupSimulation::catchupOffline(Application::pointer app, uint32_t toLedger)
...
-    lm.startCatchup(catchupConfiguration);
-    REQUIRE(!app->getClock().getIOContext().stopped());
-
-    auto caughtUp = [&]() { return lm.isSynced(); };
-    crankUntil(app, caughtUp, std::chrono::seconds{30});
-
-    auto result = caughtUp();
-    if (result)
+    Json::Value val;
+    bool success = catchup(app, catchupConfiguration, val) == 0;
```
MonsieurNicolas: this change removes the timeout check, which doesn't seem desirable.
I would add a timeout parameter to `catchup` if we really want to do this

marta-lokhova: Ok, brought the timeout back -- the condition for completion just mimics one for offline catchup (it's much faster to just check for catchup failure instead of waiting for 30 seconds timeout)

MonsieurNicolas:
r+ ae7898da0559919e2be9976259e3d78072462725

PR #2108, opened 2019-05-20
Add SCP timeout metrics

### LINKED ISSUE #1877: metrics: add metrics that track SCP timeouts
As per #1831, we're missing metrics that track how many times timeouts occur per SCP round for the various types of timers that exist (right now ballot protocol and nomination timers)
Easiest might be to have one metric per timer type, ie `scp.timeouts-X` where X is "ballot" or "nomination".
Alternative, that might be more useful is to capture those counters only at specific times:
* nomination timeouts before first prepare
* ballot timeouts before externalization

marta-lokhova:
Resolves #1877

### DIFF
--- a/src/herder/HerderSCPDriver.cpp
+++ b/src/herder/HerderSCPDriver.cpp
@@ -13,6 +13,7 @@
 #include "main/Application.h"
 #include "main/ErrorMessages.h"
 #include "scp/SCP.h"
+#include "scp/Slot.h"
 #include "util/Logging.h"
 #include "xdr/Stellar-SCP.h"
 #include "xdr/Stellar-ledger-entries.h"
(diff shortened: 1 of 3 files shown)


src/herder/HerderSCPDriver.cpp
```diff
@@ -469,6 +474,30 @@ HerderSCPDriver::timerCallbackWrapper(uint64_t slotIndex, int timerID,
...
+        if (SCPTimingIt != mSCPExecutionTimes.end())
+        {
+            auto& SCPTiming = SCPTimingIt->second;
+            if (timerID == Slot::BALLOT_PROTOCOL_TIMER)
+            {
+                if (SCPTiming.mSCPProgressState ==
+                    SCPProgressState::FIRST_PREPARE)
+                {
+                    // Timeout happened in between first prepare and externalize
+                    mSCPMetrics.mPrepareTimeout.Mark();
```
MonsieurNicolas: I think the metrics as measured are not going to be super useful: we should keep track of them in units of ledger (so maybe record the number of timeouts when we externalize using a histogram, the same way we track number of transactions/ledger)

src/herder/HerderSCPDriver.cpp
```diff
@@ -897,6 +928,7 @@ HerderSCPDriver::recordSCPExecutionMetrics(uint64_t slotIndex)
     }
 
     auto& SCPTiming = SCPTimingIt->second;
+    SCPTiming.mSCPProgressState = SCPProgressState::EXTERNALIZE;
```
MonsieurNicolas: by moving to "per ledger", you won't need to keep track of EXTERNALIZE anymore as this function will be the only place where you interact with metrics

src/herder/HerderSCPDriver.cpp
```diff
@@ -469,6 +474,30 @@ HerderSCPDriver::timerCallbackWrapper(uint64_t slotIndex, int timerID,
...
+                if (SCPTiming.mSCPProgressState ==
+                    SCPProgressState::FIRST_PREPARE)
+                {
+                    // Timeout happened in between first prepare and externalize
+                    mSCPMetrics.mPrepareTimeout.Mark();
+                }
+            }
+            else
+            {
+                if (SCPTiming.mSCPProgressState ==
```
MonsieurNicolas: you should just use `mPrepareStart` as indicator that prepare happened

marta-lokhova:
updated

MonsieurNicolas:
r+ bfa9e59bd1875ed482e79779e9eac6fd90f6a6fc

PR #2109, opened 2019-05-22
Check validity of upgrades before applying them during closeLedger

jonjove:
Fixes an issue introduced by me in 92d2f48be. That commit, part of the implementation of [CAP-0003](https://github.com/stellar/stellar-protocol/blob/master/core/cap-0003.md), introduced the possibility that an upgrade can effect ledger entries. But if a node applies an upgrade without understanding what it needs to do to ledger entries, then it will not produce the same hash as a node that does understand what it needs to do to ledger entries. Now nodes will crash when trying to apply an upgrade that they don't understand.

### DIFF
--- a/src/herder/Upgrades.cpp
+++ b/src/herder/Upgrades.cpp
@@ -245,77 +245,90 @@ Upgrades::removeUpgrades(std::vector<UpgradeType>::const_iterator beginUpdates,
     return res;
 }
 
-bool
-Upgrades::isValid(UpgradeType const& upgrade, LedgerUpgradeType& upgradeType,
-                  bool nomination, Config const& cfg,
-                  LedgerHeader const& header) const
+Upgrades::UpgradeValidity
+Upgrades::isValidForApply(UpgradeType const& opaqueUpgrade,
+                          LedgerUpgrade& upgrade, LedgerHeader const& header,
+                          uint32_t maxLedgerVersion)
 {
-    if (nomination && !timeForUpgrade(header.scpValue.closeTime))
-    {
-        return false;
-    }
-
-    LedgerUpgrade lupgrade;
-
     try
     {
-        xdr::xdr_from_opaque(upgrade, lupgrade);
+        xdr::xdr_from_opaque(opaqueUpgrade, upgrade);
     }
     catch (xdr::xdr_runtime_error&)
     {
-        return false;
+        return UpgradeValidity::XDR_INVALID;
     }
 
     bool res = true;
-    switch (lupgrade.type())
+    switch (upgrade.type())
     {
     case LEDGER_UPGRADE_VERSION:
     {
-        uint32 newVersion = lupgrade.newLedgerVersion();
-        if (nomination)
-        {
-            res = mParams.mProtocolVersion &&
-                  (newVersion == *mParams.mProtocolVersion);
-        }
+        uint32 newVersion = upgrade.newLedgerVersion();
         // only allow upgrades to a supported version of the protocol
-        res = res && (newVersion <= cfg.LEDGER_PROTOCOL_VERSION);
+        res = res && (newVersion <= maxLedgerVersion);
         // and enforce versions to be strictly monotonic
         res = res && (newVersion > header.ledgerVersion);
     }
     break;
     case LEDGER_UPGRADE_BASE_FEE:
-    {
-        uint32 newFee = lupgrade.newBaseFee();
-        if (nomination)
-        {
-            res = mParams.mBaseFee && (newFee == *mParams.mBaseFee);
-        }
-        res = res && (newFee != 0);
-    }
-    break;
+        res = res && (upgrade.newBaseFee() != 0);
+        break;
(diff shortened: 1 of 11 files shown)


vogel:
Looks good, but lacks tests for new behavior. Could you add some?

src/herder/Upgrades.cpp
```diff
@@ -246,76 +246,94 @@ Upgrades::removeUpgrades(std::vector<UpgradeType>::const_iterator beginUpdates,
...
-
-    try
-    {
-        xdr::xdr_from_opaque(upgrade, lupgrade);
-    }
-    catch (xdr::xdr_runtime_error&)
-    {
-        return false;
-    }
+    xdr::xdr_from_opaque(opaqueUpgrade, lupgrade);
```
MonsieurNicolas: why do you bubble up `xdr::xdr_runtime_error` ? that upgrade is invalid

src/herder/Upgrades.cpp
```diff
@@ -246,76 +246,94 @@ Upgrades::removeUpgrades(std::vector<UpgradeType>::const_iterator beginUpdates,
...
-        if (nomination)
-        {
-            res = mParams.mBaseReserve && (newReserve == *mParams.mBaseReserve);
-        }
-        res = res && (newReserve != 0);
+        res = isValid(upgrade, lupgrade, header, cfg.LEDGER_PROTOCOL_VERSION);
     }
-    break;
-    default:
+    catch (xdr::xdr_runtime_error&)
```
MonsieurNicolas: `isValid` should not throw exception (see my previous comment)

src/herder/Upgrades.h
```diff
@@ -95,7 +101,13 @@ class Upgrades
   private:
     UpgradeParameters mParams;
 
-    bool timeForUpgrade(uint64_t time) const;
+    static bool timeForUpgrade(uint64_t time, UpgradeParameters const& params);
```
MonsieurNicolas: I don't see a reason to make those static methods

src/ledger/LedgerManagerImpl.cpp
```diff
@@ -855,9 +855,15 @@ LedgerManagerImpl::closeLedger(LedgerCloseData const& ledgerData)
...
+            if (!Upgrades::isValid(sv.upgrades[i], lupgrade,
+                                   ltx.loadHeader().current(),
+                                   Config::CURRENT_LEDGER_PROTOCOL_VERSION))
+            {
+                throw std::runtime_error("Invalid upgrade: " +
+                                         xdr::xdr_to_string(lupgrade));
+            }
         }
-        catch (xdr::xdr_runtime_error)
+        catch (xdr::xdr_runtime_error&)
```
MonsieurNicolas: no need to have this catch handler if `isValid` doesn't throw

src/ledger/LedgerManagerImpl.cpp
```diff
@@ -855,9 +855,15 @@ LedgerManagerImpl::closeLedger(LedgerCloseData const& ledgerData)
         LedgerUpgrade lupgrade;
         try
         {
-            xdr::xdr_from_opaque(sv.upgrades[i], lupgrade);
+            if (!Upgrades::isValid(sv.upgrades[i], lupgrade,
+                                   ltx.loadHeader().current(),
+                                   Config::CURRENT_LEDGER_PROTOCOL_VERSION))
```
MonsieurNicolas: As much as possible, we should not use `Config::CURRENT_LEDGER_PROTOCOL_VERSION` (and use `cfg.LEDGER_PROTOCOL_VERSION`) as some tests are testing mixed version, and we want them to behave as close as possible to "the old version" for that protocol (ie they behave like the old code)

MonsieurNicolas:
re: `cfg.LEDGER_PROTOCOL_VERSION`

I looked at the upgrade tests and the only ones that seems wrong are "upgrade to version 10/11" and "upgrade base reserve" (all added recently):
they don't set `USE_CONFIG_FOR_GENESIS=false`.
They should instead do the same thing than what `simulateUpgrade` does: setup the app with the desired version and close a ledger with the actual upgrade to the desired version (like you're doing in the new test that you've added, except it should not fail).

jonjove:
@MonsieurNicolas fixed the issues with not using `USE_CONFIG_FOR_GENESIS` but still need to squash/clang-format before merging.

src/herder/test/UpgradesTests.cpp
```diff
@@ -214,6 +214,27 @@ toUpgradeType(LedgerUpgrade const& upgrade)
     return result;
 }
 
+LedgerHeader
+executeUpgrades(Application& app, xdr::xvector<UpgradeType, 6> const& upgrades)
```
MonsieurNicolas: we should put all those helper functions in the empty namespace as they're only used in this file

jonjove:
@MonsieurNicolas Done, should be ready to merge now.

MonsieurNicolas:
r+ 7f00d43a296e3a3254ee6d74a6168b157f585520

PR #2112, opened 2019-05-23
Persist externalize messages for older slots

### LINKED ISSUE #2107: Herder: persist/restore state should work for old ledgers
Right now nodes keep track of messages in two places:
* in memory for all validators in quorum, for the currently active slots and previous slots (up to `MAX_SLOTS_TO_REMEMBER` = 12 slots)
* on disk (`persistSCPState` and `restoreSCPState`) for the currently active ledger for that validator

On restart, the in memory state is blank but is repopulated in two ways:
* from whatever was saved last on disk
* from peers

The problem is that there is a small chance that a sequence like this happens:
1. a rapid succession of state changes by the local node (for example: `EXTERNALIZE@N` followed by `NOMINATE@N+1`), cause the node to only have the previous data (`N`) in memory and current (`N+1`) on disk
2. the node gets killed/crashes before peers receive messages for older messages (`N`)

At that point, on restart, only the latest messages (for `N+1`) can be forwarded to the network.

If messages for `N` are needed by some peers, they won't have it.

Solution is to keep not just the latest messages for the latest slot but also `EXTERNALIZE` messages for previous slots (so that the in memory cache can be repopulated for those old slots as well).

marta-lokhova:
This should resolve #2107, as `EXTERNALIZE` messages for older slots are now persisted.

### DIFF
--- a/src/herder/HerderImpl.cpp
+++ b/src/herder/HerderImpl.cpp
@@ -1065,7 +1065,7 @@ HerderImpl::persistSCPState(uint64 slot)
     std::string scpState;
     scpState = decoder::encode_b64(latestSCPData);
 
-    mApp.getPersistentState().setState(PersistentState::kLastSCPData, scpState);
+    mApp.getPersistentState().setSCPStateForSlot(slot, scpState);
 }
 
 void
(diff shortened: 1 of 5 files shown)


src/herder/HerderImpl.cpp
```diff
@@ -1047,6 +1047,25 @@ HerderImpl::persistSCPState(uint64 slot)
         }
     }
 
+    // Remember EXTERNALIZE envelopes for previous slots as well
+    auto prevSlot = slot - 1;
+    while (prevSlot > 0 && slot - prevSlot < Herder::MAX_SLOTS_TO_REMEMBER)
```
MonsieurNicolas: I think this is wrong: we need to do the same thing than in the previous loop (remember `txset` and `qset` values). We probably need to add a test case that old slots are indeed properly persisted.

Are there performance implications by implementing it like this?
it's going to save 12 transaction sets in the same blob object (potentially over 100Mb of data) every time we call `emitEnvelope` (several times per externalize).

I suspect we should:
* at a minimum only update the data for old slots when we externalize
* better, use some sort of "circular data structure", that allows to only update one of the old persisted slots (by, for example, using not just 1 entry `kLastSCPData` but 13 separate)

src/herder/HerderImpl.cpp
```diff
@@ -1086,55 +1086,66 @@ HerderImpl::restoreSCPState()
...
+            // this should be the only time we get exceptions decoding old
+            // messages.
+            CLOG(INFO, "Herder") << "Error while restoring old scp messages, "
+                                    "proceeding without them : "
+                                 << e.what();
         }
     }
-    catch (std::exception& e)
+
+    if (broadcast)
```
MonsieurNicolas: correction on broadcast flag: it's probably better to always call that function, I am not sure why it was behind that condition

src/herder/HerderImpl.cpp
```diff
@@ -1086,55 +1086,66 @@ HerderImpl::restoreSCPState()
...
+            }
+            for (auto const& qset : latestQSets)
+            {
+                Hash hash = sha256(xdr::xdr_to_opaque(qset));
+                mPendingEnvelopes.addSCPQuorumSet(hash, qset);
+            }
+            for (auto const& e : latestEnvs)
+            {
+                // Only restore all envelopes for the latest slot
+                if (e.statement.slotIndex == lcl.header.ledgerSeq ||
```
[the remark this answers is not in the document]
MonsieurNicolas: (I suspect broadcast should just be called if we have any statement, it doesn't matter)

src/herder/test/HerderTests.cpp
```diff
@@ -1348,21 +1348,28 @@ TEST_CASE("quick restart", "[herder][quickRestart]")
...
+    auto listenerCfg =
+        getTestConfig(2, Config::TestDbMode::TESTDB_ON_DISK_SQLITE);
+
     SCPQuorumSet qSet;
     qSet.threshold = 1;
     qSet.validators.push_back(validatorKey.getPublicKey());
 
-    auto validator = simulation->addNode(validatorKey, qSet);
-    auto listener = simulation->addNode(listenerKey, qSet);
+    simulation->addNode(validatorKey, qSet, &validatorCfg);
```
MonsieurNicolas: we should not change this test like you did: this test is for testing that we send the right messages when a node drops off the network for a while.

There is a different test right above: "SCP State" that should be modified. Right now it only close 1 ledger (as before this PR only 1 ledger was persisted), but now we should be able to close a few ledgers (like 5).

src/main/PersistentState.cpp
```diff
@@ -36,47 +38,94 @@ PersistentState::dropAll(Database& db)
...
-string
+std::string
 PersistentState::getStoreStateName(PersistentState::Entry n)
 {
     if (n < 0 || n >= kLastEntry)
     {
         throw out_of_range("unknown entry");
     }
-    return mapping[n];
+    auto res = mapping[n];
```
MonsieurNicolas: this is pretty ugly and also introduces a bug.

First the bug:
you're not restoring SCP state after an upgrade.

The badness:
having `getStoreStateName` rely on side effects is super counter intuitive.

I suggest instead that you add an optional parameter `int subscript` (default 0) that, when non 0, gets appended to the name of the key (otherwise don't append anything).

src/main/PersistentState.cpp
```diff
@@ -36,47 +38,94 @@ PersistentState::dropAll(Database& db)
...
+PersistentState::getSCPState()
+{
+    // First, collect all slots persisted
+    std::vector<std::string> states;
+    mSCPStateCounter = 0;
+    for (int i = 0; i <= Herder::MAX_SLOTS_TO_REMEMBER; i++)
     {
-        auto timer = db.getSelectTimer("state");
-        st.execute(true);
+        auto val = getFromDb(kLastSCPData);
```
MonsieurNicolas: I think those helper functions `updateDb/getFromDb` are better off taking the key name (string) as parameter

src/main/PersistentState.cpp
```diff
@@ -36,47 +38,94 @@ PersistentState::dropAll(Database& db)
...
+        if (!val.empty())
+        {
+            states.push_back(val);
+        }
+        mSCPStateCounter++;
     }
 
-    if (!st.got_data())
+    // Now restore SCP state count from DB to determine
```
MonsieurNicolas: you don't need this eviction logic: at this layer you just keep `MAX_SLOTS_TO_REMEMBER+1`, and let the caller override whatever it wants to

src/main/PersistentState.cpp
```diff
@@ -36,47 +38,94 @@ PersistentState::dropAll(Database& db)
...
+PersistentState::setSCPState(uint64 slot, std::string const& value)
+{
+    bool updateCounter = false;
+    auto slotIdx = static_cast<int>(slot % (Herder::MAX_SLOTS_TO_REMEMBER + 1));
+
+    // Setting SCP state for the next slot
+    if (slotIdx != mSCPStateCounter)
+    {
+        mSCPStateCounter = slotIdx;
+        updateCounter = true;
```
MonsieurNicolas: no need to use this counter

src/main/PersistentState.cpp
marta-lokhova: Updated.

src/main/PersistentState.cpp
marta-lokhova: Fixed.

src/herder/HerderImpl.cpp
marta-lokhova: Updated.

src/main/PersistentState.cpp
marta-lokhova: Ok, updated.

src/herder/HerderImpl.cpp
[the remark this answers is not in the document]
marta-lokhova: I think we only needed externalize messages for older slots, but it looks it is much simpler to just store them all entirely (then we also don't need to sort), so I updated it.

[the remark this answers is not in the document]
marta-lokhova: Seems like it's much simpler to just not filter messages at all, so I did that.

marta-lokhova:
tests updated and squashed

MonsieurNicolas:
r+ d4ad629b3c44c735dafd089ab93ca98b06c1435e

MonsieurNicolas:
@latobarita: retry

PR #2114, opened 2019-05-24
Split key verification into its own file

Geod24:
```
Before this commit, the SCP library transitively imported some modules it did not need,
such as 'main/Config.h' or 'util/RandomEvictionCache.h' because of the dependency
on 'crypto/SecretKey.h'. Splitting the verification cache in its own module fixes that.
```

In addition, it looks like some `#include` were not needed at all (e.g. `main/Config.h`).

# Checklist
- [X] Compiles
- [X] Ran all tests
- [X] ~If change impacts performance~

### DIFF
--- a/src/crypto/KeyVerification.cpp
+++ b/src/crypto/KeyVerification.cpp
@@ -1,0 +1,88 @@
+// Copyright 2014-2019 Stellar Development Foundation and contributors. Licensed
+// under the Apache License, Version 2.0. See the COPYING file at the root
+// of this distribution or at http://www.apache.org/licenses/LICENSE-2.0
+
+
+#include "crypto/KeyVerification.h"
+#include "crypto/SHA.h"
+#include "util/HashOfHash.h"
+#include "util/RandomEvictionCache.h"
+#include <mutex>
+#include <sodium.h>
+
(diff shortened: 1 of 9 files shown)


MonsieurNicolas:
Cool @Geod24 

I think that it would make sense to just completely remove the crypto dependency from SCP: I started to do it, and I suspect it's almost done.
Basically: all crypto should be handled by `HerderSCPDriver` instead of the default `SCPDriver`.
That way allows to remove the dependency on libsodium as well.

Geod24:
Does that apply only to the crypto dependency, or also to other dependencies ? E.g. easylogging++ and the JSON library.

Since your PR to move the signing code, the only torn in my side has been related to logging. `toStrKey` / `toShortString` still depend on `PublicKey` and `KeyUtils`. 

Does that also mean you don't want separation at this level and I should close this PR ?

MonsieurNicolas:
Yeah I think you should just close this PR and remove the crypto dependency by making the default `SCPDriver` crypto agnostic, which is probably done by just removing the default implementations for `toStrKey` and `toShortString`

MonsieurNicolas:
closing this PR in favor of other changes

PR #2118, opened 2019-05-28
Reduce libsodium dependency in ByteSliceHasher

Geod24:
```
The crypto_shorthash_keygen function was introduced in libsodium 1.0.12,
released 2017-03-17.
Unfortunately, this version is not available on the oldest Ubuntu LTS,
Xenial (16.04), and only libsodium18 (1.0.8, released 2015-12-26) is.
This minor change reduce the libsodium dependency for this file
and makes it usable on such boxes (including Travis CI).
```

Source: https://github.com/jedisct1/libsodium/blob/1.0.17/src/libsodium/crypto_shorthash/crypto_shorthash.c#L30-L34

### DIFF
--- a/src/crypto/ByteSliceHasher.cpp
+++ b/src/crypto/ByteSliceHasher.cpp
@@ -13,7 +13,7 @@ static unsigned char sKey[crypto_shorthash_KEYBYTES];
 void
 initialize()
 {
-    crypto_shorthash_keygen(sKey);
+    randombytes_buf(sKey, crypto_shorthash_KEYBYTES);
 }
 uint64_t
 computeHash(stellar::ByteSlice const& b)


MonsieurNicolas:
Thanks for catching this one.

The approach we've been taking is that instead of having people rely on unmaintained old packages like the one bundled in Xenial, people are better off using the version of libsodium bundled with stellar-core (we should only use the version outside of stellar-core if it's more recent).

In this case, I think that all that is missing is the proper version: right now configure checks for version 1.0.13 instead it should just check for 1.0.17

Geod24:
> The approach we've been taking is that instead of having people rely on unmaintained old packages like the one bundled in Xenial, people are better off using the version of libsodium bundled with stellar-core (we should only use the version outside of stellar-core if it's more recent).

The reason was doing so was to make our CI much faster and simpler. Just look at https://github.com/bpfkorea/agora/pull/8/files
libsodium is relatively stable that you can use an old version without too much hassle.
While it's good to have a check in `configure`, and it's probably better to have the latest version, in this case, it's just a matter of calling a function directly instead of indirectly (because `crypto_shorthash_keygen` is *just* calling `randombytes_buf` and nothing else).

MonsieurNicolas:
The best thing to do to speed up CI in this situation is to simply enable ccache (like we did): I don't think it makes sense to use a version that misses out on over 3 years worth of changes in this case (there are both performance and security related changes that went in since that version).

Geod24:
> The best thing to do to speed up CI in this situation is to simply enable ccache (like we did)

That's another topic tho. Compilation speed is much less problematic for D than it is for C++.

> I don't think it makes sense to use a version that misses out on over 3 years worth of changes in this case (there are both performance and security related changes that went in since that version).

Agree to disagree. It's just a shame that such a simple patch couldn't make it in, as it literally changes nothing for you (upstream) but simplify life for us by reducing our diff.

PR #2129, opened 2019-06-05
Fix signatures in transaction tests

vogel:
# Description

This PR fixes one issue with our tests.
For some transaction tests the following was happening:
- create transaction
- sign transaction
- update transaction envelope
- test result

All of those tests were passing, but should not. As the envelope was updated after signing, it contained old, invalid signatures. However, those envelopes also contained old, not up to date, content hashes - those were only cleared after adding new signatures and/or manually (but only in one case).

This is now fixed by making TransactionFramePtr envelope immutable - it can be only read now, not written to. It required changes to several tests that were modifying TransactionFramePtr envelope directly. Now those tests create and work on TransactionEnvelope directly, until it it done. Then conversion to TransactionFramePtr happen.

Several new methods were added to TestAccount (rawTx that makes TransactionEnvelope, variant of tx that works on Transaction and unsignedTx that creates TransactionFramePtr without any signature).

In addition three last applyXXX functions from TxTests.* were moved to TestAccount to match behavior of all of the others applyXXX functions.

# Checklist
- [x] Compiles
- [x] Ran all tests


src/transactions/test/TxResultsTests.cpp
```diff
@@ -421,40 +430,48 @@ TEST_CASE("txresults", "[tx][txresults]")
...
             SECTION("insufficent fee")
             {
-                auto tx = a.tx({payment(root, 1)});
-                tx->addSignature(a);
-                tx->getEnvelope().tx.fee--;
+                auto rawTx = a.rawTx({payment(root, 1)});
+                rawTx.fee--;
+                auto tx = a.unsignedTx(rawTx);
+                a.sign(tx);
+                a.sign(tx);
```
MonsieurNicolas: extra signature (that was a bug before)

src/transactions/test/TxResultsTests.cpp
```diff
@@ -464,17 +481,21 @@ TEST_CASE("txresults", "[tx][txresults]")
             SECTION("no account")
             {
                 auto tx = f.tx({payment(root, 1)});
-                tx->addSignature(a);
+                f.sign(tx);
```
MonsieurNicolas: isn't that a double signature as well?

src/transactions/test/TxResultsTests.cpp
```diff
@@ -464,17 +481,21 @@ TEST_CASE("txresults", "[tx][txresults]")
...
             SECTION("bad seq")
             {
-                auto tx = a.tx({payment(root, 1)});
-                tx->addSignature(a);
-                tx->getEnvelope().tx.seqNum++;
+                auto rawTx = a.rawTx({payment(root, 1)});
+                rawTx.seqNum++;
+                auto tx = a.unsignedTx(rawTx);
+                a.sign(tx);
+                a.sign(tx);
```
MonsieurNicolas: double signature as well

src/transactions/test/TxResultsTests.cpp
```diff
@@ -483,7 +504,8 @@ TEST_CASE("txresults", "[tx][txresults]")
             SECTION("insufficent balance")
             {
                 auto tx = g.tx({payment(root, 1)});
-                tx->addSignature(a);
+                g.sign(tx);
```
MonsieurNicolas: double signature?

MonsieurNicolas:
Regarding some of the duplicate signatures or bad sequence numbers in the envelope tests...

I just realized (forgot about this) that in the envelope tests, we want to not only test that failures happen but that checks happen in the right order: so we need to make sure when checking that a condition is false that we, on purpose, make subsequent checks invalid as well (in case we modify the check order by accident).
The order is specified by following the order when going through `TransactionFrame::checkValid`.

For example, imagine that we test that `txTOO_EARLY` is returned, then we need to craft a transaction with a `minTime` set too high (as you'd expect), but also
a fee too small (`txINSUFFICIENT_FEE`), a non existent account (`txNO_ACCOUNT`), a bad sequence number (`txBAD_SEQ`), etc.

That way, we know that the first check to fail is the one for `minTime` in this example.

Best is to create a helper function that makes it easy to modify a transaction as to invalidate any checks past a certain point (instead of having those invalidations sprinkled in the code base).
All those are easy to do "in place", for example invalidating signatures can be done by adding a bogus one, invalidating the seqnum is done by making it negative, etc.

src/transactions/test/TxResultsTests.cpp
vogel: This is in SECTION("double signed"), testing for which error code we will get.

src/transactions/test/TxResultsTests.cpp
vogel: As above

src/transactions/test/TxResultsTests.cpp
vogel: As above

src/transactions/test/TxResultsTests.cpp
vogel: As above

vogel:
I've updated the tests. Also removed some duplicate ones from TxEnvelopeTests.cpp (I think those in TxResultsTests.cpp are better).

I'll now check how much slower are the tests because of the extra copies.

vogel:
There is about 2% more copies of Transaction object in the new code, and 160 000 new moves of that type (with 0 before).

There is no visible time difference.

src/transactions/test/TxResultsTests.cpp
```diff
@@ -113,6 +112,236 @@ TEST_CASE("txresults", "[tx][txresults]")
...
 
+    // set up world
+    auto root = TestAccount::createRoot(*app);
+    auto nonExisting = TestAccount{*app, getAccount("nonExisting")};
+    auto nonEnoughBalance =
+        root.create("nonEnoughBalance", lm.getLastMinBalance(0));
+    auto enoughBalance = root.create("enoughBalance", startAmount);
+    auto extraSignature = TestAccount{*app, getAccount("extraSignature")};
+
+    auto validateError = [&](TestAccount const& account,
```
MonsieurNicolas: The list of checks done are in order:
```
no ops -> txMISSING_OPERATION
timebound -> txTOO_EARLY, then txTOO_LATE
no fee -> txINSUFFICIENT_FEE
no account -> txNO_ACCOUNT
seqnum -> txBAD_SEQ
bad auth -> txBAD_AUTH
account balance too low -> txINSUFFICIENT_BALANCE
op check valid -> txFAILED
extra auth -> txBAD_AUTH_EXTRA
```
I would expect this function to instead of taking `authLevel`, would take the place in that list where we expect the failure (type = `TransactionResultCode`).

Then all you need to do is something like this to consolidate all checks:
```c++

        // NB: cases are fall through here as we want to add all possible
       //  negative conditions if we expect a failure at a certain point
        switch (failFromCode)
        {
        case txMISSING_OPERATION:
            rawTx.ops.clear();
        case txTOO_EARLY:
            rawTx.timeBounds.activate().minTime = ct + 1;
        case txTOO_LATE:
            rawTx.timeBounds.activate().maxTime = ct - 1;
        case txINSUFFICIENT_FEE:
            rawTx.fee = 1;
        case txNO_ACCOUNT:
            // handled outside
        case txBAD_SEQ:
            rawTx.seqNum++;
        case txBAD_AUTH:
            badAuth = true;
        //...
        case txBAD_AUTH_EXTRA:
            extraAuth = true;
        };
        // now, sign and check the transaction
        (badAuth ? badAccount : account).sign(tx);
        if (extraAuth)
        {
            account2.sign(tx);
        }
        validateTxResults(tx, *app, validationResult, applyResult);
```
The difference with your version (on top of being more exhaustive) is that we only perform one check with all possible failures added at once (compared to your version that checks for them ~independently); it's actually fine to do this because we want to catch that none of them are tripping (we're using this in a negative context).

Then, the test cases are just things like:
```c++
SECTION("txTOO_LATE")
{
    auto rawTx = nonExisting.rawTx({payment(root, 0)}, 4);
    rawTx.timeBounds.activate().maxTime = ct - 1;
    checkError(rawTx, txINSUFFICIENT_FEE, {1, txTOO_LATE})
}
```

Also, at this layer we don't care to distinguish "no signature" from "bad signature" as what we want to make sure is that "signature verification" is not done before sequence number processing for example.

src/transactions/test/TxResultsTests.cpp
```diff
@@ -113,6 +112,236 @@ TEST_CASE("txresults", "[tx][txresults]")
...
+                }
+            }
+        });
+    };
+
+    SECTION("transaction errors")
+    {
+        SECTION("non existing")
+        {
+            SECTION("all possible bugs")
```
MonsieurNicolas: "all possible bugs" doesn't mean anything: what you're testing is `txMISSING_OPERATION`

src/transactions/test/TxResultsTests.cpp
```diff
@@ -113,6 +112,236 @@ TEST_CASE("txresults", "[tx][txresults]")
...
+            {
+                auto rawTx = nonExisting.rawTx({}, 4);
+                rawTx.timeBounds.activate().minTime = getCloseTime() + 1;
+                rawTx.timeBounds.activate().maxTime = getCloseTime() - 1;
+                rawTx.fee = 1;
+
+                validateError(nonExisting, rawTx, 0, {1, txMISSING_OPERATION});
+            }
+
+            SECTION("txMISSING_OPERATION solved")
```
MonsieurNicolas: the section names are super confusing: "XYZ solved" doesn't say what the section is testing, this one for example should be "txTOO_EARLY"

src/transactions/test/TxResultsTests.cpp
```diff
@@ -89,7 +89,7 @@ variations(int count, std::vector<T> values)
     return result;
 }
 
-TEST_CASE("txresults", "[tx][txresults]")
+TEST_CASE("txresults - errors", "[tx][txresults]")
```
MonsieurNicolas: due to limitations in the parallel test scripts, we cannot use `-` in test names, only `[A-Za-z0-9 _()]+`

src/transactions/test/TxEnvelopeTests.cpp
```diff
@@ -1022,133 +1022,6 @@ TEST_CASE("txenvelope", "[tx][envelope]")
         }
     }
 
-    SECTION("common transaction")
```
MonsieurNicolas: those are not duplicate tests so this refactor is incomplete:
they are calling "applyCheck" directly.

I suspect that the problem is that the refactored version calls `validateTxResults` which doesn't check what `apply` returns on failure

src/transactions/test/TxResultsTests.cpp
```diff
@@ -113,6 +112,236 @@ TEST_CASE("txresults", "[tx][txresults]")
...
+                    validateTxResults(tx, *app,
+                                      {validationResult.fee, txBAD_AUTH_EXTRA});
+                }
+            }
+        });
+    };
+
+    SECTION("transaction errors")
+    {
+        SECTION("non existing")
```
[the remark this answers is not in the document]
vogel: The idea was to test if existence of account changes error (i.e. if txNO_ACCOUNT can be returned even if there are no operations).

src/transactions/test/TxEnvelopeTests.cpp
vogel: So do you want to keep tests in TxResultsTests and those here as well?

MonsieurNicolas: no, it's fine to refactor (as the new approach is more systematic), you just need to make sure that we call `applyCheck` in all cases

vogel: New version calls `applyCheck` in all cases.

I don't see why in `applyCheck` we need to call `apply` even when `checkValid` failed. This causes many strange edge cases in testing code, as errors reported by `checkValid` and `apply` can be different. I propose that we remove those check and only call `apply` when `checkValid` returned true.

src/test/TxTests.cpp
```diff
@@ -191,25 +197,47 @@ applyCheck(TransactionFramePtr tx, Application& app, bool checkSeqNum)
...
+                if (tx->getResult().result.code() != txBAD_AUTH_EXTRA &&
+                    checkResult.result.code() != txBAD_AUTH_EXTRA &&
+                    (checkResult.result.code() != txBAD_SEQ ||
+                     header.current().ledgerVersion >= 10))
+                {
+                    REQUIRE(checkResult == tx->getResult());
+                }
+                else
+                {
+                    // see "merge one of signing accounts"/"by source, signed by
```
MonsieurNicolas: this question doesn't belong in the code base (discussing in PR comments instead)

src/test/TxTests.cpp
```diff
@@ -191,25 +197,47 @@ applyCheck(TransactionFramePtr tx, Application& app, bool checkSeqNum)
...
-                    REQUIRE(checkResults[i] == txResults[i]);
-                    if (checkResults[i].code() == opBAD_AUTH)
+                    try
+                    {
+                        throwIf(checkResults[i]);
+                        throwIf(txResults[i]);
+                    }
+                    catch (...)
                     {
-                        // results may not match after first opBAD_AUTH
```
MonsieurNicolas: why did you remove this comment?

src/test/TxTests.cpp
```diff
@@ -191,25 +197,47 @@ applyCheck(TransactionFramePtr tx, Application& app, bool checkSeqNum)
...
                         break;
                     }
+
+                    REQUIRE(checkResults[i] == txResults[i]);
+                }
+            }
+            else
+            {
+                auto header = ltxTx.loadHeader();
+                if (tx->getResult().result.code() != txBAD_AUTH_EXTRA &&
```
MonsieurNicolas: The way this condition is written is not super clear:
rewrite it in such a way that it's easy to understand why you're blacklisting some result permutation.

src/transactions/test/TxEnvelopeTests.cpp
MonsieurNicolas: replying in the main PR thread

src/transactions/test/TxResultsTests.cpp
```diff
@@ -108,11 +107,131 @@ TEST_CASE("txresults", "[tx][txresults]")
...
+
+        auto noAccount = false;
+        auto badAuth = false;
+        auto extraAuth = false;
+
+        // NB: cases are fall through here as we want to add all possible
+        // negative conditions if we expect a failure at a certain point
+        switch (failFromCode)
+        {
+        case txMISSING_OPERATION:
```
MonsieurNicolas: nice

MonsieurNicolas:
> New version calls applyCheck in all cases.
I don't see why in applyCheck we need to call apply even when checkValid failed. This causes many strange edge cases in testing code, as errors reported by checkValid and apply can be different. I propose that we remove those check and only call apply when checkValid returned true.

we could remove them; but then we need to add new tests. The original intent is captured in the comment:
```
        // now, check what happens when simulating what happens during a ledger
        // close and reconcile it with the return value of "apply" with the one
        // from checkValid:
        // * an invalid (as per isValid) tx is still invalid during apply (and
        // the same way)
        // * a valid tx can fail later
```

What we could do is that we don't allow `applyCheck` to be called on invalid transactions (fail tests that do this), instead test would pass in the expected *pair* of results for both `checkValid` and `apply` and maybe the list of expected side effects.
This allows `applyCheck` to not have any of the special casing (like `earlyFailure` and others).
It's true that various versions of the protocol modify how operations within a given transaction can cause the transaction to fail; with this change we would be able to properly express the expected behavior from test code.

MonsieurNicolas:
you also need to rebase this

src/test/TxTests.cpp
vogel: it is replaced with comment before - // should be equal up to first failure

vogel:
OK, I understand now. But:

1. `an invalid (as per isValid) tx is still invalid during apply (and the same way)` - this is not true anymore (and this is causing us to check for many strange cases). Transaction can be invalid in different way or even valid during apply (as in case of AUTH_EXTRA) bugs.
2. `a valid tx can fail later` that is OK, and doing the mentioned change wont affect this codepath at all

> instead test would pass in the expected pair of results for both checkValid and apply and maybe the list of expected side effects

We do that already with `validateTxResults`, but it ignores the `applyResult` is `validationResult` are not OK.

> we could remove them; but then we need to add new tests

What new tests would we need?

vogel:
Updated

MonsieurNicolas:
>    an invalid (as per isValid) tx is still invalid during apply (and the same way) - this is not true anymore (and this is causing us to check for many strange cases). Transaction can be invalid in different way or even valid during apply (as in case of AUTH_EXTRA) bugs.
    a valid tx can fail later that is OK, and doing the mentioned change wont affect this codepath at all
    instead test would pass in the expected pair of results for both checkValid and apply and maybe the list of expected side effects
We do that already with validateTxResults, but it ignores the applyResult is validationResult are not OK.

As per my previous suggestion: because checkValid and apply don't return the same thing in some (rare) cases, we can just introduce a new function that checks the values of what is being returned.
As for `validateTxResults`, I don't know what is easier: I imagine that it's the same change. We need to add a new function similar to `validateTxResults` that takes the expected results for both `checkValid` and `apply` and force the code base to use this one when the return codes are "non trivial".
I think the existing `validateTxResults` should `REQUIRE` that if `checkValid` fails it returns the same thing than `apply` (if it doesn't tests should use the new function) - this also allows to remove all the weird code from `applyCheck` that tries deal with this.

> we could remove them; but then we need to add new tests
What new tests would we need?

This was in the context of "removing checks" - the new tests that we'd need would need to cover all possible ways that `checkValid` and `apply` can return different values. I don't think we need to add those new tests if we instead change `validateTxResults` / `applyCheck` to bomb out if they are called in such situation (and use the new function for tests that are triggering different error codes in valid vs apply)

vogel:
New version of applyCheck can do it, also it can do what validateTxResults did.

src/test/TxTests.cpp
```diff
@@ -196,48 +202,29 @@ applyCheck(TransactionFramePtr tx, Application& app, bool checkSeqNum)
             tx->getResult().result.code(txINTERNAL_ERROR);
         }
         REQUIRE((!res || tx->getResultCode() == txSUCCESS));
-
-        if (!check)
+        if (applyResult)
```
MonsieurNicolas: I don't think this is what we want as this still leaves room for unchecked behavior:
the reason we have this function is to ensure that `checkValid` is consistent with `apply` (while having sane default checks that allow us to not have test code too bloated).

Right now the code still allows for conditions where we ignore the value of `checkResult` (`applyResult` set but in the case `!check` we don't check for anything if `validationResult` is not set).

Basically: if either `validationResult` or `applyResult` are not set, we have to have very strict checks in place where we have 100% confidence in the correctness of the results (so the default checks passing means we have a non exceptional behavior).

vogel: Now it is not possible to set only `validationResult` or `applyResult`, so that should be solved.

graydon:
The underlying repository of this PR was deleted. Closing.

PR #2153, opened 2019-06-18
Bug 2152 move quorum intersection checker

### LINKED ISSUE #2152: code issue: quorum intersection code depends on "app" and "herder"
I missed this during code review:

the "scp" folder is used by other projects outside of stellar and as such we try to keep it as independent as possible from the rest of stellar-core.

The way it is coded, it adds a dependency on "Herder" and on "Config".

I think that the simplest thing to do is to move those files into the "Herder" folder; there might be something cleaner to do (maybe using some new types), but I am not sure it's worth it at this time.

graydon:
Fixes #2152, should land only after #2151 since it extends the msvc fixes there.


MonsieurNicolas:
r+ 0deffc91df1a01c84c7dd9c66784aaf0a39ba3bd

graydon:
@latobarita: retry

MonsieurNicolas:
r+ f6a6567d1c33a732e3870612bfa7aa39ea17ecd9

Issue #1376, opened 2017-11-07
Investigate using afl (fuzzer) to derive transaction tests


MonsieurNicolas:
Right now we only use fuzzer builds to test overlay and hand craft transactions for testing the transaction subsystem.
As we are adding invariants, we should investigate what it would take to use afl to generate interesting transactions and/or transaction sets.
Some hurdles that need to be overcome:
* get passed signature verification - we can disable encryption (AFL builds are special anyways)
* catching unexpected results - invariants should help but that might not be enough?
* deal with the number of possible accounts - automatically map accounts based on the first byte of the public key (for example), with some number of pre-populated accounts in interesting configurations (with trustlines, without, etc).

MonsieurNicolas:
As of https://github.com/stellar/stellar-core/pull/2823 this is mostly implemented.

Areas of improvement (probably in order of priority):
 - [ ] the "setup" code that creates a few accounts and order book seems wrong.
  * The setup is unsafe. Account creation is done directly (should be done with operations) and the rest is setup by applying a transaction but is not checked for success so it's possible that the setup code leaves the ledger in a bad state or a state that is not the one we expect.
  * Balances seem wrong. The numbers are way too big, and require the fuzzer to work on many bits. Numbers should be selected so that "interesting things" can happen with a few bits, maybe 5 or 6 bits is enough (so balances setup with less than 64 stroops, right now they're setup with 100k*10e7=10e9 stroops).
  * Trustlines and orders seem "uninteresting". Setup is fully symetrical. No variations on issuer/trustline flags.
  * Missing other ledger entries like `BalanceEntry`
  - [ ] Handling of certain types like `LedgerKey` should be specialized
  * right now it's pretty much impossible for the fuzzer to generate operations that will operate on `LedgerKey` (like sponsorship), because the space is just too big.
    * a possible alternative could be to map a `LedgerKey` into 1(?) byte, and use that number as an index in the list of all `LedgerKey`s known when the operation gets executed. This can be done in a few ways:
      * The simplest way could be to just incorporate sponsorship into the setup (so have some accounts, trustlines, etc sponsored in the original Ledger state).
      * a more advanced way might be to split up the code that maps "compact binary form" and "in memory xdr". Where the first pass would do most of what it does today, and the second pass, done between operation application, would be to update some fields in a deterministic way.
  - [ ] The data generated to seed the fuzzer via the `gen-fuzz` is very low quality
[...]

PR #2155, opened 2019-06-18
Fuzzing Revival: prepwork for transaction fuzzing

robertDurst:
# Description

This PR lays the groundwork for future improvements and innovations to the fuzzing test suite. Specifically, these changes accomplish three things:
1. allow for deterministic `randombytes_buf` via conditional inclusion preprocessor directive: **FUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION**
2. fixes existing, non-terminating overlay fuzzing via replacement of `crank`  with `crankSome` and manual `shutdown` of apps
3. moves fuzzing code from `/test` to subdirectory `/test/fuzz`

Related to #1376 

# Checklist
- [x] Compiles
- [x] Ran all tests

### DIFF
--- a/src/crypto/Random.cpp
+++ b/src/crypto/Random.cpp
@@ -3,6 +3,7 @@
 // of this distribution or at http://www.apache.org/licenses/LICENSE-2.0
 
 #include "crypto/Random.h"
+#include "util/Math.h"
 #include <sodium.h>
 
 #ifdef MSAN_ENABLED
@@ -13,7 +14,16 @@ std::vector<uint8_t>
 randomBytes(size_t length)
 {
     std::vector<uint8_t> vec(length);
+
+#ifdef FUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION
+    std::uniform_int_distribution<unsigned short> dist(0, 255);
+    std::generate(vec.begin(), vec.end(), [&]() {
+        return static_cast<uint8_t>(dist(stellar::gRandomEngine));
+    });
+#else
     randombytes_buf(vec.data(), length);
+#endif
+
 #ifdef MSAN_ENABLED
     __msan_unpoison(out.key.data(), out.key.size());
 #endif
--- a/src/overlay/test/LoopbackPeer.cpp
+++ b/src/overlay/test/LoopbackPeer.cpp
@@ -113,8 +113,7 @@ LoopbackPeer::drop(std::string const& reason, DropDirection direction, DropMode)
                                  ? Peer::DropDirection::REMOTE_DROPPED_US
                                  : Peer::DropDirection::WE_DROPPED_REMOTE,
                              Peer::DropMode::IGNORE_WRITE_QUEUE);
-
-            },
+                        },
             "LoopbackPeer: drop");
     }
 }
@@ -294,6 +293,13 @@ void
 LoopbackPeer::setCorked(bool c)
 {
     mCorked = c;
+}
+
+void
+LoopbackPeer::clearInAndOutQueues()
+{
+    mOutQueue.clear();
+    mInQueue = {};
 }
 
 bool
--- a/src/overlay/test/LoopbackPeer.cpp
+++ b/src/overlay/test/LoopbackPeer.cpp
@@ -113,7 +113,7 @@ LoopbackPeer::drop(std::string const& reason, DropDirection direction, DropMode)
                                  ? Peer::DropDirection::REMOTE_DROPPED_US
                                  : Peer::DropDirection::WE_DROPPED_REMOTE,
                              Peer::DropMode::IGNORE_WRITE_QUEUE);
-                        },
+            },
             "LoopbackPeer: drop");
     }
 }
@@ -299,7 +299,7 @@ void
 LoopbackPeer::clearInAndOutQueues()
 {
     mOutQueue.clear();
-    mInQueue = {};
+    mInQueue = std::queue<xdr::msg_ptr>();
 }
 
 bool
--- a/src/simulation/Simulation.cpp
+++ b/src/simulation/Simulation.cpp
@@ -212,6 +212,26 @@ Simulation::addLoopbackConnection(NodeID initiator, NodeID acceptor)
     }
 }
 
+std::shared_ptr<LoopbackPeerConnection>
+Simulation::getLoopbackConnection(NodeID const& initiator,
+                                  NodeID const& acceptor)
+{
+    auto it = std::find_if(
+        std::begin(mLoopbackConnections), std::end(mLoopbackConnections),
+        [&](std::shared_ptr<LoopbackPeerConnection> const& conn) {
+            return conn->getInitiator()
+                           ->getApp()
+                           .getConfig()
+                           .NODE_SEED.getPublicKey() == initiator &&
+                   conn->getAcceptor()
+                           ->getApp()
+                           .getConfig()
+                           .NODE_SEED.getPublicKey() == acceptor;
+        });
+
+    return it == std::end(mLoopbackConnections) ? nullptr : *it;
+}
+
 void
 Simulation::dropLoopbackConnection(NodeID initiator, NodeID acceptor)
 {
--- a/src/test/fuzz.cpp
+++ b/src/test/fuzz.cpp
@@ -140,7 +140,12 @@ fuzz(std::string const& filename, el::Level logLevel,
         },
         std::chrono::milliseconds{500}, false);
 
+// "To make this work, the library and this shim need to be compiled in LLVM
+// mode using afl-clang-fast (other compiler wrappers will *not* work)."
+// -- AFL docs
+#ifdef AFL_LLVM_MODE
     while (__AFL_LOOP(PERSIST_MAX))
+#endif // AFL_LLVM_MODE
     {
         XDRInputFileStream in(MAX_MESSAGE_SIZE);
         in.open(filename);
--- a/src/test/fuzz.cpp
+++ b/src/test/fuzz.cpp
@@ -147,6 +147,18 @@ fuzz(std::string const& filename, el::Level logLevel,
         StellarMessage msg;
         while (tryRead(in, msg))
         {
+            // HELLO, AUTH and ERROR_MSG messages cause the connection between
+            // the peers to drop. Since peer connections are only established
+            // preceding the persistent loop, a dropped peer is not only
+            // inconvenient, it also confuses the fuzzer. Consider a msg A sent
+            // before a peer is dropped and after a peer is dropped. The two,
+            // even though the same message, will take drastically different
+            // execution paths -- the fuzzer's main metric for determinism
+            // (stability) and binary coverage.
+            if (msg.type() == HELLO || msg.type() == AUTH ||
+                msg.type() == ERROR_MSG)
+                continue;
+
             LOG(INFO) << "Fuzzer injecting message." << msgSummary(msg);
 
             seedRandomness();
@@ -193,6 +205,14 @@ genfuzz(std::string const& filename)
         try
         {
             StellarMessage m(gen(10));
+            // As more thoroughly explained above, we filter these messages
+            // since they cause peers to drop, leading to non-deterministic
+            // message injections, confusing the fuzzer.
+            while ((m.type() == HELLO || m.type() == AUTH ||
+                    m.type() == ERROR_MSG))
+            {
+                m = gen(10);
+            }
             out.writeOne(m);
             LOG(INFO) << "Message " << i << ": " << msgSummary(m);
         }
--- a/src/test/fuzz.cpp
+++ b/src/test/fuzz.cpp
@@ -69,6 +69,13 @@ tryRead(XDRInputFileStream& in, StellarMessage& m)
     }
 }
 
+void
+seedRandomness()
+{
+    srand(1);
+    gRandomEngine.seed(1);
+}
+
 #define PERSIST_MAX 10000000
 #define INITIATOR 1
 #define ACCEPTOR 0
@@ -80,6 +87,8 @@ fuzz(std::string const& filename, el::Level logLevel,
     Logging::setLogLevel(logLevel, nullptr);
     LOG(INFO) << "Fuzzing stellar-core " << STELLAR_CORE_VERSION;
     LOG(INFO) << "Fuzz input is in " << filename;
+
+    seedRandomness();
 
     Hash networkID = sha256(getTestConfig().NETWORK_PASSPHRASE);
     auto confGen = [](int instanceNumber) {
@@ -139,6 +148,8 @@ fuzz(std::string const& filename, el::Level logLevel,
         while (tryRead(in, msg))
         {
             LOG(INFO) << "Fuzzer injecting message." << msgSummary(msg);
+
+            seedRandomness();
 
             auto nodeids = simulation->getNodeIDs();
 
--- a/src/test/fuzz.cpp
+++ b/src/test/fuzz.cpp
@@ -11,9 +11,11 @@
 #include "overlay/OverlayManager.h"
 #include "overlay/TCPPeer.h"
 #include "overlay/test/LoopbackPeer.h"
+#include "simulation/Simulation.h"
 #include "test/test.h"
 #include "util/Fs.h"
 #include "util/Logging.h"
+#include "util/Math.h"
 #include "util/Timer.h"
 #include "util/XDRStream.h"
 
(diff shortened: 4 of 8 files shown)


src/test/fuzz/fuzz.cpp
```diff
@@ -131,13 +133,14 @@ fuzz(std::string const& filename, el::Level logLevel,
 restart:
 {
     VirtualClock clock;
-    Application::pointer app1 = Application::create(clock, cfg1);
-    Application::pointer app2 = Application::create(clock, cfg2);
+    auto app1 = createTestApplication(clock, cfg1);
```
MonsieurNicolas: I was expecting to start to see changes related to proper randomization both before creating configs and in the "main fuzzer loop", something (for now), like:
```
            srand(1);
            gRandomEngine.seed(1);
```

src/test/fuzz/fuzz.cpp
```diff
@@ -155,8 +158,13 @@ fuzz(std::string const& filename, el::Level logLevel,
     }
     while (loop.getAcceptor()->isConnected())
     {
-        clock.crank(true);
+        testutil::crankSome(clock);
     }
+
+    // manually shutdown everything
+    app1->getWorkScheduler().shutdown();
```
marta-lokhova: You need to crank the clock properly until WorkScheduler is in `BasicWork::State::WORK_ABORTED` state (see `testutil::shutdownWorkScheduler`)

src/test/fuzz.cpp
```diff
@@ -1,194 +0,0 @@
-// Copyright 2015 Stellar Development Foundation and contributors. Licensed
-// under the Apache License, Version 2.0. See the COPYING file at the root
-// of this distribution or at http://www.apache.org/licenses/LICENSE-2.0
-
-#include "util/asio.h"
```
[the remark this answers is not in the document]
robertDurst: Good point. Removed.

src/test/fuzz/fuzz.cpp
robertDurst: Ah ok. Added.

src/test/fuzz/fuzz.cpp
robertDurst: The original issue, and why I needed to change any code in the first place is that there are two applications, both using the same clock. Upon executing the current code, at the end of the scope, one app is able to successfully shutdown:

```
 <fuzz> [Work TRACE] Shutting down: work-scheduler [BasicWork.cpp:52]                                                               
 <fuzz> [Work DEBUG] work work-scheduler : WORK_WAITING -> WORK_ABORTING                                                            
 <fuzz> [Work TRACE] work-scheduler: waiting for children to abort. [Work.cpp:92]                                                   
 <fuzz> [Work TRACE] Abort progress for work-scheduler: done [BasicWork.cpp:342]                                                    
 <fuzz> [Work TRACE] resetting work-scheduler [BasicWork.cpp:133]
 <fuzz> [Work DEBUG] work work-scheduler : WORK_ABORTING -> WORK_ABORTED                                                            
 <fuzz> [default DEBUG] Joining 11 worker threads
 <fuzz> [default DEBUG] Joined all 11 threads
 <fuzz> [default INFO] Application destroyed
```

However, the second one just stalls:
```
<fuzz> [Work TRACE] Shutting down: work-scheduler [BasicWork.cpp:52]                                                               
<fuzz> [Work DEBUG] work work-scheduler : WORK_WAITING -> WORK_ABORTING   
```

The two solutions I came up with for this (one Marta and I discussed and one I worked out with the help of Graydon) is to:

1. Have two clocks. Two clocks will not stall on destruction, but based on comments in the code, is not recommended for simulations such as this.

2. Copy the logic from `testutil::shutdownWorkScheduler`. **In my revised code, I have decided to go with this approach**. Basically here, I just manually shutdown the two (not 100% if the stalling app is the same one each time) and then loop&crank until everything is done shutting down.

robertDurst: Furthermore, *ideally I would not want to start up and shutdown every single time I inject fuzzed messages.* This will very likely be fixed in the PR following this one -- and removal of the `goto` :)

src/crypto/Random.cpp
```diff
@@ -13,7 +14,15 @@ std::vector<uint8_t>
 randomBytes(size_t length)
 {
     std::vector<uint8_t> vec(length);
+
+#ifdef FUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION
+    std::uniform_int_distribution<unsigned char> dist;
+    auto const SEED = dist(stellar::gRandomEngine);
+    randombytes_buf_deterministic(vec.data(), length, &SEED);
```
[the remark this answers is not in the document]
robertDurst: Thanks for catching this. I will make sure to check out the docs and ensure correct input types in the future. Updated with your recommendation.

src/test/fuzz.cpp
```diff
@@ -130,9 +136,12 @@ fuzz(std::string const& filename, el::Level logLevel,
 
 restart:
 {
+    srand(1);
```
MonsieurNicolas: extract this to a function to avoid copy/paste

src/test/fuzz.cpp
```diff
@@ -157,6 +169,16 @@ fuzz(std::string const& filename, el::Level logLevel,
     {
         clock.crank(true);
     }
+
```
MonsieurNicolas: mmm, the main loop `tryRead` is a bit weird: I would expect to see not just `clock.postToCurrentCrank` but also a call to `crank` to consume the message: without this, messages are queued somewhere and there is no guarantee that they will be executed in the same order... 
I think a better option to deal with this problem (and shutdown), would be to use `Simulation` (with 2 nodes), and call in this loop something like `crankForAtMost(0.5 seconds, false)`

src/test/fuzz.cpp
```diff
@@ -157,6 +169,16 @@ fuzz(std::string const& filename, el::Level logLevel,
...
+    // properly shut everything down
+    app1->getWorkScheduler().shutdown();
+    app2->getWorkScheduler().shutdown();
+    while (
```
MonsieurNicolas: this is ugly/weird, sharing the clock between applications is very error prone and I suspect leads to all sorts of random behaviors.
Switch to using the `Simulation` class

src/test/fuzz.cpp
```diff
@@ -130,9 +136,12 @@ fuzz(std::string const& filename, el::Level logLevel,
 
 restart:
 {
+    srand(1);
+    gRandomEngine.seed(1);
+
     VirtualClock clock;
-    Application::pointer app1 = Application::create(clock, cfg1);
-    Application::pointer app2 = Application::create(clock, cfg2);
+    auto app1 = createTestApplication(clock, cfg1);
```
MonsieurNicolas: why is app creation in the `main loop` instead of being before `restart`? Question in general, maybe it's the right thing to do for overlay; I am asking about the requirements for tests executed per message (in the `while` loop) vs per "session" (for lack of better term) inside the `restart` block.

robertDurst: Moved setup before persistence loop and use LLVM mode allowing me to utilize ` while (__AFL_LOOP(PERSIST_MAX))`. Significant increase in exec/sec without compromising stability (determinism)

src/test/fuzz.cpp
robertDurst: Done

src/test/fuzz.cpp
robertDurst: Crank now after posting `sendMessage` to the crank. Thus messages actually send.

src/test/fuzz.cpp
robertDurst: Done

src/crypto/Random.cpp
```diff
@@ -13,7 +14,15 @@ std::vector<uint8_t>
 randomBytes(size_t length)
 {
     std::vector<uint8_t> vec(length);
+
+#ifdef FUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION
+    std::uniform_int_distribution<unsigned short> dist(0, 255);
+    std::generate(vec.begin(), vec.end(),
+                  [&]() { return dist(stellar::gRandomEngine); });
```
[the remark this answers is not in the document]
robertDurst: Done

src/test/fuzz/fuzz.cpp
robertDurst: Dealt with by utilizing simulation. See below comments by Nicolas.

robertDurst:
For completeness and for anyone who hasn't followed the discussion & progression of this PR, the scope has changed since the initial PR description.

# Updated Description

This PR still lays the groundwork for future improvements and innovations to the fuzzing test suite, by  improving upon the current overlay test harness. It does not just get the harness to stop crashing (original PR), but improves the determinism of the fuzz test allowing for utilization of [AFL's persistent mode](https://lcamtuf.blogspot.com/2015/06/new-in-afl-persistent-mode.html), increasing exec/sec from a measly *50* to *2000+*.

**Determinism improvements:**
1. allow for deterministic randombytes_buf via conditional inclusion preprocessor directive: FUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION
2. utilize a Simulation object for proper synchronization of two Applications instead of a single, shared, VirtualClock
3. clear persistent collections between executions:
   * LoopbackPeer in and out message queues
   * VirtualClock events
4. Ignore messages that may cause the peer connection to crash (problematic since this connection is established before the persistent loop and thus won't be re-established until many, many more messages are sent)

src/simulation/Simulation.cpp
```diff
@@ -672,4 +672,43 @@ Simulation::metricsSummary(string domain)
...
+
+        // clear all queues and cancel all events
+        initiator->clearInAndOutQueues();
+        acceptor->clearInAndOutQueues();
+
+        while (initiator->getApp().getClock().cancelAllEvents())
+            ;
+        while (acceptor->getApp().getClock().cancelAllEvents())
+            ;
+        while (mClock.cancelAllEvents())
```
MonsieurNicolas: why do you need to call `cancelAllEvents` on `mClock`?

src/test/fuzz.cpp
```diff
@@ -43,26 +45,25 @@
 namespace stellar
 {
 
-struct CfgDirGuard
+Config
+confGen(int instanceNumber)
```
MonsieurNicolas: why is this function in the global namespace? It's only used below, right?

src/test/fuzz.cpp
```diff
@@ -88,9 +89,16 @@ tryRead(XDRInputFileStream& in, StellarMessage& m)
...
-#define PERSIST_MAX 1000
-static unsigned int persist_cnt = 0;
+void
+seedRandomness()
+{
+    srand(1);
+    gRandomEngine.seed(1);
+}
 
+#define PERSIST_MAX 1000
```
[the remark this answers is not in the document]
robertDurst: This is the recommended numerical value from [AFL llvm mode docs](https://github.com/mirrorer/afl/tree/master/llvm_mode). 

> The number passed to __AFL_LOOP() controls the maximum number of
     iterations before the loop exits and the program is allowed to
     terminate normally. This limits the impact of accidental memory leaks
     and similar hiccups. 1000 is a good starting point, and going much higher
    increases the likelihood of hiccups without giving you any real performance benefits.

From what I have seen online (blogs, other projects, the docs) 1000 is a very commonly used magic number.

src/simulation/Simulation.cpp
robertDurst: Removed. Originally, I believed this to be a source of non-determinism -- events persisting between executions -- however, now that I have looked closer it looks like the issue was only with `LoopbackPeer` in/out message queues.

src/test/fuzz.cpp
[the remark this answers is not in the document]
MonsieurNicolas: OK, so it's actually not a recommended value: if you think about it, it's a way to amortize startup and setup cost
Reason I was asking about it is that it seems very low compared to our startup cost. If the goal is to avoid memory leaks, it sounds like 1000 is a setting maybe better suited for large payloads (like when people fuzz images). As a message is in the order of a few hundred bytes, I imagine that we can go a lot higher on this number (which should increase exploration rate without hurting stability).

src/test/fuzz.cpp
robertDurst: Done -- use a lambda instead now.

src/simulation/Simulation.cpp
robertDurst: I do not, remove -- double checked and events only persist on `acceptor` and `initiator`, not the idle app.

src/test/fuzz.cpp
```diff
@@ -43,26 +45,25 @@
...
+    cfg.ARTIFICIALLY_ACCELERATE_TIME_FOR_TESTING = true;
+    cfg.MANUAL_CLOSE = true;
+    cfg.CATCHUP_COMPLETE = false;
+    cfg.CATCHUP_RECENT = 0;
+    cfg.ARTIFICIALLY_GENERATE_LOAD_FOR_TESTING = false;
+    cfg.ARTIFICIALLY_SET_CLOSE_TIME_FOR_TESTING = 0;
+    cfg.PUBLIC_HTTP_PORT = false;
+    cfg.WORKER_THREADS = 1;
+    cfg.QUORUM_INTERSECTION_CHECKER = false;
+    cfg.ENTRY_CACHE_SIZE = 0;
```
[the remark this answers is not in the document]
robertDurst: Done

src/test/fuzz.cpp
```diff
@@ -43,26 +45,25 @@
...
-        clean(mConfig.BUCKET_DIR_PATH);
-    }
-};
+    Config cfg = getTestConfig(instanceNumber);
+    cfg.ARTIFICIALLY_ACCELERATE_TIME_FOR_TESTING = true;
+    cfg.MANUAL_CLOSE = true;
+    cfg.CATCHUP_COMPLETE = false;
+    cfg.CATCHUP_RECENT = 0;
+    cfg.ARTIFICIALLY_GENERATE_LOAD_FOR_TESTING = false;
+    cfg.ARTIFICIALLY_SET_CLOSE_TIME_FOR_TESTING = 0;
```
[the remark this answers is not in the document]
robertDurst: Done

src/test/fuzz.cpp
```diff
@@ -43,26 +45,25 @@
...
+    cfg.PUBLIC_HTTP_PORT = false;
+    cfg.WORKER_THREADS = 1;
+    cfg.QUORUM_INTERSECTION_CHECKER = false;
+    cfg.ENTRY_CACHE_SIZE = 0;
+    cfg.BEST_OFFERS_CACHE_SIZE = 0;
+    cfg.PREFETCH_BATCH_SIZE = 0;
```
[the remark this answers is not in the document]
robertDurst: Yes, confirmed it still works -- using `LoopbackConnection` is more of a pseudo connection, where peers are connected via weak references to each other and messages are *sent* rather manually by placing in remote's `mInQueue` directly. 

Not quite sure of other useful Overlay configs. The one I did add was: `PREFERRED_PEERS_ONLY=false`

src/simulation/Simulation.cpp
```diff
@@ -672,4 +672,43 @@ Simulation::metricsSummary(string domain)
     }
     return out.str();
 }
+
+void
+Simulation::sendMessageAndCrank(StellarMessage& msg, NodeID initiator,
```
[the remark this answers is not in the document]
robertDurst: Fixed -- all I needed was a way to get a `LoopbackPeerConnection` so I removed the `sendMessageAndCrank` in favor of a simple get method.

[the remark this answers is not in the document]
robertDurst: Done -- I needed some way of accessing a `LoopbackPeerConnection`m, so swapped this for a get method.

robertDurst:
Rebased. Should be good for another (final?) review.

src/overlay/test/LoopbackPeer.cpp
```diff
@@ -296,6 +295,14 @@ LoopbackPeer::setCorked(bool c)
     mCorked = c;
 }
 
+void
+LoopbackPeer::clearInAndOutQueues()
+{
+    mOutQueue.clear();
+    std::queue<xdr::msg_ptr> cleared;
```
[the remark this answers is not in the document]
robertDurst: Done

robertDurst:
Rebased and commits cleaned up.

MonsieurNicolas:
looks good

MonsieurNicolas:
r+ 9ccd0a39bb28e1b2606c0936a087d72a373db60b

MonsieurNicolas:
needs clang-format

robertDurst:
> needs clang-format

Done. I thought I had it all setup in my ide, but I am now questioning things.

Also, there are some files I did not touch, but when clang-format'd change. Will this also cause the build to fail?

MonsieurNicolas:
r+ 0ee9a16d1904bcd9394c60f1a29cd7da3ac78765

MonsieurNicolas:
you may be running the wrong version of clang-format @robertDurst

robertDurst:
Ok, looks like build fails because `AFL_LOOP` requires compilation with clang & `./configure --enable-afl`. May need some sort of `AFL_BUILD_MODE` or something to hide it behind.

robertDurst:
Ok, hid `AFL_LOOP` behind `AFL_LLVM_MODE` conditional inclusion directive and also tried a different way of clearing the queue, hopefully a way Travis finds acceptable.

MonsieurNicolas:
r+ 6e3627316f1e07f09697510b0e5aa2f3ae86bb4a

PR #2158, opened 2019-06-19
redo pending transactions

### LINKED ISSUE #1556: pending transactions in herder is not managed properly
When an account submits too many transactions, it looks like we only drop the oldest batch of transactions for that account:
when this happens, we keep the old batches around, with a higher sequence number, and because of that disallow resubmit of the transactions that got dropped (with a bad sequence number error).
This will probably manifest itself as a window where accounts are just in a broken state (until `triggerNextLedger` is called that ends up deleting all those invalid transactions).

We should refactor the code that deals with this (currently split between `struct TxMap` and `mPendingTransactions`) such that we track accounts across generations properly (we should not need to "recalculate" anything if we do it right too).

vogel:

# Description

Resolves #1556

With this change each behavior of transaction queue for account changes. All transactions for one account have the same "pending depth" now, so it is no longer possible that transactions with lower seqnum is removed or banned and trasaction with higher seqnum remains in queue. That situation was buggy, because account was not able to submit more valid transaction until all of them were removed from pending queue (that usually took only one ledger, so 5 seconds).

# Checklist
- [x] Compiles
- [x] Ran all tests
- [] If change impacts performance, include supporting evidence per the [performance document](https://github.com/stellar/stellar-core/blob/master/performance-eval/performance-eval.md)

### DIFF
--- a/src/herder/TransactionQueue.cpp
+++ b/src/herder/TransactionQueue.cpp
@@ -14,79 +14,14 @@
 #include <lib/util/format.h>
 #include <medida/meter.h>
 #include <medida/metrics_registry.h>
+#include <numeric>
 
 namespace stellar
 {
 
-static uint64_t
-countTxs(TransactionQueue::AccountTxMap const& acc)
-{
-    uint64_t sz = 0;
-    for (auto const& a : acc)
-    {
-        sz += a.second->mTransactions.size();
-    }
-    return sz;
-}
-
-static std::shared_ptr<TransactionQueue::TxMap>
-findOrAdd(TransactionQueue::AccountTxMap& acc, AccountID const& aid)
-{
-    std::shared_ptr<TransactionQueue::TxMap> txmap;
-    auto i = acc.find(aid);
-    if (i == acc.end())
-    {
-        txmap = std::make_shared<TransactionQueue::TxMap>();
-        acc.emplace(std::make_pair(aid, txmap));
-    }
-    else
-    {
-        txmap = i->second;
-    }
-    return txmap;
-}
-
-bool
-TransactionQueue::isBanned(Hash const& hash) const
-{
-    return std::any_of(
-        std::begin(mBannedTransactions), std::end(mBannedTransactions),
-        [&](std::unordered_set<Hash> const& transactions) {
-            return transactions.find(hash) != std::end(transactions);
-        });
-}
-
-void
-TransactionQueue::TxMap::addTx(TransactionFramePtr tx)
-{
-    auto const& h = tx->getFullHash();
-    if (mTransactions.find(h) != mTransactions.end())
-    {
-        return;
-    }
-    mTransactions.emplace(std::make_pair(h, tx));
-    mCurrentQInfo.mMaxSeq = std::max(tx->getSeqNum(), mCurrentQInfo.mMaxSeq);
-    mCurrentQInfo.mTotalFees += tx->getFeeBid();
-}
-
-void
-TransactionQueue::TxMap::recalculate()
-{
-    mCurrentQInfo.mMaxSeq = 0;
-    mCurrentQInfo.mTotalFees = 0;
-    for (auto const& pair : mTransactions)
-    {
-        mCurrentQInfo.mMaxSeq =
-            std::max(pair.second->getSeqNum(), mCurrentQInfo.mMaxSeq);
-        mCurrentQInfo.mTotalFees += pair.second->getFeeBid();
-    }
-}
-
 TransactionQueue::TransactionQueue(Application& app, int pendingDepth,
                                    int banDepth)
-    : mApp(app)
-    , mPendingTransactions(pendingDepth)
-    , mBannedTransactions(banDepth)
+    : mApp(app), mPendingDepth(pendingDepth), mBannedTransactions(banDepth)
 {
     for (auto i = 0; i < pendingDepth; i++)
     {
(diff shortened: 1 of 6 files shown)


src/herder/TransactionQueue.h
```diff
@@ -43,45 +43,54 @@ class TransactionQueue
...
     AccountTxQueueInfo
     getAccountTransactionQueueInfo(AccountID const& accountID) const;
 
     int countBanned(int index) const;
     bool isBanned(Hash const& hash) const;
+
     std::shared_ptr<TxSetFrame> toTxSet(Hash const& lclHash) const;
 
   private:
+    using PendingTransactions =
```
MonsieurNicolas: You need to put some comments on what is going on here: I had to stare for a while and reverse engineer what you're trying to do.

src/herder/TransactionQueue.cpp
```diff
@@ -216,21 +145,33 @@ TransactionQueue::shift()
     mBannedTransactions.pop_back();
     mBannedTransactions.emplace_front();
 
+    auto sizes = std::vector<int>{};
```
MonsieurNicolas: From what I understand: you moved from having a generational `deque` of maps of transactions to a single map so that you can erase all transactions for a given acount if it reaches max age?

I think this is wrong: imagine that there is an account that submits a transaction every second. What is going to happen is that those queues are going to grow but only a subset gets included in each ledger, yet the "age" associated with this account keeps growing, and after 4 ledgers, the queue gets obliterated even though the transactions were getting processed.

src/herder/TransactionQueue.cpp
```diff
@@ -123,91 +58,85 @@ TransactionQueue::tryAdd(TransactionFramePtr tx)
...
             continue;
         }
 
-        std::unordered_set<std::shared_ptr<TxMap>> toRecalculate;
-
-        for (auto const& tx : dropTxs)
+        auto& txs = it.first->second.mTransactions;
+        auto feeBid =
+            std::accumulate(it.second, std::end(txs), 0,
+                            [](int fee, TransactionFramePtr const& tx) {
```
MonsieurNicolas: overflow: need to use `int64` when adding fees

src/herder/TransactionQueue.cpp
```diff
@@ -123,91 +58,85 @@ TransactionQueue::tryAdd(TransactionFramePtr tx)
...
+                                return fee + tx->getFeeBid();
+                            });
+        it.first->second.mTotalFees -= feeBid;
+        txs.erase(it.second, std::end(txs));
```
MonsieurNicolas: should those be banned as well?

src/herder/TransactionQueue.cpp
vogel: You are right. We could fix it by resetting all age values for accounts that had their transaction processed in given ledger. I think it makes sense, as we can assume that those accounts have more valid transactions worth processing in their queues and even one bad transaction would clear that queue, so it should be safe.

What do you think?

src/herder/TransactionQueue.cpp
vogel: This function is used in two cases. First, when transaction in TransactionQueue became invalid for some reason. Second, when transaction has been applied on ledger.

I think in first case we can ban all removed transactions as well, but in second case it will result in `ADD_STATUS_TRY_AGAIN_LATER` instead of `ADD_STATUS_ERROR` when someone try to submit just applied transaction again. I don't think that it is a huge problem, or a problem at all.

So I think we should ban those as well in both cases. If you agree, I'll implement that shortly.

src/herder/TransactionQueue.h
vogel: I will add comments when all other issues are resolved, so I will not have to update it depending on chosen solutions.

src/herder/TransactionQueue.cpp
MonsieurNicolas: I think this will be good enough, yes: we'll have to make sure that this "reset" only happens when removing transactions that were in a ledger, not when invalid transactions are removed.

src/herder/TransactionQueue.cpp
MonsieurNicolas: yeah both cases are potentially bad for the validator, so banning seems to be the right strategy

vogel:
Updated

src/herder/TransactionQueue.cpp
```diff
@@ -123,91 +58,117 @@ TransactionQueue::tryAdd(TransactionFramePtr tx)
...
 
-        for (auto txm : toRecalculate)
+void
+TransactionQueue::ban(std::vector<TransactionFramePtr> const& dropTxs)
+{
+    auto& bannedFront = mBannedTransactions.front();
+    for (auto const& tx : dropTxs)
+    {
+        auto extracted = extract(tx);
+        for (auto const& tx : extracted.second)
```
MonsieurNicolas: don't shadow variables

src/herder/TransactionQueue.cpp
```diff
@@ -123,91 +58,117 @@ TransactionQueue::tryAdd(TransactionFramePtr tx)
...
-                       });
+    auto const& acc = tx->getSourceID();
+    auto accIt = mPendingTransactions.find(acc);
+    if (accIt == std::end(mPendingTransactions))
+    {
+        return {std::end(mPendingTransactions), {}};
+    }
+
+    auto& txs = accIt->second.mTransactions;
+    auto distance = tx->getSeqNum() - txs.front()->getSeqNum();
```
MonsieurNicolas: no need to overcomplicate this code: just use `find_if` to look for the transaction

src/herder/TransactionQueue.cpp
```diff
@@ -216,21 +177,33 @@ TransactionQueue::shift()
...
+            {
+                bannedFront.insert(toBan->getFullHash());
+            }
+
+            i = mPendingTransactions.erase(i);
+        }
+        else
         {
-            bannedFront.insert(toBan.first);
+            sizes[i->second.mAge] += i->second.mTransactions.size();
```
MonsieurNicolas: need to cast here: `size()` returns `size_t` which is bigger than `int`

src/herder/TransactionQueue.cpp
```diff
@@ -216,21 +177,33 @@ TransactionQueue::shift()
...
+            ++i;
         }
     }
 
-    mPendingTransactions.pop_back();
-    mPendingTransactions.emplace_front();
-
-    for (auto i = 0; i < mPendingTransactions.size(); i++)
+    for (auto i = 0; i < sizes.size(); i++)
```
MonsieurNicolas: don't shadow variables (`i`)

MonsieurNicolas:
Not sure if you saw, but there are some questions for you @vogel

vogel:
Fixed, squashed and rebased. Ready to merge ;)

src/herder/test/TransactionQueueTests.cpp
```diff
@@ -44,93 +64,69 @@ class TransactionQueueTest
...
     {
         auto txSet = mTransactionQueue.toTxSet({});
         auto expectedTxSet = TxSetFrame{{}};
-        for (auto const& tx : expected)
+        for (auto const& accountState : state.mAccountStates)
         {
-            expectedTxSet.add(tx);
+            auto& txs = accountState.mAccountTransactions;
+            auto fees =
+                std::accumulate(std::begin(txs), std::end(txs), 0,
```
MonsieurNicolas: missing include for `std::accumulate` (numeric)

vogel: updated

MonsieurNicolas:
r+ b6fdaceb99d279ba9173bc29f7df8598d46d1c96

PR #2165, opened 2019-06-20
New performance metrics

jonjove:
Added new metrics that provide holistic performance information on various stellar-core subsystems.

### DIFF
--- a/src/bucket/BucketList.cpp
+++ b/src/bucket/BucketList.cpp
@@ -154,9 +154,8 @@ BucketLevel::prepare(Application& app, uint32_t currLedger,
         }
     }
 
-    mNextCurr =
-        FutureBucket(app, curr, snap, shadows, currLedgerProtocol,
-                     BucketList::keepDeadEntries(mLevel), countMergeEvents);
+    mNextCurr = FutureBucket(app, curr, snap, shadows, currLedgerProtocol,
+                             countMergeEvents, mLevel);
     assert(mNextCurr.isMerging());
 }
 
@@ -559,7 +558,7 @@ BucketList::restartMerges(Application& app, uint32_t maxProtocolVersion)
         auto& next = level.getNext();
         if (next.hasHashes() && !next.isLive())
         {
-            next.makeLive(app, maxProtocolVersion, keepDeadEntries(i));
+            next.makeLive(app, maxProtocolVersion, i);
             if (next.isMerging())
             {
                 CLOG(INFO, "Bucket")
--- a/src/bucket/FutureBucket.cpp
+++ b/src/bucket/FutureBucket.cpp
@@ -8,6 +8,7 @@
 #include "util/asio.h"
 
 #include "bucket/Bucket.h"
+#include "bucket/BucketList.h"
 #include "bucket/BucketManager.h"
 #include "bucket/FutureBucket.h"
 #include "crypto/Hex.h"
@@ -16,6 +17,8 @@
 #include "util/LogSlowExecution.h"
 #include "util/Logging.h"
 #include "util/format.h"
+
+#include "medida/metrics_registry.h"
 
 #include <chrono>
 
@@ -26,8 +29,8 @@ FutureBucket::FutureBucket(Application& app,
                            std::shared_ptr<Bucket> const& curr,
                            std::shared_ptr<Bucket> const& snap,
                            std::vector<std::shared_ptr<Bucket>> const& shadows,
-                           uint32_t maxProtocolVersion, bool keepDeadEntries,
-                           bool countMergeEvents)
+                           uint32_t maxProtocolVersion, bool countMergeEvents,
+                           uint32_t level)
     : mState(FB_LIVE_INPUTS)
     , mInputCurrBucket(curr)
     , mInputSnapBucket(snap)
@@ -44,7 +47,7 @@ FutureBucket::FutureBucket(Application& app,
     {
         mInputShadowBucketHashes.push_back(binToHex(b->getHash()));
     }
-    startMerge(app, maxProtocolVersion, keepDeadEntries, countMergeEvents);
+    startMerge(app, maxProtocolVersion, countMergeEvents, level);
 }
 
 void
@@ -253,9 +256,20 @@ FutureBucket::getOutputHash() const
     return mOutputBucketHash;
 }
 
+static std::chrono::seconds
+getAvailableTimeForMerge(Application& app, uint32_t level)
+{
+    auto closeTime = app.getConfig().getExpectedLedgerCloseTime();
+    if (level >= 1)
+    {
+        return closeTime * BucketList::levelHalf(level - 1);
+    }
+    return closeTime;
+}
+
 void
 FutureBucket::startMerge(Application& app, uint32_t maxProtocolVersion,
-                         bool keepDeadEntries, bool countMergeEvents)
+                         bool countMergeEvents, uint32_t level)
 {
     // NB: startMerge starts with FutureBucket in a half-valid state; the inputs
     // are live but the merge is not yet running. So you can't call checkState()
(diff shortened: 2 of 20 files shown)


src/bucket/FutureBucket.cpp
```diff
@@ -253,9 +256,28 @@ FutureBucket::getOutputHash() const
     return mOutputBucketHash;
 }
 
+static double
```
graydon: Maybe use a std::chrono time here rather than double-which-is-a-count-of-something?

src/bucket/FutureBucket.cpp
```diff
@@ -276,37 +298,47 @@ FutureBucket::startMerge(Application& app, uint32_t maxProtocolVersion,
...
+        {"bucket", "merge-time", "level-" + std::to_string(level)});
 
     using task_t = std::packaged_task<std::shared_ptr<Bucket>()>;
-    std::shared_ptr<task_t> task =
-        std::make_shared<task_t>([curr, snap, &bm, shadows, maxProtocolVersion,
-                                  keepDeadEntries, countMergeEvents]() {
+    std::shared_ptr<task_t> task = std::make_shared<task_t>([
+        curr, snap, &bm, shadows, maxProtocolVersion, keepDeadEntries,
+        countMergeEvents, level, timeScope = timer.TimeScope()
+    ]() mutable {
```
graydon: TIL that you can put 'mutable' there. Wonders never cease!

graydon: (Also have initializer expressions in the capture clause? Gosh I need to keep up with new C++)

src/overlay/Peer.cpp
```diff
@@ -743,10 +750,22 @@ Peer::recvTransaction(StellarMessage const& msg)
...
             {
                 // if it's a new transaction, broadcast it
                 mApp.getOverlayManager().broadcastMessage(msg);
+
+                mUniqueBytesRecv.Mark(size);
+            }
+            else
+            {
+                mDuplicateBytesRecv.Mark(size);
+                CLOG(TRACE, "Perf")
```
graydon: As a general thing (we need to be better about this) looking up CLOG(TRACE, ...) is nontrivial cost; it's worth caching trace-being-disabled (which it usually is) or hoisting that out of a thing that's going to be called thousands of times a second.

src/overlay/Peer.cpp
```diff
@@ -743,10 +750,22 @@ Peer::recvTransaction(StellarMessage const& msg)
             // record that this peer sent us this transaction
             mApp.getOverlayManager().recvFloodedMsg(msg, shared_from_this());
 
+            size_t size = xdr::xdr_argpack_size(msg);
             if (recvRes == TransactionQueue::AddResult::ADD_STATUS_PENDING)
             {
                 // if it's a new transaction, broadcast it
                 mApp.getOverlayManager().broadcastMessage(msg);
+
+                mUniqueBytesRecv.Mark(size);
```
MonsieurNicolas: this doesn't seem to be the right place: if we receive a stream of bad (expired) transactions, we're not going to count them against "bad" data.

Easiest might be to just lookup the message in the floodmap, and if it was already there, count it as "waste" (duplicate/not needed) *before* feeding that transaction to herder.

src/overlay/Peer.cpp
```diff
@@ -799,6 +818,20 @@ Peer::recvSCPMessage(StellarMessage const& msg)
     if (res != Herder::ENVELOPE_STATUS_DISCARDED)
     {
         mApp.getOverlayManager().recvFloodedMsg(msg, shared_from_this());
+
```
MonsieurNicolas: same thing than for transactions: we need to do the accounting regardless of what the business logic decides about this SCP message. I imagine that we end up sharing the code.

For example: SCP messages get discarded for all sorts of reason; I'd rather keep track of "wasted" traffic (duplicate, but also propagated for no good reason).

MonsieurNicolas: We also need to do this accounting for the `txSet` and `quorumSet` as we ask peers for it - and we need to know if we're getting one copy back or many copies (because we ask those things to too many peers for the same thing).

As I write this, I realize that maybe this logic can be simplified quite a bit and do something pretty close to "just tracking duplicates" (regardless of what the business logic does):
what we can do is do this type of accounting in `Peer::recvMessage(xdr::msg_ptr const& msg)` instead of relying on the floodmap: all we need to do is keep a small enough cache of `HASH(msg)` (LRU) and consider a message a duplicate if it's not in cache - we can use a shorthash here that should yield "good enough" results. If we want to only do this for some message types, all we need to do is make `Peer::recvMessage(StellarMessage const& stellarMsg)` return a bool (where returning true means "do the duplicate accounting").
We can then return true for the cases `TX_SET`, `TRANSACTION`, `SCP_QUORUMSET`, `SCP_MESSAGE`.

src/bucket/FutureBucket.cpp
MonsieurNicolas: also I guess we need to have access to this as a metric so that we can compute the ratio in reports?

src/bucket/BucketList.cpp
```diff
@@ -154,9 +154,9 @@ BucketLevel::prepare(Application& app, uint32_t currLedger,
         }
     }
 
-    mNextCurr =
-        FutureBucket(app, curr, snap, shadows, currLedgerProtocol,
-                     BucketList::keepDeadEntries(mLevel), countMergeEvents);
+    mNextCurr = FutureBucket(app, curr, snap, shadows, currLedgerProtocol,
```
vogel: now BucketList::keepDeadEntries(mLevel) can be computed inside FutureBucket as well, so it is possible to get rid of one of the arguments here

src/bucket/FutureBucket.cpp
```diff
@@ -253,9 +256,28 @@ FutureBucket::getOutputHash() const
...
+getAvailableTimeForMerge(uint32_t level)
+{
+    double availableTime = 5;
+    if (level == 1)
+    {
+        availableTime = 10;
+    }
+    else if (level > 1)
+    {
+        for (uint32_t i = 0; i < level; ++i)
```
marta-lokhova: I'm not sure what this function is calculating: I _think_ the idea is to estimate how many seconds a future bucket has to perform a merge before it is needed? If so, the formula for that is `availableTime(level i)  = BucketList::levelHalf(level i - 1) * timeToCloseLedger` -- in other words, level i has to promote future bucket and start a new merge every half-size of level i-1. We also don't need to special case level 1, as there isn't anything different about it. I found this comment quite helpful in understanding the mechanics of level spilling: https://github.com/stellar/stellar-core/blob/dc5fab0e6b6ba505a615de521bbf1beff7fd0d62/src/bucket/BucketList.cpp#L475)

src/bucket/FutureBucket.cpp
jonjove: Both done.

src/ledger/LedgerManagerImpl.cpp
```diff
@@ -400,6 +419,17 @@ LedgerManagerImpl::valueExternalized(LedgerCloseData const& ledgerData)
         << ", ops=" << ledgerData.getTxSet()->sizeOp() << ", sv: "
         << stellarValueToString(mApp.getConfig(), ledgerData.getValue()) << "]";
 
+    if (mLastExternalize != std::chrono::steady_clock::time_point())
```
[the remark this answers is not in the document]
MonsieurNicolas: Oh I see so the purpose of this metric is to get the overall time spent between a ledger closing and the end of the next consensus round (right before closing the next ledger)?
I think right now we have the submetrics: time spent in nomination and time spent in ballot protocol, so this would allow us to get the number spent overall (basically the sum of those and the time spent before nomination starts).

So should this metric just be the time spent before nomination starts then (as defined in herder, in some cases "nomination starts" is just short circuited by "ballot protocol starts")?

I think that either we have metrics that represent our SLA (`mLedgerAgeClosed` tracks that we are indeed "closing a ledger every 5 seconds"), or metrics that we can easily compare/reason about (so not aggregates themselves).

MonsieurNicolas:
needs to be rebased @jonjove (and one question pending)

src/ledger/LedgerManagerImpl.cpp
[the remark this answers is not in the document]
jonjove: Okay I'll remove this metric because you can calculate the time before nomination starts from ledger age, ballot time, and nomination time. Leaving the slow ledger logging in the new place because it more accurately captures the event "this ledger was slow because SCP was slow".

MonsieurNicolas:
r+ 3da430c75d68499c1ac8e6df02d6d995f3fb87b0

MonsieurNicolas:
r+ e7b01666cc73b336392534e97a3f361d3f1fee76

PR #2170, opened 2019-06-24
herder: order qmap before hashing, reduce spurious recalculation.

graydon:
I can't say this strictly matters -- no tests reveal presently that it's a problem -- but it takes a bit of potential nondeterminism out of the calculation of the transitive qmap which could potentially trigger more recalculation than necessary of quorum intersection.

### DIFF
--- a/src/herder/HerderImpl.cpp
+++ b/src/herder/HerderImpl.cpp
@@ -1072,7 +1072,8 @@ static Hash
 getQmapHash(QuorumTracker::QuorumMap const& qmap)
 {
     std::unique_ptr<SHA256> hasher = SHA256::create();
-    for (auto const& pair : qmap)
+    std::map<NodeID, SCPQuorumSetPtr> ordered_map(qmap.begin(), qmap.end());
+    for (auto const& pair : ordered_map)
     {
         hasher->add(xdr::xdr_to_opaque(pair.first));
         if (pair.second)


src/herder/HerderImpl.cpp
```diff
@@ -1072,7 +1072,8 @@ static Hash
 getQmapHash(QuorumTracker::QuorumMap const& qmap)
 {
     std::unique_ptr<SHA256> hasher = SHA256::create();
-    for (auto const& pair : qmap)
+    std::map<NodeID, SCPQuorumSetPtr> ordered_map(qmap.begin(), qmap.end());
```
MonsieurNicolas: maybe use a set instead, something like:
```
    std::transform(m.begin(), m.end(), std::inserter(s, s.begin()), [](auto const& pair) { return pair.first; });
```

graydon: I don't understand .. we need the values of the map too. What would we use the set for?

MonsieurNicolas: For some reason, I thought the map was `NodeID -> qsetHash` (as it's what we use in some places), so we could feed the quorum set hashes to the hasher.
Not that, so current code is fine.

MonsieurNicolas:
r+ 5a1299c1f42fc9186d945c09f1c0f87ea8795d81

PR #2172, opened 2019-06-27
bug 2128: resolve any ready futures before queueing publish state.

graydon:
# Description

Resolves #2128 -- there's a long story over there but the gist is we were queueing HASs that included a lot of `FutureBucket`s in `FB_LIVE_INPUTS` state that were actually done merging, and should be resolved to `FB_LIVE_OUTPUT` state before serializing and storing to the publish queue in the DB. This causes CPU spikes (and significantly delays publishing) by redoing merges in the dequeued-and-deserialized `FutureBucket`s before publishing.

The fix is to sink eager FutureBucket resolution into the BucketList itself -- in the `addBatch` routine -- so when anyone makes a HAS copy of it it's likely to have any finished merges (especially long-running ones) resolved.

# Checklist
- [x] Reviewed the [contributing](https://github.com/stellar/stellar-core/blob/master/CONTRIBUTING.md#submitting-changes) document
- [x] Rebased on top of master (no merge commits)
- [x] Ran `clang-format` v5.0.0 (via `make format` or the Visual Studio extension)
- [x] Compiles
- [x] Ran all tests
- [x] If change impacts performance, include supporting evidence per the [performance document](https://github.com/stellar/stellar-core/blob/master/performance-eval/performance-eval.md)

Issue #1482, opened 2018-01-08
One bucket merge could be executed multiple times


vogel:
We do not cache results of bucket merges. It can cause to suboptimal usage of resources. Imagine not-yet merged `FutureBucket` that goes into `publishqueue` table. If it is at high enough level, it will be created multiple times as result of merge of the same two buckets with the same shadow set - it will lead to the same output bucket each time, which obviously could be cached.

We should add that cache to reduce IO and CPU usage.

MonsieurNicolas:
Why do we resolve pending merges in the publish queue in the first place?

vogel:
Let me look it up in commit history.

vogel:
100% lack of information there - only one commit adding whole HistoryWork stuff.

I think that the only reason this is done it to reduce time it requires nodes to catch up. Comment on `BucketList.h` mentions that, for example, for level 7 merge could take up to 45 minutes. For level 8 it would be 180 minutes, for level 9 720 minutes and for level 10 2880 minutes (2 days).

Doing that merge during catchup is suboptimal.

But also doing that merge during publish means that we do not publish for a long time.

I see 2 part solution here:
1. do not wait with publish until resolving pending merges, instead add a method to update history files when merge is done (this is 100% safe, as FutureBucket is not part of hash)
2. when doing catchup and encountering big bucket merge, go back a few checkpoints and try again to get previous level-10 or level-9 bucket, it would probably take less time to apply more transactions than to wait for a big bucket merge

MonsieurNicolas:
right, from what you are saying it seems that it's basically safer for the network to always publish (without waiting for the merge to be done), so that other nodes can catchup to the network.
Right now if we have a slow merge it potentially opens the possibility of having arbitrary long gaps between checkpoints, which doesn't sound like a good property.

During that time the bucket will show up with inputs but that seems fine. The optimization to fixup the pending merges with the completed merge seems like something we could do later on.

MonsieurNicolas:
@marta-lokhova is working on this

PR #2174, opened 2019-06-28
Inhibit double merge on publish

graydon:
# Description

Followup to #2172 to fix "the other problem" @MonsieurNicolas noted (a.k.a. bug #1482): if a merge is already running (or otherwise still has a live output) and we deserialize a FutureBucket that happens to match its input state, we will start the same merge twice when resolving that deserialized FutureBucket.

When does this happen? Unfortunately somewhat often, at least every time we do a publish. When we queue an HAS for publication, we serialize all of its FutureBuckets into the DB and then typically immediately deserialize them, duplicating any merges that were running (such as all those merges started at the ledger-close that triggered the publish).

We should (and this PR does) re-attach the deserialized FutureBucket to the running merge. It also retains a merge inputs-to-output map, to allow _resynthesizing_ a std::shared_future, should the FutureBucket expire before the merge is needed (eg. when a publish queue is so backed up that the BucketList has moved on and committed the FutureBuckets before the HAS is deserialized). This is a little convoluted but I _think_ it will work. Just needs lots of testing (which unfortunately I'm out of time for today).

I'm not sure if this should be considered a full fix for #1482, but maybe? The map here is not persistent in the DB, but I think it'll cover the main case we're concerned about.

(Still needs a test, just figured I'd post a thing to look at for early feedback.)

# Checklist
- [x] Compiles
- [x] Ran all tests

### DIFF
--- a/src/bucket/BucketManager.h
+++ b/src/bucket/BucketManager.h
@@ -7,6 +7,7 @@
 #include "bucket/Bucket.h"
 #include "overlay/StellarXDR.h"
 #include "util/NonCopyable.h"
+#include <future>
 #include <memory>
 
 #include "medida/timer_context.h"
@@ -18,6 +19,7 @@ class Application;
 class BucketList;
 class TmpDirManager;
 struct LedgerHeader;
+struct MergeKey;
 struct HistoryArchiveState;
 
 // A fine-grained merge-operation-counter structure for tracking various
@@ -28,6 +30,9 @@ struct MergeCounters
 {
     uint64_t mPreInitEntryProtocolMerges{0};
     uint64_t mPostInitEntryProtocolMerges{0};
+
+    uint64_t mRunningMergeReattachments{0};
+    uint64_t mFinishedMergeReattachments{0};
 
     uint64_t mNewMetaEntries{0};
     uint64_t mNewInitEntries{0};
@@ -115,10 +120,33 @@ class BucketManager : NonMovableOrCopyable
     // worker threads. Very carefully.
     virtual std::shared_ptr<Bucket>
     adoptFileAsBucket(std::string const& filename, uint256 const& hash,
-                      size_t nObjects, size_t nBytes) = 0;
+                      size_t nObjects, size_t nBytes,
+                      MergeKey* mergeKey = nullptr) = 0;
 
     // Return a bucket by hash if we have it, else return nullptr.
     virtual std::shared_ptr<Bucket> getBucketByHash(uint256 const& hash) = 0;
+
+    // Get a reference to a merge-future that's either running (or finished
+    // somewhat recently) from either a map of the std::shared_futures doing the
+    // merges and/or a set of records mapping merge inputs to outputs and the
+    // set of outputs held in the BucketManager. Returns an invalid future if no
+    // such future can be found or synthesized.
+    virtual std::shared_future<std::shared_ptr<Bucket>>
+    getMergeFuture(MergeKey const& key) = 0;
+
+    // Add a reference to a merge _in progress_ (not yet adopted as a file) to
+    // the BucketManager's internal map of std::shared_futures doing merges.
+    // There is no corresponding entry-removal API: the std::shared_future will
+    // be removed from the map when the merge completes and the output file is
+    // adopted.
+    virtual void
+    putMergeFuture(MergeKey const& key,
+                   std::shared_future<std::shared_ptr<Bucket>>) = 0;
+
+#ifdef BUILD_TESTS
+    // Drop all references to merge futures in progress.
+    virtual void clearMergeFuturesForTesting() = 0;
+#endif
 
     // Forget any buckets not referenced by the current BucketList. This will
     // not immediately cause the buckets to delete themselves, if someone else
--- a/src/bucket/BucketManager.h
+++ b/src/bucket/BucketManager.h
@@ -55,6 +55,7 @@ struct MergeCounters
     uint64_t mOutputIteratorBufferUpdates{0};
     uint64_t mOutputIteratorActualWrites{0};
     MergeCounters& operator+=(MergeCounters const& delta);
+    bool operator==(MergeCounters const& other) const;
 };
 
 /**
--- a/src/bucket/BucketManagerImpl.cpp
+++ b/src/bucket/BucketManagerImpl.cpp
@@ -209,6 +209,9 @@ MergeCounters::operator+=(MergeCounters const& delta)
     mPreInitEntryProtocolMerges += delta.mPreInitEntryProtocolMerges;
     mPostInitEntryProtocolMerges += delta.mPostInitEntryProtocolMerges;
 
+    mRunningMergeReattachments += delta.mRunningMergeReattachments;
+    mFinishedMergeReattachments += delta.mFinishedMergeReattachments;
+
     mNewMetaEntries += delta.mNewMetaEntries;
     mNewInitEntries += delta.mNewInitEntries;
     mNewLiveEntries += delta.mNewLiveEntries;
@@ -244,6 +247,9 @@ MergeCounters::operator==(MergeCounters const& other) const
     return (
         mPreInitEntryProtocolMerges == other.mPreInitEntryProtocolMerges &&
         mPostInitEntryProtocolMerges == other.mPostInitEntryProtocolMerges &&
+
+        mRunningMergeReattachments == other.mRunningMergeReattachments &&
+        mFinishedMergeReattachments == other.mFinishedMergeReattachments &&
 
         mNewMetaEntries == other.mNewMetaEntries &&
         mNewInitEntries == other.mNewInitEntries &&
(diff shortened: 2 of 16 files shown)


src/bucket/BucketManagerImpl.cpp
```diff
@@ -94,7 +94,7 @@ BucketManagerImpl::BucketManagerImpl(Application& app)
     , mBucketSnapMerge(app.getMetrics().NewTimer({"bucket", "snap", "merge"}))
     , mSharedBucketsSize(
           app.getMetrics().NewCounter({"bucket", "memory", "shared"}))
-
+    , mMergeFutureCache(1024)
```
MonsieurNicolas: why 1024? isn't the worst case around 10 merges? Given that you're using `weak_ptr`, we should never go beyond that (and 1024 is a huge number when translated into buckets).

In general, I don't know if a cache is the right data structure: what we really want is to remove merges that are not in use anymore (`weak_ptr` expired).

graydon: 1024 is a little arbitrary; there are enough different paths that intersect here that I figured it wouldn't hurt to give it some wiggle room.

I think the _worst_ case is when we're doing rapid replay like during catchup-with-publish: we kick off 10 merges passing a publish checkpoint, serialize an HAS (forgetting those 10), then continue setting off merges at maximum replay-speed. If the HAS is not deserialized soon enough (and it might not -- there might be a backlog of publishing) then the long running merges (say L8->L9) it was supposed to re-attach to will be purged from the cache.

Another way to look at it is that the cache should really only bother with longer-lived merges. So we could theoretically make it smaller and exclude merges younger than L4 or something? But I figured this is harmless as-is.

I don't know why a cache would be a bad data structure. A `weak_ptr` is a small thing to keep around when its corresponding `shared_ptr` is dead: just a control block.

(Oh! I just looked and [according to this](https://www.bfilipek.com/2017/12/weakptr-memory.html) it seems like `shared_ptr` will combine the control block with the pointee allocation when you use `make_shared` rather than constructing from a `new` ptr, so this might retain an entire destructed-but-allocated `Bucket` structure for the `weak_ptr`. That's still not .. especially big though. A destructed `std::string` is 32 bytes, a `Hash` is 20 bytes, and a `size_t` is 8 bytes. So a `Bucket` in that state will be 60 bytes + the control block which I think is 16 bytes? 1024 of these is ~77k.)

graydon: Note to self: followed-up in chat and the concern here isn't the size, it's that a cache might lose a live merge for no reason (regardless of size) whereas there are deterministic lifecycle edges for a FutureBucket at which a running merge can be registered and de-registered precisely.

src/bucket/BucketManagerImpl.cpp
```diff
@@ -274,6 +314,21 @@ BucketManagerImpl::adoptFileAsBucket(std::string const& filename,
         {
             mSharedBuckets.insert(std::make_pair(hash, b));
             mSharedBucketsSize.set_count(mSharedBuckets.size());
+            if (mergeKey)
+            {
+                // If this was a merge, retain records of the input->output
+                // mapping, both precisely and in full (so we can fetch it by
+                // input if someone wants it) and approximately by index (so we
+                // overapproximate what to retain when GC'ing).
+                mSharedBucketMergesByMergeKey.insert(
```
MonsieurNicolas: Use `emplace` here:
`mSharedBucketMergesByMergeKey.emplace(*mergeKey, hash);`

src/bucket/BucketManagerImpl.cpp
```diff
@@ -310,6 +365,41 @@ BucketManagerImpl::getBucketByHash(uint256 const& hash)
...
+        auto j = mSharedBucketMergesByMergeKey.find(key);
+        if (j != mSharedBucketMergesByMergeKey.end())
+        {
+            auto bucket = getBucketByHash(j->second);
+            if (bucket)
+            {
+                std::promise<std::shared_ptr<Bucket>> promise;
+                auto future = promise.get_future().share();
+                promise.set_value(bucket);
+                mLiveFutures.insert(std::make_pair(key, future));
```
MonsieurNicolas: nit: `emplace`

src/bucket/BucketManagerImpl.cpp
```diff
@@ -310,6 +365,41 @@ BucketManagerImpl::getBucketByHash(uint256 const& hash)
...
+    }
+    return i->second;
+}
+
+void
+BucketManagerImpl::putMergeFuture(
+    MergeKey const& key, std::shared_future<std::shared_ptr<Bucket>> wp)
+{
+    std::lock_guard<std::recursive_mutex> lock(mBucketMutex);
+    mLiveFutures.emplace(std::make_pair(key, wp));
```
MonsieurNicolas: emplace doesn't need the extra `std::make_pair`

src/bucket/BucketManagerImpl.cpp
```diff
@@ -435,6 +615,77 @@ BucketManagerImpl::forgetUnreferencedBuckets()
...
+
+                // It's possible for the same output to occur for multiple
+                // merge keys (eg. a+b and b+a). And the set of per-input
+                // entries should always be larger than the per-key entries.
+                assert(mSharedBucketMergesByOutput.size() <=
+                       mSharedBucketMergesByMergeKey.size());
+                assert(mSharedBucketMergesByMergeKey.size() <=
+                       mSharedBucketMergesByInput.size());
+
+                CLOG(TRACE, "Bucket")
```
MonsieurNicolas: move the `CLOG` to before the asserts in case those assert end up crashing the process

src/bucket/BucketManagerImpl.cpp
```diff
@@ -435,6 +615,77 @@ BucketManagerImpl::forgetUnreferencedBuckets()
...
+                    mergeProducingOutput->second;
+                assert(output == outputBeingDropped);
+
+                // There should be no futures alive with this output: we
+                // switched to storing only weak input/output mappings when any
+                // merge producing the bucket completed (in adoptFileAsBucket),
+                // and we believe there's only one reference to the bucket
+                // anyways -- our own in mSharedBuckets. But there might be a
+                // race we missed, so double check here.
+                assert(mLiveFutures.find(mergeKeyProducingOutput) ==
```
MonsieurNicolas: I normally prefer to not have asserts in critical code like this, maybe we should instead just leak memory and buckets? This would make it highly unlikely that all nodes crash at the same time if there is indeed some bug.

On startup, we would clean things up anyways.

src/bucket/BucketManager.h
```diff
@@ -114,11 +142,32 @@ class BucketManager : NonMovableOrCopyable
...
+    // the BucketManager's internal map of std::shared_futures doing merges.
+    // There is no corresponding entry-removal API: the std::shared_future will
+    // be removed from the map when the merge completes and the output file is
+    // adopted.
+    virtual void
+    putMergeFuture(MergeKey const& key,
+                   std::shared_future<std::shared_ptr<Bucket>>) = 0;
+
+    // Drop all references to merge futures in progress.
+    virtual void clearMergeFutures() = 0;
```
MonsieurNicolas: should that be called **ForTesting? It seems a bit dangerous

src/bucket/BucketManagerImpl.cpp
graydon: fixed

src/bucket/BucketManagerImpl.cpp
graydon: fixed

src/bucket/BucketManagerImpl.cpp
```diff
@@ -227,17 +255,69 @@ MergeCounters::operator+=(MergeCounters const& delta)
     return *this;
 }
 
+bool
+MergeCounters::operator==(MergeCounters const& other) const
```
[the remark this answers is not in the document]
graydon: Alas, not yet today!

src/bucket/BucketManager.h
graydon: Yeah fair. I'll debug-guard it too.

src/bucket/BucketManagerImpl.cpp
graydon: Fair. Kinda scared me too.

src/bucket/BucketManagerImpl.cpp
graydon: Ah good point.

src/bucket/BucketManagerImpl.cpp
```diff
@@ -272,11 +352,37 @@ BucketManagerImpl::adoptFileAsBucket(std::string const& filename,
 
         b = std::make_shared<Bucket>(canonicalName, hash);
         {
-            mSharedBuckets.insert(std::make_pair(hash, b));
+            mSharedBuckets.emplace(hash, b);
             mSharedBucketsSize.set_count(mSharedBuckets.size());
         }
     }
     assert(b);
+    if (mergeKey)
```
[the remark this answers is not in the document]
graydon: maybe. out of time today!

src/bucket/BucketManagerImpl.cpp
graydon: fixed

graydon:
Posted update with logic tidied up and more-tested, utility classes split out.

Still doesn't actually record the final merge key into an output iterator when building it, and lacks a few more tests at the integration level, but getting better.

src/bucket/BucketManagerImpl.cpp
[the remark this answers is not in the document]
graydon: Not clear but avoid adding. Done.

MonsieurNicolas:
r+ 174500dd66d22b8edffece0c1d257d1c72f65b0d

graydon:
r+ f11180f

Issue #2163, opened 2019-06-20
Refactor TransactionFrame to be immutable


jonjove:
`TransactionFrame` and `TxSetFrame` have interfaces that do not satisfy the criteria of "easy to use correctly and hard to use incorrectly". Specifically, these classes would be much more intuitive if they were immutable.

MonsieurNicolas:
Work on `TxSetFrame` was done.

Work on `TransactionFrame???` is left.
There might be more to it than just `TransactionResult` -> easiest might be to move everything mutable into a `TxResultContext` object (that contains `TransactionResult` and also things like `mCachedAccount`).

PR #2179, opened 2019-07-02
Simulate transaction application

jonjove:
This PR adds a `simulate` mode to stellar-core which permits replay of transactions from a history archive at an increased number of operations per ledger. This will be useful for testing the performance characteristics of stellar-core.

# Checklist
- [x] Compiles
- [x] Ran all tests

### DIFF
--- a/src/catchup/simulation/HistoryArchiveStream.cpp
+++ b/src/catchup/simulation/HistoryArchiveStream.cpp
@@ -1,0 +1,209 @@
+// Copyright 2019 Stellar Development Foundation and contributors. Licensed
+// under the Apache License, Version 2.0. See the COPYING file at the root
+// of this distribution or at http://www.apache.org/licenses/LICENSE-2.0
+
+#include "catchup/simulation/HistoryArchiveStream.h"
+#include "herder/TxSetFrame.h"
+#include "history/FileTransferInfo.h"
+#include "history/HistoryManager.h"
+
+namespace stellar
+{
+
+HistoryArchiveStream::HistoryArchiveStream(TmpDir const& downloadDir,
+                                           LedgerRange const& range,
+                                           HistoryManager const& hm)
+    : mDownloadDir(downloadDir), mRange(range), mHistoryManager(hm)
+{
+    if (mRange.mFirst != 1 &&
+        mHistoryManager.nextCheckpointLedger(mRange.mFirst) != mRange.mFirst)
+    {
+        uint32_t prev = mHistoryManager.prevCheckpointLedger(mRange.mFirst);
+        readHeaderHistory(std::max<uint32_t>(1, prev));
+        readTransactionHistory();
+        readResultHistory();
+        while (mHeaderHistory.header.ledgerSeq != mRange.mFirst - 1)
+        {
+            readHeaderHistory();
+            readTransactionHistory();
+            readResultHistory();
+        }
+    }
+}
+
+LedgerHeaderHistoryEntry
+HistoryArchiveStream::readHeaderHistory(uint32_t ledgerSeq)
+{
+    LedgerHeaderHistoryEntry lhhe;
+
+    if (ledgerSeq == 1 ||
+        mHistoryManager.nextCheckpointLedger(ledgerSeq) == ledgerSeq)
+    {
+        if (mHeaderStream && mHeaderStream.readOne(lhhe))
+        {
+            throw std::runtime_error(
+                "Header stream should be exhausted at end of checkpoint");
+        }
+
+        FileTransferInfo info(
+            mDownloadDir, HISTORY_FILE_TYPE_LEDGER,
+            mHistoryManager.checkpointContainingLedger(ledgerSeq));
+        mHeaderStream.close();
+        mHeaderStream.open(info.localPath_nogz());
+    }
+
+    if (!(mHeaderStream && mHeaderStream.readOne(lhhe)))
+    {
+        throw std::runtime_error(
+            "Header stream should not be exhausted before end of checkpoint");
+    }
+    assert(lhhe.header.ledgerSeq == ledgerSeq);
+
+    mHeaderHistory = lhhe;
+    return lhhe;
+}
+
+LedgerHeaderHistoryEntry
+HistoryArchiveStream::readHeaderHistory()
+{
+    return readHeaderHistory(mHeaderHistory.header.ledgerSeq + 1);
+}
+
+TransactionHistoryEntry
+HistoryArchiveStream::makeEmptyTransactionHistory()
+{
+    TransactionHistoryEntry txhe;
+    txhe.ledgerSeq = mHeaderHistory.header.ledgerSeq + 1;
+    txhe.txSet.previousLedgerHash = mHeaderHistory.hash;
+    return txhe;
+}
+
+TransactionHistoryEntry
+HistoryArchiveStream::readTransactionHistory()
+{
+    TransactionHistoryEntry txhe;
+
+    uint32_t ledgerSeq = mHeaderHistory.header.ledgerSeq;
+    if (ledgerSeq == 1 ||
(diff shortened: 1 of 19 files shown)


src/herder/simulation/SimulationTxSetFrame.cpp
```diff
@@ -0,0 +1,121 @@
...
+#include <numeric>
+
+namespace stellar
+{
+
+static Hash
+computeContentsHash(Hash const& previousLedgerHash,
+                    std::vector<TransactionEnvelope> transactions)
+{
+    std::sort(
```
MonsieurNicolas: it's a bit unfortunate that we're duplicating a lot of code from `TxSetFrame`, can we instead just have a `TxSetFrame` member that we use to delegate all those things?

src/transactions/simulation/SimulationTransactionFrame.cpp
```diff
@@ -0,0 +1,103 @@
...
+bool
+SimulationTransactionFrame::isTooLate(LedgerTxnHeader const& header) const
+{
+    return mSimulationResult.result.code() == txTOO_LATE;
+}
+
+bool
+SimulationTransactionFrame::isBadSeq(int64_t seqNum) const
+{
+    return mSimulationResult.result.code() == txBAD_SEQ;
```
MonsieurNicolas: during validation time, you'd never return false here ; only during apply time

src/main/CommandLine.cpp
```diff
@@ -797,6 +801,53 @@ runRebuildLedgerFromBuckets(CommandLineArgs const& args)
...
+    uint32_t desiredOps = 0;
+    uint32_t startLedger = 0;
+    uint32_t stopLedger = 0;
+    std::string historyArchiveGet;
+    return runWithHelp(
+        args,
+        {configurationParser(configOption),
+         clara::Opt{desiredOps, "DESIRED-OPERATIONS"}["--desired-operations"](
+             "desired number of operations per ledger, will never be less but "
+             "could be up to to 100 more"),
```
MonsieurNicolas: I think a cat messed up part of your sentence

src/main/CommandLine.cpp
```diff
@@ -797,6 +801,53 @@ runRebuildLedgerFromBuckets(CommandLineArgs const& args)
...
+{
+    CommandLine::ConfigOption configOption;
+    uint32_t desiredOps = 0;
+    uint32_t startLedger = 0;
+    uint32_t stopLedger = 0;
+    std::string historyArchiveGet;
+    return runWithHelp(
+        args,
+        {configurationParser(configOption),
+         clara::Opt{desiredOps, "DESIRED-OPERATIONS"}["--desired-operations"](
```
MonsieurNicolas: why not just calling this something a little easier to understand like `--ops-per-ledger` (and `OPS-PER-LEDGER`)

src/main/CommandLine.cpp
```diff
@@ -797,6 +801,53 @@ runRebuildLedgerFromBuckets(CommandLineArgs const& args)
...
+             "desired number of operations per ledger, will never be less but "
+             "could be up to to 100 more"),
+         clara::Opt{startLedger, "START-LEDGER"}["--start-ledger"](
+             "ledger to start reading from history archive"),
```
MonsieurNicolas: start/stop doesn't really convey a range? maybe from/to instead? Also, those are inclusive bounds, right?
Alternatively, we could use the same syntax than catchup that is "start/count"

src/main/CommandLine.cpp
```diff
@@ -797,6 +801,53 @@ runRebuildLedgerFromBuckets(CommandLineArgs const& args)
...
+             "ledger to stop reading from history archive")},
+        [&] {
+            auto config = configOption.getConfig();
+            config.setNoListen();
+
+            VirtualClock clock(VirtualClock::REAL_TIME);
+            auto app = Application::create(clock, config, false);
+            app->start();
+
+            LedgerRange lr{startLedger, stopLedger};
```
MonsieurNicolas: maybe perform some sanity check on the ledger range? This will crash if start = 0 or start > stop

jonjove: Done.

src/transactions/simulation/SimulationTransactionFrame.cpp
jonjove: `SimulationTransactionFrame` isn't used for validation, `applying` is always true on this code path.

src/catchup/simulation/HistoryArchiveStream.cpp
```diff
@@ -0,0 +1,209 @@
...
+    {
+        if (mHeaderStream && mHeaderStream.readOne(lhhe))
+        {
+            throw std::runtime_error(
+                "Header stream should be exhausted at end of checkpoint");
+        }
+
+        FileTransferInfo info(
+            mDownloadDir, HISTORY_FILE_TYPE_LEDGER,
+            mHistoryManager.nextCheckpointLedger(ledgerSeq + 1) - 1);
```
vogel: you can use checkpointContainingLedger() method for that

src/transactions/simulation/SimulationMergeOpFrame.cpp
```diff
@@ -0,0 +1,28 @@
...
+}
+
+bool
+SimulationMergeOpFrame::isSeqnumTooFar(LedgerTxnHeader const& header,
+                                       AccountEntry const& sourceAccount)
+{
+    if (mSimulationResult.code() == opINNER)
+    {
+        auto code = mSimulationResult.tr().accountMergeResult().code();
+        return code == ACCOUNT_MERGE_SEQNUM_TOO_FAR;
```
vogel: that is clever ;)

src/herder/simulation/SimulationTxSetFrame.cpp
MonsieurNicolas: yeah 3 seems fine for now, the performance impact of it should go away when #2163 is done

src/transactions/simulation/SimulationTransactionFrame.cpp
MonsieurNicolas: yes you're right: I thought we would use this simulation code in a "live" environment, but if we do this, we'll need more changes anyways so keeping this PR for "command line replay" is fine for now.

src/main/CommandLine.cpp
MonsieurNicolas: ok

src/main/CommandLine.cpp
MonsieurNicolas: I think the new code is not exactly right: you're only calling the validation code when parsing `firstLedgerInclusive`. I think it would be better instead have a `lastLedgerParser` that takes the same validation lambda (I suspect that's what you wanted to do).

src/catchup/simulation/HistoryArchiveStream.cpp
jonjove: Done.

src/main/CommandLine.cpp
jonjove: Validation first parses the entire command, then runs the validator for each parser:
https://github.com/stellar/stellar-core/blob/e93ea152aba73c2ed6d6fb27d15dd9ff6884dedf/src/main/CommandLine.cpp#L221-L237
So there isn't any advantage to having it in both parsers

src/herder/simulation/SimulationTxSetFrame.cpp
jonjove: Done.

MonsieurNicolas:
r+ fc1e49dbe57c8df0dc94a3918e280cac8e9e19c1

MonsieurNicolas:
r+ 92b85314e66c8350794026ee22bebb8d36724cc8

PR #2182, opened 2019-07-03
Introduce Transaction Subsystem Fuzzing and Common Fuzzer Interface

robertDurst:
# Some Background

Fuzzing is a test strategy that performs automated binary executions with randomly generated/mutated inputs, monitoring for program crashes. It is useful for increasing test coverage by discovery of obscure, unthought-of, and interesting edge cases.

The basic process works as so:
```
**Input**: an initial corpus of well formed data
1. Inputs fed into a program
2. Fuzzer derives more inputs (based on some gathered metrics like unique execution path)
3. Repeat
**Output**: inputs that caused crashes
```
There are many fuzz testing tools out there such as [libFuzzer](https://llvm.org/docs/LibFuzzer.html), [AFL](http://lcamtuf.coredump.cx/afl/), and [Peach Fuzzer](https://www.peach.tech/products/peach-fuzzer/). For the time being, I have implemented a test harness to work with AFL.

# Description

This builds on #2155, introducing a transaction subsystem fuzzer.

A significant portion of the work here was integrating the introduced transaction fuzzer with [AFL's persistent mode](https://lcamtuf.blogspot.com/2015/06/new-in-afl-persistent-mode.html). Persistent mode allows for multiple executions of the binary in a single long-lived process, significantly improving executions per second since the expensive state setup can be limited to every 1,000th, or even 1,000,000th execution. Areas where determinism improved include:
* foregoing signature verification (we aren't breaking ED25519 here)
* clearing caches
* seeding randomness

This PR also defines a common interface for fuzzing various subsystems, allowing one to utilize a CLI flag to switch between `Overlay` and `Transaction` fuzzing.

Finally, this PR includes an option to specify `process_id`, allowing for easy parallelization of fuzzing (as supported by AFL). 

Further progress of #1376 

# Future Work

### Generating transactions with more operations
To demonstrate functionality, I restricted the fuzzer to a subset of operations: `ACCOUNT_MERGE` and `PAYMENT` of native assets. Running against protocol 4, the fuzzer discovered the [double merge bug described here](https://www.stellar.org/developers/stellar-core/software/security-protocol-release-notes.html#v061b-not-widely-released-2017-04-06). This set will be expanded as I extend the fuzz test harness methodically, focusing first on discovery of the rest of the [historical bugs](https://www.stellar.org/developers/stellar-core/software/security-protocol-release-notes.html#v061b-not-widely-released-2017-04-06) in chronological order.

### Improving Fuzzer Efficiency
I also spent some time thinking about how to meaningfully and unobstructively limit the *fuzz space*. For example, I use a small set of pre-generated accounts for operations. Since I do this, and cause a disconnect between what the fuzzer fuzzes and what is actually fed to Stellar-core, cpu time spent incrementing and flipping bits in the `AccountID` property of XDR's is wasted time. Thus, I feed the fuzzer an initial corpus of operations with all bytes of the `AccountID`  zero'd out except for the 32nd byte to nudge it in the right direction.

***

### Note for Reviewers

Dependent on and rebased against #2155.

# Checklist
- [x] Compiles
- [x] Ran all tests

### DIFF
--- a/src/ledger/LedgerTxn.cpp
+++ b/src/ledger/LedgerTxn.cpp
@@ -1288,6 +1288,21 @@ LedgerTxnRoot::Impl::~Impl()
     }
 }
 
+#ifdef FUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION
+void
+LedgerTxnRoot::Impl::resetForFuzzer()
+{
+    mBestOffersCache.clear();
+    mEntryCache.clear();
+}
+
+void
+LedgerTxnRoot::resetForFuzzer()
+{
+    mImpl->resetForFuzzer();
+}
+#endif // FUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION
+
 void
 LedgerTxnRoot::addChild(AbstractLedgerTxn& child)
 {
--- a/src/main/CommandLine.cpp
+++ b/src/main/CommandLine.cpp
@@ -12,6 +12,7 @@
 #include "main/StellarCoreVersion.h"
 #include "main/dumpxdr.h"
 #include "scp/QuorumSetUtils.h"
+#include "test/Fuzzer.h"
 #include "util/Logging.h"
 #include "util/optional.h"
 
@@ -161,6 +162,38 @@ ParserWithValidation
 fileNameParser(std::string& string)
 {
     return requiredArgParser(string, "FILE-NAME");
+}
+
+ParserWithValidation
+fuzzerModeParser(std::string& fuzzerModeArg, FuzzerMode& fuzzerMode)
+{
+    auto validateFuzzerMode = [&] {
+        if (iequals(fuzzerModeArg, "overlay"))
+        {
+            fuzzerMode = FuzzerMode::OVERLAY;
+            return "";
+        }
+
+        if (iequals(fuzzerModeArg, "tx"))
+        {
+            fuzzerMode = FuzzerMode::TRANSACTION;
+            return "";
+        }
+
+        return "Unrecognized fuzz mode. Please select a valid mode.";
+    };
+
+    return {clara::Opt{fuzzerModeArg, "FUZZER-MODE"}["--mode"](
+                "set the fuzzer mode. Expected modes: overlay, "
+                "tx. Defaults to overlay."),
+            validateFuzzerMode};
+}
+
+clara::Opt
+processIDParser(uint& num)
+{
+    return clara::Opt{num, "PROCESS-ID"}["--process_id"](
+        "for spawning multiple instances in fuzzing parallelization");
 }
 
 clara::Opt
@@ -866,12 +899,17 @@ runFuzz(CommandLineArgs const& args)
     el::Level logLevel{el::Level::Info};
     std::vector<std::string> metrics;
     std::string fileName;
+    uint processID = 0;
+    FuzzerMode fuzzerMode{FuzzerMode::OVERLAY};
+    std::string fuzzerModeArg = "overlay";
 
     return runWithHelp(args,
                        {logLevelParser(logLevel), metricsParser(metrics),
-                        fileNameParser(fileName)},
+                        fileNameParser(fileName), processIDParser(processID),
+                        fuzzerModeParser(fuzzerModeArg, fuzzerMode)},
                        [&] {
-                           fuzz(fileName, logLevel, metrics);
+                           fuzz(fileName, logLevel, metrics, processID,
+                                fuzzerMode);
                            return 0;
                        });
 }
@@ -880,11 +918,17 @@ int
 runGenFuzz(CommandLineArgs const& args)
 {
     std::string fileName;
+    FuzzerMode fuzzerMode{FuzzerMode::OVERLAY};
+    std::string fuzzerModeArg = "overlay";
+    uint processID = 0;
 
-    return runWithHelp(args, {fileNameParser(fileName)}, [&] {
-        genfuzz(fileName);
-        return 0;
-    });
+    return runWithHelp(
+        args,
+        {fileNameParser(fileName), fuzzerModeParser(fuzzerModeArg, fuzzerMode)},
+        [&] {
+            FuzzUtils::createFuzzer(processID, fuzzerMode)->genFuzz(fileName);
+            return 0;
+        });
 }
 #endif
 
@@ -955,7 +999,7 @@ handleCommandLine(int argc, char* const* argv)
     auto commandName = fmt::format("{0} {1}", exeName, command->name());
     auto args = CommandLineArgs{exeName, commandName, command->description(),
                                 adjustedCommandLine.second};
-    if (command->name() == "run")
+    if (command->name() == "run" || command->name() == "fuzz")
     {
         // run outside of catch block so that we properly capture crashes
         return command->run(args);
--- a/src/test/FuzzerImpl.cpp
+++ b/src/test/FuzzerImpl.cpp
@@ -114,6 +114,14 @@ createFuzzTransactionFrame(PublicKey sourceAccountID,
 }
 
 bool
+isBadTransactionFuzzerInput(Operation const& op)
+{
+    return !(op.body.type() == ACCOUNT_MERGE) &&
+           !(op.body.type() == PAYMENT &&
+             op.body.paymentOp().asset.type() == ASSET_TYPE_NATIVE);
+}
+
+bool
 isBadOverlayFuzzerInput(StellarMessage const& m)
 {
     // HELLO, AUTH and ERROR_MSG messages cause the connection between
@@ -125,6 +133,125 @@ isBadOverlayFuzzerInput(StellarMessage const& m)
     // execution paths -- the fuzzer's main metric for determinism
     // (stability) and binary coverage.
     return m.type() == AUTH || m.type() == ERROR_MSG || m.type() == HELLO;
+}
+
+void
+TransactionFuzzer::initialize()
+{
+    VirtualClock clock;
+    mApp = createTestApplication(clock, getFuzzConfig(mProcessID));
+
+    resetTxInternalState(*mApp);
+    LedgerTxn ltx(mApp->getLedgerTxnRoot());
+
+    // setup the state, for this we only need to pregenerate some accounts. For
+    // now we create mNumAccounts accounts, or enough to fill the first few bits
+    // such that we have a pregenerated account for the last few bits of the
+    // 32nd byte of a public key, thus account creation is over a deterministic
+    // range of public keys
+    for (int i = 0; i < mNumAccounts; ++i)
+    {
+        PublicKey publicKey;
+        uint256 accountID;
+        accountID.at(31) = i;
+        publicKey.ed25519() = accountID;
+
+        // manually construct ledger entries, "creating" each account
+        LedgerEntry newAccountEntry;
+        newAccountEntry.data.type(ACCOUNT);
+        auto& newAccount = newAccountEntry.data.account();
+        newAccount.thresholds[0] = 1;
+        newAccount.accountID = publicKey;
+        newAccount.seqNum = 0;
+        // to create "interesting" balances we utilize powers of 2
+        newAccount.balance = 2 << i;
+        ltx.create(newAccountEntry);
+
+        // select the first pregenerated account to be the hard coded source
+        // account for all transactions
+        if (i == 0)
+            mSourceAccountID = publicKey;
+    }
+
+    // commit these to the ledger so that we have a starting, persistent state
+    // to fuzz test against -- should be the only stateful/effectful action
+    ltx.commit();
+}
+
+void
+TransactionFuzzer::inject(XDRInputFileStream& in)
+{
+    // for tryRead, in case of fuzzer creating an ill-formed xdr, generate an
+    // xdr that will trigger a non-execution path so that the fuzzer realizes it
+    // has hit an uninteresting case
+    auto tryRead = [&in](xdr::xvector<Operation>& m) {
+        try
+        {
+            return in.readOne(m);
+        }
+        catch (...)
+        {
+            m.clear(); // we ignore transactions with 0 operations
+            return true;
+        }
+    };
+
+    xdr::xvector<Operation> ops;
+    while (tryRead(ops))
+    {
+        auto wellSizedTx = ops.size() > 0 && ops.size() < 6;
+        if (std::any_of(ops.begin(), ops.end(), isBadTransactionFuzzerInput) ||
+            !wellSizedTx)
+        {
+            return;
+        }
+
+        // construct transaction
+        auto txFramePtr = createFuzzTransactionFrame(mSourceAccountID, ops,
+                                                     mApp->getNetworkID());
+
+        {
+            resetTxInternalState(*mApp);
+            LedgerTxn ltx(mApp->getLedgerTxnRoot());
+
+            // attempt to apply transaction
+            txFramePtr->attemptApplication(*mApp, ltx);
+        }
+        resetTxInternalState(*mApp);
+    }
+}
+
+int
+TransactionFuzzer::xdrSizeLimit()
+{
+    // path_payments are the largest operations, where 100k 1-3 operation
+    // transactions (more correctly xdr::xvector's) with path_payments averaged
+    // to 294 bytes, so 0x500 should be fine
+    return 0x500;
+}
+
+#define FUZZER_INITIAL_CORPUS_MAX_OPERATIONS 5
+#define FUZZER_INITIAL_CORPUS_OPERATION_GEN_UPPERBOUND 128
+void
+TransactionFuzzer::genFuzz(std::string const& filename)
+{
+    XDROutputFileStream out(/*doFsync=*/false);
+    out.open(filename);
+    autocheck::generator<Operation> gen;
+    xdr::xvector<Operation> ops;
+    auto numops =
+        autocheck::generator<uint>()(FUZZER_INITIAL_CORPUS_MAX_OPERATIONS);
+    for (int i = 0; i < numops; ++i)
+    {
+        Operation op = gen(FUZZER_INITIAL_CORPUS_OPERATION_GEN_UPPERBOUND);
+        while (isBadTransactionFuzzerInput(op))
+        {
+            op = gen(FUZZER_INITIAL_CORPUS_OPERATION_GEN_UPPERBOUND);
+        }
+        ops.emplace_back(op);
+    }
+
+    out.writeOne(ops);
 }
 
 void
--- a/src/test/FuzzerImpl.cpp
+++ b/src/test/FuzzerImpl.cpp
@@ -1,0 +1,164 @@
+// Copyright 2019 Stellar Development Foundation and contributors. Licensed
+// under the Apache License, Version 2.0. See the COPYING file at the root
+// of this distribution or at http://www.apache.org/licenses/LICENSE-2.0
+
+#include "test/FuzzerImpl.h"
+#include "ledger/LedgerTxn.h"
+#include "main/Application.h"
+#include "main/Config.h"
+#include "overlay/OverlayManager.h"
+#include "overlay/TCPPeer.h"
+#include "simulation/Simulation.h"
+#include "test/TestUtils.h"
+#include "test/fuzz.h"
+#include "test/test.h"
+#include "transactions/OperationFrame.h"
+#include "transactions/SignatureChecker.h"
+#include "util/Math.h"
+
+#include <xdrpp/autocheck.h>
+
+namespace stellar
+{
+
+// creates a generic configuration with settings rigged to maximize
+// determinism
+static Config
+getFuzzConfig(int instanceNumber)
+{
+    Config cfg = getTestConfig(instanceNumber);
+    cfg.MANUAL_CLOSE = true;
+    cfg.CATCHUP_COMPLETE = false;
+    cfg.CATCHUP_RECENT = 0;
+    cfg.ARTIFICIALLY_GENERATE_LOAD_FOR_TESTING = false;
+    cfg.ARTIFICIALLY_SET_CLOSE_TIME_FOR_TESTING = UINT32_MAX;
+    cfg.PUBLIC_HTTP_PORT = false;
+    cfg.WORKER_THREADS = 1;
+    cfg.QUORUM_INTERSECTION_CHECKER = false;
+    cfg.PREFERRED_PEERS_ONLY = false;
+    cfg.RUN_STANDALONE = true;
+
+    return cfg;
+}
+
+bool
+isBadOverlayFuzzerInput(StellarMessage const& m)
+{
+    // HELLO, AUTH and ERROR_MSG messages cause the connection between
+    // the peers to drop. Since peer connections are only established
+    // preceding the persistent loop, a dropped peer is not only
+    // inconvenient, it also confuses the fuzzer. Consider a msg A sent
+    // before a peer is dropped and after a peer is dropped. The two,
+    // even though the same message, will take drastically different
+    // execution paths -- the fuzzer's main metric for determinism
+    // (stability) and binary coverage.
+    return m.type() == AUTH || m.type() == ERROR_MSG || m.type() == HELLO;
+}
+
+void
+OverlayFuzzer::initialize()
+{
+    auto networkID = sha256(getTestConfig().NETWORK_PASSPHRASE);
+    mSimulation = std::make_shared<Simulation>(Simulation::OVER_LOOPBACK,
+                                               networkID, getFuzzConfig);
+
+    SIMULATION_CREATE_NODE(10);
+    SIMULATION_CREATE_NODE(11);
+
+    SCPQuorumSet qSet0;
+    qSet0.threshold = 2;
+    qSet0.validators.push_back(v10NodeID);
+    qSet0.validators.push_back(v11NodeID);
+
+    mSimulation->addNode(v10SecretKey, qSet0);
+    mSimulation->addNode(v11SecretKey, qSet0);
+
+    mSimulation->addPendingConnection(v10SecretKey.getPublicKey(),
+                                      v11SecretKey.getPublicKey());
+
+    mSimulation->startAllNodes();
+
+    // crank until nodes are connected
+    mSimulation->crankUntil(
+        [&]() {
+            auto nodes = mSimulation->getNodes();
+            auto numberOfSimulationConnections =
+                nodes[ACCEPTOR_INDEX]
+                    ->getOverlayManager()
+                    .getAuthenticatedPeersCount() +
+                nodes[INITIATOR_INDEX]
+                    ->getOverlayManager()
+                    .getAuthenticatedPeersCount();
+            return numberOfSimulationConnections == 2;
+        },
+        std::chrono::milliseconds{500}, false);
+}
+
+void
+OverlayFuzzer::inject(XDRInputFileStream& in)
+{
+    // see note on TransactionFuzzer's tryRead above
+    auto tryRead = [&in](StellarMessage& m) {
+        try
+        {
+            return in.readOne(m);
+        }
+        catch (...)
+        {
+            m.type(HELLO); // we ignore HELLO messages
+            return true;
+        }
+    };
+
+    StellarMessage msg;
+    while (tryRead(msg))
+    {
+        if (isBadOverlayFuzzerInput(msg))
+        {
+            return;
+        }
+
+        auto nodeids = mSimulation->getNodeIDs();
+        auto loopbackPeerConnection = mSimulation->getLoopbackConnection(
+            nodeids[INITIATOR_INDEX], nodeids[ACCEPTOR_INDEX]);
+
+        auto initiator = loopbackPeerConnection->getInitiator();
+        auto acceptor = loopbackPeerConnection->getAcceptor();
+
+        initiator->getApp().getClock().postToCurrentCrank(
+            [initiator, msg]() { initiator->Peer::sendMessage(msg); });
+
+        mSimulation->crankForAtMost(std::chrono::milliseconds{500}, false);
+
+        // clear all queues and cancel all events
+        initiator->clearInAndOutQueues();
+        acceptor->clearInAndOutQueues();
+
+        while (initiator->getApp().getClock().cancelAllEvents())
+            ;
+        while (acceptor->getApp().getClock().cancelAllEvents())
+            ;
+    }
+}
+
+int
+OverlayFuzzer::xdrSizeLimit()
+{
+    return MAX_MESSAGE_SIZE;
+}
+
+#define FUZZER_INITIAL_CORPUS_MESSAGE_GEN_UPPERBOUND 16
+void
+OverlayFuzzer::genFuzz(std::string const& filename)
+{
+    XDROutputFileStream out(/*doFsync=*/false);
+    out.open(filename);
+    autocheck::generator<StellarMessage> gen;
+    StellarMessage m(gen(FUZZER_INITIAL_CORPUS_MESSAGE_GEN_UPPERBOUND));
+    while (isBadOverlayFuzzerInput(m))
+    {
+        m = gen(FUZZER_INITIAL_CORPUS_MESSAGE_GEN_UPPERBOUND);
+    }
+    out.writeOne(m);
+}
+}
--- a/src/test/FuzzerImpl.cpp
+++ b/src/test/FuzzerImpl.cpp
@@ -361,3 +361,26 @@ OverlayFuzzer::genFuzz(std::string const& filename)
     out.writeOne(m);
 }
 }
+
+#ifdef FUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION
+namespace xdr
+{
+template <>
+void
+generator_t::operator()(stellar::PublicKey& t) const
+{
+    // generate public keys such that it is zero'd out except the last byte,
+    // hence for ED25519 public keys, an uint256 defined as opaque[32] set the
+    // 31st indexed byte. Furthermore, utilize only the first few bits of the
+    // 32nd byte, allowing us to generate these accounts against a pregeneration
+    // of such accountID's.Also, note that NUMBER_OF_PREGENERATED_ACCOUNTS here
+    // creates an off-by-one error, when generating acounts we use
+    // [0,NUMBER_OF_PREGENERATED_ACCOUNTS) exclusive versus here we use
+    // [0,NUMBER_OF_PREGENERATED_ACCOUNTS] inclusive. This allows us to generate
+    // some transactions with a source account/destination account that does
+    // not exist.
+    t.ed25519().at(31) = autocheck::generator<uint>()(
+        stellar::FuzzUtils::NUMBER_OF_PREGENERATED_ACCOUNTS);
+}
+} // namespace xdr
+#endif // FUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION
--- a/src/test/FuzzerImpl.cpp
+++ b/src/test/FuzzerImpl.cpp
@@ -41,6 +41,78 @@ getFuzzConfig(int instanceNumber)
     return cfg;
 }
 
+void
+resetTxInternalState(Application& app)
+{
+    // seed randomness
+    srand(1);
+    gRandomEngine.seed(1);
+
+// reset caches to clear persistent state
+#ifdef FUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION
+    app.getLedgerTxnRoot().resetForFuzzer();
+#endif // FUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION
+    app.getDatabase().clearPreparedStatementCache();
+}
+
+// FuzzTransactionFrame is a specialized TransactionFrame that includes
+// useful methods for fuzzing such as an attemptApplication method for reseting
+// ledger state and deterministically attempting application of transactions.
+class FuzzTransactionFrame : public TransactionFrame
+{
+  public:
+    FuzzTransactionFrame(Hash const& networkID,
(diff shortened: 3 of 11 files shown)


src/ledger/LedgerTxn.cpp
```diff
@@ -1288,6 +1289,21 @@ LedgerTxnRoot::Impl::~Impl()
     }
 }
 
+#ifdef FUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION
+void
+LedgerTxnRoot::Impl::clearBestOfferAndEntryCaches()
```
MonsieurNicolas: just call this method something more generic, like `resetForFuzzer`

src/ledger/LedgerTxn.cpp
```diff
@@ -1903,6 +1919,11 @@ LedgerTxnRoot::Impl::getNewestVersion(LedgerKey const& key) const
     }
 
     putInEntryCache(key, entry, LoadType::IMMEDIATE);
+
+#ifdef FUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION
+    mDatabase.clearPreparedStatementCache();
```
MonsieurNicolas: why isn't this done in `resetForFuzzer`?

src/main/CommandLine.cpp
```diff
@@ -159,6 +159,20 @@ fileNameParser(std::string& string)
     return requiredArgParser(string, "FILE-NAME");
 }
 
+clara::Opt
+processIDParser(uint& num)
+{
+    return clara::Opt{num, "PROCESS-ID"}["--process_id"](
+        "for spawning multiple instances in fuzzing paralellization");
```
MonsieurNicolas: sp: "parallelization"

src/main/CommandLine.cpp
```diff
@@ -159,6 +159,20 @@ fileNameParser(std::string& string)
...
+}
+
+clara::Opt
+isOverlayFuzzer(bool& overlayFuzz)
+{
+    return clara::Opt{overlayFuzz}["--overlay_fuzz"](
```
MonsieurNicolas: wouldn't it be cleaner to have something like
`fuzz --mode=overlay` or `fuzz --mode=tx` ; that way we can add as many fuzzers as we want

src/test/fuzz.cpp
```diff
@@ -1,194 +0,0 @@
-// Copyright 2015 Stellar Development Foundation and contributors. Licensed
-// under the Apache License, Version 2.0. See the COPYING file at the root
-// of this distribution or at http://www.apache.org/licenses/LICENSE-2.0
-
-#include "util/asio.h"
```
MonsieurNicolas: no need to move files to a different folder, it just adds noise to the tree

src/test/fuzz/FuzzTransaction.cpp
```diff
@@ -0,0 +1,66 @@
...
+
+void
+attemptTx(TransactionFramePtr tx, Application& app)
+{
+    LedgerTxn ltx(app.getLedgerTxnRoot());
+    ++ltx.loadHeader().current().ledgerSeq;
+
+    TransactionResultCode code;
+    {
+        LedgerTxn ltxFeeProc(ltx);
```
MonsieurNicolas: I don't think we need this (processing fees)

src/test/fuzz/FuzzTransaction.cpp
```diff
@@ -0,0 +1,66 @@
...
+    }
+
+    try
+    {
+        LedgerTxn ltxTx(ltx);
+        tx->apply(app, ltxTx);
+        // after application, clear cache so that they don't persist between
+        // executions and confuse the fuzzer, causing it to think some Tx with
+        // [op0...opN] can result in multiple different execution paths.
+        app.getLedgerTxnRoot().clearBestOfferAndEntryCaches();
```
MonsieurNicolas: I thought the `reset` code would be done *before* running a test (at the same time than reseeding)

src/test/fuzz/FuzzerImpl.cpp
```diff
@@ -0,0 +1,192 @@
...
+#include "test/fuzz/Fuzzer.h"
+#include "test/fuzz/deserialize.h"
+#include "test/test.h"
+#include "transactions/TransactionUtils.h"
+#include "xdr/Fuzz-test.h"
+
+using namespace stellar;
+
+// for a given XDRInputFileStream, continuously try to read and inject an XDR
+template <typename T, typename Injection>
```
MonsieurNicolas: you don't need templating for this

src/test/fuzz/FuzzerImpl.cpp
```diff
@@ -0,0 +1,192 @@
...
+            LOG(INFO) << "Caught XDR error '" << e.what();
+            return true;
+        }
+    };
+
+    T inp;
+    while (tryRead(in, inp))
+    {
+        Fuzzer::seedRandomness();
+        LOG(INFO) << "Fuzzer injecting message ";
```
MonsieurNicolas: I don't think logging is a good idea in the main fuzz loop

src/test/fuzz/FuzzerImpl.cpp
```diff
@@ -0,0 +1,192 @@
...
+void
+TransactionFuzzer::inject(stellar::XDRInputFileStream& in)
+{
+    tryReadLoop<FuzzTransactionInput>(
+        in, [app = mApp, accounts = mAccounts](FuzzTransactionInput inp) {
+            auto txFramePtr = deserializeFuzzTransactionInput(inp, accounts);
+
+            try
+            {
+                fuzztest::attemptTx(txFramePtr, *app);
```
MonsieurNicolas: this should just be a regular method, not calling statics like that

src/test/fuzz/deserialize.cpp
```diff
@@ -0,0 +1,94 @@
...
+    // returningthe account at that index.
+    auto end_byte = static_cast<unsigned int>(index);
+    return accounts.at(end_byte % accounts.size());
+}
+
+namespace stellar
+{
+
+TransactionFramePtr
+deserializeFuzzTransactionInput(FuzzTransactionInput const& inp,
```
MonsieurNicolas: I don't think you need this machinery:
if you generate test accounts that have a "public key" 000000XX, then the mapping happens implicitly and you don't have to scan the XDR to search/replace

src/test/fuzz/fuzz.cpp
```diff
@@ -0,0 +1,190 @@
...
+            filename, isDesiredXDR);
+    }
+}
+}
+
+#ifdef FUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION
+namespace xdr
+{
+template <>
+typename std::enable_if<xdr_traits<stellar::PublicKey>::is_union>::type
```
MonsieurNicolas: do you need this?

src/test/fuzz/fuzz.cpp
```diff
@@ -0,0 +1,190 @@
...
+generator_t::operator()(stellar::PublicKey& t) const
+{
+    const auto& vals = stellar::PublicKey::_xdr_case_values();
+    typename xdr_traits<stellar::PublicKey>::discriminant_type v;
+
+    // Just pick a random case
+    uint32_t n;
```
MonsieurNicolas: not sure I understand what you're trying to do here

src/xdr/Fuzz-test.x
```diff
@@ -1,15 +1,15 @@
...
-#include "util/Logging.h"
+%#include "xdr/Stellar-transaction.h"
 
 namespace stellar
 {
 
-void fuzz(std::string const& filename, el::Level logLevel,
-          std::vector<std::string> const& metrics);
-void genfuzz(std::string const& filename);
+struct FuzzTransactionInput
```
MonsieurNicolas: you don't need a custom type for this (and the xdr file): all you need to do is generate random operations

src/test/fuzz/deserialize.cpp
robertDurst: Ok, I see... at first I considered the impossibility of predictability generating ED25519 keypairs, but then remembered that we aren't checking signatures here, so secret keys don't matter!

robertDurst:
More or less addressed everything above. Some of the hacky-est parts here will be benefited by the work being done in #2179. Specifically, I created a method called `resetAndApplyForFuzzer` that calls `resetResults` and `applyOperations` on a `TransactionFrame`, in order to bypass fee processing. I also manually construct a transaction envelope, replicating some of the code in `TestAccount` since I only have public keys and only care about the public keys of my pregenerated accounts.

Will move this to **Ready for Review** once the PR mentioned above and #2155 are complete.

src/ledger/LedgerTxn.cpp
```diff
@@ -83,7 +83,8 @@ EntryIterator::getImpl() const
     return mImpl;
 }
 
-EntryIterator& EntryIterator::operator++()
```
vogel: looks like you are using different version of clang-format that everyone

src/test/fuzz.cpp
```diff
@@ -1,194 +1,186 @@
...
+// default "tx" if unrecognized
+FuzzerMode
+getFuzzerModefromString(std::string const& fuzzerMode)
+{
+    if (iequals(fuzzerMode, "overlay"))
+    {
+        return FuzzerMode::OVERLAY;
+    }
+
+    if (iequals(fuzzerMode, "tx") || iequals(fuzzerMode, "transaction"))
```
vogel: that seems not necessary because of default return value

src/ledger/LedgerTxn.cpp
robertDurst: Fixing, it looks like I was using 7 instead of 5 -- thanks!

src/test/fuzz.cpp
robertDurst: Good point, done.

robertDurst:
Rebased against master. With #2179 merged, going to make some changes here based on that PR and undraft later today.

robertDurst:
This PR is now ready for review! 

**Areas of uncertainty**:
* currently for initial setup, I pregenerate accounts. As it is currently implemented, I generate 16 public keys (reasons explained in comments) and add these to the ledger via creation of `LedgerEntry` objects. As I played around a bit with order book initial state setup, I do not think this elegantly scales to more complex state setup. Originally I had a ``FuzzAccount` (subset of `TestAccount`) object and utilized `TxTests`  which made this stuff a bit easier, however this comes with some overhead such as many methods having a `REQUIRE` macro that I do not want/can't use. Do note, this is only an issue for initial setup -- now that I pregenerate accounts within a predicable range, I can directly use fuzzed XDR's as input.
* Not really commentary on the current PR, but related. I have discussed with Nicolas what makes a good initial corpus for fuzzing. Ideally this would be running `genFuzz` many, many (1000's/10000's) of times and then using AFL's corpus minimization tool `afl-cmin`. Thus, there is no reason to run this fuzz initial corpus generator every single time I want to `fuzz`, may be worth having a repo or home for large, minimized initial corpuses.

robertDurst:
Appears I broke something. Not fuzzing correctly anymore. Will work on fixing tomorrow.

robertDurst:
My apologies for not finding the above bug before opening for review. Tested and everything works again as expected. Should be good for review again.

src/test/fuzz.cpp
```diff
@@ -227,4 +227,16 @@ genfuzz(std::string const& filename)
         }
     }
 }
+// default "TRANSACTION" if unrecognized or unspecified
```
MonsieurNicolas: no silent failures please: if not recognized -> help

src/test/fuzz.cpp
```diff
@@ -1,144 +1,83 @@
...
-msgSummary(StellarMessage const& m)
-{
-    xdr::detail::Printer p(0);
-    xdr::archive(p, m.type(), nullptr);
-    return p.buf_.str() + ":" + hexAbbrev(sha256(xdr::xdr_to_msg(m)));
-}
-
-bool
-tryRead(XDRInputFileStream& in, StellarMessage& m)
+template <typename XDR_TYPE, typename Predicate>
```
MonsieurNicolas: it would be cleaner to have this functionality attached to the respective fuzz class, that way you can share the validation code between genfuzz and regular fuzzing.
And if you do it like this, you won't need templates

src/test/fuzz.cpp
```diff
@@ -147,86 +86,42 @@ fuzz(std::string const& filename, el::Level logLevel,
     while (__AFL_LOOP(PERSIST_MAX))
 #endif // AFL_LLVM_MODE
     {
-        XDRInputFileStream in(MAX_MESSAGE_SIZE);
+        XDRInputFileStream in(MAX_XDR_SIZE);
```
MonsieurNicolas: `MAX_XDR_SIZE` doesn't mean anything: this is the maximum xdr object size

src/test/fuzz.cpp
```diff
@@ -1,144 +1,83 @@
...
+    }
+    }
+}
 }
 
-#define PERSIST_MAX 10000000
-#define INITIATOR 1
-#define ACCEPTOR 0
+#define PERSIST_MAX 1000000
+#define MAX_XDR_SIZE 0x1000000
```
MonsieurNicolas: I don't think you should use this value `0x1000000` for transactions (way too big)

src/test/fuzz.cpp
```diff
@@ -117,10 +116,32 @@ genfuzz(std::string const& filename, FuzzerMode fuzzerMode)
...
     {
+        // TODO: allow for more operations, not just account merge and
+        // native payments
+        auto predicate = [](Operation& op) {
+            return op.body.type() == ACCOUNT_MERGE ||
+                   (op.body.type() == PAYMENT &&
+                    op.body.paymentOp().asset.type() == ASSET_TYPE_NATIVE);
+        };
+
+        auto isDesiredXDR = [predicate](xdr::xvector<Operation>& ops) {
```
MonsieurNicolas: rework this: this inefficient and convoluted

src/test/fuzz.cpp
```diff
@@ -117,10 +116,32 @@ genfuzz(std::string const& filename, FuzzerMode fuzzerMode)
     }
     case FuzzerMode::TRANSACTION:
     {
+        // TODO: allow for more operations, not just account merge and
+        // native payments
+        auto predicate = [](Operation& op) {
```
MonsieurNicolas: this predicate should be shared (see previous comment on genfuzz)

src/test/fuzz.cpp
```diff
@@ -117,10 +116,32 @@ genfuzz(std::string const& filename, FuzzerMode fuzzerMode)
     }
     case FuzzerMode::TRANSACTION:
     {
+        // TODO: allow for more operations, not just account merge and
```
MonsieurNicolas: remove TODO from codebase

src/test/fuzz.cpp
```diff
@@ -155,3 +155,20 @@ getFuzzerModefromString(std::string const& fuzzerMode)
...
+template <>
+void
+generator_t::operator()(stellar::PublicKey& t) const
+{
+    // generate public keys such that it is zero'd out except the last byte,
+    // hence for ED25519 public keys, an uint256 defined as opaque[32] set the
+    // 31st indexed byte. Furthermore, utilize only the first 4 bits of the 32nd
+    // byte, allowing us to generate these accounts against a pregeneration of
+    // such accountID's.
+    t.ed25519().at(31) = autocheck::generator<uint8_t>()(10) % 16;
```
MonsieurNicolas: I thought the generator for scalars was taking the max value?

MonsieurNicolas: also do not use hardcoded numbers for the number of accounts

src/test/FuzzerImpl.cpp
```diff
@@ -37,6 +37,7 @@ getFuzzConfig(int instanceNumber)
     cfg.QUORUM_INTERSECTION_CHECKER = false;
     cfg.PREFERRED_PEERS_ONLY = false;
     cfg.RUN_STANDALONE = true;
+    cfg.BUCKET_DIR_PATH = "Bucket" + std::to_string(instanceNumber);
```
MonsieurNicolas: you should not need this

src/test/FuzzerImpl.cpp
```diff
@@ -114,6 +115,104 @@ class FuzzTransactionFrame : public TransactionFrame
...
+    VirtualClock clock;
+    mApp = createTestApplication(clock, getFuzzConfig(processID));
+
+    LedgerTxn ltx(mApp->getLedgerTxnRoot());
+
+    // setup the state, for this we only need to  pregenerate some accounts. For
+    // now we create 16 accounts, or enough to fill the first 4 bits such that
+    // we have a pregenerated account for the last 4 bits of the 32nd byte of a
+    // public key
+    for (int i = 0; i < 16; ++i)
```
MonsieurNicolas: do not use hardcoded numbers

src/test/FuzzerImpl.cpp
```diff
@@ -114,6 +115,104 @@ class FuzzTransactionFrame : public TransactionFrame
...
+
+        // manually construct ledger entries, "creating" each account
+        LedgerEntry newAccountEntry;
+        newAccountEntry.data.type(ACCOUNT);
+        auto& newAccount = newAccountEntry.data.account();
+        newAccount.thresholds[0] = 1;
+        newAccount.accountID = publicKey;
+        newAccount.seqNum = 0;
+        // to create "interesting" balances we utilize powers of 2
+        newAccount.balance = pow(2, i);
```
MonsieurNicolas: do not use `pow` for integers (shift operator)

src/test/fuzz.cpp
robertDurst: Ah yes... was just reading how someone 8x'd their speed by limiting to 50kb

robertDurst: Averaged XDR size of 100k `path_payment` only XDR's in the range of 1-3 ops and got `294 bytes`. So I think `0x500` should be fine.

robertDurst:
Ok, should be good for another review.

src/test/FuzzerImpl.cpp
```diff
@@ -48,6 +48,72 @@ seedRandomness()
...
+        seedRandomness();
+
+        // before application, clear cache so that they don't persist
+        // between executions and confuse the fuzzer, causing it to
+        // think some Tx with [op0...opN] can result in multiple
+        // different execution paths.
+        app.getLedgerTxnRoot().resetForFuzzer();
+        app.getDatabase().clearPreparedStatementCache();
+
+        // reset results of operations
```
jonjove: This entire block looks nearly duplicate to `TransactionFrame::resetResults`. Can you refactor to share code?

src/test/FuzzerImpl.cpp
```diff
@@ -90,6 +90,7 @@ class FuzzTransactionFrame : public TransactionFrame
         SignatureChecker signatureChecker{
             ltx.loadHeader().current().ledgerVersion, getContentsHash(),
             mEnvelope.signatures};
+        loadSourceAccount(ltx, ltx.loadHeader());
```
jonjove: Probably good to have a comment explaining why this is here.

src/test/FuzzerImpl.cpp
```diff
@@ -114,6 +115,104 @@ class FuzzTransactionFrame : public TransactionFrame
...
+    while (tryRead(in, ops))
+    {
+        // direct the fuzzer such that it will "learn" that five transactions
+        // is the maximum since nothing over five transactions will ever
+        // result in newly discovered execution paths
+        if (ops.size() == 0 || ops.size() > 5)
+            return;
+
+        // construct transaction
+        auto txFramePtr = ::FuzzTransactionFrame::create(mSourceAccountID, ops,
```
jonjove: Nitpick: `::FuzzTransactionFrame` -> `FuzzTransactionFrame`

src/test/FuzzerImpl.cpp
```diff
@@ -114,6 +115,104 @@ class FuzzTransactionFrame : public TransactionFrame
...
+
+        try
+        {
+            // attempt to apply transaction
+            LedgerTxn ltx(mApp->getLedgerTxnRoot());
+            txFramePtr->attemptApplication(*mApp, ltx);
+
+            // we need an extra db cache clear here since the db prepared
+            // statement cache may persist between tryRead loops
+            mApp->getDatabase().clearPreparedStatementCache();
```
jonjove: Why is this needed? You are already clearing this in `attemptApplication`.

src/test/FuzzerImpl.cpp
```diff
@@ -114,6 +115,104 @@ class FuzzTransactionFrame : public TransactionFrame
...
+        }
+        catch (InvariantDoesNotHold const&)
+        {
+            throw;
+        }
+        catch (TransactionResultCode const& err)
```
jonjove: I don't think we ever actually throw a `TransactionResultCode`. How is this block reached? I think to catch internal errors correctly, you need to interpret any exception other than `InvariantDoesNotHold` as an internal error. Based on the comment in the next block, this might be a non-trivial change requiring some validity checks to filter out other error sources.

src/test/FuzzerImpl.cpp
```diff
@@ -216,6 +214,39 @@ TransactionFuzzer::inject(stellar::XDRInputFileStream& in)
...
+    XDROutputFileStream out;
+    out.open(filename);
+    autocheck::generator<Operation> gen;
+    try
+    {
+        xdr::xvector<Operation> ops;
+        auto numops = autocheck::generator<uint>()(6);
+        for (int i = 0; i < numops; ++i)
+        {
+            Operation op = gen(14);
```
jonjove: Why `14` here and two lines later? Try to avoid magic numbers whenever possible

src/test/FuzzerImpl.cpp
```diff
@@ -314,4 +330,37 @@ OverlayFuzzer::inject(stellar::XDRInputFileStream& in)
...
+
+void
+OverlayFuzzer::generateFuzzerInput(const std::string& filename)
+{
+    XDROutputFileStream out;
+    out.open(filename);
+    autocheck::generator<StellarMessage> gen;
+    try
+    {
+        StellarMessage m(gen(10));
```
jonjove: Same comment about magic numbers.

src/test/FuzzerImpl.cpp
```diff
@@ -216,6 +214,39 @@ TransactionFuzzer::inject(stellar::XDRInputFileStream& in)
...
+        {
+            Operation op = gen(14);
+            while (isBadInput(op))
+                op = gen(14);
+            ops.emplace_back(op);
+        }
+
+        out.writeOne(ops);
+    }
+    catch (xdr::xdr_bad_discriminant const&)
```
jonjove: Is it possible to change the above so `xdr_bad_discriminant` never happens?

src/test/FuzzerImpl.cpp
```diff
@@ -314,4 +330,37 @@ OverlayFuzzer::inject(stellar::XDRInputFileStream& in)
...
+    autocheck::generator<StellarMessage> gen;
+    try
+    {
+        StellarMessage m(gen(10));
+        while (isBadInput(m))
+            m = gen(10);
+
+        out.writeOne(m);
+    }
+    catch (xdr::xdr_bad_discriminant const&)
```
jonjove: Same question about `xdr_bad_discriminant`.

src/test/fuzz.h
```diff
@@ -1,15 +1,23 @@
 #pragma once
 
-// Copyright 2015 Stellar Development Foundation and contributors. Licensed
+// Copyright 2019 Stellar Development Foundation and contributors. Licensed
```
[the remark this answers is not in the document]
robertDurst: So I asked Marta and the consensus seems to be, for any new files use this year, otherwise don't touch -- switching all the touched dates back.

src/test/FuzzerImpl.cpp
robertDurst: I actually do not believe this is triggered since these xdr's are generated by `autocheck`, which knows enough to generate valid structures.

src/test/FuzzerImpl.h
```diff
@@ -0,0 +1,43 @@
...
+
+namespace stellar
+{
+
+class TransactionFuzzer : public Fuzzer
+{
+  public:
+    TransactionFuzzer(int numAccounts, int processID);
+    void inject(stellar::XDRInputFileStream&) override;
+    static void generateFuzzerInput(const std::string& filename);
```
marta-lokhova: `generateFuzzerInput` and `isBadInput`should be non-member non-friend functions for better encapsulation - just make them free functions inside of a namespace, e.g. FuzzUtils. (by the way, you can read up here on why this is a better practice: http://www.drdobbs.com/cpp/how-non-member-functions-improve-encapsu/184401197)

src/test/FuzzerImpl.h
```diff
@@ -0,0 +1,43 @@
...
+
+class OverlayFuzzer : public Fuzzer
+{
+    const uint ACCEPTOR_INDEX = 0;
+    const uint INITIATOR_INDEX = 1;
+
+  public:
+    OverlayFuzzer();
+    void inject(stellar::XDRInputFileStream&) override;
+    static void generateFuzzerInput(const std::string& filename);
```
marta-lokhova: Same comment for `generateFuzzerInput ` and `isBadInput`

src/test/FuzzerImpl.cpp
```diff
@@ -0,0 +1,366 @@
...
+    mSimulation->addNode(v11SecretKey, qSet0);
+
+    mSimulation->addPendingConnection(v10SecretKey.getPublicKey(),
+                                      v11SecretKey.getPublicKey());
+
+    mSimulation->startAllNodes();
+
+    // crank until nodes are connected
+    mSimulation->crankUntil(
+        [simulation = mSimulation, acceptorIndex = ACCEPTOR_INDEX,
```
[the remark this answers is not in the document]
robertDurst: Ah yes, skimmed *Effective Modern C++* chapter on this the other day, but on closer look here there is no chance of dangling reference.

src/test/fuzz.cpp
```diff
@@ -1,144 +1,63 @@
...
-    return p.buf_.str() + ":" + hexAbbrev(sha256(xdr::xdr_to_msg(m)));
-}
-
-bool
-tryRead(XDRInputFileStream& in, StellarMessage& m)
+std::unique_ptr<Fuzzer>
+createFuzzer(int processID, FuzzerMode fuzzerMode)
 {
-    try
+    switch (fuzzerMode)
```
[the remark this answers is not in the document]
robertDurst: So originally I defaulted to `TRANSACTION` mode, however, above in one of the collapsed comments, Nicolas advised against this: *"no silent failures please: if not recognized -> help*".

Open to either.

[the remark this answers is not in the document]
marta-lokhova: what I mean here is that if a user doesn't provide a mode, default to overlay. If a user does provide a mode, but it's invalid, don't run the fuzzer, print help message. I don't see any benefit of having a fuzzer in unknown mode (we can't even do fuzzing with it, plus it's error-prone).

src/test/FuzzerImpl.h
robertDurst: Thanks for the link!

src/test/FuzzerImpl.cpp
robertDurst: Ok, so you are correct, TransactionResultCode is never thrown. It was thrown in the original version of the PR, but is no longer, so that has been removed. 

Also, I figured out that the **other** exceptions, such as the ones I note in the comment are due to me removing a `checkValid` call. I originally removed this because I though it just checked signatures and such -- however it checks if operations are well-formed as well, and as far as I can tell, the other exceptions being caught were all mal-formed XDR exceptions such as `terminate called after throwing an instance of 'xdr::xdr_wrong_union'`.

My solution for this is re-including the `checkValid` and if it fails for any operation, returning early, and not *"applying the transaction"*.

src/test/FuzzerImpl.cpp
robertDurst: I cannot quite figure out why it is needed -- for some reason, calling this before the `XDRStream` reads another `XDR` in `tryRead` makes the stability go from `91% --> 95%` even though I am fairly confident the two are wholly unrelated.

robertDurst: *Another general observation:* turns out that the first `clearPreparedStatementCache` has generally no affect on stability, given the existence of this, the second `clearPreparedStatementCache` call. This makes sense given that it clears the cache after the `tx attempted apply`, before another loop iteration, and before `XDRStream`.

robertDurst:
Ok @jonjove and @marta-lokhova, I believe I fixed up the code to adhere to your feedback. Two places I am not 100% sure is:
* for db `clearPreparedCache`, I had this in two places. I removed the first one since I believe it makes most sense at the end, or at least as much sense as the beginning -- I also could not quite figure out why it had an affect on stability when placed before `XDRStreamInput` like it did; placed at the end it not only clears the statements between loops, but also makes stability happy.
* for the `--mode` cli flag, I have duplicated code for parsing the fuzzerMode input. However I did this because I kept running into lambda capture by reference dangling reference errors when I tried to extract these to a function (*this way of passing non-const references to multiple lambdas does not seem good, but its seems to be the way to do this based on the other command parsers*).

I also cleaned up my sporadic commits. The first commit that addressed both y'all's feedback is (_13be3d1f3fd2ff34f82fbe4ec252db3415bfa2d7_) or the first one dated July 24.

src/test/FuzzerImpl.cpp
```diff
@@ -0,0 +1,358 @@
...
+    autocheck::generator<Operation> gen;
+    xdr::xvector<Operation> ops;
+    auto numops =
+        autocheck::generator<uint>()(FUZZER_INITIAL_CORPUS_MAX_OPERATIONS);
+    for (int i = 0; i < numops; ++i)
+    {
+        Operation op = gen(FUZZER_INITIAL_CORPUS_OPERATION_GEN_SIZE);
+        while (isBadTransactionFuzzerInput(op))
+        {
+            op = gen(FUZZER_INITIAL_CORPUS_OPERATION_GEN_SIZE);
```
MonsieurNicolas: same comment than for overlay: don't use `FUZZER_INITIAL_CORPUS_OPERATION_GEN_SIZE` here but some higher number

src/test/FuzzerImpl.cpp
```diff
@@ -0,0 +1,358 @@
...
+  public:
+    FuzzTransactionFrame(Hash const& networkID,
+                         TransactionEnvelope const& envelope)
+        : TransactionFrame(networkID, envelope){};
+
+    void
+    attemptApplication(Application& app, AbstractLedgerTxn& ltx)
+    {
+        // seed randomness
+        seedRandomness();
```
MonsieurNicolas: calls to `seedRandomness` and `resetForFuzzer` don't belong here, should be done in `inject`

src/test/FuzzerImpl.cpp
```diff
@@ -0,0 +1,358 @@
...
+        // think some Tx with [op0...opN] can result in multiple
+        // different execution paths.
+        app.getLedgerTxnRoot().resetForFuzzer();
+
+        // reset results of operations
+        resetResults(ltx.getHeader(), 0);
+
+        // attempt application of transaction without accounting for sequence
+        // number, processing the fee, or committing the LedgerTxn
+        mCachedAccount.reset();
```
MonsieurNicolas: why do you need to reset `mCachedAccount` ? Don't we want to let it do its thing depending on protocol version?

src/test/FuzzerImpl.cpp
MonsieurNicolas: actually, should probably be moved in some method that can be called from multiple places `resetTxInternalState`

src/test/FuzzerImpl.cpp
```diff
@@ -0,0 +1,358 @@
...
+    // (stability) and binary coverage.
+    return m.type() == AUTH || m.type() == ERROR_MSG || m.type() == HELLO;
+}
+
+TransactionFuzzer::TransactionFuzzer(int numAccounts, int processID)
+{
+    VirtualClock clock;
+    mApp = createTestApplication(clock, getFuzzConfig(processID));
+
+    LedgerTxn ltx(mApp->getLedgerTxnRoot());
```
MonsieurNicolas: you need to call `resetTxInternalState` before creating `ltx`

src/test/FuzzerImpl.cpp
```diff
@@ -0,0 +1,358 @@
...
+            return;
+        }
+
+        // construct transaction
+        auto txFramePtr = createFuzzTransactionFrame(mSourceAccountID, ops,
+                                                     mApp->getNetworkID());
+
+        try
+        {
+            // attempt to apply transaction
```
MonsieurNicolas: should call `resetTxInternalState` before creating ltx

src/test/FuzzerImpl.cpp
MonsieurNicolas: `resetTxInternalState` should call `clearPreparedStatementCache`

src/test/FuzzerImpl.cpp
```diff
@@ -0,0 +1,358 @@
...
+            // statement cache may persist between tryRead loops
+            mApp->getDatabase().clearPreparedStatementCache();
+        }
+        catch (InvariantDoesNotHold const&)
+        {
+            throw;
+        }
+        catch (...)
+        {
+            throw std::runtime_error(fmt::format("Unexpected internal error"));
```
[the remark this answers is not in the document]
MonsieurNicolas: (loss of context)

src/test/FuzzerImpl.cpp
MonsieurNicolas: I suspect you need to call `clearPreparedStatementCache` after `ltx` goes out of scope just to be sure things are reset the way you want

src/test/fuzz.cpp
```diff
@@ -147,84 +62,64 @@ fuzz(std::string const& filename, el::Level logLevel,
...
-            while (acceptor->getApp().getClock().cancelAllEvents())
-                ;
-        }
+        fuzzer->inject(in);
     }
 }
 
 void
-genfuzz(std::string const& filename)
+genfuzz(std::string const& filename, FuzzerMode fuzzerMode)
```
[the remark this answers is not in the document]
robertDurst: Not quite sure I follow here. 

Currently each fuzzer has a corresponding free function in the `FuzzerImpl` file with generation logic, that shares `isBadInput` with the fuzzer. These generation methods must either be free or static, because we do not need (and do not want) to initialize a fuzzer and start up apps just to create a random XDR. That being said, I could separate the initialization of the fuzzer to a separate method, and out of the constructor but I am not sure if this is the best route.

src/test/FuzzerImpl.cpp
robertDurst: This was part of the [`TransactionFrame::apply` method](https://github.com/stellar/stellar-core/blob/master/src/transactions/TransactionFrame.cpp#L655) I recreated here. I believe this reset for `mCachedAccount` seems to happen regardless of protocol.

src/test/FuzzerImpl.cpp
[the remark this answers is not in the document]
robertDurst: No reason. Will fix. Is the following bad practice?
```
catch(...) {
   throw;
}
```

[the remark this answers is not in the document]
MonsieurNicolas: if you're not doing anything, yes it's bad practice as it doesn't achieve anything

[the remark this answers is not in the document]
robertDurst: Ok, makes sense -- therefore, this try catch in itself is not needed since we just propagate all exceptions.

src/test/FuzzerImpl.cpp
MonsieurNicolas: I asked because `loadSourceAccount` seems to already do the right thing

src/test/FuzzerImpl.cpp
```diff
@@ -0,0 +1,360 @@
...
+                                                     mApp->getNetworkID());
+
+        resetTxInternalState(*mApp);
+        LedgerTxn ltx(mApp->getLedgerTxnRoot());
+
+        // attempt to apply transaction
+        txFramePtr->attemptApplication(*mApp, ltx);
+
+        // ensure everything reset as ltx goes out of scope
+        resetTxInternalState(*mApp);
```
MonsieurNicolas: ltx is still in scope here, you need to do something like
```c++
{
   txFramePtr->attemptApplication(*mApp, ltx);
}
resetTxInternalState(*mApp);
```

src/test/fuzz.cpp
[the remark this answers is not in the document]
MonsieurNicolas: yes, that's what I would like to see (switch on `mode` is basically what polymorphism does, in a cleaner way):
that way, each fuzzing mode is encapsulated in a dedicated class, if you move what is currently in the constructor into a method you'll have what you want.

src/test/FuzzerImpl.cpp
robertDurst: Yes, good point -- this was for a large portion of the fee processing and signature checking I left out. Removed.

src/test/fuzz.cpp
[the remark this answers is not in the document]
robertDurst: This should be addressed in my latest commits.

MonsieurNicolas:
r+ a9468013d6a4c955dba8580bba96789803698946

robertDurst:
@MonsieurNicolas accidentally turned off format for a commit, fixed now

MonsieurNicolas:
r+ 60b38b41535e1d31c092ecbe633a92bab03f9659

robertDurst:
My apologies -- the fuzzer utilized a method behind FUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION. Fixed.

MonsieurNicolas:
r+ c2cd2fc752aa68dedf3043999f1872c666b89a5e

PR #2183, opened 2019-07-04
Intersection criticality check

graydon:
Followup to #2127 : this adds a secondary check for "intersection-criticality", which is a somewhat made-up notion that attempts to capture when we're "one serious misconfiguration away" from non-intersection. The approach is described in detail in comments, but briefly:

  - We find all sets of leaf-nodes in qsets, and consider them candidate groups.
  - For each group, we run the quorum intersection checker on a modified graph in which the group's members have all been made "fickle", which is an overapproximation (for purposes of evaluating splitting risk) of "misconfiguration" that makes them willing to join (vote-for) any quorum of any node that depends on a node in the group.

This "fickleness" condition is weaker (and much more likely) than a byzantine failure: fickle nodes still model an intersecting, singly-voting node if they're a member of two quorums (i.e. we're not simulating them voting one way in one quorum and another way in another), they're just behaving as though someone set their threshold too low and/or over-connected them.

# Checklist
- [x] Compiles
- [x] Ran all tests

Still probably needs another few tests (took me forever to work out one that even works) and adjustment to the docs to describe the new field in the info / quorum endpoints. But ready for initial review I think.

### DIFF
--- a/src/herder/QuorumIntersectionCheckerImpl.cpp
+++ b/src/herder/QuorumIntersectionCheckerImpl.cpp
@@ -330,8 +330,11 @@ MinQuorumEnumerator::anyMinQuorumHasDisjointQuorum()
 ////////////////////////////////////////////////////////////////////////////////
 
 QuorumIntersectionCheckerImpl::QuorumIntersectionCheckerImpl(
-    QuorumTracker::QuorumMap const& qmap, Config const& cfg)
-    : mCfg(cfg), mLogTrace(Logging::logTrace("SCP")), mTSC(mGraph)
+    QuorumTracker::QuorumMap const& qmap, Config const& cfg, bool quiet)
+    : mCfg(cfg)
+    , mLogTrace(Logging::logTrace("SCP"))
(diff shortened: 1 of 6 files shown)


src/herder/QuorumIntersectionCheckerImpl.cpp
```diff
@@ -764,6 +764,75 @@ QuorumIntersectionCheckerImpl::networkEnjoysQuorumIntersection() const
...
+
+void
+findCriticalityCandidates(SCPQuorumSet const& p,
+                          std::set<std::set<PublicKey>>& candidates, bool root)
+{
+    // Make a singleton-set for every validator, always.
+    for (auto const& k : p.validators)
+    {
+        std::set<PublicKey> singleton{k};
+        candidates.insert(singleton);
```
MonsieurNicolas: today I learned that `operator<` is defined for `std::set` (and `std::map`)

MonsieurNicolas:
r+ a1cf409a5a8304134b0ae73920736cd6f3616e0a

PR #2184, opened 2019-07-05
Audit retries of the Work subsystem

### LINKED ISSUE #1261: Check which work items should be run as RETRY_NONE

marta-lokhova:
tl;dr we're retrying too many times, which could cause big delays

This PR does some cleanup of work retries. The downside of retrying as we currently do is that there could be some strange edge cases where all works and its children fail, and because we use random exponential backoff for each retry, it can create pretty big delays. Some important changes:
* GetAndUnzipRemoteFile work now retries less
* A few publish-related works that manage sequences do not really need to retry, as individual sequences handle that already. 
* WriteSnapshotWork was also retrying 32 times, but I suspect the biggest cause for its failure is no disk space, so it probably only needs to retry a couple times. 

This should also resolve #1261

### DIFF
--- a/src/catchup/ApplyBucketsWork.cpp
+++ b/src/catchup/ApplyBucketsWork.cpp
@@ -26,7 +26,7 @@ ApplyBucketsWork::ApplyBucketsWork(
     Application& app,
     std::map<std::string, std::shared_ptr<Bucket>> const& buckets,
     HistoryArchiveState const& applyState, uint32_t maxProtocolVersion)
-    : BasicWork(app, "apply-buckets", RETRY_A_FEW)
+    : BasicWork(app, "apply-buckets", BasicWork::RETRY_NEVER)
     , mBuckets(buckets)
     , mApplyState(applyState)
     , mApplying(false)
--- a/src/historywork/GzipFileWork.cpp
+++ b/src/historywork/GzipFileWork.cpp
@@ -10,7 +10,8 @@ namespace stellar
 
 GzipFileWork::GzipFileWork(Application& app, std::string const& filenameNoGz,
                            bool keepExisting)
-    : RunCommandWork(app, std::string("gzip-file ") + filenameNoGz)
+    : RunCommandWork(app, std::string("gzip-file ") + filenameNoGz,
+                     BasicWork::RETRY_A_LOT)
     , mFilenameNoGz(filenameNoGz)
     , mKeepExisting(keepExisting)
 {
--- a/src/historywork/PutSnapshotFilesWork.cpp
+++ b/src/historywork/PutSnapshotFilesWork.cpp
@@ -17,8 +17,11 @@ namespace stellar
 
 PutSnapshotFilesWork::PutSnapshotFilesWork(
     Application& app, std::shared_ptr<StateSnapshot> snapshot)
-    : Work(app, fmt::format("update-archives-{:08x}",
-                            snapshot->mLocalState.currentLedger))
+    : Work(app,
+           fmt::format("update-archives-{:08x}",
+                       snapshot->mLocalState.currentLedger),
+           // Each put-snapshot-sequence will retry correctly
+           BasicWork::RETRY_NEVER)
     , mSnapshot(snapshot)
 {
 }
@@ -45,8 +48,10 @@ PutSnapshotFilesWork::doWork()
 
             std::vector<std::shared_ptr<BasicWork>> seq{getState, putFiles,
                                                         putState};
-
-            addWork<WorkSequence>("put-snapshot-sequence", seq);
+            // Each inner step will retry a lot, so only retry the sequence once
+            // in case of an unexpected failure
+            addWork<WorkSequence>("put-snapshot-sequence", seq,
+                                  BasicWork::RETRY_ONCE);
         }
         mStarted = true;
     }
--- a/src/historywork/WriteSnapshotWork.cpp
+++ b/src/historywork/WriteSnapshotWork.cpp
@@ -12,9 +12,13 @@
 namespace stellar
 {
 
+// Note, WriteSnapshotWork does not have any special retry clean-up logic:
+// history items are written via XDROutputFileStream, which automatically
+// truncates any existing files.
 WriteSnapshotWork::WriteSnapshotWork(Application& app,
                                      std::shared_ptr<StateSnapshot> snapshot)
-    : BasicWork(app, "write-snapshot", Work::RETRY_A_LOT), mSnapshot(snapshot)
+    : BasicWork(app, "write-snapshot", BasicWork::RETRY_A_LOT)
+    , mSnapshot(snapshot)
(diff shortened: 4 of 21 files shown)


src/catchup/ApplyBucketsWork.cpp
```diff
@@ -26,7 +26,7 @@ ApplyBucketsWork::ApplyBucketsWork(
     Application& app,
     std::map<std::string, std::shared_ptr<Bucket>> const& buckets,
     HistoryArchiveState const& applyState, uint32_t maxProtocolVersion)
-    : BasicWork(app, "apply-buckets", RETRY_A_FEW)
```
vogel: why no RETRY_NEVER here?

vogel:
I think RETRY_A_LOT is not used anymore, so it can be removed.

vogel:
Question about `WorkSequence` instances. I see that most of them use RETRY_NEVER, but some different values. For example: `PutHistoryArchiveStateWork::spawnPublishWork` creates `WorkSequence` with two items: `MakeRemoteDirWork` and `PutRemoteFileWork`. So my understanding of it is that a following may happen:

- `MakeRemoteDirWork` succeeds after first run
- `PutRemoteFileWork` fails several times, so it cannot be retried again
- that means `WorkSequence` fails and is retries
- `MakeRemoteDirWork` is executed again, but that is not necessary as it created its dir just a while ago

I think that in most (every?) cases `WorkSequence` should not retry, only it subworks should.

vogel:
Also looking at usage of `maxRetries` arguments, those are rarely used: for `Work`, `BasicWork`, `WorkSequence` and once for `GetRemoteFileWork`. All of other classes use default values from their constructors.

marta-lokhova:
@vogel 

> I think that in most (every?) cases WorkSequence should not retry, only it subworks should.

There are cases where we want it to retry. For example, in download->verify sequences, verify should never retry, but instead re-download the file.

> Also looking at usage of maxRetries arguments, those are rarely used: for Work, BasicWork, WorkSequence and once for GetRemoteFileWork. All of other classes use default values from their constructors.

Some child works need it (e.g., `GetHistoryArchiveStateWork`), but I can clean up the ones that don't.

src/catchup/ApplyBucketsWork.cpp
marta-lokhova: i think you're right - I was under impression apply buckets can pick up where it left off and proceed in case there's a failure, but it looks like it doesn't even properly return `WORK_FAILURE`.

src/historywork/GzipFileWork.cpp
```diff
@@ -10,7 +10,8 @@ namespace stellar
 
 GzipFileWork::GzipFileWork(Application& app, std::string const& filenameNoGz,
                            bool keepExisting)
-    : RunCommandWork(app, std::string("gzip-file ") + filenameNoGz)
+    : RunCommandWork(app, std::string("gzip-file ") + filenameNoGz,
```
MonsieurNicolas: I am not sure I would not retry gzip: the only reason this would fail is because it runs out of disk space or it cannot spawn processes (we've seen both).
Retrying just that step seems desirable under those specific failure modes. Ideally it would even retry forever but there is a chance of some corruption (like for some reason the files it's trying to zip got deleted somehow) and we may be better off resetting the whole flow

src/historywork/PutSnapshotFilesWork.cpp
```diff
@@ -17,8 +17,11 @@ namespace stellar
 
 PutSnapshotFilesWork::PutSnapshotFilesWork(
     Application& app, std::shared_ptr<StateSnapshot> snapshot)
-    : Work(app, fmt::format("update-archives-{:08x}",
-                            snapshot->mLocalState.currentLedger))
+    : Work(app,
+           fmt::format("update-archives-{:08x}",
+                       snapshot->mLocalState.currentLedger),
+           // Each put-snapshot-sequence will retry correctly
+           BasicWork::RETRY_NEVER)
```
MonsieurNicolas: same thing here: uploads are likely to fail and retrying the whole thing doesn't seem necessary (so "retry a lot" was about right for this)

src/historywork/WriteSnapshotWork.cpp
```diff
@@ -12,9 +12,13 @@
 namespace stellar
 {
 
+// Note, WriteSnapshotWork does not have any special retry clean-up logic:
+// history items are written via XDROutputFileStream, which automatically
+// truncates any existing files.
 WriteSnapshotWork::WriteSnapshotWork(Application& app,
                                      std::shared_ptr<StateSnapshot> snapshot)
-    : BasicWork(app, "write-snapshot", Work::RETRY_A_LOT), mSnapshot(snapshot)
+    : BasicWork(app, "write-snapshot", BasicWork::RETRY_A_FEW)
```
MonsieurNicolas: I also do not see a reason to not retry a lot here

MonsieurNicolas: I think `PublishWork` doesn't even need to retry: it's designed to be rescheduled "forever"

src/historywork/PutSnapshotFilesWork.cpp
marta-lokhova: I'm not really sure what you mean: this work just keeps track of uploads to multiple archives. Each upload to an archive is a WorkSequence, which already handles retries in case of a failure, so I don't see a reason to "retry a lot".

src/historywork/WriteSnapshotWork.cpp
marta-lokhova: For `WriteSnapshotWork`: the only time it actually - transiently - fails is due to read/write transaction serialization order:
https://github.com/stellar/stellar-core/blob/89e994b5cffa304eb5449e40cc9fc7a35c205b2d/src/history/StateSnapshot.cpp#L138

I don't feel particularly strongly about retrying a lot here, it just seems like retrying 5 times with exponential backoff should be enough.

src/historywork/PutSnapshotFilesWork.cpp
MonsieurNicolas: yes you are right, I missed the nested work.
I think the general strategy seems to be:
* work in general should not retry
* topmost work should retry forever (typically done by rescheduling like how "PublishWork" is scheduled)
* "leaf" work should retry "a lot" (as to make progress)
* work sequences that spawn "leaf" work, may retry once (the reason a sequence like this would have to retry is because of some weird event that "should never happen" that is causing a leaf to be stuck, like a file got deleted on disk)

Now that I see it, the function `BasicWork::getRetryDelay()` allows for timeouts that are way too high: over an hour. We should cap to something more reasonable like up to 1024 seconds (so 10 instead of 12).

MonsieurNicolas:
r+ 1de4d5171aec43ec713cf94480e3c1257ef45973

PR #2193, opened 2019-07-11
Make `fullkeys` behavior consistent across endpoints

### LINKED ISSUE #2191: /quorum endpoint should allow node-id vs alias in name (Version: v11.1.0)
## Issue Description
When using the /quorum?transitive=true&fullkeys=true endpoint, the `node` property shows the alias, with no way to retrieve the node id.  This is necessary for many things including SEP20, in which i need the node IDs in the quorum to find their home domain.

## Steps to Reproduce

Fetch the following endpoint from your stellar-core node: `http://localhost:11626/quorum?transitive=true&fullkeys=true`

## Expected Result
I should see all `node` properties as their full key

## Actual Result
I still see their aliases.  This is a nice to have for humans looking over the config, but for anything programmatic, the node id is almost always better.  I would even consider creating a separate 'alias' field and only ever showing the node id in the node field.

## Your Environment and Setup
docker-quickstart v 11.1 on macOS

## Supporting Files
In the example below, i have no way to interact with node "sdf2"
```
{
         "distance" : 1,
         "heard" : 1238681,
         "node" : "sdf2",
         "qset" : {
            "t" : 2,
            "v" : [
               "GCUCJTIYXSOXKBSNFGNFWW5MUQ54HKRPGJUTQFJ5RQXZXNOLNXYDHRAP",
               "GC2V2EFSXN6SQTWVYA5EPJPBWWIMSD2XQNKUOHGEKB535AQE2I6IXV2Z",
               "GDKXE2OZMJIPOSLNA6N6F2BVCI3O777I2OOC4BV7VOYUEHYX7RTRYA7Y"
            ]
         },
         "status" : "tracking",
         "value" : "[ txH: e7cc6b, ct: 1562789100, upgrades: [ ] ]",
         "value_id" : 1
      },
```

MonsieurNicolas:
# Description

Resolves #2191 

This PR makes the behavior of `fullkeys` consistent across the entire code base:
when specified, aliases are never returned and by default we return either aliases or short keys

This PR also adds support for `fullkeys` to the "peers" endpoint

### DIFF
--- a/src/main/Config.h
+++ b/src/main/Config.h
@@ -308,8 +308,10 @@ class Config : public std::enable_shared_from_this<Config>
     void adjust();
 
     std::string toShortString(PublicKey const& pk) const;
-    std::string toStrKeyAuto(PublicKey const& pk, bool& isAlias) const;
-    std::string toStrKey(PublicKey const& pk) const;
+
+    // fullKey true => returns full StrKey corresponding to pk
+    //  otherwise, returns alias or shortString equivalent
+    std::string toStrKey(PublicKey const& pk, bool fullKey) const;
 
     bool resolveNodeID(std::string const& s, PublicKey& retKey) const;
 
--- a/src/main/Config.h
+++ b/src/main/Config.h
@@ -308,8 +308,9 @@ class Config : public std::enable_shared_from_this<Config>
     void adjust();
 
     std::string toShortString(PublicKey const& pk) const;
-    std::string toStrKey(PublicKey const& pk, bool& isAlias) const;
+    std::string toStrKeyAuto(PublicKey const& pk, bool& isAlias) const;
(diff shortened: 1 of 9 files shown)


src/main/Config.h
```diff
@@ -308,8 +308,9 @@ class Config : public std::enable_shared_from_this<Config>
     void adjust();
 
     std::string toShortString(PublicKey const& pk) const;
-    std::string toStrKey(PublicKey const& pk, bool& isAlias) const;
+    std::string toStrKeyAuto(PublicKey const& pk, bool& isAlias) const;
```
jonjove: This isn't actually used by anything at the end of this PR. Is that intentional?

graydon:
r+ 2ade539

PR #2194, opened 2019-07-11
Use ConditionalWork for history replay in catchup

### LINKED ISSUE #1951: catchup: applying transactions should be done in the same work unit than downloading

marta-lokhova:
Resolves #1951


src/work/ConditionalWork.cpp
```diff
@@ -0,0 +1,100 @@
...
+    BasicWork::shutdown();
+}
+
+bool
+ConditionalWork::onAbort()
+{
+    if (mWorkStarted && !mConditionedWork->isDone())
+    {
+        mConditionedWork->crankWork();
+        return false;
```
vogel: Does that mean that aborting started ConditionalWork is not possible?

src/work/test/WorkTests.cpp
```diff
@@ -812,3 +813,116 @@ TEST_CASE("Work batching", "[batching][work]")
...
+        while (!wm.allChildrenDone())
+        {
+            clock.crank();
+        }
+
+        REQUIRE(failedWork->getState() == BasicWork::State::WORK_FAILURE);
+        // Conditioned work was aborted
+        REQUIRE(conditionedWork->getState() == BasicWork::State::WORK_ABORTED);
+        // Dependent work was never started (internally, in PENDING state,
+        // enforced by its destructor)
```
vogel: I think you mean constructor here?

src/work/test/WorkTests.cpp
```diff
@@ -812,3 +813,116 @@ TEST_CASE("Work batching", "[batching][work]")
...
+        auto dependentWork =
+            std::make_shared<TestBasicWork>(*appPtr, "dependent-work");
+        auto conditionedWork = wm.scheduleWork<ConditionalWork>(
+            "condition-shutdown", condition, dependentWork);
+
+        while (conditionedWork->getState() != BasicWork::State::WORK_WAITING)
+        {
+            clock.crank();
+        }
+        CHECK(conditionedWork->getState() == BasicWork::State::WORK_WAITING);
```
vogel: That looks redundant.

src/catchup/ApplyCheckpointWork.cpp
```diff
@@ -38,41 +39,32 @@ ApplyLedgerChainWork::ApplyLedgerChainWork(
...
-
-    mCurrSeq = hm.checkpointContainingLedger(mRange.mFirst);
     mHdrIn.close();
     mTxIn.close();
     mFilesOpen = false;
 }
 
 void
-ApplyLedgerChainWork::openCurrentInputFiles()
+ApplyCheckpointWork::openCurrentInputFiles()
```
vogel: Could be renamed to openInputFiles. If this only works on one checkpoint there is only one set of files, so "current" is redundant.

src/work/BatchWork.cpp
```diff
@@ -45,7 +45,6 @@ BatchWork::doWork()
     }
 
     addMoreWorkIfNeeded();
-    mApp.getCatchupManager().logAndUpdateCatchupStatus(true);
```
vogel: Nice!

src/work/test/WorkTests.cpp
marta-lokhova: No, I mean destructor, which asserts that BasicWork is in proper state.

src/work/ConditionalWork.cpp
marta-lokhova: Not sure what you mean: `onAbort` returns whether Work is done aborting or not, so that BasicWork can proceed appropriately. `ConditionalWork` will keep getting scheduled to run `onAbort` until `mConditionedWork` is done aborting. At that point, `ConditionalWork` is safe to abort itself.

src/work/ConditionalWork.cpp
```diff
@@ -0,0 +1,100 @@
...
+    }
+}
+
+BasicWork::State
+ConditionalWork::onRun()
+{
+    if (mWorkStarted)
+    {
+        mConditionedWork->crankWork();
+        return mConditionedWork->getState();
```
MonsieurNicolas: I don't think this is correct:
`ConditionalWork`'s state will stay as `RUNNING` (even if `mConditionedWork` transitions to other states like WAITING).
Easiest might be to have conditional work derive from `Work` to deal with all the state transitions and proper sleeping.

src/catchup/ApplyCheckpointWork.h
```diff
@@ -38,20 +38,20 @@ struct LedgerHeaderHistoryEntry;
...
  */
 
-class ApplyLedgerChainWork : public BasicWork
+class ApplyCheckpointWork : public BasicWork
 {
     TmpDir const& mDownloadDir;
-    LedgerRange const mRange;
-    uint32_t mCurrSeq;
+    uint32_t mCheckpoint;
+    // Apply checkpoint until given ledger. Otherwise, apply checkpoint in full
```
MonsieurNicolas: pick a name that clarifies that this is a ledger number

src/catchup/ApplyCheckpointWork.h
```diff
@@ -63,10 +63,9 @@ class ApplyLedgerChainWork : public BasicWork
     bool applyHistoryOfSingleLedger();
 
   public:
-    ApplyLedgerChainWork(Application& app, TmpDir const& downloadDir,
-                         LedgerRange range,
-                         LedgerHeaderHistoryEntry& lastApplied);
-    ~ApplyLedgerChainWork() = default;
+    ApplyCheckpointWork(Application& app, TmpDir const& downloadDir,
+                        uint32_t checkpoint, uint32_t boundary = 0);
```
MonsieurNicolas: I don't like the use of 0 like this in a low level class as it's quite error prone.

I think it's probably better to keep a `LedgerRange` here, and force the caller to make a decision on what the upper range is.

This PR then just needs to enforce that that LedgerRange has the following properties:
1. low is a checkpoint
2. doesn't span multiple checkpoints

src/work/ConditionalWork.cpp
marta-lokhova: Clarified offline: `ConditionalWork` follows the state of `mConditionedWork`, so if the latter one transitions into a new state, the former work will just return that.  So this codepath should be right (we already do something similar in WorkSequence also)

src/catchup/ApplyCheckpointWork.h
marta-lokhova: Agreed that it's safer, updated.

src/catchup/ApplyCheckpointWork.cpp
```diff
@@ -244,35 +252,29 @@ ApplyLedgerChainWork::applyHistoryOfSingleLedger()
...
+        auto const& lm = mApp.getLedgerManager();
+        auto done = lm.getLastClosedLedgerNum() == mCheckpointRange.mLast;
 
-        auto& lm = mApp.getLedgerManager();
-        auto const& lclHeader = lm.getLastClosedLedgerHeader();
-        if (lclHeader.header.ledgerSeq == mRange.mLast)
+        if (result)
         {
-            return State::WORK_SUCCESS;
+            mApp.getCatchupManager().logAndUpdateCatchupStatus(true);
```
MonsieurNicolas: I'd call `logAndUpdateCatchupStatus` regardless (before the if) - it seems arbitrary to only call it if result is true.

MonsieurNicolas: I would also move the `done` case to its own it (probably done first) as to make the code easier to read (else will be deciding between running or failure)

marta-lokhova: done.

MonsieurNicolas:
r+ ad674d62e5c29e8577c154180f6b3c5529a0de2e

PR #2205, opened 2019-07-19
Remove bucketlist shadows, avoid resolving snapshots on publish 

marta-lokhova:
This PR implements [CAP-0025](https://github.com/stellar/stellar-protocol/blob/master/core/cap-0025.md), and makes a few changes to how we publish and catchup. Main change is shadow removal from the bucketlist, which changes how buckets are merged (and hence, producing different buckets, which affects ledger header hash). 

This change implies that on publish, it is no longer necessary to resolve future buckets. Moreover, it is not even necessary to store future buckets in half-merged state, since bucketlist can fully recover the merges by just using its currs and snaps. 

This change also implements the upgrade mechanism. Specifically, upon the upgrade (which will be properly timed), in-progress merges are dropped, and new-style merges are started.

### DIFF
--- a/src/bucket/Bucket.cpp
+++ b/src/bucket/Bucket.cpp
@@ -356,10 +356,18 @@ calculateMergeProtocolVersion(
     protocolVersion = std::max(oi.getMetadata().ledgerVersion,
                                ni.getMetadata().ledgerVersion);
 
+    // Starting with FIRST_PROTOCOL_SHADOWS_REMOVED,
+    // protocol version is determined as a max of curr, snap, and any shadow of
+    // version < FIRST_PROTOCOL_SHADOWS_REMOVED. This means that a bucket may
+    // still perform an old style merge despite the presence of the new protocol
+    // shadows.
     for (auto const& si : shadowIterators)
     {
-        protocolVersion =
-            std::max(si.getMetadata().ledgerVersion, protocolVersion);
+        auto version = si.getMetadata().ledgerVersion;
+        if (version < Bucket::FIRST_PROTOCOL_SHADOWS_REMOVED)
+        {
+            protocolVersion = std::max(version, protocolVersion);
+        }
     }
 
     CLOG(TRACE, "Bucket") << "Bucket merge protocolVersion=" << protocolVersion
@@ -386,6 +394,19 @@ calculateMergeProtocolVersion(
     else
     {
         ++mc.mPostInitEntryProtocolMerges;
+    }
+
+    if (protocolVersion < Bucket::FIRST_PROTOCOL_SHADOWS_REMOVED)
+    {
+        ++mc.mPreShadowRemovalProtocolMerges;
+    }
+    else
+    {
+        if (!shadowIterators.empty())
+        {
+            throw std::runtime_error("Shadows are not supported");
+        }
+        ++mc.mPostShadowRemovalProtocolMerges;
     }
 }
 
@@ -614,5 +635,13 @@ Bucket::merge(BucketManager& bucketManager, uint32_t maxProtocolVersion,
     MergeKey mk{maxProtocolVersion, keepDeadEntries, oldBucket, newBucket,
                 shadows};
     return out.getBucket(bucketManager, &mk);
+}
+
+uint32_t
+Bucket::getBucketVersion(std::shared_ptr<Bucket> const& bucket)
+{
+    assert(bucket);
+    BucketInputIterator it(bucket);
+    return it.getMetadata().ledgerVersion;
 }
 }
--- a/src/bucket/BucketList.cpp
+++ b/src/bucket/BucketList.cpp
@@ -12,6 +12,7 @@
 #include "main/Application.h"
 #include "util/Logging.h"
 #include "util/XDRStream.h"
+#include "util/format.h"
 #include "util/types.h"
 #include <cassert>
 
@@ -154,8 +155,12 @@ BucketLevel::prepare(Application& app, uint32_t currLedger,
         }
     }
 
-    mNextCurr = FutureBucket(app, curr, snap, shadows, currLedgerProtocol,
-                             countMergeEvents, mLevel);
+    auto shadowsBasedOnProtocol =
+        Bucket::getBucketVersion(snap) >= Bucket::FIRST_PROTOCOL_SHADOWS_REMOVED
+            ? std::vector<std::shared_ptr<Bucket>>()
+            : shadows;
+    mNextCurr = FutureBucket(app, curr, snap, shadowsBasedOnProtocol,
+                             currLedgerProtocol, countMergeEvents, mLevel);
     assert(mNextCurr.isMerging());
 }
 
@@ -421,6 +426,14 @@ BucketList::resolveAnyReadyFutures()
     }
 }
 
+bool
+BucketList::futuresAllResolved() const
+{
+    return std::all_of(
+        mLevels.begin(), mLevels.end(),
+        [](BucketLevel const& bl) { return !bl.getNext().isMerging(); });
+}
+
 void
 BucketList::addBatch(Application& app, uint32_t currLedger,
                      uint32_t currLedgerProtocol,
@@ -551,7 +564,8 @@ BucketList::addBatch(Application& app, uint32_t currLedger,
 }
 
 void
-BucketList::restartMerges(Application& app, uint32_t maxProtocolVersion)
+BucketList::restartMerges(Application& app, uint32_t maxProtocolVersion,
+                          uint32_t ledger)
 {
     for (uint32_t i = 0; i < static_cast<uint32>(mLevels.size()); i++)
     {
@@ -565,6 +579,57 @@ BucketList::restartMerges(Application& app, uint32_t maxProtocolVersion)
                 CLOG(INFO, "Bucket")
                     << "Restarted merge on BucketList level " << i;
             }
+        }
+        // The next block assumes we are re-starting a
+        // FIRST_PROTOCOL_SHADOWS_REMOVED or later merge, which has no shadows
+        // _and_ no stored inputs/outputs, and ensures that we are only
+        // re-starting buckets of correct version.
+        else if (next.isClear() && i > 0)
+        {
+            // Recover merge by iterating through bucketlist levels and
+            // using snaps and currs. The only time we don't use level's
+            // curr is when a level is about to snap; in that case, when the
+            // merge is needed, level's curr will snap, and merge will be
+            // promoted into curr. Therefore, when starting merge, we use an
+            // empty curr.
+            // Additionally, it is safe to recover a merge at any point
+            // before the merge is needed (meaning it should be promoted
+            // into level's curr after ledger close). This is due to the
+            // fact that future bucket inputs _do not change_ until level
+            // spill, and after such spills, new merges are started with new
+            // inputs.
+            auto snap = mLevels[i - 1].getSnap();
+            Hash const emptyHash;
+
+            // Exit early if a level has not been initialized yet;
+            // There are two possibilities for empty buckets: it is either truly
+            // untouched (meaning not enough ledgers were produced to populate
+            // given level) or it's a protocol 10-or-earlier bucket (since it
+            // does not contain a metadata entry). If we are dealing with
+            // 10-or-earlier bucket, it must have had an output published, and
+            // would be handled in the previous `if` block. Therefore, we must
+            // be dealing with an untouched level.
+            if (snap->getHash() == emptyHash)
+            {
+                return;
+            }
+
+            auto version = Bucket::getBucketVersion(snap);
+            if (version < Bucket::FIRST_PROTOCOL_SHADOWS_REMOVED)
+            {
+                auto msg =
+                    fmt::format("Invalid state: bucketlist level {} has clear "
+                                "future bucket but version {} snap",
+                                i, version);
+                throw std::runtime_error(msg);
+            }
+
+            // Round down the current ledger to when the merge was started, and
+            // re-start the merge via prepare, mimicking the logic in `addBatch`
+            auto mergeStartLedger = mask(ledger, BucketList::levelHalf(i - 1));
+            level.prepare(
+                app, mergeStartLedger, version, snap, /* shadows= */ {},
+                !app.getConfig().ARTIFICIALLY_REDUCE_MERGE_COUNTS_FOR_TESTING);
         }
     }
 }
--- a/src/bucket/FutureBucket.cpp
+++ b/src/bucket/FutureBucket.cpp
@@ -44,6 +44,15 @@ FutureBucket::FutureBucket(Application& app,
     assert(snap);
     mInputCurrBucketHash = binToHex(curr->getHash());
     mInputSnapBucketHash = binToHex(snap->getHash());
+    if (Bucket::getBucketVersion(snap) >=
+        Bucket::FIRST_PROTOCOL_SHADOWS_REMOVED)
+    {
+        if (!mInputShadowBuckets.empty())
+        {
+            throw std::runtime_error(
+                "Invalid FutureBucket: ledger version doesn't support shadows");
+        }
+    }
     for (auto const& b : mInputShadowBuckets)
     {
         mInputShadowBucketHashes.push_back(binToHex(b->getHash()));
--- a/src/bucket/FutureBucket.cpp
+++ b/src/bucket/FutureBucket.cpp
@@ -205,6 +205,12 @@ FutureBucket::hasHashes() const
 }
 
 bool
+FutureBucket::isClear() const
+{
+    return mState == FB_CLEAR;
+}
+
+bool
 FutureBucket::mergeComplete() const
 {
     assert(isLive());
--- a/src/bucket/test/BucketManagerTests.cpp
+++ b/src/bucket/test/BucketManagerTests.cpp
@@ -653,6 +653,12 @@ class StopAndRestartBucketMergesTest
                                  << mMergeCounters.mPreInitEntryProtocolMerges;
             CLOG(INFO, "Bucket") << "PostInitEntryProtocolMerges: "
                                  << mMergeCounters.mPostInitEntryProtocolMerges;
+            CLOG(INFO, "Bucket")
+                << "mPreShadowRemovalProtocolMerges: "
+                << mMergeCounters.mPreShadowRemovalProtocolMerges;
+            CLOG(INFO, "Bucket")
+                << "mPostShadowRemovalProtocolMerges: "
+                << mMergeCounters.mPostShadowRemovalProtocolMerges;
             CLOG(INFO, "Bucket") << "RunningMergeReattachments: "
                                  << mMergeCounters.mRunningMergeReattachments;
             CLOG(INFO, "Bucket") << "FinishedMergeReattachments: "
@@ -709,9 +715,17 @@ class StopAndRestartBucketMergesTest
         }
 
         void
-        checkSensiblePostInitEntryMergeCounters() const
+        checkSensiblePostInitEntryMergeCounters(uint32_t protocol) const
         {
             CHECK(mMergeCounters.mPostInitEntryProtocolMerges != 0);
+            if (protocol < Bucket::FIRST_PROTOCOL_SHADOWS_REMOVED)
+            {
+                CHECK(mMergeCounters.mPostShadowRemovalProtocolMerges == 0);
+            }
+            else
+            {
+                CHECK(mMergeCounters.mPostShadowRemovalProtocolMerges != 0);
+            }
 
             CHECK(mMergeCounters.mNewMetaEntries == 0);
             CHECK(mMergeCounters.mNewInitEntries != 0);
@@ -730,9 +744,18 @@ class StopAndRestartBucketMergesTest
             CHECK(mMergeCounters.mOldInitEntriesMergedWithNewDead != 0);
             CHECK(mMergeCounters.mNewEntriesMergedWithOldNeitherInit != 0);
 
-            CHECK(mMergeCounters.mShadowScanSteps != 0);
+            if (protocol < Bucket::FIRST_PROTOCOL_SHADOWS_REMOVED)
+            {
+                CHECK(mMergeCounters.mShadowScanSteps != 0);
+                CHECK(mMergeCounters.mLiveEntryShadowElisions != 0);
+            }
+            else
+            {
+                CHECK(mMergeCounters.mShadowScanSteps == 0);
+                CHECK(mMergeCounters.mLiveEntryShadowElisions == 0);
+            }
+
             CHECK(mMergeCounters.mMetaEntryShadowElisions == 0);
-            CHECK(mMergeCounters.mLiveEntryShadowElisions != 0);
             CHECK(mMergeCounters.mInitEntryShadowElisions == 0);
             CHECK(mMergeCounters.mDeadEntryShadowElisions == 0);
 
@@ -746,6 +769,7 @@ class StopAndRestartBucketMergesTest
         checkSensiblePreInitEntryMergeCounters() const
         {
             CHECK(mMergeCounters.mPreInitEntryProtocolMerges != 0);
+            CHECK(mMergeCounters.mPreShadowRemovalProtocolMerges != 0);
 
             CHECK(mMergeCounters.mNewMetaEntries == 0);
             CHECK(mMergeCounters.mNewInitEntries == 0);
@@ -783,6 +807,11 @@ class StopAndRestartBucketMergesTest
                   other.mMergeCounters.mPreInitEntryProtocolMerges);
             CHECK(mMergeCounters.mPostInitEntryProtocolMerges ==
                   other.mMergeCounters.mPostInitEntryProtocolMerges);
+
+            CHECK(mMergeCounters.mPreShadowRemovalProtocolMerges ==
+                  other.mMergeCounters.mPreShadowRemovalProtocolMerges);
+            CHECK(mMergeCounters.mPostShadowRemovalProtocolMerges ==
+                  other.mMergeCounters.mPostShadowRemovalProtocolMerges);
 
             CHECK(mMergeCounters.mRunningMergeReattachments ==
                   other.mMergeCounters.mRunningMergeReattachments);
@@ -1229,7 +1258,7 @@ class StopAndRestartBucketMergesTest
             mControlSurveys.rbegin()->second.dumpMergeCounters(
                 "control, Post-INITENTRY", mDesignatedLevel);
             mControlSurveys.rbegin()
-                ->second.checkSensiblePostInitEntryMergeCounters();
+                ->second.checkSensiblePostInitEntryMergeCounters(mProtocol);
         }
         else
         {
@@ -1248,7 +1277,8 @@ TEST_CASE("bucket persistence over app restart with initentry",
 {
     for (uint32_t protocol :
          {Bucket::FIRST_PROTOCOL_SUPPORTING_INITENTRY_AND_METAENTRY - 1,
-          Bucket::FIRST_PROTOCOL_SUPPORTING_INITENTRY_AND_METAENTRY})
+          Bucket::FIRST_PROTOCOL_SUPPORTING_INITENTRY_AND_METAENTRY,
+          Bucket::FIRST_PROTOCOL_SHADOWS_REMOVED})
     {
         for (uint32_t level : {2, 3})
         {
@@ -1264,7 +1294,8 @@ TEST_CASE("bucket persistence over app restart with initentry - extended",
 {
     for (uint32_t protocol :
          {Bucket::FIRST_PROTOCOL_SUPPORTING_INITENTRY_AND_METAENTRY - 1,
-          Bucket::FIRST_PROTOCOL_SUPPORTING_INITENTRY_AND_METAENTRY})
+          Bucket::FIRST_PROTOCOL_SUPPORTING_INITENTRY_AND_METAENTRY,
+          Bucket::FIRST_PROTOCOL_SHADOWS_REMOVED})
     {
         for (uint32_t level : {2, 3, 4, 5})
         {
--- a/src/bucket/test/BucketTests.cpp
+++ b/src/bucket/test/BucketTests.cpp
@@ -57,6 +57,15 @@ for_versions_with_differing_bucket_logic(
     Config const& cfg, std::function<void(Config const&)> const& f)
 {
     for_versions({Bucket::FIRST_PROTOCOL_SUPPORTING_INITENTRY_AND_METAENTRY - 1,
+                  Bucket::FIRST_PROTOCOL_SHADOWS_REMOVED},
+                 cfg, f);
+}
+
+void
+for_versions_with_differing_initentry_logic(
+    Config const& cfg, std::function<void(Config const&)> const& f)
+{
+    for_versions({Bucket::FIRST_PROTOCOL_SUPPORTING_INITENTRY_AND_METAENTRY - 1,
                   Bucket::FIRST_PROTOCOL_SUPPORTING_INITENTRY_AND_METAENTRY},
                  cfg, f);
 }
@@ -319,6 +328,73 @@ generateDifferentAccount(std::vector<LedgerEntry> const& others)
     }
 }
 
+TEST_CASE("merges proceed old-style despite newer shadows",
+          "[bucket][bucketmaxprotocol]")
+{
+    VirtualClock clock;
+    Config const& cfg = getTestConfig();
+    Application::pointer app = createTestApplication(clock, cfg);
+    auto& bm = app->getBucketManager();
+    auto v12 = Bucket::FIRST_PROTOCOL_SHADOWS_REMOVED;
+    auto v11 = v12 - 1;
+    auto v10 = v11 - 1;
+
+    LedgerEntry liveEntry = generateAccount();
+    LedgerEntry otherLiveA = generateDifferentAccount({liveEntry});
+
+    auto b10first = Bucket::fresh(bm, v10, {}, {liveEntry}, {},
+                                  /*countMergeEvents=*/true,
+                                  /*doFsync=*/true);
+    auto b10second = Bucket::fresh(bm, v10, {}, {otherLiveA}, {},
+                                   /*countMergeEvents=*/true,
+                                   /*doFsync=*/true);
+
+    auto b11first = Bucket::fresh(bm, v11, {}, {liveEntry}, {},
+                                  /*countMergeEvents=*/true,
+                                  /*doFsync=*/true);
+    auto b11second = Bucket::fresh(bm, v11, {}, {otherLiveA}, {},
+                                   /*countMergeEvents=*/true,
+                                   /*doFsync=*/true);
+
+    auto b12first =
+        Bucket::fresh(bm, v12, {}, {liveEntry}, {}, /*countMergeEvents=*/true,
+                      /*doFsync=*/true);
+    auto b12second = Bucket::fresh(bm, v12, {}, {otherLiveA}, {},
+                                   /*countMergeEvents=*/true,
+                                   /*doFsync=*/true);
+
+    SECTION("shadow version 12")
+    {
+        // With proto 12, new bucket version solely depends on the snap version
+        auto bucket = Bucket::merge(bm, v12, b11first, b11second,
+                                    /*shadows=*/{b12first},
+                                    /*keepDeadEntries=*/true,
+                                    /*countMergeEvents=*/true,
+                                    /*doFsync=*/true);
+        REQUIRE(Bucket::getBucketVersion(bucket) == v11);
+    }
+    SECTION("shadow versions mixed, pick lower")
+    {
+        // Merging older version (10) buckets, with mixed versions of shadows
+        // (11, 12) Pick initentry (11) style merge
+        auto bucket = Bucket::merge(bm, v12, b10first, b10second,
+                                    /*shadows=*/{b12first, b11second},
+                                    /*keepDeadEntries=*/true,
+                                    /*countMergeEvents=*/true,
+                                    /*doFsync=*/true);
+        REQUIRE(Bucket::getBucketVersion(bucket) == v11);
+    }
+    SECTION("refuse to merge new version with shadow")
+    {
+        REQUIRE_THROWS_AS(Bucket::merge(bm, v12, b12first, b12second,
+                                        /*shadows=*/{b12first},
+                                        /*keepDeadEntries=*/true,
+                                        /*countMergeEvents=*/true,
+                                        /*doFsync=*/true),
+                          std::runtime_error);
+    }
+}
+
 TEST_CASE("merges refuse to exceed max protocol version",
           "[bucket][bucketmaxprotocol]")
 {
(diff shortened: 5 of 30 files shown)


src/catchup/ApplyBucketsWork.cpp
```diff
@@ -145,6 +149,35 @@ ApplyBucketsWork::startLevel()
...
+            auto handler = [weak](asio::error_code const& ec) {
+                auto self = weak.lock();
+                if (self)
+                {
+                    self->wakeUp();
+                }
+            };
+
+            // Check back later
+            mDelayTimer->expires_from_now(std::chrono::milliseconds(500));
```
graydon: This seems like it'll spin twice a second for .. potentially as long as the merge runs? Current merges are short-ish but they can potentially run for minutes, hours..

src/catchup/ApplyBucketsWork.cpp
```diff
@@ -193,8 +226,22 @@ ApplyBucketsWork::onRun()
...
-    CLOG(DEBUG, "History") << "ApplyBuckets : done, restarting merges";
-    mApp.getBucketManager().assumeState(mApplyState, mMaxProtocolVersion);
+    CLOG(INFO, "History") << "ApplyBuckets : done, restarting merges";
+    // Use old-style merges from HAS
+    bool useHASFutureBuckets =
+        mMaxProtocolVersion < Bucket::FIRST_PROTOCOL_SHADOWS_REMOVED;
+    mApp.getBucketManager().assumeState(mApplyState, mMaxProtocolVersion,
+                                        useHASFutureBuckets);
+
+    // To avoid tripping off the network during online catchup post protocol 12,
```
graydon: Not entirely clear I understand this comment. Could you expand it a bit? What does "tripping off the network" mean here, and what is a promotion ledger?

src/bucket/FutureBucket.cpp
```diff
@@ -35,6 +35,7 @@ FutureBucket::FutureBucket(Application& app,
     , mInputCurrBucket(curr)
     , mInputSnapBucket(snap)
     , mInputShadowBuckets(shadows)
+    , mProtocolVersion(maxProtocolVersion)
```
graydon: Just double checking: is mProtocolVersion stored in the FutureBucket just to allow checking assertions?

src/bucket/BucketList.cpp
```diff
@@ -550,20 +573,67 @@ BucketList::addBatch(Application& app, uint32_t currLedger,
...
+            // shared state without blocking. Upon finishing, packaged_task (for
+            // now irrelevant merge) will mark shared state as ready, and will
+            // then release its reference to shared state.
+
+            next.clear();
+            if (i > 0)
             {
-                CLOG(INFO, "Bucket")
-                    << "Restarted merge on BucketList level " << i;
+                // Before the last spill
```
graydon: I know we've been over this and you convinced me before, but I'm forgetful and nervous and would love to see spelled-out in a comment here why merging an empty snap and a sometimes-empty curr is always right here. Or in general just "why this code is a correct restart of a merge".

src/catchup/ApplyBucketsWork.cpp
marta-lokhova: When buckets are applied, we won't have ready outputs before joining the network. This means that if we mark catchup as "successful", and start closing ledgers, we might block for a while, waiting for a merge to be resolved, so that the future bucket can be promoted into the bucketlist at that ledger. If resolution takes long enough, a node might lose sync. 

I'll reword this comment to avoid confusion.

src/bucket/FutureBucket.cpp
marta-lokhova: yeah, sanity check.

src/catchup/ApplyBucketsWork.cpp
marta-lokhova: The longest ones we've seen after shadow removal were under a minute. With the current state of the system, something like 0.5-1 s seems reasonable to me, although it can be tweaked in the future. What do you think?

src/historywork/ResolveSnapshotWork.cpp
```diff
@@ -16,11 +16,21 @@ ResolveSnapshotWork::ResolveSnapshotWork(
...
+    if (!mSnapshot)
+    {
+        throw std::runtime_error("ResolveSnapshotWork: invalid snapshot");
+    }
 }
 
 BasicWork::State
 ResolveSnapshotWork::onRun()
 {
+    if (mSnapshot->mLocalState.futuresAllClear())
```
MonsieurNicolas: why do we need to special case the "all clear" case?

src/herder/Upgrades.cpp
```diff
@@ -748,6 +752,13 @@ Upgrades::applyVersionUpgrade(AbstractLedgerTxn& ltx, uint32_t newVersion)
     {
         prepareLiabilities(ltx, header);
     }
+
+    if (header.current().ledgerVersion >= 12 && prevVersion < 12)
```
MonsieurNicolas: be consistent: use `Bucket::FIRST_PROTOCOL_SHADOWS_REMOVED`

src/history/HistoryArchive.cpp
```diff
@@ -289,6 +290,12 @@ HistoryArchiveState::HistoryArchiveState(uint32_t ledgerSeq,
         auto& level = buckets.getLevel(i);
         b.curr = binToHex(level.getCurr()->getHash());
         b.next = level.getNext();
+        if (clearFutureBuckets)
+        {
+            // Future buckets aren't needed in the archives
+            // post protocol 12.
```
MonsieurNicolas: use `Bucket::FIRST_PROTOCOL_SHADOWS_REMOVED` not 12

src/ledger/LedgerManager.h
```diff
@@ -106,7 +106,7 @@ class LedgerManager
 
     // return the HAS that corresponds to the last closed ledger as persisted in
     // the database
-    virtual HistoryArchiveState getLastClosedLedgerHAS() = 0;
+    virtual HistoryArchiveState getPersistentHAS() = 0;
```
MonsieurNicolas: why the rename?

marta-lokhova: there are two functions with very similar names, which are easy to confuse: `HistoryManager::getLastClosedHistoryArchiveState` and `LedgerManager::getLastClosedLedgerHAS`. The former returns HAS from an in-memory LCL, the latter queries the database for whatever last HAS was saved.

MonsieurNicolas: `LastClosedLedger` is used all over the code, calling it `persistent` makes it less clear on what it is

MonsieurNicolas: Oh I see now what you're saying. We should remove the method in `HistoryManager` instead

src/history/HistoryArchive.cpp
```diff
@@ -67,6 +67,19 @@ HistoryArchiveState::futuresAllResolved() const
     return true;
 }
 
+bool
+HistoryArchiveState::futuresAllClear() const
+{
+    for (auto const& level : currentBuckets)
```
vogel: Could be replaced by std::all_of

src/bucket/BucketList.cpp
```diff
@@ -421,6 +421,19 @@ BucketList::resolveAnyReadyFutures()
     }
 }
 
+bool
+BucketList::futuresAllResolved() const
+{
+    for (auto const& level : mLevels)
```
vogel: Could be replaced by std::all_of

src/catchup/ApplyBucketsWork.h
```diff
@@ -46,6 +46,11 @@ class ApplyBucketsWork : public BasicWork
     medida::Meter& mBucketApplyFailure;
     BucketApplicator::Counters mCounters;
 
+    // In case of an online catchup, let ApplyBucketsWork buffer ledgers
+    // which it waits for merges to resolve.
```
vogel: typo which -> while

src/history/HistoryArchive.cpp
marta-lokhova: hmm, it's quite verbose that way. "Protocol 10" and "Protocol 11" are used consistently in the comments/tests throughout the codebase, I don't see how this is any different.

src/historywork/ResolveSnapshotWork.cpp
marta-lokhova: Looks like `futuresAllResolved` already accounts for "all clear", so I think we don't need it. Removed.

src/catchup/ApplyBucketsWork.cpp
```diff
@@ -192,8 +225,19 @@ ApplyBucketsWork::onRun()
         return State::WORK_RUNNING;
     }
 
-    CLOG(DEBUG, "History") << "ApplyBuckets : done, restarting merges";
-    mApp.getBucketManager().assumeState(mApplyState, mMaxProtocolVersion);
+    CLOG(INFO, "History") << "ApplyBuckets : done, restarting merges";
+    // Use old-style merges from HAS
+    bool useHASFutureBuckets =
```
graydon: `useHASFutureBuckets` gets calculated by examining `mMaxProtocolVersion` here, but it's then passed to `BucketManager::assumeState` and on to `BucketList::restartMerges`. In that function, it's used as one half of an OR-condition, the other half of which (`removeShadows`) comes from `spillLevelWithoutShadows`, which is itself also derived by examining `mMaxProtocolVersion`. So this _seems_ redundant to me: the condition decided by `useHASFutureBuckets` is being partially re-decided in `BucketList::spillLevelWithoutShadows` and both decisions are OR'ed together and only used in that one condition in `BucketList::restartMerges`. I think one of the two calculations should go, and I _think_ it's this flag? Not sure though.

src/history/HistoryArchive.cpp
```diff
@@ -275,17 +276,35 @@ HistoryArchiveState::HistoryArchiveState() : server(STELLAR_CORE_VERSION)
...
         b.curr = binToHex(level.getCurr()->getHash());
-        b.next = level.getNext();
         b.snap = binToHex(level.getSnap()->getHash());
+        b.next = level.getNext();
+
+        // Level 0 merge always happens in-place on close ledger, so next always
+        // remains clear in between `addBatch` calls
+        if (clearFutureBuckets && prevSnap)
+        {
+            BucketInputIterator it(prevSnap);
```
graydon: You're doing this "make an iterator, sniff its version" thing in a couple places, it might be worth factoring out into a helper method on a bucket.

(Part of me also thinks maybe it should be cached, but buckets are accessed from multiple threads, and .. meh, it's probably not especially hot)

src/ledger/LedgerManagerImpl.cpp
```diff
@@ -1132,8 +1136,11 @@ LedgerManagerImpl::storeCurrentLedger(LedgerHeader const& header)
 
     // Store the current HAS in the database; this is really just to checkpoint
     // the bucketlist so we can survive a restart and re-attach to the buckets.
+
+    // Note: we do not ignore future buckets here, since we might have some
+    // merges completed which we do not want to throw away.
     HistoryArchiveState has(header.ledgerSeq,
-                            mApp.getBucketManager().getBucketList());
+                            mApp.getBucketManager().getBucketList(), false);
```
graydon: worth noting with an inline comment what `false` means here.

src/main/ApplicationUtils.cpp
```diff
@@ -235,7 +235,7 @@ rebuildLedgerFromBuckets(Config cfg)
     has.fromString(hasStr);
 
     auto applyBucketsWork = ws.executeWork<ApplyBucketsWork>(
-        localBuckets, has, Config::CURRENT_LEDGER_PROTOCOL_VERSION);
+        localBuckets, has, Config::CURRENT_LEDGER_PROTOCOL_VERSION, false);
```
graydon: also worth noting what `false` means here.

src/bucket/BucketList.cpp
```diff
@@ -551,22 +571,91 @@ BucketList::addBatch(Application& app, uint32_t currLedger,
...
+    std::shared_ptr<Bucket> snap;
     for (uint32_t i = 0; i < static_cast<uint32>(mLevels.size()); i++)
     {
         auto& level = mLevels[i];
         auto& next = level.getNext();
-        if (next.hasHashes() && !next.isLive())
+
+        // Determine whether to use future bucket from HAS, and restart merge
+        // with curr and snap for this level
+        auto removeShadows = spillLevelWithoutShadows(maxProtocolVersion, i);
```
graydon: This reads a little tricky to me, I think there are 3 things here that I'm a little nervous about:

  1. `spillLevelWithoutShadows` would read better as `spillToLevelWithoutShadows`, because we're passing `i` and it's not clear if this is to or from level `i`. I think we're discussing a spill _from_ `i-1` _to_ `i`, right? It's easy for the reader to think the opposite and be off by one, read this as a spill _from_ level `i` rather than _to_ level `i`
  2. The function is phrased as a negative condition (`Without`) and then that is tested negated in the next `if` statement, which makes the control condition a little more confusing than necessary. Could you maybe change it to `bool needsShadows = spillToLevelNeedsShadows(...)` and then test it un-negated? (And of course change the helper function itself to return the opposite sense of its current behaviour).
  3. More substantially, you're passing `maxProtocolVersion` here, but treating it as `currProtocolVersion` in the helper function. Those aren't exactly synonymous concepts and I'm not 100% sure which one we actually have on hand here (and/or whether we need to have been passed _both_). See larger comment on `maxProtocolVersion` elsewhere.

src/bucket/BucketList.cpp
```diff
@@ -551,22 +571,91 @@ BucketList::addBatch(Application& app, uint32_t currLedger,
...
+                // promoted into curr. Therefore, when starting merge, we use an
+                // empty curr.
+                // Additionally, it is safe to recover a merge at any point
+                // before the merge is needed (meaning it should be promoted
+                // into level's curr after ledger close). This is due to the
+                // fact that future bucket inputs _do not change_ until level
+                // spill, and after such spills, new merges are started with new
+                // inputs.
+                auto lastMergeLedger = ledgerSeq /
+                                       BucketList::levelHalf(i - 1) *
```
graydon: Extract subexpression `BucketList::levelHalf(i - 1)` into a local variable, to improve legibility of the arithmetic.

src/bucket/BucketList.cpp
```diff
@@ -551,22 +571,91 @@ BucketList::addBatch(Application& app, uint32_t currLedger,
...
+                                       BucketList::levelHalf(i - 1);
+                uint32_t nextSpillLedger =
+                    lastMergeLedger + BucketList::levelHalf(i - 1);
+                auto shouldClearCurr =
+                    BucketList::levelShouldSpill(nextSpillLedger, i);
+                auto newCurr = shouldClearCurr ? std::make_shared<Bucket>()
```
graydon: `newCurr` can just be `curr` here, there's no other `curr` in this function.

src/bucket/BucketList.cpp
```diff
@@ -551,22 +571,91 @@ BucketList::addBatch(Application& app, uint32_t currLedger,
...
+                                               : level.getCurr();
+                bool countMergeEvents =
+                    !app.getConfig()
+                         .ARTIFICIALLY_REDUCE_MERGE_COUNTS_FOR_TESTING;
+                auto fb = FutureBucket(app, newCurr, snap,
```
graydon: I still feel like carrying `snap` forward from one iteration of this loop to the next is a bit too non-obvious -- at least it's surprised me both times I've reviewed this loop! -- and I think it's only used in this expression here, right? Would it be possible to put the expression `mLevels[i-1].getSnap()` here, and remove the loop-carried variable `snap`?

src/bucket/BucketList.cpp
```diff
@@ -551,22 +571,91 @@ BucketList::addBatch(Application& app, uint32_t currLedger,
...
+        return false;
     }
+
+    if (level == 0)
+    {
+        return true;
+    }
+
+    auto snap = mLevels[level - 1].getSnap();
+    BucketInputIterator snapIt(snap);
```
graydon: This is the other instance of that bucket-version-sniffing idiom.

src/catchup/ApplyBucketsWork.cpp
graydon: Great, clearer now, thanks!

src/bucket/FutureBucket.cpp
graydon: Cool, thanks.

src/catchup/ApplyBucketsWork.cpp
graydon: *Shrug* ok. As you say, easy to change later.

src/catchup/ApplyBucketsWork.h
```diff
@@ -46,6 +46,27 @@ class ApplyBucketsWork : public BasicWork
     medida::Meter& mBucketApplyFailure;
     BucketApplicator::Counters mCounters;
 
+    // With protocol 12 or higher, when buckets are applied, we do not have
```
MonsieurNicolas: There is something weird here that I think is not consistent with the behavior that you're describing:
if we're catching up but not applying buckets, we should probably wait for merges to be completed before we're claiming to be "in sync".

So... what I suspect is that we need to wait for merges to be completed in more places:
1. when running `ApplyLedgerChainWork` - maybe even between ledgers?
2. after applying each ledger when replaying buffered ledgers

This way we will guarantee that we're not blocking the main thread because of pending merges.

MonsieurNicolas: NB: maybe this whole "wait to be in sync" functionality should be in its own commit as it seems non trivial (and independent from the other changes)

src/catchup/CatchupWork.cpp
```diff
@@ -106,12 +107,27 @@ CatchupWork::alreadyHaveBucketsHistoryArchiveState(uint32_t atCheckpoint) const
 WorkSeqPtr
 CatchupWork::downloadApplyBuckets()
 {
+    // If first ledger we're going to replay is of older version (before shadow
```
MonsieurNicolas: (related to the other comment: we should probably move this out of the "apply buckets" method)

src/catchup/CatchupWork.cpp
```diff
@@ -106,12 +107,27 @@ CatchupWork::alreadyHaveBucketsHistoryArchiveState(uint32_t atCheckpoint) const
...
+    // removal), it'll perform old-style merges
+    auto ledgerVersion = mVerifiedLedgerRangeStart.header.ledgerVersion;
```
MonsieurNicolas: why not just always wait for merges regardless of protocol version? The danger of blocking the main thread exists also in the current version of the protocol, right?

src/history/HistoryArchive.cpp
```diff
@@ -275,17 +276,35 @@ HistoryArchiveState::HistoryArchiveState() : server(STELLAR_CORE_VERSION)
 }
 
 HistoryArchiveState::HistoryArchiveState(uint32_t ledgerSeq,
-                                         BucketList const& buckets)
+                                         BucketList const& buckets,
+                                         bool clearFutureBuckets)
```
MonsieurNicolas: I don't like this `clearFutureBuckets` thingy: why do we set shadows in the first place?

src/history/HistoryManagerImpl.cpp
```diff
@@ -251,7 +242,18 @@ HistoryManagerImpl::maybeQueueHistoryCheckpoint()
 void
 HistoryManagerImpl::queueCurrentHistory()
 {
-    auto has = getLastClosedHistoryArchiveState();
+    // Subtle: depending on the protocol version,
```
MonsieurNicolas: related to my previous comment: I don't see why we need this here.
Basically: to queue things up we should never have to wait.

Later (actual publish), we can tell if we need to wait in `ResolveSnapshotWork` directly by for example seeing that we have shadows or not.

src/main/Config.cpp
```diff
@@ -25,7 +25,7 @@
 
 namespace stellar
 {
-const uint32 Config::CURRENT_LEDGER_PROTOCOL_VERSION = 11;
+const uint32 Config::CURRENT_LEDGER_PROTOCOL_VERSION = 12;
```
MonsieurNicolas: separate this change in a different commit (update current protocol version)

src/catchup/ApplyBucketsWork.cpp
```diff
@@ -25,7 +25,8 @@ namespace stellar
 ApplyBucketsWork::ApplyBucketsWork(
     Application& app,
     std::map<std::string, std::shared_ptr<Bucket>> const& buckets,
-    HistoryArchiveState const& applyState, uint32_t maxProtocolVersion)
+    HistoryArchiveState const& applyState, uint32_t maxProtocolVersion,
```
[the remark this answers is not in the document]
marta-lokhova: Great point. I did a check-up on the diff, and tightened the protocol versions in two places so that new-style merge changes are independent of maxProtocolVersion parameters:
* `Bucket::merge` now uses `protocolVersion` to update metrics, and perform some proto 12 sanity checks
* `BucketList::restartMerges` uses snap's protocol version to create new FutureBucket, so new-style merge restart is independent of the `maxProtocolVersion` parameter.

I also changed FutureBucket to take `protocolVersion` in its constructor, since in both usages in the code we know the exact version to use (specifically, in `BucketList::prepare` and `BucketList::restartMerges`). The remaining usages of `maxProtocolVersion` are still there, as introduced by protocol 11.

marta-lokhova:
Still got a version-related question/concern, to discuss offline

src/bucket/Bucket.cpp
```diff
@@ -356,10 +356,18 @@ calculateMergeProtocolVersion(
...
+    // protocol version is determined as a max of curr, snap, and any shadow of
+    // version < FIRST_PROTOCOL_SHADOWS_REMOVED. This means that a bucket may
+    // still perform an old style merge despite the presence of the new protocol
+    // shadows.
     for (auto const& si : shadowIterators)
     {
-        protocolVersion =
-            std::max(si.getMetadata().ledgerVersion, protocolVersion);
+        auto version = si.getMetadata().ledgerVersion;
+        if (version < Bucket::FIRST_PROTOCOL_SHADOWS_REMOVED)
```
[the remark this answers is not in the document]
marta-lokhova: we need the loop first to derive the protocol version to act on

src/bucket/BucketList.cpp
```diff
@@ -579,6 +579,10 @@ BucketList::restartMerges(Application& app, uint32_t maxProtocolVersion,
                     << "Restarted merge on BucketList level " << i;
             }
         }
+        // The next block assumes we are re-starting a
+        // FIRST_PROTOCOL_SHADOWS_REMOVED or later merge, which has no shadows
+        // _and_ no stored inputs, and ensures that we are only re-starting
```
graydon: s/no stored inputs/no stored outputs/

src/bucket/BucketList.cpp
```diff
@@ -594,9 +598,16 @@ BucketList::restartMerges(Application& app, uint32_t maxProtocolVersion,
             // spill, and after such spills, new merges are started with new
             // inputs.
             auto snap = mLevels[i - 1].getSnap();
-            Hash emptyHash;
-
-            // Level has not been touched yet, nothing to do
+            Hash const emptyHash;
+
+            // Exit early if a level has not been initialized yet;
```
graydon: Fantastic comment! Entirely reassuring now, thanks :)

src/bucket/Bucket.cpp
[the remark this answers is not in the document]
MonsieurNicolas: argh yeah misread the code :)

src/history/HistoryArchive.cpp
```diff
@@ -259,6 +262,83 @@ HistoryArchiveState::allBuckets() const
...
+        if (snap->getHash() == emptyHash)
+        {
+            continue;
+        }
+        else if (Bucket::getBucketVersion(snap) >=
+                 Bucket::FIRST_PROTOCOL_SHADOWS_REMOVED)
+        {
+            if (!level.next.isClear())
+            {
+                CLOG(INFO, "History") << "Invalid HAS: future must be cleared ";
```
graydon: `ERROR` log level (and possibly throw?)

src/history/HistoryArchive.cpp
```diff
@@ -259,6 +262,83 @@ HistoryArchiveState::allBuckets() const
...
+                return false;
+            }
+        }
+        else if (!level.next.hasOutputHash())
+        {
+            CLOG(INFO, "History")
```
graydon: `ERROR` log level (and possibly throw?)

src/catchup/CatchupWork.cpp
```diff
@@ -110,8 +111,17 @@ CatchupWork::downloadApplyBuckets()
     std::vector<std::string> hashes = has.differingBuckets(mLocalState);
     auto getBuckets = std::make_shared<DownloadBucketsWork>(
         mApp, mBuckets, hashes, *mDownloadDir);
+
+    // A consequence of FIRST_PROTOCOL_SHADOWS_REMOVED upgrade, inputs or
+    // outputs aren't published to the archives anymore: new-style merges are
+    // re-started from scratch To avoid going out of sync during online catchup,
```
graydon: Missing period between "scratch" and "To".

Also on the next line: "application work wait" should be "application work to wait".

src/bucket/BucketList.cpp
```diff
@@ -566,6 +579,53 @@ BucketList::restartMerges(Application& app, uint32_t maxProtocolVersion)
                     << "Restarted merge on BucketList level " << i;
             }
         }
+        // The next block assumes we are re-starting a
+        // FIRST_PROTOCOL_SHADOWS_REMOVED or later merge, which has no shadows
+        // _and_ no stored inputs, and ensures that we are only re-starting
```
graydon: s/no stored inputs/no stored outputs/

src/bucket/BucketList.cpp
```diff
@@ -566,6 +579,53 @@ BucketList::restartMerges(Application& app, uint32_t maxProtocolVersion)
...
+            // be dealing with an untouched level.
+            if (snap->getHash() == emptyHash)
+            {
+                return;
+            }
+
+            auto version = Bucket::getBucketVersion(snap);
+            if (version < Bucket::FIRST_PROTOCOL_SHADOWS_REMOVED)
+            {
+                throw std::runtime_error("Invalid BucketList state!");
```
graydon: Add detail to message: `invalid state: bucketlist level %d has clear "next" but version %d "snap"`

src/bucket/BucketList.cpp
```diff
@@ -566,6 +579,53 @@ BucketList::restartMerges(Application& app, uint32_t maxProtocolVersion)
...
+            }
+
+            // Round down the current ledger to when the merge was started, and
+            // re-start the merge via prepare, mimicking the logic in `addBatch`
+            auto mergeStartLedger = mask(ledger, BucketList::levelHalf(i - 1));
+            level.prepare(
+                app, mergeStartLedger, version, snap, {},
```
graydon: add `/* shadows= */` comment to `{}` in arg list.

src/catchup/ApplyBucketsWork.cpp
```diff
@@ -192,8 +232,15 @@ ApplyBucketsWork::onRun()
         return State::WORK_RUNNING;
     }
 
-    CLOG(DEBUG, "History") << "ApplyBuckets : done, restarting merges";
+    CLOG(INFO, "History") << "ApplyBuckets : done, restarting merges";
     mApp.getBucketManager().assumeState(mApplyState, mMaxProtocolVersion);
+
+    if (mResolveMerges)
```
graydon: This logic feels like it's in a strange order to me: the _first_ block checks for `mResolveMerges && mDelayTimer` and if so handles the resolving merge, then the _second_ block checks `mResolveMerges` and if so sets up `mDelayTimer`. This feels like the blocks are kinda backwards and mis-nested, perhaps? Would logic like this work? I think it's the same, but reads top-to-bottom in the order of return-visits-to-the-function.

```
if (mResolveMerges) {
    if (!mDelayTimer) {
        // This is our first time here, start the work
        bm.assumeState(...);
        mDelayTimer = make_unique<VirtualTimer>();
        return WORK_RUNNING;
    } else {
       // We're returning to running work, check to see if it's done
       bl.resolveAnyReadyFutures();
       if (!bl.futuresAllResolved()) {
           // Not yet done, queue a wait and callback
           mDelayTimer = ...
           mDelayTimer->expires_from_now(...)
           return WORK_WAITING;
        }
    }
}
// all done
return WORK_SUCCESS;

```

graydon: Update: I misunderstood that there's a big block of code between timer start and continue that github elided in the review view, this can't be readily reordered.

src/history/HistoryArchive.cpp
marta-lokhova: agreed that it should be `ERROR`, but i don't think throwing is right here: the node itself is not in a corrupted state, but the archive it downloaded from is.

graydon: 👍

src/bucket/test/BucketTests.cpp
```diff
@@ -319,6 +328,71 @@ generateDifferentAccount(std::vector<LedgerEntry> const& others)
...
 }
 
+TEST_CASE("merges proceed old-style despite newer shadows",
+          "[bucket][bucketmaxprotocol]")
+{
+    VirtualClock clock;
+    Config const& cfg = getTestConfig();
+    Application::pointer app = createTestApplication(clock, cfg);
+    auto& bm = app->getBucketManager();
+    auto vers = getAppLedgerVersion(app);
```
graydon: This looks weird: won't the test stop working in version 13? I think vers should be initialized to `Bucket::FIRST_PROTOCOL_SHADOWS_REMOVED`

src/bucket/test/BucketTests.cpp
```diff
@@ -319,6 +328,71 @@ generateDifferentAccount(std::vector<LedgerEntry> const& others)
...
+    LedgerEntry liveEntry = generateAccount();
+    LedgerEntry otherLiveA = generateDifferentAccount({liveEntry});
+
+    // Proto 10 bucket
+    auto b10first = Bucket::fresh(bm, vers - 2, {}, {liveEntry}, {},
```
graydon: also these `vers - 2` things seem to me .. not more legible than `10`, though I gather this was in response to an earlier complaint about literal `10` being written there. Consider this my disagreement with the other reviewer, at least in this one test (which is very specifically testing behaviours we expect for each of those versions).

Maybe split the difference and initialize some constants like `const v12 = FIRST_PROTOCOL_SHADOWS_REMOVED`, `const v11 = v12 - 1`, `const v10 = v11 - 1` and use the `v10` .. `v12` variables to indicate that they're versions? Bleh. I feel like such a nit-picker even mentioning this. But I'm having to remind me self on each stanza which version each expression denotes.

src/bucket/test/BucketManagerTests.cpp
```diff
@@ -508,86 +508,6 @@ TEST_CASE("bucketmanager reattach to running merge", "[bucket][bucketmanager]")
     });
 }
 
-TEST_CASE("bucketmanager reattach HAS from publish queue to finished merge",
```
graydon: Unclear why this test is being entirely eliminated; shouldn't it still be a thing we want to test when running in a pre-protocol-12 version?

src/history/test/HistoryTests.cpp
```diff
@@ -31,6 +31,32 @@
...
+numLevelsSpills(uint32_t upgradeLedger, uint32_t endLedger)
+{
+    uint32_t levels = 0;
+    uint32_t multiplier = 2;
+
+    // Given an upgrade ledger and end ledger to apply,
+    // calculate number of expected new-style merges. Note that any old-style
+    // merges in-progress complete.
+    int diff;
+    while ((diff = (endLedger / multiplier) -
```
graydon: some explanation of this arithmetic would be good; like at least what `diff` means and why `diff>0` controls the loop and why the loop is counting instances of `diff>1`.

src/history/test/HistoryTests.cpp
```diff
@@ -320,6 +346,100 @@ TEST_CASE("History publish to multiple archives", "[history]")
...
+
+        uint32_t upgradeLedger = 64;
+        catchupSimulation.setProto12UpgradeLedger(upgradeLedger);
+        auto checkpointLedger = catchupSimulation.getLastCheckpointLedger(2);
+        catchupSimulation.ensureOfflineCatchupPossible(checkpointLedger);
+        auto levelsCleared = numLevelsSpills(upgradeLedger, checkpointLedger);
+        checkClearFutureBuckets(catchupSimulation.getApp(), checkpointLedger,
+                                levelsCleared, 3);
+    }
+    SECTION("new-style merges begin mid-checkpoint")
```
graydon: Is the only difference between this section and the previous section `64` vs. `80`? If so, I'm not really sure how that corresponds to the semantic distinction between the phrases "in between checkpoints" and "mid-checkpoint". Those two phrases both seem to me like they mean the same thing, and if they don't, I don't know how they relate to `64` or `80`.

(And: if all 3 sections are identical except for the ledger number, maybe make it a loop?)

src/history/test/HistoryTests.cpp
```diff
@@ -320,6 +346,100 @@ TEST_CASE("History publish to multiple archives", "[history]")
     REQUIRE(catchupSimulation.catchupOffline(catchupApp, checkpointLedger));
 }
 
+TEST_CASE("Publish works correctly post shadow removal", "[history]")
```
graydon: In general this test and the next one look like they're testing some pretty subtle conditions and it's difficult for me to extract what you're aiming to check or how. A little commentary would help here about the overall test plan and what is essentially under investigation vs. merely incidental to the test.

(They look like good tests! I just .. don't quite understand what they're all about)

src/history/test/HistoryTests.cpp
```diff
@@ -512,6 +632,80 @@ TEST_CASE("History prefix catchup", "[history][catchup]")
...
+                                        configurator};
+
+    catchupSimulation.generateRandomLedger(oldProto);
+
+    // Different counts: with proto 12, catchup should adapt and switch merge
+    // logic
+    std::vector<uint32_t> counts = {0, std::numeric_limits<uint32_t>::max(),
+                                    60};
+
+    SECTION("Upgrade at checkpoint start")
```
graydon: Again here I _think_ all 3 of these are checking the same condition(s) just with different ledgers -- so could probably just be a loop + some commentary explaining what's special about those 3 numbers -- but if they differ in other ways I'm not seeing it and would like to know!

src/herder/test/UpgradesTests.cpp
```diff
@@ -1524,6 +1524,112 @@ TEST_CASE("upgrade to version 11", "[upgrades]")
...
+            auto lev0Snap = lev0.getSnap();
+            auto lev1Curr = lev1.getCurr();
+            auto lev1Snap = lev1.getSnap();
+            auto getVers = [](std::shared_ptr<Bucket> b) -> uint32_t {
+                return BucketInputIterator(b).getMetadata().ledgerVersion;
+            };
+            // post-shadow-removal merges
+            // We can't be quite sure how many in-progress merges actually
+            // finished (due to background thread races), so we check that the
+            // number of merges in between max possible (e.g., all kicked off
```
graydon: This seems a little odd: can't we wait here until the merges complete, before continuing?

src/history/test/HistoryTests.cpp
graydon: Ah great, I can follow it much more directly now. The commentary makes it clear how the cases differ and why you're testing them, thanks!

src/history/HistoryArchive.cpp
graydon: (Rafal isn't reviewing this anymore, but this is fixed.)

src/catchup/ApplyBucketsWork.cpp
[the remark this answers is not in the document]
graydon: Looks like this is fixed, thanks.

src/ledger/LedgerManagerImpl.cpp
graydon: obsolete comment

src/history/HistoryArchive.cpp
graydon: fixed

src/bucket/BucketList.cpp
graydon: fixed

src/bucket/BucketList.cpp
graydon: obsolete

src/bucket/BucketList.cpp
graydon: fixed

src/bucket/BucketList.cpp
graydon: fixed

src/bucket/BucketList.cpp
```diff
@@ -566,6 +579,42 @@ BucketList::restartMerges(Application& app, uint32_t maxProtocolVersion)
...
+            // Additionally, it is safe to recover a merge at any point
+            // before the merge is needed (meaning it should be promoted
+            // into level's curr after ledger close). This is due to the
+            // fact that future bucket inputs _do not change_ until level
+            // spill, and after such spills, new merges are started with new
+            // inputs.
+            auto snap = mLevels[i - 1].getSnap();
+            Hash emptyHash;
+
+            // Level has not been touched yet, nothing to do
```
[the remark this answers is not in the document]
graydon: fixed

src/invariant/test/BucketListIsConsistentWithDatabaseTests.cpp
```diff
@@ -59,9 +59,10 @@ struct BucketListGenerator
     {
         std::map<std::string, std::shared_ptr<Bucket>> buckets;
         auto has = getHistoryArchiveState();
+        has.prepareForPublish(*mAppApply);
         auto& wm = mAppApply->getWorkScheduler();
         wm.executeWork<T>(buckets, has,
-                          mAppApply->getConfig().LEDGER_PROTOCOL_VERSION,
+                          mAppApply->getConfig().LEDGER_PROTOCOL_VERSION, false,
```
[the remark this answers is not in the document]
graydon: fixed

src/history/HistoryArchive.cpp
graydon: fixed

src/bucket/BucketList.cpp
graydon: fixed

src/bucket/BucketList.cpp
graydon: fixed

src/bucket/BucketList.cpp
graydon: fixed

src/bucket/test/BucketTests.cpp
graydon: fixed

src/bucket/test/BucketTests.cpp
graydon: fixed

src/bucket/test/BucketManagerTests.cpp
graydon: fixed

src/history/test/HistoryTests.cpp
graydon: fixed

src/history/test/HistoryTests.cpp
graydon: fixed

src/herder/test/UpgradesTests.cpp
graydon: fixed

src/history/test/HistoryTests.cpp
graydon: fixed

graydon:
@marta-lokhova Thanks for all the patient editing! Looks great. @MonsieurNicolas Anything left on your end?

MonsieurNicolas:
r+ 119b6514714e962f2054f54db473f96917c5f608

Issue #2202, opened 2019-07-18
Failure to fsync on XDR stream close


graydon:
As noted in #2123 and #1395, we are seeing more instances of "missing buckets" than we expect from mere misconfiguration (which sometimes causes it, but not always).

On investigation with a node operator, we discovered a state that is best understood to be caused by our failure to fsync bucket writes:

  - Bucket went missing at moment of node crash (cloud provider killed host)
  - SQL database matched public network state (postgres WAL integrity maintained)
  - Once restored, subsequent buckets _diverge_ with difference appearing just to be missing entries or updates from lower levels (i.e. some merge restarted against a zero-sized bucket)

There are three necessary fixes for this. This bug tracks one of them: start properly fsync()'ing when we close a bucket (or any XDR file) so we don't lose them even if the OS crashes.

The other two fixes are better diagnostics (tracked with #2123) and verification of hashes while _reading_ buckets (#2203).

PR #2206, opened 2019-07-20
bug 2203: verify input bucket hashes while reading for a merge.

### LINKED ISSUE #2203: Bucket merge should verify _input_ hash while merging
As discussed in #2123, #2202 and #1395, there is some possibility (which those bugs are working to address) of an XDR merge happening with wrong inputs due to filesystem corruption. Even if those bugs are closed, the possibility of corruption remains: someone can always have a flaky disk or accidentally splat some bad bytes in the middle of a bucket file.

Bucket files, however, all have their _sha256_ written on their name. It may be unnecessarily expensive to eagerly check these values on startup, but there's little reason to avoid checking this in streaming mode _while running a merge_ that uses the bucket as an input. If the hash of an input recaltulated by the end of the merge doesn't match the _expected_ input hash, the merge should fail and throw, rather than proceeding to generate further corrupt outputs.

graydon:
# Description

Resolves #2203 

This extends the BucketInputIterator with a SHA256 instance to re-check that the bytes it's _reading_ match the purported hash (name of the bucket) it's expecting.

This is a followup to #2202 to catch data corruption in bucket files, regardless of source. It's not an instant catch -- will still lag until the next time someone reads the file for a merge -- but it tightens the window slightly (from merge-commit time back to merge-completion time) and localizes the error from an unexplained "wrong ledger hash" to a specific corrupt bucket.

# Checklist
- [x] Compiles
- [x] Ran all tests

It costs an extra ~0.5% CPU during merge operations to hash while reading. I think this is tolerable for the improved error precision, but opinions may vary.

### DIFF
--- a/src/bucket/Bucket.cpp
+++ b/src/bucket/Bucket.cpp
@@ -78,6 +78,7 @@ Bucket::containsBucketIdentity(BucketEntry const& id) const
         }
         ++iter;
     }
+    iter.checkHash();
     return false;
 }
 
@@ -561,6 +562,22 @@ mergeCasesWithEqualKeys(MergeCounters& mc, BucketInputIterator& oi,
     ++ni;
 }
 
+static void
+checkAllInputHashes(BucketInputIterator& oi, BucketInputIterator& ni,
+                    std::vector<BucketInputIterator>& shadowIterators)
+{
+    oi.checkHash();
+    ni.checkHash();
+    for (auto& shadow : shadowIterators)
+    {
+        while (shadow)
+        {
+            ++shadow;
+        }
+        shadow.checkHash();
+    }
+}
+
 std::shared_ptr<Bucket>
 Bucket::merge(BucketManager& bucketManager, uint32_t maxProtocolVersion,
               std::shared_ptr<Bucket> const& oldBucket,
@@ -609,6 +626,7 @@ Bucket::merge(BucketManager& bucketManager, uint32_t maxProtocolVersion,
     {
         bucketManager.incrMergeCounters(mc);
     }
+    checkAllInputHashes(oi, ni, shadowIterators);
     MergeKey mk{maxProtocolVersion, keepDeadEntries, oldBucket, newBucket,
                 shadows};
     return out.getBucket(bucketManager, &mk);
--- a/src/bucket/BucketInputIterator.cpp
+++ b/src/bucket/BucketInputIterator.cpp
@@ -4,6 +4,7 @@
 
 #include "bucket/BucketInputIterator.h"
 #include "bucket/Bucket.h"
+#include "util/types.h"
 
 namespace stellar
 {
@@ -14,7 +15,7 @@ namespace stellar
 void
 BucketInputIterator::loadEntry()
 {
-    if (mIn.readOne(mEntry))
+    if (mIn.readOne(mEntry, mHasher.get()))
     {
         mEntryPtr = &mEntry;
         if (mEntry.type() == METAENTRY)
@@ -85,7 +86,10 @@ BucketInputIterator::getMetadata() const
 }
 
 BucketInputIterator::BucketInputIterator(std::shared_ptr<Bucket const> bucket)
-    : mBucket(bucket), mEntryPtr(nullptr), mSeenMetadata(false)
+    : mBucket(bucket)
+    , mEntryPtr(nullptr)
+    , mHasher(SHA256::create())
+    , mSeenMetadata(false)
 {
     // In absence of metadata, we treat every bucket as though it is from ledger
     // protocol 0, which is the protocol of the genesis ledger. At very least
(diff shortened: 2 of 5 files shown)


src/bucket/Bucket.cpp
```diff
@@ -561,6 +562,22 @@ mergeCasesWithEqualKeys(MergeCounters& mc, BucketInputIterator& oi,
     ++ni;
 }
 
+static void
+checkAllInputHashes(BucketInputIterator& oi, BucketInputIterator& ni,
+                    std::vector<BucketInputIterator>& shadowIterators)
+{
+    oi.checkHash();
+    ni.checkHash();
+    for (auto& shadow : shadowIterators)
```
marta-lokhova: Curious about performance here: how does this affect IO and merge latency? Are we reading much more data, as we're iterating through all shadows during a merge? (this will become irrelevant with CAP-0025, but should be considered if this change is to be released prior to protocol 12)

graydon: I didn't compare merge latency or IO load yet. Fair to ask! Though I do expect we'll be taking CAP-0025; but it'd be good to know. Will check soon.

src/bucket/BucketInputIterator.cpp
```diff
@@ -120,4 +124,27 @@ BucketInputIterator& BucketInputIterator::operator++()
...
+    {
+        // Accept the all-zero hash as an alternative hash-name for the empty
+        // bucket. This isn't correct in the SHA-1 sense, but we use it as an
+        // equivalent sentinel value in several places.
+        return;
+    }
+    auto realHash = mHasher->finish();
+    if (realHash != mBucket->getHash())
+    {
+        std::string msg("Bucket: ");
```
MonsieurNicolas: we have `fmt::format` available to us

src/bucket/BucketInputIterator.cpp
```diff
@@ -120,4 +124,27 @@ BucketInputIterator& BucketInputIterator::operator++()
...
     return *this;
 }
+
+void
+BucketInputIterator::checkHash() const
+{
+    if (size() == 0 && isZero(mBucket->getHash()))
+    {
+        // Accept the all-zero hash as an alternative hash-name for the empty
+        // bucket. This isn't correct in the SHA-1 sense, but we use it as an
```
MonsieurNicolas: nit: not "SHA-1"

src/bucket/BucketInputIterator.cpp
```diff
@@ -120,4 +124,27 @@ BucketInputIterator& BucketInputIterator::operator++()
     }
     return *this;
 }
+
+void
+BucketInputIterator::checkHash() const
+{
```
MonsieurNicolas: should we enforce that we've reached the end of the stream `mIn` ?

MonsieurNicolas:
@graydon we can land this in version 12 so we have an opportunity for protocol change. Would the smaller change I suggested be something good enough? I am not super excited about multiplying the amount of work needed to merge buckets and would rather do something that works longer term if we can.

The termination token idea (or something similar) can include whatever we want in there if we want to increase its quality: can be stats on the bucket (nb live/mod/dead entries), which can be useful for other reasons, etc.

graydon:
@MonsieurNicolas I thought a bit about a "termination token" but ultimately I think it won't fly (and/or isn't really necessary), for the following reasons:

  - A short write that ends anywhere in the middle of an XDR value will result in a corrupt read and a thrown exception already.
  - We don't see any such short writes in practice, we only see one kind (or _saw_ in the era of failure-to-fsync anyway; we might never see this anymore) which is the zero-sized bucket.
  - A zero-sized bucket is I think actually a _legal bucket_: a pre-protocol-11 empty bucket. In theory we should never generate such a bucket (the bucket formation code treats treats an output iterator you never write to specially, and returns an empty sentinel bucket value that only has an in-memory form) but I _think_ that's just an optimization; I believe a zero-sized bucket with either the all-zero pseudo-SHA256 or the "SHA256 of zero bytes" value `e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855` should be a legitimate bucket.

  - On the other hand: a zero-sized bucket _without_ either of those two hashes definitely _is_ a corrupt bucket, and it's the kind of corruption we (a) can notice very cheaply by inspection (b) have seen in the wild at least when we were failing to fsync and (c) won't otherwise catch, as the bucket-reading code will treat it as legitimate.

So I propose that rather than hashing the input _or_ adding a termination token, we just add a very small constant-time check that any zero-sized bucket has one of those two hashes, otherwise throw with "truncated file". If this is agreeable I'll implement it. Much simpler change.

MonsieurNicolas:
@graydon 

>     A short write that ends anywhere in the middle of an XDR value will result in a corrupt read and a thrown exception already.

Can you quantify this? Maybe with I don't know what the odds are that we end up at the boundary between two XDR entries. Buffers are written in something like 4096 bytes blocks?
Actually, I just tried it on testnet buckets and it happens quite a bit: about 0.1% of the time we write an xdr object we're on a 4096 boundary.

Anyways, thinking about this a bit more... I am not sure it's worth doing at all now that we're fsyncing, as the only way we would run into this situation is if we're not using temp files, which we're not doing.

graydon:
Closing -- this is both (a) something we're not going to do and (b) somehow interfering with latobarita runs because it can't extract the PR status correctly, or something?

PR #2207, opened 2019-07-25
Allow setting upgrades at a ledger number

marta-lokhova:
This PR ensures we can upgrade at a particular ledger number. Specifically, it is useful in the context of bucketlist-related upgrades, as we can select the ledger, that is right after any large merge spills, to avoid stalling the network.

### DIFF
--- a/src/herder/Upgrades.h
+++ b/src/herder/Upgrades.h
@@ -20,34 +20,70 @@ class Database;
 struct LedgerHeader;
 struct LedgerUpgrade;
 
+class UpgradeParameters
+{
+    optional<VirtualClock::time_point> mUpgradeTime;
+    optional<uint32> mUpgradeLedger;
+
+  public:
+    UpgradeParameters()
+    {
+        VirtualClock::time_point t;
+        setUpgradeAtTime(t);
+    }
+    UpgradeParameters(Config const& cfg)
+    {
+        setUpgradeAtTime(cfg.TESTING_UPGRADE_DATETIME);
+        mProtocolVersion = make_optional<uint32>(cfg.LEDGER_PROTOCOL_VERSION);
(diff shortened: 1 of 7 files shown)


graydon:
This is cool! I somehow thought that upgrade targets were surfaced in the XDR and this was going to take a protocol change, but it looks much easier! Hurray!

src/herder/Upgrades.h
```diff
@@ -20,34 +20,66 @@ class Database;
 struct LedgerHeader;
 struct LedgerUpgrade;
 
-class Upgrades
+class UpgradeParameters
 {
+    VirtualClock::time_point mUpgradeTime;
```
graydon: The only thing I'm a little nervous about in this PR is that you're relying on the users of the UpgradeParameters class to check cases in a specific order: first call `if(upgradeAtLedger())` and only proceed to use `getUpgradeAtTime()` if the former call returns false. This is a bit accident-prone: someone could fail to check in this order and wind up accessing a (meaningless but initialized and valid) "upgrade time" in an instance that has only an upgrade ledger.

In general I'd prefer this type be made something that is more defensive about being in exactly one of two states -- time or ledger -- and prohibiting mixtures of both, or neither.

src/herder/Upgrades.h
```diff
@@ -20,34 +20,70 @@ class Database;
 struct LedgerHeader;
 struct LedgerUpgrade;
 
-class Upgrades
+class UpgradeParameters
 {
+    optional<VirtualClock::time_point> mUpgradeTime;
```
vogel: We can probably switch that to std::variant when we upgrade to C++17

marta-lokhova:
Not required for protocol 12 anymore, but we should revive this PR if we ever want to have this functionality.

PR #2208, opened 2019-07-25
Batch log queries and metrics

graydon:
This uses https://github.com/stellar/medida/pull/13 so don't approve it until that one lands.

Adds a couple helper types for caching EasyLogging++ log-level lookups and batching Medida writes. Low-ish-hanging CPU overhead fruit. Measuring this with perf is quite noisy and these results may not be especially reproducible, but it _does_ look to be a reasonable improvement to me:

Before:
~~~
  Children      Self  Shared Object                Symbol
+    3.78%     0.11%  stellar-core.nobatchmetrics  [.] medida::TimerContext::~TimerContext
+    3.32%     2.21%  stellar-core.nobatchmetrics  [.] medida::Meter::Impl::Mark

  Children      Self  Shared Object                Symbol
+    4.71%     0.00%  stellar-core.nobatchmetrics  [.] stellar::Logging::getLogLevel
~~~

After
~~~
  Children      Self  Shared Object              Symbol
+    2.15%     0.00%  stellar-core.batchmetrics  [.] medida::TimerContext::~TimerContext
     0.94%     0.48%  stellar-core.batchmetrics  [.] medida::Meter::Impl::Mark

  Children      Self  Shared Object              Symbol
+    0.76%     0.03%  stellar-core.batchmetrics  [.] stellar::Logging::getLogLevel
~~~

### DIFF
--- a/src/herder/HerderImpl.h
+++ b/src/herder/HerderImpl.h
@@ -9,6 +9,7 @@
 #include "herder/PendingEnvelopes.h"
 #include "herder/TransactionQueue.h"
 #include "herder/Upgrades.h"
+#include "util/BatchMetrics.h"
 #include "util/Timer.h"
 #include "util/XDROperators.h"
 #include <deque>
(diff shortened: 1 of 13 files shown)


MonsieurNicolas:
For context: what was core doing when you gathered those performance numbers? On a lightly loaded system, even logging bubbles up as a non trivial CPU consumer.

Also, what's the impact of the changes separately?
Batching seems a bit weird as it screws with the rate reported by the metrics

graydon:
Undesirable change, closing

PR #2210, opened 2019-07-30
OrderBookIsNotCrossed invariant

robertDurst:
# Description

This PR introduces a new invariant, named `OrderBookIsNotCrossed`. 

## Background:
Pre-protocol 10 was allowing in some cases the order book to be in a crossed state. The example from [CAP-0004](https://github.com/stellar/stellar-protocol/blob/master/core/cap-0004.md#motivation) reads:

> “In transaction 1, account A creates an offer selling 100 X at a price of 10 Y / 1 X.
In transaction 2, account B creates an offer selling 1 stroop Y at a price of 1 X / 10 Y.
As of protocol version 9, nothing is exchanged and the book remains crossed.” 

Put very simply, a crossed state is one such that the top of the book contains a highest bid higher than the lowest ask. 

## Implementation Overview

This invariant, unlike the previous five, requires *some state* to determine the highest bid and the lowest ask. For this, I utilize a nested map data structure to simulate an order book, which is updated upon each call of `checkOnOperationApply` based on the offer entries in `LedgerTxnDelta`.  Since this invariant will be used by the fuzzing code in #2182, which requires a reset of state between execution cycles, I have introduced a `resetForFuzzer` method to the base invariant class. 

## Open Questions
* Naming -- *IsNot* does not flow off the tongue
* Method of determining best order -- for now I just iterate through all orders for a given asset pair and take the highest/lowest (*O*(N)). This should be fine for small order books in fuzzing, but is not production ready.
* Initial in-memory order book state -- for now there is no way to initialize an order book. For fuzzing I believe this is fine, however it may be something we want (however, this in-memory structure in general is certainly not a production approach).

# Checklist
- [x] Compiles
- [x] Ran all tests

### DIFF
--- a/src/invariant/InvariantManagerImpl.cpp
+++ b/src/invariant/InvariantManagerImpl.cpp
@@ -227,4 +227,15 @@ InvariantManagerImpl::handleInvariantFailure(
         CLOG(ERROR, "Invariant") << REPORT_INTERNAL_BUG;
     }
 }
+
+#ifdef FUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION
+void
+InvariantManagerImpl::resetForFuzzer()
+{
+    for (auto invariant : mEnabled)
+    {
+        invariant->resetForFuzzer();
+    }
+}
+#endif // FUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION
 }
--- a/src/invariant/InvariantManagerImpl.cpp
+++ b/src/invariant/InvariantManagerImpl.cpp
@@ -232,7 +232,7 @@ InvariantManagerImpl::handleInvariantFailure(
 void
 InvariantManagerImpl::resetForFuzzer()
 {
-    for (auto invariant : mEnabled)
+    for (auto const& invariant : mEnabled)
     {
         invariant->resetForFuzzer();
     }
--- a/src/invariant/InvariantManagerImpl.cpp
+++ b/src/invariant/InvariantManagerImpl.cpp
@@ -237,6 +237,7 @@ InvariantManagerImpl::snapshotForFuzzer()
         invariant->snapshotForFuzzer();
     }
 }
+
 void
 InvariantManagerImpl::resetForFuzzer()
 {
--- a/src/invariant/InvariantManagerImpl.cpp
+++ b/src/invariant/InvariantManagerImpl.cpp
@@ -230,6 +230,14 @@ InvariantManagerImpl::handleInvariantFailure(
 
 #ifdef FUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION
 void
+InvariantManagerImpl::snapshotForFuzzer()
+{
+    for (auto const& invariant : mEnabled)
+    {
+        invariant->snapshotForFuzzer();
+    }
+}
+void
 InvariantManagerImpl::resetForFuzzer()
 {
     for (auto const& invariant : mEnabled)
--- a/src/invariant/OrderBookIsNotCrossed.cpp
+++ b/src/invariant/OrderBookIsNotCrossed.cpp
@@ -1,0 +1,162 @@
+// Copyright 2019 Stellar Development Foundation and contributors. Licensed
+// under the Apache License, Version 2.0. See the COPYING file at the root
+// of this distribution or at http://www.apache.org/licenses/LICENSE-2.0
+
+#include "invariant/OrderBookIsNotCrossed.h"
+#include "invariant/InvariantManager.h"
+#include "ledger/LedgerTxn.h"
+#include "lib/util/format.h"
+#include "main/Application.h"
+#include "xdr/Stellar-ledger-entries.h"
+
+#include <numeric>
+
+namespace stellar
+{
+AssetId
+getAssetID(Asset const& asset)
+{
+    auto r = std::string{};
+    switch (asset.type())
+    {
+    case stellar::ASSET_TYPE_NATIVE:
+        return std::string{"XLM-native"};
+    case stellar::ASSET_TYPE_CREDIT_ALPHANUM4:
+        assetCodeToStr(asset.alphaNum4().assetCode, r);
+        return r + "-" + KeyUtils::toStrKey(getIssuer(asset));
+    case stellar::ASSET_TYPE_CREDIT_ALPHANUM12:
+        assetCodeToStr(asset.alphaNum12().assetCode, r);
+        return r + "-" + KeyUtils::toStrKey(getIssuer(asset));
+    };
+}
+
+void
+deleteOffer(OrderBook& orderBook, OfferEntry const& oe)
+{
+    auto offerId = oe.offerID;
+    auto sellAssetId = getAssetID(oe.selling);
+    auto buyAssetId = getAssetID(oe.buying);
+    orderBook[sellAssetId][buyAssetId].erase(offerId);
+}
+
+void
+createOrModifyOffer(OrderBook& orderBook, OfferEntry const& oe)
+{
+    auto offerId = oe.offerID;
+    auto sellAssetId = getAssetID(oe.selling);
+    auto buyAssetId = getAssetID(oe.buying);
+    orderBook[sellAssetId][buyAssetId][offerId] = oe;
+}
+
+void
+updateOrderBook(LedgerTxnDelta const& ltxd, OrderBook& orderBook)
+{
+    for (auto const& entry : ltxd.entry)
+    {
+        // there are three possible "deltas" for an offer:
+        //      CREATED:  (nil)     -> LedgerKey
+        //      MODIFIED: LedgerKey -> LedgerKey
+        //      DELETED:  LedgerKey -> (nil)
+        if (entry.first.type() == OFFER)
+        {
+            if (entry.second.current)
+            {
+                createOrModifyOffer(orderBook,
+                                    entry.second.current->data.offer());
+            }
+            else
+            {
+                deleteOffer(orderBook, entry.second.previous->data.offer());
+            }
+        }
+    }
+}
+
+std::string
+check(Operation const& operation, OrderBook& orderBook)
+{
+    auto getCheckForCrossedMessage = [&](AssetId const& a, AssetId const& b) {
+        auto asks = orderBook[a][b];
+        auto bids = orderBook[b][a];
+
+        auto lowestAsk =
+            std::accumulate(asks.begin(), asks.end(), __DBL_MAX__,
+                            [](double lowestAsk, auto curEntry) {
+                                double curAsk =
+                                    double(curEntry.second.price.n) /
+                                    double(curEntry.second.price.d);
+
+                                return lowestAsk > curAsk ? curAsk : lowestAsk;
+                            });
+
+        auto highestBid = std::accumulate(
+            bids.begin(), bids.end(), 0.0,
+            [](double highestBid, auto curEntry) {
+                double curBid = 1.0 / (double(curEntry.second.price.n) /
+                                       double(curEntry.second.price.d));
+
+                return highestBid < curBid ? curBid : highestBid;
+            });
+
+        if (highestBid >= lowestAsk)
+        {
+            return fmt::format(
+                "Order book is in a crossed state for {} - {} asset pair.\nTop "
+                "of the book is:\n\tAsk price: {}\n\tBid price: {}\n\nWhere {} "
+                ">= {}!",
+                a, b, lowestAsk, highestBid, highestBid, lowestAsk);
+        }
+
+        return std::string{};
+    };
+
+    if (operation.body.type() == MANAGE_BUY_OFFER)
+    {
+        return getCheckForCrossedMessage(
+            getAssetID(operation.body.manageBuyOfferOp().buying),
+            getAssetID(operation.body.manageBuyOfferOp().selling));
+    }
+    else if (operation.body.type() == MANAGE_SELL_OFFER)
+    {
+        return getCheckForCrossedMessage(
+            getAssetID(operation.body.manageSellOfferOp().buying),
+            getAssetID(operation.body.manageSellOfferOp().selling));
+    }
+    else if (operation.body.type() == CREATE_PASSIVE_SELL_OFFER)
+    {
+        return getCheckForCrossedMessage(
+            getAssetID(operation.body.createPassiveSellOfferOp().buying),
+            getAssetID(operation.body.createPassiveSellOfferOp().selling));
+    }
+
+    return {};
+}
+
+std::shared_ptr<Invariant>
+OrderBookIsNotCrossed::registerInvariant(Application& app)
+{
+    return app.getInvariantManager().registerInvariant<OrderBookIsNotCrossed>();
+}
+std::string
+OrderBookIsNotCrossed::getName() const
+{
+    return "OrderBookIsNotCrossed";
+}
+
+std::string
+OrderBookIsNotCrossed::checkOnOperationApply(Operation const& operation,
+                                             OperationResult const& result,
+                                             LedgerTxnDelta const& ltxDelta)
+{
+    updateOrderBook(ltxDelta, mOrderBook);
+    return check(operation, mOrderBook);
+}
+
+#ifdef FUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION
+void
+OrderBookIsNotCrossed::resetForFuzzer()
+{
+    mOrderBook = {};
+};
+#endif // FUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION
+}
--- a/src/invariant/OrderBookIsNotCrossed.cpp
+++ b/src/invariant/OrderBookIsNotCrossed.cpp
@@ -9,6 +9,7 @@
 #include "lib/util/format.h"
 #include "main/Application.h"
 #include "xdr/Stellar-ledger-entries.h"
+#include <xdrpp/printer.h>
 
 namespace stellar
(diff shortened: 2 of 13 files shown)


src/invariant/test/OrderBookIsNotCrossedTests.cpp
```diff
@@ -0,0 +1,274 @@
...
+        REQUIRE(orders.size() == 1);
+        REQUIRE(orders.at(offer1.data.offer().offerID).amount == 3);
+        REQUIRE(orders.at(offer1.data.offer().offerID).price.n == 3);
+        REQUIRE(orders.at(offer1.data.offer().offerID).price.d == 2);
+
+        ltxPtr->commit();
+
+        SECTION("Then modify offer")
+        {
+            auto offer2 = generateOffer(cur1, cur2, 2, Price{5, 3});
```
jonjove: Why not just copy `offer1` and modify it directly?

src/invariant/test/OrderBookIsNotCrossedTests.cpp
```diff
@@ -0,0 +1,274 @@
...
+        Operation op{};
+        OperationResult opRes{};
+        auto offer1 = generateOffer(cur1, cur2, 3, Price{3, 2});
+
+        auto ltxStore = std::make_unique<LedgerTxn>(app->getLedgerTxnRoot());
+        auto ltxPtr = ltxStore.get();
+        auto entry = ltxPtr->create(offer1);
+
+        invariant->checkOnOperationApply(op, opRes, ltxPtr->getDelta());
+        auto orderbook = invariant->getOrderBook();
```
jonjove: You want `auto const&` here and many other places in this file.

src/invariant/test/OrderBookIsNotCrossedTests.cpp
```diff
@@ -0,0 +1,274 @@
...
+    genApplyCheckCreateOffer(cur1, cur2, 2, Price{5, 3}, *app, SHOULD_PASS);
+    genApplyCheckCreateOffer(cur2, cur1, 4, Price{3, 4}, *app, SHOULD_PASS);
+    genApplyCheckCreateOffer(cur2, cur1, 1, Price{4, 5}, *app, SHOULD_PASS);
+
+    SECTION("Not crossed when highest bid < lowest ask")
+    {
+        SECTION("Create offer")
+        {
+            // Offer 5: 7 A @  8/5 (B/A)
+            auto offer5 = generateOffer(cur1, cur2, 7, Price{8, 5});
```
jonjove: Why not use `genApplyCheckCreateOffer` here? (This question applies in multiple places in this file)

src/invariant/test/OrderBookIsNotCrossedTests.cpp
```diff
@@ -0,0 +1,274 @@
...
+    auto offer = generateOffer(ask, bid, amount, price);
+    auto op = txtest::manageOffer(offer.data.offer().offerID, ask, bid, price,
+                                  amount);
+    std::vector<LedgerEntry> current{offer};
+    auto updates = makeUpdateList(current, nullptr);
+    REQUIRE(XNOR(store(app, updates, nullptr, nullptr, &op), shouldPass));
+}
+
+void
+genApplyCheckModifyOffer(Asset& ask, Asset& bid, int64 amount, Price price,
```
jonjove: Passing around the operations is kind of clunky. The invariant only depends on the assets in the operation, so you should be able to prepare them in these functions.

src/invariant/test/OrderBookIsNotCrossedTests.cpp
```diff
@@ -0,0 +1,274 @@
...
+    //     1 B @  4/5 (A/B)
+    // where A = cur1 and B = cur2
+    genApplyCheckCreateOffer(cur1, cur2, 3, Price{3, 2}, *app, SHOULD_PASS);
+    genApplyCheckCreateOffer(cur1, cur2, 2, Price{5, 3}, *app, SHOULD_PASS);
+    genApplyCheckCreateOffer(cur2, cur1, 4, Price{3, 4}, *app, SHOULD_PASS);
+    genApplyCheckCreateOffer(cur2, cur1, 1, Price{4, 5}, *app, SHOULD_PASS);
+
+    SECTION("Not crossed when highest bid < lowest ask")
+    {
+        SECTION("Create offer")
```
jonjove: Why put a section directly inside another section? (This applies in multiple places in this file)

src/invariant/InvariantManagerImpl.cpp
```diff
@@ -227,4 +227,15 @@ InvariantManagerImpl::handleInvariantFailure(
         CLOG(ERROR, "Invariant") << REPORT_INTERNAL_BUG;
     }
 }
+
+#ifdef FUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION
+void
+InvariantManagerImpl::resetForFuzzer()
+{
+    for (auto invariant : mEnabled)
```
marta-lokhova: no need to copy `shared_ptr`s here, use const ref

src/invariant/OrderBookIsNotCrossed.cpp
```diff
@@ -0,0 +1,162 @@
...
+    switch (asset.type())
+    {
+    case stellar::ASSET_TYPE_NATIVE:
+        return std::string{"XLM-native"};
+    case stellar::ASSET_TYPE_CREDIT_ALPHANUM4:
+        assetCodeToStr(asset.alphaNum4().assetCode, r);
+        return r + "-" + KeyUtils::toStrKey(getIssuer(asset));
+    case stellar::ASSET_TYPE_CREDIT_ALPHANUM12:
+        assetCodeToStr(asset.alphaNum12().assetCode, r);
+        return r + "-" + KeyUtils::toStrKey(getIssuer(asset));
```
marta-lokhova: I think you need to add a `default` to avoid compiler warnings, even though all cases are covered here. Usually it's an `abort` call, see `BasicWork::getStatus()` as an example

src/invariant/OrderBookIsNotCrossed.h
```diff
@@ -0,0 +1,62 @@
+#pragma once
+
+// Copyright 2019 Stellar Development Foundation and contributors. Licensed
+// under the Apache License, Version 2.0. See the COPYING file at the root
+// of this distribution or at http://www.apache.org/licenses/LICENSE-2.0
+
+#include "invariant/Invariant.h"
+#include "xdr/Stellar-ledger-entries.h"
+
+#include <unordered_map>
```
marta-lokhova: I don't think you need this include

src/invariant/OrderBookIsNotCrossed.h
```diff
@@ -0,0 +1,62 @@
...
+using Orders = std::unordered_map<int64_t, OfferEntry>;
+// AssetOrders, orders by asset, a map of AssetId --> map of Orders
+using AssetOrders = std::unordered_map<AssetId, Orders>;
+// OrderBook is a mapping of AssetId --> map of asset pair's orders
+using OrderBook = std::unordered_map<AssetId, AssetOrders>;
+
+class OrderBookIsNotCrossed : public Invariant
+{
+  public:
+    explicit OrderBookIsNotCrossed() : Invariant(false)
```
marta-lokhova: Why is this invariant not strict? I thought we want to throw (and crash)?

src/invariant/test/OrderBookIsNotCrossedTests.cpp
```diff
@@ -0,0 +1,274 @@
...
+#include "invariant/InvariantManager.h"
+#include "invariant/OrderBookIsNotCrossed.h"
+#include "invariant/test/InvariantTestUtils.h"
+#include "ledger/LedgerTxn.h"
+#include "test/TestUtils.h"
+#include "test/TxTests.h"
+#include "test/test.h"
+
+using namespace stellar;
+using namespace stellar::InvariantTestUtils;
```
marta-lokhova: I don't think you need this.

src/main/ApplicationImpl.cpp
```diff
@@ -142,6 +143,7 @@ ApplicationImpl::initialize(bool createNewDB)
     ConservationOfLumens::registerInvariant(*this);
     LedgerEntryIsValid::registerInvariant(*this);
     LiabilitiesMatchOffers::registerInvariant(*this);
+    OrderBookIsNotCrossed::registerInvariant(*this);
```
marta-lokhova: I don't think we want to allow people to enable this invariant in production (at least right now :) )

src/invariant/OrderBookIsNotCrossed.h
robertDurst: I think I do. `Invariant.h` only includes:
```
#include <memory>
#include <string>
```

src/invariant/test/OrderBookIsNotCrossedTests.cpp
robertDurst: As far as I can tell I do, won't compile without and it is also in most of the other invariant tests.

robertDurst:
@marta-lokhova and @jonjove I have refactored per your feedback. Thanks -- I think this is quite a bit cleaner now! 

Since some of the refactors were more involved, the diffs may make review a bit harder. To address this I tried to sensibly organize my changes into a few commits labeled `[pr fix]`, however if a) this organization does not make sense or b) it would be easier if I just re-org'd everything together, let me know and I can fix up ASAP.

src/invariant/OrderBookIsNotCrossed.h
```diff
@@ -0,0 +1,62 @@
...
+struct LedgerTxnDelta;
+struct OfferEntry;
+
+// AssetId is defined as a concatenation of issuer and asset code:
+// ASSET_CODE | '-' | ISSUER_ACCOUNT_ID
+using AssetId = std::string;
+// Orders is a map of OfferId (int64_t) --> OfferEntry
+using Orders = std::unordered_map<int64_t, OfferEntry>;
+// AssetOrders, orders by asset, a map of AssetId --> map of Orders
+using AssetOrders = std::unordered_map<AssetId, Orders>;
```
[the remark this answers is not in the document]
jonjove: The change you made is not what I meant, let's discuss further offline.

robertDurst:
Updated for proper template specialization for `std::hash` for `Asset` instead of `getAssetId`--> `std::string` approach.

robertDurst:
Changed header to WIP -- realize this does not properly account for `PathPayment`. It makes the assumption only a single asset pair will be modified and thus only checks if a single asset pair's orders are crossed. Will change header back when fixed.

robertDurst:
Latest commit 86e3826 is a bit substantial (hence WIP in header). The most notable changes from the initial PR include:
* utilizing `Operation` to derive asset pairs for which to check order book is crossed
* proper accounting for all operations that touch the order book (`ManageBuyOffer`, `ManageSellOffer`, `CreatePassiveOffer`, `PathPayment`, and `AllowTrust`)
* utilize set for Orders (`std::set<OfferEntry, OfferEntryCmp`) instead of `std::unordered_map<offerId, OfferEntry>` since I do not need to look up individual offers by id, but rather am interested in the least priced, or best offers (thinking ahead a bit here to _BestOffersTakenInvariant_)
* decoupled `genApplyCheck` methods in testing to `gen` and `applyCheck` allowing for multiple offers to be `applyCheck`'d and also separated `updateOrderBook` and `check` to allow for more complex checking later on... again thinking about _BestOffersTakenInvariant_

src/invariant/OrderBookIsNotCrossed.cpp
```diff
@@ -0,0 +1,239 @@
...
+
+std::vector<std::pair<Asset, Asset>>
+extractAssetPairs(Operation const& op, LedgerTxnDelta const& ltxd)
+{
+    switch (op.body.type())
+    {
+    case MANAGE_BUY_OFFER:
+    {
+        auto const& offer = op.body.manageBuyOfferOp();
+        return {std::pair<Asset, Asset>(offer.selling, offer.buying)};
```
vogel: std::make_pair can infer types for arguments, so until C++17 please use that

src/invariant/OrderBookIsNotCrossed.cpp
```diff
@@ -0,0 +1,239 @@
...
+        {
+            return {std::pair<Asset, Asset>(pp.sendAsset, pp.destAsset)};
+        }
+
+        std::vector<std::pair<Asset, Asset>> assets;
+        // For send, dest, {A, B} we get: {send, A}, {A, B}, {B, dest}
+        for (int i = 0; i < pp.path.size(); ++i)
+        {
+            // beginning: send -> A
+            if (i == 0)
```
vogel: is does not make sense for those conditions to be inside loop
you can loop from 1 to pp.path.size() - 1 and have those special cases outside of loop

src/invariant/OrderBookIsNotCrossed.cpp
YanaSneg23: S

robertDurst: Ok will do. Ah cool I see what you mean: https://www.codingame.com/playgrounds/2205/7-features-of-c17-that-will-simplify-your-code/template-argument-deduction-for-class-templates

src/invariant/OrderBookIsNotCrossed.cpp
robertDurst: Good point. Fixed, simpler now.

src/invariant/OrderBookIsNotCrossed.cpp
```diff
@@ -13,81 +13,31 @@
...
-            // LedgerTxnDelta entries
-            for (auto const& entry : ltxd.entry)
+            auto const& offer = entry.second.previous
+                                    ? entry.second.previous->data.offer()
+                                    : entry.second.current->data.offer();
+
+            auto assetPair = std::make_pair(offer.selling, offer.buying);
+            auto oppositeAssetPair =
+                std::make_pair(offer.buying, offer.selling);
+            if (assets.find(oppositeAssetPair) == assets.end() &&
```
jonjove: Technically, you only need to check `oppositeAssetPair` here. The insert on line 35 will do nothing if `assetPair` is already included.

src/invariant/OrderBookIsNotCrossed.cpp
```diff
@@ -13,81 +13,31 @@
...
-        // deleted
-        if (!at.authorize)
+        if (entry.first.type() == OFFER)
         {
-            std::vector<std::pair<Asset, Asset>> assets;
-            // since we only get one side of the asset pair from the operation,
-            // we derive the rest of the information from the
-            // LedgerTxnDelta entries
-            for (auto const& entry : ltxd.entry)
+            auto const& offer = entry.second.previous
```
jonjove: Looking closely at this, I'm not sure that this makes sense as implemented. An important piece of context here: it _is_ possible to modify the `buying` or `selling` assets on an offer. In those cases, `previous` and `current` won't have the same asset pairs.

But in any case, I'm wondering why we should ever care about the `previous` state? If an order book wasn't crossed and an order was deleted, then it still isn't crossed. If an order book wasn't crossed and an order was modified such that it has the same asset pair, then it will be gathered from the current state. If an order book wasn't crossed and an order was modified such that it doesn't have the same asset pair, then it still isn't crossed. In any of these cases, the `previous` asset pair isn't useful. Am I missing something here?

src/invariant/OrderBookIsNotCrossed.h
```diff
@@ -83,7 +83,7 @@ class OrderBookIsNotCrossed : public Invariant
     void deleteFromOrderBook(OfferEntry const& oe);
     void addToOrderBook(OfferEntry const& oe);
     void updateOrderBook(LedgerTxnDelta const& ltxd);
-    std::string check(std::vector<std::pair<Asset, Asset>> assetPairs);
+    std::string check(std::set<std::pair<Asset, Asset>> assetPairs);
```
jonjove: Nitpick: `... const& assetPairs`

src/invariant/OrderBookIsNotCrossed.cpp
```diff
@@ -0,0 +1,237 @@
...
+getMinOfferPrice(Orders const& orders)
+{
+    return orders.cbegin() == orders.cend() ? __DBL_MAX__
+                                            : price(*orders.cbegin());
+}
+
+std::string
+checkCrossed(Asset const& a, Asset const& b, OrderBook const& orderBook)
+{
+    // if either side of order book for asset pair empty or does not yet exist,
```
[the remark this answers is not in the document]
jonjove: I think you forgot to fill in the rest of the code starting from `// ...` (I only gave you the first part of it). Should need to make some changes through `auto const& bids = ...`.

src/invariant/OrderBookIsNotCrossed.cpp
robertDurst: Good point -- fixed.

src/invariant/OrderBookIsNotCrossed.cpp
robertDurst: Thanks, I did not realize you can change the asset, I thought you could only change the price and amount. Good catch.

Very good point -- if a `set A` of orders on an order book is not crossed, there is **no** `subset B` such that it may become crossed. The spread between the lowest ask and the highest bid for the subset `B` is `>=` the spread before the deletion of any order `A`.

robertDurst: **TLDR** they can only be in a less or equally crossed state so if the state was not crossed before it is not crossed now.

MonsieurNicolas:
will reopen later

PR #2204, opened 2019-07-18
XDR: fsync on close (also bucket dir), with option to disable.

graydon:
# Description

Resolves #2202

# Checklist
- [x] Reviewed the [contributing](https://github.com/stellar/stellar-core/blob/master/CONTRIBUTING.md#submitting-changes) document
- [x] Rebased on top of master (no merge commits)
- [x] Ran `clang-format` v5.0.0 (via `make format` or the Visual Studio extension)
- [x] Compiles
- [x] Ran all tests
- [x] If change impacts performance, include supporting evidence per the [performance document](https://github.com/stellar/stellar-core/blob/master/performance-eval/performance-eval.md)

Performance is worse but not hugely. Benchmarks that hammer on the bucket system non-stop run about 10x slower wall time on local NVMe and 100x worse on a spindle. And at the _very smallest_ end of the merge times, slowdown is indeed also 10x-100x: 1ms merge becomes 10ms on NVMe or even 100ms on spindle. At level 0, the 100x case is still only taking 1-3% of allowed time for the merge -- we had a ton of headroom -- but it's worse for sure.

However, it also amortizes-away pretty smoothly as we get into bigger merges. By level 4 we're back down to 0.03% of available time and at level 5 down to 0.009%. That is: the fsync is adding a definitely-noticable 100ms to a merge running on the spindle, but it's not adding _proportionally_ much more as we scale to _somewhat_ larger buckets.

To measure whether this vanishes into the noise _completely_ when scaling up (or if, say, full buffers / actually having some real data to write changes things again), I made a synthetic 1.4gb XDR test and ran it against the spindle in a loop. This gives not super wonderful but also not super deadly results. The large files _do_ definitely show different effects that are not "constant 100ms totally vanishing into the noise":

~~~
2019-07-18T14:56:41.781 <test> [Fs INFO] wrote 1410837608 bytes to fsync file in 26094ms
2019-07-18T14:56:52.930 <test> [Fs INFO] wrote 1410837608 bytes to no-fsync file in 11149ms
2019-07-18T14:57:28.145 <test> [Fs INFO] wrote 1410837608 bytes to fsync file in 35214ms
2019-07-18T14:57:40.637 <test> [Fs INFO] wrote 1410837608 bytes to no-fsync file in 12491ms
2019-07-18T14:58:01.238 <test> [Fs INFO] wrote 1410837608 bytes to fsync file in 20600ms
[...]

PR #2211, opened 2019-08-01
Refactor publish-related work to avoid races

### LINKED ISSUE #2190: Publishing to multiple archives has races that may lead to corrupt archives
Looking at `GzipAndPutFilesWork`, it looks like `GzipFileWork` is run *in parallel* on the same files (each archive tries to compress the same bucket file for example).

I suspect this may lead to strange publish failures and even corruptions: it's possible that a gzip process is spawned right before a different archives spawns a "put" of the same file, if this happens an empty/partial file may get uploaded (also, we use stdout redirection to the final name in "process manager", not a safer "redirect to temp file + rename").

marta-lokhova:
Resolves #2190


src/historywork/PutFilesWork.cpp
```diff
@@ -0,0 +1,61 @@
...
+#include "work/WorkSequence.h"
+
+namespace stellar
+{
+
+PutFilesWork::PutFilesWork(Application& app,
+                           std::shared_ptr<HistoryArchive> archive,
+                           std::shared_ptr<StateSnapshot> snapshot,
+                           HistoryArchiveState const& remoteState)
+    // Each gzip-and-put-file sequence will retry correctly
```
MonsieurNicolas: comment still says "gzip"

src/historywork/PutFilesWork.cpp
```diff
@@ -0,0 +1,61 @@
...
+            addWork<WorkSequence>("mkdir-and-put-file-" + f->localPath_gz(),
+                                  seq, BasicWork::RETRY_ONCE);
+        }
+        mChildrenSpawned = true;
+    }
+    else
+    {
+        return checkChildrenStatus();
+    }
+    return State::WORK_RUNNING;
```
MonsieurNicolas: avoid the 2 "return" like this, use a return code instead or move this one up to its related control block

src/catchup/CatchupWork.cpp
```diff
@@ -181,14 +181,15 @@ CatchupWork::doWork()
                       mCatchupConfiguration.toLedger() + 1) -
                       1;
         mGetHistoryArchiveStateWork =
-            addWork<GetHistoryArchiveStateWork>(mRemoteState, toCheckpoint);
+            addWork<GetHistoryArchiveStateWork>(toCheckpoint);
```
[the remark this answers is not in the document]
marta-lokhova: Previously we were passing a reference to HAS to `GetHistoryArchiveStateWork` to load the state into (which other works, like CatchupWork, would later use), which is quite error-prone. Instead, we're now explicitly querying `GetHistoryArchiveStateWork` for the loaded state instead of relying on the reference.

src/process/ProcessManagerImpl.cpp
```diff
@@ -72,12 +75,27 @@ class ProcessExitEvent::Impl
         , mOuterEc(outerEc)
         , mCmdLine(cmdLine)
         , mOutFile(outFile)
+        , mTempFile(binToHex(randomBytes(5)))
```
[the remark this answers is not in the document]
marta-lokhova: Ah yes, you're right, didn't realize that process files are cleaned up manually in relevant works. Will update.

src/catchup/CatchupWork.cpp
```diff
@@ -181,14 +181,15 @@ CatchupWork::doWork()
...
         return State::WORK_RUNNING;
     }
     else if (mGetHistoryArchiveStateWork->getState() != State::WORK_SUCCESS)
     {
         return mGetHistoryArchiveStateWork->getState();
     }
 
+    mRemoteState = mGetHistoryArchiveStateWork->getHistoryArchiveState();
```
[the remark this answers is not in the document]
marta-lokhova: Good idea, updated.

src/process/ProcessManagerImpl.cpp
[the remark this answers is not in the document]
marta-lokhova: As for the file names collisions, we're bounded by the max number of processes allowed (whereas TmpDir isn't)

src/process/ProcessManagerImpl.cpp
```diff
@@ -72,12 +75,27 @@ class ProcessExitEvent::Impl
...
     }
+
+    void
+    finish()
+    {
+        if (!mOutFile.empty() && mRunning)
+        {
+            if (std::rename(mTempFile.c_str(), mOutFile.c_str()))
+            {
+                throw std::runtime_error(
```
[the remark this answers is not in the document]
marta-lokhova: Good point, updating this logic to return bool, and set an error code in case of a failure.

src/historywork/PutSnapshotFilesWork.cpp
```diff
@@ -29,53 +31,96 @@ PutSnapshotFilesWork::PutSnapshotFilesWork(
...
+PutSnapshotFilesWork::getFilesToZip()
+{
+    // Sanity check: there are states for all archives
+    if (mGetStateWorks.size() !=
+        mApp.getHistoryArchiveManager().getWritableHistoryArchives().size())
+    {
+        throw std::runtime_error("Corrupted GetHistoryArchiveStateWork");
+    }
+
+    std::unordered_set<std::shared_ptr<FileTransferInfo>> filesToZip{};
```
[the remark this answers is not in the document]
marta-lokhova: hmm looks like the change I had for this function got lost in rebase somehow, let me update it.

marta-lokhova:
Ok, comments should be addressed now: currently all changes are in a separate commit, I'll squash and rebase when we're ready to land.

src/process/ProcessManagerImpl.cpp
```diff
@@ -70,30 +72,40 @@ class ProcessExitEvent::Impl
...
         {
-            if (std::rename(mTempFile.c_str(), mOutFile.c_str()))
+            if (fs::exists(mOutFile))
+            {
+                CLOG(WARNING, "Process")
+                    << "Outfile already exists: " << mOutFile;
+                return false;
+            }
+            else if (mRunning &&
+                     std::rename(mTempFile.c_str(), mOutFile.c_str()))
```
MonsieurNicolas: this will need to be updated to use PR #2204 semantics

src/process/ProcessManagerImpl.cpp
```diff
@@ -70,30 +72,40 @@ class ProcessExitEvent::Impl
...
     {
-        if (!mOutFile.empty() && mRunning)
+        if (!mOutFile.empty())
         {
-            if (std::rename(mTempFile.c_str(), mOutFile.c_str()))
+            if (fs::exists(mOutFile))
+            {
+                CLOG(WARNING, "Process")
+                    << "Outfile already exists: " << mOutFile;
+                return false;
```
MonsieurNicolas: should this function return an `ec` instead of a `bool`?

src/process/ProcessManagerImpl.cpp
```diff
@@ -523,15 +543,15 @@ ProcessManagerImpl::handleSignalWait()
...
+bool
 ProcessManagerImpl::handleProcessTermination(int pid, int status)
 {
     std::lock_guard<std::recursive_mutex> guard(mProcessesMutex);
     auto pair = mProcesses.find(pid);
     if (pair == mProcesses.end())
     {
         CLOG(DEBUG, "Process") << "failed to find process with pid " << pid;
-        return;
+        return false;
```
MonsieurNicolas: same than Win32: should probably return an `ec` instead of `bool`

src/process/ProcessManagerImpl.cpp
```diff
@@ -701,13 +726,23 @@ std::weak_ptr<ProcessExitEvent>
 ProcessManagerImpl::runProcess(std::string const& cmdLine, std::string outFile)
 {
     std::lock_guard<std::recursive_mutex> guard(mProcessesMutex);
+    if (!mTmpDir)
+    {
+        mTmpDir =
+            std::make_unique<TmpDir>(mApp.getTmpDirManager().tmpDir("process"));
```
MonsieurNicolas: maybe you should have a `mTmpDirManager` instead of `mApp` ?
Actually - maybe we should just create the tmp dir in the constructor? that way we avoid the singleton code

src/process/ProcessManagerImpl.cpp
```diff
@@ -701,13 +726,23 @@ std::weak_ptr<ProcessExitEvent>
...
+    }
+
     auto pe =
         std::shared_ptr<ProcessExitEvent>(new ProcessExitEvent(mIOContext));
 
     std::weak_ptr<ProcessManagerImpl> weakSelf(
         std::static_pointer_cast<ProcessManagerImpl>(shared_from_this()));
+
+    auto tempFile =
+        mTmpDir->getName() + "/temp-" + std::to_string(mTempFileCount++);
```
MonsieurNicolas: `mTempFileCount` is incremented from inside a lock_guard, so it's safe to use a regular int as opposed to `std::atomic`

src/process/ProcessManagerImpl.cpp
```diff
@@ -726,6 +761,18 @@ ProcessManagerImpl::maybeRunPendingProcesses()
     {
         auto i = mPending.front();
         mPending.pop_front();
+
+        if (fs::exists(i->mImpl->mOutFile))
+        {
+            i->mImpl->cancel(asio::error_code(asio::error::invalid_argument,
```
MonsieurNicolas: the better way (avoids duplicating code and logs properly) is to just throw a `runtime_error` with that error message

src/process/ProcessManagerImpl.cpp
marta-lokhova: that's what i initially did, but our tests are setup in a funny way: they create an application once and potentially run newdb on the same application instance. This causes `bucket` dir to be deleted, and then re-created (see `BucketManagerImpl::dropAll`). At that point, process manager is already initialized, but its temp dir is deleted. A solution would be to spin up an application twice in our tests, once to do `newdb`, and then to actually run it. I can move tmp dir creation to the constructor if we're ok with the overhead.

MonsieurNicolas: From what I see it's `ApplicationImpl::initialize` that's written in a funny way:
maybe all we need to do is move `    mProcessManager = ProcessManager::create(*this);`
 to after the newdb/upgrade block.

src/historywork/GetHistoryArchiveStateWork.h
```diff
@@ -31,11 +31,23 @@ class GetHistoryArchiveStateWork : public Work
 
   public:
     GetHistoryArchiveStateWork(
```
vogel: I would be happy to see that change everywhere we are passing references to work constructors (like VerifyBucketWork).

marta-lokhova: Agreed, but it's outside of the scope of this PR. I can open an issue to track this though

src/process/ProcessManagerImpl.cpp
MonsieurNicolas: I was saying that so that you can return different error codes based on the type of error instead of performing a blanket conversion

MonsieurNicolas:
Just one pending question @marta-lokhova

src/process/ProcessManagerImpl.cpp
marta-lokhova: ok let me update it

MonsieurNicolas:
r+ c056dbfd30cab72ad50a75c92e243739e4dcb1b3

PR #2213, opened 2019-08-02
Cut network from state

graydon:
# Description

This is a preliminary "very coarse cut" through the middle of the network/storage (more specifically herder/ledger) axis of stellar-core, attempting to produce a new `RelayApplication` subclass of `ApplicationImpl` that _only_ does networking-stuff. It has no database, bucketlist, etc.

Most of this consists of virtualizing methods and construction, and putting stubs in subclasses. Plus a little splitting or moving existing logic into shapes that support that. And making the database initialization a little lazier.

The _purpose_ of this is just testing at this point: we'd like to do some work on optimizing the networking system _alone_, and we expect it might be helpful to be able to test and experiment on it in isolation. We may also find some use for this type of not-even-watching "pure network relay" node in future network topologies, but that's speculation at this point.

Update: now passing tests, ready for review.

# Checklist
- [x] Compiles
- [x] Ran all tests

### DIFF
--- a/src/main/ApplicationUtils.cpp
+++ b/src/main/ApplicationUtils.cpp
@@ -48,7 +48,8 @@ runWithConfig(Config cfg)
     Application::pointer app;
     try
     {
-        app = Application::create(clock, cfg, false);
+        app = Application::create(
+            clock, cfg, Application::InitialDBMode::APP_DB_UPGRADE_EXISTING);
 
         if (!app->getHistoryArchiveManager().checkSensibleConfig())
         {
@@ -140,7 +141,8 @@ setForceSCPFlag(Config cfg, bool set)
 {
     VirtualClock clock;
     cfg.setNoListen();
-    Application::pointer app = Application::create(clock, cfg, false);
+    Application::pointer app = Application::create(
+        clock, cfg, Application::InitialDBMode::APP_DB_UPGRADE_EXISTING);
 
     app->getPersistentState().setState(PersistentState::kForceSCPOnNextLaunch,
                                        (set ? "true" : "false"));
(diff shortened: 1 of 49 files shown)


src/main/ApplicationUtils.cpp
```diff
@@ -25,6 +26,53 @@
...
+    LOG(INFO) << "Starting relay-mode stellar-core " << STELLAR_CORE_VERSION;
+    VirtualClock clock(VirtualClock::REAL_TIME);
+    Application::pointer app;
+    try
+    {
+        app = Application::create<RelayApplication>(
+            clock, cfg, Application::InitialDBMode::APP_DB_DONT_OPEN);
+
+        // Relays are never validating.
+        cfg.NODE_IS_VALIDATOR = false;
```
vogel: I think you may want to set that before creating application

src/main/CommandLine.cpp
```diff
@@ -940,6 +962,7 @@ handleCommandLine(int argc, char* const* argv)
          {"upgrade-db", "upgade database schema to current version",
           runUpgradeDB},
 #ifdef BUILD_TESTS
+         {"run-relay", "run relay-mode stellar-core node", runAsRelay},
```
vogel: I think that in future it may be better to have different NODE_TYPE options in configuration file (VALIDATOR, RELAY, NON-VALIDATOR?), instead of making that a new test-only command.

src/main/ApplicationUtils.cpp
graydon: Nice catch! Thanks, fixed (that was the biggest remaining bug!)

src/main/CommandLine.cpp
graydon: Possible. I'm going to not do this for now unless you feel really strongly; the initialization path is a bit different, we'd need to change config (currently use NODE_IS_VALIDATOR), and I'm not even sure we want to surface relay mode to users yet. Maybe in a followup PR if we decide to keep it / make it a supported mode.

vogel: No, I don't feel strong about it. That is just maybe-nice-to-have-in-future note.

YanaSneg23:
1

graydon:
Obsolete, @marta-lokhova is continuing this over in https://github.com/stellar/stellar-core/pull/2348 with a less-invasive approach.

PR #2216, opened 2019-08-07
Improve naming consistency of federated voting functions

marta-lokhova:
A bit of a nit, but I think some of the SCP function names are confusing. Reading the white paper, it looks like ballot protocol stages are: "accept prepared", "confirm prepared", "accept commit", "confirm commit". This change makes the code consistent with the whitepaper.


MonsieurNicolas:
r+ 3bf757e96bb82cf94a40b694d878d6458b8e7572

PR #2220, opened 2019-08-11
Fail loadgen when load is too high, remove Simulation dependency

### LINKED ISSUE #2199: Loadgen not working as expected
The built-in load generator used for testing has a few problems.

There is a design flow in that after txs get submitted it moves to a completely different phase  `waitTillComplete`, assuming that transactions were submitted succesfuly and just need to be processed by the network:
1. a big problem there is that if nothing was actually submitted (which can happen, see below), that phase will succeed
2. if some txs are dropped, there is no way it can recover and an observer of loadgen has no way of quantifying the amount of dropped transactions

A few more details:
1. in general, it doesn't handle when transactions get dropped by the validator's queue for whatever reason. The logic there that "just retries" is too simplistic to recover.
2. it doesn't handle `ADD_STATUS_TRY_AGAIN_LATER` (banned transactions). I suspect this should be handled the same way that we deal with 1 anyways

Here are my recommendations:
1. load generator should only succeed when work is complete. An observer (such as test automation), only needs to wait for `loadgen.run.complete` to be set without having to know how to check for completion. We leave the option of "timing out" to the observer (so "stopping" loadgen" should work in this situation)
2. We already have logic that uses the list of all accounts used in the simulation to sign and generate proper sequence numbers, we can expand on this to a process that can guarantee progress instead.

Potential updated way of generating load:
1. start simulation: setup world
    a. synchronize all source accounts used during the entire simulation run (`mAccounts`)
    b. compute the list of pairs `expected` `<simulation_account_ID=uint64, sequence_number>` of expected sequence numbers for each source account *at the end of the simulation*. For "create", it's the single account, something like (rounding) `current_seq_num+nbAccount/batchSize`, for payments we can assume round robin over the set of `mAccounts` so something like `current_seq_num+nTx/nAccounts`
    c. initialize the lists of accounts `done` and `backlog` to `empty`
2. load generation step (inject `txPerStep` transactions), loop is something like:
    a. generate `backlog` if needed (ie, `backlog` is empty)
         i. iterate over `expected`, remove accounts that already have the right sequence number; otherwise add to `backlog`
         ii. if `expected` is empty, simulation is "done"
         ii. shuffle `backlog`
    b. pick an account from `backlog`, generate and submit one transaction for that account
         i. the logic that we need here is something like we have right now: duplicate -> skip (ie, pick the next account from `backlog`) on error, synchronize account
         ii. NB: `generate` has to be fully deterministic based on source account and sequence number (this is already the case)
         iii. we have to break if no progress was made even after rebuilding the backlog (can happen with `create` if the account gets banned)

This may generate a few extra transactions at the end, but this should not really matter (as they would be rejected by the validator with `duplicate`). We get "retries" for free between steps.

marta-lokhova:
Resolves #2199

### DIFF
--- a/src/simulation/LoadGenerator.cpp
+++ b/src/simulation/LoadGenerator.cpp
@@ -36,20 +36,28 @@ namespace stellar
 using namespace std;
 using namespace txtest;
 
-// Units of load are is scheduled at 100ms intervals.
+// Units of load are scheduled at 100ms intervals.
 const uint32_t LoadGenerator::STEP_MSECS = 100;
-//
-const uint32_t LoadGenerator::TX_SUBMIT_MAX_TRIES = 1000;
+
+// If submission fails with txBAD_SEQ, attempt refreshing the account or
+// re-submitting a new payment
+const uint32_t LoadGenerator::TX_SUBMIT_MAX_TRIES = 10;
+
+// After successfully submitting desired load, wait a bit to let it get into the
+// ledger.
+const uint32_t LoadGenerator::TIMEOUT_NUM_LEDGERS = 20;
 
 LoadGenerator::LoadGenerator(Application& app)
-    : mMinBalance(0), mLastSecond(0), mApp(app), mTotalSubmitted(0)
+    : mMinBalance(0)
+    , mLastSecond(0)
+    , mApp(app)
+    , mTotalSubmitted(0)
+    , mLoadgenComplete(
+          mApp.getMetrics().NewMeter({"loadgen", "run", "complete"}, "run"))
+    , mLoadgenFail(
+          mApp.getMetrics().NewMeter({"loadgen", "run", "failed"}, "run"))
 {
     createRootAccount();
-}
-
-LoadGenerator::~LoadGenerator()
-{
-    clear();
 }
 
 void
@@ -92,12 +100,14 @@ LoadGenerator::getTxPerStep(uint32_t txRate)
 }
 
 void
-LoadGenerator::clear()
+LoadGenerator::reset()
 {
     mAccounts.clear();
     mRoot.reset();
     mStartTime.reset();
     mTotalSubmitted = 0;
+    mWaitTillCompleteForLedgers = 0;
+    mFailed = false;
 }
 
 // Schedule a callback to generateLoad() STEP_MSECS miliseconds from now.
@@ -106,6 +116,18 @@ LoadGenerator::scheduleLoadGeneration(bool isCreate, uint32_t nAccounts,
                                       uint32_t offset, uint32_t nTxs,
                                       uint32_t txRate, uint32_t batchSize)
 {
+    // If previously scheduled step of load did not succeed, fail this loadgen
+    // run.
+    if (mFailed)
+    {
+        CLOG(ERROR, "LoadGen") << "Load generation failed, ensure correct "
+                                  "number parameters are set and accounts are "
+                                  "created, or retry with smaller tx rate.";
+        mLoadgenFail.Mark();
+        reset();
+        return;
+    }
+
     if (!mLoadTimer)
     {
         mLoadTimer = std::make_unique<VirtualTimer>(mApp.getClock());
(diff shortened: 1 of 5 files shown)


src/simulation/LoadGenerator.cpp
```diff
@@ -409,13 +438,13 @@ LoadGenerator::findAccount(uint64_t accountId, uint32_t ledgerNum)
...
-        auto account = TestAccount{mApp, txtest::getAccount(name.c_str()), sn};
-        newAccountPtr = make_shared<TestAccount>(account);
+        auto name = "TestAccount-" + std::to_string(accountId);
+        newAccountPtr =
+            std::make_shared<TestAccount>(mApp, txtest::getAccount(name), sn);
 
         if (!loadAccount(newAccountPtr, mApp))
         {
-            std::runtime_error(
+            throw std::runtime_error(
```
MonsieurNicolas: wow, that was quite a miss :)

src/simulation/LoadGenerator.cpp
```diff
@@ -463,52 +492,73 @@ LoadGenerator::handleFailedSubmission(TestAccountPtr sourceAccount,
...
-        auto reloadRes = loadAccount(account, app);
-        // reload the account
-        if (!reloadRes || currentSeqNum != account->getLastSequenceNumber())
+        auto accountFromDB = *account;
+
+        auto reloadRes = loadAccount(accountFromDB, app);
+        // For account creation, reload accounts from the DB
+        // For payments, ensure that the sequence number matches expected
+        // seqnum. Timeout after 20 ledgers.
+        if (!reloadRes && isCreate)
```
MonsieurNicolas: code would be easier to read if it was just written with a different nesting:

```c++
if (isCreate)
{
    if (!reloadRes)
    {
          // not created yet
    }
}
else if (!reloadRes)
{
// what should happen here? 
}
else if (account->getLastSequenceNumber() !=
                accountFromDB.getLastSequenceNumber())
{
// not in sync
}
```
note that there is a case that I *think* should not happen (so we should fail the simulation in this case?)

src/simulation/LoadGenerator.cpp
```diff
@@ -463,52 +492,73 @@ LoadGenerator::handleFailedSubmission(TestAccountPtr sourceAccount,
...
-LoadGenerator::waitTillComplete()
+LoadGenerator::waitTillComplete(bool isCreate)
 {
     if (!mLoadTimer)
     {
         mLoadTimer = std::make_unique<VirtualTimer>(mApp.getClock());
     }
     vector<TestAccountPtr> inconsistencies;
-    inconsistencies = checkAccountSynced(mApp);
+    inconsistencies = checkAccountSynced(mApp, isCreate);
```
MonsieurNicolas: Do we have a problem where `checkAccountSynced` is blocking for too long because it rescans all of `mAccounts` instead of the ones that didn't reach their goal? We may want to open an issue for this: if we run a simulation for millions of accounts, it may cause problems.

src/simulation/LoadGenerator.h
```diff
@@ -29,20 +29,77 @@ class VirtualTimer;
 class LoadGenerator
 {
   public:
+    using TestAccountPtr = std::shared_ptr<TestAccount>;
     LoadGenerator(Application& app);
-    ~LoadGenerator();
-    void clear();
+    ~LoadGenerator() = default;
```
MonsieurNicolas: do we need the destructor at all here?

src/simulation/LoadGenerator.cpp
marta-lokhova: I didn't encounter problems of this kind during benchmarking runs, but opened an issue as it's worth taking a look: https://github.com/stellar/stellar-core/issues/2223

marta-lokhova:
Ok, comments should be addressed now (and changes squashed)

MonsieurNicolas:
r+ dc7248e88207cf30bff3a8a30fe1b3fc9e0e2d3f

PR #2225, opened 2019-08-14
Squash db schema version 10

### LINKED ISSUE #2200: squash sql schema changes up to schema version 10
Database schema version 10 was introduced in version 11.0.0 and now that all nodes are running versions more recent than this, we should "squash" all schema upgrades.

This would allow us to remove some of the confusing code in Database.cpp like https://github.com/stellar/stellar-core/blob/master/src/database/Database.cpp#L212

MonsieurNicolas:
# Description

Resolves #2200 

This PR consolidates database changes to the version of the schema introduced in version 11.0.0
(as the network has been running protocol 11 for a while, it is safe to assume that upgrades from versions 10.x or older don't need to be supported anymore)


src/ledger/LedgerTxnAccountSQL.cpp
```diff
@@ -530,15 +530,18 @@ LedgerTxnRoot::Impl::dropAccounts()
...
-           "homedomain      VARCHAR(32)  NOT NULL,"
-           "thresholds      TEXT         NOT NULL,"
-           "flags           INT          NOT NULL,"
-           "lastmodified    INT          NOT NULL"
+           "accountid          VARCHAR(56)  PRIMARY KEY,"
+           "balance            BIGINT       NOT NULL CHECK (balance >= 0),"
+           "buyingliabilities  BIGINT CHECK (buyingliabilities >= 0),"
+           "sellingliabilities BIGINT CHECK (sellingliabilities >= 0),"
+           "seqnum             BIGINT       NOT NULL,"
+           "numsubentries      INT          NOT NULL CHECK (numsubentries >= "
```
marta-lokhova: why the formatting change?

src/ledger/LedgerTxnAccountSQL.cpp
```diff
@@ -541,6 +473,7 @@ LedgerTxnRoot::Impl::dropAccounts()
            "homedomain         VARCHAR(32)  NOT NULL,"
            "thresholds         TEXT         NOT NULL,"
            "flags              INT          NOT NULL,"
+           "signers            TEXT,"
            "lastmodified       INT          NOT NULL"
            ");";
     mDatabase.getSession() << "CREATE TABLE signers"
```
marta-lokhova: I don't think we need this or the index creation code below.

src/database/Database.cpp
```diff
@@ -189,15 +189,11 @@ Database::applySchemaUpgrade(unsigned long vers)
...
-        // add tracking table information
-        mApp.getHerderPersistence().createQuorumTrackingTable(mSession);
-        break;
     default:
-        if (vers <= 9)
+        if (vers <= 10)
         {
             throw std::runtime_error(
-                "Database version is too old, must use at least 10");
+                "Database version is too old, must use at least 11");
```
marta-lokhova: but version 11 doesn't exist yet :) `applySchemaUpgrade` should just be a no-op if `vers == SCHEMA_VERSION`

src/ledger/LedgerTxnAccountSQL.cpp
MonsieurNicolas: column shift

src/ledger/LedgerTxnAccountSQL.cpp
MonsieurNicolas: good catch

src/database/Database.cpp
MonsieurNicolas: Easier to maintain like this; that said, let me take out the hardcoded 11

MonsieurNicolas:
updated @marta-lokhova

marta-lokhova:
r+ b36409a5bf0e6d40d0a5b458adebef29c2b05613

Issue #2123, opened 2019-05-30
Missing bucket recovery when user deletes bucket dir has bad diagnostic & recovery


graydon:
I've seen this happen multiple times with operators:

  - Using a container or some config that fails to mount BUCKET_DIR persistently (esp. during execution of new-db, which isn't especially obvious about the fact that it writes to BUCKET_DIR)
  - Or just deleting BUCKET_DIR because operator thinks it's transient
  - Node tries to start up, sees missing buckets
  - Last-gasp repair attempt tries to contact history to get missing buckets
  - User gets the impression history is damaged because it's trying to fetch a bucket that wasn't (and never should have been) published.
  - User wastes time trying to figure out what's wrong with the network, history archives, get commands, etc. etc. rather than being directed to correct diagnosis of the problem as "wrong lifetime of BUCKET_DIR"

I'm not entirely sure the right approach to issuing better diagnostics here would be. Maybe we can add a few special cases to the bucket-repair code. Like if the bucket dir is _empty_ that's a pretty clear sign that there's misconfiguration. Similarly if a large fraction of the bucket list is missing (say we have the genesis bucket but nothing else). Or perhaps it's possible to work out in reverse when a bucket shouldn't actually exist in a history archive (because it's not the input-to, shadow-of or output-of a checkpoint state) and avoid fetching it? Idk. This happens a lot though and current diagnostic story is bad and frustrating.

MonsieurNicolas:
Maybe this code (download missing buckets) should be removed? I mean the chances of being on a snapshot (where this would work) is low (as in 1/64 = 1.5%).

graydon:
This is happening more often than is likely due to operator error. I suspect we've got a bug where we're deleting unreferenced buckets too eagerly in exception cases, eg. when failing to connect to SQL.

I put some manual recovery instructions in another bug (here: https://github.com/stellar/stellar-core/issues/1395#issuecomment-511580240) but this needs to (a) stop happening altogether and (b) probably gain a new automatic recovery mode. Nobody should ever have to do those manual-recovery steps.

MonsieurNicolas:
So we need 2 issues then: one that tracks stopping to try to download missing buckets and the other that tracks the investigation so that we stop getting into that state in the first place. Right @graydon ?

graydon:
Opened #2202 and #2203 for fixes, having determined (I think) the root cause. Keeping this one open for "stopping trying to download missing buckets" since that's really the nature of the "bad diagnostic" here.

PR #2232, opened 2019-08-16
bug 2123: do not try to (confusingly) "repair" missing buckets, just fail

graydon:
As mentioned in #2123, the "repair" code is too clever for its own good: it's a sort of wishful-thinking about content addressed buckets that fails to recognize that it has _at best_ a one-in-64 chance of working, and would only ever work when a bucket went missing _after_ being published (and before being replaced with some other local one).

In practice this never happened: the only cases of bucket corruption or missing buckets we've seen in the field were fsync failures and people misconfiguring their BUCKET_DIR, and the "repair" work that kicks in is lost in the error logs, causing the user to totally misinterpret their stalled server as a broken history archive (because the work "can't find" the buckets it's looking for in the archive, and emits an error about that failure-to-download).

This PR just removes the attempt at repair. It is never going to work anyways, and it's better to just be honest and tell the user that with a clear exception. A corrupt bucket dir is actually corrupt and the best repair approach is to kill the node and fast-catch-up to a state before the failure, then let it replay forwards.


MonsieurNicolas:
r+ cbbbb1a5e99725bb6f7ae1b60dd4ceb2db43b553

PR #2233, opened 2019-08-16
Fix quorum checks (runtime and startup) for automatically generated quorums

### LINKED ISSUE #2209: Stellar-core new autoconfig code can't handle example config (Version 11.2.0)
In https://github.com/stellar/stellar-core/pull/2125 a new config was added docs/stellar-core_example_validators.cfg which, if you run stellar-core with it, fails assertion `isQuorumSetSane`.

This is because we defined quorum sets as sane only if they're 3 nesting levels or shallower, and the new auto-config code groups nodes by homeDomain before putting them into nesting levels. This makes for a 4-level nesting overall, not 3.

A trivial fix for this is to change the test in QuorumSanityChecker::checkSanity from `depth > 2` to `depth > 3` but I'm not sure if that's _actually_ the semantics we want.

MonsieurNicolas:
# Description

Resolves #2209 

Turns out that the validity check was not in the right place (too late), so tests never caught this.

We do want to increase the limit here as to be able to mix all quality levels.

### DIFF
--- a/docs/stellar-core_testnet.cfg
+++ b/docs/stellar-core_testnet.cfg
@@ -6,9 +6,12 @@ NETWORK_PASSPHRASE="Test SDF Network ; September 2015"
 DATABASE="sqlite3://stellar.db"
 
 # Stellar Testnet validators
+[[HOME_DOMAINS]]
+HOME_DOMAIN="testnet.stellar.org"
+QUALITY="HIGH"
+
 [[VALIDATORS]]
 NAME="sdftest1"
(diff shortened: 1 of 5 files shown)


graydon:
r+ 3efab57

MonsieurNicolas:
forgot to update some tests

MonsieurNicolas:
r+ aa45d9eaa4ece2d2d6f9b2bcbdc5898ba5d18080

Issue #2234, opened 2019-08-16
database credentials may get leaked into logs


MonsieurNicolas:
Looks like f2069efa41168372ee47747f8b7350c90becb37a doesn't deal with some config formats:

```
DATABASE="postgresql://dbname=stellar user=stellar password=thisshouldbesecret host=/var/run/postgresql/"
```

yields:

```
2019-08-16T18:18:29.203 GDKXE [Database INFO] Connecting to: postgresql://dbname=stellar user=stellar password=thisshouldbesecret host=/var/run/postgresql/
```

graydon:
Oh, we're being too narrow with the definition of a bareword. Postgres allows [anything non-whitespace](https://www.postgresql.org/docs/9.4/libpq-connect.html#LIBPQ-CONNSTRING). Easy fix.

PR #2237, opened 2019-08-16
Add a bit more explanation on the last test.

graydon:
Just a slightly longer explanation in code and comments for what bug #2234 was and how it was fixed. cc @marta-lokhova


marta-lokhova:
r+ c32cf09c84092fa03d3edf6249514840adf1a90c

PR #2239, opened 2019-08-18
Sort metrics by name in metrics.md

Subho13:
sort the metrics according to dictionary order of the metric name


MonsieurNicolas:
Thanks!

PR #2240, opened 2019-08-19
Fix exit code issue when pandoc is not installed

robertDurst:
# Description

If pandoc is not installed, pandoc's exit code is ignored by the `-` preceding it. However, the final `sed` command tries to reference the output file of `pandoc` and its exit code is not ignored.

Thus, this PR adds the `-` prefix to the final `sed`.

For anyone curious about `@`, `-`, and `+` prefixes: https://stackoverflow.com/questions/3477292/what-do-and-do-as-prefixes-to-recipe-lines-in-make

# Checklist
- [x] Compiles


MonsieurNicolas:
r+ 42190a04ad2395abb6233eb535015a9ea780b537

PR #2242, opened 2019-08-21
Conditional work fixes

marta-lokhova:
Reset condition in ConditionalWork once it's satisfied, as this work is intended to be used once


MonsieurNicolas:
r+ b0a4bdbf5fbeafbdb9f086aa10f4cdc01e73e2c0

PR #2246, opened 2019-08-22
Fuzzing docs and makefile update

robertDurst:
# Description

Following #2182 , this updates the **fuzzing.md** doc with what I have learned and the updates I have made. This also updates the shell script so that we generate a lot of inputs and then minimize them with `afl-cmin`.

### DIFF
--- a/docs/fuzzing.md
+++ b/docs/fuzzing.md
@@ -11,11 +11,11 @@ this road we go, so you're encouraged to give this a try.
 
 AFL is a black-box fuzzer -- it knows nothing much about the thing it's fuzzing
 -- that uses an evolutionary algorithm to breed malicious input under a bunch of
-heuristic mutations (bit flips and such). It thus uses a corpus of small inputs
-to the victim program, which it expands as it finds inputs that take the program
-into new portions of its control-flow space and/or trigger crashes. It writes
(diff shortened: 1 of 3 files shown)


docs/fuzzing.md
```diff
@@ -25,53 +25,77 @@ branch-tuple counts produces a control-flow "signature" for a run, which is used
...
 
 
 ## Building an instrumented stellar-core
 
-Start with a clean workspace, `make clean` or cleaner; then just do `./configure
---enable-afl && make` and make sure you have not enabled `asan` and `ccache`;
-the former is incompatible, and the latter doesn't interoperate with the
-compiler wrappers.
+Start with a clean workspace, `make clean` or cleaner; enable a few compilation flags **WHICH SHOULD
+NEVER BE ENABLED IN PRODUCTION**: `FUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION` and `AFL_LLVM_MODE`; and
```
MonsieurNicolas: are you sure about this? `FUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION` seems to be set automatically by afl.

As for `AFL_LLVM_MODE`, if it's not automatically set, we should just set it when configuring with `--enable-afl` in the configure script

PR #2248, opened 2019-08-22
Fix computation of overlay parameters

MonsieurNicolas:
# Description

This PR changes the way we compute overlay's parameters (how many connections are allowed in/out and how many can be pending):
the old  behavior was overly aggressive in trying to enforce the default of 500 pending connections.

Instead, I modified the behavior to bias towards enforcing how many established connections the instance has and only reduce those numbers if limits are causing those targets to not be met


marta-lokhova:
r+ a32f9206a09b3caba028743cfa513e1e80efebe6

PR #2252, opened 2019-08-26
Initial State Improvement for Transaction Fuzzer

robertDurst:
# Description

This PR seeks to expand the possible coverage and search area of the transaction fuzzer by introducing the following:
* elimination of op type restriction for initial corpus generation and operation application
* raise initial xlm balance per account and set to a fixed number so that we can easily setup an initial state with orders and trustlines
* initialize trustlines for four assets against all accounts except for source
* set up initial offers so that offer based operations may be valid with less complex fuzzed inputs

## Order Book Initial State

The initial state of the order book is based on the following, per asset structure (also described in a comment in the PR). 

Selling Asset → Buying Asset | Account | Amount | Numerator (n) | Denominator (d) | Converted Price (in terms of A)
-- | -- | -- | -- | -- | --
A → B | 1 | 1000 | 3 (B) | 2 (A) | 0.6
A → B | 2 (Issuer) | 5000 | 3 (B) | 2 (A) | 0.6
A → B | 3 | 10000 | 1 (B) | 1 (A) | 1.0
B → A | 4 (Issuer) | 1000 | 10 (A) | 9 (B) | 1.1
B → A | 5 | 5000 | 10 (A) | 9 (B) | 1.1
B → A | 1 | 10000 | 22 (A) | 7 (B) | 3.14 = π

Offers on Order Book Properties:
- Not crossed
- Multiple for some depth (2 “levels”)
- Offers from issuers
- Multiple offers at the same “level”
- Should be some prices that promote rounding
- Sell and buy from same account

# Checklist
- [x] Compiles

### DIFF
--- a/src/test/FuzzerImpl.cpp
+++ b/src/test/FuzzerImpl.cpp
@@ -24,6 +24,7 @@ namespace FuzzUtils
 {
 auto const FUZZER_MAX_OPERATIONS = 5;
 auto const INITIAL_LUMEN_AND_ASSET_BALANCE = 100000LL;
+auto const NUMBER_OF_ASSETS_TO_ISSUE = 4;
 
 // generates a public key such that it's comprised of bytes reading [0,0,...,i]
 PublicKey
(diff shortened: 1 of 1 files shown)


MonsieurNicolas:
needs to be cleaned up

PR #2262, opened 2019-08-30
Update admin.md

rice2000:
I think this will fix the validator quality diagram so that it appears correctly both on Github and on the stellar.org docs.


MonsieurNicolas:
I don't think this works: if you browse to your branch's version, the diagram is not visible

https://github.com/rice2000/stellar-core/blob/patch-3/docs/software/admin.md#automatic-quorum-set-generation

rice2000:
You're right.  I see what the problem is.  I'll work on a fix.

PR #2270, opened 2019-09-10
Windows handle fix part 3

MonsieurNicolas:
# Description

This PR is the latest attempt at fixing a race condition when spawning processes in Windows.

Previous attempts were #2253 and #1987

Third time should do it:

using `PROC_THREAD_ATTRIBUTE_HANDLE_LIST` requires to set `bInheritHandles =TRUE` in `CreateProcess` (so #2253 was wrong, and the pieces that were working were by pure luck) 
https://docs.microsoft.com/en-us/windows/win32/api/processthreadsapi/nf-processthreadsapi-updateprocthreadattribute

the problem was actually that #1987 didn't set `PROC_THREAD_ATTRIBUTE_HANDLE_LIST` when redirecting `stdout`, this was causing all inheritable handles to be inherited.

### DIFF
--- a/src/process/ProcessManagerImpl.cpp
+++ b/src/process/ProcessManagerImpl.cpp
@@ -28,6 +28,8 @@
 
 #ifdef _WIN32
 #include <Windows.h>
+
+#include "util/FileSystemException.h"
 #else
 #include <errno.h>
 #include <fcntl.h>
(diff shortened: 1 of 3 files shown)


graydon:
r+ dffc866

MonsieurNicolas:
@latobarita: retry

MonsieurNicolas:
(travis timed out talking to ubuntu servers)

PR #2276, opened 2019-09-17
Overlay logging improvement

MonsieurNicolas:
# Description

This PR:
* fixes a few logging issues in overlay
   * logging of uninitialized member variables during `Hello`   
   * missing context, useful when running tests that spin up multiple applications
* bump overlay version (allows to boot old core versions later on that can't even catchup to the network post protocol upgrade)
* adds a new property to connections to report how long a connection has been active (only used for reporting in this PR)


src/overlay/OverlayManagerImpl.cpp
```diff
@@ -201,10 +209,33 @@ OverlayManagerImpl::PeersList::acceptAuthenticatedPeer(Peer::pointer peer)
 
     CLOG(INFO, "Overlay") << "Non preferred " << mDirectionString
                           << " authenticated peer " << peer->toString()
-                          << " rejected because all available slots are taken.";
+                          << " rejected because all available slots are taken."
```
jonjove: I think these lines are out of order. Adding a function for this also would have avoided this issue.

src/overlay/OverlayManagerImpl.cpp
```diff
@@ -113,7 +114,8 @@ OverlayManagerImpl::PeersList::removePeer(Peer* peer)
     if (pendingIt != std::end(mPending))
     {
         CLOG(DEBUG, "Overlay") << "Dropping pending " << mDirectionString
-                               << " peer: " << peer->toString();
+                               << " peer: " << peer->toString() << " @"
```
jonjove: Did you intend for no space between "@" and the port? This is the case in every log line.

src/overlay/OverlayManagerImpl.cpp
```diff
@@ -142,7 +146,8 @@ OverlayManagerImpl::PeersList::moveToAuthenticated(Peer::pointer peer)
     {
         CLOG(WARNING, "Overlay")
             << "Trying to move non-pending " << mDirectionString << " peer "
-            << peer->toString() << " to authenticated list";
+            << peer->toString() << " to authenticated list"
```
jonjove: I think you meant to have "to authenticated list" after the port. In order to avoid issues like this, I think it would be better to use a function like `std::string Peer::toStringWithPort() const`.

src/overlay/OverlayManagerImpl.cpp
```diff
@@ -201,10 +209,33 @@ OverlayManagerImpl::PeersList::acceptAuthenticatedPeer(Peer::pointer peer)
...
+
+    if (Logging::logTrace("Overlay"))
+    {
+        CLOG(TRACE, "Overlay") << fmt::format(
+            "limit: {}, pending: {}, authenticated: {}", mMaxAuthenticatedCount,
+            mPending.size(), mAuthenticated.size());
+        std::stringstream pending, authenticated;
+        for (auto p : mPending)
+        {
+            pending << p->toString();
```
jonjove: Do you want to add the port here as well?

src/overlay/OverlayManagerImpl.cpp
MonsieurNicolas: yes, makes it simpler to grep

src/overlay/OverlayManagerImpl.cpp
MonsieurNicolas: no, the port is not related to `peer` but to the application, and I'd rather keep this at the end of the line to limit the confusion

src/overlay/OverlayManagerImpl.cpp
MonsieurNicolas: same thing: I do want this "@..." thing at the end of the line

src/overlay/OverlayManagerImpl.cpp
MonsieurNicolas: no: port is unrelated to `p` and there is already context from the `INFO` entry above

jonjove:
r+ 45cd965

MonsieurNicolas:
@latobarita: retry

MonsieurNicolas:
@latobarita: retry

MonsieurNicolas:
r+ 04dcaf5530e5a0cb5a3704b24f14bc9520cfcc61

MonsieurNicolas:
Issue was that I forgot to update one of the tests when I removed some of the constraints on `PeerRecord`.

MonsieurNicolas:
r+ 6556353b730846755a0b2d932ef332e0182f2065

PR #2278, opened 2019-09-18
Bug 2267 quorum intersection max scc topological order

### LINKED ISSUE #2267: Potentially incorrect quorum intersection check (Version: 11.4.0)
While implementing the quorum intersection check in Python in [stellar-observatory](https://github.com/andrenarchy/stellar-observatory) (as part of an SDF-supported academic research project) with @liamsi we noticed a potential issue in the corresponding Stellar Core implementation. It seems to be implemented incorrectly in the part where strongly connected components of the induced dependency/reachability graph are handled:

* In the [paper by Lachowski](https://arxiv.org/abs/1902.06493v1) the "greatest element" or "top" has a special meaning: it is meant with respect to the topological (reachability) order, i.e., the "top" component element cannot "reach" other components. See also the [discussion on PaperHive](https://paperhive.org/documents/items/ACdED1EekRhH?a=d:-_U0EJUs92Ng).
* In Core it seems to use the biggest component (i.e., by the number of elements), see [here](https://github.com/stellar/stellar-core/blob/27576172e99d89cbacfe6571f807a5e85746f618/src/herder/QuorumIntersectionCheckerImpl.cpp#L680).

This means that the algorithm would be looking at the wrong strongly connected component if there is a component that is bigger than the top component in the topological order, thus giving wrong results.

graydon:
# Description

Resolves #2267: (partially) order SCCs topologically / by reachability, take a maximally-reachable element in the partial order.

# Checklist
- [x] Compiles
- [x] Ran all tests

### DIFF
--- a/src/herder/QuorumIntersectionCheckerImpl.cpp
+++ b/src/herder/QuorumIntersectionCheckerImpl.cpp
@@ -8,7 +8,17 @@
 #include "util/Logging.h"
 #include "util/Math.h"
 
-namespace
+namespace stellar
+{
+std::shared_ptr<QuorumIntersectionChecker>
+QuorumIntersectionChecker::create(QuorumTracker::QuorumMap const& qmap,
+                                  Config const& cfg, bool quiet)
+{
+    using namespace quorum_intersection;
+    return std::make_shared<QuorumIntersectionCheckerImpl>(qmap, cfg, quiet);
+}
+
+namespace quorum_intersection
 {
 
 ////////////////////////////////////////////////////////////////////////////////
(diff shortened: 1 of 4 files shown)


src/herder/QuorumIntersectionCheckerImpl.cpp
```diff
@@ -670,14 +680,69 @@ QuorumIntersectionCheckerImpl::buildGraph(QuorumTracker::QuorumMap const& qmap)
...
 QuorumIntersectionCheckerImpl::buildSCCs()
 {
+    SCCReachabilityCalculator reach(mGraph);
+    reach.calculateReachability();
     mTSC.calculateSCCs();
     mMaxSCC.clear();
     for (auto const& scc : mTSC.mSCCs)
     {
-        if (scc.count() > mMaxSCC.count())
+        if (mMaxSCC.empty() || reach.canReach(mMaxSCC, scc))
```
Liamsi: Quick note: doesn't Tarjan give you the SCCs in the right (topological) order? So you could simply set maxSCC to the last component found (or actually first, I do not remember the order) instead of iterating here and computing the reachability?

graydon:
This is overcomplex -- see #2279 for simpler fix.

PR #2279, opened 2019-09-18
Bug 2267 quorum intersection pick any scc with a quorum

### LINKED ISSUE #2267: Potentially incorrect quorum intersection check (Version: 11.4.0)

graydon:
# Description

Resolves #2267 in a different way than #2278. Simpler: just pick the first quorum-containing SCC we find, and fail if there's more than one (as we were already doing). Avoids artificially picking an SCC with no quorums that happens to just be "big".

# Checklist
- [x] Compiles
- [x] Ran all tests

### DIFF
--- a/src/herder/QuorumIntersectionCheckerImpl.cpp
+++ b/src/herder/QuorumIntersectionCheckerImpl.cpp
@@ -181,15 +181,16 @@ MinQuorumEnumerator::pickSplitNode() const
 size_t
 MinQuorumEnumerator::maxCommit() const
 {
-    return (mQic.mMaxSCC.count() / 2) + 1;
+    return (mScanSCC.count() / 2) + 1;
 }
 
 MinQuorumEnumerator::MinQuorumEnumerator(
-    BitSet const& committed, BitSet const& remaining,
+    BitSet const& committed, BitSet const& remaining, BitSet const& scanSCC,
     QuorumIntersectionCheckerImpl const& qic)
     : mCommitted(committed)
     , mRemaining(remaining)
     , mPerimeter(committed | remaining)
+    , mScanSCC(scanSCC)
     , mQic(qic)
 {
 }
@@ -245,7 +246,7 @@ MinQuorumEnumerator::anyMinQuorumHasDisjointQuorum()
                     << "early exit 3.1: minimal quorum=" << committedQuorum;
             }
             mQic.mStats.mEarlyExit31s++;
-            return mQic.hasDisjointQuorum(committedQuorum);
+            return hasDisjointQuorum(committedQuorum);
         }
         if (mQic.mLogTrace)
         {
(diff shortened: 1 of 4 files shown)


graydon:
@fixxxedpoint is this more along the correct line of behaviour?

src/herder/QuorumIntersectionCheckerImpl.cpp
```diff
@@ -674,18 +674,19 @@ void
 QuorumIntersectionCheckerImpl::buildSCCs()
 {
     mTSC.calculateSCCs();
-    mMaxSCC.clear();
+    mScanSCC.clear();
     for (auto const& scc : mTSC.mSCCs)
```
fixxxedpoint: I would just remove that loop together with mScanSCC class variable and merge it with that main loop in networkEnjoysQuorumIntersection(). Name suggests it is only responsible for building SCCs and the whole thing will be also more easier to understand (as we talked before, you don't need to search for that greatest element). It is not a big deal but current version also iterates all SCCs twice.

graydon: @fixxxedpoint Well, mScanSCC is used in a couple other helper methods throughout the implementation, I can't easily eliminate it entirely without turning it into a parameter I have to pass everywhere; but I can merge the setup loops as you suggest. Will do so. Otherwise I just wanted to double check: style issues aside, this at least looks _correct_ to you?

Thanks again for all the input and hand-holding; I just want to make sure I'm faithfully transplanting your work here :)

graydon: (Update: actually making the variable local / sinking it into a member into the helper class was relatively low-impact; patch updated with the results!)

MonsieurNicolas:
r+ c2d9b9e564ed4520f205a0b1cd6016652a99152f

PR #2287, opened 2019-09-23
Use all bucket hashes from HAS instead of relying on StateSnapshot fo…

### LINKED ISSUE #2286: some buckets are leaked from the publish queue
Looks like we're not deleting certain buckets that are referenced by the publish queue.

Repro: run a node, wait for publish to complete, expected number of buckets in the buckets folder should be 31.

marta-lokhova:
Resolves #2286


MonsieurNicolas:
r+ 78ae9a61a2dcf4d5801b588cb8b211de4c83bd6d

PR #2291, opened 2019-09-26
Introduce PROBATION state in overlay

### LINKED ISSUE #2097: flaky test "connecting to saturated nodes"
Ran into this a few times:
```
"connecting to saturated nodes" overlay/test/OverlayTests.cpp:962
local:2/21/100%/12.0s 1	./usr/include/c++/6/bits/unique_ptr.h:295: typename std::add_lvalue_reference<_Tp>::type std::unique_ptr<_Tp, _Dp>::operator*() const [with _Tp = Catch::AssertionInfo; _Dp = std::default_delete<Catch::AssertionInfo>; typename std::add_lvalue_reference<_Tp>::type = Catch::AssertionInfo&]: Assertion 'get() != pointer()' failed.
```

compiled with g++ with extra-checks enabled.

MonsieurNicolas:
# Description

Resolves #2097 

this closes a possible race condition when peers connect to each other but one of
them is at capacity (therefore dropping the connection).

The approach is to add an intermediate state to the connection `GOT_AUTH_PROBATION` that tells overlay to allow that connection to go through *not* counting towards the total number of connections (ie, it's still counted as a "pending" connection).

When the grace period elapses, if the peer is still connected, it gets promoted to `GOT_AUTH_ACTIVE` and if there are other connections in probation in the other direction, they get dropped.

### DIFF
--- a/docs/software/admin.md
+++ b/docs/software/admin.md
@@ -686,6 +686,7 @@ This list is the result of both inbound connections from other peers and outboun
            "elapsed" : 6,
            "id" : "sdf1",
            "olver" : 5,
+           "probation" : false,
            "ver" : "v9.1.0"
         }
      ],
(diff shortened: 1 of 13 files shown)


MonsieurNicolas:
I don't think we should do this, we should find some other solution. this is too complicated

PR #2292, opened 2019-09-26
Publish throttles catchup

marta-lokhova:
Update offline catchup flow to avoid creating huge backlog of checkpoints to publish

### DIFF
--- a/src/catchup/ApplyCheckpointWork.cpp
+++ b/src/catchup/ApplyCheckpointWork.cpp
@@ -31,7 +31,7 @@ ApplyCheckpointWork::ApplyCheckpointWork(Application& app,
                     fmt::format("{}-{}", range.mFirst, range.mLast),
                 BasicWork::RETRY_NEVER)
     , mDownloadDir(downloadDir)
-    , mCheckpointRange(range)
+    , mLedgerRange(range)
     , mCheckpoint(
           app.getHistoryManager().checkpointContainingLedger(range.mFirst))
     , mApplyLedgerSuccess(app.getMetrics().NewMeter(
@@ -43,12 +43,12 @@ ApplyCheckpointWork::ApplyCheckpointWork(Application& app,
     auto const& hm = mApp.getHistoryManager();
     auto low = std::max(LedgerManager::GENESIS_LEDGER_SEQ,
                         hm.prevCheckpointLedger(mCheckpoint));
-    if (mCheckpointRange.mFirst != low)
+    if (mLedgerRange.mFirst != low)
     {
         throw std::runtime_error(
             "Ledger range start must be aligned with checkpoint start");
     }
-    if (mCheckpointRange.mLast > mCheckpoint)
+    if (mLedgerRange.mLast > mCheckpoint)
     {
         throw std::runtime_error(
             "Ledger range must span at most 1 checkpoint worth of ledgers");
@@ -267,7 +267,7 @@ ApplyCheckpointWork::onRun()
 
         auto result = applyHistoryOfSingleLedger();
         auto const& lm = mApp.getLedgerManager();
-        auto done = lm.getLastClosedLedgerNum() == mCheckpointRange.mLast;
+        auto done = lm.getLastClosedLedgerNum() == mLedgerRange.mLast;
 
         if (done)
         {
--- a/src/catchup/ApplyCheckpointWork.h
+++ b/src/catchup/ApplyCheckpointWork.h
@@ -45,7 +45,7 @@ struct LedgerHeaderHistoryEntry;
 class ApplyCheckpointWork : public BasicWork
 {
     TmpDir const& mDownloadDir;
-    LedgerRange const mCheckpointRange;
+    LedgerRange const mLedgerRange;
     uint32_t const mCheckpoint;
 
     XDRInputFileStream mHdrIn;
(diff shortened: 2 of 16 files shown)


src/catchup/ApplyCheckpointWork.cpp
```diff
@@ -266,9 +302,22 @@ ApplyCheckpointWork::onRun()
         }
 
         auto result = applyHistoryOfSingleLedger();
-        auto const& lm = mApp.getLedgerManager();
-        auto done = lm.getLastClosedLedgerNum() == mCheckpointRange.mLast;
+        auto lcl = mApp.getLedgerManager().getLastClosedLedgerNum();
+
+        // In case we just closed a ledger that unblocked publishing of the
```
MonsieurNicolas: why do we wait at all if we're done applying the checkpoint?

src/catchup/ApplyCheckpointWork.cpp
```diff
@@ -266,9 +302,22 @@ ApplyCheckpointWork::onRun()
...
+        // previous checkpoint, wait for publish to complete before
+        // continuing application
+        if (result && mWaitForPublish)
+        {
+            auto ledger = hm.prevCheckpointLedger(lcl);
+            if (ledger == lcl && hm.publishQueueLength() > 0)
```
MonsieurNicolas: isn't the condition on the size of the publish queue enough?

src/catchup/ApplyCheckpointWork.h
```diff
@@ -56,14 +56,16 @@ class ApplyCheckpointWork : public BasicWork
     medida::Meter& mApplyLedgerFailure;
 
     bool mFilesOpen{false};
+    bool const mWaitForPublish;
+    std::unique_ptr<VirtualTimer> mPublishTimer;
```
MonsieurNicolas: technically it's really `mWaitForPublishQueueTimer`

src/catchup/ApplyCheckpointWork.cpp
```diff
@@ -137,6 +140,13 @@ ApplyCheckpointWork::applyHistoryOfSingleLedger()
         return false;
     }
 
+    if (header.ledgerSeq > mLedgerRange.mLast)
```
MonsieurNicolas: unless you have a bug, this should never happen

MonsieurNicolas: Oh I see: that's because you're calling `applyHistoryOfSingleLedger` when coming back from a `wait for publish` even though you may be `done` (see my other comment)

src/catchup/ApplyCheckpointWork.cpp
```diff
@@ -258,6 +268,32 @@ ApplyCheckpointWork::applyHistoryOfSingleLedger()
 BasicWork::State
 ApplyCheckpointWork::onRun()
 {
+    auto& hm = mApp.getHistoryManager();
+    if (mPublishTimer)
+    {
+        if (hm.publishQueueLength() > 0)
```
MonsieurNicolas: you should use constants instead of `0` in both locations, and take the opportunity to define the two thresholds (0 is too aggressive, I think):
* increase the size of the queue needed to trigger a wait - let's say 32
* define the size of the queue to unblock applying ledgers - let's say 16 for now

src/catchup/ApplyCheckpointWork.cpp
MonsieurNicolas: now that I think of it we should move this code into a precondition and use `ConditionalWork` to wrap `ApplyCheckpointWork`

marta-lokhova: > isn't the condition on the size of the publish queue enough?

The sequence of events is as follows: a ledger that is the last one in a checkpoint is closed, triggering snapshot being queued for publish. Then `ResolveSnapshotWork` waits for the next ledger (aka first ledger in the next checkpoint) to be closed in order to proceed. So the "publish queue is not empty" condition is not enough here, as we need to also close the next ledger to unblock `ResolveSnapshotWork` before going into `WORK_WAITING` state.

> now that I think of it we should move this code into a precondition and use ConditionalWork to wrap ApplyCheckpointWork

Right, I think initially this was a bit tricky to do, since I did not allow any drift in catchup vs publish. But with the min/max queue size to trigger wait/applying that you suggested in the other comment, this should be cleaner to do with ConditionalWork (since we're waiting for checkpoint far enough in the past to complete publishing, so I _think_ it's not going to have the `ResolveSnapshotWork` problem I mentioned above)

src/catchup/ApplyCheckpointWork.cpp
marta-lokhova: Seems like the way ConditionalWork is implemented, it'd require changing state inside the lambda to know whether to wait or apply when publish queue size it in between 16 and 32 (which I think is quite error-prone and also doesn't comply with ConditionFn contract). I added the lower boundary of 16 for now to preserve the const-ness of lambda, what do you think?

MonsieurNicolas: What do you mean by changing state inside the lambda @marta-lokhova ? You mean the conditional is stateful? I don't think this is a problem: we do this all the time when we capture a pointer to a class.

MonsieurNicolas:
r+ 4cec6afeaa4df30391ab9fed14472c4861514dc9

PR #2293, opened 2019-09-26
Remove assert

marta-lokhova:
i don't think we can actually guarantee this


MonsieurNicolas:
r+ 1b0eaa53edfdec3becd745adb94a48aefd9b0665

PR #2302, opened 2019-10-02
Remove per-key prefetch metrics

jonjove:
This PR removes per-key prefetch metrics. They aren't currently accessible from any external interface, and the way they were implemented isn't compatible with incremental prefetching (coming in a future PR).

# Checklist
- [x] Compiles
- [x] Ran all tests

### DIFF
--- a/src/ledger/LedgerTxn.cpp
+++ b/src/ledger/LedgerTxn.cpp
@@ -1461,7 +1461,6 @@ LedgerTxnRoot::Impl::commitChild(EntryIterator iter, LedgerTxnConsistency cons)
     // Clearing the cache does not throw
     mBestOffersCache.clear();
     mEntryCache.clear();
-    mPrefetchMetrics.clear();
 
     // std::unique_ptr<...>::reset does not throw
     mTransaction.reset();
@@ -1878,11 +1877,6 @@ LedgerTxnRoot::Impl::getNewestVersion(LedgerKey const& key) const
     {
         return getFromEntryCache(key);
     }
-    else
-    {
-        auto& metrics = mPrefetchMetrics[key];
-        ++metrics.misses;
-    }
 
     std::shared_ptr<LedgerEntry const> entry;
     try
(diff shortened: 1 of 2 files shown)


jonjove:
In the PR description, I mentioned that the way the per-key metrics was implemented isn't compatible with incremental prefetching. @marta-lokhova inquired about how the incremental prefetching will work and what is causing the incompatibility. 

#### Incremental Prefetching
The idea behind incremental prefetching (in this context) is to prefetch the accounts and trustlines associated with offers that are fetched as part of `LedgerTxnRoot::getBestOffer`. This makes sense because we only load best offers in the context of offer book operations (eg. `ManageOffer` and `PathPayment`) and in these cases we will always need the account and trustlines for every offer that is processed during `convertWithOffers`. We already fetch multiple best offers at once in order to reduce database round-trip time. Prefetching in these cases is an extension of this optimization.

#### Why is this implementation incompatible?
The existing implementation marks `mPrefetchMetrics` whenever something has been prefetched. The current implementation is very careful to never request-for-prefetch the same key more than once (eg. we prefetch source accounts once, then transaction data other than source accounts once). With incremental prefetching as described above, it is possible for the same key to be requested-for-prefetch more than once. This can happen if two offers far apart in the best offers order relation have the same `sellerID` for example. If the result of the previous prefetch has already been evicted from the entry cache, then the second request-for-prefetch will trigger an actual prefetch and will then throw when it finds that `mPrefetchMetrics` has already been marked. With small entry cache size (particularly with size 0), this is likely to occur. 

#### Why remove this instead of making it compatible?
It's not externally accessible; maintaining it requires resources. If we want to access this information at some future date, we can make it accessible and compatible with whatever prefetching solution exists at the time.

marta-lokhova:
That makes sense, thank you for the detailed comment. To add to the last point, I think having these stats is valuable, but we can re-add those later when we have an idea of how the prefetching logic would look like.

marta-lokhova:
r+ cf52fb9e115bba38c7d8ad6f6ba997146a875a2e

PR #2303, opened 2019-10-04
Reset shared_future in FutureBucket when it's ready

### LINKED ISSUE #2298: Some buckets are leaked
Looks like fixing #2286 was not enough, and we still have bucket number constantly fluctuating (with total number of files being over expected 31). Checking the logs, the number of buckets referenced by the bucketlist, publishqueue and LCL is correct, so it must be failing the `refcount == 1` check in `forgetUnreferencedBuckets`

marta-lokhova:
This should resolve #2298 -- some standard library implementations (libstdc++) store task lambdas in the shared state, which we used to keep alive for a while (specifically, until level spill which could take months for low level buckets). This change explicitly resets shared_future once it's done.

The second change is a minor update to catchup, to stop referencing buckets earlier (currently we drop bucket references at the very end of a potentially long CatchupWork)

### DIFF
--- a/src/bucket/FutureBucket.cpp
+++ b/src/bucket/FutureBucket.cpp
@@ -65,16 +65,12 @@ FutureBucket::setLiveOutput(std::shared_ptr<Bucket> output)
 {
     mState = FB_LIVE_OUTPUT;
     mOutputBucketHash = binToHex(output->getHash());
-    // Given an output bucket, fake-up a promise for it connected to
-    // the future so that it can be immediately retrieved.
-    std::promise<std::shared_ptr<Bucket>> promise;
-    mOutputBucket = promise.get_future().share();
-    promise.set_value(output);
+    mOutputBucket = output;
     checkState();
 }
 
 static void
-checkHashEq(std::shared_ptr<Bucket> b, std::string const& h)
+checkHashEq(std::shared_ptr<Bucket> const& b, std::string const& h)
 {
     assert(b->getHash() == hexToBin256(h));
 }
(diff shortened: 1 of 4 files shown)


cocoi241116:
Oke thanks br

src/bucket/FutureBucket.h
```diff
@@ -59,7 +59,8 @@ class FutureBucket
     std::shared_ptr<Bucket> mInputCurrBucket;
     std::shared_ptr<Bucket> mInputSnapBucket;
     std::vector<std::shared_ptr<Bucket>> mInputShadowBuckets;
-    std::shared_future<std::shared_ptr<Bucket>> mOutputBucket;
+    std::shared_ptr<Bucket> mOutputBucket;
+    std::shared_future<std::shared_ptr<Bucket>> mOutputBucketFuture;
```
MonsieurNicolas: no test update?

marta-lokhova: We do not test the internals of FutureBucket, only its public interface, which functionally did not change.

src/bucket/FutureBucket.cpp
```diff
@@ -299,7 +315,7 @@ FutureBucket::startMerge(Application& app, uint32_t maxProtocolVersion,
 
     assert(curr);
     assert(snap);
-    assert(!mOutputBucket.valid());
+    assert(!mOutputBucketFuture.valid());
```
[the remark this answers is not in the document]
marta-lokhova: Good idea, fixed (plus one other small simplification). If the change looks ok, I can squash the commits.

src/bucket/FutureBucket.cpp
```diff
@@ -233,24 +240,29 @@ FutureBucket::resolve()
...
+        // Explicitly reset shared_future to ensure destruction of shared state.
+        // Some compilers store packaged_task lambdas in the shared state,
+        // keeping its captures alive as long as the future is alive.
+        mOutputBucketFuture = std::shared_future<std::shared_ptr<Bucket>>();
     }
 
     if (mOutputBucketHash.empty())
     {
-        mOutputBucketHash = binToHex(bucket->getHash());
+        mOutputBucketHash = binToHex(mOutputBucket->getHash());
```
MonsieurNicolas: `mOutputBucket` is only set above, right?

So why not setting `mOutputBucketHash` in the same place?
Actually maybe `resolve` should just start with `if (LIVE_OUTPUT) { return mOutputBucket; } ` right now the code is written in an almost idempotent way, which doesn't make much sense

MonsieurNicolas:
r+ d50d4a69b796a1e37c55458521741503972fc432

PR #2307, opened 2019-10-07
Fast best offers

jonjove:
# Description

This PR completely overhauls the implementation details of `LedgerTxn::loadBestOffer` and `LedgerTxn::getBestOffer` by leveraging an incrementally maintained in-memory order book and a newly recognized invariant satisfied by `loadBestOffer`. This is built on top of #2306, so it should not be merged until that PR is merged.

# Checklist
- [x] Compiles
- [x] Ran all tests

### DIFF
--- a/src/ledger/LedgerTxn.cpp
+++ b/src/ledger/LedgerTxn.cpp
@@ -510,68 +510,135 @@ LedgerTxn::Impl::getAllOffers()
 }
 
 std::shared_ptr<LedgerEntry const>
-LedgerTxn::getBestOffer(Asset const& buying, Asset const& selling,
-                        std::unordered_set<LedgerKey>& exclude)
+LedgerTxn::getBestOffer(Asset const& buying, Asset const& selling)
 {
-    return getImpl()->getBestOffer(buying, selling, exclude);
+    return getImpl()->getBestOffer(buying, selling);
 }
 
 std::shared_ptr<LedgerEntry const>
-LedgerTxn::Impl::getBestOffer(Asset const& buying, Asset const& selling,
-                              std::unordered_set<LedgerKey>& exclude)
+LedgerTxn::Impl::getBestOffer(Asset const& buying, Asset const& selling)
 {
-    auto end = mEntry.cend();
-    auto bestOfferIter = end;
-    if (exclude.size() + mEntry.size() > exclude.bucket_count())
+    if (!mActive.empty())
     {
-        // taking a best guess as a starting point for how big exclude can get
-        // this avoids rehashing when processing transactions after a lot of
-        // changes already occured
-        exclude.reserve(exclude.size() + mEntry.size());
+        throw std::runtime_error("active entries when getting best offer");
     }
-    for (auto iter = mEntry.cbegin(); iter != end; ++iter)
+
+    AssetPair const assets{buying, selling};
+
+    std::shared_ptr<LedgerEntry const> selfBest;
+    auto mobIter = mMultiOrderBook.find(assets);
+    if (mobIter != mMultiOrderBook.end())
     {
-        auto const& key = iter->first;
-        auto const& entry = iter->second;
-        if (key.type() != OFFER)
+        auto const& offers = mobIter->second;
+        if (!offers.empty())
         {
-            continue;
-        }
-
-        if (!exclude.insert(key).second)
-        {
-            continue;
-        }
-
-        if (!(entry && entry->data.offer().buying == buying &&
-              entry->data.offer().selling == selling))
-        {
-            continue;
-        }
-
-        if ((bestOfferIter == end) ||
-            isBetterOffer(*entry, *bestOfferIter->second))
-        {
-            bestOfferIter = iter;
+            auto entryIter = mEntry.find(offers.begin()->second);
+            if (entryIter == mEntry.end() || !entryIter->second)
+            {
+                throw std::runtime_error("invalid order book state");
+            }
+            selfBest = std::make_shared<LedgerEntry const>(*entryIter->second);
         }
     }
 
-    std::shared_ptr<LedgerEntry const> bestOffer;
-    if (bestOfferIter != end)
+    auto parentBest = mParent.getBestOffer(buying, selling);
+    // If parentBest is not nullptr, then the parent contains at least one offer
+    // with the specified asset pair. If selfBest is not nullptr and parentBest
+    // is not better than selfBest, then we must return selfBest because
+    // parentBest cannot be the best offer. Otherwise, either selfBest is
+    // nullptr or parentBest is better than selfBest. In the first case, self
+    // has no recorded offer with the specified asset pair. In the second case,
+    // self has a recorded offer with the specified asset pair but it is worse
+    // than parentBest. Either way, we should return parentBest if it is not
+    // recorded in self. If it is recorded in self, then one of the following is
+    // true:
+    // - it has been erased, in which case it is not the best offer
+    // - it has been modified to a different asset pair or a worse price, in
+    //   which case it is not the best offer
+    // - it has been modified to a price no worse than its price in the parent,
+    //   but this possibility is excluded by the fact that either selfBest is
+    //   nullptr or parentBest is better than selfBest
+    // If parentBest is recorded in self, then we must ask parent for the next
+    // best offer that is worse than parentBest and repeat.
+    while (parentBest && (!selfBest || isBetterOffer(*parentBest, *selfBest)))
     {
-        bestOffer = std::make_shared<LedgerEntry const>(*bestOfferIter->second);
+        if (mEntry.find(LedgerEntryKey(*parentBest)) == mEntry.end())
+        {
+            return parentBest;
+        }
+        parentBest = mParent.getBestOffer(*parentBest);
+    }
+    return selfBest;
+}
+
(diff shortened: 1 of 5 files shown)


src/ledger/LedgerTxn.cpp
```diff
@@ -372,6 +400,17 @@ LedgerTxn::Impl::deactivate(LedgerKey const& key)
     {
         throw std::runtime_error("Key is not active");
     }
+
+    auto entryIter = mEntry.find(key);
```
MonsieurNicolas: could we maybe rewrite those code changes into a call to a method `updateOrderBook` or sorts that can by itself reason about `mActive` and `mEntry` vs `mOrderBook`. That way it's easier to understand which invariant are supposed to be met. Right now it's pretty hard to see if we've missed an update.
Basically, if we could just call a method every time we update `mActive` or `mEntry` that ensures that the in memory order book has the right state the change would be much easier to validate.

src/ledger/LedgerTxn.cpp
```diff
@@ -1089,9 +1184,19 @@ LedgerTxn::Impl::loadBestOffer(LedgerTxn& self, Asset const& buying,
...
-    std::unordered_set<LedgerKey> exclude;
-    auto le = getBestOffer(buying, selling, exclude);
-    return le ? load(self, LedgerEntryKey(*le)) : LedgerTxnEntry();
+    auto le = getBestOffer(buying, selling);
+    auto res = le ? load(self, LedgerEntryKey(*le)) : LedgerTxnEntry();
+
+    try
+    {
+        AssetPair assets{buying, selling};
+        mWorstBestOffer[assets] = le;
```
MonsieurNicolas: I don't know if we can give `mWorstBestOffer` this value: `le` may have as value some random entry from `mEntry` (via `mOrderBook`).

If this happens, this will cause `loadBestOffer` to return duplicates, where as the previous implementation would never do this as whatever was in `mEntry` was always excluded.

I don't know if we should detect this. I guess the reason is that `getBestOffer` has a different behavior now in this situation. I don't think the current code cares (we call `getBestOffer` right after creating a `LedgerTxn` scope)

src/ledger/LedgerTxn.cpp
```diff
@@ -1874,27 +2065,28 @@ findIncludedOffer(std::list<LedgerEntry>::const_iterator iter,
...
 
     size_t const BATCH_SIZE = 5;
     while (!res && !cached->allLoaded)
     {
         std::list<LedgerEntry>::const_iterator newOfferIter;
         try
         {
-            newOfferIter = loadBestOffers(offers, buying, selling, BATCH_SIZE,
-                                          offers.size());
+            newOfferIter = loadBestOffers(offers, oe.buying, oe.selling,
```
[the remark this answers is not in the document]
jonjove: I've already improved the SQL query considerably in a future commit, taking advantage of the index and the knowledge of `worseThan`. But it is built on top of this PR, so I'll publish it once we've worked out the issues in this one.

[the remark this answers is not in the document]
MonsieurNicolas: cool. I realized later that we can't actually use `worseThan` in this version as it may cause unwanted gaps in `offers`. I'll watch for the change in the later PR that makes changes to this area.

src/ledger/LedgerTxn.cpp
```diff
@@ -1225,6 +1496,192 @@ LedgerTxn::Impl::maybeUpdateLastModifiedThenInvokeThenSeal(
...
+    if (mobIterPrev != mMultiOrderBook.end() &&
+        obIterPrev != mobIterPrev->second.end())
+    {
+        mobIterPrev->second.erase(obIterPrev);
+    }
+}
+
+void
+LedgerTxn::Impl::updateWorstBestOffer(
+    AssetPair const& assets, std::shared_ptr<OfferDescriptor const> worstBest)
```
MonsieurNicolas: comments say `entry` so rename `worstBest` into `entry`

src/ledger/LedgerTxn.cpp
```diff
@@ -1225,6 +1496,192 @@ LedgerTxn::Impl::maybeUpdateLastModifiedThenInvokeThenSeal(
...
+{
+    // Update mWorstBestOffer if
+    // - assets is currently not in mWorstBestOffer
+    // - entry is nullptr, because if any offers exist they would be returned
+    //   before nullptr
+    // - assets is in mWorstBestOffer, entry is not nullptr, and *entry is
+    //   worse than mWorstBestOffer[assets] because then mWorstBestOffer[assets]
+    //   would be returned before *entry
+    auto iter = mWorstBestOffer.find(assets);
+    if (iter == mWorstBestOffer.end() || !worstBest ||
```
MonsieurNicolas: per definition of "worse than", should the condition be:
```
if (iter == mWorstBestOffer.end() || !entry ||
        !iter->second || isBetterOffer(*iter->second, *entry))
```

MonsieurNicolas: I think current code doesn't override `null` because of this, which I think is a (performance only) bug

src/ledger/LedgerTxn.cpp
```diff
@@ -271,19 +352,67 @@ LedgerTxn::Impl::commitChild(EntryIterator iter, LedgerTxnConsistency cons)
...
-                mEntry[key] = nullptr;
+            {
+                updateEntry(key, nullptr);
             }
         }
+
+        // We will show that the following update procedure leaves the parent
+        // worst best offer map in a correct state.
+        //
+        // Fix an asset pair P in the child worst best offer map, and let V be
```
MonsieurNicolas: It would be clearer/shorter to use terminology in the context of "this" as "parent" can be confusing.

Maybe use the proposed notation/definition that I proposed too?

```
        // By definition, child shadows all offers <=? V from its parents (including
        // this LedgerTxn), as a consequence:
        // (1) After the commit, this LedgerTxn will shadow all offers <=? V
        //     from its parents.
        // We need to update the worst offer map to reflect this fact.
        //
        // There are three possibilities prior to the commit:
        // 1. P is not in this LedgerTxn worst best offer map. Because of (1),
        //    V can be associated to this LedgerTxn worst best offer map
        // 2. P is in the parent worst best offer map with associated value W,
        //    and W <=? V. ...
```

src/ledger/LedgerTxnImpl.h
```diff
@@ -158,6 +165,33 @@ class LedgerTxn::Impl
...
+    // mMultiOrderbook is an in-memory representation of the order book that
+    // contains an entry if and only if it is recorded in this LedgerTxn and is
+    // not active. It is grouped by asset pair, and for each asset pair all
+    // entries are sorted according to the better offer relation.
+    MultiOrderBook mMultiOrderBook;
+
+    typedef std::unordered_map<
+        AssetPair, std::shared_ptr<OfferDescriptor const>, AssetPairHash>
+        WorstBestOfferMap;
+    // In the following definition, we say that an offer A is worse than a
```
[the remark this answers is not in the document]
MonsieurNicolas: Maybe `mShadowedOfferThreshold` ?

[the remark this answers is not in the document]
MonsieurNicolas: Actually, I think we should define "A worst than B" between pointer-to-offer

I think it needs to look like this:
```
    // In the following definition, we say that pointer-to-offer A is worse
    // than pointer-to-offer B if A && (!B || isBetterOffer(*B, A))
    // noted A >? B
    // NB: any non null offer A satisfies A >? null
```

We can then change `updateWorstBestOffer` to properly use `A >? B`, right now there is some weird logic in there that doesn't map to this definition (ie, I would like  `updateWorstBestOffer` to update the map if the argument is worst than what is stored - or - insert if the map didn't have anything)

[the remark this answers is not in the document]
jonjove: Introduced notation (not exactly what you proposed, but similar) to the definitions to increase clarity in the proofs. This was an excellent suggestion!

src/ledger/LedgerTxn.cpp
jonjove: Updated to use better notation, and also switched the naming convention to "parent", "self", and "child" where child is being committed into self.

src/ledger/LedgerTxnImpl.h
```diff
@@ -183,13 +183,32 @@ class LedgerTxn::Impl
...
+    // We say an offer-descriptor A is worse than an offer-descriptor B if
+    //     A.price > B.price || (A.price == B.price && A.offerID > B.offerID)
+    // We write this as A > B, and write !(A > B) as B <= A to denote that B is
+    // not worse than A.
+    //
+    // We say a pointer-to-offer-descriptor A is worse than a
+    // pointer-to-offer-descriptor B if
+    //     B && (!A || *A > *B)
+    // We again write this as A > B, and write !(A > B) as B <= A to denote that
+    // B is not worse than A. That nullptr < &B for any offer-descriptor B is
```
MonsieurNicolas: you mean `That nullptr > &B`

jonjove: Yes, typo. Fixed.

MonsieurNicolas:
r+ 3b0fd5e612436441029bf3b1721eef2c73b698d3

MonsieurNicolas:
@latobarita: retry

PR #2308, opened 2019-10-08
Check apply-state validity only once per bucket-apply.

graydon:
Small change to ensure we don't re-check HAS validity on every crank while applying level 10. We discussed this last week, not a huge deal but worth fixing.


MonsieurNicolas:
r+ 9355c66c54c839ff04dedf2e44dbc579248a41cb

PR #2309, opened 2019-10-08
Fix multiple keys referencing the same merge

### LINKED ISSUE #2299: Same merges produce different mergeKeys, causing merge races
It looks like maxProtocolVersion is used to create MergeKey in the merge re-attachment map. During normal operation in addBatch it uses current protocol version, but when FutureBucket calls makeLive is over-approximates it to Config::CURRENT_LEDGER_PROTOCOL_VERSION, so I think same merges could produce different keys in the merge map, if supported protocol version differs from current protocol version. I think this may cause redoing some of the merges, since we won't be able to retrieve finished/running merge due to differing keys.

marta-lokhova:
This should resolve #2299, as we don't rely on `maxProtocolVersion` for merge uniqueness anymore.

### DIFF
--- a/src/bucket/Bucket.cpp
+++ b/src/bucket/Bucket.cpp
@@ -632,8 +632,7 @@ Bucket::merge(BucketManager& bucketManager, uint32_t maxProtocolVersion,
     {
         bucketManager.incrMergeCounters(mc);
     }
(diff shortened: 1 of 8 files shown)


MonsieurNicolas:
r+ 7fecb9af8d223a00a622a82a5e5e01753fa4204c

PR #2315, opened 2019-10-15
Fix missing information from TransactionMeta in older protocol versions

### LINKED ISSUE #2217: Transaction meta is missing one time signer changes in older versions of the protocol
When `removeUsedOneTimeSignerKeys` was added in 
https://github.com/stellar/stellar-core/commit/840787d13777cc8cd78fff5ada5828037129fe08#diff-94d5ac6726dabe26918718d13fd44b9cR540

the meta format was not updated to reflect those changes (this would have required a "post transaction" meta of sorts).

This was later fixed in protocol 10 (presigned transactions changes are tracked in `txChanges` as they are done before applying operations):
https://github.com/stellar/stellar-core/issues/1464#issuecomment-361773464

Proposed fix:
in order to generate meta that contains all changes, a simple update to pre-protocol 10 ledgers would be to modify the meta of the last operation to include the one time signer change (NB: this is a merge of 2 meta).
It's a bit ugly as technically speaking the change doesn't belong to the last operation, but it seems to be the simplest we can do without adding "post tx" meta.

jonjove:
Resolves #2217 

This PR introduces a new version (version 2) of `TransactionMeta` and allows `stellar-core` to be configured to generate either version 1 or version 2.

# Checklist
- [x] Compiles
- [x] Ran all tests

### DIFF
--- a/docs/stellar-core_example.cfg
+++ b/docs/stellar-core_example.cfg
@@ -378,6 +378,12 @@ MAXIMUM_LEDGER_CLOSETIME_DRIFT=50
 # you want to make that trade.
 DISABLE_XDR_FSYNC=false
 
+# SUPPORTED_META_VERSION (1 or 2) defaults to 1.
+# If set to 1, stellar-core will emit TransactionMeta as TransactionMetaV1. If
+# set to 2, stellar-core will emit TransactionMeta as TransactionMetaV2. Do not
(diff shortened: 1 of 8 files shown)


MonsieurNicolas:
r+ b299b3a458a15f592352c67d4da69baa6e8fbb6a

PR #2319, opened 2019-10-19
Allow passing archive name in command-line catchup

### LINKED ISSUE #2311: Command line catchup should allow passing in an archive
This is quite a simple change, but will be beneficial for archive monitoring work we're doing.

marta-lokhova:
Resolves #2311


MonsieurNicolas:
Cool. Curious to understand the trade off you made @marta-lokhova :
In this proposal, you first classed the ability to "stick" to a given archive for catchup.
A simpler solution would be to filter out archives from the configuration file that would not require making changes to any of the logic (leaving room for different filtering where the filtered set is not a single archive)

MonsieurNicolas:
r+ 012168fa695a6d7ba465577b0aed9ea360b42429

PR #2320, opened 2019-10-21
Remove pandoc dependency

### LINKED ISSUE #2301: remove pandoc dependency
Right now we have a pandoc dependency to generate man pages, but:
* that pandoc dependency pulls a lot of packages (a significant subset of the internet it seems)
* those man pages are pretty much useless

We can instead just keep a hardcoded dummy man page (or remove it completely if it doesn't break things).

MarcPow:
# Description

Resolves #2301 

# Checklist
- [. ] Reviewed the [contributing](https://github.com/stellar/stellar-core/blob/master/CONTRIBUTING.md#submitting-changes) document
- [ .] Rebased on top of master (no merge commits)
- [. ] Compiles
- [ ] Ran all tests


.devcontainer/Dockerfile
```diff
@@ -0,0 +1,70 @@
...
+# Install common compilation tools
+RUN sudo apt-get -y install git build-essential pkg-config autoconf automake libtool bison flex libpq-dev parallel
+
+# Update compiler tools
+RUN sudo apt-get -y install gcc-6 g++-6 cpp-6 libstdc++6
+ENV CC=gcc-6
+ENV CXX=g++-6
+
+# Install pandoc (to build manpages)
+RUN sudo apt-get -y install pandoc
```
MarcPow: Note that pandoc must remain in the dev environment until it is also removed from xdrpp.

.devcontainer/README.md
```diff
@@ -0,0 +1,50 @@
...
+* This feature pre-configures a dev environment with all the tools you'll need to build stellar-core
+
+
+## Pre-requisites: What do I need to use it?
+* Install [Visual Studio Code](https://code.visualstudio.com/Download) for your platform
+* Install the [Remote Development Extension Pack](https://marketplace.visualstudio.com/items?itemName=ms-vscode-remote.vscode-remote-extensionpack)
+* Install [Docker](https://www.docker.com) for your platform
+  * If you are installing Docker on Windows, please make sure your settings have enabled Linux, rather than Windows containers.
+* Special note for Windows users:
+  * This feature mounts your git repository into the container so that you can use it from Linux.
```
MonsieurNicolas: Windows: it seems that the git working folder is all borked because of line ending differences between platforms. Not sure what the best way around this (assuming the shared folder should be usable from both sides)

.devcontainer/README.md
```diff
@@ -0,0 +1,50 @@
...
+- Type `./autogen.sh`.
+- Type `./configure`   
+- Type `make` or `make -j<N>` (where `<N>` is the number of parallel builds, a number less than the number of CPU cores available, e.g. `make -j3`)
+- Type `su vscode` to switch to a non-root user for test execution purposes.   (Tests that rely on postgresql will not run under root)
+- Type `make check` to run tests.
+- Type `make install` to install.
+
+## Known issues
+* Test failures on Windows:
+  * bucket/test/BucketManagerTests.cpp:1310 will fail
```
MonsieurNicolas: workaround is to add `thisConfig.DISABLE_XDR_FSYNC = true;` in `src/test/test.cpp` right after `thisConfig.UNSAFE_QUORUM = true;`

MonsieurNicolas: (we could also add a configure flag, but that seems overkill)

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,31 @@
...
+	"dockerFile": "Dockerfile",
+	// The optional 'runArgs' property can be used to specify additional runtime arguments.
+	"runArgs": [
+		// Uncomment the line if you will use a ptrace-based debugger like C++, Go, and Rust.
+	    "--cap-add=SYS_PTRACE", "--security-opt", "seccomp=unconfined",
+
+		// Uncomment the next line to use a non-root user. On Linux, this will prevent
+		// new files getting created as root, but you may need to update the USER_UID
+		// and USER_GID in .devcontainer/Dockerfile to match your user if not 1000.
+		// "-u", "vscode"
```
MonsieurNicolas: why not enable the `vscode` user by default as it allows to run tests?

docs/stellar-core.1
```diff
@@ -0,0 +1,389 @@
+.\" Automatically generated by Pandoc 1.16.0.2
```
MonsieurNicolas: I would not bother with this and just have a minimalist file that doesn't need to be maintained

MonsieurNicolas: I *think* you can actually remove pandoc from the image as it's a non fatal error for xdrpp

MonsieurNicolas:
The replacement PR that included your commits was merged. Thanks again!

PR #2327, opened 2019-10-30
Remove quorum set cache

### LINKED ISSUE #509: Deal with when LRU pushes out things that SCP still needs

MonsieurNicolas:
# Description

Part 1 (qset) to fix #509

This PR replaces the cache used to keep track of quorum sets with a deterministic data structure.

This allows to make (close to) optimal use of memory when it comes to quorum sets.

A similar change will be done for TxSet in a subsequent PR (wanted to first test the water with QSets)


MonsieurNicolas:
replaced by #2369

PR #2340, opened 2019-11-12
Fast best offers improvements

jonjove:
This PR adds some incremental improvements on top of #2307, which are largely focused on the interaction between stellar-core and the database. Do not merge until #2307 has been merged.

# Checklist
- [x] Compiles
- [x] Ran all tests

### DIFF
--- a/src/ledger/LedgerTxn.cpp
+++ b/src/ledger/LedgerTxn.cpp
@@ -11,6 +11,7 @@
 #include "ledger/LedgerTxnEntry.h"
 #include "ledger/LedgerTxnHeader.h"
 #include "ledger/LedgerTxnImpl.h"
+#include "transactions/TransactionUtils.h"
 #include "util/GlobalChecks.h"
 #include "util/XDROperators.h"
 #include "util/types.h"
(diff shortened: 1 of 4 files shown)


src/ledger/LedgerTxnOfferSQL.cpp
```diff
@@ -65,27 +65,81 @@ LedgerTxnRoot::Impl::loadAllOffers() const
...
+    // here. We are far from this limit anyway.
+    if (worseThan.offerID == INT64_MAX)
+    {
+        throw std::runtime_error("maximum offerID encountered");
+    }
+
     // price is an approximation of the actual n/d (truncated math, 15 digits)
     // ordering by offerid gives precendence to older offers for fairness
-    sql += " ORDER BY price, offerid LIMIT :n OFFSET :o";
+    std::string sql =
```
MonsieurNicolas: nit: don't use `SELECT *` (use explicit columns)

src/ledger/LedgerTxnOfferSQL.cpp
```diff
@@ -65,27 +65,81 @@ LedgerTxnRoot::Impl::loadAllOffers() const
...
+        "WITH r1 AS (SELECT * FROM offers "
+        "WHERE sellingasset = :v1 AND buyingasset = :v2 "
```
MonsieurNicolas: this seems to be a very complicated query...

can't we just do something like:
```
                      "WHERE sellingasset = :v1 AND buyingasset = :v2 " AND price >= :wp
                      "ORDER BY price, offerid LIMIT :n OFFSET :o";
```
where `wp` binds to `worseThanPrice`
and `o` binds to the number of offers in `offers` with a price equal to `worseThanPrice`

src/ledger/LedgerTxnOfferSQL.cpp
jonjove: Good idea. Changed the query to only use `SELECT *` when querying a table that is part of the `WITH` (meaning `r1` and `r2`) because at that point we already have explicit columns. This matches what we do in other queries.

src/ledger/LedgerTxn.cpp
```diff
@@ -2293,18 +2293,14 @@ LedgerTxnRoot::getBestOffer(Asset const& buying, Asset const& selling,
...
-    {
-        if (isBetterOffer(worseThan, *iter))
-        {
-            return std::make_shared<LedgerEntry const>(*iter);
-        }
-    }
-    return {};
+    iter = std::upper_bound(
+        iter, end, worseThan,
+        (bool (*)(OfferDescriptor const&, LedgerEntry const&))isBetterOffer);
```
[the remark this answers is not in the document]
jonjove: Done.

src/ledger/LedgerTxn.cpp
```diff
@@ -2317,9 +2317,12 @@ LedgerTxnRoot::Impl::getBestOffer(Asset const& buying, Asset const& selling,
 
     auto res = findIncludedOffer(offers.cbegin(), offers.cend(), worseThan);
 
-    size_t const BATCH_SIZE = 5;
+    size_t const MAX_BATCH_SIZE = 1024;
     while (!res && !cached->allLoaded)
     {
+        size_t const BATCH_SIZE =
+            std::min(MAX_BATCH_SIZE, std::max<size_t>(1, offers.size()));
```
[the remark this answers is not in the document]
jonjove: This is a good point, I don't really know what a good value for this is. The old value of 5 seems reasonable, so I will change to that.

MonsieurNicolas:
r+ 231519434e0a8c4f3e44f6a61560ced4343901de

PR #2358, opened 2019-11-25
bump overlay min version to 10

MonsieurNicolas:
# Description

This PR bumps the minimum version for overlay to 10 (as the network is now running version 12.0.0 and above that have overlay version 10, this is safe to do)


graydon:
r+ c22af66

Issue #2226, opened 2019-08-14
Consider evaluating a mode to generate meta only


MonsieurNicolas:
(starting a discussion here, this may be totally stupid)

To speed up rebuilding data for Horizon, we may benefit from a special replay mode to only rebuild tx meta.

I am thinking something relatively simple like:
1. keep ledger information in memory (or in a local sqlite database)
    * disable any durability guarantees (for on disk database)
2. disable buckets and their merges (of course publish to history would also be disabled)
3. emit meta either in the database (when on disk) or if there is an in-memory ledger, use a different database for the meta (on disk).

Would that be useful? I don't know if people need just the meta or the meta + history (for a full node)

graydon:
Update: hacked up a version with a "not even sqlite in memory" LedgerTxnRoot, using plain std::unordered_maps, and can replay 100k ledgers in 19 minutes. A little more RAM used but still in the ballpark of 2GB. There are still several hotspots to attend to (signature verification, txn history logging that still hits the sqlite-in-memory db). I've got some other stuff to do this afternoon but I'll come back to it.

graydon:
Removing more points of sqlite contact, touching some bucket-manager mutexes, log-level lookup and such gets down to 15 minutes. Signature verification is proving a little tricky to disable, looking into how the simulation stuff works. Also getting some tricky profiles because we're running thousands of curl and gzip subprocesses which is collectively taking >50% of on-core time, but it's hard to tell whether all the kernel work and user work associated with that is actually managing to parallelize across cores or whether we're hurting the main thread's ability.

graydon:
Down to 11 minutes with signature checking disabled. Interesting note: it's _much_ faster for older segments of history, due to their having far fewer ops and txs per ledger. Only 3 minutes to replay 100k ledgers back at ledger 15000000. So the first few years of history will probably replay 3-4x as fast as the most recent year.

graydon:
Added a (very crude) TxMeta-streaming XDROutputFileStream, causes only a 0.4% slowdown, generates 4GB of XDR for 100k ledgers. I was expecting we'd have to do this on a background thread but it doesn't look like it's expensive enough to even matter.

ire-and-curses:
Loving these updates!

graydon:
At the level of detail this bug discusses, this is done. File subsequent bugs for specific issues with the implementation.

PR #2361, opened 2019-12-06
Missing piece for bug #2226, need newdb in --replay-in-memory

graydon:
Silly mistake -- in the process of merging the command into a flag, I broke the flag. --replay-in-memory needs to imply newdb, otherwise we crash on startup with "no ledger".


MonsieurNicolas:
r+ 66c430395a69e7e8d548b996e11d569048c30168

MonsieurNicolas:
@latobarita: retry

graydon:
@latobarita: retry

PR #2362, opened 2019-12-07
Missing piece for bug #2226, METADATA_OUTPUT_STREAM should deny NODE_…

graydon:
Another small missing piece. We deny NODE_IS_VALIDATOR for non-DB mode, but also ought to deny it any time METADATA_OUTPUT_STREAM is set.


MonsieurNicolas:
ah yes you're right. At first I thought it would be OK to let people experiment with this, but they should experiment with non validating nodes instead, otherwise we're leaving a foot gun

MonsieurNicolas:
r+ 6f1da86385cc772a78cfb4fd03ce8aee66afaacf

MonsieurNicolas:
r+ c075b38443452089e9d81e4adaf0da5f28f59a89

graydon:
r+ c075b38443452089e9d81e4adaf0da5f28f59a89

PR #2366, opened 2019-12-12
Make the order of applying changes more clear

bartekn:
I found edited part to be confusing. Maybe because of indentation I understood that one should process fee and then meta of each tx one after another, however you should process fee changes of ALL txs first before processing meta changes.


MonsieurNicolas:
The indentation means "meta associated with processing the transaction set", and fees are part of this so this change as it is makes things even more confusing.

Maybe "(fees are processed before everything else)" should be replaced by "(fees are processed before applying transactions, see below)"?

MonsieurNicolas:
closing this PR: if you have a different suggestion, please re-open a PR and this time do not create a branch here

PR #2376, opened 2020-01-08
bug 2318: print LCL on startup

### LINKED ISSUE #2318: print LCL on startup
When we start up we must always print the LCL as soon as possible -- before we even attempt to check for missing buckets. When buckets are missing we currently crash before printing LCL which makes it challenging for operators to know which LCL to repair the node's buckets directory to (they have to run an SQL query).

graydon:
# Description

Resolves #2318

This just improves the startup logging a bit (from the perspective of a node operator) to print out information about the LCL at multiple fallible stages, especially "before trying to restore buckets". The idea here is that the operator can learn enough from looking at a crash log on node with corrupt buckets to know which ledger the node is trying to start up at, so the operator can go fetch the correct buckets (eg. from an archive).

# Checklist
- [x] Compiles
- [x] Ran all tests

### DIFF
--- a/src/ledger/LedgerManagerImpl.cpp
+++ b/src/ledger/LedgerManagerImpl.cpp
@@ -268,11 +268,13 @@ LedgerManagerImpl::loadLastKnownLedger(
 
     if (lastLedger.empty())
     {
-        throw std::runtime_error("No ledger in the DB");
+        throw std::runtime_error(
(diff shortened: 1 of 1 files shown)


MonsieurNicolas:
r+ 89272532c02b4fa2d3ee50d9f6e06720cbbd6ebb

PR #2377, opened 2020-01-09
Bump xdrpp to take fix for static destructor fiasco

graydon:
Just takes a change to xdrc to make it not-crash in a case it used to crash in.


MonsieurNicolas:
r+ 0c249eb2e3153695bd586a97a8f5a59c27ecb6ae

MonsieurNicolas:
@latobarita: retry

PR #2378, opened 2020-01-09
Use simple "C" collation on varchar / text index columns.

graydon:
# Description

This is a more-or-less free perf win on postgresql: switch text-typed columns that participate in indexing (eg. primary keys) from default collation (which uses strcoll_l language-aware collation) to "C" byte-level collation. Cuts livenet bucket-apply / fast-catchup time from 5 minutes to 4 minutes on my workstation.

Perhaps surprisingly, the ALTER TABLE bit during schema upgrade doesn't take especially long, only a few seconds. I believe it causes a REINDEX which does something like a sequential page-by-page rebuild of the index btrees, the largest of which are like 700MB; but it's also possibly the case that it notices there's no actual moving-of-data to do, everything's already in the correct order? Anyway it does not seem like it's operationally dangerous to be doing as part of a normal schema upgrade.

I'll do some timings of tx latency tomorrow -- bucket-apply is probably the worst possible case, so I expect a less noticable win -- but I _expect_ this should be neutral-to-positive for basically all cases.

# Checklist
- [x] Compiles
- [x] Ran all tests

### DIFF
--- a/src/database/Database.cpp
+++ b/src/database/Database.cpp
@@ -57,7 +57,7 @@ bool Database::gDriversRegistered = false;
 
 // smallest schema version supported
 static unsigned long const MIN_SCHEMA_VERSION = 9;
-static unsigned long const SCHEMA_VERSION = 11;
+static unsigned long const SCHEMA_VERSION = 12;
 
 // These should always match our compiled version precisely, since we are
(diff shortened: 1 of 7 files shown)


MonsieurNicolas:
r+ 3324a201bd061012f8d91386e246bb8b1f8c0a57

PR #2382, opened 2020-01-11
Update default testing behavior to actually be pseudo random by default

MonsieurNicolas:
This PR updates Catch to the latest version (which doesn't ignore certain sections as reported by @jonjove )

It also changes the default behavior for random seeds:
* it now uses a single default seed even by default
   * the previous behavior made it impossible to use a command line seed when running into an issue
   * this also allows to always report the seed to use in order to repro issues
* now it uses time by default instead of a mix of seeds
   * this increases the chances that people actually run into issues

### DIFF
--- a/lib/catch.hpp
+++ b/lib/catch.hpp
@@ -1,6 +1,6 @@
 /*
- *  Catch v2.7.0
- *  Generated: 2019-03-07 21:34:30.252164
+ *  Catch v2.11.1
+ *  Generated: 2019-12-28 21:22:11.930976
  *  ----------------------------------------------------------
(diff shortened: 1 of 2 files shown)


MonsieurNicolas:
@graydon updated with your suggestion (well making the seed by default stable for 24 hours)

graydon:
r+ 6692c3a

graydon:
r+ 7469cd2

MonsieurNicolas:
forgot to run clang-format during the last update

MonsieurNicolas:
r+ ba6bca3c7ee957849f927dcd849c20e56cefbb3c

PR #2383, opened 2020-01-11
ensure that operation level meta is cleared during an internal error

MonsieurNicolas:
This PR fixes an issue in the meta generation when encountering `txINTERNAL_ERROR`:
for a transaction with more than one operation, the code was keeping meta for operations that succeeded before the operation that triggered the internal error, even though the actual changes were never committed.

Thanks @bartekn for the report

### DIFF
--- a/src/test/TxTests.cpp
+++ b/src/test/TxTests.cpp
@@ -171,9 +171,10 @@ applyCheck(TransactionFramePtr tx, Application& app, bool checkSeqNum)
     bool res = false;
     {
         LedgerTxn ltxTx(ltx);
+        TransactionMeta tm(2);
         try
         {
-            res = tx->apply(app, ltxTx);
+            res = tx->apply(app, ltxTx, tm);
         }
         catch (...)
         {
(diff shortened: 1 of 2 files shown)


jonjove:
While this change looks correct, I don't think it is the best solution. I think it would be better to no-throw swap the meta into place as the last step of the function for two reasons:
1. It's the idiomatic way to write exception-safe code
2. Makes it extremely clear that all code paths must reach that point in order to change the meta

As an example of why my proposed approach would be better, consider the following segment of `TransactionFrame::applyOperations`
https://github.com/stellar/stellar-core/blob/c5ca2a3167fb2b8c64310587ffad0899adddf83a/src/transactions/TransactionFrame.cpp#L634-L640
The implementation says it should never happen, but we have code there to handle it. Did it happen? I have no idea, but if it did your change would still permit the wrong meta to be generated. My proposed change would generate no meta in this case, as expected. A future maintainer could similarly add other fast-fail paths, not recognizing that it would lead to incorrect meta generation.

MonsieurNicolas:
Yeah I was going for the minimal change but your suggestion cleans up the code a little bit, thanks for the suggestion @jonjove

jonjove:
r+ affb766

PR #2390, opened 2020-01-17
Improve logs and metrics

marta-lokhova:
Main issue that I've encountered while running network simulations is our logs being too verbose but not very detailed at the same time. This PR attempts to make a few improvements (such as tracking life cycle of a txset, and consolidating info that used to take multiple logs in different places into a single log entry).

Also added a new metric to record duration of individual items being fetched: problem with `scp.fetch.duration` was that if we had outliers that for some reason got stuck, and then cleaned up, we never surfaced them. The ItemFetcher code is kind of messy (like we don't actually delete trackers right away, so I _think_ there might be some strange edge case where PendingEnvelopes reuses the same tracker object to fetch an item), but we can at least surface the outliers with this metric.

### DIFF
--- a/src/overlay/Tracker.cpp
+++ b/src/overlay/Tracker.cpp
@@ -28,6 +28,7 @@ Tracker::Tracker(Application& app, Hash const& hash, AskPeer& askPeer)
     , mItemHash(hash)
     , mTryNextPeer(app.getMetrics().NewMeter(
           {"overlay", "item-fetcher", "next-peer"}, "item-fetcher"))
+    , mFetchTime("fetch-" + hexAbbrev(hash), LogSlowExecution::Mode::MANUAL)
 {
     assert(mAskPeer);
 }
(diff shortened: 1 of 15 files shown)


src/overlay/Tracker.cpp
```diff
@@ -28,6 +28,9 @@ Tracker::Tracker(Application& app, Hash const& hash, AskPeer& askPeer)
     , mItemHash(hash)
     , mTryNextPeer(app.getMetrics().NewMeter(
           {"overlay", "item-fetcher", "next-peer"}, "item-fetcher"))
+    , mFetchItemDuration(
```
MonsieurNicolas: let's make this
* type aware, ie `overlay.fetch.txset` and `overlay.fetch.qset`
* track successful downloads only (instead of including cancels) - a new method needs to be added for this

marta-lokhova: updated

MonsieurNicolas:
r+ 48977d466767f4cc32e6c7192407eca3f5ca2f14

PR #2392, opened 2020-01-22
fix broken test: don't use wrapped value if we don't need to

MonsieurNicolas:
(values are not known yet by SCP, so wrapping fails)

broken by ea32447b469f0e501a3c318420901b17b6d4364a


MonsieurNicolas:
r+ 9de903190257d2c28510da7404b70c004f1c4df8

PR #2394, opened 2020-01-22
Add dev container docker settings

### LINKED ISSUE #2301: remove pandoc dependency

MonsieurNicolas:
This PR replaces https://github.com/stellar/stellar-core/pull/2320

Resolves #2301 

changes compared to #2320 
* use `vscode` user instead of defaulting to `root`
* actually remove `pandoc` dependency
* added better instructions on how to use it on Windows


MonsieurNicolas:
r+ 8ef5bc440c184a882c4d4d8af29ca80364dff0fb

MarcPow:
:thumbsup:

PR #2399, opened 2020-01-23
refresh pending envelopes metrics

MonsieurNicolas:
broken by 88de5b743a9f96dd9587b0daa491e62bc80c5986

metrics have to be updated every time we change something


marta-lokhova:
r+ ca0cde692f7096274bc40af0474468806b58a26e

PR #2401, opened 2020-01-23
SCP: ensure that we really call nominatingValue with all new valid values

MonsieurNicolas:
Fixes a crash introduced by ea32447b469f0e501a3c318420901b17b6d4364a
itself caused by 88f8f4c1aa01c33a56cd67debf5d3151d70c5248

Crash would trigger with logging enabled at DEBUG or more verbose in the `Herder` partition


marta-lokhova:
r+ 60e06411b6cdec8d645fe34275b99595f3b35781

Issue #2404, opened 2020-01-24
Global test configs sometimes cause collisions 


marta-lokhova:
## Issue Description
Looks like we sometimes have test config collisions on latest master. Specifically, if different tests have the same instance number and on-disk database, they'll get mapped to the same config.

## Steps to Reproduce
run `stellar-core test "[db]"`, observe "sqlite MVCC test" fail

## Expected Result
"database on-disk smoketest" should use a different tmp folder than "sqlite MVCC test"

## Actual Result
"database on-disk smoketest" and "sqlite MVCC test"  end up using the same tmp folder due to the same config.

MonsieurNicolas:
Can you clarify here on the broader issue not fixed in #2405?
I think that what you ran into is that some (buggy) tests are not setting themselves up properly (normally they're supposed to do things like "new-db" to delete any stale data on disk).

marta-lokhova:
do you mean that it is expected for many tests to re-use the same test folder?

MonsieurNicolas:
yes tests are supposed to be resilient to stale data present in their temp folder.
In theory, we could have all tests created under a common folder, that we'd delete at the beginning of every test but I suspect this would add some significant overhead (most of the time, nothing gets created in there).
Maybe worth trying? I opened a PR that should make that change even easier as we now create all test folders under the same root folder.

marta-lokhova:
Closing: we haven't seen this issue in a long time.

PR #2405, opened 2020-01-24
fix tests

marta-lokhova:
This fixes currently failing tests by ensuring we re-create the database if it exists. I also opened https://github.com/stellar/stellar-core/issues/2404 which is the cause of the issue


MonsieurNicolas:
in general I would expect tests to, in addition to set things up properly (what you're doing here), to also cleanup after themselves (which doesn't seem to be the case). So I would suggest adding the `drop table` at the end of the test as well.

marta-lokhova:
ok, updated

MonsieurNicolas:
r+ f06168d49d55b69e4e0f83da4eaa8a9c7572cfd6

PR #2406, opened 2020-01-28
Clarify location of folders and logging when testing

MonsieurNicolas:
There was some confusion in #2404 about the organization of the test folders.

This PR doesn't functionally do much other than add an extra hint in the temp folder name containing the instance number and makes it obvious that there is one test folder per base instance - but that there is only one log file per stellar-core instance.


marta-lokhova:
r+ ddeda6946bbfbee6e52b2143f81d5eadd1be5575

PR #2407, opened 2020-01-28
absorb medida fix to exp-decay reservoir (thus: percentiles) in metrics

graydon:
Absorb change to medida submodule that fixes bogus exponential decay reservoir (and thus: percentiles of most metrics).

See https://github.com/stellar/medida/pull/14


graydon:
r+ 67783b4

PR #2408, opened 2020-01-28
IWYU bucket/* and related fallout

graydon:
An initial run + manual application of the suggestions of [IWYU ("include what you use")](https://include-what-you-use.org/) on the bucket directory.

IWYU follows a simple rule that does cause more includes (typically) to need to be listed explicitly in .cpp files, but the goal is to cut down the cross-inclusion of .h files, thus limiting the number of compilation units that get rebuilt when you touch any given file. You can't _entirely_ do it directory-at-a-time but you can do "a set of files plus the transitive effects in others", which is not _so_ bad if you try to go bottom-up.

Posting for discussion but also, if doing this (even in pieces) is agreeable, ~a request to merge this one piece: it'll likely bitrot pretty quickly~. I am happy to do a few of these and/or show others how to do it and/or post the IWYU output I got from my own run here so you can just work from what it says to do. Its suggestions are mostly correct, if occasionally stylistically off a bit.

EDIT: actually I want to work on automating this a little more before we merge it. But I'd appreciate any initial thoughts as it'll be a big bulk-edit in any case, and there's no point doing it if everyone hates it.

### DIFF
--- a/src/bucket/Bucket.cpp
+++ b/src/bucket/Bucket.cpp
@@ -8,25 +8,22 @@
 #include "util/asio.h"
 #include "bucket/Bucket.h"
 #include "bucket/BucketApplicator.h"
-#include "bucket/BucketList.h"
 #include "bucket/BucketManager.h"
 #include "bucket/BucketOutputIterator.h"
 #include "bucket/LedgerCmp.h"
 #include "bucket/MergeKey.h"
-#include "crypto/Hex.h"
-#include "crypto/Random.h"
-#include "crypto/SHA.h"
-#include "database/Database.h"
 #include "lib/util/format.h"
 #include "main/Application.h"
+#include "main/Config.h"
 #include "medida/timer.h"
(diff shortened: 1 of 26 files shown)


MonsieurNicolas:
That seems reasonable to do.

A few questions:
* this seems quite tricky to test reliably when it comes to system headers:
   * from what I remember clang MacOS includes different things than clang Linux? I think `<algorithm>` comes to mind
   * what about libstd++ vs libc++ ?
   * what about different versions of the headers?
   * people in the past opened issues related to building on other systems (FreeBSD if I recall well)
   * and Windows?
* if this can be automated, should we gate the CI in the same way we gate `clang-format` (so that there is no drift)? Basically, people must ensure that their includes were updated

Samertrad:
Great job well done

graydon:
Closing this as it's just noise sitting in the queue. We can revisit sometime if interested.

Issue #2196, opened 2019-07-12
Simulate mode is a little too willing to write to a live archive


graydon:
I ran simulation mode on what I wanted to be "real input data" so used the config file I typically use to run a test listener node. Unfortunately that listener config is also configured to write to a test history archive. Since simulation mode runs over a "real" ledger-number range but replays "not the real ledger history" (re-batching the real history at a different granularity) this means it wrote a bunch of simulation junk to a live ledger range in my test history archive.

No serious harm done -- it's a test archive, I can rebuild it -- but it seems like a relatively easy error to make, given the proximity we expect between "real" and "simulated" configs. Might be nice to make the default behaviour exit with an error if the user runs simulate with a writable history archive (perhaps overridable with a no-really-I-want-to-write-faux-history flag like `--write-simulated-history-to-archives` flag or such)

MonsieurNicolas:
Maybe we could solve that in a more general way: archives should contain the network passphrase somewhere (I've had close calls in the past switching between test and prod configs), and we should ensure that it matches whatever we're doing.
Simulation mode would add `* simulation` to the passphrase. The plan is to not actually keep the original transactions (and their signatures) from the transaction stream during a simulation and instead only used "derived accounts".

PR #2409, opened 2020-01-28
Ignore any archive write commands in simulation mode

marta-lokhova:
After discussing a bit with @jonjove, it looks like it makes sense to skip publishing in tx simulation mode, since its unclear how to interpret its performance impact given that we are not closing 5-second ledgers (publish during catchup tends to fall behind anyway). It is also a bit too easy to publish to an archive by mistake right now, so I suggest we avoid publishing, and if in the future we can always add a flag like `--publish` if we want to test that. If we go this route, this should also resolve https://github.com/stellar/stellar-core/issues/2196

### DIFF
--- a/src/main/CommandLine.cpp
+++ b/src/main/CommandLine.cpp
@@ -937,6 +937,8 @@ runSimulate(CommandLineArgs const& args)
         [&] {
             auto config = configOption.getConfig();
             config.setNoListen();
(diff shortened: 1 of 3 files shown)


graydon:
Yes please!

MonsieurNicolas:
r+ 1c10249b69992e3840e5a9df6797f631766981d8

PR #2410, opened 2020-01-28
Add timers for queueing delays from send queue and asio queue.

graydon:
This adds two new timers that track the amount of time outgoing messages are delayed due to queueing in our TCPPeer class' internal queue and then again in ASIO between write-issue and completion.


MonsieurNicolas:
looks good, forgot to update docs/metrics.md

MonsieurNicolas:
r+ 246943260cdef2ab70ebaa7611961e47344b9a46

PR #2414, opened 2020-01-30
Add ledger close stats and timers

marta-lokhova:
Started putting together some stats during ledger application, this may help with slow application investigations and finding the bottlenecks (e.g., prefetching, actual entry application, best offers etc)

### DIFF
--- a/src/ledger/LedgerManagerImpl.cpp
+++ b/src/ledger/LedgerManagerImpl.cpp
@@ -1164,6 +1164,10 @@ LedgerManagerImpl::applyTransactions(
     }
 
     prefetchTransactionData(txs);
+    auto applyTimer =
+        LogSlowExecution{"LedgerManagerImpl::applyTransactions",
+                         LogSlowExecution::Mode::AUTOMATIC_RAII, "took",
+                         std::chrono::milliseconds::zero()};
 
     for (auto tx : txs)
     {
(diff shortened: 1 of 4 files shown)


src/ledger/LedgerManagerImpl.cpp
```diff
@@ -1164,6 +1164,10 @@ LedgerManagerImpl::applyTransactions(
     }
 
     prefetchTransactionData(txs);
+    auto applyTimer =
+        LogSlowExecution{"LedgerManagerImpl::applyTransactions",
```
MonsieurNicolas: I don't think we should be adding performance information by default: it's getting too noisy.
Maybe we need to rethink how the "perf" partition works.

I sketched a change in https://github.com/MonsieurNicolas/stellar-core/commit/ab4be49baee4db2782d7a2cf3c2d282820c421a3 that may be closer to what we want.
Reason I sketched it was that I was getting too many `fetch-XYZ` messages even though my log level was `INFO`.
I didn't have time to find the right way to log (so ended up logging with the `INFO` level even though it's not).

src/ledger/LedgerManagerImpl.cpp
```diff
@@ -1244,12 +1248,15 @@ LedgerManagerImpl::logTxApplyMetrics(AbstractLedgerTxn& ltx, size_t numTxs,
...
-    CLOG(DEBUG, "Ledger") << "Ledger: " << ledgerSeq << " txs: " << numTxs
-                          << ", ops: " << numOps
-                          << ", prefetch hit rate (%): " << hitRate;
-
     // We lose a bit of precision here, as medida only accepts int64_t
     mPrefetchHitRate.set_count(std::llround(hitRate));
+
+    if (Logging::logTrace("Perf"))
+    {
+        CLOG(TRACE, "Perf")
```
MonsieurNicolas: related to my question on the perf partition: TRACE doesn't look right for this. As this is done once per ledger, DEBUG seemed OK. TRACE would be "I want to see performance information at the operation level" type of thing

src/ledger/LedgerTxn.cpp
```diff
@@ -2242,6 +2253,9 @@ uint32_t
 LedgerTxnRoot::Impl::prefetch(std::unordered_set<LedgerKey> const& keys)
 {
     uint32_t total = 0;
+    auto prefetchTimer =
```
MonsieurNicolas: Is the reason the code is written like this to make it "exception safe" when not in TRACE mode?
It looks a bit cludgy: it actually performs a bunch of computation that we never use even if we're not in TRACE mode.
Maybe this is related to what I was asking in https://github.com/stellar/stellar-core/pull/2386#issuecomment-584435308
Basically, maybe this (unsafe logging) should only be something that we can enable on test builds.

src/ledger/LedgerTxn.cpp
```diff
@@ -2672,7 +2711,24 @@ LedgerTxnRoot::Impl::putInEntryCache(
 {
     try
     {
-        mEntryCache.put(key, {entry, type});
+        mEntryCache.put(key, CacheEntry{entry, type});
```
MonsieurNicolas: I think those metrics got removed in a previous commit, and I am not sure I'd want to pay the price for them in the general case (this cache is very performance sensitive).

So either find a way to only compute those things on demand, or don't compute them at all.

MonsieurNicolas:
I think we don't need this anymore as we've got Tracy instrumentation

PR #2415, opened 2020-01-30
Prioritize write traffic

graydon:
This builds on https://github.com/stellar/stellar-core/pull/2413, adding a dual-priority scheme to outgoing traffic: transactions are flood priority, everything else is fetch. Fetch-priority traffic gets sent before flood-priority 3 times out of every 4 sends (so flood eventually always gets sent -- just might be held up a bit).

@marta-lokhova would you be willing to run another load test to see how this goes?


MonsieurNicolas:
as per our chat, closing this for now

PR #2421, opened 2020-02-05
[WIP] Expeimental account creation for simulation modes

marta-lokhova:
Simulation modes with disabled bucketlist allow bypassing consensus + ledger close when creating accounts. This change adds functionality to generate test accounts on a node. Still need to test a few things and run our performance tests, so it's probably not ready for review yet.


marta-lokhova:
closing, as it seems like we actually don't need this and can run perf tests with basic account creation functionality.

PR #2429, opened 2020-02-10
Fix missing reference value scp

MonsieurNicolas:
This PR fixes an issue introduced a long time ago ba4dd4a5071003681e1ec3b5874879487ec7e49f that became a real problem with the move to https://github.com/stellar/stellar-core/pull/2369

Basically: in a prepare statement, p' should be tracked like any other value (it was ignored).

This was causing tracking of related values to be wrong, which causes all sorts of downstream problems.


marta-lokhova:
r+ d8f1d451a93a48fc43ef84dba3c531c9865814cc

MonsieurNicolas:
mixed signal: I added an extra commit from a different PR

MonsieurNicolas:
r+ 7b8b8eed58c4456c98cb402cdd5464f479ab23b8

PR #2430, opened 2020-02-11
Bump overlay to 12

MonsieurNicolas:
As 12.4.0 will have significant improvement on the overlay front, we need to be able to migrate the network over time to a version of overlay that performs in the way we want


marta-lokhova:
r+ f7bb8e9665ecb7948800edaca4110e83b922d2e8

PR #2434, opened 2020-02-12
Initial implementation of ledger scaling simulation

### LINKED ISSUE #2456: SimulationTransactionFrame can throw XDR exception
## Issue Description
`SimulationTransactionFrame` can throw an XDR bad-access exception (subsequently logged via `LedgerManagerImpl`) when preparing a `SimulationMergeOpFrame` on https://github.com/stellar/stellar-core/blob/a39a512137bd9d32a47eefd694bf5a7f1553623a/src/transactions/simulation/SimulationTransactionFrame.cpp#L41-L42 This happens because `SimulationTransactionFrame` assumes that the `TransactionResult` was either `txSUCCESS` or `txFAILED` but other codes are possible.

## Steps to Reproduce
Catch up to ledger 10010000 then run
```
stellar-core simulate --first-ledger-inclusive 10010001 --last-ledger-inclusive 10020000 --ops-per-ledger 1
```

## Expected Result
The simulation should not crash and the transaction should be simulated correctly.

## Actual Result
The simulation crashed with the following exception
```
[Ledger FATAL] processFeesSeqNums error @ 1 : _result_t: results accessed when not selected [LedgerManagerImpl.cpp:1105]
```

marta-lokhova:
This PR implements an initial version of ledger scaling simulation, which allows scaling the bucketlist and simulate transactions application on it (by constructing fake transactions that mimic the real ones). This change includes the order book scaling as well. 

Also resolves #2456


marta-lokhova:
quick note: for now, I did not include salt for key pair generation. It's a simple change to add it back if we decide that we need it.

src/main/CommandLine.cpp
```diff
@@ -964,6 +965,78 @@ runSimulate(CommandLineArgs const& args)
...
+
+            std::vector<std::shared_ptr<BasicWork>> seq{simulateBuckets,
+                                                        downloadHeaders};
+
+            auto w = app->getWorkScheduler().executeWork<WorkSequence>(
+                "simulate-bucketlist-application", seq);
+
+            if (w->getState() == BasicWork::State::WORK_SUCCESS)
+            {
+                // Ensure that LedgerManager's view of LCL if consistent with
```
jonjove: Nitpick: typo "if" -> "is"

src/main/CommandLine.cpp
```diff
@@ -964,6 +965,78 @@ runSimulate(CommandLineArgs const& args)
...
+                // the bucketlist
+                FileTransferInfo ft(dir, HISTORY_FILE_TYPE_LEDGER, checkpoint);
+                XDRInputFileStream hdrIn;
+                hdrIn.open(ft.localPath_nogz());
+                LedgerHeaderHistoryEntry curr;
+                while (hdrIn && hdrIn.readOne(curr))
+                    ;
```
jonjove: Is this loop supposed to be empty? If so, then I don't really understand what this is doing so perhaps a comment would be helpful.

src/catchup/simulation/GenerateBucketsWork.cpp
```diff
@@ -0,0 +1,260 @@
...
+            key = getNewSecret(le.data.trustLine().accountID, count);
+            newEntry.data.trustLine().accountID = key.getPublicKey();
+            break;
+        case DATA:
+            key = getNewSecret(le.data.data().accountID, count);
+            newEntry.data.data().accountID = key.getPublicKey();
+            break;
+        case OFFER:
+            key = getNewSecret(le.data.offer().sellerID, count);
+            newEntry.data.offer().sellerID = key.getPublicKey();
```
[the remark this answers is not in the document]
marta-lokhova: yeah, this will be resolved with the offer `offerID` change I'm currently testing.

src/main/CommandLine.cpp
marta-lokhova: yes. it's reading the last ledger header

src/catchup/simulation/GenerateBucketsWork.cpp
```diff
@@ -0,0 +1,260 @@
...
+            key = getNewSecret(le.data.account().accountID, count);
+            newEntry.data.account().accountID = key.getPublicKey();
+
+            for (auto const& signer : le.data.account().signers)
+            {
+                if (signer.key.type() == SIGNER_KEY_TYPE_ED25519)
+                {
+                    auto pubKey = KeyUtils::convertKey<PublicKey>(signer.key);
+                    key = getNewSecret(pubKey, count);
+                    newEntry.data.account().signers.push_back(
```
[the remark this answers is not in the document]
marta-lokhova: ah good catch!

marta-lokhova:
@jonjove to follow-up on our previous discussion regarding command line user experience, I added a new option to pass history state. If provided, bucket generation is bypassed, and buckets are applied from a previously generated HAS file.

marta-lokhova:
Alright, i think the next set of changes to be done here are:
* Improve memory consumption during bucketlist generation (via creating smaller buckets then merging them)
* Mitigate simulated transaction signing overhead (i.e., persist all the transactions and allow reading from that stream during replay)

src/catchup/simulation/ApplyTransactionsWork.cpp
```diff
@@ -23,7 +31,184 @@ ApplyTransactionsWork::ApplyTransactionsWork(
...
+    std::vector<TransactionResultPair>& results,
+    std::vector<UpgradeType>& upgrades)
+{
+    assert(mTransactionIter != mTransactionHistory.txSet.txs.cend());
+    assert(mResultIter != mResultHistory.txResultSet.results.cend());
+
+    LedgerTxn ltx(mApp.getLedgerTxnRoot());
+
+    // No amplification needed, simply return existing transactions and results
+    if (mCount == 0)
```
jonjove: `mCount` should be an argument here and a local variable at the call site, rather than an instance variable. It will make it easier to reason about because then maintainers won't be searching through the file for other uses like I just was.

src/catchup/simulation/ApplyTransactionsWork.cpp
```diff
@@ -64,20 +259,33 @@ ApplyTransactionsWork::getNextLedger(
...
+    };
+
     while (true)
     {
+        // sustained: mMaxOperations > 0,
+        // scaled ledger: avoid checking nOps <= mMaxOperations, mMaxOperations
+        // = 0
         while (mTransactionIter != mTransactionHistory.txSet.txs.cend() &&
-               nOps <= mMaxOperations)
+               moreOps())
```
jonjove: This doesn't seem to work the way I would expect? The old implementation was set up to guarantee that you would get between `mMaxOperations` and `mMaxOperations + 100` operations per ledger. Now I think you're forcing the operation-rate to scale with the ledger size. But they should be able to scale independently.

src/catchup/simulation/ApplyTransactionsWork.h
```diff
@@ -26,24 +26,37 @@ class ApplyTransactionsWork : public BasicWork
     std::vector<TransactionResultPair>::const_iterator mResultIter;
 
     uint32_t const mMaxOperations;
+    uint32_t const mMultiplier;
```
jonjove: How is `mSimulateOnly` different from `mMultiplier == 1`? If they're the same, I'd prefer to use only `mMultiplier = 1` as the canonical "simulate-only".

src/catchup/simulation/ApplyTransactionsWork.cpp
marta-lokhova: how so? Old functionality is preserved, but there is a new mode where you scale ledgers with a multiplier. It depends on what `mMaxOperations` is configured to (anything above 0 will yield sustained operation-rate, while 0 just scales each ledger based on the multiplier)

jonjove: Sorry, I don't think what I wrote was a good description of what I noticed. This implementation scales the variation in the operation rate. As noted above, the old version guaranteed that it would be between `mMaxOperations` and `mMaxOperations + 100`. Your implementation reduces that guarantee to being between `mMaxOperations` and `mMaxOperations + mMultiplier * 100`. If we wanted to simulate 10x ledger size but 1x operation rate (eg. ~100M entries and ~200 ops/ledger) this would amount to a variation in the operation rate 5 times bigger than the actual desired operation rate. I don't really see that as a constant operation rate at all.

src/catchup/simulation/GenerateBucketsWork.h
```diff
@@ -0,0 +1,66 @@
...
+
+    // New HAS is populated incrementally
+    HistoryArchiveState mGeneratedApplyState;
+    uint32_t const mMultiplier;
+    uint32_t mLevel;
+
+    std::shared_ptr<Bucket> mPrevSnap;
+
+    // Keep references to created buckets for cleanup
+    std::vector<std::shared_ptr<Bucket>> mBuckets;
```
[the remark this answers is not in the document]
marta-lokhova: Agreed that the comment isn't super clear. The reason I added this is to ensure downstream works can't forget newly created buckets (actually ran into this issue, since `GenerateBucketsWork` populates a HAS, which only keeps bucket hash strings). Re-visiting this, I've made two changes: 1. update the comment to make it clear why these buckets need to remembered, and 2. pass a reference to the generated buckets map to `GenerateBucketsWork` to make the ownership a bit more explicit (i.e. let `SimulateApplyBucketsWork` track generated buckets)

src/catchup/simulation/GenerateBucketsWork.cpp
```diff
@@ -0,0 +1,224 @@
...
+            auto b1 = simulated.front();
+            simulated.pop_front();
+            auto b2 = simulated.front();
+            simulated.pop_front();
+            fbs.emplace_back(mApp, b1, b2, /* shadows */ shadows,
+                             Config::CURRENT_LEDGER_PROTOCOL_VERSION, false,
+                             mLevel);
+        }
+
+        std::transform(fbs.begin(), fbs.end(), std::back_inserter(simulated),
```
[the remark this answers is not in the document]
marta-lokhova: Good suggestion, updated.

src/catchup/simulation/GenerateBucketsWork.cpp
```diff
@@ -0,0 +1,224 @@
...
+#include "catchup/simulation/GenerateBucketsWork.h"
+#include "bucket/Bucket.h"
+#include "bucket/BucketInputIterator.h"
+#include "bucket/BucketList.h"
+#include "bucket/BucketManager.h"
+#include "bucket/BucketOutputIterator.h"
+#include "crypto/SecretKey.h"
+#include "crypto/SignerKey.h"
+#include "history/HistoryArchive.h"
+#include "util/SimulationUtils.h"
```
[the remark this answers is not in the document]
marta-lokhova: Good call. I'll address this one when I do the final rebase (so we can first review the commit I'm about to push that addresses your feedback)

src/util/SimulationUtils.cpp
```diff
@@ -0,0 +1,284 @@
...
+        SimulationUtils::updateAccountID(op.body.allowTrustOp().trustor, n);
+        break;
+    case ACCOUNT_MERGE:
+        SimulationUtils::updateAccountID(op.body.destination(), n);
+        break;
+    case MANAGE_BUY_OFFER:
+        offer = op.body.manageBuyOfferOp().offerID;
+        SimulationUtils::replaceIssuer(op.body.manageBuyOfferOp().selling, n);
+        SimulationUtils::replaceIssuer(op.body.manageBuyOfferOp().buying, n);
+        if (offer > 0)
```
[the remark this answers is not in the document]
marta-lokhova: `getNewOfferID` throws if it sees an invalid offer if (0), whereas here this is a valid op and we want to just leave it at 0. That being said, now I'm thinking that perhaps `getNewOfferID` is unnecessarily aggressive and doesn't need to throw on 0, so I updated it to just return 0 instead of throwing.

src/catchup/simulation/ApplyTransactionsWork.cpp
marta-lokhova: I agree it's not ideal to scale it this way, but I'm not sure how much it matters here: the multiplier scaling is done during transaction generation, after which simulated transactions are publishes to a special archive. When we run `simulate-transactions` , we actually don't do any scaling, but replay simulated transactions from the archive with whatever ops-per-ledger , so I do not think we'd have this problem with operation rate (command flow for this is described in CommandLine.cpp right above `runGenerateOrSimulateTxs` function)

src/catchup/simulation/ApplyTransactionsWork.cpp
```diff
@@ -23,7 +31,184 @@ ApplyTransactionsWork::ApplyTransactionsWork(
...
+    newTx.operations.clear();
+    auto& operations = newTx.operations;
+
+    for (auto const& op : env.tx.operations)
+    {
+        Operation newOp = op;
+        SimulationUtils::updateOperation(newOp, mCount);
+
+        // We might have just added a signer; check if its signature is required
+        if (op.body.type() == SET_OPTIONS && op.body.setOptionsOp().signer)
```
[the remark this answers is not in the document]
marta-lokhova: Added a comment. I'm not sure if we care about old versions (likely we don't), but I'm going to keep it for now. Once we establish how exactly we're going to use the simulation code and for which ranges, I'm happy to re-visit this.

src/catchup/simulation/ApplyTransactionsWork.cpp
```diff
@@ -23,7 +31,184 @@ ApplyTransactionsWork::ApplyTransactionsWork(
...
+    }
+
+    auto const& env = *mTransactionIter;
+    auto txHash = mResultIter->transactionHash;
+    TransactionEnvelope newEnv = env;
+    Transaction& newTx = newEnv.tx;
+    auto key = SimulationUtils::updateAccountID(newTx.sourceAccount, mCount);
+
+    // Keep track of accounts that need to sign
+    std::vector<SecretKey> keys;
```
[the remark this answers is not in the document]
marta-lokhova: Done.

src/catchup/simulation/ApplyTransactionsWork.cpp
```diff
@@ -77,7 +78,8 @@ ApplyTransactionsWork::hasSig(PublicKey const& account,
         return false;
     }
 
-    for (auto const& sig : oldEnvelope.signatures)
+    auto env = oldEnvelope;
```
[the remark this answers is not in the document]
marta-lokhova: While implementing the signature fix, I actually updated `hasSig` to take signatures as an argument, so there is no envelope copying needed anymore.

src/transactions/simulation/SimulationFeeBumpTransactionFrame.h
```diff
@@ -0,0 +1,32 @@
...
+
+#include "transactions/FeeBumpTransactionFrame.h"
+#include "transactions/simulation/SimulationTransactionFrame.h"
+
+namespace stellar
+{
+class SimulationFeeBumpTransactionFrame : public FeeBumpTransactionFrame
+{
+    TransactionResult mSimulationResult;
+    std::shared_ptr<SimulationTransactionFrame> mSimulationInnerTx;
```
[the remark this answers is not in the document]
marta-lokhova: Good idea, done.

src/catchup/simulation/ApplyTransactionsWork.cpp
```diff
@@ -63,14 +64,125 @@ checkResults(Application& app, uint32_t ledger,
...
+                res.result.result.code(), dbRes.code(),
                 binToHex(res.transactionHash));
             CLOG(ERROR, "History") << msg;
+            return;
+        }
+
+        // Now check each operation result
+        auto const& archiveRes = results[i].result.result;
+
+        if (dbRes.code() == txFAILED || dbRes.code() == txSUCCESS)
```
jonjove, line 81: This doesn't handle fee-bump transactions at all. Is that the intention?

src/catchup/simulation/ApplyTransactionsWork.cpp
```diff
@@ -63,14 +64,125 @@ checkResults(Application& app, uint32_t ledger,
...
+                {
+                    continue;
+                }
+
+                // Otherwise, compare each operation
+                auto archiveOpRes = archiveRes.results()[j].tr();
+                auto dbOpRes = dbRes.results()[j].tr();
+
+                bool match = false;
+                std::string type = xdr::xdr_to_string(archiveOpRes.type());
```
jonjove, line 107: I think this is unused.

src/catchup/simulation/ApplyTransactionsWork.cpp
```diff
@@ -63,14 +64,125 @@ checkResults(Application& app, uint32_t ledger,
...
             CLOG(ERROR, "History") << msg;
+            return;
+        }
+
+        // Now check each operation result
+        auto const& archiveRes = results[i].result.result;
+
+        if (dbRes.code() == txFAILED || dbRes.code() == txSUCCESS)
+        {
+            assert(archiveRes.code() == txFAILED ||
```
jonjove, line 83: I don't think you can always guarantee this. It's possible, for example, that a transaction that changes signers succeeds in production but fails in simulation, then causes another transaction that failed in production with `txFAILED` to fail in simulation with `txBAD_AUTH`.

src/catchup/simulation/ApplyTransactionsWork.cpp
```diff
@@ -63,14 +64,125 @@ checkResults(Application& app, uint32_t ledger,
...
+                   archiveRes.code() == txSUCCESS);
+            assert(archiveRes.results().size() == dbRes.results().size());
+
+            for (size_t j = 0; j < archiveRes.results().size(); j++)
+            {
+                if (archiveRes.results()[j].code() != dbRes.results()[j].code())
+                {
+                    throw std::runtime_error(
```
jonjove, line 91: Why throw here instead of logging?

src/catchup/simulation/ApplyTransactionsWork.cpp
```diff
@@ -63,14 +64,125 @@ checkResults(Application& app, uint32_t ledger,
...
+                if (archiveRes.results()[j].code() != opINNER)
+                {
+                    continue;
+                }
+
+                // Otherwise, compare each operation
+                auto archiveOpRes = archiveRes.results()[j].tr();
+                auto dbOpRes = dbRes.results()[j].tr();
+
+                bool match = false;
```
jonjove, line 106: nitpick: move this closer to where it's actually used.

src/catchup/simulation/ApplyTransactionsWork.cpp
```diff
@@ -63,14 +64,125 @@ checkResults(Application& app, uint32_t ledger,
...
+                   archiveRes.code() == txSUCCESS);
+            assert(archiveRes.results().size() == dbRes.results().size());
+
+            for (size_t j = 0; j < archiveRes.results().size(); j++)
+            {
+                if (archiveRes.results()[j].code() != dbRes.results()[j].code())
+                {
+                    throw std::runtime_error(
+                        fmt::format("Expected {} but got {}",
+                                    archiveRes.results()[j].code(),
```
jonjove, line 93: This is going to log numbers like "Expected -3 but got -5". Is that what you intended? It's possible to get the actual strings if you want them.

src/catchup/simulation/ApplyTransactionsWork.cpp
marta-lokhova: it's already handled on line 68 (case where the codes aren't equal). the assert is redundant here, so i removed it.

src/catchup/simulation/ApplyTransactionsWork.cpp
```diff
@@ -35,17 +49,190 @@ ApplyTransactionsWork::ApplyTransactionsWork(
...
+        if (PubKeyUtils::verifySig(account, sig.signature, hash))
+        {
+            return true;
+        }
+    }
+    return false;
+}
+
+void
+ApplyTransactionsWork::addSignerKeys(AccountID const& acc,
```
MonsieurNicolas: this seems computationally intensive, do we need to do this because we need to "match" & replace signatures for partitions?

In any case, not a big deal if this is only done when we generate the stream (it would be if this was used while replaying).

src/catchup/simulation/ApplyTransactionsWork.cpp
```diff
@@ -35,17 +49,190 @@ ApplyTransactionsWork::ApplyTransactionsWork(
...
+            if (hasSig(pubKey, oldEnvelope, txHash))
+            {
+                keys.emplace_back(SimulationUtils::getNewSecret(pubKey, n));
+            }
+        }
+    }
+}
+
+uint32_t
+ApplyTransactionsWork::scaleLedger(
```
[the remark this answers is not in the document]
MonsieurNicolas: Maybe the notes on TxSimulation should be in their own `TxSimulation.md` file somewhere (that we can point to from the perf docs) instead of scattered in the code base

src/catchup/simulation/ApplyTransactionsWork.cpp
```diff
@@ -35,17 +49,190 @@ ApplyTransactionsWork::ApplyTransactionsWork(
...
+
+    for (auto& op : newTx.operations)
+    {
+        // Add signer keys where needed before simulating the operation
+        if (op.sourceAccount)
+        {
+            addSignerKeys(*op.sourceAccount, ltx, keys, env, n);
+        }
+
+        // Prior to protocol 10, it is possible that we might have just added a
```
MonsieurNicolas, line 322: NB: I don't know if we care about simulation for older versions of the protocol (older than more than a year ago let's say), this adds complexity that we don't need.

src/catchup/simulation/ApplyTransactionsWork.cpp
```diff
@@ -35,17 +49,190 @@ ApplyTransactionsWork::ApplyTransactionsWork(
...
+    // Is the signature of this account present in the envelope we're
+    // simulating?
+    if (!mUsedPubKeys.insert(account).second)
+    {
+        return false;
+    }
+
+    for (auto const& sig : oldEnvelope.signatures)
+    {
+        if (PubKeyUtils::verifySig(account, sig.signature, hash))
```
MonsieurNicolas: checking signatures like this bypasses using the "hint" - see `SignatureUtils::verify`

src/herder/simulation/SimulationTxSetFrame.cpp
```diff
@@ -75,12 +76,23 @@ SimulationTxSetFrame::sortForApply()
...
 
     auto resultIter = mResults.cbegin();
+    uint32_t count = 0;
     for (auto const& txEnv : mTransactions)
     {
         res.emplace_back(SimulationTransactionFrame::makeTransactionFromWire(
-            mNetworkID, txEnv, resultIter->result));
+            mNetworkID, txEnv, resultIter->result, count));
         ++resultIter;
+        // Transaction generation guarantees that the number of transactions in
```
MonsieurNicolas, line 106: actually that's not exactly the reason: see my comment on how we generate the stream (and how we need a comment that explains what's going on, ie round robin over partitions).

src/main/CommandLine.cpp
```diff
@@ -902,15 +903,48 @@ runRebuildLedgerFromBuckets(CommandLineArgs const& args)
...
+// - transactions from the archives replayed at a sustained ops-per-ledger
+// rate
+// - real and simulated transactions replayed at a sustained ops-per-ledger rate
+// - real and simulated transactions replayed with real ledger sizes scaled by
+// the multiplier
+
+// There are a few steps involved in prepping the state before we can simulate
+// transactions without state generation overhead. Typically, the sequence of
+// commands is as follows:
+// simulate-bucketlist --simulate-bucketlist <L> --multiplier <M>
```
MonsieurNicolas, line 916: this comment belongs in a readme of sorts (I was suggesting creating a "transaction-simulation.md" file)

src/catchup/simulation/SimulateApplyBucketsWork.cpp
```diff
@@ -0,0 +1,92 @@
...
+#include "catchup/ApplyBucketsWork.h"
+#include "catchup/simulation/GenerateBucketsWork.h"
+#include "historywork/DownloadBucketsWork.h"
+#include "historywork/GetHistoryArchiveStateWork.h"
+#include "ledger/LedgerManager.h"
+#include "ledger/LedgerTxn.h"
+
+namespace stellar
+{
+SimulateApplyBucketsWork::SimulateApplyBucketsWork(
```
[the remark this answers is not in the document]
MonsieurNicolas: actually, you need it - I thought you were publishing the scaled bucket list, instead you're doing an "in place" scaling.

Should this be called `TxSimScaleBucketList` instead then?

src/main/CommandLine.cpp
```diff
@@ -928,46 +962,190 @@ runSimulate(CommandLineArgs const& args)
...
+    auto opsPerLedgerParser = [](uint32_t& opsPerLedger) {
+        return clara::Opt{opsPerLedger, "OPS-PER-LEDGER"}["--ops-per-ledger"](
+            "desired number of operations per ledger, will never be less but "
+            "could be up to 100 * multiplier more. If 0, real ledger sizes "
+            "from the archive stream are used");
+    };
+
+    auto multiplierParser = [](uint32_t& multiplier) {
+        return clara::Opt{multiplier,
+                          "N"}["--multiplier"]("aplification factor, "
```
MonsieurNicolas, line 974: typo "aplification"

src/catchup/simulation/ApplyTransactionsWork.cpp
```diff
@@ -543,41 +545,85 @@ ApplyTransactionsWork::onReset()
...
-    if (mVerifyResults)
-    {
-        checkResults(mApp, header.ledgerSeq, results);
-    }
+    auto applyLedger = std::make_shared<ApplyLedgerWork>(mApp, closeData);
+
+    auto const& ham = mApp.getHistoryArchiveManager();
+    auto const& hm = mApp.getHistoryManager();
+    bool waitForPublish = false;
+    auto condition = [&lm, &ham, &hm, waitForPublish]() mutable {
```
MonsieurNicolas, line 606: can you explain why we need this? I think it was because we would create a huge backlog of things to publish?

src/catchup/simulation/ApplyTransactionsWork.cpp
marta-lokhova: Yes, we need to make sure that all scaled transaction envelopes have correct signatures. It's unfortunate that this is pretty expensive, but we don't pay the price during the actual simulation (only during generation)

src/catchup/simulation/ApplyTransactionsWork.cpp
marta-lokhova: Yes, as discussed earlier, publish can't keep up during replay.

src/util/SimulationUtils.cpp
```diff
@@ -0,0 +1,269 @@
...
+getNewSecret(AccountID const& key, uint32_t n)
+{
+    return SecretKey::fromSeed(
+        sha256(KeyUtils::toStrKey(key) + std::to_string(n)));
+}
+
+int64_t
+getNewOfferID(int64_t offerId, uint32_t n)
+{
+    if (n == 0 || offerId == 0)
```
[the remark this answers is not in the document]
marta-lokhova: The code used to throw on offerID == 0, but I changed it since 0 is a valid offer ID when creating/deleting offers used in operation simulation code (see `SimulationUtils::updateOperation`)

src/util/SimulationUtils.cpp
```diff
@@ -0,0 +1,284 @@
...
+}
+
+void
+updateOperation(Operation& op, uint32_t n)
+{
+    uint32_t offer;
+    PublicKey signerKey;
+    Signer signer;
+
+    switch (op.body.type())
```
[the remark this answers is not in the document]
marta-lokhova: https://github.com/stellar/stellar-core/issues/2520

MonsieurNicolas:
r+ 103f5fd5110498d3bceac8e175e44de028e4d597

PR #2439, opened 2020-02-19
{Ballot,Nomination}Protocol::processEnvelope: Use early return to simplify flow

Geod24:
# Description

```
This diff is better viewed with whitespace changes ignored.

By using early returns, we simplify the code flow,
making it more obvious which path are error path and which are not.
It also lessen the burden on the programmer to keep context in mind,
and reduce the indenting level as an added bonus.

While NominationProtocol was fairly consistent in its usage of a 'res' value,
BallotProtocol used a mixed of both styles.
Not that some error handling has also been hoisted as a result of the refactoring.
```

# Checklist
- [X] Compiles
- [X] Ran all tests
- [X] ~If change impacts performance, [...]~

### DIFF
--- a/src/scp/BallotProtocol.cpp
+++ b/src/scp/BallotProtocol.cpp
@@ -185,51 +185,10 @@ BallotProtocol::processEnvelope(SCPEnvelopeWrapperPtr envelope, bool self)
     }
 
     auto validationRes = validateValues(statement);
-    if (validationRes != SCPDriver::kInvalidValue)
+
+    // If the value is not valid, we just ignore it.
+    if (validationRes == SCPDriver::kInvalidValue)
     {
-        bool processed = false;
-
-        if (mPhase != SCP_PHASE_EXTERNALIZE)
-        {
-            if (validationRes == SCPDriver::kMaybeValidValue)
-            {
-                mSlot.setFullyValidated(false);
-            }
-
-            recordEnvelope(envelope);
-            processed = true;
-            advanceSlot(statement);
-            res = SCP::EnvelopeState::VALID;
-        }
-
-        if (!processed)
-        {
-            // note: this handles also our own messages
-            // in particular our final EXTERNALIZE message
-            if (mPhase == SCP_PHASE_EXTERNALIZE &&
-                mCommit->getBallot().value == getWorkingBallot(statement).value)
-            {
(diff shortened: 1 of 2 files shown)


Geod24:
Ping @MonsieurNicolas ?

MonsieurNicolas:
Yeah sorry, I'm going to take a look for our next release @Geod24 . Thanks!

src/scp/BallotProtocol.cpp
```diff
@@ -2119,12 +2107,12 @@ BallotProtocol::getJsonQuorumInfo(NodeID const& id, bool summary, bool fullKeys)
...
 
-    auto f = LocalNode::findClosestVBlocking(*qSet, mLatestEnvelopes,
-                                             [&](SCPStatement const& st) {
-                                                 return areBallotsCompatible(
-                                                     getWorkingBallot(st), b);
-                                             },
-                                             &id);
+    auto f = LocalNode::findClosestVBlocking(
+        *qSet, mLatestEnvelopes,
+        [&](SCPStatement const& st) {
```
MonsieurNicolas: not sure why this changed? Maybe you're using a different version of clang-format?

MonsieurNicolas:
r+ 3ab66ebf33ddc29a94c30549f703d3fd5166ac97

src/scp/BallotProtocol.cpp
Geod24: I couldn't really get to replicate your setup of clang-format (maybe because I'm on Mac?).
Even with clang-format 5 I have diff with HEAD... Should I fix it according to what the CI is saying ?

Geod24: It's complaining in 3 places: https://travis-ci.org/stellar/stellar-core/jobs/658035681#L1257

MonsieurNicolas: yeah the CI gives you the diff, so you can apply it and you should be good to go.
( in the future, you need to run `clang-format-5.0` not `clang-format` that is likely to pick up the wrong version )

Geod24:
Rebased on master, applied formatting diff.

MonsieurNicolas:
r+ cc2bf2c7a5bbcf18b0108268a2e52a577e6653e4

Issue #2441, opened 2020-02-20
delays in receiving SCP messages from connected peers


MonsieurNicolas:
I don't know if this is a different incarnation of #2368 

On watcher node that I have:
```
2020-02-20T22:01:58.365 GABA5 [Ledger INFO] Closed ledger: [seq=28319360, hash=4a23ad]
2020-02-20T22:02:09.164 GABA5 [Herder INFO] Quorum information for 28319359 : {"agree":3,"delayed":0,"disagree":0,"fail_at":2,"hash":"75b44e","ledger":28319359,"missing":0,"phase":"EXTERNALIZE"}

2020-02-20T22:02:09.168 GABA5 [Ledger INFO] Got consensus: [seq=28319361, prev=4a23ad, txs=117, ops=234, sv: [ txH: 232f01, ct: 1582236123, upgrades: [ ] ]]
2020-02-20T22:02:09.175 GABA5 [Tx INFO] applying ledger 28319361 (txs:117, ops:234)
2020-02-20T22:02:09.236 GABA5 [Ledger INFO] Closed ledger: [seq=28319361, hash=f8ac19]
```

Some messages that the watcher node saw:
```
            [
               1582236123,
               "{ENV@GAAV2 |  i: 28319361 | NOMINATE | D: 341a0a | X: {'[ SIGNED@GAAV2 txH: 232f01, ct: 1582236123, upgrades: [ ] ]'} | Y: {} }",
               false
            ],
...
[
               1582236125,
               "{ENV@GDKWE |  i: 28319361 | EXTERNALIZE | c: (1,[ txH: 232f01, ct: 1582236123, upgrades: [ ] ]) | h.n: 1 | (lastD): 341a0a }",
               false
            ],
            [
               1582236125,
               "{ENV@GA35T |  i: 28319361 | EXTERNALIZE | c: (1,[ txH: 232f01, ct: 1582236123, upgrades: [ ] ]) | h.n: 1 | (lastD): 341a0a }",
               false
            ],
            [
               1582236125,
               "{ENV@GCFON |  i: 28319361 | EXTERNALIZE | c: (1,[ txH: 232f01, ct: 1582236123, upgrades: [ ] ]) | h.n: 1 | (lastD): 341a0a }",
               false
            ],
            [
               1582236125,
[...]

PR #2446, opened 2020-02-23
Dial down the YieldTimer on synchronous read-dispatch loop

graydon:
Experimental change to attempt to address #2441 by limiting time limit / iteration count of the synchronous read  dispatching loop.


graydon:
Abandoning this because it's just one of several random attempts we didn't pursue to fix the slow writes.

PR #2447, opened 2020-02-24
Update overlay default

MonsieurNicolas:
# Description

This PR updates the defaults used to batch reads to be less aggressive.
It also schedules the first write right away instead of trying to wait for more writes.


graydon:
r+ d34138a

PR #2448, opened 2020-02-24
TCPPeer: avoid bouncing off ASIO before calling messageSender at first

graydon:
This is plausibly causing more latency than it's helping in throughput (see ongoing saga of https://github.com/stellar/stellar-core/issues/2441)


graydon:
Abandoning this because it's just one of several random attempts we didn't pursue to fix the slow writes.

PR #2449, opened 2020-02-26
Check medida math

graydon:
# Description

This adds some testing machinery to confirm that medida is calculating percentiles properly, and that when it runs a sliding-window reservoir sampler it's not corrupting the samples too much.

Includes a submodule bump of libmedida that absorbs https://github.com/stellar/medida/pull/15 and https://github.com/stellar/medida/pull/16 which correct the quantile calculation and implement the sliding window, respectively.


MonsieurNicolas:
r+ b78e65a98eba7bcf24a426ddd4badfdfad268dc2

PR #2450, opened 2020-02-27
Fix seqnum processing in transaction simulation

marta-lokhova:
It appears that `simulate` command does not work correctly when `ops-per-ledger` is less max ops per ledger config (i.e., 1000), since it is possible to split multiple operations for an account across multiple ledgers and fail in `processSeqNum`. This PR overrides `processSeqNum` similarly to how it's done for `processFeeSeqNum`.

### DIFF
--- a/src/transactions/simulation/SimulationTransactionFrame.cpp
+++ b/src/transactions/simulation/SimulationTransactionFrame.cpp
@@ -100,4 +100,16 @@ SimulationTransactionFrame::processFeeSeqNum(AbstractLedgerTxn& ltx,
         acc.seqNum = mEnvelope.v0().tx.seqNum;
     }
 }
+
+void
+SimulationTransactionFrame::processSeqNum(AbstractLedgerTxn& ltx)
+{
+    auto header = ltx.loadHeader();
+    if (header.current().ledgerVersion >= 10)
+    {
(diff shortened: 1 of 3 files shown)


src/transactions/simulation/SimulationTransactionFrame.cpp
```diff
@@ -100,4 +100,19 @@ SimulationTransactionFrame::processFeeSeqNum(AbstractLedgerTxn& ltx,
...
+{
+    auto header = ltx.loadHeader();
+    if (header.current().ledgerVersion >= 10)
+    {
+        auto sourceAccount = loadSourceAccount(ltx, header);
+        if (sourceAccount.current().data.account().seqNum > mEnvelope.tx.seqNum)
+        {
+            return;
+        }
+        sourceAccount.current().data.account().seqNum = mEnvelope.tx.seqNum;
```
jonjove: I think lines 111-114 should be removed, making this analogous to lines 98-101.

marta-lokhova: Done.

MonsieurNicolas:
r+ 68d10dbdd6990daef57ed3180cc01c3ad0fd05cc

PR #2454, opened 2020-03-03
Quorum check interrupt

### LINKED ISSUE #2189: Consider making "quorum intersection critically check" interruptible or multi-part
https://github.com/stellar/stellar-core/pull/2183 introduced an extended check for quorum intersection.

That extended check can take a lot longer than the main quorum intersection check.

One of the problems with this is that if a change in quorum configuration happens while the check is still running, we're not going to perform the "high priority" work of checking quorum intersection.

Potential solutions:
* (simplest) use different threads for quorum intersection and extended quorum intersection
* add some sort of time limit to the extended check/or only explore the space in multiple rounds, maybe combined with a randomization of the "critical groups" (returned by `getIntersectionCriticalGroups`), so that the exploration over time covers "all of the potential groups", even when changes happen in succession

### LINKED ISSUE #2443: Hang on shutdown computing quorum intersection
## Issue Description

Quite often, I find that core seems to be stuck when shutting down.

I attached a debugger and it seems that it's computing quorum intersection.

## Steps to Reproduce
Start a node watching the public network, wait for consensus to be acquired and 

## Expected Result
core stops when asked to, aborting computation of quorum intersection and other costly tasks

## Actual Result
core keeps computing quorum intersection

### stellar-core Version
12.3.0

graydon:
# Description

This adds an asynchronous inter-thread interruption mechanism to the quorum intersection checking subsystem (which runs on a background thread) and triggers it on both server shutdown concurrent quorum changes during a running calculation. Rather than thread an interruption status all the way in and out of every line of the subsystem, I just have it throw and catch an exception. This shouldn't happen especially often.

Also includes two fixes that both address (in different ways) a sporadic bug I've seen lately concerning deleting tmpdirs in the wrong order during shutdown.

Resolves #2189 
Resolves #2443

# Checklist
- [x] Compiles
- [x] Ran all tests

### DIFF
--- a/src/herder/Herder.h
+++ b/src/herder/Herder.h
@@ -89,6 +89,7 @@ class Herder
     virtual void syncMetrics() = 0;
 
     virtual void bootstrap() = 0;
+    virtual void shutdown() = 0;
 
     // restores Herder's state from disk
     virtual void restoreState() = 0;
--- a/src/herder/HerderImpl.cpp
+++ b/src/herder/HerderImpl.cpp
@@ -132,6 +132,19 @@ HerderImpl::bootstrap()
 }
 
 void
+HerderImpl::shutdown()
+{
+    if (mLastQuorumMapIntersectionState.mRecalculating)
+    {
+        // We want to interrupt any calculation-in-progress at shutdown to
+        // avoid a long pause joining worker threads.
+        CLOG(DEBUG, "Herder")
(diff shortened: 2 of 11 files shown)


MonsieurNicolas:
r+ 6ef010e9f88909a0cbe961ce4750ce937d834cd0

PR #2457, opened 2020-03-04
Do not fsync metadata output stream, pipes do not support.

graydon:
This just changes the metadata output stream to not fsync on close: pipes don't support that, and we really can't tell when we have or don't-have a pipe.


MonsieurNicolas:
r+ 09c653bf596b2d8833590c120bd61c2fbb8b9a10

PR #2466, opened 2020-03-11
Delayed externalize metric

### LINKED ISSUE #2451: consider improving metrics that report "delayed" validators
Right now "delayed" is reported in a very opaque way: any message received post externalize is considered "delayed".

The problem is that we don't actually get a good idea of the variance of those delays: some validators could be milliseconds behind, where as others could be seconds behind.

MonsieurNicolas:
# Description

Resolves #2451

This PR adds two new metrics to help track
* the "perceived lag" between network closing (modeled as delay between "first externalize event" and actual externalize)
* validators that are "behind" (modeled as delay between the node externalizing and receiving externalize messages from other nodes)

# Checklist
- [ ] Compiles
- [ ] Ran all tests

### DIFF
--- a/docs/metrics.md
+++ b/docs/metrics.md
@@ -116,8 +116,10 @@ scp.pending.ready                        | counter   | number of envelopes ready
 scp.sync.lost                            | meter     | validator lost sync
 scp.timeout.nominate                     | meter     | timeouts in nomination
 scp.timeout.prepare                      | meter     | timeouts in ballot protocol
-scp.timing.externalized                  | timer     | time spent in ballot protocol
 scp.timing.nominated                     | timer     | time spent in nomination
+scp.timing.externalized                  | timer     | time spent in ballot protocol
+scp.timing.externalize-lag               | timer     | delay between first externalize message and local node externalizing
+scp.timing.externalize-delay             | timer     | delay between local node externalizing and later externalize messages
 scp.value.invalid                        | meter     | SCP value is invalid
 scp.value.valid                          | meter     | SCP value is valid
 
--- a/src/herder/HerderSCPDriver.cpp
+++ b/src/herder/HerderSCPDriver.cpp
@@ -934,6 +934,24 @@ HerderSCPDriver::recordSCPEvent(uint64_t slotIndex, bool isNomination)
 }
 
 void
+HerderSCPDriver::recordLogTiming(VirtualClock::time_point start,
+                                 VirtualClock::time_point end,
+                                 medida::Timer& timer,
+                                 std::string const& logStr,
+                                 std::chrono::nanoseconds threshold,
+                                 uint64_t slotIndex)
+{
+    auto delta =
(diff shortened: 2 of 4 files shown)


src/herder/HerderSCPDriver.h
```diff
@@ -88,6 +88,8 @@ class HerderSCPDriver : public SCPDriver
 
     void recordSCPExecutionMetrics(uint64_t slotIndex);
     void recordSCPEvent(uint64_t slotIndex, bool isNomination);
+    void recordSCPExternalizeEvent(uint64_t slotIndex, NodeID const& id,
+                                   bool forceupdateself);
```
marta-lokhova: nit: we typically use camel case

src/herder/HerderSCPDriver.cpp
```diff
@@ -933,6 +945,48 @@ HerderSCPDriver::recordSCPEvent(uint64_t slotIndex, bool isNomination)
...
+    {
+        timing.mFirstExternalize = make_optional<VirtualClock::time_point>(now);
+    }
+
+    bool self = (id == mSCP.getLocalNodeID());
+    if (self)
+    {
+        if (!timing.mSelfExternalize)
+        {
+            if (timing.mFirstExternalize)
```
marta-lokhova: `timing.mFirstExternalize` is either present already or set on line 959, so I don't think you need this check.

docs/metrics.md
```diff
@@ -116,8 +116,10 @@ scp.pending.ready                        | counter   | number of envelopes ready
 scp.sync.lost                            | meter     | validator lost sync
 scp.timeout.nominate                     | meter     | timeouts in nomination
 scp.timeout.prepare                      | meter     | timeouts in ballot protocol
-scp.timing.externalized                  | timer     | time spent in ballot protocol
 scp.timing.nominated                     | timer     | time spent in nomination
+scp.timing.externalized                  | timer     | time spent in ballot protocol
+scp.timing.externalize-lag               | timer     | delay between first externalize message received and node externalizing
+scp.timing.externalize-delay             | timer     | delay between node externalizing and other externalize messages
```
marta-lokhova: Can we slightly tweak this description to make it clear that "other externalize messages" are the externalize messages received after the node itself externalizes?

src/herder/HerderSCPDriver.cpp
MonsieurNicolas: yeah I wanted to use `first` here (so bug), but I think it's better to always capture. so I removed this condition

docs/metrics.md
MonsieurNicolas: done

marta-lokhova: Looks like the description is still the same?

MonsieurNicolas: I thought you were asking about "node externalizing" that I clarified as "local node externalizing", but yeah I replaced "other" with "later"

marta-lokhova:
r+ 8b0e5d9de72168e04c9f5d51b9621c02594c00db

PR #2471, opened 2020-03-17
prefetch hit-rate metric updates

### LINKED ISSUE #2470: Improve Prefetch Hit-Rate Metric
The current prefetch hit-rate metric is an average over the entire time that the node has been running. After the node has been running for several days, it becomes essentially constant. It would be far more useful if this were a measure of the recent (last 5 minutes, for example) prefetch hit-rate so variations could more easily be correlated against other events.

marta-lokhova:
Resolves #2470 

With this change, we now collect prefetch hit-rates per ledger, and track them via a histogram (5-minute sliding window)

### DIFF
--- a/src/ledger/LedgerTxn.cpp
+++ b/src/ledger/LedgerTxn.cpp
@@ -2123,6 +2123,9 @@ LedgerTxnRoot::Impl::commitChild(EntryIterator iter, LedgerTxnConsistency cons)
     // std::unique_ptr<...>::swap does not throw
     mHeader.swap(childHeader);
     mChild = nullptr;
+
+    mPrefetchHits = 0;
+    mPrefetchMisses = 0;
 }
 
 std::string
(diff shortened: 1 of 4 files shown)


src/ledger/LedgerTxn.cpp
```diff
@@ -2123,6 +2123,9 @@ LedgerTxnRoot::Impl::commitChild(EntryIterator iter, LedgerTxnConsistency cons)
     // std::unique_ptr<...>::swap does not throw
     mHeader.swap(childHeader);
     mChild = nullptr;
+
+    mPrefetchHits = 0;
```
jonjove: I think you missed the same in `LedgerTxnRoot::Impl::rollbackChild`.

marta-lokhova: Good catch, updated.

MonsieurNicolas:
r+ edc77e4892fab7d9d785cbc945d9e9b94cb81fe2

PR #2488, opened 2020-03-31
Tx queue delay instrumentation

### LINKED ISSUE #2469: rethink mempool instrumentation
Right now the mempool reports the following metrics:
number of transaction in each generation aka age with "herder.pending-txs.age<N>".

While this is maybe interesting, it doesn't actually give information on the impact on end users, just a vague idea on the node's backlog.

Proposed update: as ledgers are denominated in operations, we should probably update those metrics to be in operations and not transactions.

I think we should also add the following metrics.

## Number of transactions getting banned

I think "banned" is enough to capture transactions that were added to the mempool but got dropped *without* being included in a ledger.

This can happen for the following reasons (we should have logs to get this break down):
* transactions sitting in the mempool for too long (traffic too high vs inbound, surge pricing)
* mempool reached capacity
* transaction getting dropped because of conflicting transactions getting included in a ledger (account running out of funds, getting deleted, invalidating the sequence number, etc)

## Inclusion lag

Here we would measure the time spent between the events "inserted in mempool" and "included in a ledger" (updated every ledger).

I am not sure we'd need any other additional logging for this scenario.

MonsieurNicolas:
# Description

Resolves #2469 

This PR updates the transaction queue instrumentation to track items in a more meaningful way.

# Checklist
- [ ] Compiles
- [ ] Ran all tests

### DIFF
--- a/src/herder/TransactionQueue.cpp
+++ b/src/herder/TransactionQueue.cpp
@@ -17,6 +17,7 @@
 #include <lib/util/format.h>
 #include <medida/meter.h>
 #include <medida/metrics_registry.h>
+#include <medida/timer.h>
 #include <numeric>
 
 namespace stellar
@@ -33,6 +34,8 @@ TransactionQueue::TransactionQueue(Application& app, int pendingDepth,
                          .header.ledgerVersion)
     , mBannedTransactionsCounter(
           app.getMetrics().NewCounter({"herder", "pending-txs", "banned"}))
+    , mTransactionsDelay(
+          app.getMetrics().NewTimer({"herder", "pending-txs", "delay"}))
     , mPoolLedgerMultiplier(poolLedgerMultiplier)
 {
     for (auto i = 0; i < pendingDepth; i++)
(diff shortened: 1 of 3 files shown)


src/herder/TransactionQueue.cpp
```diff
@@ -240,14 +244,14 @@ TransactionQueue::tryAdd(TransactionFrameBasePtr tx)
...
     if (oldTxIter != stateIter->second.mTransactions.end())
     {
-        releaseFeeMaybeEraseAccountState(*oldTxIter);
-        stateIter->second.mQueueSizeOps -= (*oldTxIter)->getNumOperations();
-        mQueueSizeOps -= (*oldTxIter)->getNumOperations();
-        *oldTxIter = tx;
+        releaseFeeMaybeEraseAccountState(oldTxIter->mTx);
+        stateIter->second.mQueueSizeOps -= oldTxIter->mTx->getNumOperations();
+        mQueueSizeOps -= oldTxIter->mTx->getNumOperations();
+        *oldTxIter = {tx, mApp.getClock().now()};
```
jonjove, line 250: I'm unsure about this. Is this conceptually what we are trying to measure here? Specifically, this allows the slightly unusual property that the timestamps for a single account might not be monotonically increasing. Do we want to measure "how long did it take for specific transactions to get included" or "how long did it take for specific combination of source account and sequence number to get included"?

src/herder/TransactionQueue.cpp
```diff
@@ -378,7 +380,10 @@ TransactionQueue::ban(Transactions const& banTxs)
     {
         auto& transactions = transactionsByAccount[tx->getSourceID()];
         transactions.emplace_back(tx);
-        bannedFront.emplace(tx->getFullHash());
+        if (bannedFront.emplace(tx->getFullHash()).second)
```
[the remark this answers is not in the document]
MonsieurNicolas: I thought about this too but decided to not to it: those methods would still not stop people from manipulating the queue directly, so if we really wanted to do this, we would have to create a wrapper class of sorts that enforces this contract.
Given that "banning" is the simplest part of this class and that all code is internal to this class, I decided to not do this for now.

src/herder/TransactionQueue.cpp
MonsieurNicolas: per the issue and metric definition, it's tracking how long a specific transaction has been sitting in the queue. The notion of "sequence number slot" is irrelevant per this definition.

jonjove:
r+ 6daac5e

jonjove:
@latobarita: retry

PR #2491, opened 2020-04-04
bump overlay versions to min=11 cur=13

MonsieurNicolas:
# Description

This PR bumps overlay to version 13 (as we'll want a clean cut when the network switches to protocol 13).

It also updates the minimum version to only support clients more recent than 12.3.0 (3 months old when 13 goes out the door).

# Checklist
- [ ] Compiles
- [ ] Ran all tests


graydon:
r+ fa94e71

MonsieurNicolas:
Added an extra commit that drops support for peers that don't support surveying (as this is not possible with the bump of min protocol)

MonsieurNicolas:
r+ 630a16a25d7b43a726b5c83fef97d1513e099d32

PR #2493, opened 2020-04-06
Retry interrupted io

graydon:
# Description

This PR improves the XDRStream type to accommodate its use in streaming LedgerCloseMeta records over a pipe to a host process when running in "captive" mode (as in https://github.com/stellar/go/pull/2322)

The following improvements are made:
  - To deal with the fact that pipe writes fail with EINTR sometimes -- and that this is specified as leaving the contents of `FILE*` write buffers in an undefined state, which we've in fact observed as data corruption when streaming large test ranges -- we switch to using asio sychronous stream types (`asio::posix::stream_descriptor` and `asio::windows::stream_handle`).
  - To deal with the fact that these types need to be associated with an `asio::io_context`, one is threaded through all the construction sites. Occasionally a new one is constructed in a test.
  - To deal with the fact that these types can still be interrupted when writing (fail with `asio::error::interrupted`) the write calls are turned into loops that test for this code; the advantage is that they do _not_ invalidate their buffers (thereby corrupting output) when interrupted.
  - To deal with the fact that on windows only named pipes should be streams, and files need to use the `asio::windows::random_access_handle` type to avoid writing all output at the same offset (0) the `XDRStream` type is given _two_ different handles, one stream and one random access, and a crude test is made on the filename when opening on windows to see if it's in the named-pipe namespace or not.

This is all somewhat ridiculous but it does allow us to pass all tests and stream to a parent horizon process on both linux and windows.

# Checklist
- [x] Compiles
- [x] Ran all tests

### DIFF
--- a/Builds/VisualStudio/stellar-core.vcxproj
+++ b/Builds/VisualStudio/stellar-core.vcxproj
@@ -273,6 +273,7 @@ exit /b 0
     <ClCompile Include="..\..\src\catchup\ApplyLedgerWork.cpp" />
     <ClCompile Include="..\..\src\catchup\CatchupConfiguration.cpp" />
     <ClCompile Include="..\..\src\catchup\CatchupManagerImpl.cpp" />
+    <ClCompile Include="..\..\src\catchup\CatchupRange.cpp" />
     <ClCompile Include="..\..\src\catchup\CatchupWork.cpp" />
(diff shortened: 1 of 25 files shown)


MonsieurNicolas:
r+ a56c3278fff08de64d460e5476c89eb9b6dce99d

PR #2506, opened 2020-04-20
targeted scheduler fix: avoid throttling work when STANDALONE

graydon:
This adds a small, temporary, targeted fix to the scheduling mix that avoids throttling-down work when we're running STANDALONE (eg. when we're doing command-line catchup).


MonsieurNicolas:
r+ da1ffff3a628712c2118e946375634ba629878c5

PR #2508, opened 2020-04-22
Command line publish: forget unreferenced buckets when finished

### LINKED ISSUE #2507: Publish sometimes leaves stale files
Observed while running simulations which utilize `publish` command. The two problems are:
* `publish` does not forget unreferenced buckets. Typically, this is done when closing ledgers, but since the command just runs publishing in a loop, cleanup never happens
* Some stale zipped files are left in the buckets directory. This is probably related to one of the underlying works not cleaning up zipped files

marta-lokhova:
Resolves #2507

### DIFF
--- a/src/main/ApplicationUtils.cpp
+++ b/src/main/ApplicationUtils.cpp
@@ -4,6 +4,7 @@
 
 #include "main/ApplicationUtils.h"
 #include "bucket/Bucket.h"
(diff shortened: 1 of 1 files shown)


MonsieurNicolas:
r+ a302a624f215258ab561d2fdefed1ef29775e1d8

PR #2510, opened 2020-04-23
Add shutdown in ApplyBufferedLedgersWork

### LINKED ISSUE #2509: ApplyBufferedLedgersWork does not shutdown correctly
It looks like ApplyBufferedLedgersWork is missing implementations of `shutdown` and `onAbort` (returns true currently). The work maintains a conditional work, so it must call `shutdown` on it and keep cranking it until done in `onAbort`. Reason we don't see the problem in tests is because on SIGINT we don't go through graceful application shutdown, but call `exit` instead.

marta-lokhova:
Resolves #2509


MonsieurNicolas:
r+ 704d89a2084dbcad02c98187becea2cdb409aa3e

PR #2513, opened 2020-04-26
Add scripts/Dockerfile.testing for one-off test images

graydon:
# Description

This adds a Dockerfile for building one-off test images. It lets developers bypass the CI/CD system we have for official images (based on signed debian packages) and instead just scoops up the current source directory and builds it into a deployable stellar-core for testing in a k8s sandbox.

This is different from the existing `.devcontainer/Dockerfile` -- that file builds a container which has tools, compilers and libraries _with which you can build_ `stellar-core`.

This PR's `scripts/Dockerfile.testing` is a multi-step build: it first builds such a dev container, then builds `stellar-core` in it, then builds a second _deployment_ container and copies-forward the runnable `stellar-core` built in the first step, so that it can be tagged and executed directly on your local machine's k3s.

(An earlier version of this PR used alpine linux for size, but that seems not to get along with k3s; for compatibility and uniformity sake I've moved it back to using ubuntu)


.dockerignore
```diff
@@ -0,0 +1,107 @@
+**/*.a
```
MonsieurNicolas, line 1: is it possible to just symlink to .gitignore?

MonsieurNicolas (changes requested):

.dockerignore
graydon: Unfortunately they use [subtly incompatible formats](https://zzz.buzz/2018/05/23/differences-of-rules-between-gitignore-and-dockerignore/). Also they identify slightly different sets of files (I excluded some additional testing stuff that technically _is_ under revision control but it's not worth copying over).

scripts/Dockerfile.testing
```diff
@@ -0,0 +1,56 @@
...
+
+ARG CC=clang-8
+ARG CXX=clang++-8
+ARG CFLAGS='-O3 -g1 -fno-omit-frame-pointer'
+ARG CXXFLAGS='-O3 -g1 -fno-omit-frame-pointer -stdlib=libc++'
+
+RUN ./autogen.sh
+RUN ./configure CC="${CC}" CXX="${CXX}" CFLAGS="${CFLAGS}" CXXFLAGS="${CXXFLAGS}"
+RUN sh -c 'make -j $(nproc)'
+RUN make install
```
hidenori-shinohara, line 45: I was wondering if it'd make sense to put `-j $(nproc)` for install as well.

graydon: possibly? install is very fast -- doesn't really matter.

MonsieurNicolas:
I think you need to bump it to llvm-10 @graydon

graydon:
@MonsieurNicolas ah good catch, fixed!

MonsieurNicolas:
r+ af7e3326fe8a8c9a7cd21fda5a596e250b22dcfa

PR #2516, opened 2020-05-01
database: add PostgreSQL "notice receiver"

### LINKED ISSUE #191: implement a new "notice processor" in postgreSQL to intercept warnings
Per http://www.postgresql.org/docs/9.4/static/libpq-exec.html

A result of status PGRES_NONFATAL_ERROR will never be returned directly by PQexec or other query execution functions; results of this kind are instead passed to the notice processor 

There are a couple issues with this approach:
- default handler spews out to stderr instead of using our logger
- some errors deemed "non fatal" are invisible to the caller (they register as PGRES_COMMAND_OK). "nested BEGIN statements fall into this category even though the behavior is just wrong. We may want to just fail all the time in the presence of a warning.

santegoeds:
# Registers a notice receiver with PostgreSQL sessions.

With reference to #191 

PostgreSQL can emit notice messages with severity INFO, NOTICE or WARNING that per default are written to `stderr`.  The receiver in this PR forwards the message to the stellar-core log with level WARNING for component "Database".

While running tests *only*; an `soci::soci_error` for notice messages with severity WARNING is thrown.

See https://www.postgresql.org/docs/current/libpq-notice-processing.html for further details about PostgreSQL notice processing.

### DIFF
--- a/src/database/Database.cpp
+++ b/src/database/Database.cpp
@@ -92,7 +92,17 @@ badPgVersion(int vers)
         << '.' << MIN_POSTGRESQL_MINOR_VERSION;
     return msg.str();
 }
-#endif
+
+static void
+pgNoticeReceiver(void* arg, const PGresult* res)
+{
+    if (res == nullptr)
+    {
+        return;
+    }
+    CLOG(WARNING, "Database") << PQresultErrorMessage(res);
+}
+#endif // USE_POSTGRES
 
 static std::string
 badSqliteVersion(int vers)
@@ -168,6 +178,9 @@ class DatabaseConfigureSessionOp : public DatabaseTypeSpecificOperation<void>
         {
             throw std::runtime_error(badPgVersion(vers));
         }
+
+        PQsetNoticeReceiver(pg->conn_, pgNoticeReceiver, nullptr);
+
         mSession
             << "SET SESSION CHARACTERISTICS AS TRANSACTION ISOLATION LEVEL "
                "SERIALIZABLE";
--- a/src/database/test/DatabaseTests.cpp
+++ b/src/database/test/DatabaseTests.cpp
@@ -244,6 +244,51 @@ TEST_CASE("postgres smoketest", "[db]")
             app->getDatabase().getSession() << "drop table if exists test";
             checkMVCCIsolation(app);
         }
+
+        SECTION("postgres notify")
+        {
+            auto pg = static_cast<soci::postgresql_session_backend*>(
+                session.get_backend());
+
+            int callCount = 0;
+            auto savedPgNoticeReceiver =
+                PQsetNoticeReceiver(pg->conn_,
+                                    [](void* arg, const PGresult* res) {
+                                        if (res == nullptr)
+                                        {
+                                            return;
+                                        }
+                                        CLOG(WARNING, "Database")
+                                            << PQresultErrorMessage(res);
+
+                                        int* cc = static_cast<int*>(arg);
+                                        ++(*cc);
+                                    },
+                                    &callCount);
+
+            try
+            {
+                session << "do $$ "
+                           "begin "
+                           "raise info 'a test message'; "
+                           "end; "
+                           "$$;";
+
+                session << "drop table if exists non_existing_table;";
+
+                session << "begin; begin; "
+                           "select 1; "
+                           "end;";
+
+                PQsetNoticeReceiver(pg->conn_, savedPgNoticeReceiver, nullptr);
+                REQUIRE(callCount == 3);
+            }
+            catch (...)
+            {
+                PQsetNoticeReceiver(pg->conn_, savedPgNoticeReceiver, nullptr);
+                throw;
+            }
+        }
     }
     catch (soci::soci_error& err)
     {


MonsieurNicolas:
ah cool thank you.

I think the original issue was suggesting causing the call sites to fail (so throw?) but this change should allow us to see if this makes sense or not (change your description above so that the issue does not get resolved).

I see that you're capturing the value of `PG_DIAG_SEVERITY` in test mode. Should we add it to the logged error message in non test mode (I don't know if the error message contains its value or not all the time)?

santegoeds:
Updated the PR description...

I considered to have the call sites fail and at first implemented it that way; but then reconsidered.

The reason is that I don't consider the notice messages part of the PostgreSQL API and I don't like the idea that changing the release of PostgreSQL causes stellar-core to fail.

The logged message already contains the value of `PG_DIAG_SEVERITY`.  See the output of the section of the "postgres smoketest" below...

Notice that the message for warnings will "stutter" when not in test mode.  But I decided to leave the message as generated by postgres because it will be more familiar for someone with postgres experience.

I just now notice that the messages have 2 newlines.  One from the postgres and another one from stellar-core.  Let me know if you'd like me to trim them.

Lastly, I considered logging the messages with "Postgres" instead of "Database" to make it more explicit that the message originates not from stellar-core, but from an external source.  But I didn't want to break some (informal) convention.

**log extract from postgres smoketest**
```
2020-05-03T10:27:17.852 <test> [Database WARNING] INFO:  a test message

2020-05-03T10:27:17.852 <test> [Database WARNING] NOTICE:  table "non_existing_table" does not exist, skipping

2020-05-03T10:27:17.852 <test> [default INFO] WARNING:  there is already a transaction in progress

```

MonsieurNicolas:
yes this makes sense. Thank you. Merging!

MonsieurNicolas:
r+ 8e8477bd09368ca2429994540982b5aacb744f80

MonsieurNicolas:
Weird. Looks like this test doesn't pass in travis.

```
1	"postgres smoketest" database/test/DatabaseTests.cpp:204
local:2/29/100%/4.2s 1	........
1~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
1	stellar-core test is a Catch v2.11.3 host application.
1	Run with -? for options
1	
1	Randomness seeded to: 18387
1	
1	-------------------------------------------------------------------------------
1	postgres smoketest
1	  postgres notify
1	-------------------------------------------------------------------------------
1	database/test/DatabaseTests.cpp:248
1	...............................................................................
1	
1	database/test/DatabaseTests.cpp:269: FAILED:
1	  REQUIRE( 0 )
1	
1	<done>
```

santegoeds:
I have added another commit for this.

The notice is delivered as part of the statement result.  The last statement did not wait for a result so my expectation is that in some cases the test asserted that an exception was thrown before the result was received.

This last commit collects the result of the statement before asserting.  This should assure that the notice is always received before the test assertion is executed.

Let me know if I should squash the commits...

MonsieurNicolas:
Yeah that's fine, no need to squash

MonsieurNicolas:
r+ 945bb8116e78a027e74c2f494ab119ce418f10dc

MonsieurNicolas:
you forgot to run `clang-format` (so you can squash commits and do it)

santegoeds:
Apologies... formatted and squashed..

MonsieurNicolas:
r+ 20de23fcb9901ebc62beafa4b81a7a35b14754d2

santegoeds:
So it failed again. :-(

Same compiler; I wonder if that is a coincidence.  I will see if I can reproduce this or otherwise can find a possible cause.

MonsieurNicolas:
I wonder if the problem is that it's actually not safe to throw an exception from those handlers: who knows the kind of cleanup / code that they're doing within the postgres libraries.

As you only use this for testing, the better option might be to set a different handler for those specific tests.

santegoeds:
I expect that you are correct. I will try to dig into the code that calls the handler during the weekend.

MonsieurNicolas:
@santegoeds probably worth diving into postgres' code: even if it was true today, there is no guarantee that it will stay safe in the future. Basically, I would thread this code as some opaque C library (so can't throw in the callback)

santegoeds:
Too late because I already took a look during the weekend. :-)  But you are correct - the approach that I'm pursuing here is a no-go.  (Well... unless `libpq` is statically linked into `stellar-core`, e.g. by adding it to `lib`)

Realising that, I've looked at some alternatives to force a postgres error for the statement that generated the warning; but haven't found any options so far.

- I thought to update the `PGresult` in the handler to force an error.  But the`PGresult` that the callback receives is a copy of the one that is checked for the status of the query.  So that doesn't work.
- Cancelling the query using `PGcancel` won't work. By the time that the callback runs the query has already finished.
- The `soci` wrapper is not configurable and the details are inaccessible.  So it's not possible to force a query failure there.  At least not without patching the library or writing a separate driver just for handling these warnings.

I'm currently exploring whether it's possible to hook into `Catch`, similar to what the default posix signal handlers do.  But so far without success.

Of course it is always possible to abort/exit from the handler.  Which is ugly but it will show the issue. But even then there is the additional complication that the warning is communicated in text form and I learned that the `WARNING` message is subject to localization.

I'll spend a little more time on the approach via `Catch`, but I expect that my recommendation will be to keep the callback just for the purpose of redirecting the notice message to the stellar log instead of having them go to `stderr`.

MonsieurNicolas:
Maybe the simplest thing to do here is just override the handler just for that test, then you can do whatever you want in there (including setting some variable). It's not exactly testing that the logging handler is getting called, but it's close enough.

santegoeds:
@MonsieurNicolas - I added a commit as per your suggestion.

Let me know if I misunderstood of if you'd like to see it implemented differently.  I'll squash the commits if you're happy with it.

MonsieurNicolas:
yes I think this should work this time. Thank you.
you can squash and I'll merge

santegoeds:
Done..

src/main/Config.h
```diff
@@ -374,7 +374,8 @@ class Config : public std::enable_shared_from_this<Config>
 #ifdef BUILD_TESTS
     // If set to true, the application will be aware this run is for a test
     // case.  This is used right now in the signal handler to exit() instead of
-    // doing a graceful shutdown
+    // doing a graceful shutdown and in the Postgresql notification receiver
+    // to throw an exception for warning notifications.
```
MonsieurNicolas, line 378: you forgot to undo this change

MonsieurNicolas:
sorry, I missed that you forgot to undo the comment in config.h during the last round of updates

santegoeds:
Good catch, but I should've reviewed the diff before pushing.  Comments reverted too.

santegoeds:
@MonsieurNicolas - I'd appreciate it if you can take another look at this.

I removed some more comments that were no longer applicable and added another commit to change the approach to testing.

I don't like the approach in the 1st commit much because it tests `PQsetNoticeReceiver` more than anything else, and I feel that belongs to the PostgreSQL domain and not Stellar-core.

Tbh - the approach in the 2nd commit is only marginally different.  So please let me know which one you prefer and I can either squash or drop the 2nd commit accordingly.

This time I've double-check the PR twice to make sure I haven't left any invalid comments in *and* I've made sure to run `make format`.

MonsieurNicolas:
as both commits require changing the handler, I'd stick with the first one that doesn't have test code mixed with regular code.

santegoeds:
:+1: I dropped the extra commit...

MonsieurNicolas:
r+ 1005b7a187a9f83fd024b8c5ae2f4e94824523a4

MonsieurNicolas:
Looks like it's the same failure than before...
```
database/test/DatabaseTests.cpp:284: FAILED:
2	  REQUIRE( callCount == 3 )
2	with expansion:
2	  1 == 3
```

to repro this, you may have to spin up a docker container with the same version of Ubuntu+Postgres.

santegoeds:
Pff...  that boggles my mind.  I'll follow your suggestion and see if I can understand.

santegoeds:
Closing this for now.  I've had little time to work on this recently but the attempts I did make failed to reproduce the issue locally.

I'll create a new PR when and if I've reproduced and solved the issue.

MonsieurNicolas:
sorry about this @santegoeds ; we're going to update our toolchain in the next few weeks, this may make this easier to diagnose

PR #2521, opened 2020-05-06
Add minimum Fee Bump replacement fee to `POST /tx`'s  response

### LINKED ISSUE #2505: Add Fee-Recommendation to Transaction Submission Response
After the upgrade to protocol 13, it will be possible to submit fee-bump transactions. Fee-bump transactions will participate in replace-by-fee semantics with back-off. It is possible, as a consequence, to submit a fee-bump transaction to stellar-core and have it rejected as invalid due to an insufficiently large ratio between its fee and the fee of a transaction in the queue. It is currently difficult (perhaps impossible) to know what fee to submit if this happens, so we should add information about this in the response.

2opremio:
# Description

Fixes #2505

When the fee of a Fee Bump transaction isn't high enough to satisfy the replace-by-fee requirement, include ~a fee-rate hint to~ the minimum Fee Bump replacement fee to:

1. Be more precise on why the fee wasn't enough.
2. Give the user an opportunity to recreate a transaction with an acceptable fee (although the fee requirement is dynamic, so, there are no guarantees).

The ~hint~ minimum fee is represented as an ~rational number (i.e. providing a numerator and denominator)~  integer in the JSON response. Here is an example:

```
"min_fee": 11,
```

This approach is reactive (i.e. the user needs to try submitting a transaction in order get ~a hint~ the minimum).

Alternatively we could add a separate endpoint (or a query parameter) to only check the fee required. I think being reactive is sufficient and adding a preliminary check adds unnecessary complexity, but, I am happy to be proven wrong.

I am really looking forward to your feedback. This is my first Core PR and I haven't done C++ in 12 years, so be prepared to read cringy code.

Tests will be added once a get a GA on the approach.

# Checklist
- [x] Compiles
- [x] Ran all tests

### DIFF
--- a/src/herder/TransactionQueue.cpp
+++ b/src/herder/TransactionQueue.cpp
@@ -60,7 +60,8 @@ canReplaceByFee(TransactionFrameBasePtr tx, TransactionFrameBasePtr oldTx)
     // FEE_MULTIPLIER * v2 does not overflow uint128_t because fees are bounded
     // by INT64_MAX, while number of operations and FEE_MULTIPLIER are small.
     auto v1 = bigMultiply(newFee, oldNumOps);
-    auto v2 = uint128_t(TransactionQueue::FEE_MULTIPLIER) * bigMultiply(oldFee, newNumOps);
+    auto v2 = uint128_t(TransactionQueue::FEE_MULTIPLIER) *
+              bigMultiply(oldFee, newNumOps);
     std::pair<bool, int64_t> result;
     result.first = v1 >= v2;
     if (!result.first)
@@ -124,10 +125,11 @@ TransactionQueue::canAdd(TransactionFrameBasePtr tx,
                          TimestampedTransactions::iterator& oldTxIter)
 {
     TransactionQueue::AddResult result;
-    result.mFeeRecommendation = 0;
+    result.mFeeBumpMinFee = 0;
     if (isBanned(tx->getFullHash()))
     {
-        result.mStatus =TransactionQueue::AddStatus::ADD_STATUS_TRY_AGAIN_LATER;
+        result.mStatus =
+            TransactionQueue::AddStatus::ADD_STATUS_TRY_AGAIN_LATER;
         return result;
     }
 
@@ -149,7 +151,8 @@ TransactionQueue::canAdd(TransactionFrameBasePtr tx,
                 if (findBySeq(tx, transactions, iter) &&
                     iter != transactions.end() && isDuplicateTx(iter->mTx, tx))
                 {
-                    result.mStatus = TransactionQueue::AddStatus::ADD_STATUS_DUPLICATE;
+                    result.mStatus =
+                        TransactionQueue::AddStatus::ADD_STATUS_DUPLICATE;
                     return result;
                 }
 
@@ -160,7 +163,8 @@ TransactionQueue::canAdd(TransactionFrameBasePtr tx,
                 if (!findBySeq(tx, transactions, oldTxIter))
                 {
                     tx->getResult().result.code(txBAD_SEQ);
-                    result.mStatus= TransactionQueue::AddStatus::ADD_STATUS_ERROR;
+                    result.mStatus =
+                        TransactionQueue::AddStatus::ADD_STATUS_ERROR;
                     return result;
                 }
 
@@ -169,15 +173,17 @@ TransactionQueue::canAdd(TransactionFrameBasePtr tx,
                     // Replace-by-fee logic
                     if (isDuplicateTx(oldTxIter->mTx, tx))
                     {
-                        result.mStatus = TransactionQueue::AddStatus::ADD_STATUS_DUPLICATE;
+                        result.mStatus =
+                            TransactionQueue::AddStatus::ADD_STATUS_DUPLICATE;
                         return result;
                     }
                     auto replace = canReplaceByFee(tx, oldTxIter->mTx);
                     if (!replace.first)
                     {
                         tx->getResult().result.code(txINSUFFICIENT_FEE);
-                        result.mStatus = TransactionQueue::AddStatus::ADD_STATUS_BAD_REPLACE_BY_FEE;
-                        result.mFeeRecommendation = replace.second;
+                        result.mStatus = TransactionQueue::AddStatus::
+                            ADD_STATUS_BAD_REPLACE_BY_FEE;
+                        result.mFeeBumpMinFee = replace.second;
                         return result;
                     }
 
@@ -199,7 +205,8 @@ TransactionQueue::canAdd(TransactionFrameBasePtr tx,
     if (netOps + mQueueSizeOps > maxQueueSizeOps())
     {
         ban({tx});
-        result.mStatus = TransactionQueue::AddStatus::ADD_STATUS_TRY_AGAIN_LATER;
+        result.mStatus =
+            TransactionQueue::AddStatus::ADD_STATUS_TRY_AGAIN_LATER;
         return result;
     }
 
--- a/src/herder/TransactionQueue.cpp
+++ b/src/herder/TransactionQueue.cpp
@@ -45,7 +45,7 @@ TransactionQueue::TransactionQueue(Application& app, int pendingDepth,
     }
 }
 
-static std::pair<bool, TransactionQueue::FeeRateRecommendation>
+static std::pair<bool, int64_t>
 canReplaceByFee(TransactionFrameBasePtr tx, TransactionFrameBasePtr oldTx)
 {
     int64_t newFee = tx->getFeeBid();
@@ -60,15 +60,16 @@ canReplaceByFee(TransactionFrameBasePtr tx, TransactionFrameBasePtr oldTx)
     // FEE_MULTIPLIER * v2 does not overflow uint128_t because fees are bounded
     // by INT64_MAX, while number of operations and FEE_MULTIPLIER are small.
     auto v1 = bigMultiply(newFee, oldNumOps);
-    auto v2 = TransactionQueue::FEE_MULTIPLIER * bigMultiply(oldFee, newNumOps);
-    std::pair<bool, TransactionQueue::FeeRateRecommendation> result;
+    auto v2 = uint128_t(TransactionQueue::FEE_MULTIPLIER) * bigMultiply(oldFee, newNumOps);
+    std::pair<bool, int64_t> result;
     result.first = v1 >= v2;
     if (!result.first)
     {
-        // Add a fee rate recommendation for future attempts
-        // TODO: overflow?
-        result.second.n = TransactionQueue::FEE_MULTIPLIER * oldFee;
-        result.second.d = oldNumOps;
+        // Add a fee recommendation for future attempts
+        // recommendation = FEE_MULTIPLIER * oldFeeRate * newNumOps
+        //                = FEE_MULTIPLIER * (oldFee / oldNumOps) * newNumOps
+        //                = (FEE_MULTIPLIER * oldFee * newNumOps) / oldNumOps
+        result.second = bigDivide(v2, oldNumOps, ROUND_UP);
     }
     return result;
 }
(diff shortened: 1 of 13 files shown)


jonjove:
Thanks for submitting this! I don't really see any big C++ issues here @2opremio. However, I do have a few comments/questions about the PR:

1. It doesn't build on my local machine (errors in HerderTests.cpp)
2. I think we should probably add a new `AddResult` enumeration for this bad replace-by-fee case with a name like `ADD_STATUS_BAD_REPLACE_BY_FEE` or `ADD_STATUS_REPLACE_BY_FEE_INSUFFICIENT`. The HTTP endpoint shouldn't return the `TransactionResult` in this case, it should just return the fee recommendation
3. I think you can mitigate a lot of collateral damage by having two overloads of `Herder::recvTransaction`: the one that already exists, and one that uses pass-by-reference to return the `feeRec`. Most existing code doesn't care about the `feeRec` and so will continue to use the original overload, avoiding changes in those places. The code that does care (eg. in `CommandHandler::tx`) will use the new overload with `feeRec`. In general, using pass-by-reference in this PR might simplify a lot of things.
4. What advantage do you see in returning the recommended fee-per-operation (as a rational) instead of the recommended fee (as an integer)? I imagine that in basically every single case you will want to know the recommended fee for the transaction that you submitted, rather than using it to calculate the recommended fee for some other transaction. In fact, you generally don't want to use this for other transactions because they will only have the same fee recommendation if they have the same source account and sequence number. But maybe I'm missing something here?

2opremio:
Thanks for the review!

> 1. It doesn't build on my local machine (errors in HerderTests.cpp)

I fixed that. But now I am getting a segmentation fault in one of the tests:

```
1	-------------------------------------------------------------------------------
1	LedgerCloseMetaStream file descriptor - REPLAY_IN_MEMORY
1	-------------------------------------------------------------------------------
1	ledger/test/LedgerCloseMetaStreamTests.cpp:105
1	...............................................................................
1	
1	ledger/test/LedgerCloseMetaStreamTests.cpp:105: FAILED:
1	due to a fatal error condition:
1	  SIGSEGV - Segmentation violation signal
1	
1	<done>
```

I doubt my new could has caused incorrect memory accesses. Any suggestions on how to debug this?

> 2. I think we should probably add a new `AddResult` enumeration for this bad replace-by-fee case with a name like `ADD_STATUS_BAD_REPLACE_BY_FEE` or `ADD_STATUS_REPLACE_BY_FEE_INSUFFICIENT`. The HTTP endpoint shouldn't return the `TransactionResult` in this case, it should just return the fee recommendation

Done

> 3. I think you can mitigate a lot of collateral damage by having two overloads of `Herder::recvTransaction`: the one that already exists, and one that uses pass-by-reference to return the `feeRec`. Most existing code doesn't care about the `feeRec` and so will continue to use the original overload, avoiding changes in those places. The code that does care (eg. in `CommandHandler::tx`) will use the new overload with `feeRec`. In general, using pass-by-reference in this PR might simplify a lot of things.

I've done this, but I am not sure it's an improvement (the PR has become even larger).

> 4. What advantage do you see in returning the recommended fee-per-operation (as a rational) instead of the recommended fee (as an integer)? I imagine that in basically every single case you will want to know the recommended fee for the transaction that you submitted, rather than using it to calculate the recommended fee for some other transaction. In fact, you generally don't want to use this for other transactions because they will only have the same fee recommendation if they have the same source account and sequence number. But maybe I'm missing something here?

After thinking about it, there is no advantage. I changed the code.

MonsieurNicolas:
great.

wrt what we return:
* agree with Jon, there is no need to expose fee rate or some new notation for fees. If what we return is the smallest fee amount that would allow the fee bump to be accepted, we can just use an amount.
   *  (I see that you fixed that already)
* (minor) we should not call this a recommendation internally or in the contract exposed to the outside, it's a statement about the current state. I'd suggest using terminology like `feeBumpAcceptableFee` or `feeBumpMinFee`. A recommendation could be derived from this at the SDK level by applying an extra multiplier on top but this is not a core concern.

src/herder/TransactionQueue.h
```diff
@@ -59,13 +59,20 @@ class TransactionQueue
...
-    enum class AddResult
+    enum class AddStatus
     {
         ADD_STATUS_PENDING = 0,
         ADD_STATUS_DUPLICATE,
         ADD_STATUS_ERROR,
         ADD_STATUS_TRY_AGAIN_LATER,
-        ADD_STATUS_COUNT
+        ADD_STATUS_COUNT,
+        ADD_STATUS_BAD_REPLACE_BY_FEE,
```
MonsieurNicolas, line 69: this seems wrong: `ADD_STATUS_COUNT` is used as the marker for the upper bound for the enum values

2opremio: Good catch, thanks. It should be fixed now

src/simulation/LoadGenerator.cpp
```diff
@@ -630,14 +630,14 @@ LoadGenerator::TxInfo::execute(Application& app, bool isCreate,
     msg.transaction() = txf->getEnvelope();
     txm.mTxnBytes.Mark(xdr::xdr_argpack_size(msg));
 
-    auto status = app.getHerder().recvTransaction(txf);
-    if (status != TransactionQueue::AddResult::ADD_STATUS_PENDING)
+    auto result = app.getHerder().recvTransaction(txf);
```
jonjove, line 633: Nitpick: why did you rename this from `status` to `result`? It is antagonistic to your other change of renaming `AddResult` to `AddStatus`.

src/overlay/Peer.cpp
```diff
@@ -784,15 +784,16 @@ Peer::recvTransaction(StellarMessage const& msg)
     {
         // add it to our current set
         // and make sure it is valid
-        auto recvRes = mApp.getHerder().recvTransaction(transaction);
+        auto recvResStatus = mApp.getHerder().recvTransaction(transaction);
```
jonjove, line 787: Nitpick: The name reads as "receive result status" which is weird. How about just `recvStatus`?

src/main/CommandHandler.cpp
```diff
@@ -568,22 +568,28 @@ CommandHandler::tx(std::string const& params, std::string& retStr)
...
             }
-
-            output << "{"
-                   << "\"status\": "
-                   << "\"" << TX_STATUS_STRING[static_cast<int>(status)]
-                   << "\"";
-            if (status == TransactionQueue::AddResult::ADD_STATUS_ERROR)
+            break;
+            case TransactionQueue::AddStatus::ADD_STATUS_BAD_REPLACE_BY_FEE:
+                output << " , \"min_fee\": " << feeBumpMinFee;
```
jonjove, line 590: Do we need to use a string here? Otherwise, will this work correctly with languages that don't support `int64` natively?

src/herder/TransactionQueue.cpp
```diff
@@ -60,8 +60,19 @@ canReplaceByFee(TransactionFrameBasePtr tx, TransactionFrameBasePtr oldTx)
...
     auto v1 = bigMultiply(newFee, oldNumOps);
-    auto v2 = bigMultiply(oldFee, newNumOps);
-    return v1 >= TransactionQueue::FEE_MULTIPLIER * v2;
+    auto v2 = uint128_t(TransactionQueue::FEE_MULTIPLIER) *
+              bigMultiply(oldFee, newNumOps);
+    std::pair<bool, int64_t> result;
+    result.first = v1 >= v2;
+    if (!result.first)
+    {
+        // Add a fee recommendation for future attempts
```
jonjove, line 69: As @MonsieurNicolas pointed out, we shouldn't use the term "fee recommendation" anywhere.

src/herder/TransactionQueue.cpp
```diff
@@ -60,8 +60,19 @@ canReplaceByFee(TransactionFrameBasePtr tx, TransactionFrameBasePtr oldTx)
...
+        // recommendation = FEE_MULTIPLIER * oldFeeRate * newNumOps
+        //                = FEE_MULTIPLIER * (oldFee / oldNumOps) * newNumOps
+        //                = (FEE_MULTIPLIER * oldFee * newNumOps) / oldNumOps
+        result.second = bigDivide(v2, oldNumOps, ROUND_UP);
```
MonsieurNicolas, line 73: Can we maybe reorganize the code a bit differently:
only one code path cares about this computation, so we could move it to a function `computeMinFee` of its own that gets called by the caller at the top if it runs into the `txINSUFFICIENT_FEE` error (see my other commend in the `tx` endpoint, related).

Reason this came up during my review is that `bigDivide` can throw and I'd hate to crash the other code paths.

src/main/CommandHandler.cpp
```diff
@@ -568,22 +568,28 @@ CommandHandler::tx(std::string const& params, std::string& retStr)
...
                 mApp.getOverlayManager().broadcastMessage(msg);
             }
-
-            output << "{"
-                   << "\"status\": "
-                   << "\"" << TX_STATUS_STRING[static_cast<int>(status)]
-                   << "\"";
-            if (status == TransactionQueue::AddResult::ADD_STATUS_ERROR)
+            break;
+            case TransactionQueue::AddStatus::ADD_STATUS_BAD_REPLACE_BY_FEE:
```
MonsieurNicolas, line 589: Sorry, I should have noticed this much earlier. I just realized something that impacts the overall structure of this PR.

I don't think we need to add a new internal status result... or even a `minFee` element in the json:

what we really want here is that if the transaction fails with `txINSUFFICIENT_FEE`, we provide the "mininum fee" required for that transaction.

For replace by fee, it's the computation that you've done, for others, it depends on context (grep for `tx`).

OK, so if we want to always return a min fee with this kind of error... how do we return it?

For that, I suspect we can simply use `feeCharged` (in the result) to propagate the min fee on error from anywhere in the code (NB: it may already be the case for the other failures), instead of having special handling for fee bump failures coming from the mempool.

It's probably safe to do as fees are always charged before applying transactions (true now and for all older protocol versions).

Having this included in the result is actually quite nice if we can do it because I imagine that we already propagate back the raw result to the SDKs so the changes to Horizon & clients should be minimal.

src/main/CommandHandler.cpp
MonsieurNicolas: +1: this is unsafe

src/main/CommandHandler.cpp
jonjove: I don’t really think this is a good idea. First, I think it would be unnecessary. If the fee is insufficient for any reason other than replace-by-fee backoff then you can just calculate the required fee using the base fee from the ledger header. In that case, stellar-core doesn’t need to give you any new information. 

Second, I think there is already a huge amount of confusion regarding the results of transaction submission and validation. Your proposal definitely wouldn’t help reduce that confusion, and it could make it worse. For example, people might wonder why stellar-core is charging a fee higher than the fee specified in the transaction (it isn’t, of course, it’s just telling you what fee would have been required). This is just a single way this could be misinterpreted. But good interfaces should be easy to use correctly and hard to use incorrectly. The interface as described in @2opremio PR is obvious and unambiguous:

- You only get that result if you submit a fee-bump (this was part of your motivation of making replace-by-fee only happen for fee-bumps) so there are no backwards compatibility issues
- The result clearly indicates what happened and cannot be misinterpreted

If you really wanted to always provide fee information whenever a transaction is rejected because the fee is bad, I would actually argue that we should add `ADD_STATUS_BAD_FEE` instead of `ADD_STATUS_BAD_REPLACE_BY_FEE` and provide information there. But again, I think this is unnecessary.

MonsieurNicolas: There are a few things here:
* the main point is that this proposal is "replace by fee" centric, and is not future proof.
* on the contract front
  * we are already returning a transaction result, and `feeCharged` is already set to some arbitrary thing. If we were to actually formally define what the value of `feeCharged` is on failure, I think it would be a net improvement over the current situation where people don't realize that it's the same.
  * that said, if we were to add a new field on failure because we're not happy with transaction result, we should just make the interface harder to use wrong (and it would not have two "fees" returned in one call). I had a proposal to remove using transaction results altogether for this kind of failure but there was not much appetite for having to support a different contract from the Horizon team (I think because it involves changing SDKs). This change requires changing SDKs etc - so maybe it's the right time.

jonjove: @2opremio After looking at the transaction result contract more carefully, I think the approach mentioned by @MonsieurNicolas is a good idea. It'll be a lot less code, and downstream systems will already be wired up for it. Long run, I still think we should be making this interface less confusing as noted in #1999. Sorry about the change of plans.

2opremio:
closing in favor of https://github.com/stellar/stellar-core/pull/2624

PR #2523, opened 2020-05-11
Clear metrics when running simulation replay

marta-lokhova:
Allow replay simulation to ramp up a bit before gathering metrics for analysis

### DIFF
--- a/src/catchup/simulation/TxSimApplyTransactionsWork.cpp
+++ b/src/catchup/simulation/TxSimApplyTransactionsWork.cpp
@@ -24,6 +24,7 @@ namespace stellar
 {
 namespace txsimulation
 {
+constexpr uint32_t const CLEAR_METRICS_AFTER_NUM_LEDGERS = 100;
 
 TxSimApplyTransactionsWork::TxSimApplyTransactionsWork(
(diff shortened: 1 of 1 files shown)


src/catchup/simulation/TxSimApplyTransactionsWork.cpp
```diff
@@ -555,9 +558,21 @@ TxSimApplyTransactionsWork::onRun()
...
-        else if (mVerifyResults)
+        else
         {
-            checkResults(mApp, header.ledgerSeq, mResults);
+            if (mVerifyResults)
+            {
+                checkResults(mApp, header.ledgerSeq, mResults);
+            }
+            auto& closed =
+                mApp.getMetrics().NewTimer({"ledger", "ledger", "close"});
```
MonsieurNicolas, line 568: Why not use a condition on lcl? We have the starting ledger, so we know where it's going to be after 100 ledgers.
This would be less error prone than relying on metric names (that change relatively often, so we'd break this and only notice much later).

marta-lokhova: Done.

MonsieurNicolas:
r+ 5d01f39cb75d073f82bd2a0a72163828d04e73fc

PR #2532, opened 2020-05-18
add an explicit error message when running into a crypto error

MonsieurNicolas:
When running into libsodium errors, we were throwing the generic "runtime_error", this PR gives them a dedicated type.


jonjove:
r+ a2f9cdd

jonjove:
@latobarita: retry

PR #2545, opened 2020-05-23
Fix how we schedule trigger timer

MonsieurNicolas:
This PR fixes a potential issue introduced in b28fca6fa71d11fd644d4e0d5e8de9bdc3a764d3 : 

in that change, we schedule starting nomination from within the callback that gets posted when a ledger closes.

The problem is that if for some reason the execution of the callback is delayed (the scheduler may decide to prioritize other actions), the node may miss its window to start nomination (timers do not get throttled the same way).

### DIFF
--- a/src/herder/HerderImpl.cpp
+++ b/src/herder/HerderImpl.cpp
@@ -653,79 +653,73 @@ HerderImpl::getTransactionQueue()
 #endif
 
 void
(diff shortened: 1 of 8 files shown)


marta-lokhova:
r+ d75aae445c2cd3b60c29733db32318370c057068

Issue #864, opened 2015-10-21
New install, Mac El Capitan, fatal error: 'libpq-fe.h' file not found


ChristopherA:
When I follow instructions at https://github.com/stellar/stellar-core/blob/master/INSTALL.md the first command that fails is "make .". Subsequent make commands result in fatal error: 'libpq-fe.h' file not found

jedmccaleb:
can you try `./configure --disable-postgres` and then `make`

olivier-w:
@jedmccaleb Thanks worked for me!

PR #2549, opened 2020-05-27
fix(INSTALL.md): fix libpq-fe.h file not found

yenchihliao:
# Description
Per description of [this](https://github.com/stellar/stellar-core/issues/864) issue. When installing Stellar-core on macOS Mojave, the solution to it indeed avoids compiling errors. However, some components can't work as expected after adopting the solution. With the line, `brew install postgresql`, instead, MacOS users can now install the system without potentially missing anything.
Resolves #X

# Checklist
- [x] Compiles
- [x] Ran all tests

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -91,6 +91,7 @@ When building on OSX, here's some dependencies you'll need:
 - brew install automake
 - brew install pkg-config
 - brew install libpqxx *(If ./configure later complains about libpq missing, try PKG_CONFIG_PATH='/usr/local/lib/pkgconfig')*
+- brew install postgresql *(deal with fatal error: 'libpq-fe.h' file not found stellar)
 - brew install parallel (required for running tests)
 
 ### Windows


MonsieurNicolas:
Somebody with a Mac can comment on this? This seems weird that we'd have to install `postgres` (the service) in order to compile client code; the line above that installs `libpqxx` may not be correct.

overcat:
@MonsieurNicolas 

I am not familiar with C++ compiler tools, but I tried to compile stellar-core on MacOS(10.15), I hope the information is helpful.

I successfully compiled through the following steps.
- Install xcode(This may not be necessary.)
- Install homebrew
- brew install libsodium
- brew install libtool
- brew install automake
- brew install pkg-config(This may not be necessary.)
- brew install libpqxx (P.S. PKG_CONFIG_PATH='/usr/local/lib/pkgconfig' doesn't work.)
- ln -s /usr/local/opt/libpq/include/* /usr/local/include/ (I believe there is a better way, but I am not familiar with C++ compilation tools 😥)
-  ./autogen.sh && ./configure && make -j6


libpq information. (**libpq is keg-only**)
```bash
> brew info libpq
libpq: stable 12.3 (bottled) [keg-only]
Postgres C API library
https://www.postgresql.org/docs/12/libpq.html
/usr/local/Cellar/libpq/12.3 (2,241 files, 23.8MB)
  Poured from bottle on 2020-07-15 at 14:43:00
From: https://github.com/Homebrew/homebrew-core/blob/HEAD/Formula/libpq.rb
==> Dependencies
Required: krb5 ✔, openssl@1.1 ✔
==> Caveats
libpq is keg-only, which means it was not symlinked into /usr/local,
because conflicts with postgres formula.

If you need to have libpq first in your PATH run:
  echo 'export PATH="/usr/local/opt/libpq/bin:$PATH"' >> ~/.zshrc

For compilers to find libpq you may need to set:
  export LDFLAGS="-L/usr/local/opt/libpq/lib"
  export CPPFLAGS="-I/usr/local/opt/libpq/include"

For pkg-config to find libpq you may need to set:
  export PKG_CONFIG_PATH="/usr/local/opt/libpq/lib/pkgconfig"

==> Analytics
install: 59,629 (30 days), 199,241 (90 days), 805,542 (365 days)
install-on-request: 11,515 (30 days), 31,422 (90 days), 91,460 (365 days)
build-error: 0 (30 days)
```

`export PATH="/usr/local/opt/libpq/bin:$PATH" && export LDFLAGS="-L/usr/local/opt/libpq/lib" && export CPPFLAGS="-I/usr/local/opt/libpq/include" && export PKG_CONFIG_PATH="/usr/local/opt/libpq/lib/pkgconfig"` also doesn't work.

MonsieurNicolas:
Thanks @overcat - I think you confirmed that the instructions are not super clear/have something missing and that we don't need to require people to install postgresql in order to build core (where this PR was going)

Issue #2548, opened 2020-05-26
Implement CAP0033


MonsieurNicolas:
https://github.com/stellar/stellar-protocol/blob/master/core/cap-0033.md

PR #2551, opened 2020-05-28
Generalize ledger txn

jonjove:
# Description
First step toward #2548 

# Checklist
- [x] Compiles
- [x] Ran all tests

### DIFF
--- a/src/ledger/GeneralizedLedgerEntry.h
+++ b/src/ledger/GeneralizedLedgerEntry.h
@@ -1,0 +1,104 @@
+// Copyright 2020 Stellar Development Foundation and contributors. Licensed
+// under the Apache License, Version 2.0. See the COPYING file at the root
+// of this distribution or at http://www.apache.org/licenses/LICENSE-2.0
+
+#pragma once
+
+#include "xdr/Stellar-ledger-entries.h"
+#include "xdr/Stellar-ledger.h"
+
+namespace stellar
+{
+
+enum class GeneralizedLedgerEntryType
+{
+    LEDGER_ENTRY
+};
+
+class GeneralizedLedgerKey
+{
+  private:
+    GeneralizedLedgerEntryType mType;
+    union {
+        LedgerKey mLedgerKey;
+    };
+
+    void assign(GeneralizedLedgerKey const& glk);
+    void assign(GeneralizedLedgerKey&& glk);
+    void construct();
+    void destruct();
+
+    void checkDiscriminant(GeneralizedLedgerEntryType expected) const;
+
+  public:
+    GeneralizedLedgerKey();
+    explicit GeneralizedLedgerKey(GeneralizedLedgerEntryType t);
+
+    GeneralizedLedgerKey(LedgerKey const& lk);
+
+    GeneralizedLedgerKey(GeneralizedLedgerKey const& glk);
+    GeneralizedLedgerKey(GeneralizedLedgerKey&& glk);
+
+    GeneralizedLedgerKey& operator=(GeneralizedLedgerKey const& glk);
+    GeneralizedLedgerKey& operator=(GeneralizedLedgerKey&& glk);
+
+    ~GeneralizedLedgerKey();
+
+    void type(GeneralizedLedgerEntryType t);
+    GeneralizedLedgerEntryType type() const;
+
+    LedgerKey& ledgerKey();
+    LedgerKey const& ledgerKey() const;
+
+    std::string toString() const;
+};
+
+class GeneralizedLedgerEntry
+{
+  private:
+    GeneralizedLedgerEntryType mType;
+    union {
+        LedgerEntry mLedgerEntry;
+    };
+
+    void assign(GeneralizedLedgerEntry const& gle);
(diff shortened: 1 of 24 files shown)


src/ledger/GeneralizedLedgerEntry.h
```diff
@@ -0,0 +1,90 @@
...
+
+#pragma once
+
+#include "xdr/Stellar-ledger-entries.h"
+#include "xdr/Stellar-ledger.h"
+
+namespace stellar
+{
+
+enum class GeneralizedLedgerEntryType
```
MonsieurNicolas, line 13: maybe just `Internal` instead of `Generalized`?

src/ledger/GeneralizedLedgerEntry.h
```diff
@@ -0,0 +1,90 @@
...
+    };
+
+    void assign(GeneralizedLedgerKey const& glk);
+    void construct();
+    void destruct();
+
+    void checkDiscriminant(GeneralizedLedgerEntryType expected) const;
+
+  public:
+    GeneralizedLedgerKey(LedgerKey const& lk);
```
MonsieurNicolas, line 33: shouldn't all this stuff be `explicit`?

src/ledger/GeneralizedLedgerEntry.h
```diff
@@ -0,0 +1,90 @@
...
+    GeneralizedLedgerKey(GeneralizedLedgerKey const& glk);
```
MonsieurNicolas, line 34: no "move" constructor?

src/ledger/GeneralizedLedgerEntry.h
```diff
@@ -0,0 +1,90 @@
...
+
+class GeneralizedLedgerKey
+{
+  private:
+    GeneralizedLedgerEntryType mType;
+    union {
+        LedgerKey mLedgerKey;
+    };
+
+    void assign(GeneralizedLedgerKey const& glk);
```
MonsieurNicolas, line 26: Could do this, or use pointers and avoid dealing with unions for now? (we'll move to C++ 17 shortly and can use variants).

I know this might be faster, but custom new/delete operators tends to trip asan etc and come with some many warning labels...

MonsieurNicolas:
Approach looks doable to me

src/ledger/GeneralizedLedgerEntry.h
jonjove: Well, it depends. I considered all approaches. In order to get compatibility between "generalized" and "non-generalized" we need to do one of the following:
1. do not use explicit here 
2. add overloads everywhere that we want compatibility
3. change all the existing operations to explicitly convert to "generalized"

Path 3 is a painful option. Path 2 is a possible route but seems like a lot of work compared to path 1. We can always change this retroactively. Choosing path 1 means we aren't creating extra work by doing path 2 or path 3 in the future.

It's important to note that all the other constructors (eg. for CAP-33 sponsorship keys) will be explicit because there is no existing code that depends on it.

src/ledger/GeneralizedLedgerEntry.h
jonjove: It'll be in the next update. I've been plugging all the holes as I go. I'm also adding a default constructor and a constructor that takes `GeneralizedLedgerEntryType`.

src/ledger/GeneralizedLedgerEntry.h
jonjove: `Internal` seems better. I'm going to hold off on changing this until everything else is settled here, then global-search-and-replace.

graydon:
So after some orientation I think I see what you're trying to do; I'm not especially well-equipped to spot low-level C++-semantics-violating issues though I can and will try, but I will also comment now from the position of high-level thoughts:

First I agree with the (minor) terminology point that `GeneralizedLedgerEntry` seems a bit of a mouthful. I can scratch my head / search a thesaurus for alternatives, but there's a bigger / more fundamental question I am finding myself thinking while I read this: does a new _base_ type need to be introduced at all, to generalize over the existing case + the new case (which I guess we might call `TempLedgerEntry` or something)? Or is there already a base type that can be extended, somewhere where it's a type that's already essentially confined to the `LedgerTxn` subsystem, as we desire?

For example, we already have wrapper types `LedgerTxnEntry` and `LedgerTxnEntry::Impl` to mediate and track access to `LedgerEntry`s, and the latter has an existing subclass `ConstLedgerTxnEntry::Impl` (which I think is there to support `loadWithoutRecord`?); would it be possible and semantically less-messy to make a new type here instead, like say `TempLedgerTxnEntry::Impl` that lives adjacent to it, just for these ephemeral outputs? I am particularly wondering this because of the number of places this PR has something like `assert(e.entry().type() == GeneralizedLedgerEntryType::LEDGER_ENTRY)` -- those asserts feel to me like the GLE type you're introducing here is leaking too far out of the context you want it for, and maybe you want something more like the other internal values in the LedgerTxn system.

MonsieurNicolas:
@jonjove I think the only thing left from this PR is that we never closed on the name "Generalized*"

jonjove:
Obsoleted by #2623

PR #2552, opened 2020-05-28
Fix formatting

marta-lokhova:


src/main/CommandHandler.cpp
```diff
@@ -123,7 +123,7 @@ CommandHandler::safeRouter(CommandHandler::HandlerRoute route,
     }
     catch (std::exception const& e)
     {
-        retStr = fmt::format(R"({"exception": "{}"})", e.what());
+        retStr = fmt::format(R"({{"exception": "{}"}})", e.what());
```
MonsieurNicolas, line 126: same problem in the `catch(...)` right below?

MonsieurNicolas:
r+ b3137050f9d31892c571e0a461ccc43572311eaf

PR #2559, opened 2020-06-01
Track delayed nodes in qset

marta-lokhova:
this change adds tracking of nodes in the qset that are delayed

### DIFF
--- a/docs/software/admin.md
+++ b/docs/software/admin.md
@@ -544,6 +544,7 @@ After observing consensus, a new field `quorum` will be set with information on 
             "disagree" : 0,
             "fail_at" : 3,
             "hash" : "980a24",
+            "lag_ms" : 430,
             "missing" : 0,
             "phase" : "EXTERNALIZE"
          },
@@ -638,6 +639,7 @@ The output will look something like
             "disagree" : 0,
             "fail_at" : 2,
             "hash" : "d5c247",
+            "lag_ms" : 430,
             "ledger" : 24311579,
             "missing" : 1,
             "phase" : "EXTERNALIZE"
@@ -805,7 +807,7 @@ The output looks something like:
 
 ```json
 {
-   "node" : "GCTSFJ36M7ZMTSX7ZKG6VJKPIDBDA26IEWRGV65DVX7YVVLBPE5ZWMIO",
+   "node" : "GCTSF",
    "qset" : {
       "agree" : 6,
       "delayed" : null,
@@ -813,6 +815,15 @@ The output looks something like:
       "fail_at" : 2,
       "fail_with" : [ "sdf_watcher1", "sdf_watcher2" ],
       "hash" : "d5c247",
+      "lag_ms" : {
+         "sdf_watcher1" : 192,
+         "sdf_watcher2" : 215,
+         "sdf_watcher3" : 79,
+         "stronghold1" : 321,
+         "eno" : 266,
+         "tempo.eu.com" : 225,
+         "satoshipay" : 249,
+      },
       "ledger" : 24311847,
       "missing" : [ "stronghold1" ],
       "phase" : "EXTERNALIZE",
@@ -831,7 +842,7 @@ The output looks something like:
    },
    "transitive" : {
       "critical": [
-         [ "GDM7M262ZJJPV4BZ5SLGYYUTJGIGM25ID2XGKI3M6IDN6QLSTWQKTXQM" ]
+         [ "GDM7M" ]
       ],
       "intersection" : true,
       "last_check_ledger" : 24311536,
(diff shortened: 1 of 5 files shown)


src/scp/SCP.h
```diff
@@ -24,6 +26,8 @@ typedef std::shared_ptr<SCPQuorumSet> SCPQuorumSetPtr;
 class SCP
 {
     SCPDriver& mDriver;
+    std::unordered_map<NodeID, std::shared_ptr<medida::Histogram>>
```
marta-lokhova, line 29: I found it convenient to use histograms in order to track nodes voting overtime. However, right now the sliding window defaults to 5 min of data, and I'm wondering if this is too short (we could change medida to allow passing in a longer time window)

marta-lokhova:
example output in `quorum` endpoint:
```
   "lag_ms" : {
      "LOBSTR 1 (Europe)" : 132,
      "LOBSTR 2 (Europe)" : 164,
      "LOBSTR 3 (North America)" : 158,
      "LOBSTR 4 (Asia)" : 389,
      "LOBSTR 5 (Australia)" : 359,
      "SDF 1" : 285,
      "SDF 2" : 271,
      "SDF 3" : 281,
      "SatoshiPay (DE, Frankfurt)" : 163,
      "SatoshiPay (SG, Singapore)" : 307,
      "SatoshiPay (US, Iowa)" : 203,
      "StellarExpert-V1" : 267,
      "StellarExpert-V2" : 238,
      "StellarExpert-V3" : 240
```

src/herder/HerderSCPDriver.cpp
```diff
@@ -976,11 +1018,24 @@ HerderSCPDriver::recordSCPExternalizeEvent(uint64_t slotIndex, NodeID const& id,
...
+        {
+            recordLogTiming(*timing.mSelfExternalize, now,
+                            mSCPMetrics.mExternalizeDelay,
+                            fmt::format("externalize delay ({})",
+                                        mApp.getConfig().toShortString(id)),
+                            std::chrono::nanoseconds::zero(), slotIndex);
+        }
+
+        // Record lag for other nodes
+        auto recordedLag =
```
MonsieurNicolas, line 1034: why not just use `medida::Timer` directly and avoid `shared_ptr`, then you can just do
`auto &vLag = mQSetLag[id];`

src/herder/HerderSCPDriver.h
```diff
@@ -202,6 +206,9 @@ class HerderSCPDriver : public SCPDriver
     // Prepare timeouts per ledger
     medida::Histogram& mPrepareTimeout;
 
+    // Externalize lag tracking for nodes in qset
+    std::unordered_map<NodeID, std::shared_ptr<medida::Timer>> mQsetLag;
```
MonsieurNicolas, line 210: should be spelled `mQSetLag`

src/herder/HerderImpl.cpp
```diff
@@ -1187,9 +1187,14 @@ HerderImpl::getJsonQuorumInfo(NodeID const& id, bool summary, bool fullKeys,
     ret["node"] = mApp.getConfig().toStrKey(id, fullKeys);
     ret["qset"] = getSCP().getJsonQuorumInfo(id, summary, fullKeys, index);
     bool isSelf = id == mApp.getConfig().NODE_SEED.getPublicKey();
-    if (isSelf && mLastQuorumMapIntersectionState.hasAnyResults())
+    if (isSelf)
     {
-        ret["transitive"] = getJsonTransitiveQuorumIntersectionInfo(fullKeys);
+        ret["lag_ms"] = getHerderSCPDriver().getQsetLagInfo(summary, fullKeys);
```
MonsieurNicolas, line 1192: this should be under the `qset` node.

Also, you need to update the `admin.md` file that contains examples (both compact via `info` and normal)

src/herder/HerderSCPDriver.cpp
```diff
@@ -932,6 +932,48 @@ HerderSCPDriver::getPrepareStart(uint64_t slotIndex)
...
+    auto qSet = getSCP().getLocalQuorumSet();
+    LocalNode::forAllNodes(qSet, [&](NodeID const& n) {
+        auto lag = getExternalizeLag(n);
+        if (lag > 0)
+        {
+            if (!summary)
+            {
+                ret[toStrKey(n, fullKeys)] = static_cast<uint64_t>(lag);
+            }
+            totalLag += lag;
```
MonsieurNicolas, line 951: from what I see we only need to update those when `summary` is true

MonsieurNicolas (changes requested):
this looks good to me, only a few minor changes

marta-lokhova:
addressed your feedback @MonsieurNicolas. This has also been squashed/rebased now.

docs/software/admin.md
```diff
@@ -638,6 +639,7 @@ The output will look something like
             "disagree" : 0,
             "fail_at" : 2,
             "hash" : "d5c247",
+            "lag_ms" : 430,
```
MonsieurNicolas, line 642: you forgot to update the output of the `quorum` in the section `Quorum set diagnostics`

MonsieurNicolas (changes requested):
minor miss in the docs and we can merge it

docs/software/admin.md
marta-lokhova: ah yes, grepped for the wrong thing. Fixed!

MonsieurNicolas:
r+ a3fca83663f612d5091120ce71ebddd1c50646ac

MonsieurNicolas:
r+ 7f0f5a59cddf11a5e194070a1683f13c6a9d6eae

PR #2571, opened 2020-06-05
SCP: properly clear state during internal transition

MonsieurNicolas:
This PR fixes an edge case that was not covered in a fix from a couple years ago in 675382d8e9c81244d9408b96a8463739e070a0bc

Basically, as the SCP state machine goes through a few internal transitions, it has to meet all invariants; yet in this case we forgot to reset `c` (that may or may not be set again in a later transition).

### DIFF
--- a/src/scp/test/SCPTests.cpp
+++ b/src/scp/test/SCPTests.cpp
@@ -16,6 +16,12 @@
 #include "xdrpp/printer.h"
 #include <fmt/format.h>
 
+// General convention in this file is that numbers in parenthesis
+// refer to the rule number in the related protocol from the white paper
+// For example (2) in the ballot protocol refers to:
+// If phi = PREPARE and m lets v confirm new higher ballots prepared,
+// then raise h to the highest such ballot and set z = h.x
+
 namespace stellar
 {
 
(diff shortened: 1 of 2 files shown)


src/scp/test/SCPTests.cpp
```diff
@@ -2452,6 +2510,53 @@ TEST_CASE("ballot protocol core3", "[scp][ballotprotocol]")
             }
         }
     }
+    SECTION("prepared A1 with timeout")
```
marta-lokhova, line 2513: maybe i'm missing something, but the name seems confusing: at the end of the test, B2 was prepared

src/scp/test/SCPTests.cpp
```diff
@@ -2452,6 +2510,53 @@ TEST_CASE("ballot protocol core3", "[scp][ballotprotocol]")
...
+        // p=B2, p'=A1 (1)
+        // computed_h = B2 (2)
+        //   does not update h as b < computed_h
+        // v-blocking ahead -> b = computed_h = B2 (9)
+        // h = B2 (2) (now possible)
+        // c = 0 (1)
+        verifyPrepare(scp.mEnvs[2], v0SecretKey, qSetHash0, 0, B2, &A2, 0, 2,
+                      &B2);
+    }
+    SECTION("third party timing")
```
marta-lokhova, line 2537: also not quite sure what this name means

MonsieurNicolas: I don't have a good name for this, I called it third party as it's a watcher node that doesn't have itself in its quorum set. I guess I'm going to rename it to something else

src/scp/test/SCPTests.cpp
MonsieurNicolas: section starts with `prepared A1` and then timeouts

src/scp/test/SCPTests.cpp
```diff
@@ -1090,6 +1090,30 @@ TEST_CASE("ballot protocol core5", "[scp][ballotprotocol]")
...
                                           0, 3, B3, 2, 2);
                         }
+                        SECTION("higher counter mixed")
+                        {
+                            recvVBlocking(
+                                makePrepareGen(qSetHash, A3, &B3, 0, 2, &A2));
+                            REQUIRE(scp.mEnvs.size() == 6);
+                            // h still A2
+                            // v-blocking
+                            //     prepared B3 -> p = B3, p'=A2 (1)
```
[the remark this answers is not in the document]
MonsieurNicolas: added a comment at the top of the file to explain this convention (there were already test cases written like this)

MonsieurNicolas:
@marta-lokhova addressed your feedback

marta-lokhova:
r+ f67bbf50d8571f7b3b9fd55abb019443c246d2ab

MonsieurNicolas:
@latobarita: retry

PR #2574, opened 2020-06-10
bug 2573: support --metadata-output-stream command-line override for config.

graydon:
This adds a command-line flag `--metadata-output-stream` to `run` and `catchup` which allows users to set the `METADATA_OUTPUT_STREAM` config variable if it's not set in the config file. This helps enable horizon to run captive core while still using a user-provided config file (with user-provided quorum and so on).

### DIFF
--- a/src/main/CommandLine.cpp
+++ b/src/main/CommandLine.cpp
@@ -233,6 +233,28 @@ configurationParser(CommandLine::ConfigOption& configOption)
                "default 'stellar-core.cfg')");
 }
 
+clara::Opt
+metadataOutputStreamParser(std::string& stream)
+{
+    return clara::Opt(stream, "STREAM")["--metadata-output-stream"](
+        "Filename or file-descriptor number 'fd:N' to stream metadata to");
+}
+
+void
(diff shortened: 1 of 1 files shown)


src/main/CommandLine.cpp
```diff
@@ -554,6 +563,13 @@ runCatchup(CommandLineArgs const& args)
                 config.DISABLE_XDR_FSYNC = true;
             }
 
+            if (!stream.empty())
+            {
+                // Allow overriding config METADATA_OUTPUT_STREAM using
+                // command line --metadata-output-stream
+                config.METADATA_OUTPUT_STREAM = stream;
```
MonsieurNicolas, line 570: Should it just blow up if the setting was already in the config file? I kinda hate the notion of "override" that just creates ambiguity.

graydon: Fair, pushed variant that's exclusive.

MonsieurNicolas:
r+ 5990358d9b66e1d42bb3826b30732061dedf193c

PR #2575, opened 2020-06-10
Clean up buckets in in-memory mode in BucketManager::shutdown, not runCatchup

graydon:
In 0c423487b7a1441dc8a0ed612cb5cb2aad8ad2c1 the `runCatchup` code was modified to "clean up" the buckets folder at the end of an in-memory-replay catchup by constructing a second application in newdb=true mode, and then immediately shutting it down. This does technically work but it means that the app is started-and-stopped _twice_ In a catchup run, and if there's anything in the catchup run that's only supposed to happen once this will try to happen twice, and potentially fail the second time.

Specifically: if we run with a metadata output stream set to a file descriptor, it'll be closed on the first application shutdown and then you'll get a spurious error when the second application attempts to re-open it.

In general I think this "run the app twice" approach probably isn't ideal. At very least not what a casual reader of ApplicationImpl is going to expect. So this PR suggests removing it and achieving the same cleanup by teaching the BucketManagerImpl to purge the bucket directory on shutdown in in-memory-replay mode instead.

### DIFF
--- a/src/bucket/BucketManagerImpl.h
+++ b/src/bucket/BucketManagerImpl.h
@@ -48,6 +48,8 @@ class BucketManagerImpl : public BucketManager
     medida::Counter& mSharedBucketsSize;
     MergeCounters mMergeCounters;
 
+    bool const mDeleteEntireBucketDirInDtor;
+
     // Records bucket-merges that are currently _live_ in some FutureBucket, in
     // the sense of either running, or finished (with or without the
     // FutureBucket being resolved). Entries in this map will be cleared when
@@ -64,7 +66,8 @@ class BucketManagerImpl : public BucketManager
     BucketMergeMap mFinishedMerges;
 
     void cleanupStaleFiles();
-    void cleanDir();
+    void deleteTmpDirAndUnlockBucketDir();
+    void deleteEntireBucketDir();
     bool renameBucket(std::string const& src, std::string const& dst);
 
 #ifdef BUILD_TESTS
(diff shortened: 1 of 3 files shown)


graydon:
@marta-lokhova I've pushed a new version here that leaves more of the existing structure in place -- BucketManagerImpl::shutdown is called only in ApplicationImpl::gracefulStop so that other subsystems will still be alive, and only forgets unreferenced buckets -- and extracted the Bucket-dir-deletion code into its own method which is called both from the dropAll path (run on new-db) and the ~BucketManagerImpl path, parameterized by a boolean flag initialized at BucketManagerImpl construction time from the config.MODE_USES_IN_MEMORY_LEDGER value.

I also renamed and added comments to some stuff, and reordered the tmpdir-deletion path to unlock the bucket dir only _after_ deleting the tmpdir (not that this likely ever resulted in a real error, but it seemed like a potential race).

src/bucket/BucketManagerImpl.h
```diff
@@ -48,6 +48,8 @@ class BucketManagerImpl : public BucketManager
     medida::Counter& mSharedBucketsSize;
     MergeCounters mMergeCounters;
 
+    bool mDeleteEntireBucketDirInDtor;
```
marta-lokhova, line 51: safer to use `const`

graydon: Good catch, fixed (and rebased!)

graydon:
r+ 52168c3

(self-r+'ing given approval noted above)

PR #2576, opened 2020-06-11
Convert OverlaySurvey.py to Python 3

hidenori-shinohara:
# Description

Convert `OverlaySurvey.py` to Python 3. Unfortunately, this script is rather hard to test because the output is different when this script is run multiple times.

However, I tried running this with Python 2 and Python 3 at the same time on testnet, and the results seem fairly close by looking at the output files using yEd.

On top of that, the only change in the code is the parentheses, so it seems reasonable to assume that this change will not affect the performance.

# Checklist
- [x] Compiles
- [x] Ran all tests

### DIFF
--- a/scripts/OverlaySurvey.py
+++ b/scripts/OverlaySurvey.py
@@ -180,7 +180,7 @@ def run_survey(args):
                 peer_list.append(key)
 
     if nx.is_empty(G):
-        print "Graph is empty!"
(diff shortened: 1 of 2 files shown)


MonsieurNicolas:
r+ 417f4a46059d2e6f9e3ff0f020b33eb285ac880f

PR #2580, opened 2020-06-12
bump min overlay version to 13

MonsieurNicolas:
This bumps min overlay version to 13 as only v13.X clients are supported at this time


graydon:
r+ d064390

Issue #2518, opened 2020-05-04
Metric `overlay.delay.write-queue` has negative values (Version: 12.5.0-32)


jacekn:
## Issue Description

I have stellar-core 12.5.0-32 that exposes negative values of the `overlay.delay.write-queue` metric (for example sum or mean):
```
,"overlay.delay.write-queue":{
"type":"timer",
"count":12216602352,
"event_type":"calls",
"rate_unit":"s",
"mean_rate":4504.95,
"1_min_rate":6277.53,
"5_min_rate":6355.12,
"15_min_rate":6257.9,
"duration_unit":"ms",
"min":0,
"max":149469,
"mean":-633.807,
"stddev":7763.77,
"sum":-7.74296e+12,
"median":0,
"75%":0.001,
"95%":5.3996,
"98%":2962.81,
"99%":14027.3,
"99.9%":66916.1
}
```

## Steps to Reproduce

On core 12.5.0-32: `curl 127.0.0.1:11626/metrics` and look at metrics

## Expected Result

Timers have positive values

## Actual Result

`overlay.delay.write-queue` has negative sum and mean. In my case this is the only affected metric.

## Your Environment and Setup

* stellar-core 12.5.0-32
* Ubuntu 16.04.6 LTS
* installed using apt.stellar.org deb package

## Additional context

I noticed this because the [stellar-core-prometheus-exporter](https://github.com/stellar/stellar-core-prometheus-exporter) stops returning data when it sees negative values for sum.

MonsieurNicolas:
Weird, looks `TimestampedMessage::recordWriteTiming` is called when it shouldn't or we overflow duration somehow on the target system?

MonsieurNicolas:
@graydon of note, I think that the change to steady clock may not be enough for this particular one: we suspect overflow in this case as we accumulate timing for all writes (look at the value of `count: 12216602352`).

Basically, writes performed take in the order of a second, which in nano is 1e9, and as we have many writes done, we end up with `12216602352*1e9 ~= 1.22e19 > 2^63`.

The fix might be to either
* change unit for that metric (or class of metrics). not sure how that works: medida seems to be hardcoded to use nanoseconds.
* change medida to automatically reset metrics that overflow. Downstream systems like prometheus actually detect when this happens and compensate.

graydon:
[...]

PR #2583, opened 2020-06-16
Bump medida for #2518 (metrics overflow) and add test.

graydon:
Update medida sumodule to take https://github.com/stellar/medida/pull/20 and add a test that fails before but passes with the update.


MonsieurNicolas:
r+ 452444f79adbe7b94c0b4410a8d29277c964a367

PR #2584, opened 2020-06-17
Issue 2570 pr 1

### LINKED ISSUE #2570: Simplify Schema for LedgerEntry Extensions
Some ledger entries, specifically `AccountEntry` and `TrustLineEntry`, have non-trivial extensions. With https://github.com/stellar/stellar-protocol/blob/master/core/cap-0033.md, every `LedgerEntry` will have an extension and `AccountEntry` will have two extensions.

Right now, our database schema deconstructs the extension fields. For example, https://github.com/stellar/stellar-core/blob/c097a3fe36f108bcbb52e4995b81bfe617f3109a/src/ledger/LedgerTxnAccountSQL.cpp#L470-L486 separates the liabilities out of the extension. This is unnecessary because we never query by those columns. As a consequence, our database interactions must have special handling for the extension fields which are not guaranteed to exist.

Instead, we should store the extension fields as base-64 encoded XDR-opaque. There are several advantages to this approach:
1. There won’t be any need for special handling of whether the extension fields exist since this will be determined via the typical XDR encoding/decoding logic.
2. We won’t need to update the database code for new extensions, it will “just work” automatically in the future.

The `LedgerEntry`-level extension should be one column, and the `AccountEntry`/`TrustLineEntry`-level should be another column. There is no need to include the `OfferEntry`/`DataEntry` extensions since they are always trivial now; we can update the schema if that changes.

This change is parallel to past schema changes such as #1957 (for assets) and #1866 (for signers).

rokopt:
# Update DB schema to make all ledger entry extensions opaque

Resolves #2570 

This pull request changes the database schema as proposed in issue #2570 by converting the extension fields of AccountEntry, TrustLineEntry, and LedgerEntry into opaque XDR.  It upgrades old databases by copying the existing transparent extensions (which are Liabilities fields in the case of AccountEntry and TrustLineEntry, and empty in the case of LedgerEntry) into new columns, and then deleting the column containing the transparent extensions.  (Since the LedgerEntry extension is empty, it had no old column; it simply gains a new one.)

The PR introduces a test which creates a mechanism for injecting code into points during application creation, in this case, between the creation of the database (which we do with an old schema) or the loading of an old database and the upgrading of the database to the new schema.  It injects code to use raw SQL to create AccountEntry and TrustLineEntry entries with Liabilities fields while the database is in the old format.  Then, after the application initialization has completed, and the database has been upgraded, it uses the standard interfaces to load the accounts and trustlines and validate that they have been upgraded as expected (accounts or trustlines with NULL Liabilities fields will now have NULL extension fields, while those with non-NULL Liabilities will have non-NULL opaque extensions which decode to Liabilities fields equal to the old ones).

I'm starting out by filing this PR as a draft because I'd still like to do some performance investigations on large, real-world databases; the code might be complete if the performance turns out to be fine, though, so the draft will allow a head-start on code-review.

# Checklist
- [x] Compiles
- [x] Ran all tests


src/main/ApplicationImpl.cpp
```diff
@@ -902,4 +902,9 @@ ApplicationImpl::getLedgerTxnRoot()
     return mConfig.MODE_USES_IN_MEMORY_LEDGER ? *mNeverCommittingLedgerTxn
                                               : *mLedgerTxnRoot;
 }
+
+void
+ApplicationImpl::actBeforeDBSchemaUpgrade()
```
jonjove, line 907: I'm not sure this is the cleanest approach. It has the annoying overhead of needing to create a derived class for each variation you want. Perhaps it would be better to add a test-only setting (eg. one that can't be specified in the configuration file) to `Config` called something like `UPGRADE_TO_SCHEMA_VERSION` (default value is the current schema version). Then instead of `ApplicationImpl` calling `mDatabase->upgradeToCurrentSchema` you can call `mDatabase->upgradeToSchema(mConfig.UPGRADE_TO_SCHEMA_VERSION)`. With this, you can just create an application at the desired schema version then upgrade it again later. Do you think this is better or worse?

src/database/test/DatabaseTests.cpp
```diff
@@ -353,14 +353,25 @@ TEST_CASE("schema test", "[db]")
 class SchemaUpgradeTestApplication : public TestApplication
 {
   public:
-    SchemaUpgradeTestApplication(VirtualClock& clock, Config const& cfg)
+    using PreUpgradeFunc = std::function<void(SchemaUpgradeTestApplication&)>;
+
+  private:
+    PreUpgradeFunc const preUpgradeFunc;
```
jonjove, line 364: Nitpick: code-style conventions, should be `mPreUpgradeFunc`

src/database/test/DatabaseTests.cpp
```diff
@@ -353,14 +353,25 @@ TEST_CASE("schema test", "[db]")
...
+
+  public:
+    SchemaUpgradeTestApplication(VirtualClock& clock, Config const& cfg,
+                                 PreUpgradeFunc _preUpgradeFunc)
```
jonjove, line 368: Nitpick: this isn't mentioned in the code-style conventions, but we don't typical use the "prefix-underscore" because it isn't necessary with the "prefix-m" member variable convention

src/database/test/DatabaseTests.cpp
```diff
@@ -377,13 +377,56 @@ class SchemaUpgradeTestApplication : public TestApplication
...
+            session
+                << "INSERT INTO accounts ( "
+                   "accountid, balance, seqnum, numsubentries, inflationdest,"
+                   "homedomain, thresholds, signers, flags, lastmodified, "
+                   "buyingliabilities, sellingliabilities "
+                   ") VALUES ( "
+                   ":id, :v1, :v2, :v3, :v4, :v5, :v6, :v7, :v8, :v9, :v10, "
+                   ":v11 "
+                   ")",
+                soci::use(mAccountID, "id"), soci::use(mBalance, "v1"),
```
jonjove: You should use positional substitution instead of named-value substitution, it's considerably faster. Obviously this is a test so performance is much less important, but that is the reason we use positional substitution everywhere else.

src/database/test/DatabaseTests.cpp
```diff
@@ -377,13 +377,56 @@ class SchemaUpgradeTestApplication : public TestApplication
 
 TEST_CASE("schema upgrade test", "[db]")
 {
-    auto test_one_db_mode = [](Config::TestDbMode const db_mode) {
+    auto prepOldSchemaDB = [](SchemaUpgradeTestApplication& app) {
+        auto& session = app.getDatabase().getSession();
+        {
+            std::string const mAccountID = "account";
```
jonjove: Nitpick: code-style conventions again, local variables are camel-case with first letter lower.

src/database/test/DatabaseTests.cpp
```diff
@@ -428,6 +428,35 @@ TEST_CASE("schema upgrade test", "[db]")
                                mSellingLiabilities0);
         addOneOldSchemaAccount(app, mAccountID1, mBuyingLiabilities1,
                                mSellingLiabilities1);
+        auto addOneOldSchemaTrustLine =
+            [](SchemaUpgradeTestApplication& app, std::string const mAccountID,
```
jonjove: This is getting a little hard on the eyes. Can you avoid the nested lambdas? I'm not even sure if these lambdas are adding a lot of value compared to just looping.

src/database/test/DatabaseTests.cpp
```diff
@@ -378,26 +382,26 @@ class SchemaUpgradeTestApplication : public TestApplication
...
-            std::string const mThreshold = "threshold";
-            std::string const mSigners = "autograph";
-            int32_t const mFlags = 0;
-            int32_t const mLastModified = 3;
+            std::string accountIDStr =
+                KeyUtils::toStrKey<PublicKey>(ae.accountID);
+            std::string inflationDestStr =
+                KeyUtils::toStrKey<PublicKey>(*ae.inflationDest);
+            std::string homeDomainStr;
+            homeDomainStr = decoder::encode_b64(ae.homeDomain);
```
jonjove, line 422: Nitpick: can just combine this line into the previous line. Also, these can use `auto`.

src/database/Database.cpp
```diff
@@ -289,6 +302,202 @@ Database::upgradeToCurrentSchema()
...
 
+void
+Database::addTextColumnIfNotPresent(std::string const table,
+                                    std::string const column)
+{
+    std::string addColumnStr("ALTER TABLE " + table + " ADD " + column +
+                             " TEXT;");
+    CLOG(INFO, "Database") << "Adding column with string '" << addColumnStr
+                           << "'";
+    // SQLite doesn't give us a way of doing an "ALTER TABLE" only if some
```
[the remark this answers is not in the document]
rokopt: Yes, good catch (no pun intended -- at least, not at first...) -- I agree with that general statement about exception handling; I wrote that code early on, even though I didn't like it, when I misunderstood how our upgrade code worked and I thought that there was no way to distinguish in that code path whether we'd already performed an upgrade or not (so I thought the code it had to be idempotent).  Now I see that that should be called exactly once per database, so we can expect it to be called only when the column doesn't exist yet.  I've removed that try/catch.

src/database/Database.cpp
```diff
@@ -289,6 +302,202 @@ Database::upgradeToCurrentSchema()
     assert(vers == SCHEMA_VERSION);
 }
 
+void
+Database::addTextColumnIfNotPresent(std::string const table,
+                                    std::string const column)
+{
+    std::string addColumnStr("ALTER TABLE " + table + " ADD " + column +
```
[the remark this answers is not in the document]
rokopt: Good question -- as far as I can tell, we're only currently using collation for columns used as keys, as you mention, but I could experiment with this on owlbear (once I have figured out how to do that) if you think it's potentially worthwhile.

src/database/Database.cpp
```diff
@@ -289,6 +302,202 @@ Database::upgradeToCurrentSchema()
...
+            << "Adding column '" << column << "' to table '" << table
+            << "' failed: '" << e.what()
+            << "' (possibly the table already has the column)";
+    }
+}
+
+void
+Database::dropTextColumn(std::string const table, std::string const column)
+{
+    // SQLite doesn't give us a way of dropping a column with a single
```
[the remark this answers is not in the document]
rokopt: I like the idea of NULLing out all the cells of unused columns if we can't drop them altogether (i.e. when using SQLite), so for now I've implemented that.  As for replacing the updates with a rewrite, I'll look into that along with other performance investigations.

src/database/Database.cpp
```diff
@@ -289,6 +302,202 @@ Database::upgradeToCurrentSchema()
...
+    auto prep_select = getPreparedOldLiabilitySelect("accounts", "accountid");
+    auto& st_select = prep_select.statement();
+    st_select.exchange(soci::into(accountIDStrKey));
+    st_select.exchange(
+        soci::into(extension.liabilities.buying, buyingLiabilitiesInd));
+    st_select.exchange(
+        soci::into(extension.liabilities.selling, sellingLiabilitiesInd));
+    st_select.define_and_bind();
+    {
+        auto timer = getSelectTimer("account-ext-to-opaque");
```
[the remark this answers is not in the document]
rokopt: Okay, I've removed it.

src/database/Database.cpp
```diff
@@ -289,6 +302,202 @@ Database::upgradeToCurrentSchema()
...
+    for (; st_select.got_data(); st_select.fetch())
+    {
+        // We've only selected accounts which have at least one of
+        // buying liabilities or selling liabilities present, and if
+        // either is present, then both should be.
+        assert(buyingLiabilitiesInd == soci::i_ok);
+        assert(sellingLiabilitiesInd == soci::i_ok);
+        std::string opaqueExtension(
+            decoder::encode_b64(xdr::xdr_to_opaque(extension)));
+        auto prep_update = getPreparedStatement(
```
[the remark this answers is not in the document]
rokopt: Ah, I didn't realize it was possible to reuse a prepared statement.  I've now done that.

However, having done so, I'm now not sure about it.  If I understand what that's doing correctly, it's what this page refers to as "Statement caching":

http://soci.sourceforge.net/doc/master/statements/

And that page says that to reuse the prepared statement, we should follow the ```execute()``` call with a call to ```bind_clean_up()```.  But I can't find a ```bind_clean_up()``` in the version of soci that we're using.

As for the interleaving, I wondered about that myself, thought I'd convinced myself that it was safe, and after your question am now unsure again.  Perhaps it's easier to confirm that we'll be able to hold all the changes in memory for the real database than it is to confirm that this code is always safe (now that my Googling has also failed to produce a definitive answer).  I'll change that and experiment with it.

src/database/Database.cpp
```diff
@@ -289,6 +302,202 @@ Database::upgradeToCurrentSchema()
...
+    st_select.exchange(soci::into(accountIDStrKey));
+    st_select.exchange(soci::into(issuerStrKey));
+    st_select.exchange(soci::into(assetStrKey));
+    st_select.exchange(
+        soci::into(extension.liabilities.buying, buyingLiabilitiesInd));
+    st_select.exchange(
+        soci::into(extension.liabilities.selling, sellingLiabilitiesInd));
+    st_select.define_and_bind();
+    {
+        auto timer = getSelectTimer("trustline-ext-to-opaque");
```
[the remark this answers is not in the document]
rokopt: Removed that one too.

src/database/Database.cpp
```diff
@@ -289,6 +302,202 @@ Database::upgradeToCurrentSchema()
...
+    for (; st_select.got_data(); st_select.fetch())
+    {
+        // We've only selected trustlines which have at least one of
+        // buying liabilities or selling liabilities present, and if
+        // either is present, then both should be.
+        assert(buyingLiabilitiesInd == soci::i_ok);
+        assert(sellingLiabilitiesInd == soci::i_ok);
+        std::string opaqueExtension(
+            decoder::encode_b64(xdr::xdr_to_opaque(extension)));
+        auto prep_update = getPreparedStatement(
```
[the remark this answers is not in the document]
rokopt: Changed that one too, to prepare the statement only once, though both will probably change again to avoid the interleaved processing.

src/database/Database.h
```diff
@@ -103,6 +103,22 @@ class Database : NonMovableOrCopyable
     static void registerDrivers();
     void applySchemaUpgrade(unsigned long vers);
 
+    // Convert the accounts table from using explicit entries for
+    // extension fields into storing the entire extension as opaque XDR.
+    void convertAccountExtensionsToOpaqueXDR();
+    void copyIndividualAccountExtensionFieldsToOpaqueXDR();
+
+    StatementContext getPreparedOldLiabilitySelect(std::string const table,
```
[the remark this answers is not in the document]
rokopt: Changed (in ```getPreparedOldLiabilitySelect()```, ```addTextColumnIfNotPresent()```, and ```dropTextColumn()```).

src/database/test/DatabaseTests.cpp
rokopt: Fixed.

src/database/test/DatabaseTests.cpp
rokopt: Fixed.

src/database/test/DatabaseTests.cpp
rokopt: Fixed.  Made all the declarations in that block into "auto"s, while I was at it.

src/ledger/LedgerTxnAccountSQL.cpp
```diff
@@ -21,80 +21,87 @@ namespace stellar
...
-    st.exchange(soci::into(liabilities.selling, sellingLiabilitiesInd));
-    st.exchange(soci::into(signers, signersInd));
-    st.exchange(soci::use(actIDStrKey));
-    st.define_and_bind();
-    {
-        auto timer = mDatabase.getSelectTimer("account");
-        st.execute(true);
-    }
-    if (!st.got_data())
+    try
```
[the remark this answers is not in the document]
rokopt: The only reason for those catches and re-throws was to get a function name into the logs so as to narrow down the place where a problem occurred, should one happen in the future -- I _did_ have some during tests, when I had problems such as SQL syntax errors, so I thought that might be useful.  I'm not attached to them; I could remove them altogether if you prefer, or do them in a separate commit as you suggest, or leave them as they are if github's ```"?w=1"``` suffices to clarify the changes, as Graydon discovered in about five seconds (and benevolently refrained from directing me to "letmegooglethatforyou").

src/ledger/LedgerTxnAccountSQL.cpp
```diff
@@ -282,82 +282,84 @@ class BulkUpsertAccountsOperation : public DatabaseTypeSpecificOperation<void>
...
-        st.exchange(soci::use(strInflationDests));
-        st.exchange(soci::use(strHomeDomains));
-        st.exchange(soci::use(strThresholds));
-        st.exchange(soci::use(strSigners));
-        st.exchange(soci::use(strFlags));
-        st.exchange(soci::use(strLastModifieds));
-        st.exchange(soci::use(strBuyingLiabilities));
-        st.exchange(soci::use(strSellingLiabilities));
-        st.define_and_bind();
+        try
```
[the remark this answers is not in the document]
rokopt: Same reply and options for you as above :)

src/ledger/LedgerTxnAccountSQL.cpp
```diff
@@ -412,24 +414,33 @@ class BulkDeleteAccountsOperation : public DatabaseTypeSpecificOperation<void>
...
-        std::string strAccountIDs;
-        marshalToPGArray(conn, strAccountIDs, mAccountIDs);
-        std::string sql =
-            "WITH r AS (SELECT unnest(:ids::TEXT[])) "
-            "DELETE FROM accounts WHERE accountid IN (SELECT * FROM r)";
-        auto prep = mDB.getPreparedStatement(sql);
-        soci::statement& st = prep.statement();
-        st.exchange(soci::use(strAccountIDs));
-        st.define_and_bind();
+        try
```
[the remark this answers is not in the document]
rokopt: Same reply and options for you as above :)

src/main/ApplicationImpl.cpp
rokopt: I considered a Config option at first, and I might be misunderstanding something but I don't think that it would have worked:  the application creation itself, when it creates a new database, ends up calling LedgerManagerImpl::startNewLedger(), which creates a root account and then closes the ledger, which in turn updates the database, and the database update code expects the database to be in the current schema already.  So we can't wait until after createTestApplication() returns before upgrading the schema version.

Furthermore, even if a Config option would have worked, I'd still like what I did better, for several reasons:

- Because the derived class constructor can take whatever parameters it wants (because createTestApplication() and Application::create() take variadic parameter packs), and in particular can take lambdas for one or more parameters, we don't need to create a new derived class for each different function that we might want to inject into the code path between database creation and schema upgrade; we can just pass a different lambda for each different function, as illustrated in the code I've already written using the SchemaUpgradeTestApplication derived class, and each different function could take different sets of (and numbers of) parameters.  We only need to create a new derived class if and when we reuse the mechanism to add test-code-injection functionality to _other_ code paths, and that would also be true of a Config option -- if we wanted to add test-code-injection functionality to more code paths in the future, we'd have to add new Config options for those.

- The code that handles the weird C++ variadic-parameter syntax only has to be written once (and is written in this check-in) -- clients (as illustrated by SchemaUpgradeTestApplication) don't have to use that syntax; they simply write their constructors with however many parameters they want, and pass a matching list of parameters to createTestApplication().

- With this code, since the variadic parameters propagate to Application::create(), any other future derived classes of Application itself, or future users of Application::create with different template parameters, would also get the option of having arbitrary-length-and-type parameter lists passed to their constructors as part of Application::create().

- With this code, each call to createTestApplication() can store a different set of parameters in its template parameter class (derived from TestApplication or Application), which would then be accessible to the callback from the injection point, so that that same callback would work, with the same list of parameter types (including an empty list, as in this case), for different invocations of createTestApplication() using the same derived class.  To accomplish that with a Config option without deriving from TestApplication, I think that we'd have to do something such as including potential parameters to the injection function within TestApplication itself, and setting them manually in the client before calling createTestApplication(), which would both litter the base TestApplication with members having to do with injections that were only used by certain individual tests, and require clients to make multiple ->setParameter() calls to set those parameters prior to calling createTestApplication(), rather than simply using them as parameters to createTestApplication() itself.

In short, I don't think a Config option works, because of the code called within Application::create() itself that relies on the schema already having been upgraded, but even if it did work, I think that what I did would provide a lot more flexibility for clean future test-code injections, so I'd still prefer it.  But of course I'm open to counter-arguments!

src/database/test/DatabaseTests.cpp
```diff
@@ -349,3 +349,29 @@ TEST_CASE("schema test", "[db]")
     auto av = db.getAppSchemaVersion();
     REQUIRE(dbv == av);
 }
+
+TEST_CASE("schema upgrade test", "[db]")
+{
+    auto test_one_db_mode = [](Config::TestDbMode const db_mode) {
```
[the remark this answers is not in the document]
rokopt: Yeah, I caught myself with ```test_one_db_mode```.  Now I see that I should change ```db_mode``` to ```dbMode```, too.

I think I'll also do a little refactoring there that might either get rid of that lambda or make it more obviously useful.  :)

src/database/test/DatabaseTests.cpp
rokopt: Yeah, I caught myself there in a later commit too (yet another bit of review I could have saved you by squashing commits).

src/database/test/DatabaseTests.cpp
rokopt: Ah, I somehow failed to notice that I was doing that and the existing code wasn't.  Fixed.

src/database/test/DatabaseTests.cpp
rokopt: Caught myself there too.  :)

src/database/test/DatabaseTests.cpp
```diff
@@ -377,13 +377,56 @@ class SchemaUpgradeTestApplication : public TestApplication
 
 TEST_CASE("schema upgrade test", "[db]")
 {
-    auto test_one_db_mode = [](Config::TestDbMode const db_mode) {
+    auto prepOldSchemaDB = [](SchemaUpgradeTestApplication& app) {
+        auto& session = app.getDatabase().getSession();
+        {
```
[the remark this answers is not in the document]
rokopt: Agreed, I moved it into addOneOldSchema* in a later version.

src/database/test/DatabaseTests.cpp
rokopt: Yes, you're right, those lambdas can be un-nested, and doing so makes them more readable.  Changed.

I am going to leave them as lambdas, because I want to factor out that test code so that it doesn't take three hard-coded parameters, but a vector instead.  That ought to make it shorter and more readable still.

rokopt:
Jon, I think I've gone over all your code-review comments (thank you!), and addressed most.  Here's what I believe remains, which I'll work on next:

- Separating the select from the update in the upgrade code, rather than interleaving them, since we're not sure that the interleaving is guaranteed to be correct.

- Doing some refactoring in the upgrade test to use the lambdas to avoid fixed-length test-case lists.

- Doing some refactoring out of commonality in the upgrade code (once it's been changed to separate select from upgrade).

- Testing various aspects of performance, and possibly making changes based on them:
    - Experimenting with collation in PostgreSQL.
    - Measuring upgrade time on the real-world database.
    - Comparing pre-upgrade and post-upgrade performance on the real-world database.
    - Measuring how long a full rewrite would take instead, and whether that would help with post-upgrade performance.

- Making a new branch in which to squash commits into a small number of independently-reviewable and non-overlapping changes.

Does that sound right?

src/database/test/DatabaseTests.cpp
rokopt: Okay, having already un-nested the lambdas, I've now made them useful by fully parameterizing the test on the Liabilities vectors to be used for accounts and trustlines.  Modify those vectors and the test will now automatically change to create the new correct numbers of accounts and trustlines and validate the new expected results.  I think the code is more readable as a result too -- we just pass vectors around instead of long fixed-size parameter lists of accounts, trustlines, and liabilities.

src/database/Database.cpp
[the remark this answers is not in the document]
rokopt: I have now changed the upgrade code to separate the select from the update rather than interleaving them.  While I was at it, I also factored out the commonality between the account and trustline updates, so that there's one place (a templated function called ```copyIndividualExtensionFieldsToOpaqueXDR()```) where we can experiment with the algorithm.

src/database/test/DatabaseTests.cpp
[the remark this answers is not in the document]
rokopt: (That refactoring that I wanted to do is now done.)

rokopt:
> Jon, I think I've gone over all your code-review comments (thank you!), and addressed most. Here's what I believe remains, which I'll work on next:
> 
> * Separating the select from the update in the upgrade code, rather than interleaving them, since we're not sure that the interleaving is guaranteed to be correct.
> * Doing some refactoring in the upgrade test to use the lambdas to avoid fixed-length test-case lists.
> * Doing some refactoring out of commonality in the upgrade code (once it's been changed to separate select from upgrade).
> * Testing various aspects of performance, and possibly making changes based on them:
>   
>   * Experimenting with collation in PostgreSQL.
>   * Measuring upgrade time on the real-world database.
>   * Comparing pre-upgrade and post-upgrade performance on the real-world database.
>   * Measuring how long a full rewrite would take instead, and whether that would help with post-upgrade performance.
> * Making a new branch in which to squash commits into a small number of independently-reviewable and non-overlapping changes.
> 
> Does that sound right?

Following up on this:  the select/update separation and the test and upgrade refactoring that I wanted to do are now done.  I think that what remains are the performance investigations (and any changes that might come out of them) and then the commit-squashing (in a separate branch) for readability.

rokopt:
This PR has been made obsolete by https://github.com/stellar/stellar-core/pull/2593, which contains extensive revisions and a greatly simplified history.

PR #2587, opened 2020-06-18
Path payment test fixes

rokopt:
# Description

Split "offers in a loop" path payment tests into separate TEST_CASES
so that they can use Application objects without USE_CONFIG_FOR_GENESIS.

The path payment and strict path payment "use all offers in a loop"
tests perform modifications (an account merge and a ledger close)
in a for_versions() loop when and only when the protocol version is
13 or greater.  In master currently, it is 13, so the modifications are
performed only in version 13, which, because it is the current version,
is also the maximum protocol version of any bucket.

When the version is bumped to 14, however, there will be version-14
buckets from previous tests when the "use all offers in a loop" tests
try to perform modifications while running protocol 13.  This will
cause the bucket merge code to crash, as it will find existing buckets
of version 14 but it will be running with a maximum supported
protocol of 13 (during that iteration of for_versions()).

Other tests have faced this problem, and there is a test-only Config
option which works around it: disabling USE_CONFIG_FOR_GENESIS,
which those other tests do.

However, many of the path payment and strict path payment tests
depend upon USE_CONFIG_FOR_GENESIS, so we can't just turn it
off throughout all the path payment tests without changing many
other tests.  Therefore, this pull request fixes the problem by splitting
the "use all offers in a loop" tests, which are currently SECTION()s,
into their own TEST_CASE()s, with their own Application objects,
which use Config objects with USE_CONFIG_FOR_GENESIS disabled.

# Checklist
- [x] Compiles
- [x] Ran all tests

(Regarding performance, only very short-running tests are changed.)

### DIFF
--- a/src/transactions/test/PathPaymentStrictSendTests.cpp
+++ b/src/transactions/test/PathPaymentStrictSendTests.cpp
@@ -1916,108 +1916,6 @@ TEST_CASE("pathpayment strict send", "[tx][pathpayment]")
         });
     }
 
-    SECTION("uses all offers in a loop")
-    {
-        auto useAllOffersInLoop = [&](TestAccount* issuerToDelete) {
-            auto market = TestMarket{*app};
-            auto source = root.create("source", minBalance4);
-            auto destination = root.create("destination", minBalance1);
-            auto mm12 = root.create("mm12", minBalance3);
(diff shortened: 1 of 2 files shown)


jonjove:
r+ 2e3886e

PR #2588, opened 2020-06-18
Use ms for consensus deltas logging

### LINKED ISSUE #2538: Display consensus round latencies in appropriate units
Right now our logs can be a bit hard to read, because we use nanoseconds. 
```
externalize lag delta for slot 29758332 is 189298938 ns.
Nominate delta for slot 29758332 is 13858470148 ns.
Prepare delta for slot 29758332 is 50538432 ns.
```
It doesn't seem like we need this much precision, and it would be much easier to parse this if we used milliseconds instead.

marta-lokhova:
Resolves #2538


graydon:
r+ f6efa23

PR #2593, opened 2020-06-23
Issue #2570: DB schema change to make extensions opaque

### LINKED ISSUE #2570: Simplify Schema for LedgerEntry Extensions

rokopt:
# Update DB schema to make all ledger entry extensions opaque

Resolves #2570 

This pull request changes the database schema as proposed in issue #2570 by converting the extension fields of AccountEntry, TrustLineEntry, DataEntry, OfferEntry, and LedgerEntry into opaque XDR.  It upgrades old databases by copying the existing transparent extensions (which are Liabilities fields in the case of AccountEntry and TrustLineEntry, and empty in the other cases) into new columns, and then deleting the column containing the transparent extensions.  (Extensions that were empty had no old columns; they simply gain new ones.)

The PR introduces a test which creates a mechanism for injecting code into points during application creation, in this case, between the creation of the database (which we do with an old schema) or the loading of an old database and the upgrading of the database to the new schema.  It injects code to use raw SQL to create ledger entries of all types while the database is in the old format.  Then, after the application initialization has completed, and the database has been upgraded, it uses the standard interfaces to load the accounts, trustlines, key-value data, and offers, and validate that they have been upgraded as expected (accounts or trustlines which had NULL Liabilities fields will now have NULL extension fields, while those which had non-NULL Liabilities will have non-NULL opaque extensions which decode to Liabilities fields equal to the old ones).

See https://github.com/stellar/stellar-core/pull/2584 for conversations about a previous version of this pull request, which is made obsolete by this one (which has had its history cleaned up, as well as having had extensive revisions since PR 2584).

I'm starting out by filing this PR as a draft because I still have to do some performance investigations on large, real-world databases; possible questions include:
- How long does upgrade take on the real database on a system such as owlbear?
- Does live performance on the real database change measurably after an upgrade on a system such as owlbear?
- Does PostgreSQL collation make any difference to live performance?  (Or does that only help with fields used as keys?)
- How long would a full database rewrite take as opposed to the column upgrade done by this code, and would there be a difference between the two in live performance afterwards?  (And if, for example, the full rewrite would take much longer, but would also produce better performance afterwards, is there a way that people can choose to take the time to do a full rewrite if they wish to?)

# Checklist
- [x] Compiles
- [x] Ran all tests

### DIFF
--- a/docs/db-schema.md
+++ b/docs/db-schema.md
@@ -51,8 +51,8 @@ homedomain | VARCHAR(44) | (BASE64)
 thresholds | TEXT | (BASE64)
 flags | INT NOT NULL |
 lastmodified | INT NOT NULL | lastModifiedLedgerSeq
-buyingliabilities | BIGINT CHECK (buyingliabilities >= 0) |
-sellingliabilities | BIGINT CHECK (sellingliabilities >= 0) |
+extension | TEXT | Extension specific to AccountEntry (XDR)
+ledgerext | TEXT | Extension common to all LedgerEntry types (XDR)
 signers | TEXT | (XDR)
 
 ## offers
@@ -73,6 +73,8 @@ priced | INT NOT NULL | Price.d
 price | DOUBLE PRECISION NOT NULL | computed price n/d, used for ordering offers
 flags | INT NOT NULL |
 lastmodified | INT NOT NULL | lastModifiedLedgerSeq
+extension | TEXT | Extension specific to OfferEntry (XDR)
+ledgerext | TEXT | Extension common to all LedgerEntry types (XDR)
 (offerid) | PRIMARY KEY |
 
 ## trustlines
@@ -91,8 +93,8 @@ tlimit | BIGINT NOT NULL DEFAULT 0 CHECK (tlimit >= 0) | limit
 balance | BIGINT NOT NULL DEFAULT 0 CHECK (balance >= 0) |
 flags | INT NOT NULL |
 lastmodified | INT NOT NULL | lastModifiedLedgerSeq
-buyingliabilities | BIGINT CHECK (buyingliabilities >= 0) |
-sellingliabilities | BIGINT CHECK (sellingliabilities >= 0) |
+extension | TEXT | Extension specific to TrustLineEntry (XDR)
+ledgerext | TEXT | Extension common to all LedgerEntry types (XDR)
 (accountid, issuer, assetcode) | PRIMARY KEY |
 
 ## accountdata
@@ -107,6 +109,8 @@ accountid | VARCHAR(56) NOT NULL | (STRKEY)
 dataname | VARCHAR(88) NOT NULL | (BASE64)
 datavalue | VARCHAR(112) NOT NULL | (BASE64)
 lastmodified | INT NOT NULL | lastModifiedLedgerSeq
+extension | TEXT | Extension specific to DataEntry (XDR)
+ledgerext | TEXT | Extension common to all LedgerEntry types (XDR)
 (accountid, dataname) | PRIMARY KEY |
 
 ## txhistory
--- a/src/database/Database.cpp
+++ b/src/database/Database.cpp
@@ -277,6 +277,7 @@ Database::upgradeToCurrentSchema()
                          std::to_string(SCHEMA_VERSION));
         throw std::runtime_error(s);
     }
+    actBeforeDBSchemaUpgrade();
     while (vers < SCHEMA_VERSION)
     {
         ++vers;
--- a/src/database/Database.cpp
+++ b/src/database/Database.cpp
@@ -14,6 +14,8 @@
 #include "util/Logging.h"
 #include "util/Timer.h"
 #include "util/types.h"
+#include <error.h>
+#include <fmt/format.h>
 
 #include "bucket/BucketManager.h"
 #include "herder/HerderPersistence.h"
@@ -31,6 +33,7 @@
 #include "medida/counter.h"
 #include "medida/metrics_registry.h"
 #include "medida/timer.h"
+#include "xdr/Stellar-ledger-entries.h"
 
 #include <lib/soci/src/backends/sqlite3/soci-sqlite3.h>
 #include <string>
@@ -59,7 +62,7 @@ bool Database::gDriversRegistered = false;
 
 // smallest schema version supported
 static unsigned long const MIN_SCHEMA_VERSION = 9;
-static unsigned long const SCHEMA_VERSION = 12;
+static unsigned long const SCHEMA_VERSION = 13;
 
 // These should always match our compiled version precisely, since we are
 // using a bundled version to get access to carray(). But in case someone
@@ -254,6 +257,33 @@ Database::applySchemaUpgrade(unsigned long vers)
         // the accountbalances index around.
         mSession << "DROP INDEX IF EXISTS accountbalances";
         break;
+    case 13:
+        if (!mApp.getConfig().MODE_USES_IN_MEMORY_LEDGER)
+        {
+            // Add columns for the LedgerEntry extension to each of
+            // the tables that stores a type of ledger entry.
+            addTextColumn("accounts", "ledgerext");
+            addTextColumn("trustlines", "ledgerext");
+            addTextColumn("accountdata", "ledgerext");
+            addTextColumn("offers", "ledgerext");
+            // Absorb the explicit columns of the extension fields of
+            // AccountEntry and TrustLineEntry into single opaque
+            // blobs of XDR each of which represents an entire extension.
+            convertAccountExtensionsToOpaqueXDR();
+            convertTrustLineExtensionsToOpaqueXDR();
+            // Neither earlier schema versions nor the one that we're upgrading
+            // to now had any extension columns in the offers or accountdata
+            // tables, but we add columns in this version, even though we're not
+            // going to use them for anything other than writing out opaque
+            // base64-encoded empty v0 XDR extensions, so that, as with the
+            // other LedgerEntry extensions, we'll be able to add such
+            // extensions in the future without bumping the database schema
+            // version, writing any upgrade code, or changing the SQL that reads
+            // and writes those tables.
+            addTextColumn("offers", "extension");
+            addTextColumn("accountdata", "extension");
+        }
+        break;
     default:
         throw std::runtime_error("Unknown DB schema version");
     }
@@ -290,6 +320,189 @@ Database::upgradeToCurrentSchema()
     }
     CLOG(INFO, "Database") << "DB schema is in current version";
     assert(vers == SCHEMA_VERSION);
+}
+
+void
+Database::addTextColumn(std::string const& table, std::string const& column)
+{
+    std::string addColumnStr("ALTER TABLE " + table + " ADD " + column +
+                             " TEXT;");
+    CLOG(INFO, "Database") << "Adding column '"
+                           << "' to table '" << table << "'";
+    mSession << addColumnStr;
+}
+
+void
+Database::dropNullableColumn(std::string const& table,
+                             std::string const& column)
+{
+    // SQLite doesn't give us a way of dropping a column with a single
+    // SQL command.  If we need it in production, we could re-create the
+    // table without the column and drop the old one.  Since we currently
+    // use SQLite only for testing and PostgreSQL in production, we simply
+    // leave the unused columm around in SQLite at the moment, and NULL
+    // out all of the cells in that column.
+    if (!isSqlite())
+    {
+        std::string dropColumnStr("ALTER TABLE " + table + " DROP COLUMN " +
+                                  column);
+        CLOG(INFO, "Database") << "Dropping column '" << column
+                               << "' from table '" << table << "'";
+
+        mSession << dropColumnStr;
+    }
+    else
+    {
+        std::string nullColumnStr("UPDATE " + table + " SET " + column +
+                                  " = NULL");
+        CLOG(INFO, "Database") << "Setting all cells of column '" << column
+                               << "' in table '" << table << "' to NULL";
+
+        mSession << nullColumnStr;
+    }
+}
+
+std::string
+Database::getOldLiabilitySelect(std::string const& table,
+                                std::string const& fields)
+{
+    return fmt::format("SELECT {}, "
+                       "buyingliabilities, sellingliabilities FROM {} WHERE "
+                       "buyingliabilities IS NOT NULL OR "
+                       "sellingliabilities IS NOT NULL",
+                       fields, table);
+}
+
+void
+Database::convertAccountExtensionsToOpaqueXDR()
+{
+    addTextColumn("accounts", "extension");
+    copyIndividualAccountExtensionFieldsToOpaqueXDR();
+    dropNullableColumn("accounts", "buyingliabilities");
+    dropNullableColumn("accounts", "sellingliabilities");
+}
+
+void
+Database::convertTrustLineExtensionsToOpaqueXDR()
+{
+    addTextColumn("trustlines", "extension");
+    copyIndividualTrustLineExtensionFieldsToOpaqueXDR();
+    dropNullableColumn("trustlines", "buyingliabilities");
+    dropNullableColumn("trustlines", "sellingliabilities");
+}
+
+void
+Database::copyIndividualAccountExtensionFieldsToOpaqueXDR()
+{
+    std::string const tableStr = "accounts";
+
+    CLOG(INFO, "Database") << "Updating extension schema for " << tableStr;
+
+    // <accountID, extension>
+    struct Fields
+    {
+        std::string mAccountID;
+        std::string mExtension;
+    };
+
+    std::string const fieldsStr = "accountid";
+    std::string const selectStr = getOldLiabilitySelect(tableStr, fieldsStr);
+    auto makeFields = [](soci::row const& row) {
+        AccountEntry::_ext_t extension;
+        // getOldLiabilitySelect() places the buying and selling extension
+        // column names after the key field in the SQL select string.
+        extension.v(1);
+        extension.v1().liabilities.buying = row.get<long long>(1);
+        extension.v1().liabilities.selling = row.get<long long>(2);
+        return Fields{.mAccountID = row.get<std::string>(0),
+                      .mExtension =
+                          decoder::encode_b64(xdr::xdr_to_opaque(extension))};
+    };
+
+    std::string const updateStr =
+        "UPDATE accounts SET extension = :ext WHERE accountID = :id";
+    auto prepUpdate = [](soci::statement& st_update, Fields const& data) {
+        st_update.exchange(soci::use(data.mExtension)),
+            st_update.exchange(soci::use(data.mAccountID));
+    };
+
+    auto postUpdate = [](long long const affected_rows, Fields const& data) {
+        if (affected_rows != 1)
+        {
+            throw std::runtime_error(fmt::format(
+                "{}: updating account with account ID {} affected {} row(s) ",
+                __func__, data.mAccountID, affected_rows));
+        }
+    };
+
+    size_t numUpdated = selectUpdateMap<Fields>(
+        *this, selectStr, makeFields, updateStr, prepUpdate, postUpdate);
+
+    CLOG(INFO, "Database") << __func__ << ": updated " << numUpdated
+                           << " records(s) with liabilities in " << tableStr
+                           << " table";
+}
+
+void
+Database::copyIndividualTrustLineExtensionFieldsToOpaqueXDR()
+{
+    std::string const tableStr = "trustlines";
+
+    CLOG(INFO, "Database") << __func__ << ": updating extension schema for "
+                           << tableStr;
+
+    // <accountID, issuer_id, asset_id, extension>
+    struct Fields
+    {
+        std::string mAccountID;
+        std::string mIssuerID;
+        std::string mAssetID;
+        std::string mExtension;
+    };
+
+    std::string const fieldsStr = "accountid, issuer, assetcode";
+    std::string const selectStr = getOldLiabilitySelect(tableStr, fieldsStr);
+    auto makeFields = [](soci::row const& row) {
+        TrustLineEntry::_ext_t extension;
+        // getOldLiabilitySelect() places the buying and selling extension
+        // column names after the three key fields in the SQL select string.
+        extension.v(1);
+        extension.v1().liabilities.buying = row.get<long long>(3);
+        extension.v1().liabilities.selling = row.get<long long>(4);
+        return Fields{.mAccountID = row.get<std::string>(0),
+                      .mIssuerID = row.get<std::string>(1),
+                      .mAssetID = row.get<std::string>(2),
+                      .mExtension =
+                          decoder::encode_b64(xdr::xdr_to_opaque(extension))};
+    };
+
+    std::string const updateStr =
+        "UPDATE trustlines SET extension = :ext WHERE accountID = :id "
+        "AND issuer = :issuer_id AND assetcode = :asset_id";
+    auto prepUpdate = [](soci::statement& st_update, Fields const& data) {
+        st_update.exchange(soci::use(data.mExtension));
+        st_update.exchange(soci::use(data.mAccountID));
+        st_update.exchange(soci::use(data.mIssuerID));
+        st_update.exchange(soci::use(data.mAssetID));
+    };
+
+    auto postUpdate = [](long long const affected_rows, Fields const& data) {
+        if (affected_rows != 1)
+        {
+            throw std::runtime_error(fmt::format(
+                "{}: updating trustline with account ID {}, issuer {}, and "
+                "asset {} affected {} row(s)",
+                __func__, data.mAccountID, data.mIssuerID, data.mAssetID,
+                affected_rows));
+        }
+    };
+
+    size_t numUpdated = selectUpdateMap<Fields>(
+        *this, selectStr, makeFields, updateStr, prepUpdate, postUpdate);
+
+    CLOG(INFO, "Database") << __func__ << ": updated " << numUpdated
+                           << " records(s) with liabilities in " << tableStr
+                           << " table";
 }
 
 void
--- a/src/database/Database.cpp
+++ b/src/database/Database.cpp
@@ -9,6 +9,7 @@
 #include "main/Application.h"
 #include "main/Config.h"
 #include "overlay/StellarXDR.h"
+#include "util/Decoder.h"
 #include "util/GlobalChecks.h"
 #include "util/Logging.h"
 #include "util/Timer.h"
@@ -32,6 +33,7 @@
 #include "medida/timer.h"
 
 #include <lib/soci/src/backends/sqlite3/soci-sqlite3.h>
+#include <string>
 #ifdef USE_POSTGRES
 #include <lib/soci/src/backends/postgresql/soci-postgresql.h>
 #endif
--- a/src/database/Database.h
+++ b/src/database/Database.h
@@ -248,6 +248,72 @@ class DBTimeExcluder : NonCopyable
     ~DBTimeExcluder();
 };
 
+// Select a set of records using a client-defined query string, then map
+// each record into an element of a client-defined datatype by applying a
+// client-defined function (the records are accumulated in the "out"
+// vector).
+template <typename T>
+void
+selectMap(Database& db, std::string const& selectStr,
+          std::function<T(soci::row const&)> makeT, std::vector<T>& out)
+{
+    soci::rowset<soci::row> rs = (db.getSession().prepare << selectStr);
+
+    std::transform(rs.begin(), rs.end(), std::back_inserter(out), makeT);
+}
+
+// Map each element in the given vector of a client-defined datatype into a
+// SQL update command by applying a client-defined function, then send those
+// update strings to the database.
+//
+// The "postUpdate" function receives the number of records affected
+// by the given update, as well as the element of the client-defined
+// datatype which generated that update.
+template <typename T>
+void updateMap(Database& db, std::vector<T> const& in,
+               std::string const& updateStr,
+               std::function<void(soci::statement&, T const&)> prepUpdate,
+               std::function<void(long long const, T const&)> postUpdate);
+template <typename T>
+void
+updateMap(Database& db, std::vector<T> const& in, std::string const& updateStr,
+          std::function<void(soci::statement&, T const&)> prepUpdate,
+          std::function<void(long long const, T const&)> postUpdate)
+{
+    auto st_update = db.getPreparedStatement(updateStr).statement();
+
+    for (auto& recT : in)
+    {
+        prepUpdate(st_update, recT);
+        st_update.define_and_bind();
+        st_update.execute(true);
+        auto affected_rows = st_update.get_affected_rows();
+        st_update.clean_up(false);
+        postUpdate(affected_rows, recT);
+    }
+}
+
+// The composition of updateMap() following selectMap().
+//
+// Returns the number of records selected by selectMap() (all of which were
+// then passed through updateMap() before the selectUpdateMap() call
+// returned).
+template <typename T>
+size_t
+selectUpdateMap(Database& db, std::string const& selectStr,
+                std::function<T(soci::row const&)> makeT,
+                std::string const& updateStr,
+                std::function<void(soci::statement&, T const&)> prepUpdate,
+                std::function<void(long long const, T const&)> postUpdate)
+{
+    std::vector<T> vecT;
+
+    selectMap<T>(db, selectStr, makeT, vecT);
+    updateMap<T>(db, vecT, updateStr, prepUpdate, postUpdate);
+
+    return vecT.size();
+}
+
 template <typename T>
 void
 decodeOpaqueXDR(std::string const& in, T& out)
--- a/src/database/Database.h
+++ b/src/database/Database.h
@@ -247,4 +247,23 @@ class DBTimeExcluder : NonCopyable
     DBTimeExcluder(Application& mApp);
     ~DBTimeExcluder();
 };
+
+template <typename T>
+void
+decodeOpaqueXDR(std::string const& in, T& out)
+{
+    std::vector<uint8_t> opaque;
+    decoder::decode_b64(in, opaque);
+    xdr::xdr_from_opaque(opaque, out);
+}
+
+template <typename T>
+void
+decodeOpaqueXDR(std::string const& in, soci::indicator const& ind, T& out)
+{
+    if (ind == soci::i_ok)
+    {
+        decodeOpaqueXDR(in, out);
+    }
+}
 }
--- a/src/database/Database.h
+++ b/src/database/Database.h
@@ -7,11 +7,14 @@
 #include "database/DatabaseTypeSpecificOperation.h"
 #include "medida/timer_context.h"
 #include "overlay/StellarXDR.h"
+#include "util/Decoder.h"
 #include "util/NonCopyable.h"
 #include "util/Timer.h"
+#include <functional>
 #include <set>
 #include <soci.h>
 #include <string>
+#include <xdrpp/marshal.h>
 
 namespace medida
 {
@@ -108,6 +111,10 @@ class Database : NonMovableOrCopyable
     // if there is a connection error, this will throw.
     Database(Application& app);
 
+    virtual ~Database()
+    {
+    }
+
     // Return a crude meter of total queries to the db, for use in
     // overlay/LoadManager.
     medida::Meter& getQueryMeter();
@@ -196,6 +203,16 @@ class Database : NonMovableOrCopyable
     // Access the optional SOCI connection pool available for worker
     // threads. Throws an error if !canUsePool().
     soci::connection_pool& getPool();
+
+  protected:
+    // Give clients the opportunity to perform operations on databases while
+    // they're still using old schemas (prior to the upgrade that occurs either
+    // immediately after database creation or after loading a version of
+    // stellar-core that introduces a new schema).
+    virtual void
+    actBeforeDBSchemaUpgrade()
+    {
+    }
 };
 
 template <typename T>
--- a/src/database/Database.h
+++ b/src/database/Database.h
@@ -106,6 +106,22 @@ class Database : NonMovableOrCopyable
     static void registerDrivers();
     void applySchemaUpgrade(unsigned long vers);
 
+    // Convert the accounts table from using explicit entries for
+    // extension fields into storing the entire extension as opaque XDR.
+    void convertAccountExtensionsToOpaqueXDR();
+    void copyIndividualAccountExtensionFieldsToOpaqueXDR();
+
+    std::string getOldLiabilitySelect(std::string const& table,
+                                      std::string const& fields);
+    void addTextColumn(std::string const& table, std::string const& column);
+    void dropNullableColumn(std::string const& table,
+                            std::string const& column);
+
+    // Convert the trustlines table from using explicit entries for
+    // extension fields into storing the entire extension as opaque XDR.
+    void convertTrustLineExtensionsToOpaqueXDR();
+    void copyIndividualTrustLineExtensionFieldsToOpaqueXDR();
+
   public:
     // Instantiate object and connect to app.getConfig().DATABASE;
     // if there is a connection error, this will throw.
--- a/src/database/test/DatabaseTests.cpp
+++ b/src/database/test/DatabaseTests.cpp
@@ -349,3 +349,48 @@ TEST_CASE("schema test", "[db]")
     auto av = db.getAppSchemaVersion();
     REQUIRE(dbv == av);
 }
+
+class SchemaUpgradeTestApplication : public TestApplication
+{
+  public:
+    using PreUpgradeFunc = std::function<void(SchemaUpgradeTestApplication&)>;
+
+  private:
+    class SchemaUpgradeTestDatabase : public Database
+    {
+        SchemaUpgradeTestApplication& mApp;
+        PreUpgradeFunc const mPreUpgradeFunc;
+
+      public:
+        SchemaUpgradeTestDatabase(SchemaUpgradeTestApplication& app,
+                                  PreUpgradeFunc preUpgradeFunc)
+            : Database(app), mApp(app), mPreUpgradeFunc(preUpgradeFunc)
+        {
+        }
+
+        virtual void
+        actBeforeDBSchemaUpgrade() override
+        {
+            if (mPreUpgradeFunc)
+            {
+                mPreUpgradeFunc(mApp);
+            }
+        }
+    };
+
+    PreUpgradeFunc const mPreUpgradeFunc;
+
+  public:
+    SchemaUpgradeTestApplication(VirtualClock& clock, Config const& cfg,
+                                 PreUpgradeFunc preUpgradeFunc)
+        : TestApplication(clock, cfg), mPreUpgradeFunc(preUpgradeFunc)
+    {
+    }
+
+    virtual std::unique_ptr<Database>
+    createDatabase() override
+    {
+        return std::make_unique<SchemaUpgradeTestDatabase>(*this,
+                                                           mPreUpgradeFunc);
+    }
+};
--- a/src/database/test/DatabaseTests.cpp
+++ b/src/database/test/DatabaseTests.cpp
@@ -4,16 +4,22 @@
 
 #include "util/asio.h"
 #include "crypto/Hex.h"
+#include "crypto/KeyUtils.h"
 #include "database/Database.h"
+#include "ledger/LedgerTxn.h"
+#include "ledger/test/LedgerTestUtils.h"
 #include "lib/catch.hpp"
 #include "main/Application.h"
 #include "main/Config.h"
 #include "test/TestUtils.h"
 #include "test/test.h"
+#include "util/Decoder.h"
 #include "util/Logging.h"
 #include "util/Math.h"
 #include "util/Timer.h"
 #include "util/TmpDir.h"
+#include "util/optional.h"
+#include <algorithm>
 #include <random>
 
 using namespace stellar;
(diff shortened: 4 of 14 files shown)


rokopt:
There must be something that I'm failing to do during the upgrade.  I can create a pubnet database with schema 12, run on it for a while, and then upgrade it with my new code.  And I can create a new database with schema 13 with my new code, and run on that.  But if I run on schema 12 for a while and then upgrade with my new code, I can't run with the upgraded database; this happens on startup:

```2020-06-23T12:42:07.369 GDJYT [Ledger INFO] Last closed ledger (LCL) hash is 78d269aaddb5f15a51c56873e2178168a3b57eff3cdbfcbccc5faa0b7f6fc22b
2020-06-23T12:42:07.370 GDJYT [Ledger INFO] Loaded LCL header from database: [seq=30215056, hash=78d269]
2020-06-23T12:42:07.371 GDJYT [Ledger ERROR] 30 buckets are missing from bucket directory 'buckets' [LedgerManagerImpl.cpp:329]
2020-06-23T12:42:07.372 GDJYT [default FATAL] Got an exception: Bucket directory is corrupt [ApplicationUtils.cpp:75]```

**Comment by rokopt** [comment]
> There must be something that I'm failing to do during the upgrade. I can create a pubnet database with schema 12, run on it for a while, and then upgrade it with my new code. And I can create a new database with schema 13 with my new code, and run on that. But if I run on schema 12 for a while and then upgrade with my new code, I can't run with the upgraded database; this happens on startup:
> 
> ```
> 2020-06-23T12:42:07.369 GDJYT [Ledger INFO] Last closed ledger (LCL) hash is 78d269aaddb5f15a51c56873e2178168a3b57eff3cdbfcbccc5faa0b7f6fc22b
> 2020-06-23T12:42:07.370 GDJYT [Ledger INFO] Loaded LCL header from database: [seq=30215056, hash=78d269]
> 2020-06-23T12:42:07.371 GDJYT [Ledger ERROR] 30 buckets are missing from bucket directory 'buckets' [LedgerManagerImpl.cpp:329]
> 2020-06-23T12:42:07.372 GDJYT [default FATAL] Got an exception: Bucket directory is corrupt [ApplicationUtils.cpp:75]
> ```

Oh, I think that was operator rather than code error on my part -- I did the two runs (with the old code and then the new code) from two different working directories, and I think that by default the "buckets" directory is placed in the current working directory.  So I was using the wrong buckets directory for the run with the new code.

**Comment by rokopt** [comment]
> > There must be something that I'm failing to do during the upgrade. I can create a pubnet database with schema 12, run on it for a while, and then upgrade it with my new code. And I can create a new database with schema 13 with my new code, and run on that. But if I run on schema 12 for a while and then upgrade with my new code, I can't run with the upgraded database; this happens on startup:
> > ```
> > 2020-06-23T12:42:07.369 GDJYT [Ledger INFO] Last closed ledger (LCL) hash is 78d269aaddb5f15a51c56873e2178168a3b57eff3cdbfcbccc5faa0b7f6fc22b
> > 2020-06-23T12:42:07.370 GDJYT [Ledger INFO] Loaded LCL header from database: [seq=30215056, hash=78d269]
> > 2020-06-23T12:42:07.371 GDJYT [Ledger ERROR] 30 buckets are missing from bucket directory 'buckets' [LedgerManagerImpl.cpp:329]
> > 2020-06-23T12:42:07.372 GDJYT [default FATAL] Got an exception: Bucket directory is corrupt [ApplicationUtils.cpp:75]
> > ```
> 
> Oh, I think that was operator rather than code error on my part -- I did the two runs (with the old code and then the new code) from two different working directories, and I think that by default the "buckets" directory is placed in the current working directory. So I was using the wrong buckets directory for the run with the new code.

Confirmed, that was my operator error, using different bucket directories for the two runs.

**Comment by rokopt** [comment]
Performance note:

After running overnight on the pubnet with DB schema v12 (prior to this PR), I switched over to v13 (a binary with the changes from this PR), and, as it consistently has so far when I've tested upgrades on owlbear, it took ~20s to upgrade the database schema, which involved rewriting ~60K accounts records and ~120K trustlines records (that's how many have liabilities in the current database).  Including that time and the ensuing catchup, it took ~2m40s for the node to sync back up on v13.

File: src/database/test/DatabaseTests.cpp [hunk 1]
```diff
@@ -349,3 +355,430 @@ TEST_CASE("schema test", "[db]")
...
+            soci::use(inflationDestStr, inflationDestInd),
+            soci::use(homeDomainStr), soci::use(thresholdsStr),
+            soci::use(signersStr), soci::use(ae.flags),
+            soci::use(app.getLedgerManager().getLastClosedLedgerNum()),
+            soci::use(liabilities ? liabilities->buying : 0),
+            soci::use(liabilities ? liabilities->selling : 0);
+
+        if (!liabilities)
+        {
+            session << "UPDATE accounts SET buyingliabilities = NULL, "
```
jonjove, line 444: Why not just use indicators on line 439, 440 instead of this?

src/database/test/DatabaseTests.cpp
```diff
@@ -349,3 +355,430 @@ TEST_CASE("schema test", "[db]")
...
+            soci::use(accountIDStr), soci::use(assetType), soci::use(issuerStr),
+            soci::use(assetCodeStr), soci::use(tl.limit), soci::use(tl.balance),
+            soci::use(tl.flags),
+            soci::use(app.getLedgerManager().getLastClosedLedgerNum()),
+            soci::use(liabilities ? liabilities->buying : 0),
+            soci::use(liabilities ? liabilities->selling : 0);
+
+        if (!liabilities)
+        {
+            session << "UPDATE trustlines SET buyingliabilities = NULL, "
```
jonjove, line 481: Same thing about indicators.

src/database/test/DatabaseTests.cpp
```diff
@@ -349,3 +355,430 @@ TEST_CASE("schema test", "[db]")
...
+            CLOG(FATAL, "Database") << __func__ << ": exception " << e.what()
+                                    << " while adding old-schema offer";
+            throw;
+        }
+    };
+
+    auto testOneDBMode = [prepOldSchemaDB](Config::TestDbMode const dbMode) {
+        // A vector of optional Liabilities entries, for each of which the test
+        // will generate a valid account.
+        auto const accOptLiabilities = {
```
jonjove, line 596: As mentioned about the types above, I don't really understand why this approach is valuable. But regardless: why use different liabilities for accounts and trust lines, and why have repeats? Also you're missing the important (0, 0) case for accounts.

src/database/test/DatabaseTests.cpp
```diff
@@ -349,3 +355,430 @@ TEST_CASE("schema test", "[db]")
...
+
+        // Validate that the accounts and trustlines have the expected
+        // liabilities and ledger entry extensions now that the database upgrade
+        // has completed.
+
+        LedgerTxn ltx(app->getLedgerTxnRoot());
+
+        for (auto aol : accOptLiabilitiesVec)
+        {
+            LedgerKey key;
```
jonjove, line 665: Use the `LedgerEntryKey` function (applies in multiple places).

src/database/test/DatabaseTests.cpp
```diff
@@ -349,3 +355,430 @@ TEST_CASE("schema test", "[db]")
...
+            key.type(ACCOUNT);
+            key.account().accountID = aol.first.accountID;
+            auto acc = ltx.load(key);
+            REQUIRE(acc.current().data.type() == ACCOUNT);
```
jonjove, line 669: Why effectively write your own comparison operator? Use the built-in XDR comparison operators for `LedgerEntry` instead. It'll save a lot of effort, reduce the risk of missing something, and improve clarity. (This applies in multiple places)

docs/db-schema.md
```diff
@@ -51,8 +51,8 @@ homedomain | VARCHAR(44) | (BASE64)
 thresholds | TEXT | (BASE64)
 flags | INT NOT NULL |
 lastmodified | INT NOT NULL | lastModifiedLedgerSeq
-buyingliabilities | BIGINT CHECK (buyingliabilities >= 0) |
-sellingliabilities | BIGINT CHECK (sellingliabilities >= 0) |
+extension | TEXT | Opaque XDR extension specific to AccountEntry
```
jonjove, line 54: Nitpick: for consistency with the rest of the file these should read like "Extension specific to AccountEntry (XDR)". See the legend at the top of this file.

src/database/Database.cpp
```diff
@@ -346,18 +334,17 @@ Database::addTextColumn(std::string const& table, std::string const& column)
 }
 
 void
-Database::addTextColumnWithDefault(std::string const& table,
-                                   std::string const& column,
-                                   std::string const& defaultVal)
+Database::addTextColumnAndFill(std::string const& table,
```
jonjove, line 337: Let's remove this if it isn't used anymore. I don't believe in maintaining code that we aren't using, because it requires effort for no reward. Furthermore, I believe heavily in testing and if code is unused then it is _definitely_ untested.

src/main/Application.h
```diff
@@ -278,15 +278,20 @@ class Application
 
     virtual AbstractLedgerTxnParent& getLedgerTxnRoot() = 0;
 
+    // For tests to perform operations on databases while they're
+    // still using old schemas.
+    virtual void actBeforeDBSchemaUpgrade() = 0;
```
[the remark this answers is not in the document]
rokopt: I think it's the comment that should change -- there's nothing about it that specifically requires the client to be a test; it could be any client of Application.  I'll do that.

src/database/Database.h
```diff
@@ -196,6 +197,40 @@ class Database : NonMovableOrCopyable
     // Access the optional SOCI connection pool available for worker
     // threads. Throws an error if !canUsePool().
     soci::connection_pool& getPool();
+
+    // Select a set of records using a client-defined query string, then map
+    // each record into an element of a client-defined datatype by applying a
+    // client-defined function (the records are accumulated in the "out"
+    // vector).
+    template <typename T>
+    void selectMap(std::string const& selectStr,
```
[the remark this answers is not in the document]
rokopt: > Therefore, attempting to use these template functions by including Database.h anywhere except Database.cpp will lead to a link failure. That being said, these functions aren't used outside of Database.cpp right now so you may as well just remove the declarations from the header.

I don't understand what you mean about removing the declarations from the header -- that leads to:

```
database/Database.cpp:436:11: error: out-of-line definition of 'selectMap' does not match any declaration in 'stellar::Database'
Database::selectMap(std::string const& selectStr,
          ^~~~~~~~~
```

Also, I wrote those generically with the intent that we might well be able to start using them to factor out some of the repeated code that currently exists in the *SQL.cpp files, or at least to avoid further repetition in the future.  If we do end up using selectMap\<Foo\> in some source file other than Database.cpp, can't we just make it link by also adding the following to Database.cpp?

```
template void Database::selectMap<Foo>(std::string const& selectStr,
                   std::function<Foo(soci::row const&)> makeT,
                   std::vector<Foo>& out);
```

(See for example https://raymii.org/s/snippets/Cpp_template_definitions_in_a_cpp_file_instead_of_header.html.)

src/database/Database.cpp
```diff
@@ -290,6 +335,115 @@ Database::upgradeToCurrentSchema()
...
+{
+    std::string addColumnStr("ALTER TABLE " + table + " ADD " + column +
+                             " TEXT;");
+    CLOG(INFO, "Database") << "Adding column with string '" << addColumnStr
+                           << "'";
+    mSession << addColumnStr;
+}
+
+void
+Database::addTextColumnWithDefault(std::string const& table,
```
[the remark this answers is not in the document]
rokopt: You mentioned this comment in person, and I agreed, and have changed it accordingly.

src/database/Database.h
```diff
@@ -191,6 +209,11 @@ class Database : NonMovableOrCopyable
     // Check schema version and apply any upgrades if necessary.
     void upgradeToCurrentSchema();
 
+    // The name of the column containing the ledger entry extension,
+    // which appears in all the tables which contain any type of
+    // ledger entry.
+    static std::string const ledgerExtName;
```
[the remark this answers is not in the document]
rokopt: > Does this really add clarity though? For example, to remember the name of the constant you must remember "ledgerExt" _plus_ an additional word, whereas the constant is just "ledgerext". After reviewing more of the PR, I'm becoming more convinced this is not adding clarity.

Which additional word?  Do you mean ```Database::```?

I agree that it does not read particularly well, but I did that because I wanted to have the compiler ensure first that the LedgerEntry extension had the same name in each table in which it appears (that part _is_ just for clarity) and that the compiler would prevent any misspellings.  But it does make the client code more verbose and less readable.  I'll change it if you think it's a better tradeoff the other way.  (If you think the compiler-time consistency is worth the client-code verbosity, then I'll just change the constant name to match our convention.)

What I'd prefer to both options would be for there to be a separate module that factors out all the raw SQL generation and allowed clients to write all their uses from methods such as ```generateGenericSelectSQL()``` and ```generatePostgresUpdateSQL()```, which would protect from both misspellings and argument mismatches (in particular, all those client wrapper methods could be unit-tested -- the missing "v12" that you caught below in code review only turned up in practice when the buckets were >70% through being replayed, so some strings are quite rarely used!) and make the client code easier rather than harder to read (standard C++ method calls rather than long strings of not-quite-repeated raw SQL throughout all the *SQL.cpp files).  But that'd be a PR of its own, not something to clutter this one with.

src/database/Database.cpp
```diff
@@ -252,6 +258,45 @@ Database::applySchemaUpgrade(unsigned long vers)
...
+            // Neither earlier schema versions nor the one that we're upgrading
+            // to now had any extension columns in the offers or accountdata
+            // tables, but we add columns in this version, even though we're not
+            // going to use them for anything other than writing out opaque
+            // base64-encoded empty v0 XDR extensions, so that, as with the
+            // other LedgerEntry extensions, we'll be able to add such
+            // extensions in the future without bumping the database schema
+            // version, writing any upgrade code, or changing the SQL that reads
+            // and writes those tables.
+            addTextColumnWithDefault(
```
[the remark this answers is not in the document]
rokopt: Now that the new columns aren't being filled, but will just be NULL, perhaps that overhead will be less?  (Of course, I'll implement whatever you and @MonsieurNicolas decide you prefer.)

src/database/Database.cpp
```diff
@@ -252,6 +258,45 @@ Database::applySchemaUpgrade(unsigned long vers)
...
         mSession << "DROP INDEX IF EXISTS accountbalances";
         break;
+    case 13:
+        if (!mApp.getConfig().MODE_USES_IN_MEMORY_LEDGER)
+        {
+            // Add columns for the LedgerEntry extension to each of
+            // the tables that stores a type of ledger entry.
+            addTextColumnWithDefault(
+                "accounts", ledgerExtName,
+                decoder::encode_b64(xdr::xdr_to_opaque(LedgerEntry::_ext_t())));
```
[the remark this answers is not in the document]
rokopt: Yeah -- the time to do the upgrade with the filling code on the real pubnet database was ~2m.  Without the filling, as the code stands now, it's ~20s.  I don't yet have enough context to know whether ~2m is a big deal or not, but I don't think that not-filling is likely to be a big deal, so we probably may as well go with the ~20s (i.e. the current code, where the filling has already been removed).

src/database/Database.cpp
```diff
@@ -290,6 +335,115 @@ Database::upgradeToCurrentSchema()
...
+                                  column);
+        CLOG(INFO, "Database")
+            << "Dropping column '" << column << "' with string '"
+            << dropColumnStr << "' from table '" << table << "'";
+
+        mSession << dropColumnStr;
+    }
+    else
+    {
+        std::string nullColumnStr("UPDATE " + table + " SET " + column +
```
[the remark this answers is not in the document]
rokopt: See above (the ~2m and ~20s figures are for the complete upgrade, including the filling of all the new columns in the first case).

src/ledger/LedgerTxnOfferSQL.cpp
```diff
@@ -407,21 +469,25 @@ class BulkUpsertOffersOperation : public DatabaseTypeSpecificOperation<void>
...
+            ") ON CONFLICT (offerid) DO UPDATE SET "
+            "sellerid = excluded.sellerid, "
+            "sellingasset = excluded.sellingasset, "
+            "buyingasset = excluded.buyingasset, "
+            "amount = excluded.amount, "
+            "pricen = excluded.pricen, "
+            "priced = excluded.priced, "
+            "price = excluded.price, "
+            "flags = excluded.flags, "
+            "lastmodified = excluded.lastmodified, " +
```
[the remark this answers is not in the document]
rokopt: Removed (in my local repo, soon to be pushed).

src/ledger/LedgerTxnOfferSQL.cpp
```diff
@@ -473,33 +541,40 @@ class BulkUpsertOffersOperation : public DatabaseTypeSpecificOperation<void>
...
+            "unnest(:v2::BIGINT[]), "
+            "unnest(:v3::TEXT[]), "
+            "unnest(:v4::TEXT[]), "
+            "unnest(:v5::BIGINT[]), "
+            "unnest(:v6::INT[]), "
+            "unnest(:v7::INT[]), "
+            "unnest(:v8::DOUBLE PRECISION[]), "
+            "unnest(:v9::INT[]), "
+            "unnest(:v10::INT[]), "
+            "unnest(:v11::TEXT[]) "
```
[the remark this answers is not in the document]
rokopt: Yep, fixed later.

src/database/Database.cpp
rokopt: It was tested quite a bit while it _was_ being used, and anyone who were to use it in the future certainly ought to add a test for that use.  :)  But I'll remove it.

docs/db-schema.md
rokopt: Done.

src/ledger/LedgerTxnAccountSQL.cpp
```diff
@@ -21,80 +21,98 @@ namespace stellar
...
-    st.exchange(soci::into(homeDomain));
-    st.exchange(soci::into(thresholds));
-    st.exchange(soci::into(account.flags));
-    st.exchange(soci::into(le.lastModifiedLedgerSeq));
-    st.exchange(soci::into(liabilities.buying, buyingLiabilitiesInd));
-    st.exchange(soci::into(liabilities.selling, sellingLiabilitiesInd));
-    st.exchange(soci::into(signers, signersInd));
-    st.exchange(soci::use(actIDStrKey));
-    st.define_and_bind();
+    try
```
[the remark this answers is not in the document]
rokopt: Agreed, I was having trouble getting gdb working in the context of "make check", but I think I got that working eventually, and either way that shouldn't lead to any artefacts in the production code.  I've removed the try/catch/throw blocks.

src/ledger/LedgerTxnTrustLineSQL.cpp
```diff
@@ -48,53 +50,73 @@ getTrustLineStrings(AccountID const& accountID, Asset const& asset,
...
-    st.exchange(soci::into(tl.balance));
-    st.exchange(soci::into(tl.flags));
-    st.exchange(soci::into(le.lastModifiedLedgerSeq));
-    st.exchange(soci::into(liabilities.buying, buyingLiabilitiesInd));
-    st.exchange(soci::into(liabilities.selling, sellingLiabilitiesInd));
-    st.exchange(soci::use(accountIDStr));
-    st.exchange(soci::use(issuerStr));
-    st.exchange(soci::use(assetStr));
-    st.define_and_bind();
+    try
```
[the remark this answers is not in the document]
rokopt: Also removed.

src/database/test/DatabaseTests.cpp
```diff
@@ -349,3 +355,430 @@ TEST_CASE("schema test", "[db]")
...
+        : TestApplication(clock, cfg), mPreUpgradeFunc(preUpgradeFunc)
+    {
+    }
+
+    virtual void
+    actBeforeDBSchemaUpgrade() override
+    {
+        if (mPreUpgradeFunc)
+        {
+            try
```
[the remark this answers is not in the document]
rokopt: Agreed; removed.

src/database/test/DatabaseTests.cpp
rokopt: Added (0,0) for accounts.  As for there being two different lists, I had two possibly-silly reasons:  I didn't want to give the impression that there was any dependency on their being the same, and I figured I'd take advantage of the there being two different types of ledger entries with liabilities by testing lists with different characteristics (different lengths, one beginning with a nullopt but not ending with one while the other was the other way around).  For the moment, having explained why, I'll leave it that way, but I'll squash the two lists into one if you still prefer I do so.

src/database/test/DatabaseTests.cpp
rokopt: I wanted to test with XDR's ==, but as the code was, that comparison didn't hold!  The extensions could have differed, because the loaded one could have been changed by the optional Liabilities paired with the given entries.  However, now that I've implemented your suggestion of just using the extensions themselves to represent the optional<>, the operator== works.  So your earlier idea led to a cascade of cleanups.  Thanks!

src/database/test/DatabaseTests.cpp
rokopt: Done.

src/database/test/DatabaseTests.cpp
rokopt: I couldn't figure out how to get it to typecheck!  The compiler was telling me that I was trying to convert a ```size_t``` into an ```indicator``` for the second argument, and I was baffled as to why (```soci::i_ok``` sure looks like an ```indicator``` to me).

Since you asked, I went back and fought with the compiler for a while longer.  Eventually I figured out that the problem was that the overload that I was _actually_ trying to use declares the input ```indicator``` as a _non-_```const``` reference.  It did not occur to me that it might do that, since on an update, I figured, we were only reading from the indicator parameter, not modifying it.

So, I made a non-```const``` local to hold ```soci::i_ok``` (or ```soci::i_null```, as appropriate), and with that I've been able to implement your suggestion.

src/database/test/DatabaseTests.cpp
rokopt: Same resolution as above.

src/database/test/DatabaseTests.cpp
```diff
@@ -349,3 +355,430 @@ TEST_CASE("schema test", "[db]")
...
+    }
+};
+
+// This tests upgrading from MIN_SCHEMA_VERSION to SCHEMA_VERSION by creating
+// ledger entries with the old MIN_SCHEMA_VERSION after database creation, then
+// validating their contents after the upgrade to SCHEMA_VERSION.
+TEST_CASE("schema upgrade test", "[db]")
+{
+    using OptLiabilities = stellar::optional<Liabilities>;
+    using AccOptLiabilitiesPair = std::pair<AccountEntry, OptLiabilities>;
```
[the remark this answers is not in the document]
rokopt: Good idea -- it was the way it was because in an earlier version of the code I was generating the ```AccountEntry``` and ```TrustLineEntry``` objects within the lambdas, and only passing in ```optional<Liabilities>```.  After I started generating the ```AccountEntry```/```TrustLineEntry``` objects in the caller of the lambas and passing them into the lambdas, it didn't occur to me that that could allow me to remove the ```optional<Liabilities>``` from the parameter lists.  I have now done so.

src/ledger/LedgerTxnTrustLineSQL.cpp
```diff
@@ -48,53 +50,73 @@ getTrustLineStrings(AccountID const& accountID, Asset const& asset,
...
+        tl.asset = key.trustLine().asset;
 
-    tl.accountID = key.trustLine().accountID;
-    tl.asset = key.trustLine().asset;
+        if (extensionInd == soci::i_ok)
+        {
+            tl.ext.v(1);
+            std::vector<uint8_t> extensionOpaque;
+            decoder::decode_b64(extensionStr, extensionOpaque);
+            xdr::xdr_from_opaque(extensionOpaque, tl.ext.v1());
```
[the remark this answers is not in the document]
rokopt: Agreed, I like the consistency of your suggestion better.  Changed.

src/ledger/LedgerTxnOfferSQL.cpp
```diff
@@ -275,6 +293,20 @@ LedgerTxnRoot::Impl::loadOffers(StatementContext& prep,
         oe.selling = processAsset(sellingAsset);
         oe.buying = processAsset(buyingAsset);
 
+        if (extensionInd == soci::i_ok)
```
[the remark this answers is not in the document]
rokopt: Agreed, that repetition was annoying me too.  Your comment has spurred me to factor it out.  :)

rokopt:
I _think_ I've addressed all the review comments so far, mostly by just making the suggested changes, and in a few cases proposing and implementing different ones or offering reasons to leave it as it is.  I've finally marked this as ready for review, as opposed to a draft, though further review and discussion will almost certainly lead to some further changes.

rokopt:
Here's a raw benchmark log from owlbear, comparing a series of ledger-entry-related [bench] tests, with a v12 run first and then a v13 run.  I have not done enough to get an idea of the variance, but at a first glance nothing has jumped out at me as untoward.

[benchmarks-v12-vs-v13.txt](https://github.com/stellar/stellar-core/files/4829233/benchmarks-v12-vs-v13.txt)

src/database/Database.h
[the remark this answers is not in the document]
jonjove: I mean remembering the name "ledgerExtName" means you have already remembered "ledgerext", so it doesn't save any mental effort at all. I don't think the compile-time consistency helps much here: if there were a mis-spelling then the code definitely will not pass our acceptance tests.

I agree that factoring out raw SQL generation sounds like a really good idea. Almost all of our queries have a few standard parts (`WITH` statements for example) that we can use as components. But as you note, "that'd be a PR of its own." How about you open an issue to address the bigger problem, and stick with hard coded here?

src/database/test/DatabaseTests.cpp
```diff
@@ -612,105 +627,58 @@ TEST_CASE("schema upgrade test", "[db]")
...
-                    aol.second->selling);
-            }
-            else
-            {
-                REQUIRE(acc.current().data.account().ext.v() == 0);
-            }
+            LedgerEntry entry;
+            entry.data.type(ACCOUNT);
+            entry.data.account() = ae;
+            key = LedgerEntryKey(entry);
```
jonjove, line 636: Nitpick: could declare `key` at this point (applies in multiple places)

src/database/test/DatabaseTests.cpp
```diff
@@ -612,105 +627,58 @@ TEST_CASE("schema upgrade test", "[db]")
...
+            auto ae_upgraded = ltx.load(key);
```
jonjove, line 637: Nitpick: naming convention, should be `aeUpgraded` (applies in multiple places)

src/database/test/DatabaseTests.cpp
```diff
@@ -612,105 +627,58 @@ TEST_CASE("schema upgrade test", "[db]")
...
+            REQUIRE(ae_upgraded.current().data.type() == ACCOUNT);
```
jonjove, line 638: I still think you're working too hard here. It would be more compact and reliable to use
```c++
operator==(LedgerEntry const&, LedgerEntry const&)
```

Is the reason you aren't doing this that those comparisons fail because you don't have the right `lastModifiedLedgerSeq`? If so, it would probably be better to just fix the root cause. Or is it something else?

src/database/Database.cpp
```diff
@@ -443,6 +444,44 @@ Database::selectUpdateMap(
     return vecT.size();
 }
 
+template <typename T>
+void
+Database::decodeOpaqueXDR(std::string const& in, T& out)
```
jonjove, line 449: These should be free functions. I'm not even certain that this first overload should be defined in a "database" file; it's a generic utility function.

src/database/Database.cpp
```diff
@@ -443,6 +444,44 @@ Database::selectUpdateMap(
...
+                          T& out)
+{
+    if (ind == soci::i_ok)
+    {
+        decodeOpaqueXDR(in, out);
+    }
+}
+
+// Explicit instantiations of decodeOpaqueXDR() for clients to link with.
+template void
```
jonjove, line 468: This is the same thing as the `selectMap` functions. Publicly accessible template functions should be declared _and_ defined in a header file. Then you don't need explicit instantiations. For perspective: the template functions you wrote have either 3 line or 4 line implementations, yet most of these declarations are 4 lines! Furthermore, doing this leads to bloat: there won't be a compile error or link error if I stop using one of these functions yet it will persist in my executable.

src/database/Database.h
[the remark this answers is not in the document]
jonjove: You can do that, but it doesn't mean you should. The link you shared starts with the following paragraph:

> In this snippet I'll show you how to place your C++ template definitions in a seperate .cpp file. I'd recommend you to just put template definitions in your header file, or a .hpp file, but if you really want to there is a trick to get them in a seperate .cpp file. The trick is to explicitly instanciate every template you're going to use at the end of the .cpp file. With many different templates and types this becomes cumbersome, but for certain usecases it could be useful.

Basically, if you want to take this approach the burden of proof is on you to justify why it is the right thing to do. I laid out some reasons in a separate comment about why I don't generally support this approach. 

----
My comment was confusing, sorry. You are right that removing them from the header wouldn't work immediately. What I meant was that you could remove them from the header and make them free functions. You'll see that I really like free functions whenever possible. The reason is that it makes it extremely clear that you are respecting encapsulation, and that a function doesn't depend on implementation details. In fact, this change presents a beautiful example in action. Your function `selectMap` is a member function, and therefore proceeds to access private variables such as `mSession`. But `mSession` has a getter `getSession` with a non-trivial implementation! There is a check in there to make sure you are on the main thread. You're bypassing that check, but a free function would not have been able to do so.

To me this is a compelling argument for changing to free functions. But if you aren't convinced there are some other options:
- You can at least make the functions private. As noted in the previous paragraph, they present a foot-gun so they shouldn't be public (it might be safe to bypass the check in your case, but it probably isn't in general).
- You can fix the implementations and move the definitions into the header.

src/main/ApplicationImpl.h
```diff
@@ -116,6 +116,8 @@ class ApplicationImpl : public Application
 
     virtual AbstractLedgerTxnParent& getLedgerTxnRoot() override;
 
+    virtual void actBeforeDBSchemaUpgrade() override;
```
jonjove, line 119: Do you think this should be a virtual function of `Database` instead of `Application`? Seems like separation of concerns.

src/database/Database.h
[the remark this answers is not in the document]
rokopt: > How about you open an issue to address the bigger problem, and stick with hard coded here?

That sounds good to me.  I've changed to code to use a hardcoded "ledgerext", as with other columns, and have filed issue #2594 .

src/database/Database.h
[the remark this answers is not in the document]
rokopt: > Basically, if you want to take this approach the burden of proof is on you to justify why it is the right thing to do. I laid out some reasons in a separate comment about why I don't generally support this approach.

I was mildly afraid of increasing compilation times and maybe even bloating the object code if the linker isn't paying enough attention (as discussed, for example, [here](https://docs.microsoft.com/en-us/cpp/cpp/source-code-organization-cpp-templates?view=vs-2019).  But I do agree with you that placing the definitions in the header file is stylistically superior, so I'm fine with doing so.

> To me this is a compelling argument for changing to free functions. But if you aren't convinced there are some other options:
> - You can at least make the functions private. As noted in the previous paragraph, they present a foot-gun so they shouldn't be public (it might be safe to bypass the check in your case, but it probably isn't in general).
>- You can fix the implementations and move the definitions into the header.

I am convinced, actually; I think that is a good argument for free rather than member functions.  But I would still prefer them public rather than private, because I did intend them as generic, and, as you show, making them free functions would remove the foot-gun, since it would force them to get at the session through ```Database::getSession()```.  So if they become free functions, then I think that addresses the problems with making them private (at least the one you raised in this conversation).

So how about I take your free-functions suggestion _and_ your one alternative suggestion (making it an adjunct rather than an alternative!) to move the definitions into the header?  As far as I can tell, free functions in the header would satisfy all your arguments in this conversation, as well as my argument that, as generic functions, they should be available to any client of Database, not only to members of Database itself.

That's what I've done in the latest code; please let me know whether that all made sense.

src/database/Database.cpp
rokopt: Done.

src/main/ApplicationImpl.h
rokopt: The idea was that Application would be the one class which dictated where its clients could inject code.  If ```actBeforeDBSchemaUpgrade()``` were to move to the Database class, how are you envisioning clients creating Application instances that injected code into Database methods?

src/database/Database.cpp
rokopt: Okay, I've made them into free functions.

I wasn't sure about putting the first overload in Database either, but I also wasn't sure where _else_ to put it.  Neither Decoder nor XDR knows about the other, and I figured they probably shouldn't, so that the new utility belonged in some third place.  And Database seemed like the place that already knew about decoding _and_ XDR, and indeed it was precisely the *SQL.cpp modules that needed decodeOpaqueXDR(), which I thought supported the idea of the functions belonging there.  But I'm open to suggestions for other places to move one or both of them to.

src/database/test/DatabaseTests.cpp
rokopt: Fixed.

src/database/test/DatabaseTests.cpp
rokopt: Fixed.

src/database/test/DatabaseTests.cpp
rokopt: Oh!  The comparisons that you're suggesting work; I'm just not used to the types, and it simply hadn't occurred to me that ```ae_upgraded``` (now ```aeUpgraded```) and ```entry``` were comparable.  I've made that change.

rokopt:
I'm attaching a couple "metrics" collections from running for a while on each of v12 and v13.  This is my first time trying to interpret our metrics, but after looking over the database.* and ledger.* ones that appear to be measuring the database operations that involve the tables that I've changed, I'm not seeing anything that looks like a statistically
significant regression.

[metrics-v12-long-run.txt](https://github.com/stellar/stellar-core/files/4833034/metrics-v12-long-run.txt)
[metrics-v13-long-run.txt](https://github.com/stellar/stellar-core/files/4833033/metrics-v13-long-run.txt)

src/database/Database.cpp
jonjove: I think it's fine for now. Right now we are only using it for the database. Maybe at a later time we'll have more clarity about it.

src/main/ApplicationImpl.h
jonjove: I would add a virtual function `ApplicationImpl::createDatabase` (see `createInvariantManager`, `createOverlayManager`, and `createOverlayManager` for reference) and overload this in the derived class. Then you could use basically the same injection mechanism as you have here, where the derived `Application` takes a functor which passes it to the derived `Database` upon construction, and the functor is called before schema upgrade.

I admit it's a bit less direct. It has the disadvantage of requiring a bit more boiler-plate. But it has the advantage of having better separation of concerns and being more composable. I'm not sure how much it matters here, but I figured we should at least discuss it.

rokopt: Aha -- I had briefly mulled over trying to work out how to do that, but I was afraid it would come across as too heavyweight, and didn't get as far as noticing that we had precedent!

Though the result is more verbose, "having better separation of concerns and being more composable" is music to my ears, so I've taken your suggestion.  Does the result look sane?

rokopt:
Jon pointed out that, although we mostly expect PostgreSQL to be our production-use database, we're not sure that no one's using SQLite in production, and the latter behaves differently during the upgrade because it doesn't let us drop columns, so we NULL out that field of all rows instead, which is a much larger operation than anything we do in the PostgreSQL upgrade.  Therefore, I'd better test the SQLite upgrade on the real pubnet database too.

I've now done so, and on owlbear, the SQLite schema upgrade took 2m13s.  I'd happened to guess based on other measurements that it would be somewhere around 2m, and asked Nicolas whether it would be alright if it were, and he said that that would be fine, so I am venturing to hope that 2m13s is fine too.

rokopt:
Jon pointed out that catchup performance would provide a strongly apples-to-apples comparison between v12 and v13, since it would involve applying exactly the same buckets; he suggested the command `stellar-core catchup 30000000/10000`.  I tried that on owlbear using PostgreSQL, and v13 was ~20s slower out of a ~65m test (0.5%).  I haven't repeated the test to try to get an idea of the variance yet, but even if that's significant, I imagine it's probably worthwhile for the future flexibility.

[catchup-perf-v12.txt](https://github.com/stellar/stellar-core/files/4834629/catchup-perf-v12.txt)
[catchup-perf-v13.txt](https://github.com/stellar/stellar-core/files/4834628/catchup-perf-v13.txt)

rokopt:
Commits have been re-squashed.  ...Non-destructively, I hope.

src/database/Database.cpp
```diff
@@ -263,10 +262,10 @@ Database::applySchemaUpgrade(unsigned long vers)
         {
             // Add columns for the LedgerEntry extension to each of
             // the tables that stores a type of ledger entry.
-            addTextColumn("accounts", ledgerExtName);
-            addTextColumn("trustlines", ledgerExtName);
-            addTextColumn("accountdata", ledgerExtName);
-            addTextColumn("offers", ledgerExtName);
+            addTextColumn("accounts", "ledgerext");
```
MonsieurNicolas, line 265: it looks like this commit also contains changes for the previous commit (cleanup?)

src/database/Database.cpp
```diff
@@ -290,6 +323,98 @@ Database::upgradeToCurrentSchema()
...
+    try
+    {
+        addTextColumn("accounts", "extension");
+        copyIndividualAccountExtensionFieldsToOpaqueXDR();
+        dropTextColumn("accounts", "buyingliabilities");
+        dropTextColumn("accounts", "sellingliabilities");
+    }
+    catch (std::exception& e)
+    {
+        CLOG(FATAL, "Database") << __func__ << ": exception " << e.what()
```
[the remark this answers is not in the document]
MonsieurNicolas: nm: I see that you cleaned this up in a later commit and forgot to squash relevant changes

src/database/Database.h
```diff
@@ -285,4 +248,89 @@ class DBTimeExcluder : NonCopyable
...
     ~DBTimeExcluder();
 };
+
+// Select a set of records using a client-defined query string, then map
+// each record into an element of a client-defined datatype by applying a
+// client-defined function (the records are accumulated in the "out"
+// vector).
+template <typename T>
+void
+selectMap(Database& db, std::string const& selectStr,
```
MonsieurNicolas, line 272: many of those changes should be squashed into the commit that introduced the change in the first place - this causes churn, making it harder to check what is going on

src/ledger/LedgerTxnDataSQL.cpp
```diff
@@ -102,19 +119,25 @@ class BulkUpsertDataOperation : public DatabaseTypeSpecificOperation<void>
...
+            "INSERT INTO accountdata ( "
+            "accountid, dataname, datavalue, lastmodified, extension, "
+            "ledgerext "
+            ") VALUES ( "
+            ":id, :v1, :v2, :v3, :v4, :v5 "
+            ") ON CONFLICT (accountid, dataname) DO UPDATE SET "
+            "datavalue = excluded.datavalue, "
+            "lastmodified = excluded.lastmodified, "
+            "extension = excluded.extension, ledgerext "
+            " = excluded.ledgerext";
```
MonsieurNicolas, line 132: not sure if this is a clang-format thing, but it would be nicer to have `ledgerext = excluded.ledgerext` on its own line

src/ledger/LedgerTxnDataSQL.cpp
```diff
@@ -136,31 +159,41 @@ class BulkUpsertDataOperation : public DatabaseTypeSpecificOperation<void>
...
+            "unnest(:v5::TEXT[]) "
+            ")"
+            "INSERT INTO accountdata ( "
+            "accountid, dataname, datavalue, lastmodified, extension, "
+            "ledgerext "
+            ") SELECT * FROM r "
+            "ON CONFLICT (accountid, dataname) DO UPDATE SET "
+            "datavalue = excluded.datavalue, "
+            "lastmodified = excluded.lastmodified, "
+            "extension = excluded.extension, ledgerext "
```
MonsieurNicolas, line 187: same here `ledgerext = excluded.ledgerext`

src/ledger/LedgerTxnOfferSQL.cpp
```diff
@@ -473,33 +514,41 @@ class BulkUpsertOffersOperation : public DatabaseTypeSpecificOperation<void>
...
+            "sellerid = excluded.sellerid, "
+            "sellingasset = excluded.sellingasset, "
+            "buyingasset = excluded.buyingasset, "
+            "amount = excluded.amount, "
+            "pricen = excluded.pricen, "
+            "priced = excluded.priced, "
+            "price = excluded.price, "
+            "flags = excluded.flags, "
+            "lastmodified = excluded.lastmodified, "
+            "extension = excluded.extension, ledgerext "
```
MonsieurNicolas, line 550: ` ledgerext = excluded.ledgerext`

src/main/Application.h
```diff
@@ -278,8 +278,10 @@ class Application
 
     virtual AbstractLedgerTxnParent& getLedgerTxnRoot() = 0;
 
-    // For tests to perform operations on databases while they're
-    // still using old schemas.
+    // Give clients the opportunity to perform operations on databases while
```
MonsieurNicolas: wrong commit

src/database/Database.h
```diff
@@ -214,6 +218,16 @@ class Database : NonMovableOrCopyable
     // Access the optional SOCI connection pool available for worker
     // threads. Throws an error if !canUsePool().
     soci::connection_pool& getPool();
+
+  protected:
+    // Give clients the opportunity to perform operations on databases while
```
MonsieurNicolas, line 223: the application test hook that you removed here should just be removed from history as you never used it (it's just extra commits and churn for no reason)

docs/db-schema.md
```diff
@@ -51,8 +51,8 @@ homedomain | VARCHAR(44) | (BASE64)
 thresholds | TEXT | (BASE64)
 flags | INT NOT NULL |
 lastmodified | INT NOT NULL | lastModifiedLedgerSeq
-buyingliabilities | BIGINT CHECK (buyingliabilities >= 0) |
-sellingliabilities | BIGINT CHECK (sellingliabilities >= 0) |
+extension | TEXT | Extension specific to AccountEntry (XDR)
```
MonsieurNicolas, line 54: squash this commit into the commit with the schema change

rokopt:
> I didn't find actual bugs, good job.
> 
> Mostly cosmetic issues from what I see.
> 
> In particular, the commit history is quite hard to follow as it's only partially squashed.
> 
> Given the changes, I think it's probably easier to squash everything and resplit instead of propagating a subset of changes across commits (but you can try to do this as an exercise if you've never done it).

Yeah, I wasn't sure what I should do about later commits that affected multiple previous ones.  I wondered whether I should try to squash them all into one and then split them up again.  Should I do that before merge?

src/database/Database.cpp
```diff
@@ -252,6 +258,33 @@ Database::applySchemaUpgrade(unsigned long vers)
         // the accountbalances index around.
         mSession << "DROP INDEX IF EXISTS accountbalances";
         break;
+    case 13:
```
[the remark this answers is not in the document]
rokopt: Done; see issue #2601 .

src/database/Database.cpp
```diff
@@ -9,10 +9,13 @@
 #include "main/Application.h"
 #include "main/Config.h"
 #include "overlay/StellarXDR.h"
+#include "util/Decoder.h"
```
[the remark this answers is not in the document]
rokopt: Okay, I'll keep that in mind when I squash and re-split.  Verbosity is a habit it might take me some time to break, though.  ;)

src/database/Database.cpp
```diff
@@ -252,6 +258,33 @@ Database::applySchemaUpgrade(unsigned long vers)
         // the accountbalances index around.
         mSession << "DROP INDEX IF EXISTS accountbalances";
         break;
+    case 13:
+        if (!mApp.getConfig().MODE_USES_IN_MEMORY_LEDGER)
+        {
+            // Add columns for the LedgerEntry extension to each of
+            // the tables that stores a type of ledger entry.
+            addTextColumn("accounts", ledgerExtName);
```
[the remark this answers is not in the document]
rokopt: I think all the changes got reflected in my later commit to `db-schema.md`, right?

src/database/Database.cpp
```diff
@@ -290,6 +323,98 @@ Database::upgradeToCurrentSchema()
     assert(vers == SCHEMA_VERSION);
 }
 
+void
+Database::addTextColumn(std::string const& table, std::string const& column)
+{
+    std::string addColumnStr("ALTER TABLE " + table + " ADD " + column +
+                             " TEXT;");
+    CLOG(INFO, "Database") << "Adding column with string '" << addColumnStr
```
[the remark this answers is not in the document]
rokopt: Okay.  I did it that way because I occasionally wrote syntax errors and wanted to see exactly what I'd done wrong (the `soci` exceptions generally don't say).  But it makes sense that for a final version, operators wouldn't care.

rokopt:
> Given the changes, I think it's probably easier to squash everything and resplit instead of propagating a subset of changes across commits (but you can try to do this as an exercise if you've never done it).

Done; commits have been squashed and re-split.

src/database/Database.cpp
```diff
@@ -290,6 +290,52 @@ Database::upgradeToCurrentSchema()
...
+
+template <typename T>
+void
+Database::updateMap(std::vector<T> const& in, std::string const& updateStr,
+                    std::function<void(soci::statement&, T const&)> prepUpdate,
+                    std::function<void(long long const, T const&)> postUpdate)
+{
+    auto st_update = getPreparedStatement(updateStr).statement();
+
+    for (auto recT : in)
```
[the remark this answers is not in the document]
rokopt: Done.

src/database/Database.cpp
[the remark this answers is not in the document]
rokopt: ...I didn't actually shorten them much this time.  Sorry about that.

src/database/Database.cpp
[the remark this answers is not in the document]
rokopt: I've updated the log messages in that file to try to constrain them to "what" and not "how".

src/database/Database.cpp
```diff
@@ -290,6 +323,98 @@ Database::upgradeToCurrentSchema()
...
+            << "Dropping column '" << column << "' with string '"
+            << dropColumnStr << "' from table '" << table << "'";
+
+        mSession << dropColumnStr;
+    }
+    else
+    {
+        std::string nullColumnStr("UPDATE " + table + " SET " + column +
+                                  " = NULL");
+        CLOG(INFO, "Database") << "SQLite does not support dropping column '"
```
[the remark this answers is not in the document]
rokopt: Done.

src/database/Database.cpp
```diff
@@ -290,6 +323,98 @@ Database::upgradeToCurrentSchema()
...
+{
+    std::string addColumnStr("ALTER TABLE " + table + " ADD " + column +
+                             " TEXT;");
+    CLOG(INFO, "Database") << "Adding column with string '" << addColumnStr
+                           << "'";
+    mSession << addColumnStr;
+}
+
+void
+Database::dropTextColumn(std::string const& table, std::string const& column)
```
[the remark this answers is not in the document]
rokopt: Changed as you suggest.

src/database/Database.cpp
```diff
@@ -290,6 +323,98 @@ Database::upgradeToCurrentSchema()
...
+
+        mSession << nullColumnStr;
+    }
+}
+
+std::string
+Database::getOldLiabilitySelect(std::string const& table,
+                                std::string const& fields)
+{
+    return "SELECT " + fields + "," +
```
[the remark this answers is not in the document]
rokopt: Done.

src/database/Database.cpp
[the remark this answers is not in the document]
rokopt: Yes.  Commits have been re-squashed and re-split now.

src/database/Database.cpp
```diff
@@ -290,6 +323,98 @@ Database::upgradeToCurrentSchema()
...
+    try
+    {
+        addTextColumn("trustlines", "extension");
+        copyIndividualTrustLineExtensionFieldsToOpaqueXDR();
+        dropTextColumn("trustlines", "buyingliabilities");
+        dropTextColumn("trustlines", "sellingliabilities");
+    }
+    catch (std::exception& e)
+    {
+        CLOG(FATAL, "Database") << __func__ << ": exception " << e.what()
```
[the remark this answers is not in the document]
rokopt: Same answer.

src/database/Database.cpp
```diff
@@ -336,6 +461,109 @@ Database::selectUpdateMap(
...
+void
+Database::copyIndividualAccountExtensionFieldsToOpaqueXDR()
+{
+    std::string const tableStr = "accounts";
+
+    CLOG(INFO, "Database") << __func__ << ": updating extension schema for "
+                           << tableStr;
+
+    // <accountID, extension>
+    using T = std::tuple<std::string, std::string>;
```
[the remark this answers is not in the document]
rokopt: Good points.  I've changed them to structures with meaningful names.

src/database/Database.cpp
rokopt: Yep, commits have been cleaned up.

src/database/Database.h
rokopt: Done.

src/ledger/LedgerTxnDataSQL.cpp
rokopt: Changed.  clang-format didn't disturb the change.

src/ledger/LedgerTxnDataSQL.cpp
rokopt: Changed.

src/ledger/LedgerTxnOfferSQL.cpp
rokopt: Changed.

src/main/Application.h
rokopt: Fixed, I think.

src/database/Database.h
rokopt: Done.

docs/db-schema.md
rokopt: Done.

MonsieurNicolas:
r+ 5801f31f822d74c9cd2c0af623e3ed52c85d5cea

PR #2605, opened 2020-07-01
Remove gHasher and use stack-local incremental SHA256 class.

### LINKED ISSUE #2604: Why isn't gHasher in src/crypto/SecretKey.cpp protected by a mutex?
### Explain in detail the incorrect or additional documentation you'd like to see.

I went to look at how Stellar Core does Ed25519 validation and came to this function:

https://github.com/stellar/stellar-core/blob/fdfbd76382df602db491288b928896ef3e27c5ef/src/crypto/SecretKey.cpp#L313-L346

If I understand correctly, it's maintaining a global cache of signature verification results here:

https://github.com/stellar/stellar-core/blob/fdfbd76382df602db491288b928896ef3e27c5ef/src/crypto/SecretKey.cpp#L36-L40

I don't really understand why a cache helps unless the code is repeatedly verifying the same signature, but I assume this is because of my lack of knowledge about Stellar's architecture.

What I'm concerned about is the potential for race conditions on the hasher.  The `gVerifySigCacheMutex` is used to control access to the cache.  The cache key is computed as
```
auto cacheKey = verifySigCacheKey(key, signature, bin);
```
before the cache lookup, which is in a sub-scope protected by a mutex guard:
```
    {
        std::lock_guard<std::mutex> guard(gVerifySigCacheMutex);
        if (gVerifySigCache.exists(cacheKey)) { /* ... */ }
    }
```
However, `verifySigCacheKey` uses `gHasher`, defined as `static std::unique_ptr<SHA256> gHasher = SHA256::create();`, using an IUF API that mutates the `gHasher` state.

What protects the `gHasher` from concurrent mutation? It appears that concurrent access is possible (otherwise, there would be no need to protect the cache with a mutex), but nothing seems to prevent concurrent calls to `verifySigCacheKey` from simultaneously writing the same hasher state, causing corruption in the hash calculation.

graydon:
Resolves #2604 -- removes gHasher and switches to using a local (per-call) instance of the SHA256 class, which itself is made stack-allocatable.

gHasher wasn't accessed by multiple threads as far as I can tell already but removing it entirely eliminates the problem better than adding another mutex.

### DIFF
--- a/src/bucket/BucketList.cpp
+++ b/src/bucket/BucketList.cpp
@@ -30,10 +30,10 @@ BucketLevel::BucketLevel(uint32_t i)
 uint256
 BucketLevel::getHash() const
 {
-    auto hsh = SHA256::create();
-    hsh->add(mCurr->getHash());
-    hsh->add(mSnap->getHash());
-    return hsh->finish();
+    SHA256 hsh;
+    hsh.add(mCurr->getHash());
+    hsh.add(mSnap->getHash());
+    return hsh.finish();
 }
 
 FutureBucket const&
@@ -367,12 +367,12 @@ uint256
 BucketList::getHash() const
 {
     ZoneScoped;
-    auto hsh = SHA256::create();
+    SHA256 hsh;
     for (auto const& lev : mLevels)
     {
-        hsh->add(lev.getHash());
+        hsh.add(lev.getHash());
     }
-    return hsh->finish();
+    return hsh.finish();
 }
 
 // levelShouldSpill is the set of boundaries at which each level should spill,
--- a/src/bucket/BucketOutputIterator.cpp
+++ b/src/bucket/BucketOutputIterator.cpp
@@ -42,7 +42,6 @@ BucketOutputIterator::BucketOutputIterator(std::string const& tmpDir,
     : mFilename(randomBucketName(tmpDir))
     , mOut(ctx, doFsync)
     , mBuf(nullptr)
-    , mHasher(SHA256::create())
     , mKeepDeadEntries(keepDeadEntries)
     , mMeta(meta)
     , mMergeCounters(mc)
@@ -96,7 +95,7 @@ BucketOutputIterator::put(BucketEntry const& e)
         if (mCmp(*mBuf, e))
         {
             ++mMergeCounters.mOutputIteratorActualWrites;
-            mOut.writeOne(*mBuf, mHasher.get(), &mBytesPut);
+            mOut.writeOne(*mBuf, &mHasher, &mBytesPut);
             mObjectsPut++;
         }
     }
@@ -117,7 +116,7 @@ BucketOutputIterator::getBucket(BucketManager& bucketManager,
     ZoneScoped;
     if (mBuf)
     {
-        mOut.writeOne(*mBuf, mHasher.get(), &mBytesPut);
+        mOut.writeOne(*mBuf, &mHasher, &mBytesPut);
         mObjectsPut++;
         mBuf.reset();
     }
@@ -135,7 +134,7 @@ BucketOutputIterator::getBucket(BucketManager& bucketManager,
         }
         return std::make_shared<Bucket>();
     }
-    return bucketManager.adoptFileAsBucket(mFilename, mHasher->finish(),
+    return bucketManager.adoptFileAsBucket(mFilename, mHasher.finish(),
                                            mObjectsPut, mBytesPut, mergeKey);
 }
 }
--- a/src/bucket/BucketOutputIterator.h
+++ b/src/bucket/BucketOutputIterator.h
@@ -27,7 +27,7 @@ class BucketOutputIterator
     XDROutputFileStream mOut;
     BucketEntryIdCmp mCmp;
     std::unique_ptr<BucketEntry> mBuf;
-    std::unique_ptr<SHA256> mHasher;
+    SHA256 mHasher;
     size_t mBytesPut{0};
     size_t mObjectsPut{0};
     bool mKeepDeadEntries{true};
--- a/src/crypto/SHA.cpp
+++ b/src/crypto/SHA.cpp
@@ -25,31 +25,13 @@ sha256(ByteSlice const& bin)
     return out;
 }
 
-class SHA256Impl : public SHA256, NonCopyable
-{
-    crypto_hash_sha256_state mState;
-    bool mFinished;
-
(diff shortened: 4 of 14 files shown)


jonjove:
This change looks right. But it's not obvious to me that `gHasher` should be global in the first place. I generally take the stance that anything global bears the burden of justifying why it should be global.

The constructor for `SHA256Impl` (which in my opinion is not a great name because this is not using the `pImpl` idiom) just calls `reset`... which is exactly what we are doing explicitly in `verifySigCacheKey` anyway. Other than that, constructing `SHA256Impl` means allocating `crypto_hash_sha_256_state` which is approximately 100 bytes. It's possible this makes a tangible performance difference, but I'd want evidence of that.

graydon:
Well, here's `perf stat -r 3 stellar-core test 'sign and verify benchmarking'`:

~~~
with SHA1 cache:                                                                                                                                                                
                                                                                                                                                                                
Performance counter stats for './src/stellar-core test sign and verify benchmarking' (3 runs):                                                                                  
                                                                                                                                                                                
      14099.986837      task-clock (msec)         #    0.987 CPUs utilized            ( +-  2.52% )                                                                             
             2,145      context-switches          #    0.152 K/sec                    ( +-  2.39% )                                                                             
                 9      cpu-migrations            #    0.001 K/sec                    ( +-  6.42% )                                                                             
            63,424      page-faults               #    0.004 M/sec                    ( +-  0.22% )                                                                             
    47,165,524,808      cycles                    #    3.345 GHz                      ( +-  0.11% )                                                                             
     8,123,588,237      stalled-cycles-frontend   #   17.22% frontend cycles idle     ( +-  0.98% )                                                                             
   115,641,971,821      instructions              #    2.45  insn per cycle                                                                                                     
                                                  #    0.07  stalled cycles per insn  ( +-  0.00% )                                                                             
     1,458,902,624      branches                  #  103.468 M/sec                    ( +-  0.00% )                                                                             
        39,142,008      branch-misses             #    2.68% of all branches          ( +-  0.12% )                                                                             
                                                                                                                                                                                
      14.287589810 seconds time elapsed                                          ( +-  2.49% )                                                                                  
                                                                                                                                                                                
without SHA1 cache:                                                                                                                                                             
                                                                                                                                                                                
 Performance counter stats for './src/stellar-core test sign and verify benchmarking' (3 runs):                                                                                 
                                                                                                                                                                                
      14525.457126      task-clock (msec)         #    0.987 CPUs utilized            ( +-  3.05% )                                                                             
             2,225      context-switches          #    0.153 K/sec                    ( +-  3.13% )                                                                             
                12      cpu-migrations            #    0.001 K/sec                    ( +- 20.00% )                                                                             
            60,371      page-faults               #    0.004 M/sec                    ( +-  2.83% )                                                                             
    47,114,685,352      cycles                    #    3.244 GHz                      ( +-  0.20% )                                                                             
     8,028,694,902      stalled-cycles-frontend   #   17.04% frontend cycles idle     ( +-  1.01% )                                                                             
   115,651,189,437      instructions              #    2.45  insn per cycle                                                                                                     
                                                  #    0.07  stalled cycles per insn  ( +-  0.00% )                                                                             
     1,461,180,220      branches                  #  100.594 M/sec                    ( +-  0.02% )                                                                             
        39,244,894      branch-misses             #    2.69% of all branches          ( +-  0.12% )                                                                             
                                                                                                                                                                                
      14.720780018 seconds time elapsed                                          ( +-  3.00% )                                                                                  
~~~

Unfortunately that's a benchmark that's 100% misses so it's going to bury the difference of the cache in the cost of the verifies. I will work up a better test.

MonsieurNicolas:
+1 on removing this global.

I would argue that we should ditch that SHA256 Impl class entirely: we should collapse those classes, make SHA256 the actual implementation that way: no virtual calls when hashing and the class can be stack allocated.

graydon:
Ok here's a better benchmark that's composed entirely of verify-hits with a before/after comparison with "remove the entire shared global and use a stack local", removing the indirection to a private class and not worrying about a little bleed-through of a small set of definitions from sodium headers (I checked and it's quite small):

~~~
shared global:

 Performance counter stats for './src/stellar-core test verify-hit benchmarking' (30 runs):

       2700.300198      task-clock (msec)         #    0.923 CPUs utilized            ( +-  1.21% )
               467      context-switches          #    0.173 K/sec                    ( +-  3.02% )
                 2      cpu-migrations            #    0.001 K/sec                    ( +- 12.80% )
             2,183      page-faults               #    0.808 K/sec                    ( +-  0.02% )
     7,844,273,314      cycles                    #    2.905 GHz                      ( +-  0.06% )
       567,236,379      stalled-cycles-frontend   #    7.23% frontend cycles idle     ( +-  0.96% )
    23,785,821,606      instructions              #    3.03  insn per cycle         
                                                  #    0.02  stalled cycles per insn  ( +-  0.00% )
       418,395,890      branches                  #  154.944 M/sec                    ( +-  0.02% )
           698,821      branch-misses             #    0.17% of all branches          ( +- 14.44% )

       2.925243781 seconds time elapsed                                          ( +-  1.13% )

stack local:

 Performance counter stats for './src/stellar-core test verify-hit benchmarking' (30 runs):

       2283.762593      task-clock (msec)         #    0.909 CPUs utilized            ( +-  0.22% )
               409      context-switches          #    0.179 K/sec                    ( +-  3.97% )
                 2      cpu-migrations            #    0.001 K/sec                    ( +- 12.97% )
             2,182      page-faults               #    0.956 K/sec                    ( +-  0.02% )
     7,788,646,217      cycles                    #    3.410 GHz                      ( +-  0.08% )
       556,192,231      stalled-cycles-frontend   #    7.14% frontend cycles idle     ( +-  1.20% )
    23,704,973,281      instructions              #    3.04  insn per cycle         
                                                  #    0.02  stalled cycles per insn  ( +-  0.01% )
       400,531,120      branches                  #  175.382 M/sec                    ( +-  0.05% )
           636,040      branch-misses             #    0.16% of all branches          ( +- 13.01% )

       2.511696622 seconds time elapsed                                          ( +-  0.23% )
~~~

In other words it's a perf win to do so, and it's simpler. Everyone should be happy!

MonsieurNicolas:
This looks great. Squash this up and we'll merge it. @graydon

MonsieurNicolas:
r+ 20b490c300f08a0dff448bd0d8d0e176054fa115

PR #2610, opened 2020-07-07
Trigger next ledger when node gets back in sync

### LINKED ISSUE #2527: Trigger next ledger as soon as node goes back in sync
See discussion here: https://github.com/stellar/stellar-core/pull/2512#discussion_r423399774

Right now when node goes back in sync, it waits until the next ledger from the network to trigger a new ledger. We should change this behavior to allow triggering the next ledger as soon as the last buffered ledger is applied.

marta-lokhova:
Resolves #2527

### DIFF
--- a/src/herder/HerderImpl.cpp
+++ b/src/herder/HerderImpl.cpp
@@ -261,7 +261,17 @@ HerderImpl::valueExternalized(uint64 slotIndex, StellarValue const& value)
     }
     else
     {
+        // This may trigger getting back in sync (buffered ledgers
+        // application)
         processExternalized(slotIndex, value);
+
+        // Any ledgers processed by Herder must have been buffered in LM.
+        // If LM applied them all, Herder and LM must now be consistent with
+        // each other (i.e., track the same ledger)
+        if (mLedgerManager.isSynced())
+        {
+            maybeTriggerNextLedger(false);
+        }
     }
 }
 
@@ -693,26 +703,15 @@ HerderImpl::getTransactionQueue()
 #endif
 
 void
-HerderImpl::ledgerClosed(bool synchronous)
+HerderImpl::maybeTriggerNextLedger(bool synchronous)
 {
-    // this method is triggered every time the most recent ledger is
-    // externalized it performs some cleanup and also decides if it needs to
-    // schedule triggering the next ledger
-
-    ZoneScoped;
     mTriggerTimer.cancel();
 
-    CLOG(TRACE, "Herder") << "HerderImpl::ledgerClosed";
-
-    auto lastIndex = mHerderSCPDriver.lastConsensusLedgerIndex();
     uint64_t nextIndex = mHerderSCPDriver.nextConsensusLedgerIndex();
-
-    mPendingEnvelopes.slotClosed(lastIndex);
-
-    mApp.getOverlayManager().ledgerClosed(lastIndex);
-
     if (mLedgerManager.isSynced())
     {
+        auto lastIndex = mHerderSCPDriver.lastConsensusLedgerIndex();
+
         // if we're in sync, we setup mTriggerTimer
         // it may get cancelled if a more recent ledger externalizes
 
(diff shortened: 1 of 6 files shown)


src/herder/HerderImpl.cpp
```diff
@@ -764,6 +763,27 @@ HerderImpl::ledgerClosed(bool synchronous)
     }
 }
 
+void
+HerderImpl::ledgerClosed(bool synchronous)
```
MonsieurNicolas, line 767: nit: (for future) - the diff would have made more sense if you defined `ledgerClosed` before `maybeTriggerNextLedger`

src/herder/test/HerderTests.cpp
```diff
@@ -1620,100 +1620,126 @@ TEST_CASE("SCP State", "[herder][acceptance]")
...
+    REQUIRE(currentCLedger() == currentLedger);
 
-    // disconnect listener
-    simulation->dropConnection(validatorKey.getPublicKey(),
-                               listenerKey.getPublicKey());
+    // disconnect C
+    simulation->dropConnection(validatorAKey.getPublicKey(),
+                               validatorCKey.getPublicKey());
+
+    HerderImpl& AHerder = *static_cast<HerderImpl*>(&A->getHerder());
```
MonsieurNicolas, line 1694: names should be things like `herderA` (this also avoids not following our coding standard for variable names)

src/herder/test/HerderTests.cpp
```diff
@@ -1620,100 +1620,126 @@ TEST_CASE("SCP State", "[herder][acceptance]")
...
-    auto const& cm = listener->getCatchupManager();
+    auto const& lm = C->getLedgerManager();
+    auto const& cm = C->getCatchupManager();
 
     // Now construct a few future externalize messages
     // Make sure out of order messages are still within the validity range
     std::map<uint32_t, std::pair<SCPEnvelope, TxSetFramePtr>>
-        validatorSCPMessages;
-    Hash qSetHash = sha256(xdr::xdr_to_opaque(qset));
+        AValidatorSCPMessages;
```
MonsieurNicolas, line 1704: same thing here wrt names

src/herder/test/HerderTests.cpp
```diff
@@ -1620,100 +1620,126 @@ TEST_CASE("SCP State", "[herder][acceptance]")
...
+
+    HerderImpl& AHerder = *static_cast<HerderImpl*>(&A->getHerder());
+    HerderImpl& BHerder = *static_cast<HerderImpl*>(&B->getHerder());
+    HerderImpl& CHerder = *static_cast<HerderImpl*>(&C->getHerder());
 
-    HerderImpl& herder = *static_cast<HerderImpl*>(&listener->getHerder());
-    HerderImpl& valHerder = *static_cast<HerderImpl*>(&validator->getHerder());
-    auto const& lm = listener->getLedgerManager();
-    auto const& cm = listener->getCatchupManager();
+    auto const& lm = C->getLedgerManager();
```
MonsieurNicolas, line 1698: maybe rename `lm`/`cm` into `lmC`/`cmC` ?

src/herder/test/HerderTests.cpp
```diff
@@ -1620,100 +1620,126 @@ TEST_CASE("SCP State", "[herder][acceptance]")
...
                 REQUIRE(txset);
-                validatorSCPMessages[start] = std::make_pair(env, txset);
+                BValidatorSCPMessages[start] = std::make_pair(env, txset);
             }
         }
     }
 
-    REQUIRE(herder.getCurrentLedgerSeq() == currentListenerLedger());
+    REQUIRE(AValidatorSCPMessages.size() == BValidatorSCPMessages.size());
+    REQUIRE(CHerder.getCurrentLedgerSeq() == currentCLedger());
```
MonsieurNicolas, line 1742: also need to re-check that ` REQUIRE(currentCLedger() == currentLedger);` in case there was a bug in the test (and C was still connected)

src/herder/test/HerderTests.cpp
```diff
@@ -1800,6 +1848,25 @@ TEST_CASE("values externalized out of order", "[herder][quickRestart]")
     {
         testOutOfOrder(/* partial */ true);
     }
+    SECTION("C goes back in sync and unsticks the network")
+    {
+        testOutOfOrder(/* partial */ false);
+
+        // Now that C is back in sync and triggered next ledger
```
MonsieurNicolas, line 1855: Do we need to reconnect here? I think there is a slight chance that by connecting `A` and `C`, `C` externalizes ledgers that `A` and `B` externalized together.

I think we should just wait a few seconds and we should see a `NOMINATION` message from `C` for that new ledger before connecting `A` to `C` (maybe add this to the end of `testOutOfOrder`).

marta-lokhova:
@MonsieurNicolas addressed your feedback (changes already squashed into the last commit)

MonsieurNicolas:
r+ e38eb66736f47049c966166edc688f0dcbaa911b

PR #2611, opened 2020-07-08
Fix compile error (non C++ 14 compliant code)

MonsieurNicolas:
Struct initialization is a C99 extension https://en.cppreference.com/w/c/language/struct_initialization

in C++ this was added in C++ 20 as designator initializers https://en.cppreference.com/w/cpp/language/aggregate_initialization

For some reason both clang and g++ both allow this to compile in standard compliance mode, and the only way to reveal this (as a warning) seems to build with `-Wc99-extensions` .


MonsieurNicolas:
r+ f41504d33e383abba316e33090e98bf773431ac6

PR #2612, opened 2020-07-08
Validating nodes default to force_scp=true unless specified otherwise

### LINKED ISSUE #2542: Node should start nominating if it sees that it's on the right ledger
Right now we decide to nominate on startup only based on `FORCE_SCP`.

But if the network is on the same ledger, the local node should be able to start nomination as well.

This would allow (alongside #2081 ) for a node to fully participate to consensus on restart.

The tricky question is deciding what "on the same ledger" means:
it's possible the network is stuck nominating the same slot, but it could also be that the network moved on to a different slot.

A possibility could be to deprecate `FORCE_SCP`, but this could cause other problems in catastrophic situations (ie bad network fork) - this needs to be analyzed (we added this early on before validating SCP). I suspect it's not needed anymore.

We could go with a less aggressive stance: `FORCE_SCP=false` can trigger if we don't see messages from our quorum set at a later slot (this criteria should take into account the possibility of dealing with byzantine nodes so it's really a small subset of nodes that could be ahead, like a v-blocking set).

marta-lokhova:
Resolves #2542

### DIFF
--- a/src/main/Config.cpp
+++ b/src/main/Config.cpp
@@ -992,6 +992,10 @@ Config::processConfig(std::shared_ptr<cpptoml::table> t)
                 throw std::invalid_argument(err);
             }
         }
+
+        // Validators default to starting the network from local state
+        FORCE_SCP = NODE_IS_VALIDATOR;
+
         // process elements that potentially depend on others
         if (t->contains("VALIDATORS"))
         {
--- a/src/main/Config.h
+++ b/src/main/Config.h
@@ -121,12 +121,12 @@ class Config : public std::enable_shared_from_this<Config>
     // application config
 
     // The default way stellar-core starts is to load the state from disk and
-    // catch up to the network before starting SCP. If you need different
-    // behavior you need to use new-db or force-scp which sets the following
-    // flags:
+    // start a consensus round (if node is validating), then maybe trigger
+    // catchup. If you need different behavior you need to use new-db or
+    // --wait-for-consensus option which sets the following flag to false:
 
     // SCP will start running immediately using the current local state to
-    // participate in consensus. DO NOT INCLUDE THIS IN A CONFIG FILE
+    // participate in consensus
     bool FORCE_SCP;
 
     // This is a mode for testing. It prevents you from trying to connect to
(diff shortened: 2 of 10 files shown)


src/main/Config.cpp
```diff
@@ -992,6 +992,10 @@ Config::processConfig(std::shared_ptr<cpptoml::table> t)
                 throw std::invalid_argument(err);
             }
         }
+
+        // Validators default to starting the network from local state
+        FORCE_SCP = NODE_IS_VALIDATOR;
```
MonsieurNicolas, line 997: grep for `force-scp`; it looks like you need a few updates:
* comments in config.h
* commands.md - (need to document instead the new option command)
* testnet.md - behavior described is wrong, also needs to be updated to describe when to use the new option

marta-lokhova: Good call, updated. Next thing I'm going to do is try it with several nodes like you suggested, and also see if this change breaks anything in the testing pipeline.

marta-lokhova:
Local tests I did with two different quorum setups (all nodes are validating):

* node1 <--- node2
  * Both start nominating (`FORCE_SCP=true`) values for the same slot. Both nodes close ledgers normally
  * Node1 boots first, starts nominating, and closes a few ledgers. Node2 boots next, attempts to vote for its current slot, but hears from Node1 and closes several ledgers as well. Both nodes proceed to close ledgers normally
  * Node1 boots and starts nominating, Node2 boots with `--wait-for-consensus` flag, the network closes ledger normally
  * Node1 boots with `--wait-for-consensus` flag, Node2 boots and starts nominating, the network is stuck

* node1 <---> node2
  * Both nodes boot, start nominating and close ledgers normally
  * Node1 boots with `--wait-for-consensus` flag, the network is stuck

@MonsieurNicolas  Are there more complex scenarios you'd like to try out?

src/main/Config.h
```diff
@@ -121,9 +121,9 @@ class Config : public std::enable_shared_from_this<Config>
     // application config
 
     // The default way stellar-core starts is to load the state from disk and
-    // catch up to the network before starting SCP. If you need different
-    // behavior you need to use new-db or force-scp which sets the following
-    // flags:
+    // start a consensus round (if node is validating), then maybe trigger
+    // catchup. If you need different behavior you need to use new-db or
+    // --wait-for-consensus option which sets the following flag:
```
MonsieurNicolas, line 126: comment was not properly updated. "wait-for-consensus" _clears_ the flag ; also here is a "all caps" that this should not be set in a config file - we should remove that as well

marta-lokhova: updated (squashed as well)

MonsieurNicolas:
r+ 1a6c0e2635f8d079dd2fc6a22ddfebd5c6495473

PR #2622, opened 2020-07-14
Do not wait for in-progress merges on shutdown

### LINKED ISSUE #2618: pending merges should be cancelled on shutdown
ran into this on v13.2.0

basically, I ran a a command line catchup that looked like it was getting stuck on shutdown. Turns out that it waits for a merge to complete. I don't think there is a reason for this?

repro steps (against prod):
 * setup - `stellar-core catchup 30389842/10`
 * close a few ledgers `stellar-core catchup 30389845/10`

actual:
Get stuck on `2020-07-10T12:14:18.385 GAAVD [default INFO] Application destructing`

expected:
stops as soon as possible (merge will be resumed next time it starts)

marta-lokhova:
Resolves #2618

### DIFF
--- a/src/bucket/Bucket.cpp
+++ b/src/bucket/Bucket.cpp
@@ -622,8 +622,24 @@ Bucket::merge(BucketManager& bucketManager, uint32_t maxProtocolVersion,
                              mc, ctx, doFsync);
 
     BucketEntryIdCmp cmp;
+    size_t iter = 0;
+
     while (oi || ni)
     {
+        // Check if the merge should be stopped every few entries
+        if (++iter >= 1000)
+        {
+            iter = 0;
+            if (bucketManager.isShutdown())
+            {
+                // Stop merging, as BucketManager is now shutdown
+                // This is safe as temp file has not been adopted yet,
+                // so it will be removed with the tmp dir
+                throw std::runtime_error(
+                    "Incomplete bucket merge due to BucketManager shutdown");
+            }
+        }
+
         if (!mergeCasesWithDefaultAcceptance(cmp, mc, oi, ni, out,
                                              shadowIterators, protocolVersion,
                                              keepShadowedLifecycleEntries))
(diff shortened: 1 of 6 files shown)


src/bucket/Bucket.cpp
```diff
@@ -622,8 +622,23 @@ Bucket::merge(BucketManager& bucketManager, uint32_t maxProtocolVersion,
...
+        // Check if the merge should be stopped every few entries
+        if (++iter >= 1000)
+        {
+            iter = 0;
+            if (bucketManager.isShutdown())
+            {
+                // Stop merging, as BucketManager is now shutdown
+                // This is safe as temp file has not been adopted yet,
+                // so it will be removed with the tmp dir
+                return nullptr;
```
graydon, line 638: I don't think this is generally a safe thing to do -- FutureBucket's contract is that its mOutputBucketFuture returns a non-null shared pointer. See for example FutureBucket.cpp:260, in the function FutureBucket::resolve() where we call getHash() on the returned value from mOutputBucketFuture.get().

And more generally FutureBucket::resolve() is supposed to also return a non-null shared_ptr. The whole lifecycle of FutureBucket assumes that it runs to completion, not that it can be interrupted and left in an invalid state, eg. BucketLevel::commit() calls setCurr(mNextCurr.resolve()) and so on.

Granted none of these things are _likely_ to happen if we're in the process of shutting down, but changing the contract this way feels .. kinda worrying to me.

Maybe we could approach it by throwing an exception in the merge function, in this shutdown condition? I _believe_ this will be caught by the std::packaged_task doing the merge and will re-propagate on access to its caller (eg. if anyone calls get() it'll throw). It might also be worth adding a release assert that the future is valid() in any access paths; it looks like we're not doing that currently and we should probably fail if it's ever not.

marta-lokhova: yeah, that's a good idea. I don't think right now we do any future bucket manipulations on shutdown, as the main io service is stopped first before we allow merges to finish, but it's probably better not to rely on that. I've updated the code to throw instead of returning nullptr.

Also, we're already checking future validity in `FutureBucket::checkState`, or are you referring to something else?

graydon: Ah, I somehow missed that checkState checks validity. Well and good!

graydon:
r+ 7148ea6

PR #2624, opened 2020-07-14
Set `feeCharged` with the min fee for core to accept a transaction

### LINKED ISSUE #2505: Add Fee-Recommendation to Transaction Submission Response

MonsieurNicolas:
# Description

Resolves #2505 

replaces #2521

This PR tweaks the behavior when checking validity such that `feeCharged` in the `TransactionResult` is always set with the minimum fee that core expects for that transaction to be accepted in the transaction queue and to pass validity in general.

Core was *almost* doing this, this PR makes it that it's doing it consistently for all instances of `txINSUFFICIENT_FEE`.

I decided to *not* touch the code at all in the "apply" case even though I *think* that with today's logic it's actually not possible to reach some of the conditions that are in the codebase. Given how critical fees are in the system, I figured this was the more conservative approach and that has less chance of breaking in the future.

### DIFF
--- a/lib/util/uint128_t.cpp
+++ b/lib/util/uint128_t.cpp
@@ -28,6 +28,7 @@ uint128_t::operator bool() const{
 uint128_t::operator char() const{
     return (char) LOWER;
 }
+
 uint128_t::operator int() const{
     return (int) LOWER;
 }
--- a/lib/util/uint128_t.h
+++ b/lib/util/uint128_t.h
@@ -28,6 +28,18 @@ a bug in operator*.
 
 Thanks to Fran�ois Dessenne for convincing me
 to do a general rewrite of this class.
+
+when defining UNSAFE_UINT128_OPS additional behaviors are enabled.
+
+for example:
+// implicit typecast Operators,
+// short for (short)uint128_t(y).lower()
+// or (char)uint128_t(y)
+char x = uint128_t(y);
+
+// arithmetic operators with short type as lhs,
+// short for 2 * uint128_t(y).lower()
+auto x = 2 * uint128_t(y);
 */
 
 #ifndef __UINT128_T__
@@ -38,6 +50,12 @@ to do a general rewrite of this class.
 #include <stdint.h>
 #include <utility>
 #include <string>
+
+#ifdef UNSAFE_UINT128_OPS
+#define IMPLICIT_UNSAFE_UINT128_OPS
+#else
+#define IMPLICIT_UNSAFE_UINT128_OPS explicit
+#endif
 
 class uint128_t{
     private:
@@ -70,13 +88,13 @@ class uint128_t{
         }
 
         // Typecast Operators
-        operator bool() const;
-        operator char() const;
-        operator int() const;
-        operator uint8_t() const;
-        operator uint16_t() const;
-        operator uint32_t() const;
-        operator uint64_t() const;
+        IMPLICIT_UNSAFE_UINT128_OPS operator bool() const;
+        IMPLICIT_UNSAFE_UINT128_OPS operator char() const;
+        IMPLICIT_UNSAFE_UINT128_OPS operator int() const;
+        IMPLICIT_UNSAFE_UINT128_OPS operator uint8_t() const;
+        IMPLICIT_UNSAFE_UINT128_OPS operator uint16_t() const;
+        IMPLICIT_UNSAFE_UINT128_OPS operator uint32_t() const;
+        IMPLICIT_UNSAFE_UINT128_OPS operator uint64_t() const;
 
         // Bitwise Operators
         uint128_t operator&(const uint128_t & rhs) const;
@@ -334,6 +352,7 @@ template <typename T> bool operator<=(const T & lhs, const uint128_t & rhs){
     return ((uint64_t) lhs <= rhs.lower());
 }
 
+#ifdef UNSAFE_UINT128_OPS
 // Arithmetic Operators
 template <typename T> T operator+(const T & lhs, const uint128_t & rhs){
     return (T) (rhs + lhs);
@@ -379,6 +398,7 @@ template <typename T> T & operator%=(T & lhs, const uint128_t & rhs){
     lhs = (T) (uint128_t(lhs) % rhs);
     return lhs;
 }
+#endif
 
 // IO Operator
 std::ostream & operator<<(std::ostream & stream, const uint128_t & rhs);
--- a/src/herder/TransactionQueue.cpp
+++ b/src/herder/TransactionQueue.cpp
@@ -46,8 +46,12 @@ TransactionQueue::TransactionQueue(Application& app, int pendingDepth,
     }
 }
 
+// returns true, if a transaction can be replaced by another
+// `minFee` is set when returning false, and is the smallest fee
+// that would allow replace by fee to succeed in this situation
 static bool
-canReplaceByFee(TransactionFrameBasePtr tx, TransactionFrameBasePtr oldTx)
+canReplaceByFee(TransactionFrameBasePtr tx, TransactionFrameBasePtr oldTx,
+                int64_t& minFee)
 {
     int64_t newFee = tx->getFeeBid();
     uint32_t newNumOps = std::max<uint32_t>(1, tx->getNumOperations());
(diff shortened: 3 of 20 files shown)


src/herder/TransactionQueue.cpp
```diff
@@ -62,7 +66,14 @@ canReplaceByFee(TransactionFrameBasePtr tx, TransactionFrameBasePtr oldTx)
     // by INT64_MAX, while number of operations and FEE_MULTIPLIER are small.
     auto v1 = bigMultiply(newFee, oldNumOps);
     auto v2 = bigMultiply(oldFee, newNumOps);
-    return v1 >= TransactionQueue::FEE_MULTIPLIER * v2;
+    auto minFeeN = v2 * TransactionQueue::FEE_MULTIPLIER;
+    bool res = v1 >= minFeeN;
+    if (!res)
+    {
+        // this will not throw
+        minFee = bigDivide(minFeeN, int64_t(oldNumOps), Rounding::ROUND_UP);
```
jonjove, line 74: I assume you are relying on the fact that the total amount of native asset is bounded at about 50B? I think with that constraint this can't overflow.

I think it would probably be clearer and safer to use the non-throwing variant and just explicitly handle capping the `minFee` at `INT64_MAX` if it would have overflowed. This way we don't have to re-check this math if we change the fee-multiplier, or the max number of operations, or anything else.

src/herder/test/TransactionQueueTests.cpp
```diff
@@ -1037,8 +1037,8 @@ TEST_CASE("replace by fee", "[herder][transactionqueue]")
         {
             for (uint32_t i = 1; i <= 3; ++i)
             {
-                test.add(transaction(*app, account1, i, 1, 199),
-                         TransactionQueue::AddResult::ADD_STATUS_ERROR);
+                auto tx = transaction(*app, account1, i, 1, 199);
```
jonjove, line 1040: Was this changed for style reasons, or is there something about the semantics that required this?

src/transactions/FeeBumpTransactionFrame.cpp
```diff
@@ -194,6 +194,8 @@ FeeBumpTransactionFrame::commonValidPreSeqNum(AbstractLedgerTxn& ltx)
     auto v2 = bigMultiply(mInnerTx->getFeeBid(), getMinFee(lh));
     if (v1 < v2)
     {
+        getResult().feeCharged =
```
jonjove, line 197: It is not at all obvious to me that this one can't overflow, even if you use the bound of approximately 50B native asset. The reason is that the `baseFee` could conceivably be `UINT32_MAX`, which makes `50B * UINT32_MAX * 100` the approximate worst case value of `v2`.

src/transactions/TransactionFrame.cpp
```diff
@@ -148,18 +148,26 @@ TransactionFrame::getMinFee(LedgerHeader const& header) const
...
-TransactionFrame::getFee(LedgerHeader const& header, int64_t baseFee) const
+TransactionFrame::getFee(LedgerHeader const& header, int64_t baseFee,
+                         bool applying) const
 {
-    if (header.ledgerVersion < 11)
-    {
-        return getFeeBid();
-    }
-    else
+    if (header.ledgerVersion >= 11 || !applying)
```
jonjove, line 154: I'm having a hard time understanding this. I _think_ this works even before version 11 because in that case `baseFee = lh.baseFee`. I also _think_ this works correctly for fee-bump inner transactions which will have `baseFee = 0`. It's pretty hard to see on preliminary inspection though. Can you add some commentary explaining why this is right?

I'm also not sure why this matters, given that no fee recommendations from before version 11 can actually be accessed (I suppose this might happen on the testnet during a reset?).

lib/util/uint128_t.cpp
```diff
@@ -21,13 +21,15 @@ uint128_t uint128_t::operator=(const uint128_t & rhs){
     return *this;
 }
 
-uint128_t::operator bool() const{
+uint128_t::operator bool() const
+{
```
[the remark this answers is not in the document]
MonsieurNicolas: I ran clang-format on the file. I can revert it if you want, but I figured we're diverging anyways from upstream and it's just easier to always run clang-format going forward

lib/util/uint128_t.h
```diff
@@ -70,13 +88,13 @@ class uint128_t{
...
+        IMPLICIT_UNSAFE_UINT128_OPS operator bool() const;
+        IMPLICIT_UNSAFE_UINT128_OPS operator char() const;
+        IMPLICIT_UNSAFE_UINT128_OPS operator int() const;
+        IMPLICIT_UNSAFE_UINT128_OPS operator uint8_t() const;
+        IMPLICIT_UNSAFE_UINT128_OPS operator uint16_t() const;
+        IMPLICIT_UNSAFE_UINT128_OPS operator uint32_t() const;
+        IMPLICIT_UNSAFE_UINT128_OPS operator uint64_t() const;
 
         // Bitwise Operators
         uint128_t operator&(const uint128_t & rhs) const;
```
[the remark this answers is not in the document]
MonsieurNicolas: yeah I only changed the ones that could potentially be used in the bug I fixed. We're going to have to take a much blunter instrument for the next round of changes to that class but that's outside of the scope of that PR.

src/herder/TransactionQueue.cpp
MonsieurNicolas: sure, fixed

src/herder/test/TransactionQueueTests.cpp
MonsieurNicolas: nope, just leftover: I thought I was going to hit `txINSUFFICIENT_FEE` in those cases, but that's not the case. I will remove this when I do the final rebase

src/transactions/FeeBumpTransactionFrame.cpp
MonsieurNicolas: Theoretically possible sure. I'll clamp it the same way for clarity.

src/transactions/TransactionFrame.cpp
MonsieurNicolas: I'm not 100% sure I understand your question.

There are two changes here:
* use the same code path when validating regardless of protocol version (as validation is not part of the protocol, I'd rather take the simpler route). This is just to simplify testing and not have to special case this in SDKs/Horizon
* not clamping down during validation. I think this is necessary because it's possible to have `adjustedFee > getFeeBid()` because the effective base fee (during surge pricing) is computed with a round up and I didn't want to potentially give a bid that was potentially too small

lib/util/uint128_t.cpp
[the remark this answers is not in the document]
MonsieurNicolas: Mmm, actually it doesn't look like clang-format did this. So I am going to revert that change

lib/util/uint128_t.h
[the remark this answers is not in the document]
jonjove: Alright I'm fine with that. Let's open an issue for this.

src/transactions/FeeBumpTransactionFrame.cpp
```diff
@@ -194,8 +194,11 @@ FeeBumpTransactionFrame::commonValidPreSeqNum(AbstractLedgerTxn& ltx)
     auto v2 = bigMultiply(mInnerTx->getFeeBid(), getMinFee(lh));
     if (v1 < v2)
     {
-        getResult().feeCharged =
-            bigDivide(v2, mInnerTx->getMinFee(lh), Rounding::ROUND_UP);
+        if (!bigDivide(getResult().feeCharged, v2, mInnerTx->getMinFee(lh),
+                       Rounding::ROUND_UP))
+        {
+            getResult().feeCharged = UINT64_MAX;
```
jonjove, line 200: `feeCharged` is `int64_t`, not `uint64_t`.

MonsieurNicolas: nice... I should have gotten a warning here

jonjove:
r+ 0e5c2a1

MonsieurNicolas:
had to disable a check in `checkApply` for protocol versions less than 11: the fee recommendation was not higher in that case because we always take the fee bid in older versions of the protocol

MonsieurNicolas:
I also added a commit to fix the build (doesn't compile when tests are disabled)

MonsieurNicolas:
r+ 8f356f7cf956fdeabc788c5e3846b4de79f4e191

PR #2628, opened 2020-07-17
Simulation: advance bucketlist level even for empty buckets

marta-lokhova:
Simulation bugfix, which only applies to really old (< 11) bucket versions.


MonsieurNicolas:
r+ 9ac04eb5fca4d9bd0fd3b3135e8d21d115989d1c

PR #2631, opened 2020-07-20
do not use temp variables in SQL queries

MonsieurNicolas:
fixes a stack corruption bug in test


MonsieurNicolas:
r+ ccb2a2abbf25bb2001713efafa6e63881c28f571

PR #2637, opened 2020-07-24
Set fail and/or badbit exception masks on fstreams.

graydon:
This just sets badbit on ifstream exception masks, and failbit|badbit on ofstreams.

Failbit is not set on ifstreams because read operations typically signal _completion_ by reading at-the-end and failing, setting both failbit and eof. Yet more magic and beauty of the iostreams library.


MonsieurNicolas:
r+ 6dab5dbad1bb62b532400656a0cb38b2dcf8aea8

PR #2638, opened 2020-07-25
Manual close end point now allows to close manually once

marta-lokhova:
This isn't well tested yet (and probably not simplified either), but opening a preliminary PR so that I can do some simulation testing.

### DIFF
--- a/src/herder/HerderImpl.cpp
+++ b/src/herder/HerderImpl.cpp
@@ -129,7 +129,7 @@ HerderImpl::bootstrap()
     assert(getSCP().isValidator());
     assert(mApp.getConfig().FORCE_SCP);
 
-    mLedgerManager.bootstrap();
+    mLedgerManager.moveToSynced();
     mHerderSCPDriver.bootstrap();
 
     ledgerClosed(true);
@@ -737,7 +737,7 @@ HerderImpl::ledgerClosed(bool synchronous)
         if (!mApp.getConfig().MANUAL_CLOSE)
             mTriggerTimer.async_wait(
                 std::bind(&HerderImpl::triggerNextLedger, this,
-                          static_cast<uint32_t>(nextIndex)),
+                          static_cast<uint32_t>(nextIndex), true),
                 &VirtualTimer::onFailureNoop);
     }
     else
@@ -828,15 +828,42 @@ HerderImpl::getMaxSeqInPendingTxs(AccountID const& acc)
     return mTransactionQueue.getAccountTransactionQueueInfo(acc).mMaxSeq;
 }
 
+void
+HerderImpl::setInSyncAndTriggerNextLedger()
+{
+    // We either have not set trigger timer, or we're in the
+    // middle of a consensus round. Either way, we do not want
+    // to trigger ledger, as the node is already making progress
+    if (mTriggerTimer.seq() > 0)
+    {
+        CLOG(DEBUG, "Herder") << "Skipping setInSyncAndTriggerNextLedger: "
+                                 "trigger timer already set";
+        return;
+    }
+
+    // Bring Herder and LM in sync in case they aren't
+    if (mLedgerManager.getState() == LedgerManager::LM_BOOTING_STATE)
+    {
+        mLedgerManager.moveToSynced();
+    }
+
+    // Trigger next ledger, without requiring Herder to properly track SCP
+    auto lcl = mLedgerManager.getLastClosedLedgerNum();
+    triggerNextLedger(lcl + 1, false);
+}
+
 // called to take a position during the next round
 // uses the state in LedgerManager to derive a starting position
 void
-HerderImpl::triggerNextLedger(uint32_t ledgerSeqToTrigger)
+HerderImpl::triggerNextLedger(uint32_t ledgerSeqToTrigger,
+                              bool checkTrackingSCP)
 {
     ZoneScoped;
     ZoneValue(static_cast<int64_t>(ledgerSeqToTrigger));
 
-    if (!mHerderSCPDriver.trackingSCP() || !mLedgerManager.isSynced())
+    auto isTrackingValid = mHerderSCPDriver.trackingSCP() || !checkTrackingSCP;
+
+    if (!isTrackingValid || !mLedgerManager.isSynced())
     {
         CLOG(DEBUG, "Herder") << "triggerNextLedger: skipping (out of sync) : "
                               << mApp.getStateHuman();
--- a/src/main/CommandHandler.cpp
+++ b/src/main/CommandHandler.cpp
@@ -161,13 +161,14 @@ CommandHandler::manualClose(std::string const& params, std::string& retStr)
 {
     if (mApp.manualClose())
     {
-        retStr = "Forcing ledger to close...";
+        retStr = "Triggering a new consensus round";
     }
     else
     {
-        retStr =
-            "Set MANUAL_CLOSE=true in the stellar-core.cfg if you want this "
-            "behavior";
+        retStr = "Set MANUAL_CLOSE=true in the stellar-core.cfg if you want to "
+                 "close every ledger manually. Otherwise, run stellar-core "
+                 "with --wait-for-consensus flag to close ledger once and "
+                 "trigger consensus. Ensure NODE_IS_VALIDATOR is set to true.";
     }
 }
 
--- a/docs/software/commands.md
+++ b/docs/software/commands.md
@@ -275,8 +275,10 @@ format.
   When a nonzero I is given, a spike will occur every I seconds injecting S transactions on top of `txrate`.
 
 * **manualclose**
-  If MANUAL_CLOSE is set to true in the .cfg file. This will cause the current
-  ledger to close.
+  If MANUAL_CLOSE is set to true in the .cfg file, this will cause the current
+  ledger to close. If MANUAL_CLOSE is set to false, allows a validating node
+  that is waiting to hear about consensus from the network to force ledger close,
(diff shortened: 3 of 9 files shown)


graydon:
Looks fairly good but I am a little concerned with a few points:
  - The names "once" and "all" are a little uninformative / non-obvious to users, and passing a boolean flag named `once` that we check the _negation_ of seems a little hard to follow; especially when combined with the fact that "all" needs the flag to be set, and `once` needs FORCE_SCP to _not_ be set. Overall it just feels confusing. This is subjective and you're welcome to tell me "this is the best that can be done!" but it's sorta a gut sense I get reading it, that I'll have a hard time editing the code in the future / telling what it's supposed to do in which combinations.
  - More worryingly: "once" can be called more than once, and I'm not sure it's actually safe to call Herder::bootstrap more than once. Maybe? Might also cause havoc.
  - I'm not 100% clear on why you need to restoreScpState in the non-FORCE_SCP path of bootstrap. What's different about it? 
  - Overall I suspect the close-once flag might serve a subtly-different purpose than the `MANUAL_CLOSE` flag. If I understand correctly, `MANUAL_CLOSE` was added as a kind of escape hatch to let running networks that have "lost leadership" get forced back into operation by an operator. This is, I believe, why it is also not hidden behind `BUILD_TESTS`: so that production nodes can be nudged back into SCP when they have given up on it or are all waiting for one another. This might no longer be necessary at all (given the changes you landed in https://github.com/stellar/stellar-core/pull/2612) but if it is true (or if we want to keep it around for purposes of backward compatibility with scripts, or mere paranoia) then I think that's different than what you're adding here. I think what you're adding here is a one-time, idempotent trigger that _negates_ `--wait-for-consensus` mode and says ok, stop waiting now. If that's true I might suggest:
    - Move `--wait-for-consensus` behind `BUILD_TESTS` if we think nobody needs to do this in production (I think .. not?)
    - Rename `Config.FORCE_SCP` to `Config.WAIT_FOR_CONSENSUS` to reduce the number of double negatives. Limit the residue of the `FORCE_SCP` name to the deprecation warnings and back-compat config option parser. 
    - Add a new HTTP method called, say, `bootstrap` that _is_ behind `BUILD_TESTS` that calls bootstrap iff `WAIT_FOR_CONSENSUS` is true and bootstrap hasn't happened yet. If bootstrap has happened, return an error indicating that.

How does that sound to you?

marta-lokhova:
Thanks for taking a look! I made changes to address the first two bullets, I agree that keeping track of all the flags was confusing.

Regarding "restoreScpState" in bootstrap: when we startup with --wait-for-consensus, we're not in sync technically, but `ledgerClosed` expects herder to be in sync. So with this change we basically pretend like we got in sync, and closed ledger, so we can trigger a new one.

For the last bullet, I haven't implemented all of your suggestions yet, as I realized that I'm confused by the existing manual_close functionality. So I'd like to hash out some questions on that first. 
My main issue is that I don't quite understand how MANUAL_CLOSE is supposed to be used, as the documentation seems to contradict the implementation.
* `manualclose` endpoint says that it forces ledger close. In reality, I don't think it's true -- nodes trigger a consensus round. When MANUAL_CLOSE is set to true, nodes discard scp messages from peers, so I don't understand how it is possible to close a ledger for networks that consist of more than 1 node.
* I don't think manualclose was intended to be a production endpoint. `commands.md` and `Application::manualClose` both claim that it is for testing only.

So I guess my question is: do we want to keep manualclose as is, and just have `bootstrap` endpoint like you suggested, or do we want to re-visit manualclose, and change its behavior to the new bootstrapping logic? (@graydon @MonsieurNicolas )

graydon:
Fine questions! I await @MonsieurNicolas input as I do not have a clear sense of the answers myself, clearly :)

src/main/CommandHandler.cpp
```diff
@@ -104,6 +104,7 @@ CommandHandler::CommandHandler(Application& app) : mApp(app)
     addRoute("generateload", &CommandHandler::generateLoad);
     addRoute("testacc", &CommandHandler::testAcc);
     addRoute("testtx", &CommandHandler::testTx);
+    addRoute("bootstrap", &CommandHandler::bootstrap);
```
MonsieurNicolas, line 107: should we always expose this? This seems to be useful in general to spin up nodes, so maybe shouldn't be limited to test builds?

src/herder/HerderImpl.cpp
```diff
@@ -125,14 +125,25 @@ HerderImpl::getStateHuman() const
...
-    assert(getSCP().isValidator());
-    assert(mApp.getConfig().FORCE_SCP);
+    if (!mBootstrapped)
+    {
+        mBootstrapped = true;
+
+        CLOG(INFO, "Herder") << "Force joining SCP with local state";
+        assert(getSCP().isValidator());
+
+        if (!mApp.getConfig().FORCE_SCP)
```
MonsieurNicolas, line 135: Should we just always restore SCP state in the code path `mHerder->restoreState()` and not do it arbitrarly late: the only difference between `FORCE_SCP` and no `FORCE_SCP` will then only be that we trigger a ledger  or not.

src/herder/HerderImpl.cpp
```diff
@@ -125,14 +125,25 @@ HerderImpl::getStateHuman() const
...
+        {
+            auto const& lcl = mLedgerManager.getLastClosedLedgerHeader();
+            mHerderSCPDriver.restoreSCPState(lcl.header.ledgerSeq,
+                                             lcl.header.scpValue);
+        }
 
-    mLedgerManager.bootstrap();
-    mHerderSCPDriver.bootstrap();
+        mLedgerManager.bootstrap();
```
MonsieurNicolas, line 142: I am really not sure this is right.
we're introducing a race between whatever that node is doing and calling `bootstrap`: those bootstrap calls may reset a bunch of things that the node was actually doing.

I thought the less dangerous version of this PR was to just allow `manualclose` to be called at any time?

src/herder/HerderImpl.cpp
marta-lokhova: I did that initially, but then moved "restore state" path to later for the following reasons:
* If we move this to `mHerder->restoreState()`, we'd need to setup a timer to properly go out of sync as `restoreState` is used for normal node operation as well. For the simulation purposes, I don't think we want that, as nodes should stay "in sync" until manualclose is called (which can take arbitrarily long) in order to trigger the next ledger 
* minor: there is an existing optimization to not go in sync when we're at the genesis ledger and waiting for consensus  

I could change it as per your suggestion, but it seems like this way we get more flexibility. What do you think?

marta-lokhova:
That makes sense @MonsieurNicolas. I think I initially started with triggering the next ledger instead of bootstrapping, but as you pointed out, there is some prep required to bring the node into proper state to trigger, and bootstrap was already doing some of it. I'll update the code to check timers, then get in sync, then call `triggerNextLedger`.

src/herder/HerderImpl.cpp
```diff
@@ -825,6 +825,36 @@ HerderImpl::getMaxSeqInPendingTxs(AccountID const& acc)
...
+    // middle of a consensus round. Either way, we do not want
+    // to trigger ledger, as the node is already making progress
+    if (mTriggerTimer.seq() > 0)
+    {
+        CLOG(DEBUG, "Herder") << "Skipping setInSyncAndTriggerNextLedger: "
+                                 "trigger timer already set";
+        return;
+    }
+
+    // Bring Herder and LM in sync in case they aren't
```
MonsieurNicolas, line 841: Yikes, it ends up being not as straightforward as I thought 😅 

As far as I can tell, `restoreSCPState` doesn't do what you want when lcl is genesis (so starting fresh) as it no-ops in that case.

More importantly, this code is still unsafe, imagine the following sequence takes place:
1. node waits for consensus
2. node receives enough messages to externalize
3. (node may transition to catchup)
4. node goes out of sync
5. we invoke the `manualclose` endpoint

At this point
* `restoreSCPState` will perform potentially unsafe operations (overrides a bunch of SCP related fields, potentially rolls back certain things in SCP itself)
* `mLedgerManager.bootstrap` potentially sets the wrong state (if catching up)

Now, do we need to actually do this?
* for `mLedgerManager`
   * if it's "catching up" we have to no-op
   * otherwise, we can call `mLedgerManager.bootstrap()` (that will either no-op or perform the BOOTING->SYNCED transition) - maybe that function should be renamed `forceTransitionToSynced`?
* for SCP, the only thing we care about is to make sure that we're nominating against the local state (so `lcl`) so that it can properly validate values. I don't think we need to mess with "tracking"
   * basically in this particular situation, we can bypass the check `!mHerderSCPDriver.trackingSCP()` ; we still need to ensure that `mLedgerManager.isSynced()` is true though
   * we could call a variation of `restoreSCP` on startup, but that is a minor improvement. This would make it safer if we're using this endpoint when restarting the network from a state that's not genesis
       * we'd remove the condition on `FORCE_SCP` (ie, always restore when not on genesis)
       * make it no-op if the node is not validating
       * always restores the SCP state
       * skips setting tracking state when `FORCE_SCP` is not set (doesn't call `HerderSCPDriver::restoreSCPState`)

marta-lokhova: Updated.

As discussed offline, I did not change `restoreSCPState`, as it already handles non-genesis ledgers properly.

MonsieurNicolas:
r+ b513585af87a110624ccb79f461a0457a1839931

MonsieurNicolas:
@latobarita: retry

PR #2640, opened 2020-07-31
require a recent libsodium version or default to the one supplied

MonsieurNicolas:


graydon:
r+ 4b7b33d

MonsieurNicolas:
@latobarita: retry

PR #2645, opened 2020-08-01
fix crash on shutdown (as to get the proper error message)

MonsieurNicolas:
Fixes a crash on shutdown occurring when running `new-hist` without a database

This was introduced by #2622 : when there is no database, `shutdown` throws, which bypasses joining threads (and deleting a thread object when a thread is running violates spec).


graydon:
r+ 7c74acf

PR #2652, opened 2020-08-05
replace all remaining LRUcache by RandomEvictionCache

MonsieurNicolas:
lrucache is tripping builds with "extrachecks" with libc++ enabled (on clang),
There might be a bug in its implementation.

Moving to `RandomEvictionCache`, I can no longer cause core to crash.


graydon:
r+ 84435aa

PR #2655, opened 2020-08-05
Increase reliability when fetching SCP envelopes

MonsieurNicolas:
This PR makes a small adjustment to the way we manage envelopes that are still in "fetching" state:
when in that state, we ensure that we always fetch data. Old behavior may in some rare occasion leave envelopes in a "half fetch" state with no fetchers.


graydon:
r+ b9abf4f

MonsieurNicolas:
r+ 833558ee0329eb59de52f674429541bd8f216579

MonsieurNicolas:
(missed a test)

PR #2662, opened 2020-08-07
improve how extrachecks are enabled

MonsieurNicolas:
* libstdc++ allows for better checks
* libc++ needs less strict checks by default

Tested with clang++ by performing several catchups against the live network


MonsieurNicolas:
r+ 60bb8c4834ceed8d5165cce087b197e6d7e7d791

Issue #2579, opened 2020-06-12
Add network_passphrase to stellar-history.json (Version: N/A)


bartekn:
## Issue Description

Currently, JSON in a HAS file in history archives does not have a `network_passphrase` field. This can be helpful to check if history archive actually contains the data for a specified network.

Since history archive is not specified by any CAP I'm adding this issue here.

## Additional context

One of Horizon users used pubnet config (connected to pubnet Stellar-Core) but testnet archive. Because of that Horizon was waiting indefinitely for a checkpoint ledger after a given pubnet ledger but it never appeared because testnet ledger sequence numbers are much smaller than in pubnet.

hidenori-shinohara:
@MonsieurNicolas Can you assign this to me? I don't seem to have write access to assignees.

PR #2666, opened 2020-08-10
Add networkPassphrase to HAS

hidenori-shinohara:
# Description

https://github.com/stellar/stellar-core/issues/2579

Add `networkPassphrase` to HAS (e.g., This will add `networkPassphrase` to `.well-known/stellar-history.json`)

# Checklist
- [x] Compiles
- [x] Ran all tests


src/history/HistoryArchive.cpp
```diff
@@ -36,7 +36,7 @@
 namespace stellar
 {
 
-unsigned const HistoryArchiveState::HISTORY_ARCHIVE_STATE_VERSION = 1;
+unsigned const HistoryArchiveState::HISTORY_ARCHIVE_STATE_VERSION = 2;
```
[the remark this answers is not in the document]
marta-lokhova: Just to be sure, we're okay not guarding this with a protocol upgrade, which may result in HAS files for the same ledger to be different on different nodes?

src/history/HistoryArchive.h
```diff
@@ -64,12 +64,14 @@ struct HistoryArchiveState
 
     unsigned version{HISTORY_ARCHIVE_STATE_VERSION};
     std::string server;
+    std::string networkPassphrase;
```
marta-lokhova, line 67: Now that we're including the network passphrase, should we have a sanity check when processing HAS files in catchup?

hidenori-shinohara: Done!

src/history/HistoryArchive.cpp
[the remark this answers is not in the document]
graydon: They already can differ -- they have the version of the node written into them (the `server` field).

I agree we might want to get rid of `next` with this change.

src/catchup/CatchupWork.cpp
```diff
@@ -242,6 +242,13 @@ CatchupWork::runCatchupStep()
     }
 
     auto const& has = mGetHistoryArchiveStateWork->getHistoryArchiveState();
+    if (has.networkPassphrase != mApp.getConfig().NETWORK_PASSPHRASE)
```
graydon, line 245: Won't this trip when catching up to a HAS that has an empty networkPassphrase (from older versions)?

src/testdata/stellar-history.testnet.6714240.json
```diff
@@ -0,0 +1,95 @@
+{
```
MonsieurNicolas, line 1: I recommend moving the testing of the new file format to its own commit *before* making any other code change, that way you can verify that the old version can indeed consume this new format.

hidenori-shinohara:
* As for `next`, I'm not sure if I can get rid of it in this version bump. Simply removing `CEREAL_NVP(next)` from the serialize function for `HistoryStateBucket` might not be a good idea because nodes running the current version fails if it sees HAS files without next. (`JSON Parsing failed - provided NVP (next) not found`) I was trying to figure out a way to use the version upgrade, but the serialize function for `HistoryArchiveState` and the one for `HistoryStateBucket` are different and the serialize function for `HistoryStateBucket` doesn't seem to know the version in `HistoryArchiveState `.
 The best alternative I came up with is to include `CEREAL_NVP(next)` in a try-catch statement, so we will be able to stop outputting it in the future once every node is running the version of code with `CEREAL_NVP(next)' in a try-catch statement.
https://github.com/stellar/stellar-core/pull/2666/commits/bb92da4f3cbdf3534f1d630bec3d98ffe414a6d2

* As for the `networkPassphrase`,  I noticed that `HistoryArchiveState::load` only runs when `version == HISTORY_ARCHIVE_STATE_VERSION`. However, if we bump the version of HAS right now, nodes running the current version will fail when they see version-2 HAS. One way that I learned from Marta that would solve this is to guard this with a protocol upgrade. @MonsieurNicolas What are your thoughts on this?

MonsieurNicolas:
> 
> 
>     * As for `next`, I'm not sure if I can get rid of it in this version bump. Simply removing `CEREAL_NVP(next)` from the serialize function for `HistoryStateBucket` might not be a good idea because nodes running the current version fails if it sees HAS files without next. (`JSON Parsing failed - provided NVP (next) not found`) I was trying to figure out a way to use the version upgrade, but the serialize function for `HistoryArchiveState` and the one for `HistoryStateBucket` are different and the serialize function for `HistoryStateBucket` doesn't seem to know the version in `HistoryArchiveState `.
>       The best alternative I came up with is to include `CEREAL_NVP(next)` in a try-catch statement, so we will be able to stop outputting it in the future once every node is running the version of code with `CEREAL_NVP(next)' in a try-catch statement.
>       [bb92da4](https://github.com/stellar/stellar-core/commit/bb92da4f3cbdf3534f1d630bec3d98ffe414a6d2)

I don't think we can remove `next` support. We use `HAS` for two things:
* publish to archives
   * old snapshots may have `next` set?
* persist the bucket list state when running
   * for sure we don't resolve merges every time we close a ledger, so `next` may be set in this situation
   * we should open an issue to decouple "HAS used in archives" and "HAS used to persist the local state"

In any case, I think that we should *keep* `next` but perform stricter runtime validations after we unmarshal it from json:
this way, we don't have to mess with the schema and yet can detect when there are unexpected items in there.
For example, I think that when we read a HAS from archive for protocol version 13 we should never have "next" set to anything but "CLEAR"? @marta-lokhova to confirm.

> 
>     * As for the `networkPassphrase`,  I noticed that `HistoryArchiveState::load` only runs when `version == HISTORY_ARCHIVE_STATE_VERSION`. However, if we bump the version of HAS right now, nodes running the current version will fail when they see version-2 HAS. One way that I learned from Marta that would solve this is to guard this with a protocol upgrade. @MonsieurNicolas What are your thoughts on this?

I am not sure we need to bump the schema version: we're only *adding* fields and most JSON parsers have as default behavior to just ignore the extra properties.
In our case, we'd simply only compare the passphrase to the one from the local node if the passphrase in the archive is not empty.

That said, it looks like something is off with your PR: the first commit, which adds files with the new format, seems to "work just fine" even though old code clearly cannot handle this new format. So... something is wrong in that commit: I'd expect to see some tests updated to ensure that we properly fail (or maybe they fail and you didn't try?). I would first correct this situation before reverting the schema change.

graydon:
We should not do a version bump here, no. First of all, as @hidenori-shinohara pointed out to me with reference to this code:

https://github.com/stellar/stellar-core/blob/master/src/history/HistoryArchive.cpp#L140

There's not actually any code in any versions of core that can handle version changes. If the version doesn't match, core just throws / crashes. We haven't written code to deal with non-identical versions yet. So we'd need to write some version-awareness code first.

Secondly, concerning this bit:

> I think that when we read a HAS from archive for protocol version 13 we should never have "next" set to anything but "CLEAR"? @marta-lokhova to confirm.

The post-protocol-13 HASs do indeed have the `next` field set to an empty state `{ "state": 0 }` but this can't be reliably turned into a validity check, because (a) there's no indication in an HAS of which protocol version it was composed with (b) each HAS consists of _buckets_ from multiple different versions (they roll over over an extended period of time) and (c) fetching a bucket's protocol version requires having it on disk to open and sniff, and we often parse HAS files before we have downloaded a bucket to local disk. So it's like .. triply impossible :(

I agree that 

  - _after_ we've split the HAS class into separate classes for local-database and an archive-only use cases, we can possibly remove `next`-reading / storing-in-memory from the archive-only use case (assuming cereal doesn't mind ignored fields).
  - _after_ we've added machinery to gracefully handle multiple HAS versions, rather than just throwing, we might consider also no longer _producing_ a `next` field and bumping the HAS version.

But until either of those conditions occurs, I think we ought to leave the `next` field alone.

hidenori-shinohara:
I made several changes from another branch and force pushed it. Specifically,

* I got rid of the version bump.
   * Because of this, now the sanity check will confirm that the network passphrase is either empty or the same as the one in the config. (Previously, I checked if the version is old or the network passphrase matches) 
* I got rid of any changes related to `next`.

> That said, it looks like something is off with your PR: the first commit, which adds files with the new format, seems to "work just fine" even though old code clearly cannot handle this new format. So... something is wrong in that commit: I'd expect to see some tests updated to ensure that we properly fail (or maybe they fail and you didn't try?). I would first correct this situation before reverting the schema change.

(Though we talked about this over a video call, I'll document this for future reference) This was because the `load` function wasn't tested in the Serialization round trip test and it was the `fromString` function that was tested. `fromString` doesn't seem to care about the version, but `load` does. I've added `load` to the Serialization round trip test so that it is tested now.

src/history/HistoryArchive.h
```diff
@@ -96,16 +98,34 @@ struct HistoryArchiveState
...
     serialize(Archive& ar)
     {
-        ar(CEREAL_NVP(version), CEREAL_NVP(server), CEREAL_NVP(currentLedger),
-           CEREAL_NVP(currentBuckets));
+        ar(CEREAL_NVP(version), CEREAL_NVP(server), CEREAL_NVP(currentLedger));
+        try
+        {
+            ar(CEREAL_NVP(networkPassphrase));
+        }
+        catch (...)
```
MonsieurNicolas, line 106: do not use `catch(...)` but instead catch the exception that you expect in this case (security issue)

src/history/HistoryArchive.h
```diff
@@ -96,16 +98,34 @@ struct HistoryArchiveState
...
     }
 
     template <class Archive>
     void
     serialize(Archive& ar) const
     {
-        ar(CEREAL_NVP(version), CEREAL_NVP(server), CEREAL_NVP(currentLedger),
-           CEREAL_NVP(currentBuckets));
+        ar(CEREAL_NVP(version), CEREAL_NVP(server), CEREAL_NVP(currentLedger));
+        try
```
MonsieurNicolas, line 119: this is the *writer* specialization, so we don't need this try/catch

src/history/test/SerializeTests.cpp
```diff
@@ -12,23 +12,40 @@ using namespace stellar;
...
 {
-    std::vector<std::string> testFiles = {
-        "stellar-history.testnet.6714239.json",
-        "stellar-history.livenet.15686975.json"};
-    for (auto const& fn : testFiles)
+    std::vector<std::string> inputFiles = {
+        "stellar-history.testnet.6714239.input.json",
+        "stellar-history.livenet.15686975.input.json",
+        "stellar-history.testnet.6714239.networkPassphrase.json"};
+    std::vector<std::string> outputFiles = {
```
MonsieurNicolas, line 19: you should just add some prefix string to the output file name (so you don't need this vector) and that way we have a clear mapping between files.

src/history/test/SerializeTests.cpp
```diff
@@ -12,23 +12,40 @@ using namespace stellar;
...
             HistoryArchiveState has;
-            has.fromString(fromFile);
-            REQUIRE(fromFile == has.toString());
+
+            // Test fromString
+            has.fromString(input);
+            REQUIRE(output == has.toString());
+
+            // Test load
+            has.load(inputPath);
```
MonsieurNicolas, line 47: do not use the same `has` for reading and writing in case something goes wrong (like load no-ops)

src/testdata/stellar-history.livenet.15686975.input.json
```diff
@@ -0,0 +1,94 @@
+{
```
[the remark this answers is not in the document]
MonsieurNicolas: Also, I don't see a reason to rename the existing files, so keep `stellar-history.livenet.15686975.json` as is

src/history/test/SerializeTests.cpp
MonsieurNicolas: also, I think files should be named `expected-output-XYZ.json` - output is a lie :)

src/history/test/SerializeTests.cpp
```diff
@@ -12,23 +12,40 @@ using namespace stellar;
...
+        "stellar-history.testnet.6714239.output.json",
+        "stellar-history.livenet.15686975.output.json",
+        "stellar-history.testnet.6714239.networkPassphrase.json"};
+    for (int i = 0; i < inputFiles.size(); i++)
     {
         std::string fnPath = "testdata/";
-        fnPath += fn;
-        SECTION("Serialize " + fnPath)
+        std::string inputPath = fnPath + inputFiles[i];
+        std::string outputPath = fnPath + outputFiles[i];
```
MonsieurNicolas, line 27: same here, variable should be called `expectedOutputPath`

src/history/test/SerializeTests.cpp
```diff
@@ -12,23 +12,40 @@ using namespace stellar;
...
         {
-            std::ifstream in(fnPath);
+            std::ifstream in(inputPath);
             in.exceptions(std::ios::badbit);
-            std::string fromFile((std::istreambuf_iterator<char>(in)),
-                                 std::istreambuf_iterator<char>());
+            std::string input((std::istreambuf_iterator<char>(in)),
+                              std::istreambuf_iterator<char>());
+
+            std::ifstream out(outputPath);
```
MonsieurNicolas, line 35: `expectedOutputStream` instead of `out`

src/history/test/SerializeTests.cpp
```diff
@@ -12,23 +12,40 @@ using namespace stellar;
...
+            out.exceptions(std::ios::badbit);
+            std::string output((std::istreambuf_iterator<char>(out)),
```
MonsieurNicolas, line 37: `expectedOutput` instead of `output`

src/history/test/SerializeTests.cpp
MonsieurNicolas: (and yes this will cause the one with the passphrase to be duplicated but that's fine)

src/history/HistoryArchive.h
MonsieurNicolas: instead what you can do is only call `ar` on `networkPassphrase` when it's not empty (and avoid putting empty strings in the json file)

MonsieurNicolas (changes requested):

src/history/HistoryArchive.h
MonsieurNicolas: I think it's `cereal::Exception` here

src/testdata/stellar-history.livenet.15686975.input.json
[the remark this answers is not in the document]
hidenori-shinohara: Moved all the json files to `src/history/serialize-tests` and updated `src/Makefile.am` accordingly! I also reverted the change that renames the file.

src/history/HistoryArchive.h
hidenori-shinohara: Done!

src/history/HistoryArchive.h
hidenori-shinohara: Fixed it!

src/history/test/SerializeTests.cpp
hidenori-shinohara: Renamed the file to `expected-output-...`.

After adding the logic to print `networkPassphrase` only if it's non-empty, the expected output file is only necessary when the _current_ code reads a file _with_ `networkPassphrase` and output a file _without_ `networkPassphrase`. In all other cases, the input and expected output are the same. So I decided to keep these two vectors in order to avoid copying a lot of input files. Let me know what you think!

src/history/test/SerializeTests.cpp
hidenori-shinohara: Makes sense! I created a new object `hasLoad` to test `load`.

src/history/test/SerializeTests.cpp
hidenori-shinohara: Done!

src/history/test/SerializeTests.cpp
hidenori-shinohara: Done!

src/history/test/SerializeTests.cpp
hidenori-shinohara: Done!

src/history/HistoryArchive.h
```diff
@@ -96,16 +98,29 @@ struct HistoryArchiveState
...
     }
 
     template <class Archive>
     void
     serialize(Archive& ar) const
     {
-        ar(CEREAL_NVP(version), CEREAL_NVP(server), CEREAL_NVP(currentLedger),
-           CEREAL_NVP(currentBuckets));
+        ar(CEREAL_NVP(version), CEREAL_NVP(server), CEREAL_NVP(currentLedger));
+        if (networkPassphrase != "")
```
MonsieurNicolas, line 119: the proper way to check this is to use `.empty()`

MonsieurNicolas (changes requested):
looks pretty much ready to be merged. Just a minor change

src/history/HistoryArchive.h
hidenori-shinohara: Done!

MonsieurNicolas:
r+ dcf57dc0d96c27725b3c389b44b8cea825b837aa

PR #2684, opened 2020-08-27
Bump tracy submodule and build with -DTRACY_ONLY_IPV4

graydon:
The goal here is to have tracy only listen on IPv4 interfaces so that we can talk to it through the IPv4-only kubernetes port-forwarding system.

The submodule bump includes tracy v0.7 => v0.7.1 because the latter is the version the patch to upstream is made against.


MonsieurNicolas:
r+ 678c21908c830176431c2a4372fd7a2a153513f1

PR #2685, opened 2020-08-27
allow access to HTTP endpoint -- /info in particular -- during catchup.

graydon:
This very-slightly relaxes the configuration during command-line catchup so that the HTTP endpoint is enabled. Specifically so that kubernetes health checks and the like can call the /info endpoint and monitor the progress of a long catchup.

I included a couple tightenings of the HTTP endpoints that one might want to _not_ be allowed to call in such a circumstance -- disabling the /tx endpoint when not synced and the /connect endpoint when standalone -- which shouldn't change semantics as far as I can tell, but I can drop those if they're contentious.

### DIFF
--- a/src/main/CommandHandler.cpp
+++ b/src/main/CommandHandler.cpp
@@ -81,29 +81,37 @@ CommandHandler::CommandHandler(Application& app) : mApp(app)
         addRoute("maintenance", &CommandHandler::maintenance);
     }
 
-    addRoute("bans", &CommandHandler::bans);
+    if (!mApp.getConfig().RUN_STANDALONE)
+    {
+        addRoute("bans", &CommandHandler::bans);
+        addRoute("connect", &CommandHandler::connect);
+        addRoute("droppeer", &CommandHandler::dropPeer);
+        addRoute("peers", &CommandHandler::peers);
+        addRoute("quorum", &CommandHandler::quorum);
+        addRoute("scp", &CommandHandler::scpInfo);
+        addRoute("stopsurvey", &CommandHandler::stopSurvey);
+#ifndef BUILD_TESTS
+        addRoute("getsurveyresult", &CommandHandler::getSurveyResult);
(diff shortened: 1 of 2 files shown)


src/main/CommandHandler.cpp
```diff
@@ -559,6 +564,11 @@ CommandHandler::tx(std::string const& params, std::string& retStr)
     ZoneScoped;
     std::ostringstream output;
 
+    if (mApp.getState() != Application::State::APP_SYNCED_STATE)
```
MonsieurNicolas, line 567: This is a breaking change for Horizon ; do we really need this as the only thing it's doing is to stop accumulating transactions in memory

src/main/CommandHandler.cpp
```diff
@@ -290,6 +290,11 @@ CommandHandler::logRotate(std::string const& params, std::string& retStr)
 void
 CommandHandler::connect(std::string const& params, std::string& retStr)
 {
+    if (mApp.getConfig().RUN_STANDALONE)
```
MonsieurNicolas, line 293: we should just not register (ie `addRoute`) the overlay related endpoints in standalone mode:
connect, droppeer, peers, quorum, scp, surveytopology, stopsurvey, getsruveyresult, unban

src/main/CommandHandler.cpp
graydon: No, we do not need this. I will revert it.

src/main/CommandHandler.cpp
graydon: ok.

MonsieurNicolas:
r+ 7932d561ec48018ea4d304e06825a7abe38ed220

MonsieurNicolas:
Looks like some tests are maybe impacted by the route changes? @graydon

graydon:
Derp yeah sorry the endpoint has to be enabled when BUILD_TESTS. Fixed, re-landing.

graydon:
r+ 20acb82

PR #2690, opened 2020-09-01
Restructure is-a-component-of relationship among some Herder classes

jonjove:
Restructures the is-a-component-of relationship among some Herder classes.


MonsieurNicolas:
Sorry this code is a mess, and this change is not going in the right direction:
* from an "ownership" point of view, herder has to be one owning future envelopes as data can be "owned" by SCP or by overlay.
* the issue is that we're using in many places "SCP/SCPDriver" to gather data like "transaction sets", but we should really not be using that, and instead go through "Herder"

As any refactor is conflicting with other changes I opened https://github.com/stellar/stellar-core/pull/2693 that should fix the issue you're trying to fix in a more surgical way.

jonjove:
Closing in favor of #2693

PR #2693, opened 2020-09-02
make quorum tracker use only what it needs

MonsieurNicolas:
This PR fixes a small issue in how herder is initialized.

Approach taken here is to simply not use members that may not be fully initialized

### DIFF
--- a/src/herder/QuorumTracker.h
+++ b/src/herder/QuorumTracker.h
@@ -24,11 +24,11 @@ class QuorumTracker : public NonMovableOrCopyable
     using QuorumMap = std::unordered_map<NodeID, SCPQuorumSetPtr>;
 
   private:
-    SCP& mSCP;
+    NodeID const mLocalNodeID;
     QuorumMap mQuorum;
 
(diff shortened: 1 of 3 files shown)


src/herder/QuorumTracker.h
```diff
@@ -24,11 +24,11 @@ class QuorumTracker : public NonMovableOrCopyable
     using QuorumMap = std::unordered_map<NodeID, SCPQuorumSetPtr>;
 
   private:
-    SCP& mSCP;
+    NodeID mLocalNodeID;
```
marta-lokhova, line 27: safer to use const

marta-lokhova (approved):
one minor thing, otherwise LGTM

src/herder/QuorumTracker.h
MonsieurNicolas: fixed!

marta-lokhova:
r+ ba92760c34f79c75757ec5a48bf786c3ecf409fd

Issue #2695, opened 2020-09-03
[Stellar Playground: A browser environment for testing Stellar Network client code] (Version: [14.1.0])


rokopt:
## Description

The concept of the "Stellar Playground", a testing and learning tool for people developing or considering developing on the Stellar Network, has evolved from ideas of @bartekn through a series of conversations (see below).

## Prior related issues / PRs

- [Dry run mode for transaction submission · Issue #1391 · stellar/stellar-core](https://github.com/stellar/stellar-core/issues/1391)

- [Green: A Language for Building Stellar Ledger Scenarios](https://github.com/stellar/stellar-protocol/pull/604) ([Draft SEP](https://github.com/stellar/stellar-protocol/blob/46cd39b5dc191ae302dce89df93dfc756d3f71bc/ecosystem/sep_green.md))

- [change "manual close" to use `VIRTUAL_TIME` · Issue #1760 · stellar/stellar-core](https://github.com/stellar/stellar-core/issues/1760)

- [Run in-memory / start-from](https://github.com/stellar/stellar-core/issues/2486)

- [Captive Stellar-Core wrapper server](https://github.com/stellar/go/issues/2841) (PR: [Add HTTP server for managing a remote captive core instance](https://github.com/stellar/go/pull/2896))

- Example of the kind of question to which we’d like to be able to demonstrate an answer: https://stellar.stackexchange.com/questions/4093/which-horizon-post-transaction-responses-are-final-and-immutable

- [Captive Stellar-Core: Stream meta data starting from any ledger](https://github.com/stellar/stellar-core/issues/2634)

- [Switch to `stellar-core run-from` command when added to Stellar-Core](https://github.com/stellar/go/issues/2842)

- [Remove ambiguous result from "tx" endpoint · Issue #1999 · stellar/stellar-core](https://github.com/stellar/stellar-core/issues/1999)

## External references / models

The following are examples of the kinds of things we might want to build over time:

- [Albedo playground](https://albedo.link/playground): At a glance, it looks to me as though its “Playground” is far more constrained than what we're envisioning here; it's more like the Stellar Laboratory, allowing you to enter constrained sets of parameters to generate various operations.  It doesn't give you an interpreter.

- [Go playground & tutorials](https://play.golang.org/)

- [Katacoda](https://www.katacoda.com/)

- [repl.it](https://repl.it/)

[...]

Issue #1760, opened 2018-08-20
change "manual close" to use `VIRTUAL_TIME`


MonsieurNicolas:
manual close mode is used when testing, and can be used to replay transactions or debug issues;

there are a too many "gotcha's" that come with the way manual close works right now (for example, if the caller attempts to close too fast, some failures occur, or certain timeouts trigger).

Best option seems to be:
* to use `VIRTUAL_TIME`
* move the clock forward by 5 seconds when closing
* allow to set the clock to an arbitrary date (optional parameter to `manualclose`, probably in the same unit than the `close time` used in the ledger header).
  * This allows to replay transaction sets exactly the same way than found on a live network for example.

rokopt:
A couple of notes:

- We should only use `VIRTUAL_TIME` when `RUN_STANDALONE` is specified in addition to `MANUAL_CLOSE`.  In that case, we can allow `manualclose` with parameters to override the defaults (such as the next ledger sequence number and close time).  There are networked `supercluster` tests that rely on parameter-less `manualclose`.

- When we do use `VIRTUAL_TIME` as a result of both `MANUAL_CLOSE` and `RUN_STANDALONE` being set, we must also validate that automatic maintenance is off (`AUTOMATIC_MAINTENANCE_COUNT`/`AUTOMATIC_MAINTENANCE_PERIOD` are `0`).  Otherwise, we end up trimming history in a tight loop around `ExternalQueue::deleteOldEntries()`.

rokopt:
Here's an example config file that I've been using to test `VIRTUAL_TIME` induced by `MANUAL_CLOSE` and `RUN_STANDALONE` (see PR #2696 ):

```
NODE_SEED="SA7FGJMMUIHNE3ZPI2UO5I632A7O5FBAZTXFAIEVFA4DSSGLHXACLAIT"
NODE_HOME_DOMAIN="developer-testing.stellar.org"
NODE_IS_VALIDATOR=true
NETWORK_PASSPHRASE="Standalone 'network'"
DATABASE="sqlite3:///tmp/developer-testing.db"

MANUAL_CLOSE=true
RUN_STANDALONE=true
AUTOMATIC_MAINTENANCE_PERIOD=0
AUTOMATIC_MAINTENANCE_COUNT=0
UNSAFE_QUORUM=true

[[HOME_DOMAINS]]
HOME_DOMAIN="developer-testing.stellar.org"
QUALITY="MEDIUM"

# History archives

# Stellar.org history store
[...]

PR #2696, opened 2020-09-03
Stellar playground enhanced manual close

rokopt:
# Description

The `stellar-core` component of #2695 .

# Checklist
- [x] Compiles
- [x] Ran all tests *(but new tests must be added before this PR comes out of draft)*
- [ ] If change impacts performance, include supporting evidence per the [performance document](https://github.com/stellar/stellar-core/blob/master/performance-eval/performance-eval.md) **(N/A)**

### DIFF
--- a/docs/quick-reference.md
+++ b/docs/quick-reference.md
@@ -78,6 +78,11 @@ The `manualclose` command allows to close a ledger, this is used when the instan
 configured with `MANUAL_CLOSE=true`.
 With this you can test various scenario where ledger boundaries could be important.
 
+If `RUN_STANDALONE=true` along with `MANUAL_CLOSE=true`, then
+the `manualclose` command may also control some of the parameters of the
+ledger that it causes to close.  (The `closeTime` parameter may be no later
+than the first second of the year 2200, GMT.)
+
 On private networks, you can use the `generateload` command to create accounts and generate
 synthetic traffic.
 
--- a/src/herder/HerderImpl.cpp
+++ b/src/herder/HerderImpl.cpp
@@ -318,10 +318,11 @@ HerderImpl::broadcast(SCPEnvelope const& e)
 void
 HerderImpl::startRebroadcastTimer()
 {
-    if (mApp.getConfig().RUN_STANDALONE)
+    if (mApp.getConfig().MANUAL_CLOSE && mApp.getConfig().RUN_STANDALONE)
     {
         return;
     }
+
     mRebroadcastTimer.expires_from_now(std::chrono::seconds(2));
 
     mRebroadcastTimer.async_wait(std::bind(&HerderImpl::rebroadcast, this),
--- a/src/herder/HerderImpl.cpp
+++ b/src/herder/HerderImpl.cpp
@@ -318,6 +318,10 @@ HerderImpl::broadcast(SCPEnvelope const& e)
 void
 HerderImpl::startRebroadcastTimer()
 {
+    if (mApp.getConfig().RUN_STANDALONE)
+    {
+        return;
+    }
     mRebroadcastTimer.expires_from_now(std::chrono::seconds(2));
 
     mRebroadcastTimer.async_wait(std::bind(&HerderImpl::rebroadcast, this),
@@ -1082,6 +1086,13 @@ std::string
 HerderImpl::getUpgradesJson()
 {
     return mUpgrades.getParameters().toJson();
+}
+
+void
+HerderImpl::forceSCPStateIntoSyncWithLastClosedLedger()
+{
+    auto const& header = mLedgerManager.getLastClosedLedgerHeader().header;
+    mHerderSCPDriver.restoreSCPState(header.ledgerSeq, header.scpValue);
 }
 
 bool
--- a/src/main/Application.h
+++ b/src/main/Application.h
@@ -5,6 +5,8 @@
 // of this distribution or at http://www.apache.org/licenses/LICENSE-2.0
 
 #include "main/Config.h"
+#include "util/optional.h"
+#include "xdr/Stellar-ledger-entries.h"
 #include "xdr/Stellar-types.h"
 #include <lib/json/json.h>
 #include <memory>
@@ -240,9 +242,14 @@ class Application
     // returns.
     virtual void joinAllThreads() = 0;
 
-    // If config.MANUAL_MODE=true, force the current ledger to close and return
+    // If config.MANUAL_CLOSE=true, force the current ledger to close and return
     // true. Otherwise return false. This method exists only for testing.
-    virtual bool manualClose() = 0;
+    //
+    // Non-default parameters may be specified only if additionally
+    // config.RUN_STANDALONE=true.
+    virtual std::string
+    manualClose(optional<uint32_t> const& manualLedgerSeq,
+                optional<TimePoint> const& manualCloseTime) = 0;
 
 #ifdef BUILD_TESTS
     // If config.ARTIFICIALLY_GENERATE_LOAD_FOR_TESTING=true, generate some load
--- a/src/main/ApplicationImpl.cpp
+++ b/src/main/ApplicationImpl.cpp
@@ -2,6 +2,7 @@
 // under the Apache License, Version 2.0. See the COPYING file at the root
 // of this distribution or at http://www.apache.org/licenses/LICENSE-2.0
 
+#include <limits>
 #define STELLAR_CORE_REAL_TIMER_FOR_CERTAIN_NOT_JUST_VIRTUAL_TIME
 #include "ApplicationImpl.h"
 
@@ -394,25 +395,27 @@ ApplicationImpl::start()
 
     if (mConfig.FORCE_SCP && !mConfig.NODE_IS_VALIDATOR)
     {
-        LOG(ERROR) << "Starting stellar-core in FORCE_SCP mode requires "
-                      "NODE_IS_VALIDATOR to be set";
-        throw std::invalid_argument("NODE_IS_VALIDATOR not set");
+        throw std::invalid_argument(
+            "FORCE_SCP is set but NODE_IS_VALIDATOR not set");
     }
 
-    if (mConfig.METADATA_OUTPUT_STREAM != "" && mConfig.NODE_IS_VALIDATOR)
+    auto const isNetworkedValidator =
+        mConfig.NODE_IS_VALIDATOR && !mConfig.RUN_STANDALONE;
+
+    if (mConfig.METADATA_OUTPUT_STREAM != "" && isNetworkedValidator)
     {
-        LOG(ERROR) << "Starting stellar-core with METADATA_OUTPUT_STREAM "
-                      "requires NODE_IS_VALIDATOR to be unset";
-        throw std::invalid_argument("NODE_IS_VALIDATOR is set");
+        throw std::invalid_argument(
+            "METADATA_OUTPUT_STREAM is set, NODE_IS_VALIDATOR is set, and "
+            "RUN_STANDALONE is not set");
     }
 
-    if (mConfig.NODE_IS_VALIDATOR)
+    if (isNetworkedValidator)
     {
         if (mConfig.MODE_USES_IN_MEMORY_LEDGER)
         {
-            LOG(ERROR) << "Starting stellar-core with in-memory ledger mode "
-                          "requires NODE_IS_VALIDATOR to be unset";
-            throw std::invalid_argument("NODE_IS_VALIDATOR is set");
+            throw std::invalid_argument(
+                "MODE_USES_IN_MEMORY_LEDGER is set, NODE_IS_VALIDATOR is set, "
+                "and RUN_STANDALONE is not set");
         }
     }
 
@@ -420,10 +423,8 @@ ApplicationImpl::start()
     {
         if (!mConfig.MODE_STORES_HISTORY)
         {
-            LOG(ERROR) << "Starting stellar-core in a mode that does not store "
-                          "history data requires all history archives to be "
-                          "non-writable";
-            throw std::invalid_argument("some history archives are writable");
+            throw std::invalid_argument("MODE_STORES_HISTORY is not set, but "
+                                        "some history archives are writable");
         }
     }
 
@@ -552,27 +553,195 @@ ApplicationImpl::joinAllThreads()
     LOG(DEBUG) << "Joined all " << mWorkerThreads.size() << " threads";
 }
 
-bool
-ApplicationImpl::manualClose()
+std::string
+ApplicationImpl::manualClose(optional<uint32_t> const& manualLedgerSeq,
+                             optional<TimePoint> const& manualCloseTime)
 {
+    assertThreadIsMain();
+
     // Manual close only makes sense for validating nodes
     if (!mConfig.NODE_IS_VALIDATOR)
     {
-        return false;
+        throw std::invalid_argument(
+            "Issuing a manual ledger close requires NODE_IS_VALIDATOR to be "
+            "set to true.");
     }
 
     if (mConfig.MANUAL_CLOSE)
     {
-        mHerder->triggerNextLedger(mLedgerManager->getLastClosedLedgerNum() + 1,
-                                   true);
-        return true;
+        auto const targetLedgerSeq =
+            targetManualCloseLedgerSeqNum(manualLedgerSeq);
+
+        if (mConfig.RUN_STANDALONE)
+        {
+            setManualCloseVirtualTime(manualCloseTime);
+            advanceToLedgerBeforeManualCloseTarget(targetLedgerSeq);
+        }
+        else
+        {
+            // These may only be specified if RUN_STANDALONE is configured.
+            releaseAssert(!manualCloseTime);
+            releaseAssert(!manualLedgerSeq);
+        }
+
+        mHerder->triggerNextLedger(targetLedgerSeq, true);
+
+        if (mConfig.RUN_STANDALONE)
+        {
+            auto const newLedgerSeq =
+                getLedgerManager().getLastClosedLedgerNum();
+            if (newLedgerSeq != targetLedgerSeq)
+            {
+                throw std::runtime_error(fmt::format(
+                    FMT_STRING(
+                        "Standalone manual close failed to produce expected "
+                        "sequence number increment (expected {}; got {})"),
+                    targetLedgerSeq, newLedgerSeq));
+            }
+
+            return fmt::format(
+                FMT_STRING("Manually closed ledger with sequence "
+                           "number {} and closeTime {}"),
+                targetLedgerSeq,
+                getLedgerManager()
+                    .getLastClosedLedgerHeader()
+                    .header.scpValue.closeTime);
+        }
+
+        return fmt::format(
+            FMT_STRING("Manually triggered a ledger close with sequence "
+                       "number {}"),
+            targetLedgerSeq);
     }
     else if (!mConfig.FORCE_SCP)
     {
         mHerder->setInSyncAndTriggerNextLedger();
-        return true;
+        return "Triggered a new consensus round";
     }
-    return false;
+
+    throw std::invalid_argument(
+        "Set MANUAL_CLOSE=true in the stellar-core.cfg if you want to "
+        "close every ledger manually. Otherwise, run stellar-core "
+        "with --wait-for-consensus flag to close ledger once and "
+        "trigger consensus. Ensure NODE_IS_VALIDATOR is set to true.");
+}
+
+uint32_t
+ApplicationImpl::targetManualCloseLedgerSeqNum(
+    optional<uint32_t> const& explicitlyProvidedSeqNum)
+{
+    auto const startLedgerSeq = getLedgerManager().getLastClosedLedgerNum();
+
+    // The "scphistory" stores ledger sequence numbers as INTs.
+    if (startLedgerSeq >= std::numeric_limits<int32_t>::max())
+    {
+        throw std::invalid_argument(fmt::format(
+            FMT_STRING(
+                "Manually closed ledger sequence number ({}) already at max"),
+            startLedgerSeq));
+    }
+
+    auto const nextLedgerSeq = startLedgerSeq + 1;
+
+    if (explicitlyProvidedSeqNum)
+    {
+        if (*explicitlyProvidedSeqNum > std::numeric_limits<int32_t>::max())
+        {
+            // The "scphistory" stores ledger sequence numbers as INTs.
+            throw std::invalid_argument(fmt::format(
+                FMT_STRING("Manual close ledger sequence number {} beyond max"),
+                *explicitlyProvidedSeqNum,
+                std::numeric_limits<int32_t>::max()));
+        }
+
+        if (*explicitlyProvidedSeqNum <= startLedgerSeq)
+        {
+            throw std::invalid_argument(fmt::format(
+                FMT_STRING("Invalid manual close ledger sequence number {} "
+                           "(must exceed current sequence number {})"),
+                *explicitlyProvidedSeqNum, startLedgerSeq));
+        }
+    }
+
+    return explicitlyProvidedSeqNum ? *explicitlyProvidedSeqNum : nextLedgerSeq;
+}
+
+void
+ApplicationImpl::setManualCloseVirtualTime(
+    optional<TimePoint> const& explicitlyProvidedCloseTime)
+{
+    TimePoint constexpr firstSecondOfYear2200GMT = 7'258'118'400ULL;
+
+    static_assert(
+        firstSecondOfYear2200GMT <=
+            std::min(static_cast<TimePoint>(
+                         std::chrono::duration_cast<std::chrono::seconds>(
+                             VirtualClock::system_time_point::duration::max())
+                             .count()),
+                     static_cast<TimePoint>(
+                         std::numeric_limits<std::time_t>::max())) -
+                Herder::MAX_TIME_SLIP_SECONDS.count(),
+        "Documented limit (first second of year 2200 GMT) of manual "
+        "close time would allow runtime overflow in Herder");
+
+    TimePoint const lastCloseTime = getLedgerManager()
+                                        .getLastClosedLedgerHeader()
+                                        .header.scpValue.closeTime;
+    uint64_t const now = VirtualClock::to_time_t(getClock().system_now());
+    uint64_t nextCloseTime = std::max(now, lastCloseTime + 1);
+
+    TimePoint const maxCloseTime = firstSecondOfYear2200GMT;
+
+    if (explicitlyProvidedCloseTime)
+    {
+        if (*explicitlyProvidedCloseTime < nextCloseTime)
+        {
+            throw std::invalid_argument(
+                fmt::format(FMT_STRING("Manual close time {} too early (last "
+                                       "close time={}, now={})"),
+                            *explicitlyProvidedCloseTime, lastCloseTime, now));
+        }
+        nextCloseTime = *explicitlyProvidedCloseTime;
+    }
+    else
+    {
+        // Without an explicitly-provided closeTime, imitate what
+        // HerderImpl::triggerNextLedger() would do in REAL_TIME
+        // (unless that would move the virtual clock backwards).
+        uint64_t const defaultNextCloseTime =
+            VirtualClock::to_time_t(std::chrono::system_clock::now());
+        nextCloseTime = std::max(nextCloseTime, defaultNextCloseTime);
+    }
+
+    if (nextCloseTime > maxCloseTime)
+    {
+        throw std::invalid_argument(
+            fmt::format(FMT_STRING("New close time {} would exceed max ({})"),
+                        nextCloseTime, maxCloseTime));
+    }
+
+    getClock().setCurrentVirtualTime(VirtualClock::from_time_t(nextCloseTime));
+}
+
+void
+ApplicationImpl::advanceToLedgerBeforeManualCloseTarget(
+    uint32_t const& targetLedgerSeq)
+{
+    if (targetLedgerSeq != getLedgerManager().getLastClosedLedgerNum() + 1)
+    {
+        LOG(INFO) << "manually advancing to ledger with sequence number "
+                  << targetLedgerSeq - 1 << " to prepare to close number "
+                  << targetLedgerSeq << " (last ledger to close was number "
+                  << getLedgerManager().getLastClosedLedgerNum() << ")";
+
+        LedgerTxn ltx(getLedgerTxnRoot());
+        ltx.loadHeader().current().ledgerSeq = targetLedgerSeq - 1;
+        getLedgerManager().manuallyAdvanceLedgerHeader(
+            ltx.loadHeader().current());
+        ltx.commit();
+
+        getHerder().forceSCPStateIntoSyncWithLastClosedLedger();
+    }
 }
 
 #ifdef BUILD_TESTS
--- a/src/main/ApplicationUtils.cpp
+++ b/src/main/ApplicationUtils.cpp
@@ -30,6 +30,8 @@ namespace stellar
 int
 runWithConfig(Config cfg, optional<CatchupConfiguration> cc)
 {
+    VirtualClock::Mode clockMode = VirtualClock::REAL_TIME;
+
     if (cfg.MANUAL_CLOSE)
     {
         if (!cfg.NODE_IS_VALIDATOR)
@@ -38,10 +40,27 @@ runWithConfig(Config cfg, optional<CatchupConfiguration> cc)
                           "NODE_IS_VALIDATOR to be set";
             return 1;
         }
+        if (cfg.RUN_STANDALONE)
+        {
+            clockMode = VirtualClock::VIRTUAL_TIME;
+            if (cfg.AUTOMATIC_MAINTENANCE_COUNT != 0 ||
+                cfg.AUTOMATIC_MAINTENANCE_PERIOD.count() != 0)
+            {
+                LOG(WARNING)
+                    << "Using MANUAL_CLOSE and RUN_STANDALONE together "
+                       "induces virtual time, which requires automatic "
+                       "maintenance to be disabled.  "
+                       "AUTOMATIC_MAINTENANCE_COUNT and "
+                       "AUTOMATIC_MAINTENANCE_PERIOD are being overridden to "
+                       "0.";
+                cfg.AUTOMATIC_MAINTENANCE_COUNT = 0;
+                cfg.AUTOMATIC_MAINTENANCE_PERIOD = std::chrono::seconds{0};
+            }
+        }
     }
 
     LOG(INFO) << "Starting stellar-core " << STELLAR_CORE_VERSION;
-    VirtualClock clock(VirtualClock::REAL_TIME);
+    VirtualClock clock(clockMode);
     Application::pointer app;
     try
     {
--- a/src/main/CommandHandler.cpp
+++ b/src/main/CommandHandler.cpp
@@ -164,22 +164,6 @@ CommandHandler::fileNotFound(std::string const& params, std::string& retStr)
         "<p>Have fun!</p>";
 }
 
-void
-CommandHandler::manualClose(std::string const& params, std::string& retStr)
-{
-    if (mApp.manualClose())
-    {
-        retStr = "Triggering a new consensus round";
-    }
-    else
-    {
-        retStr = "Set MANUAL_CLOSE=true in the stellar-core.cfg if you want to "
-                 "close every ledger manually. Otherwise, run stellar-core "
-                 "with --wait-for-consensus flag to close ledger once and "
-                 "trigger consensus. Ensure NODE_IS_VALIDATOR is set to true.";
-    }
-}
-
 template <typename T>
 optional<T>
 maybeParseParam(std::map<std::string, std::string> const& map,
@@ -217,6 +201,28 @@ parseParam(std::map<std::string, std::string> const& map,
         throw std::runtime_error(errorMsg);
     }
     return val;
+}
+
+void
+CommandHandler::manualClose(std::string const& params, std::string& retStr)
+{
+    ZoneScoped;
+    std::map<std::string, std::string> retMap;
+    http::server::server::parseParams(params, retMap);
+
+    if (!retMap.empty() && !mApp.getConfig().RUN_STANDALONE)
+    {
+        throw std::invalid_argument(
+            "The 'manualclose' command accepts parameters only if the "
+            "configuration includes RUN_STANDALONE=true.");
+    }
+
+    uint32_t ledgerSeq;
+    auto manualLedgerSeq = maybeParseParam(retMap, "ledgerSeq", ledgerSeq);
+    TimePoint closeTime;
+    auto manualCloseTime = maybeParseParam(retMap, "closeTime", closeTime);
+
+    retStr = mApp.manualClose(manualLedgerSeq, manualCloseTime);
 }
 
 void
--- a/src/main/test/CommandHandlerTests.cpp
+++ b/src/main/test/CommandHandlerTests.cpp
@@ -2,6 +2,8 @@
 // under the Apache License, Version 2.0. See the COPYING file at the root
 // of this distribution or at http://www.apache.org/licenses/LICENSE-2.0
 
+#include "ledger/LedgerTxn.h"
+#include "ledger/test/LedgerTestUtils.h"
 #include "lib/catch.hpp"
 #include "main/Application.h"
 #include "main/CommandHandler.h"
@@ -12,7 +14,13 @@
 #include "transactions/SignatureUtils.h"
 #include "transactions/TransactionUtils.h"
 #include "util/Decoder.h"
+#include "util/Math.h"
+#include "util/optional.h"
+#include "xdr/Stellar-ledger-entries.h"
+#include "xdr/Stellar-transaction.h"
 #include "xdrpp/marshal.h"
+#include <fmt/format.h>
+#include <stdexcept>
 
 using namespace stellar;
 using namespace stellar::txtest;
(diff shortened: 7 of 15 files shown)


src/herder/HerderImpl.cpp
```diff
@@ -1008,7 +1073,7 @@ HerderImpl::triggerNextLedger(uint32_t ledgerSeqToTrigger,
```
jonjove, line 1073: Are we using `slotIndex` here correctly in the case when we're closing with a distant `ledgerSeq`? (Wasn't allowed to comment on the line I wanted, but it's right above this)

src/main/ApplicationImpl.cpp
```diff
@@ -587,7 +587,7 @@ ApplicationImpl::manualClose(std::map<std::string, std::string> const& retMap,
         retStr = "ENHANCED_MANUAL_CLOSE not enabled";
         return;
     }
-    throw std::runtime_error("Enhanced manual close not implemented yet");
+    mHerder->triggerEnhancedManualCloseLedger(retMap, retStr);
```
jonjove, line 590: This doesn't seem to do all the same things as the other `manualClose`. Ideally these should share the same code path (after all, if `retMap.empty()` then they should be _identical_). You can either remove one, or have one invoke the other. 

Furthermore, as with `Herder::triggerEnhancedManualCloseLedger` this interface is not appropriate here. Handling the key-value pairs should be happening higher up in the call-stack.

jonjove (changes requested):
Some general issues:

- Commit time-stamps are out of order (although GitHub seems to be showing them in the right order... did that behavior change?)
- fdb9fd6f doesn't seem to have anything to do with this PR, so it should be moved to a separate PR

Overall this seems like a good start. I had a couple of comments about how various interfaces were designed. One thing I was pretty unsure about was modifying `triggerNextLedger`, although I didn't see a better path. That function is already too long. Not sure what to do with that.

rokopt:
> * Commit time-stamps are out of order (although GitHub seems to be showing them in the right order... did that behavior change?)

Okay, I've rebased them into order (with `--ignore-date`).  I'm not sure whether GitHub changed anything.

jonjove:
If you liked the old commit order (which I was fine with), then you can also just rewrite the dates. I use a command like `git rebase --force-rebase --exec 'sleep 2' --exec 'git commit --amend --no-edit --date=now' <ref>` and I believe other people on our team have similar commands.

rokopt:
> * [fdb9fd6](https://github.com/stellar/stellar-core/commit/fdb9fd6ffbee628df48d7bcd9077bb5ef2682d0a) doesn't seem to have anything to do with this PR, so it should be moved to a separate PR

Okay, I thought I was going to depend on that, but I ended up not doing so.  I've rebased it away.  It already has PR #2626 .

src/main/ApplicationUtils.cpp
```diff
@@ -39,6 +39,16 @@ runWithConfig(Config cfg)
         }
     }
 
+    if (cfg.ENHANCED_MANUAL_CLOSE &&
+        (!cfg.MANUAL_CLOSE || !cfg.isStandaloneValidator()))
+    {
+        LOG(ERROR)
+            << "Starting stellar-core in ENHANCED_MANUAL_CLOSE mode requires "
+               "MANUAL_CLOSE to be set, NODE_IS_VALIDATOR to be set, and "
+               "RUN_STANDALONE to be unset";
```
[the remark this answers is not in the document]
rokopt: Good catch.  Fixed.

rokopt:
> One thing I was pretty unsure about was modifying `triggerNextLedger`, although I didn't see a better path. That function is already too long. Not sure what to do with that.

Same here!

src/main/ApplicationImpl.cpp
```diff
@@ -578,6 +578,18 @@ ApplicationImpl::manualClose()
     return false;
 }
 
+void
+ApplicationImpl::manualClose(std::map<std::string, std::string> const& retMap,
+                             std::string& retStr)
+{
+    if (!mConfig.ENHANCED_MANUAL_CLOSE)
```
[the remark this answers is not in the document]
rokopt: I ended up doing it the way I did primarily because it means that if we ever decide to change the conditions under which `ENHANCED_MANUAL_CLOSE` is allowed to be set, we'll only have to change those checks, and not have to remember to change anything here.

Paranoia about such things isn't bad, though.  Still, if we _do_ end up changing the conditions, it could be to make them _more_ restrictive, in which case there could be the risk in forgetting to update them here that we later lead someone who notices the inconsistency to decide that the old, more lax requirements are the right ones.  Overall I think I favor leaving it and trusting ourselves to have done the validation correctly when we first set the flag.

rokopt:
> If you liked the old commit order (which I was fine with), then you can also just rewrite the dates. I use a command like `git rebase --force-rebase --exec 'sleep 2' --exec 'git commit --amend --no-edit --date=now' <ref>` and I believe other people on our team have similar commands.

I _think_ I preserved the commit order (I believe I'd have had at least one conflict if I hadn't), but all the commits ended up within the same second.  I do like how that command introduces a visible difference; I'll copy that one down.

src/main/ApplicationImpl.cpp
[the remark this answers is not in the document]
jonjove: I'm not sure you understood quite what I meant (not that I necessarily disagree with your decision, I just want to make sure we're speaking the same language). I'm talking about purely internal usage, like
```
Config cfg;
cfg.ENHANCED_MANUAL_CLOSE = true;
...
```
The checks that you added are only in `runWithConfig` which is only used via the command line.

src/herder/HerderImpl.cpp
```diff
@@ -949,17 +1006,25 @@ HerderImpl::triggerNextLedger(uint32_t ledgerSeqToTrigger,
...
+    {
+        nextCloseTime = minCloseTime;
+    }
 
-    // Ensure we're about to nominate a value with valid close time
-    auto isCtValid =
-        ctValidityOffset(nextCloseTime) == std::chrono::milliseconds::zero();
-
-    if (!isCtValid)
+    auto const enhancedManualClose = mApp.getConfig().ENHANCED_MANUAL_CLOSE;
```
MonsieurNicolas, line 1014: I do not like this:
1. either we have full fidelity with `triggerNextLedger` (so no need to diverge from the existing code), or
2. we hook things up at the `ledgerClose` layer and aim to get partial fidelity
  * we're bypassing SCP but when there is a single value, this may not be that much of a big deal and the goal is to force the close time to take on a specific value anyways

"1" seems to be the best, but for that we need https://github.com/stellar/stellar-core/issues/1760
* any other way will require special handling (what you're doing here, and more) as the close time specified may not be compatible with the current date (when using non virtual time), and we need values to be valid during the SCP round (the part that you missed) - there might be even more special cases, and we don't want to go chase this stuff.
* a potential downside with this approach is that the "close time" specified may or may not be the actual time in the resulting ledger (as we'd be using all rules)

"2" will require extracting parts of `triggerNextLedger` so that we can share the code that calculates the candidate value, but instead of using SCP we would just invoke the close ledger code directly.

Both of those approaches require some special code that ensures that lcl is the "right one" when the ledger seqnum jumps ahead, otherwise SCP won't close that ledger and ledger close will crash (because it increments the ledger number by 1).

src/main/ApplicationImpl.cpp
```diff
@@ -399,20 +399,26 @@ ApplicationImpl::start()
         throw std::invalid_argument("NODE_IS_VALIDATOR not set");
     }
 
-    if (mConfig.METADATA_OUTPUT_STREAM != "" && mConfig.NODE_IS_VALIDATOR)
+    if (mConfig.METADATA_OUTPUT_STREAM != "" && mConfig.isNetworkedValidator())
     {
         LOG(ERROR) << "Starting stellar-core with METADATA_OUTPUT_STREAM "
-                      "requires NODE_IS_VALIDATOR to be unset";
-        throw std::invalid_argument("NODE_IS_VALIDATOR is set");
+                      "requires NODE_IS_VALIDATOR to be unset or "
```
MonsieurNicolas, line 405: I'm not sure why we log before throwing, I'm pretty sure we log the exception anyways, so logging here seems redundant

src/main/ApplicationImpl.cpp
```diff
@@ -399,20 +399,26 @@ ApplicationImpl::start()
         throw std::invalid_argument("NODE_IS_VALIDATOR not set");
     }
 
-    if (mConfig.METADATA_OUTPUT_STREAM != "" && mConfig.NODE_IS_VALIDATOR)
+    if (mConfig.METADATA_OUTPUT_STREAM != "" && mConfig.isNetworkedValidator())
```
MonsieurNicolas, line 402: it seems that this is the only place in the entire code base that we use `isNetworkedValidator`, I think that adding a variable with that name here (and using it for both conditions) avoids introducing a new concept to the code base

src/main/ApplicationUtils.cpp
```diff
@@ -39,6 +39,15 @@ runWithConfig(Config cfg)
         }
     }
 
+    if (cfg.ENHANCED_MANUAL_CLOSE &&
```
MonsieurNicolas, line 42: see my other comment on the new flag: the right checks are already performed if we don't introduce this new config element, so we can remove this

src/main/CommandHandler.cpp
```diff
@@ -167,6 +167,15 @@ CommandHandler::fileNotFound(std::string const& params, std::string& retStr)
 void
 CommandHandler::manualClose(std::string const& params, std::string& retStr)
 {
+    ZoneScoped;
+    std::map<std::string, std::string> retMap;
+    http::server::server::parseParams(params, retMap);
+
+    if (!retMap.empty())
+    {
+        mApp.manualClose(retMap, retStr);
```
MonsieurNicolas, line 176: we should really share the code with manual close (I think this will become evident if we use virtual time).
I would just expect the behavior to be additive: for example, if we need to "jump ahead" the ledger seq num, we would do that before performing the existing "manualClose" behavior, same thing with the close time.

MonsieurNicolas (changes requested):

src/main/ApplicationImpl.cpp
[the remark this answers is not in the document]
rokopt: Oh, I see.  Yes, I did fail to gather that.

I think I am inclined to let purely internal users (potentially) shoot themselves in the foot arbitrarily, but it's not a strong preference so I would be fine with going the other way if you'd rather.

src/herder/HerderImpl.cpp
```diff
@@ -917,11 +918,67 @@ HerderImpl::setInSyncAndTriggerNextLedger()
...
+    auto const minCloseTime =
+        mLedgerManager.getLastClosedLedgerHeader().header.scpValue.closeTime +
+        1;
+    auto closeTime = minCloseTime;
+
+    try
+    {
+        setNumericParam(
+            ledgerSeq, "ledgerSeq",
+            [](std::string stringVal) { return stoul(stringVal); },
```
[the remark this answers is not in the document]
rokopt: Also a good catch.  When I started writing that function, I meant to compensate for the lack of a `std::stou()`, but by the time I got there, I'd forgotten.  Fixed (I hope in a sane way).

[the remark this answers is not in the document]
MonsieurNicolas: This code is redundant - if you do the parsing in the http handler, you can use things like `maybeParseParam` that performs the right checks

src/herder/HerderImpl.cpp
rokopt: Hmm, I'm not sure.  From my first reading I thought that when we're in `RUN_STANDALONE` mode, it sufficed for `slotIndex` to be increasing, but maybe it does have to match the new ledger sequence number.

I've changed it to be the latter because that now looks more plausible to me, but I'm still not sure what are all the ways that `slotIndex` is used.  This is probably a question for @MonsieurNicolas : if an enhanced manual ledger close increases the sequence number by more than `1`, should the `slotIndex` be one greater than the old sequence number, or equal to the new, even greater sequence number?  Or does it require subtler handling than either of those extremes?

src/herder/HerderImpl.cpp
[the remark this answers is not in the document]
rokopt: Thanks, `maybeParseParam()` is nice.  I'm using that now that I've moved the argument-parsing into `CommandHandler.cpp`.

src/herder/HerderImpl.cpp
```diff
@@ -917,11 +918,67 @@ HerderImpl::setInSyncAndTriggerNextLedger()
     triggerNextLedger(lcl + 1, false);
 }
 
+void
+HerderImpl::triggerEnhancedManualCloseLedger(
+    std::map<std::string, std::string> const& retMap, std::string& retStr)
```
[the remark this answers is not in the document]
rokopt: Makes sense.  I've moved the argument parsing into `CommandHandler.cpp`.

src/main/ApplicationImpl.cpp
rokopt: I've moved the key-value-pair handling to `CommandHandler.cpp`, and used the resulting restructuring to unify the `manualClose()` and `triggerNextLedger()` variants into single interfaces.

src/main/CommandHandler.cpp
rokopt: If I understand what you're suggesting correctly, then this is done now.  `Application::manualClose()` is now a single interface with optional parameters for the sequence number and close time, and it's called in just one place from `CommandHandler::manualClose()`.

src/main/ApplicationImpl.cpp
rokopt: Alright, I've done that.

src/main/ApplicationUtils.cpp
rokopt: I've removed `isStandaloneValidator()` as well.

src/main/ApplicationImpl.cpp
rokopt: Good question.  It was like that already; do you want me to change it?

src/main/ApplicationUtils.cpp
rokopt: And now I've removed `ENHANCED_MANUAL_CLOSE` altogether.

docs/quick-reference.md
```diff
@@ -78,6 +78,11 @@ The `manualclose` command allows to close a ledger, this is used when the instan
 configured with `MANUAL_CLOSE=true`.
 With this you can test various scenario where ledger boundaries could be important.
 
+Setting `ENHANCED_MANUAL_CLOSE=true` (along with `MANUAL_CLOSE=true`) further
```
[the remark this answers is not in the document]
rokopt: Alright, I took it out.

I do still validate that we're in `RUN_STANDALONE` before allowing the specification of any parameters to the `manualclose` HTTP command.  Does that make sense?  I don't think we'd want one validator in a network to be able to try to push forward the close time or ledger sequence number arbitrarily, would we?

src/herder/HerderImpl.cpp
rokopt: I thought that enforcing `RUN_STANDALONE` would allow us to set the sequence number forward arbitrarily since there would be no nodes proposing different ones.  I didn't realize that it was automatic that each close just incremented the number.

I _think_ we want advancing the sequence number to be one of the things that we're letting people test, right?  Are you saying we'll need to change the sequence number in the header in `triggerNextLedger()` if it jumps ahead because of a manual request, to satisfy SCP?

Are you also saying that we should force the use of virtual time when MANUAL_CLOSE is specified (issue #1760) as part of the work for this issue?

src/main/ApplicationUtils.cpp
```diff
@@ -40,7 +40,8 @@ runWithConfig(Config cfg)
     }
 
     LOG(INFO) << "Starting stellar-core " << STELLAR_CORE_VERSION;
-    VirtualClock clock(VirtualClock::REAL_TIME);
+    VirtualClock clock(cfg.MANUAL_CLOSE ? VirtualClock::VIRTUAL_TIME
```
jonjove, line 43: This doesn't seem to implement everything in #1760.

src/herder/HerderImpl.cpp
```diff
@@ -949,8 +949,16 @@ HerderImpl::triggerNextLedger(optional<uint32_t> ledgerSeqToTrigger,
     // We pick as next close time the current time unless it's before the last
     // close time. We don't know how much time it will take to reach consensus
     // so this is the most appropriate value to use as closeTime.
+    auto const usingManualCloseWithoutExplicitCloseTime =
```
jonjove, line 953: Nitpick: `bool` and `auto` are the same length, why not be explicit?

src/herder/HerderImpl.cpp
```diff
@@ -949,8 +949,16 @@ HerderImpl::triggerNextLedger(optional<uint32_t> ledgerSeqToTrigger,
...
+        mApp.getConfig().MANUAL_CLOSE && !minCloseTime;
+    auto static const manualCloseDefaultPerLedgerTimeTIncrement =
```
jonjove, line 955: I'm not really understanding this constant. The `closeTime` has units of seconds, so you can just add 5. You don't need to convert it to a `time_t` or a `time_point` (which is not a duration).

Nitpick: We don't generally use local `static` variables, and I don't see any serious need for it here. Also can we change the name? It's a bit of an eye-sore and doesn't help me understand what it means.

src/main/ApplicationUtils.cpp
jonjove: Ah I see, it's in the next commit. Probably should squash them.

src/herder/HerderImpl.cpp
```diff
@@ -949,8 +949,16 @@ HerderImpl::triggerNextLedger(optional<uint32_t> ledgerSeqToTrigger,
...
+        VirtualClock::to_time_t(
+            std::chrono::system_clock::time_point(std::chrono::seconds{5}));
     uint64_t nextCloseTime =
-        VirtualClock::to_time_t(mApp.getClock().system_now());
+        usingManualCloseWithoutExplicitCloseTime
```
jonjove, line 959: More generally, does this code even need to be in `triggerNextLedger` or could it be in `manualClose` (transmitted via the `minTime` parameter)?

src/herder/HerderImpl.cpp
```diff
@@ -1033,7 +1034,16 @@ HerderImpl::triggerNextLedger(optional<uint32_t> ledgerSeqToTrigger,
     }
     if (*ledgerSeqToTrigger > slotIndex)
     {
+        if (!usingManualClose)
+        {
+            throw std::logic_error(
+                "Allowing sequence number jump without MANUAL_CLOSE");
+        }
         slotIndex = *ledgerSeqToTrigger;
+        LedgerTxn updateLastSeqnoToPredecessor(mApp.getLedgerTxnRoot());
```
jonjove, line 1043: Nitpick: perhaps just `ltx` (that's the name we conventionally use everywhere) unless you have some compelling reason to have a long detailed name?

jonjove (commented):
A few new comments, mostly cosmetic. Squashing the MANUAL_CLOSE commits into one would probably add clarity for reviewers.

src/herder/HerderImpl.cpp
```diff
@@ -914,13 +918,15 @@ HerderImpl::setInSyncAndTriggerNextLedger()
...
-    triggerNextLedger(lcl + 1, false);
+    triggerNextLedger(make_optional<uint32_t>(lcl + 1), nullopt<TimePoint>(),
+                      false);
 }
 
 // called to take a position during the next round
 // uses the state in LedgerManager to derive a starting position
 void
-HerderImpl::triggerNextLedger(uint32_t ledgerSeqToTrigger,
+HerderImpl::triggerNextLedger(optional<uint32_t> ledgerSeqToTrigger,
```
MonsieurNicolas, line 928: It looks that from the conversation https://github.com/stellar/stellar-core/pull/2696#discussion_r483277469 you went with the "virtual time" option.

In this case, you don't actually need to (or want to) change `triggerNextLedger` that should just work "out of the box" as to provide full fidelity; in fact I don't even think that you need to make any modification to herder and just make sure that you setup the "initial state" properly:

* if a close time is specified by `manualclose`, you should just "jump time ahead" (see `setCurrentVirtualTime`)
* if a ledger seq num is specified (and it requires "jumping ahead" from a ledger point of view), this should be done outside of `triggerNextLedger` by directly adjusting lcl

src/main/ApplicationImpl.cpp
```diff
@@ -549,27 +558,109 @@ ApplicationImpl::joinAllThreads()
...
+        {
+            throw std::invalid_argument(
+                fmt::format(FMT_STRING("May not skip ledgers (requested "
+                                       "sequence number {}; next {}) while "
+                                       "transaction queue is non-empty"),
+                            targetLedgerSeq, minLedgerSeq));
+        }
+        auto nextLedgerSeqToTrigger = startLedgerSeq;
+        bool lastClose;
+        do
```
MonsieurNicolas, line 604: This loop seems overly complicated, why not just closing one ledger like before?

I think all you need to make this work is ensure that we meet the conditions so that "herder" can trigger the ledger properly (what I mentioned twice already in different comments).

Glancing at the code, I think that all you need is to call `restoreSCPState` with the right `index`/`StellarValue` when you modify lcl (NB: from `StellarValue`, we only use the close time).

MonsieurNicolas (changes requested):

src/herder/HerderImpl.cpp
rokopt: I'm hoping that this has been addressed in the latest implementation, which reverts to using `triggerNextLedger()` only to increment the sequence number by `1` (it currently uses `advanceLedgerPointers()` to do the big jump).

src/herder/HerderImpl.cpp
rokopt: I'm now changing the sequence number in a separate modification before `triggerNextLedger()`, and I am indeed using `VIRTUAL_TIME` when `MANUAL_CLOSE` is specified, as suggested by issue #1760 .

src/main/ApplicationImpl.cpp
rokopt: Done.

src/main/ApplicationUtils.cpp
rokopt: Squashed.

src/herder/HerderImpl.cpp
rokopt: That sounds fine to me, though as it turns out, later changes have eliminated that code altogether. :)

src/herder/HerderImpl.cpp
rokopt: The code that implements the five-second default has been moved (to `ApplicationImpl::manualClose()`), and it no longer uses any conversions (the constant is just a `5`).

src/herder/HerderImpl.cpp
rokopt: It has, indeed, since been moved out of `triggerNextLedger()` and into `manualClose()`.

src/herder/HerderImpl.cpp
rokopt: The code that did this has moved; it's called `ltx` there now.

src/herder/HerderImpl.cpp
rokopt: Yes, you're right, once I went to `VIRTUAL_TIME`, the seeming need to modify `triggerNextLedger()` melted away.  :)

The new implementation is  in `ApplicationImpl::manualClose()`, and it does use `setCurrentVirtualTime()` and explicit modification of `loadHeader().current()`.

src/main/ApplicationImpl.cpp
rokopt: >This loop seems overly complicated, why not just closing one ledger like before?

> I think all you need to make this work is ensure that we meet the conditions so that "herder" can trigger the ledger properly (what I mentioned twice already in different comments).

It was because I hadn't figured out how to meet the conditions to skip ledger numbers yet.  I wanted to see if I could get _something_ working first.

> Glancing at the code, I think that all you need is to call restoreSCPState with the right index/StellarValue when you modify lcl (NB: from StellarValue, we only use the close time).

Currently I'm taking Jon's suggestion of using (a slight variation on) `LedgerManager::advanceLedgerPointers()`, and thus far it appears to be working, but I can try the `restoreSCPState()` approach if that's better (maybe with just `advanceLedgerPointers()` there's something that's supposed to be happening that isn't?).

src/main/ApplicationImpl.cpp
```diff
@@ -549,27 +558,109 @@ ApplicationImpl::joinAllThreads()
...
+        auto const targetLedgerSeq =
+            manualLedgerSeq ? *manualLedgerSeq : minLedgerSeq;
+        if (targetLedgerSeq < minLedgerSeq)
+        {
+            throw std::invalid_argument(
+                fmt::format(FMT_STRING("Invalid manual close ledger "
+                                       "sequence number {} (minimum {})"),
+                            targetLedgerSeq, minLedgerSeq));
+        }
+        if (!mHerder->isTransactionQueueEmpty() &&
```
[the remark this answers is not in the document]
rokopt: > Why do we need this check? Because we already use the normal code when triggering ledgers we'll do the right thing with any potential transaction that gets invalidated by the new close time / ledger seq

When we manually close a ledger, there might be more transactions in the queue than would fit in a single ledger.  So to skip ledgers in that state, we have to close ledgers one at a time until the queue is empty, to get the right semantics regarding which transactions get into which ledgers (which may have further consequences).

So if we want to skip ahead, we have to choose where to do the one-by-one closing to empty the transaction queue first:  the client (usually Horizon), or the core.  The core could do it, or it could return an error and leave it to the client.  I think that it'd be basically the same code in either place, and I gather it's a generally-agreed principle (it would certainly make sense to me) that, all other things being equal, complexity is better outside `stellar-core` than inside it, so I went with leaving it to the client.

I could stick the loop in `stellar-core` itself if you prefer.

src/main/ApplicationImpl.cpp
rokopt: > Glancing at the code, I think that all you need is to call restoreSCPState with the right index/StellarValue when you modify lcl (NB: from StellarValue, we only use the close time).

Although some tests were working before using just Jon's `advanceLedgerPointers()` suggestion, there is at least this interface-violating way of calling `restoreSCPState()` as well, in case that is also necessary:  c23a2d05d370d2a67ff4e43cf3c33df9b9b6eb2c

For now, I've left that in a separate commit for independent review, since I'm not sure whether it's necessary or correct, and even if it's both, it should probably be exposed via some new Herder-interface abstract method, though I'm not sure with exactly what parameters or what to call it.

rokopt:
I think that I've answered the pending review comments as of now, so that this is ready for another round of reviews.

src/main/ApplicationImpl.cpp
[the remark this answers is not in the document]
MonsieurNicolas: not really: this is an arbitrary rule that should be removed.

In a real deployment of stellar-core, the decisions made wrt the transaction queue by the local node are pretty much independent of whatever happens during consensus. In fact this kind of thing happens pretty often as nodes on the network are not closing ledgers at exactly the same time but still accept/flood transactions.

Like you said, we should leave the decision on what to simulate to the caller, and this rule gets in the way.

src/main/ApplicationImpl.cpp
```diff
@@ -549,27 +550,123 @@ ApplicationImpl::joinAllThreads()
...
+                      << targetLedgerSeq - 1 << " on behalf of target "
+                      << targetLedgerSeq;
+            LedgerTxn ltx(getLedgerTxnRoot());
+            ltx.loadHeader().current().ledgerSeq = targetLedgerSeq - 1;
+            getLedgerManager().manuallyAdvanceLedgerHeader(
+                ltx.loadHeader().current());
+            ltx.commit();
+            auto const& header =
+                getLedgerManager().getLastClosedLedgerHeader().header;
+            dynamic_cast<HerderImpl&>(getHerder())
```
MonsieurNicolas, line 635: yes we need to do this.
I'd suggest adding a new method to Herder to expose this in a slightly more sane way and with a name that yells "unsafe".

Maybe something like `forceSyncedState()` (no parameters) that calls `restoreSCPState` with data gathered from lcl.

src/main/ApplicationImpl.cpp
```diff
@@ -549,27 +550,123 @@ ApplicationImpl::joinAllThreads()
...
+                .getHerderSCPDriver()
+                .restoreSCPState(header.ledgerSeq, header.scpValue);
+        }
+        if (now < nextCloseTime)
+        {
+            getClock().setCurrentVirtualTime(
+                VirtualClock::time_point(std::chrono::seconds(nextCloseTime)));
+        }
+        mHerder->triggerNextLedger(targetLedgerSeq, true);
+        while (getClock().crank(true) != 0 &&
```
MonsieurNicolas, line 645: This doesn't look right:
we should not call `crank` from here as this code is already being called from the main `crank` and I don't think we make any guarantees that this function is re-entrant.

Because `triggerNextLedger` posts an event, things should "just work", did you have any problems without this code?

MonsieurNicolas (changes requested):

src/main/ApplicationImpl.cpp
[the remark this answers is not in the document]
rokopt: > In a real deployment of stellar-core, the decisions made wrt the transaction queue by the local node are pretty much independent of whatever happens during consensus. In fact this kind of thing happens pretty often as nodes on the network are not closing ledgers at exactly the same time but still accept/flood transactions.

This is what I had failed to understand:  we might close a bunch of ledgers without taking any transactions from our queue, as it's consensus that dictates which transactions get in.  I had been stuck in "standalone mode" myself, thinking that the node would deterministically take some given set of transactions from its queue on each close.

I've removed `isTransactionQueueEmpty()`.

src/main/ApplicationImpl.cpp
```diff
@@ -549,27 +550,123 @@ ApplicationImpl::joinAllThreads()
...
+            throw std::invalid_argument(
+                fmt::format(FMT_STRING("Invalid manual close ledger "
+                                       "sequence number {} (minimum {})"),
+                            targetLedgerSeq, minLedgerSeq));
+        }
+        auto const lastCloseTime = getLedgerManager()
+                                       .getLastClosedLedgerHeader()
+                                       .header.scpValue.closeTime;
+        auto const now = VirtualClock::to_time_t(getClock().system_now());
+        uint32_t const defaultManualCloseTimeInterval = 5;
```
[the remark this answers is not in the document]
rokopt: Thanks, I didn't know we had that.  I'm using it now.

src/main/ApplicationImpl.cpp
```diff
@@ -549,27 +550,123 @@ ApplicationImpl::joinAllThreads()
...
+        auto nextCloseTime = lastCloseTime + 1;
+        auto const defaultNextCloseTime =
```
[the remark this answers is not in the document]
rokopt: Good point, done.

src/main/ApplicationImpl.cpp
rokopt: I've given it the descriptive-but-hostile-sounding name `forceSCPStateIntoSyncWithLastClosedLedger()`, though I'll shorten it to `forceSyncedState()` (or whatever) if you prefer.

src/main/ApplicationImpl.cpp
rokopt: > did you have any problems without this code?

No, it _does_ just work.  This was me not adequately understanding how `VIRTUAL_TIME` worked.  I thought we'd have to manually `crank` to process the event(s) posted by `triggerNextLedger()` and the rest of the SCP processing resulting therefrom.  I've removed the `crank` loop.

jonjove (commented):
Once `manualClose` has been refactored, I'll  want to review again to make sure you're covering all the right edge cases.

src/main/ApplicationImpl.cpp
```diff
@@ -549,27 +549,109 @@ ApplicationImpl::joinAllThreads()
...
+            "Issuing a manual ledger close requires NODE_IS_VALIDATOR to be "
+            "set to true.");
     }
 
     if (mConfig.MANUAL_CLOSE)
     {
-        mHerder->triggerNextLedger(mLedgerManager->getLastClosedLedgerNum() + 1,
-                                   true);
-        return true;
+        auto const startLedgerSeq = getLedgerManager().getLastClosedLedgerNum();
```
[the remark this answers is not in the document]
rokopt: Good idea; I've factored out three private utility methods from `ApplicationImpl::manualClose()`.

src/main/ApplicationImpl.cpp
```diff
@@ -549,27 +549,109 @@ ApplicationImpl::joinAllThreads()
...
+                                       "sequence number {} (minimum {})"),
+                            targetLedgerSeq, minLedgerSeq));
+        }
+        auto const lastCloseTime = getLedgerManager()
+                                       .getLastClosedLedgerHeader()
+                                       .header.scpValue.closeTime;
+        auto const now = VirtualClock::to_time_t(getClock().system_now());
+        uint32_t const defaultManualCloseTimeInterval =
+            getConfig().getExpectedLedgerCloseTime().count();
+        auto nextCloseTime = lastCloseTime + 1;
```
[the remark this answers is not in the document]
rokopt: It turns out it's messier than either:  `HerderImpl::triggerNextLedger()` uses `VirtualClock::to_time_t()` on the `closeTime`, so we need to do the same to validate against the last closed header's `closeTime`, and possibly to produce a `closeTime` derived from that.  Then we convert it back to a `system_time_point` to call `setCurrentVirtualTime()`.  So a `closeTime` that overflows is one that isn't stable under all that converting.  That's nine-point-something billion.  I think that the best way to check for that is to do that stability test itself, so that's what the code does now.

src/main/ApplicationImpl.cpp
```diff
@@ -549,27 +549,109 @@ ApplicationImpl::joinAllThreads()
...
+            auto const defaultNextCloseTime =
+                lastCloseTime + defaultManualCloseTimeInterval;
+            if (nextCloseTime < defaultNextCloseTime)
+            {
+                nextCloseTime = defaultNextCloseTime;
+            }
+        }
+        if (targetLedgerSeq != minLedgerSeq)
+        {
+            LOG(INFO) << "closing ledger with sequence number "
```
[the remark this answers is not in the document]
rokopt: Yeah, I don't like that one either.  I've changed it.

src/main/ApplicationImpl.cpp
```diff
@@ -549,27 +549,109 @@ ApplicationImpl::joinAllThreads()
...
+            "set to true.");
     }
 
     if (mConfig.MANUAL_CLOSE)
     {
-        mHerder->triggerNextLedger(mLedgerManager->getLastClosedLedgerNum() + 1,
-                                   true);
-        return true;
+        auto const startLedgerSeq = getLedgerManager().getLastClosedLedgerNum();
+        if (startLedgerSeq == std::numeric_limits<uint32_t>::max())
```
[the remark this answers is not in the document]
rokopt: Right, I noticed in testing that the `scphistory` database crashes on such values because it stores the ledger number as an `INT`.  I've updated the validation to reflect that.

src/main/ApplicationImpl.cpp
```diff
@@ -549,27 +549,109 @@ ApplicationImpl::joinAllThreads()
...
+                            targetLedgerSeq, minLedgerSeq));
+        }
+        auto const lastCloseTime = getLedgerManager()
+                                       .getLastClosedLedgerHeader()
+                                       .header.scpValue.closeTime;
+        auto const now = VirtualClock::to_time_t(getClock().system_now());
+        uint32_t const defaultManualCloseTimeInterval =
+            getConfig().getExpectedLedgerCloseTime().count();
+        auto nextCloseTime = lastCloseTime + 1;
+        if (nextCloseTime < now)
```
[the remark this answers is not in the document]
rokopt: Done.

src/main/ApplicationImpl.cpp
```diff
@@ -549,27 +549,109 @@ ApplicationImpl::joinAllThreads()
...
+                throw std::invalid_argument(
+                    fmt::format(FMT_STRING("Invalid manual close time {} (last "
+                                           "close time={}, now={})"),
+                                *manualCloseTime, lastCloseTime, now));
+            }
+            nextCloseTime = *manualCloseTime;
+        }
+        else
+        {
+            auto const defaultNextCloseTime =
```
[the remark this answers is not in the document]
rokopt: The code to check for the weird overflow condition is now unified between those two cases.

src/main/ApplicationImpl.cpp
```diff
@@ -549,27 +549,109 @@ ApplicationImpl::joinAllThreads()
...
+                lastCloseTime + defaultManualCloseTimeInterval;
+            if (nextCloseTime < defaultNextCloseTime)
```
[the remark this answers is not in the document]
rokopt: Done.

src/main/ApplicationImpl.cpp
```diff
@@ -549,27 +549,109 @@ ApplicationImpl::joinAllThreads()
...
+        auto const newLedgerSeq = getLedgerManager().getLastClosedLedgerNum();
+        if (newLedgerSeq != targetLedgerSeq)
+        {
+            throw std::runtime_error(fmt::format(
+                FMT_STRING("Manual close failed to produce expected "
+                           "sequence number increment (expected {}; got {})"),
+                targetLedgerSeq, newLedgerSeq));
+        }
+        return fmt::format(
+            FMT_STRING("Closed ledgers up to sequence number {}"),
```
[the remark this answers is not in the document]
rokopt: True.  I've changed it.

src/main/test/CommandHandlerTests.cpp
```diff
@@ -137,3 +139,106 @@ TEST_CASE("transaction envelope bridge", "[commandhandler]")
         });
     }
 }
+
+TEST_CASE("manualclose", "[commandhandler]")
```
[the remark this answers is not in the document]
rokopt: I've added tests for those.  (I still have some more to write, such as to validate that the combinations of config parameters that I expect to be rejected are indeed rejected, but I think those tests should just be additions; the existing test code is reviewable now, as far as I know.  I have validated the config combinations manually with `curl`.)

rokopt:
I've dealt with code review comments again, and squashed commits.

I've also added a separate commit, d04230d1384901759d88b5878f0465334fe16dc9 , after discovering that when I would start `stellar-core` with `MANUAL_CLOSE` and `RUN_STANDALONE`, it would consume 100% of CPU time, and the virtual "now" would increase continually.  That turned out to be because the Herder's rebroadcast timer would continually fire and re-set itself when in virtual time with the clock being continually cranked.

src/herder/HerderImpl.cpp
```diff
@@ -318,6 +318,10 @@ HerderImpl::broadcast(SCPEnvelope const& e)
 void
 HerderImpl::startRebroadcastTimer()
 {
+    if (mApp.getConfig().MANUAL_CLOSE && mApp.getConfig().RUN_STANDALONE)
```
jonjove, line 321: More generally, why do we do this at all if we are running with `RUN_STANDALONE`? There is no one to receive the broadcast.

jonjove (commented):
Getting very close now.

src/herder/HerderImpl.cpp
rokopt: I wondered if it might be okay to change that regardless of `MANUAL_CLOSE`.  I've done so now.

src/main/ApplicationImpl.cpp
```diff
@@ -549,27 +550,177 @@ ApplicationImpl::joinAllThreads()
...
+        "with --wait-for-consensus flag to close ledger once and "
+        "trigger consensus. Ensure NODE_IS_VALIDATOR is set to true.");
+}
+
+uint32_t
+ApplicationImpl::targetManualCloseLedgerSeqNum(
+    optional<uint32_t> const& userSpecifiedSeqNum)
+{
+    auto const startLedgerSeq = getLedgerManager().getLastClosedLedgerNum();
+    if (startLedgerSeq == std::numeric_limits<uint32_t>::max())
```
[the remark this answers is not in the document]
rokopt: > If lcl.ledgerSeq > INT32_MAX, then haven't we already had an issue with the scphistory table?

Yes, we would have.

> Why not just check [ ... ]

I've now done effectively that, though I've broken it up into separate tests so as to provide different log messages for each.

src/main/ApplicationImpl.cpp
```diff
@@ -549,27 +550,177 @@ ApplicationImpl::joinAllThreads()
...
+    }
+    throw std::invalid_argument(
+        "Set MANUAL_CLOSE=true in the stellar-core.cfg if you want to "
+        "close every ledger manually. Otherwise, run stellar-core "
+        "with --wait-for-consensus flag to close ledger once and "
+        "trigger consensus. Ensure NODE_IS_VALIDATOR is set to true.");
+}
+
+uint32_t
+ApplicationImpl::targetManualCloseLedgerSeqNum(
```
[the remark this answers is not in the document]
rokopt: I've broken the code up into more "paragraphs".  Please let me know if it reads well to you now.

src/main/ApplicationImpl.cpp
```diff
@@ -549,27 +550,177 @@ ApplicationImpl::joinAllThreads()
...
+}
+
+void
+ApplicationImpl::setManualCloseVirtualTime(
+    optional<TimePoint> const& userSpecifiedCloseTime)
+{
+    uint64_t const lastCloseTime = getLedgerManager()
+                                       .getLastClosedLedgerHeader()
+                                       .header.scpValue.closeTime;
+    if (lastCloseTime == std::numeric_limits<uint64_t>::max())
```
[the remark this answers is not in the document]
rokopt: I'm now using `std::numeric_limits<std::time_t>::max()`.

And there's more:

- The overflow that I was actually running into was not of `time_t`, but of `system_time_point` in units of `std::chrono::seconds`.  I'm now checking against that as well as against `std::numeric_limits<std::time_t>::max()`.

- `HerderImpl::triggerNextLedger()` does validation that involves adding `Herder::MAX_TIME_SLIP_SECONDS` to the candidate close time.  So I have to subtract that from the minimum limit imposed by the `time_t` and `system_time_point` types to obtain the limit against which I need to validate.  I'm now doing that, and testing for it in the `"manualclose"` `TEST_CASE` in `CommandHandlerTests.cpp`.

src/main/ApplicationImpl.cpp
```diff
@@ -549,27 +550,224 @@ ApplicationImpl::joinAllThreads()
...
+            throw std::invalid_argument(
+                fmt::format(FMT_STRING("Invalid manual close time {} (last "
+                                       "close time={}, now={})"),
+                            *userSpecifiedCloseTime, lastCloseTime, now));
+        }
+        nextCloseTime = *userSpecifiedCloseTime;
+    }
+    else
+    {
+        uint64_t const defaultManualCloseTimeInterval =
```
MonsieurNicolas, line 732: just realized that this is a breaking change:
* existing `manualclose` users are going to see a clock from the 1970s instead of a current clock, this may break clients that are setting time bounds in their transactions
* a non breaking change would be to set `defaultNextCloseTime` to the current *real* clock's value (so calling `std::chrono::steady_clock::now()` directly)

src/main/ApplicationImpl.cpp
```diff
@@ -549,27 +550,224 @@ ApplicationImpl::joinAllThreads()
...
+            .count();
+    TimePoint const maxTimeT = std::numeric_limits<std::time_t>::max();
+    TimePoint const maxCloseTime =
+        std::min(maxSystemTimePointSeconds, maxTimeT) -
+        Herder::MAX_TIME_SLIP_SECONDS.count();
+
+    TimePoint const lastCloseTime = getLedgerManager()
+                                        .getLastClosedLedgerHeader()
+                                        .header.scpValue.closeTime;
+    if (lastCloseTime >= maxCloseTime)
```
MonsieurNicolas, line 689: Why do we care to compare `lastCloseTime` to the max? We should just check that the new time is within bounds and avoid duplicate code?
(this is true for both checks)

src/main/ApplicationImpl.cpp
```diff
@@ -549,27 +550,224 @@ ApplicationImpl::joinAllThreads()
...
+                                   "close further ledgers"),
+                        maxCloseTime));
+    }
+
+    uint64_t const now = VirtualClock::to_time_t(systemNow);
+    uint64_t nextCloseTime = std::max(now, lastCloseTime + 1);
+
+    if (userSpecifiedCloseTime)
+    {
+        if (*userSpecifiedCloseTime > maxCloseTime)
```
MonsieurNicolas, line 715: I think we should just move those checks before calling `setCurrentVirtualTime` (and tweak the wording a bit as it's not necessarily provided by the user) - that way we perform the checks on the validity of `nextCloseTime` regardless of default behavior or not.

MonsieurNicolas (changes requested):
Looks pretty close to me as well. I have a few final suggestions to cleanup/simplify the code.

src/main/Application.h
```diff
@@ -5,6 +5,8 @@
 // of this distribution or at http://www.apache.org/licenses/LICENSE-2.0
 
 #include "main/Config.h"
+#include "util/optional.h"
+#include "xdr/Stellar-ledger-entries.h"
```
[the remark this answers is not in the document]
rokopt: `TimePoint`.

src/main/ApplicationImpl.cpp
```diff
@@ -549,27 +550,224 @@ ApplicationImpl::joinAllThreads()
...
+        if (mConfig.RUN_STANDALONE)
+        {
+            setManualCloseVirtualTime(manualCloseTime);
+            advanceToLedgerBeforeManualCloseTarget(targetLedgerSeq);
+        }
+        else
+        {
+            if (manualCloseTime)
+            {
+                throw std::logic_error("Specifying manual close time should "
```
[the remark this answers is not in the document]
rokopt: I don't think so -- if any options had been specified without `RUN_STANDALONE` being configured, then `CommandHandler::manualClose()` should have caught it:

```
    if (!retMap.empty() && !mApp.getConfig().RUN_STANDALONE)
    {
        throw std::invalid_argument(
            "The 'manualclose' command accepts parameters only if the "
            "configuration includes RUN_STANDALONE=true.");
    }
```

But `ApplicationImpl::manualClose()` is in a different class, so I figured that it should make it clear to any other potential users that they must also avoid providing parameters if `RUN_STANDALONE` isn't configured.

src/main/test/CommandHandlerTests.cpp
```diff
@@ -137,3 +144,366 @@ TEST_CASE("transaction envelope bridge", "[commandhandler]")
...
+{
+    auto testManualCloseConfig = [](auto configure, auto issue) {
```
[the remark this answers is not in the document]
rokopt: You're welcome.  Have another.  ;)  (I added a mini-stress test.)

src/main/ApplicationImpl.cpp
rokopt: The salient part of the `lastCloseTime` comparison is with `std::numeric_limits<TimePoint>::max()` -- maybe I should just test for equality with that.  If I don't, then the `lastCloseTime + 1` later might overflow, and that happens before we can check the new time.  With `secondsSinceEpoch`, that's making sure that `VirtualClock::to_time_t(systemNow)` won't overflow, which as Jon pointed out, could be undefined behavior.  So I might be able to change at least the first check to be against a different maximum (but I figured that as long as I was checking it, I may as well check it against the strictest thing that I could, as there would be no point continuing if it failed that check anyway).  Can I actually avoid either of these somehow without risking either of those overflows?

src/main/ApplicationImpl.cpp
```diff
@@ -549,27 +550,224 @@ ApplicationImpl::joinAllThreads()
...
+void
+ApplicationImpl::setManualCloseVirtualTime(
+    optional<TimePoint> const& userSpecifiedCloseTime)
+{
+    TimePoint const maxSystemTimePointSeconds =
+        std::chrono::duration_cast<std::chrono::seconds>(
+            VirtualClock::system_time_point::duration::max())
+            .count();
+    TimePoint const maxTimeT = std::numeric_limits<std::time_t>::max();
+    TimePoint const maxCloseTime =
```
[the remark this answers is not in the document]
rokopt: > So instead, we should just pick a sensible constant easy to document for maxCloseTime like 1/1/2200 00:00:00 (the c++ spec says "at least 292 years" for second resolution, so we're well within that range relative to epoch).

Our code seems to fall short of the standard -- the value that produces overflow, `9'223'371'977`, is in the year 2262.  (But that does at least mean that the beginning of 2200 would work.  That's `7'258'118'400`, GMT.)

src/main/ApplicationImpl.cpp
MonsieurNicolas: That's because you wrote the code that way, if we use the current clock you won't do any arithmetic on time and can just enforce that `nextCloseTime > `lcl.ct`.

rokopt: > That's because you wrote the code that way, if we use the current clock you won't do any arithmetic on time and can just enforce that nextCloseTime > lcl.ct`.

Don't we also have to enforce that a manually-specified close time is greater than the _virtual_ system now, regardless of whether we use the system clock as a default, because `HerderImpl::triggerNextLedger()` is going to use the virtual time?

src/main/ApplicationImpl.cpp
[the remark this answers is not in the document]
rokopt: (I'll impose the beginning-of-2200 limit, and add a `static_assert` that that doesn't exceed the value that would produce overflow in `HerderImpl::triggerNextLedger()`.)

src/main/ApplicationImpl.cpp
rokopt: > calling `std::chrono::steady_clock::now()` directly

Why `std::chrono::steady_clock::now()` as opposed to `std::chrono::system_clock::now()`?  The latter is what `VirtualClock::system_now()` uses in `REAL_TIME`, and `VirtualClock::system_now()` is what `HerderImpl::triggerNextLedger()` uses.

src/main/ApplicationImpl.cpp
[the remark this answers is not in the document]
rokopt: This is done in 6d1b919c14ad649d127e86040869b5deb0ae2706.

src/main/ApplicationImpl.cpp
```diff
@@ -549,27 +550,224 @@ ApplicationImpl::joinAllThreads()
...
+            getConfig().getExpectedLedgerCloseTime().count();
+        if (maxCloseTime - defaultManualCloseTimeInterval >= lastCloseTime)
+        {
+            uint64_t const defaultNextCloseTime =
+                lastCloseTime + defaultManualCloseTimeInterval;
+            nextCloseTime = std::max(nextCloseTime, defaultNextCloseTime);
+        }
+    }
+
+    auto const nextCloseSystemTime = VirtualClock::from_time_t(nextCloseTime);
```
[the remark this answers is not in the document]
rokopt: I've used the hardcoded constant as mentioned above.  I made the `releaseAssert` change in 0545209546bf8826bf1a3ec51d9b25a1e9e795e9.

src/main/ApplicationUtils.cpp
```diff
@@ -37,10 +39,24 @@ runWithConfig(Config cfg)
                           "NODE_IS_VALIDATOR to be set";
             return 1;
         }
+        if (cfg.RUN_STANDALONE)
+        {
+            clockMode = VirtualClock::VIRTUAL_TIME;
+            if (cfg.AUTOMATIC_MAINTENANCE_COUNT != 0 ||
+                cfg.AUTOMATIC_MAINTENANCE_PERIOD.count() != 0)
+            {
+                LOG(ERROR) << "Using MANUAL_CLOSE and RUN_STANDALONE together "
```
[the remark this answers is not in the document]
rokopt: Done in 819e6cea355a5ba73e71a179d5f25976962fd6b0.

src/main/ApplicationImpl.cpp
rokopt: Done in 8f353399bbdac17b5645942260f5809e4cb7031f, though as I mentioned, I used `std::chrono::system_clock::now()`, not `std::chrono::steady_clock::now()`.  Please let me know if I'm misunderstanding something there.

src/main/ApplicationImpl.cpp
rokopt: I _think_ I've made the kind of rearrangement that you are requesting in 4932dab679e9f802c211ee35dd8d1c83655bb6e5.  Please let me know whether that is what you were looking for.

src/main/ApplicationImpl.cpp
rokopt: I think that has also happened as part of 4932dab679e9f802c211ee35dd8d1c83655bb6e5.

rokopt:
I've made several more changes (without squashing commits this time) in response to Nicolas's latest comments.

I also noticed a bug myself, and fixed it in 1b4eafccfc65899f7914fe18d0fc426ea2aa0dd0.

I think it's ready for re-review.

src/main/ApplicationImpl.cpp
```diff
@@ -552,27 +553,224 @@ ApplicationImpl::joinAllThreads()
...
+    if (VirtualClock::to_time_t(nextCloseSystemTime) != nextCloseTime)
```
jonjove: I know you have already removed these checks, but I want to emphasize that they don't make that much sense to begin with. As I was saying in my previous comment, `time_t` can be signed so you simply _cannot_ overflow it under any circumstances. You can't try to retroactively catch it, because overflowing it is already undefined behavior.

As an analogy, the following code is wrong
```
bool doesAdditionOverflow(int x, int y)
{
    int z = x + y;
    return (z - y) != x;
}
```
because the undefined behavior has already occurred by the time you maybe noticed (and I say maybe, because nothing is guaranteed!).

There is no action required on this, so feel free to resolve it once you have read it.

jonjove (commented):
Just some nitpicks and no-action-required comments left from me.

src/main/ApplicationImpl.cpp
```diff
@@ -552,27 +553,224 @@ ApplicationImpl::joinAllThreads()
...
+    optional<uint32_t> const& userSpecifiedSeqNum)
+{
+    auto const startLedgerSeq = getLedgerManager().getLastClosedLedgerNum();
+
+    // The "scphistory" stores ledger sequence numbers as INTs.
+    if (startLedgerSeq >= std::numeric_limits<int32_t>::max())
+    {
+        throw std::invalid_argument(fmt::format(
+            FMT_STRING("Manually closed ledger sequence number ({}) can not be "
+                       "increased without overflowing database entry"),
```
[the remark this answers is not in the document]
rokopt: Fixed.

src/main/ApplicationImpl.cpp
```diff
@@ -552,27 +553,224 @@ ApplicationImpl::joinAllThreads()
...
+
+    auto const nextLedgerSeq = startLedgerSeq + 1;
+
+    if (userSpecifiedSeqNum)
+    {
+        if (*userSpecifiedSeqNum > std::numeric_limits<int32_t>::max())
+        {
+            throw std::invalid_argument(fmt::format(
+                FMT_STRING("Manual close ledger sequence number {} too large "
+                           "for database (maximum {})"),
```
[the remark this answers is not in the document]
rokopt: Fixed.

src/main/ApplicationImpl.cpp
rokopt: Makes sense, thanks.  Resolving.

rokopt:
> Referencing specific hashes is generally not the most useful approach on GitHub, because that hash is no longer in this PR (at least GitHub still has a reference to it now).

I thought that in the short term it might save a bit of time in review, but I can avoid it in future, of course.

src/main/ApplicationImpl.cpp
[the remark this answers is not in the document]
MonsieurNicolas: Oh didn't see that... Mmmm I guess you can just use asserts then, those error messages make it look like they're for the end user, not a core developer and calling into "not implemented" behavior.

src/main/ApplicationImpl.cpp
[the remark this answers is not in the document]
MonsieurNicolas: no we're within standard: 292 years is for duration, and epoch starts in 1970, so if you try to add 1970+292 you end up at 2262.

src/main/ApplicationImpl.cpp
MonsieurNicolas: This is fine yes, I was worried that the clock would go backwards, but we're guarding against that now, so people would just have to wait a bit if this happens

src/main/test/CommandHandlerTests.cpp
```diff
@@ -137,3 +145,447 @@ TEST_CASE("transaction envelope bridge", "[commandhandler]")
...
+                submitClose(noLedgerSeq, make_optional<TimePoint>(maxTime + 1),
+                            retStr);
+
+                LedgerTxn lookupTx(app->getLedgerTxnRoot());
+                auto entry = lookupTx.load(LedgerEntryKey(dle));
+                REQUIRE(!entry);
+            }
+        }
+
+        SECTION("Large transaction queues may be emptied over the course of "
```
MonsieurNicolas, line 504: this test is getting way too deep into testing "everything that core does", we already have tests for the transaction queue, and this is just duplicating that so there is no real added test coverage, on the other hand we're making our test suite slower.

src/main/ApplicationImpl.cpp
MonsieurNicolas: yes things look good

src/main/test/CommandHandlerTests.cpp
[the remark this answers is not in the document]
MonsieurNicolas: let's not do stress testing as part of our unit tests:
* our test suite is already slow enough
* if we ever care about performance of this particular scenario, we'll tackle it then, when we actually understand the sequence of inputs that are causing performance issues

src/main/ApplicationImpl.cpp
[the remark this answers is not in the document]
rokopt: Okay, I've changed them to `releaseAssert()`s.

src/main/ApplicationImpl.cpp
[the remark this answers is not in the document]
rokopt: Oh, right, I forgot that meant 292 years from _epoch_, not from _now_.

src/main/test/CommandHandlerTests.cpp
rokopt: Okay, I've removed it.

src/main/test/CommandHandlerTests.cpp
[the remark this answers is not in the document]
rokopt: Alright, that one's gone.

rokopt:
I've squashed commits after addressing Nicolas's latest review comments.

MonsieurNicolas:
r+ b91dd1b360073ee153669696d9c0214f117db2b3

src/herder/HerderImpl.cpp
rokopt: Actually, disabling it when only `RUN_STANDALONE` is present breaks, at least, the "3 nodes 2 running threshold 2"/"Over loopback" acceptance test.  I've pushed a change to revert to the original version which disables the timer only if both `MANUAL_CLOSE` and `RUN_STANDALONE` are configured.

MonsieurNicolas:
r+ 023ec9969997a4ebc31054015fd32e1c6d2fd86a

src/herder/HerderImpl.cpp
jonjove: @MonsieurNicolas can you open an issue for that with some useful context about why it is like this? It strongly defies my expectations that we might need to broadcast when running standalone.

MonsieurNicolas: Yeah @jonjove I opened https://github.com/stellar/stellar-core/issues/2719

Issue #2634, opened 2020-07-21
Captive Stellar-Core: Stream meta data starting from any ledger


bartekn:
## Description

Stellar-Core 12.3.0 added a great feature of streaming metadata (`LedgerCloseMeta`s) via a filesystem pipe:
> an experimental feature for generating meta data without inserting it into the database while replaying in memory, this allows to speed up meta data generation (for historical data) by several orders of magnitude

This really helped Horizon team to speed reingestion of entire pubnet history from weeks to around 1.5 days. Additional advantage of this new feature is decoupling Stellar-Core from Horizon. Currently Horizon is connecting to Stellar-Core database directly to get ledger data however it's causing many issues connected to operations (ex. missing ledgers if cursor is not updated).

It would be great to be able to use this mode not only in reingestion (offline mode) but also when ingesting newly closed ledgers (online mode).

### Explain why this feature is important

Allows completely decoupling Stellar-Core and Horizon and removing operational burden connected to running Stellar-Core used by Horizon.

### Describe the solution you'd like

We propose to add a new command to stellar-core: `run-from [starting_ledger]` that would start streaming from `starting_ledger` to the latest closed by a network and then would continue streaming new ledgers as they are closed.

### Describe alternatives you've considered

The current workaround we use when starting a new captive session is the following chain of commands that use a persistent DB:
* `stellar-core new-db`,
* `stellar-core catchup [starting_ledger-1]/0`,
* `stellar-core run`.

While it works fine, it has some disadvantages:
* It requires specifying a DB connection and maintaining a DB.
* Applying buckets stage is much slower when using a disk vs RAM.
* `stellar-core run` execution can be slow when it needs to wait for the next checkpoint.

The new command would solve all the issues above.

MonsieurNicolas:
I don't know if we need a new subcommand with those semantics.

It seems to me that:
* the main functionality is actually related to `run` with the in-memory streaming mode, so maybe `run --in-memory`?
* we need is an additional option that allows to override the behavior when starting up from fresh (right now this is controlled by `CATCHUP_RECENT`).

I *think* that we also need to make sure that we pass in not only the ledger number but its hash as well as a way to have a contract that guarantees that states in different systems are in agreement.

We probably need to track what other features are needed for a "live" in memory watcher node. For example, we have (I think?) to ensure that the overlay database is persisted in that mode as to avoid (potentially very) long startup times when the node gets restarted.

bartekn:
[...]

PR #2700, opened 2020-09-10
Bug 2634 run in memory from checkpoint

graydon:
Initial implementation of bug #2634 that allows one to run something like:

```
stellar-core --conf ~/core.cfg run --in-memory --start-at-ledger 31534975 --start-at-hash 407dc8a57ec308c12fa292f11d98080898b403fafc8dc2c12eb48d70977258bf
```

### DIFF
--- a/src/main/ApplicationUtils.cpp
+++ b/src/main/ApplicationUtils.cpp
@@ -17,6 +17,7 @@
 #include "main/Maintainer.h"
 #include "main/PersistentState.h"
 #include "main/StellarCoreVersion.h"
+#include "overlay/OverlayManager.h"
 #include "util/Logging.h"
 #include "work/WorkScheduler.h"
 
@@ -27,7 +28,7 @@ namespace stellar
 {
 
 int
-runWithConfig(Config cfg)
+runWithConfig(Config cfg, optional<CatchupConfiguration> cc)
 {
     if (cfg.MANUAL_CLOSE)
     {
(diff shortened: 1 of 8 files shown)


src/main/CommandLine.cpp
```diff
@@ -273,6 +273,19 @@ maybeSetMetadataOutputStream(Config& cfg, std::string const& stream)
     }
 }
 
+void
+maybeEnableInMemoryLedgerMode(Config& config, bool inMemory)
+{
+    // Adjust configs for in-memory-replay mode
```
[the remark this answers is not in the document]
graydon: Oh, whoops, good point! Left out a little more validation -- fixed!

src/main/ApplicationUtils.cpp
```diff
@@ -56,7 +57,26 @@ runWithConfig(Config cfg)
...
-        app->start();
+        if (cc)
+        {
+            Json::Value catchupInfo;
+            std::shared_ptr<HistoryArchive> archive;
+            auto const& ham = app->getHistoryArchiveManager();
+            archive = ham.selectRandomReadableHistoryArchive();
+            if (catchup(app, *cc, catchupInfo, archive) == 0)
+            {
+                // Pre-run catchup defers overlay startup.
```
MonsieurNicolas, line 68: nit: looks to me that this code should be moved to line 80 with something like `if (!cfg.MODE_AUTO_STARTS_OVERLAY) {...}` as to make it future proof

src/main/ApplicationUtils.cpp
```diff
@@ -56,7 +57,26 @@ runWithConfig(Config cfg)
...
+                app->getOverlayManager().start();
+            }
+            else
+            {
+                return 1;
```
MonsieurNicolas, line 73: Should we return the error code returned by catchup? It returns 3 on failure from what I see and as a caller it can be nice to know which stage failed, Horizon may decide to retry differently if catchup failed for example

MonsieurNicolas (changes requested):
looks pretty close to landing. Just a couple suggestions

src/main/ApplicationUtils.cpp
graydon: Yep, good call. Fixed.

src/main/ApplicationUtils.cpp
graydon: Yes, fixed.

MonsieurNicolas:
r+ 36c5eef84b529aff880da769009683b6174019fc

PR #2702, opened 2020-09-11
Small cleanups and comments

graydon:
Avoid a couple no-ops that nonetheless trigger diagnostics, and add a couple asserts and comments.

### DIFF
--- a/src/crypto/XDRHasher.h
+++ b/src/crypto/XDRHasher.h
@@ -26,6 +26,15 @@ template <typename Derived> struct XDRHasher
     void
     queueOrHash(unsigned char const* u, size_t sz)
     {
+        if (u == nullptr)
+        {
+            if (sz != 0)
+            {
+                throw std::invalid_argument(
+                    "nullptr input with nonzero length");
+            }
+            return;
+        }
         if (sz > available())
         {
             flush();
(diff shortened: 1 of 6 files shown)


src/crypto/XDRHasher.h
```diff
@@ -35,8 +35,11 @@ template <typename Derived> struct XDRHasher
                 return;
             }
         }
-        memcpy(mBuf + mLen, u, sz);
-        mLen += sz;
+        if (sz != 0)
+        {
+            memcpy(mBuf + mLen, u, sz);
```
MonsieurNicolas, line 40: My reading of the spec is that UB is when the pointers are invalid/null, `sz == 0` by itself is not UB. As `mBuf` is a valid pointer, we should just have a guard at the top of that function against `u == nullptr`

graydon: > My reading of the spec is that UB is when the pointers are invalid/null, `sz == 0` by itself is not UB. As `mBuf` is a valid pointer, we should just have a guard at the top of that function against `u == nullptr`

That's the thing: u is sometimes nullptr (exactly when sz == 0). I can _branch_ on either, but I can't exclude either as a possibility (unless it's a bug in xdrpp -- looking).

graydon: Yeah this isn't a bug in xdrpp, it's just a consequence of the way we're called when processing a string field that's optional / missing (or otherwise zero-length). I can branch on nullptr and assert zero size, if you prefer that?

MonsieurNicolas:
r+ 9093c01431b7570b28c964fe0e032fba11ddd79c

graydon:
r+ 07d1998

graydon:
(previous failure was a missing clang-format)

Issue #2681, opened 2020-08-26
move to "real" json for "dump-xdr"/"print-xdr"


MonsieurNicolas:
An option `--json` was added to `dump-xdr` some time ago as to use proper json when using subcommands.

This was partially added: the `print-xdr` command doesn't have this option.

We should:
* remove that option and move to using that real "json" formatter the default behavior for both `dumpt-xdr` and `print-xdr`
* ensure that there are the right overrides when dealing with `MuxedAccount`
  * it looks like the json formatter doesn't have the override that "xdr_printer" has `xdr_printer(MuxedAccount const& muxedAccount)`

NB: I did a quick performance test, calling `dump-xdr` on a bucket file, and there doesn't seem to be any difference in runtime between the json and the old-style version.

MonsieurNicolas:
I think we should add a new option `--compact` to those commands, that would cause to pass in `cereal::JSONOutputArchive::Options::NoIndent()` as an option to `cereal::JSONOutputArchive` : this speeds things up by over 50% and scripts only need that.

jonjove:
I don't think this issue covers quite everything that #2543 covered. #2543 was partially directed at other uses of `xdr::xdr_to_string` (such as logging in an invariant failure). The `JSONOutputArchive` functions will serve that purpose well, but we should

a. Move those functions into a utility header so they can be accessed wherever they are needed
b. Stop using `xdr::xdr_to_string` everywhere since it doesn't handle custom multi-line output well

(a) should definitely be part of this issue. (b) can be done in a separate issue if you prefer, I'm indifferent.

MonsieurNicolas:
Yeah that sounds right. This issue can track both. When we replace `xdr_to_string` we need to use the version that indents.

jonjove:
Part (b) in my last comment hasn't been addressed yet. Should we reopen this?

MonsieurNicolas:
ah yes you are right - forgot we wanted to do this sweep.

OK, I am re-opening it.

I also noticed that we left some `xdr_printer` functions in `dumpxdr.cpp` that are probably not used.

PR #2703, opened 2020-09-11
Make print-xdr and dump-xdr use Cereal to print

hidenori-shinohara:
# Description

Resolves Part (a) of #2681.

This PR makes print-xdr and dump-xdr use Cereal to improve readability such as multiline printing, and also adds the `compact` flag which removes all indentation.

After:
```
{
    "value0": {
        "type": "ENVELOPE_TYPE_TX",
        "v1": {
            "tx": {
                "sourceAccount": "GAY6WPTYIMNRRFEFOKUE3KRRKIMAWRB3XLSO2M6HNMIABBPDR5DTA3IY",
                "fee": 100,
                "seqNum": 903274572021761,
                "timeBounds": null,
                "memo": {
                    "type": "MEMO_NONE"
                },
                "operations": [
                    {
                        "sourceAccount": {
                            "id": 0,
                            "accountID": "GAY6WPTYIMNRRFEFOKUE3KRRKIMAWRB3XLSO2M6HNMIABBPDR5DTA3IY"
                        },
                        "body": {
                            "type": "PAYMENT",
                            "paymentOp": {
                                "destination": {
                                    "id": 123,
                                    "accountID": "GD5FAOMDOCXDYNNY3QHGRQZIJ42FWYJBFW7OJQMUWCHIFJMDW6RFS6XQ"
                                },
                                "asset": "XLM",
                                "amount": 1000
                            }
                        }
                    }
                ],
                "ext": {
                    "v": 0
                }
            },
            "signatures": [
                {
                    "hint": "e38f4730",
                    "signature": "39344833d02394318b530efbfbc9c7d7e7ace3d6910ee21ca37cafcaa38d1fc87cb2dfb4a142ee2ccf79b23b99266449f82c8f388cc0b05c263271f8319f1308"
                }
            ]
        }
    }
}
```

Before:
```
TransactionEnvelope = {
  type = ENVELOPE_TYPE_TX,
  v1 = {
    tx = {
      sourceAccount = GAY6WPTYIMNRRFEFOKUE3KRRKIMAWRB3XLSO2M6HNMIABBPDR5DTA3IY,
      fee = 100,
      seqNum = 903274572021761,
      timeBounds = NULL,
      memo = {
        type = MEMO_NONE
      },
      operations = [
        { sourceAccount = { id = 0, accountID = GAY6WPTYIMNRRFEFOKUE3KRRKIMAWRB3XLSO2M6HNMIABBPDR5DTA3IY },
          body = {
            type = PAYMENT,
            paymentOp = {
              destination = { id = 123, accountID = GD5FAOMDOCXDYNNY3QHGRQZIJ42FWYJBFW7OJQMUWCHIFJMDW6RFS6XQ },
              asset = {
                type = ASSET_TYPE_NATIVE
              },
              amount = 1000
            }
          } }
      ],
      ext = {
        v = 0
      }
    },
    signatures = [
      { hint = e38f4730,
        signature = 39344833d02394318b530efbfbc9c7d7e7ace3d6910ee21ca37cafcaa38d1fc87cb2dfb4a142ee2ccf79b23b99266449f82c8f388cc0b05c263271f8319f1308 }
    ]
  }
}
```

# Checklist
- [x] Compiles
- [x] Ran all tests

### DIFF
--- a/src/main/dumpxdr.cpp
+++ b/src/main/dumpxdr.cpp
@@ -164,29 +164,21 @@ xdr_printer(MuxedAccount const& muxedAccount)
 
 template <typename T>
 void
-dumpstream(XDRInputFileStream& in, bool json)
+dumpstream(XDRInputFileStream& in, bool compact)
 {
     T tmp;
-    if (json)
+    cereal::JSONOutputArchive archive(
+        std::cout, compact ? cereal::JSONOutputArchive::Options::NoIndent()
+                           : cereal::JSONOutputArchive::Options::Default());
+    archive.makeArray();
+    while (in && in.readOne(tmp))
     {
-        cereal::JSONOutputArchive archive(std::cout);
-        archive.makeArray();
-        while (in && in.readOne(tmp))
-        {
-            archive(tmp);
-        }
-    }
-    else
-    {
-        while (in && in.readOne(tmp))
-        {
-            std::cout << xdr::xdr_to_string(tmp) << std::endl;
-        }
+        archive(tmp);
     }
 }
 
 void
-dumpXdrStream(std::string const& filename, bool json)
+dumpXdrStream(std::string const& filename, bool compact)
 {
     std::regex rx(
         ".*(ledger|bucket|transactions|results|scp)-[[:xdigit:]]+\\.xdr");
@@ -198,24 +190,24 @@ dumpXdrStream(std::string const& filename, bool json)
 
         if (sm[1] == "ledger")
         {
-            dumpstream<LedgerHeaderHistoryEntry>(in, json);
+            dumpstream<LedgerHeaderHistoryEntry>(in, compact);
         }
         else if (sm[1] == "bucket")
         {
-            dumpstream<BucketEntry>(in, json);
+            dumpstream<BucketEntry>(in, compact);
         }
         else if (sm[1] == "transactions")
         {
-            dumpstream<TransactionHistoryEntry>(in, json);
+            dumpstream<TransactionHistoryEntry>(in, compact);
         }
         else if (sm[1] == "results")
         {
-            dumpstream<TransactionHistoryResultEntry>(in, json);
+            dumpstream<TransactionHistoryResultEntry>(in, compact);
         }
         else
         {
             assert(sm[1] == "scp");
-            dumpstream<SCPHistoryEntry>(in, json);
+            dumpstream<SCPHistoryEntry>(in, compact);
         }
     }
     else
@@ -256,20 +248,23 @@ readFile(const std::string& filename, bool base64 = false)
 
 template <typename T>
 void
-printOneXdr(xdr::opaque_vec<> const& o, std::string const& desc)
+printOneXdr(xdr::opaque_vec<> const& o, std::string const& desc, bool compact)
 {
     T tmp;
     xdr::xdr_from_opaque(o, tmp);
-    cereal::JSONOutputArchive archive(std::cout);
-    archive(tmp);
+    cereal::JSONOutputArchive ar(
+        std::cout, compact ? cereal::JSONOutputArchive::Options::NoIndent()
+                           : cereal::JSONOutputArchive::Options::Default());
+    ar(tmp);
 }
 
 void
-printXdr(std::string const& filename, std::string const& filetype, bool base64)
+printXdr(std::string const& filename, std::string const& filetype, bool base64,
+         bool compact)
 {
 // need to use this pattern as there is no good way to get a human readable
 // type name from a type
-#define PRINTONEXDR(T) std::bind(printOneXdr<T>, _1, #T)
+#define PRINTONEXDR(T) std::bind(printOneXdr<T>, _1, #T, compact)
     auto dumpMap =
         std::map<std::string, std::function<void(xdr::opaque_vec<> const&)>>{
             {"ledgerheader", PRINTONEXDR(LedgerHeader)},
--- a/src/main/dumpxdr.cpp
+++ b/src/main/dumpxdr.cpp
@@ -78,6 +78,32 @@ cereal_override(cereal::JSONOutputArchive& ar, const stellar::PublicKey& s,
 {
     xdr::archive(ar, stellar::KeyUtils::toStrKey<stellar::PublicKey>(s), field);
 }
+
+void
+cereal_override(cereal::JSONOutputArchive& ar,
+                const stellar::MuxedAccount& muxedAccount, const char* field)
+{
+    switch (muxedAccount.type())
+    {
+    case stellar::KEY_TYPE_ED25519:
+        xdr::archive(ar, stellar::KeyUtils::toStrKey(toAccountID(muxedAccount)),
+                     field);
+        return;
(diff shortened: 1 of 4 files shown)


src/main/dumpxdr.cpp
```diff
@@ -78,6 +78,43 @@ cereal_override(cereal::JSONOutputArchive& ar, const stellar::PublicKey& s,
 {
     xdr::archive(ar, stellar::KeyUtils::toStrKey<stellar::PublicKey>(s), field);
 }
+
+// A struct used to archive MuxedAccount
+// when type = KEY_TYPE_MUXED_ED25519.
+struct MuxedEd25519
```
jonjove: I don't understand the purpose of this type? Perhaps I'm missing something, but if you want to archive this information then why don't you just archive it directly from the relevant `cereal_overload`?

src/main/dumpxdr.cpp
```diff
@@ -78,6 +78,43 @@ cereal_override(cereal::JSONOutputArchive& ar, const stellar::PublicKey& s,
...
+    {
+        ar(CEREAL_NVP(id), CEREAL_NVP(accountID));
+    }
+};
+
+void
+cereal_override(cereal::JSONOutputArchive& ar,
+                const stellar::MuxedAccount& muxedAccount, const char* field)
+{
+    MuxedEd25519 account;
```
jonjove: Nitpick: It is generally good policy to declare variables as close to where they are needed as possible (see Scott Meyers' Effective C++, Third Edition, Item 26). This should be declared in the `KEY_TYPE_MUXED_ED25519` branch, which will require you to add an explicit scope there.

That being said, I don't think you need this at all -- see the previous comment.

src/util/XDRCereal.h
```diff
@@ -97,6 +97,14 @@ cereal_override(cereal::JSONOutputArchive& ar, const stellar::Asset& s,
     xdr::archive(ar, stellar::assetToString(s), field);
 }
 
+template <typename T>
+typename std::enable_if<xdr::xdr_traits<T>::is_enum>::type
+cereal_override(cereal::JSONOutputArchive& ar, const T& t, const char* field)
```
jonjove, line 102: One thing I'm wondering is whether we should make it optional to print names (for human readable purposes) or values (for computer purposes). Another option would be to always print names and optionally print values. Perhaps this behavior could be controlled by `--compact`. @MonsieurNicolas where do you stand on this issue?

src/main/dumpxdr.cpp
hidenori-shinohara: `MuxedAccount` seems to have different levels of "nesting" based on the type.

For `KEY_TYPE_ED25519`, we print

```
...
              "sourceAccount": "GAY6WPTYIMNRRFEFOKUE3KRRKIMAWRB3XLSO2M6HNMIABBPDR5DTA3IY",
...
```

but, for `KEY_TYPE_MUXED_ED25519`, we print

```
...
                        "sourceAccount": {
                            "id": 0,
                            "accountID": "GAY6WPTYIMNRRFEFOKUE3KRRKIMAWRB3XLSO2M6HNMIABBPDR5DTA3IY"
                        },
                        "body": {
                            "type": "PAYMENT",
...
```

If I instead do `xdr::archive(ar, id, "id");` and `xdr::archive(ar, accountID, "accountID");`, it doesn't print "sourceAccount" and I end up with

```
...
                            "id": 0,
                            "accountID": "GAY6WPTYIMNRRFEFOKUE3KRRKIMAWRB3XLSO2M6HNMIABBPDR5DTA3IY",
                            "body": {
                                "type": "PAYMENT",
...
```

If I put the "id" and "accountID" in one string value and do `xdr::archive(ar, rendered_id_accountID, "sourceAccount")`, then I end with the multi-line printing issue because I won't know how much to indent.

This solution seems pretty weird, and I feel like I'm doing something very unnecessary, but I couldn't find other ways to do this.

src/util/XDRCereal.h
hidenori-shinohara: I thought about it after our conversation, but I left it out for now because the current code seems to only print human readable names (both `print-xdr` and `dump-xdr`).

src/main/dumpxdr.cpp
hidenori-shinohara: Ah, that makes sense! There's no need for `account` to be visible outside the branch. I'll keep that there for now since I might delete it after all.

src/util/XDRCereal.h
```diff
@@ -0,0 +1,111 @@
...
+                const stellar::MuxedAccount& muxedAccount, const char* field)
+{
+    switch (muxedAccount.type())
+    {
+    case stellar::KEY_TYPE_ED25519:
+        xdr::archive(ar, stellar::KeyUtils::toStrKey(toAccountID(muxedAccount)),
+                     field);
+        return;
+    case stellar::KEY_TYPE_MUXED_ED25519:
+        xdr::archive(ar,
```
jonjove, line 68: Nitpick: please don't move code while simultaneously modifying it, because it makes reviewing much more difficult. Can you split this as either "move then modify" or "modify then move" (which is the approach you had previously taken)?

hidenori-shinohara: Makes sense! Changed it in https://github.com/stellar/stellar-core/pull/2703/commits/a486bd80caae2d8dbf05e116b503f6265695e2fe

src/main/dumpxdr.cpp
hidenori-shinohara: Removed

src/main/dumpxdr.cpp
hidenori-shinohara: Removed!

src/main/dumpxdr.cpp
```diff
@@ -78,6 +78,31 @@ cereal_override(cereal::JSONOutputArchive& ar, const stellar::PublicKey& s,
...
+    {
+    case stellar::KEY_TYPE_ED25519:
+        xdr::archive(ar, stellar::KeyUtils::toStrKey(toAccountID(muxedAccount)),
+                     field);
+        return;
+    case stellar::KEY_TYPE_MUXED_ED25519:
+        xdr::archive(ar,
+                std::make_tuple(
+                    cereal::make_nvp("id", muxedAccount.med25519().id),
+                    cereal::make_nvp("accountID", stellar::KeyUtils::toStrKey(toAccountID(muxedAccount)))
```
jonjove: Missing `clang-format`

jonjove (commented):
Generally looks good to me.

src/util/XDRCereal.h
MonsieurNicolas: Yeah we can keep just friendly names for now.
If we need automation to do something with those I suspect that they can just treat them as strings and are very unlikely to need to convert it back to binary. If we ever need this, we can consider adding a new option when we need this.

src/main/dumpxdr.cpp
hidenori-shinohara: Done!

jonjove:
r+ 183fe98

PR #2710, opened 2020-09-16
SCP: remove dependency on config.h

### LINKED ISSUE #2707: SCP should not depend on Config.h
as noted in https://github.com/stellar/stellar-core/pull/2602#issuecomment-691929175

we should not have a dependency back to the main app's config.

In this case, the setting should be propagated down instead

MonsieurNicolas:
# Description

Resolves #2707

Also made a few additional cleanups in the code base


marta-lokhova:
r+ 414b8a83f7989c351d306150f3368d2faddb9156

AndrejMitrovic:
@MonsieurNicolas sorry to bump an old PR. I think the previous behavior in `validateValues` was actually buggy, correct? It was setting `kMaybeValidValue` or `kInvalidValue` based entirely on what the **last** item in the `values` set was, regardless of any previous ones.

Is my understanding correct that the function should return the following?:

- `kValidValue` -> only if **all** values are valid
- else, `kMaybeValidValue` if **at least one** value is `kMaybeValidValue`
- else `kInvalidValue`

PR #2711, opened 2020-09-16
add new option to `catchup` command to active http endpoint

MonsieurNicolas:
Default was changed in 20acb825832fdc2114d238211ac1833ab91d7557 which ended up being a breaking change, this makes that behavior "opt-in".


graydon:
Hmm, unsure; I might prefer to do it via the existing HTTP_PORT / PUBLIC_HTTP_PORT config settings? I mean, I guess the question is whether those config settings implicitly mean ".. only when doing run, not catchup"?

What exactly is going wrong with the previous change?

PR #2713, opened 2020-09-16
Allow to set HTTP port in config to 0 as a way to disable it

MonsieurNicolas:
The change introduced in 20acb82 introduces a dependency on setting the HTTP_PORT to a unique value.

This change allows to disable the HTTP port via configuration when core is used exclusively via command line.

Replaces #2696


jonjove:
r+ ecd8518

PR #2716, opened 2020-09-17
Simulation doc updates and bucket generation bugfix

marta-lokhova:
@rokopt pointed out a few typos in the simulation docs

This also fixes a bug introduced in https://github.com/stellar/stellar-core/pull/2628 
Instead, this change doesn't special case the empty buckets.


rokopt (approved):

rokopt (approved):

src/transactions/simulation/TxSimGenerateBucketsWork.cpp
```diff
@@ -152,14 +149,9 @@ TxSimGenerateBucketsWork::onRun()
...
             mApp.getBucketManager().getBucketByHash(hexToBin256(bucketHash));
-        Hash emptyHash;
-        if (bucket->getHash() != emptyHash)
-        {
-            CLOG(INFO, "History")
-                << "Simulating " << (mIsCurr ? "curr" : "snap")
-                << " bucketlist level: " << mLevel;
-            startBucketGeneration(bucket);
-        }
+        CLOG(INFO, "History") << "Simulating " << (mIsCurr ? "curr" : "snap")
```
MonsieurNicolas, line 152: Can you explain which issue is getting fixed?

The other change is a straight rollback of 9ac04eb5fca4d9bd0fd3b3135e8d21d115989d1c - but what is the actual problem here?

From what I see
* the new behavior is invoking `startBucketGeneration` with an empty bucket to populate `mIntermediateBuckets` with a bunch of empty buckets that end up getting merged into an empty bucket - so we don't special case "empty bucket" anymore
* but the old behavior was using `mIntermediateBuckets.empty()` as a proxy for "empty bucket"

I can't tell why those two behaviors end up producing different results.

marta-lokhova: The problem here is not related to empty buckets, it's just that by trying to accommodate empty buckets, I broke generation of normal buckets. The issue with https://github.com/stellar/stellar-core/commit/9ac04eb5fca4d9bd0fd3b3135e8d21d115989d1c is that on the first `onRun` it'd skip the first bucket -- `curr` of level 0. So there were two cases where `mIntermediateBuckets.empty()`: at the very beginning (we haven't started merging), and when we encounter empty buckets (we just skip that bucket). 

I don't think handling empty buckets really matters here: transaction replay requires protocol 12 or higher (so no buckets with empty hashes), and performance-wise, it's the buckets with contents that make a difference.

src/transactions/simulation/TxSimGenerateBucketsWork.cpp
```diff
@@ -124,12 +124,9 @@ TxSimGenerateBucketsWork::onRun()
...
-    else
+    else if (!mIntermediateBuckets.empty())
     {
-        if (!mIntermediateBuckets.empty())
-        {
-            processGeneratedBucket();
-        }
+        processGeneratedBucket();
 
         if (mLevel < BucketList::kNumLevels - 1)
```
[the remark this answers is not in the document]
marta-lokhova: Skipping that portion is part of the fix -- here we're constantly checking if there are any finished bucket merges (mIntermediateBuckets), and if so, we advance the bucketlist to the next bucket. The only time we should not advance the bucketlist is at the very beginning, when we haven't started merging.

src/transactions/simulation/TxSimGenerateBucketsWork.cpp
```diff
@@ -124,12 +124,9 @@ TxSimGenerateBucketsWork::onRun()
         mTimer->async_wait(wakeSelfUpCallback(), &VirtualTimer::onFailureNoop);
         return BasicWork::State::WORK_WAITING;
     }
-    else
+    else if (!mIntermediateBuckets.empty())
```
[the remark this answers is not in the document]
marta-lokhova: Updated the code (quick search shows that we use "else after a return" quite often, though I don't think we strictly prefer one convention over the other)

src/transactions/simulation/TxSimGenerateBucketsWork.cpp
MonsieurNicolas: ah yes the first iteration was different

MonsieurNicolas:
r+ 6a097904cb3c36559da1f259588f893833d0d0eb

PR #2717, opened 2020-09-18
Set "fail" bit on streams

MonsieurNicolas:
In a few places we were not enabling `failbit` checks.
This causes all sorts of unexpected behavior: in particular, if the file cannot be opened, trying to read from it will just return "no data" and the failure will be silent.

### DIFF
--- a/src/history/HistoryArchive.cpp
+++ b/src/history/HistoryArchive.cpp
@@ -108,8 +108,9 @@ void
 HistoryArchiveState::save(std::string const& outFile) const
 {
     ZoneScoped;
-    std::ofstream out(outFile);
+    std::ofstream out;
     out.exceptions(std::ios::failbit | std::ios::badbit);
+    out.open(outFile);
     cereal::JSONOutputArchive ar(out);
     serialize(ar);
 }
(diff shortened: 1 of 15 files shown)


src/history/HistoryArchive.cpp
```diff
@@ -108,8 +108,9 @@ void
 HistoryArchiveState::save(std::string const& outFile) const
 {
     ZoneScoped;
-    std::ofstream out(outFile);
+    std::ofstream out;
     out.exceptions(std::ios::failbit | std::ios::badbit);
```
marta-lokhova, line 112: I know this was here before, but we should update TxSimGenerateBucketsWork::onSuccess, which is using this function and can throw. 
* we should guard the call to `save` with try/catch and return WORK_FAILURE if there's a problem
* To make the change above work, it needs to be moved to `onRun` https://github.com/stellar/stellar-core/blob/24d1a612aa0ee99db993ee97ceb1891d9a00e9cd/src/transactions/simulation/TxSimGenerateBucketsWork.cpp#L141

marta-lokhova: As for the other subsystems, from what i see, catchup/publish are already properly using `load`/`save`.

MonsieurNicolas: updated

marta-lokhova:
r+ f2c84c8c3ed9c893b5b81233910feea92dbb191c

Issue #1753, opened 2018-08-10
Update peer table to only include reacheable peers over time


MonsieurNicolas:
Right now the `peers` table contains a cesspool of peers received from other peers and the only indicator that peers may be bad is the `numfailures` column.

This makes peer discovery quite suboptimal as peers forwarded to other peers may not be reacheable.

I think we should make the following changes to the way we track peers:
* define a clear "validation" process for peers.
   * `new`: we do not know anything about this peer
   * `outbound_valid`: we could connect to this peer at some point
   * preferred and known peers would set the flags such that they would assume to be "healthy"
* only use `numfailures` for outbound connections, right now, I *think* we reset it when a peer connects to us, this is wrong
* only forward peers with `outbound_valid`
* probably delete peers that we could not connect to after X retries (probably should be equivalent to 1 or 2 weeks of attempts)

Second, we should ensure that we have logic in place to properly scrub the peers table:
right now we only attempt connections when we are below `TARGET_PEER_CONNECTIONS`, causing the entries in that table to rot away.

A few potential options:
* always connect to a bunch of peers, but drop them later if we find that we were already above our target and we were the initiator of those connections.
* add a "peer reaper" sub task that purpose is to scrub the peer table, I am not sure how we can ensure that this doesn't potentially negatively impact the normal connection process

vogel:
I'm wondering if another solution, not based on states, would not be better.

I propose to save in PeerRecord several additional fields (all nullable datetimes):

- lastOutboundConnectionAttempt (set on connectTo)
- lastOutboundConnectionSuccess (set recvAuth when mRole = WE_CALLED_REMOTE, before getOverlayManager().acceptAuthenticatedPeer())
- lastInboundConnectionAttempt (set on recvHello when mRole = REMOTE_CALLED_US)
- lastInboundConnectionSuccess (set on recvAuth when mRole = REMOTE_CALLED_US, before getOverlayManager().acceptAuthenticatedPeer())

Of course when lastOutboundConnectionAttempt is NULL, also lastOutboundConnectionSuccess is NULL and when lastInboundConnectionAttempt is NULL, lastInboundConnectionSuccess is NULL.

Following states can be deducted from those values:

node is new when:
lastOutboundConnectionAttempt == NULL && lastInboundConnectionAttempt == NULL

node never accepted outbound connection:
lastOutboundConnectionAttempt != NULL && lastOutboundConnectionSuccess == NULL

node properly accepted last outbound connection:
[...]

PR #2720, opened 2020-09-21
Improve connection management flow

marta-lokhova:
* Limit number of connection attempts by the number of available pending slots: I believe this change is strictly better than master because in `OverlayManagerImpl::addOutboundConnection`, we drop any pending connection if we do not have enough pending slots (currently, we initiate connections and update backoff anyway)
* Minor cleanup of the `tick` function to make the code more readable

### DIFF
--- a/src/overlay/OverlayManagerImpl.cpp
+++ b/src/overlay/OverlayManagerImpl.cpp
@@ -532,42 +532,49 @@ OverlayManagerImpl::tick()
     }
 
     auto availablePendingSlots = availableOutboundPendingSlots();
-    auto availableAuthenticatedSlots = availableOutboundAuthenticatedSlots();
-    auto nonPreferredCount = nonPreferredAuthenticatedCount();
-
-    // if all of our outbound peers are preferred, we don't need to try any
-    // longer
-    if (availableAuthenticatedSlots > 0 || nonPreferredCount > 0)
+    if (availablePendingSlots == 0)
     {
-        // try to replace all connections with preferred peers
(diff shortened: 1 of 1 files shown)


src/overlay/OverlayManagerImpl.cpp
```diff
@@ -532,42 +532,49 @@ OverlayManagerImpl::tick()
...
+    // PREFERRED_PEER_ONLY is set, to support key-based preferred peers mode
+    // (see PREFERRED_PEER_KEYS). When PREFERRED_PEER_ONLY is set and we connect
+    // to a non-preferred peer, drop it and backoff during handshake.
+    if (availablePendingSlots > 0 && availableAuthenticatedSlots > 0)
+    {
+        // try to leave at least some pending slots for peer promotion
+        constexpr const auto RESERVED_FOR_PROMOTION = 1;
+        auto outboundToConnect =
+            availablePendingSlots > RESERVED_FOR_PROMOTION
+                ? std::min(availablePendingSlots - RESERVED_FOR_PROMOTION,
```
marta-lokhova, line 568: @MonsieurNicolas I didn't change this calculation, as I'm still not sure about the contract between outbound vs inbound-to-outbound promotions. Based on the discussion in #1753 (comment), it seems like we should prioritize outbound target before attempting promotions, but this is not what the code is doing here (as it subtracts the const from pending before calculating min)

MonsieurNicolas:
r+ b11afcc51e56d76ca63032445a46b67da156359c

PR #2725, opened 2020-09-24
isQuorumSetSane: Add error string parameter

### LINKED ISSUE #2723: Making isQuorumSetSane more informative
Hi,

We have some downstream changes to SCP which we'd like to upstream into `stellar-core`. One of them is making `isQuorumSetSane` more informative so it's possible to see why a quorum set is considered unsafe to use.

In our fork we used `char**` for this reason (https://github.com/bpfkorea/agora/pull/549), however I can see alternative implementions:

- Use `std::string` for the `reason` parameter
- Alternatively, use logging instead

Do you have any preference, and would you accept such a feature if we provided the patch?

Thanks!

AndrejMitrovic:
Please feel free to suggest improvements. Thanks!

Resolves #2723

### DIFF
--- a/src/main/Config.cpp
+++ b/src/main/Config.cpp
@@ -1258,12 +1258,10 @@ Config::validateConfig(ValidationThresholdLevels thresholdLevel)
         throw;
     }
 
-    if (!isQuorumSetSane(QUORUM_SET, !UNSAFE_QUORUM))
+    const char* errString = nullptr;
+    if (!isQuorumSetSane(QUORUM_SET, !UNSAFE_QUORUM, errString))
     {
-        LOG(FATAL) << fmt::format("Invalid QUORUM_SET: check nesting, "
-                                  "duplicate entries and thresholds (must be "
-                                  "between {} and 100)",
-                                  UNSAFE_QUORUM ? 1 : 51);
+        LOG(FATAL) << fmt::format("Invalid QUORUM_SET: {}", errString);
         throw std::invalid_argument("Invalid QUORUM_SET");
     }
 }
--- a/src/scp/BallotProtocol.cpp
+++ b/src/scp/BallotProtocol.cpp
@@ -242,10 +242,13 @@ bool
 BallotProtocol::isStatementSane(SCPStatement const& st, bool self)
 {
     auto qSet = mSlot.getQuorumSetFromStatement(st);
-    bool res = qSet != nullptr && isQuorumSetSane(*qSet, false);
+    const char* errString = nullptr;
+    bool res = qSet != nullptr && isQuorumSetSane(*qSet, false, errString);
     if (!res)
     {
-        CLOG(DEBUG, "SCP") << "Invalid quorum set received";
+        CLOG(DEBUG, "SCP") << "Invalid quorum set received : "
+                           << (errString ? errString : "<empty>");
+
         return false;
     }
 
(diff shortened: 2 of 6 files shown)


src/scp/BallotProtocol.cpp
```diff
@@ -242,10 +242,17 @@ bool
...
+    const char* errString = nullptr;
+    bool res = qSet != nullptr && isQuorumSetSane(*qSet, false, errString);
     if (!res)
     {
         CLOG(DEBUG, "SCP") << "Invalid quorum set received";
+        if (errString != nullptr)
+        {
+            std::string msg(errString);
+            CLOG(DEBUG, "SCP") << msg;
+        }
```
AndrejMitrovic, line 254: Can someone suggest a way to inline this into the previous `CLOG` call?

src/main/Config.cpp
```diff
@@ -1258,12 +1258,11 @@ Config::validateConfig(ValidationThresholdLevels thresholdLevel)
...
-    if (!isQuorumSetSane(QUORUM_SET, !UNSAFE_QUORUM))
+    const char* errString = nullptr;
+    if (!isQuorumSetSane(QUORUM_SET, !UNSAFE_QUORUM, errString))
     {
-        LOG(FATAL) << fmt::format("Invalid QUORUM_SET: check nesting, "
-                                  "duplicate entries and thresholds (must be "
-                                  "between {} and 100)",
-                                  UNSAFE_QUORUM ? 1 : 51);
+        std::string msg(errString);
+        LOG(FATAL) << "Invalid QUORUM_SET: " << msg;
```
MonsieurNicolas, line 1265: why not just keep `fmt::format("Invalid QUORUM_SET: {}", errString)`

src/scp/BallotProtocol.cpp
MonsieurNicolas: I would just write this as
`CLOG(DEBUG, "SCP") << "Invalid quorum set received : " << (errString?errString:"<empty>");` and be done with it (also avoids a new variable).

MonsieurNicolas (changes requested):
thanks. just a couple suggestions

MonsieurNicolas:
Also, before we can merge your changes, please fill in the agreement per https://github.com/stellar/.github/blob/master/CONTRIBUTING.md#submitting-changes

AndrejMitrovic:
Thanks! I've filled in the agreement and applied the changes.

MonsieurNicolas:
Thank you!

MonsieurNicolas:
r+ 8e2a98896b95de94acf99a6ab92fe9e878b08695

AndrejMitrovic:
I've re-run clang formatter on the code. Sorry about that!

MonsieurNicolas:
r+ e44e88a11c22db126481d303911781db70bdd7d0

AndrejMitrovic:
It looks like Travis uses clag-format 5, and I used the one installed with brew (v10) which has a slightly different output. I'll try to get v5 installed and re-run the formatter.

AndrejMitrovic:
Installing clang-format 5 seems to be tediously slow (tried https://gist.github.com/ffeu/6ffb75d8e8c7d92c0fbeb4b036599c33), so I applied to diff manually based on the results from Travis.

Geod24:
@MonsieurNicolas : Since you got Travis set up, I wonder, why doesn't it trigger on PRs? Not having the CI trigger before approval (and only once per committer's approval) can make contributing quite tedious.

MonsieurNicolas:
@AndrejMitrovic yes this is the way to do it, the error message is an actual patch that can be applied if you don't have clang-format-5 installed. Let's see if it works this time around...

MonsieurNicolas:
@Geod24 good idea. I just enabled it for future PRs, from what I remember it was just too slow (we had it enabled a long time ago), but maybe it's not that bad now.

MonsieurNicolas:
r+ cbefaa9b1cddc2abcb765be44f4a6d6ed9724d46

Issue #2641, opened 2020-07-31
Flaky test "LedgerTxnRoot prefetch"


MonsieurNicolas:
Ran into this https://travis-ci.org/github/stellar/stellar-core/jobs/713841372

```
stellar-core test is a Catch v2.11.3 host application.
2	Run with -? for options
2	
2	Randomness seeded to: 18474
2	
2	-------------------------------------------------------------------------------
2	LedgerTxnRoot prefetch
2	  postgresql
2	  stop prefetching as cache fills up
2	-------------------------------------------------------------------------------
2	ledger/test/LedgerTxnTests.cpp:2388
2	...............................................................................
2	
2	ledger/test/LedgerTxnTests.cpp:2392: FAILED:
2	  REQUIRE( root.prefetch(keysToPrefetch) == (cfg.ENTRY_CACHE_SIZE / 2) )
2	with expansion:
2	  1000 (0x3e8) == 500 (0x1f4)
2	
2	<done>
2	===============================================================================
2	test cases:  2 |  1 passed | 1 failed
2	assertions: 22 | 21 passed | 1 failed
```

rokopt:
It's possible that the test is flaky; the production code definitely is.  There's a copy/paste error in `LedgerTxnRoot::Impl::prefetch()`:

<pre><code>
[ ... ]
        case TRUSTLINE:
            insertIfNotLoaded(trustlines, key);
            if (trustlines.size() == mBulkLoadBatchSize)
            {
                cacheResult(bulkLoadTrustLines(trustlines));
                trustlines.clear();
            }
            break;
        case DATA:
            insertIfNotLoaded(data, key);
            if (data.size() == mBulkLoadBatchSize)
            {
                cacheResult(bulkLoadData(data));
                data.clear();
            }
            break;
        case CLAIMABLE_BALANCE:
            insertIfNotLoaded(claimablebalance, key);
            if (<b>data</b>.size() == mBulkLoadBatchSize)
            {
                cacheResult(bulkLoadClaimableBalance(claimablebalance));
                claimablebalance.clear();
            }
            break;
[ ... ]
</code></pre>

rokopt:
> It's possible that the test is flaky; the production code definitely is.

The reason that I say it's possible that the test is flaky is that it reuses a reference to `getLedgerTxnRoot()` after a `commit()`:

```
        auto& root = app->getLedgerTxnRoot();

        auto entries = LedgerTestUtils::generateValidLedgerEntries(1000);
[...]

PR #2728, opened 2020-09-25
Reseed autocheck::rng() along with other RNGs

### LINKED ISSUE #2663: autocheck is not integrated with RNG seed (Version: [1.32])
Issue #2641 is an example of a bug which turned out to be intermittent even in unit tests, which use virtual time and a reusable random-number-generator seed with the intent of allowing reproducible runs.

We've traced the cause of the non-determinism to the use within the `autocheck` library of a `std::random_device` which is not connected to the RNG with the fixed seed used by the rest of `stellar-core`.  `xdrpp` uses `autocheck` to generate random XDR values, and we in turn use `xdrpp`'s generators to generate random ledger entries.  So any unit tests which generate random ledger entries are subject to non-determinism.

As an experiment, I tried replacing `autocheck`'s `std::random_device` with `stellar::gRandomEngine`, which uses the seed from the command line.  With that change, instead of, for example, seed 18474 failing 1-2% of the time during my attempts to reproduce issue #2641 , I've now seen seed 18474 pass every one of 400 runs, whereas seed 32, for example, has failed every one of 750 runs.

That encapsulation-breaking change is obviously not one we will commit, but we should find a way to do this properly, as we're trying to eliminate all the sources of non-determinism in Catch unit tests that we can find.  Possibilities include:

- Using inline namespaces
- Modifying autocheck to parameterize it on an RNG, and submitting that code upstream

rokopt:
Make ReseedPRNGListener::reseed() reseed autocheck's
mt19937 RNG along with the others that it already
reseeds (std::rand, gRandomEngine, and Catch::rng()), to
remove a source of non-determinism (one which we first
noticed because it made issue #2641 intermittent).

# Description

Resolves #2663 

We originally noticed the non-determinism resulting from our use of `autocheck` to generate pseudo-random `LedgerEntry` structures when we found that issue #2641 was intermittent.  The problem is that `autocheck::rng()` is seeded from a `std::random_device` which has no connection with `stellar-core`'s `--rng-seed`.

I originally thought that we might need to make `autocheck` share our internal RNG, `gRandomEngine`, but it occurs to me that all we need to do to make it deterministic is to seed `autocheck::rng()` at the same time as we seed `gRandomEngine`.

That's what this code does.  However, I'm actually mystified that it works, because I'd have expected the `inline` `autocheck::rng()` to have a separate instance in each translation unit, and consequently one Meyers "singleton" per translation unit.  But dumping the address of it from different translation units shows that they all use the same instance, and wrapping that instance in a class that logs in the constructor shows that it's only constructed once.

So this is, embarrassingly, a smaller change than I expected because of something that I'm misunderstanding about the C++ standard.

# Checklist
- [x] Compiles
- [x] Ran all tests

### DIFF
--- a/src/test/test.cpp
+++ b/src/test/test.cpp
@@ -5,6 +5,7 @@
 #define CATCH_CONFIG_RUNNER
 
 #include "util/asio.h"
+#include <autocheck/autocheck.hpp>
 
 #include "ledger/LedgerTxn.h"
 #include "ledger/LedgerTxnHeader.h"
@@ -56,6 +57,7 @@ struct ReseedPRNGListener : Catch::TestEventListenerBase
         srand(sCommandLineSeed);
         gRandomEngine.seed(sCommandLineSeed);
         Catch::rng().seed(sCommandLineSeed);
+        autocheck::rng().seed(sCommandLineSeed);
     }
     virtual void
     testCaseStarting(Catch::TestCaseInfo const& testInfo) override


jonjove:
Your remarks here caused me to dig deeper into the standard and clarify my understanding of the function specifier `inline` (which I almost never use). The main discussion in the C++14 standard seems to be in section 7.1.2. The essential quote is

> An inline function with external linkage shall have the same address in all translation units. A static local variable in an extern inline function always refers to the same object.

although there is a considerable amount of interesting information around it. I also found https://stackoverflow.com/questions/185624/static-variables-in-an-inlined-function which discusses the topic a bit.

jonjove:
r+ 4fb6832

rokopt:
> Your remarks here caused me to dig deeper into the standard and clarify my understanding of the function specifier `inline` (which I almost never use). The main discussion in the C++14 standard seems to be in section 7.1.2. The essential quote is
> 
> > An inline function with external linkage shall have the same address in all translation units. A static local variable in an extern inline function always refers to the same object.
> 
> although there is a considerable amount of interesting information around it. I also found https://stackoverflow.com/questions/185624/static-variables-in-an-inlined-function which discusses the topic a bit.

Thanks for digging that up!  I have learned something.

I had found some information about inline _variables_ -- see for example https://stackoverflow.com/a/38043566 -- but they came in with C++17, and I didn't realize that they were extending to variables a property that earlier standards already ensured for functions.

PR #2729, opened 2020-09-25
SCPDriver: Make hashing routine user-definable

### LINKED ISSUE #2724: Making the hash types and hashing routines user-definable in SCP
Hi,

We have a downstream change to SCP which we're considering upstreaming to `stellar-core`. We wanted to use our own hashing routines. And secondarily we wanted to define our own Hash type. In our use-case we use `BLAKE2b` and require a 64-byte hash, compared to 32 byte hashes used in stellar-core.

The way we've modified SCP is as follows. We've introduced prototype functions `getHashOf` in a new header file `Hash.h`, and then defined them on our side. Here is [the diff](https://gist.github.com/AndrejMitrovic/c374651809a3b746e18531c0a579a817).

For changing the hashes into 64 bytes we made a less flexible change, by hardcoding the type to 64-bytes. The diff [is here](https://gist.github.com/AndrejMitrovic/356280d445bc6af551049bc25a3f056c). We want to make this flexible so the client code can define it. So this part of code will need more work.

Would you be open to accepting a patch which makes the hash type and hashing routines definable by the client code?

Thanks!

AndrejMitrovic:
@MonsieurNicolas I ended up going with your [original suggestion](https://github.com/stellar/stellar-core/issues/2724#issuecomment-698039259).

Let me know if you have any suggestions. Cheers!
 
Resolves #2724

### DIFF
--- a/src/herder/HerderSCPDriver.cpp
+++ b/src/herder/HerderSCPDriver.cpp
@@ -32,6 +32,17 @@ namespace stellar
 uint32_t const HerderSCPDriver::FIRST_PROTOCOL_WITH_TXSET_CLOSETIME_AFFINITY =
     14;
 
+Hash
+HerderSCPDriver::getHashOf(std::vector<xdr::opaque_vec<>> const& vals) const
+{
+    SHA256 hasher;
+    for (auto const& v : vals)
+    {
+        hasher.add(v);
+    }
+    return hasher.finish();
+}
+
 HerderSCPDriver::SCPMetrics::SCPMetrics(Application& app)
     : mEnvelopeSign(
           app.getMetrics().NewMeter({"scp", "envelope", "sign"}, "envelope"))
--- a/src/scp/SCPDriver.cpp
+++ b/src/scp/SCPDriver.cpp
@@ -8,7 +8,6 @@
 
 #include "crypto/Hex.h"
 #include "crypto/KeyUtils.h"
-#include "crypto/SHA.h"
 #include "crypto/SecretKey.h"
 #include "xdrpp/marshal.h"
 
@@ -56,7 +55,7 @@ SCPDriver::wrapValue(Value const& value)
 std::string
 SCPDriver::getValueString(Value const& v) const
 {
-    uint256 valueHash = sha256(xdr::xdr_to_opaque(v));
+    Hash valueHash = getHashOf({xdr::xdr_to_opaque(v)});
 
     return hexAbbrev(valueHash);
 }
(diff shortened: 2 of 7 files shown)


src/scp/SCPDriver.h
```diff
@@ -63,6 +64,11 @@ class SCPEnvelopeWrapper : public NonMovableOrCopyable
 
 typedef std::shared_ptr<SCPEnvelopeWrapper> SCPEnvelopeWrapperPtr;
 
+// values used to switch hash function between priority and neighborhood checks
```
AndrejMitrovic, line 67: I wasn't sure whether to keep these here or to move them to `HerderSCPDriver` (and by extension to `SCPUnitTests.cpp`).

They're trivial enough that they could be considered noise in here. Let me know what you think.

MonsieurNicolas (changes requested):

src/herder/HerderSCPDriver.cpp
```diff
@@ -32,6 +32,52 @@ namespace stellar
 uint32_t const HerderSCPDriver::FIRST_PROTOCOL_WITH_TXSET_CLOSETIME_AFFINITY =
     14;
 
+static uint64
```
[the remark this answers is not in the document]
AndrejMitrovic: Oh yes that's not a bad idea. Thank you.

AndrejMitrovic:
@MonsieurNicolas applied your suggestions. Thank you for the help!

src/scp/SCPDriver.cpp
```diff
@@ -80,13 +79,14 @@ static const uint32 hash_K = 3;
 
 static uint64
 hashHelper(uint64 slotIndex, Value const& prev,
-           std::function<void(SHA256&)> extra)
+           std::function<void(std::vector<xdr::opaque_vec<>>&)> extra,
+           SCPDriver const& driver)
```
MonsieurNicolas, line 83: as you're in the SCPDriver class, you can simply define this helper function as a private method

MonsieurNicolas (changes requested):
minor nit, and we can merge. Thank you

AndrejMitrovic:
Updated. Hopefully the formatter doesn't complain, but since I'm still using the incompatible v10 version it might fail the build. I'll fix it up if it does.

src/scp/SCPDriver.cpp
AndrejMitrovic: Thanks. In retrospect that should have been obvious. :)

MonsieurNicolas:
r+ 726057dbbb782db71ea821115390c6e886713117

AndrejMitrovic:
Fixed the two styling nits.

MonsieurNicolas:
r+ 2a91d38731e7aac22a29b10ddbf112102a4a0a39

Issue #2719, opened 2020-09-19
Confusing behavior of RUN_STANDALONE in tests


MonsieurNicolas:
The user facing `RUN_STANDALONE` setting is used as a way to start a node without overlay enabled.

What is confusing is that internally, `RUN_STANDALONE` is more of a proxy for "not starting overlay over TCP".

Some (many) test cases, start nodes that are connected to each other via the `LoopbackPeer` class instead of TCP, and to do that, they set `RUN_STANDALONE=false` and manage peer connections from the test cases themselves.

The condition https://github.com/stellar/stellar-core/blob/49a14db54ecb15154f5a7af9b7c0d2e8b6862707/src/herder/HerderImpl.cpp#L321

was modified as to avoid triggering the rebroadcast timer when using a standalone node with manual close enabled - to avoid a busy loop that would advance virtual time, the condition was added.

We *could* just change that condition to just `MANUAL_CLOSE`, still a bit confusing nevertheless.

graydon:
I would vote for something stronger here: I'd like to deprecate `RUN_STANDALONE` as a config flag. It currently means too many things, few of which are clear from its name. I'd like to make separate flags for each of the things it means, and then leave the config setting as backward-compatibility support to flip all those settings individually.

MonsieurNicolas:
Agree on "meaning too many things" - while for the config file, we may want to keep that setting. Internally we're probably much better off splitting this into different flags.

For example, what the survey test is looking for is to be in control of the network topology (so disable most of what `tick` does to discover new peers/promote peers).

PR #2733, opened 2020-09-28
remove extra survey routes

MonsieurNicolas:
those were added by accident in 20acb825832fdc2114d238211ac1833ab91d7557

fixes a minor issue: test enabled builds had survey related endpoints always enabled

### DIFF
--- a/src/main/CommandHandler.cpp
+++ b/src/main/CommandHandler.cpp
@@ -90,10 +90,8 @@ CommandHandler::CommandHandler(Application& app) : mApp(app)
         addRoute("quorum", &CommandHandler::quorum);
         addRoute("scp", &CommandHandler::scpInfo);
         addRoute("stopsurvey", &CommandHandler::stopSurvey);
-#ifndef BUILD_TESTS
         addRoute("getsurveyresult", &CommandHandler::getSurveyResult);
         addRoute("surveytopology", &CommandHandler::surveyTopology);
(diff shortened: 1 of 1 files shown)


graydon:
Hmm, the previous hunk looks fairly intentional. I do not remember the intent -- but are you sure it's wrong?

MonsieurNicolas:
Yeah no idea why you made this change, I missed it when I r+ the PR.

Before 20acb82 survey routes were always active, with 20acb82 the behavior became conditional for non test builds, based on `RUN_STANDALONE`.

I think the change for `RUN_STANDALONE` makes sense (surveying a node without network doesn't seem useful).

Why would we not do this for test enabled builds is a bit of a mystery to me... Maybe some tests over loopback were failing? As much as possible, I prefer to keep the behavior the same between test enabled builds and regular builds.

graydon (approved):
Ok. This is a mystery to me still but maybe next time we change it we will discover the actual cause! Or maybe it was a mistake.

graydon:
r+ c95f3c9

MonsieurNicolas:
OK, mystery solved: the test `topology survey` is indeed using a loopback configuration (as to ensure that the network topology stays stable) while testing the survey endpoint.

One more to add to the list for https://github.com/stellar/stellar-core/issues/2719

I am closing this PR for now.

PR #2740, opened 2020-10-01
 Give a sensible default value for operations error code

MonsieurNicolas:
This PR fixes an edge case in how we apply transactions when replaying from history, that can cause nodes to fail catching up to the  latest ledger with a "hash mismatch" error.

The fix ensures that invoking `TransactionFrame::apply` yields the same error code on failure regardless of calling `TransactionFrame::checkValid` before (we don't call `checkValid` when replaying from history as we rely on hashes).

### DIFF
--- a/src/transactions/test/TxEnvelopeTests.cpp
+++ b/src/transactions/test/TxEnvelopeTests.cpp
@@ -408,7 +408,7 @@ TEST_CASE("txenvelope", "[tx][envelope]")
         {
             std::string name;
             bool autoRemove;
-            std::function<SignerKey(TransactionFrame const&)> createSigner;
+            std::function<SignerKey(TransactionFrame&)> createSigner;
             std::function<void(TransactionFrame&)> sign;
         };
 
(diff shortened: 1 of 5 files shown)


src/transactions/test/TxEnvelopeTests.cpp
```diff
@@ -1933,32 +1942,73 @@ TEST_CASE("txenvelope", "[tx][envelope]")
...
+                                      expectedResult(baseFee * 2, 2, txSUCCESS,
+                                                     {SET_OPTIONS_SUCCESS,
+                                                      SET_OPTIONS_SUCCESS}));
+                });
+                }
+                SECTION("missing signature")
+                {
+                    for_versions_from(10, *app, [&] {
+                        auto tx =
+                            a.tx({payment(root, 1000),
```
jonjove, line 1980: Nitpick: Formatting here is really weird, is this what clang-format produced?

src/transactions/TransactionFrame.cpp
```diff
@@ -66,6 +66,12 @@ TransactionFrame::getFullHash() const
 Hash const&
 TransactionFrame::getContentsHash() const
 {
+#ifdef _DEBUG
+    // force recompute
+    Hash oldHash;
+    std::swap(mContentsHash, oldHash);
```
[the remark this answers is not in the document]
MonsieurNicolas: couple things:
* `_DEBUG` is not defined anywhere, so it's not enabled in any build by default. I was planning on maybe adding this to `extrachecks` at some point, but there are tradeoffs.
* "a bug in caching" as in we don't cache (so perf impact?)?
   * What is a lot more likely (and allowed me to find many bugs in tests) is that people mutate state even though we pre-computed the hash (ie, they forgot to call `clearCached` for example). With this change you get an assert the next time you try to access the supposedly cached hash instead of returning a stale value for the hash.

In any case I don't want to keep discussing this further in this PR - I am removing the commit that I thought was non controversial because of the dependency on `_DEBUG` and will open a different PR with that change.

src/transactions/test/TxEnvelopeTests.cpp
MonsieurNicolas: very strange indeed - yeah it was a clang-format issue

jonjove:
r+ 13f4697

Issue #2741, opened 2020-10-01
Implementation-dependent choice of `std::uniform_int_distribution` introduces test non-determinism across configurations (Version: [14.1])


rokopt:
While investigating issue #2739 , which we found while invesigating issue #2736 , which we encountered while investigating issue #2730 , in a branch with the fix for issue #2663 , @sisuresh and I found another source of non-determinism between test runs with the same RNG seed but built against different libraries:  `std::uniform_int_distribution`.  It produces different values for the two of us given the same RNG.

Siddarth found the following relevant post:

https://stackoverflow.com/questions/32746092/random-output-different-between-implementations/32746283#32746283

He also wrote the following unit test:

```
#include <random>
#include <iostream>
int
main(void)
{
    std::mt19937 rand(0);
    std::uniform_int_distribution<uint32_t> dist(
        1, 10
    );
    for (auto i = 0; i < 10; ++i)
    	std::cout << dist(rand) << std::endl;
}
```

He gets:
```
6
1
4
4
8
10
4
6
3
5
```

I get:
```
6
6
8
9
7
9
6
9
5
7
```

Issue #2739, opened 2020-10-01
Implementation-dependent choice of `std::rand()` introduces test non-determinism across configurations (Version: [14.1])


rokopt:
While investigating issue #2736 , which we encountered while investigating issue #2730 , in a branch with the fix for issue #2663 , @sisuresh and I found another source of non-determinism between test runs with the same RNG seed but built against different libraries:  `std::rand()`.  We reseed that with `std::srand()` in `ReseedPRNGListener`, but across different libraries, `std::rand()` isn't guaranteed to return the same sequence given the same seed.  We don't appear to have a huge number of uses, but I found these:

`src/overlay/PeerManager.cpp`:
```
static std::chrono::seconds
computeBackoff(int numFailures)
{
[ ... ]
    auto nsecs = std::chrono::seconds(
        std::rand() % int(std::pow(2, backoffCount) * SECONDS_PER_BACKOFF) + 1);
[ ... ]
}
```

`src/scp/test/SCPTests.cpp`:
```
static void
setupValues()
{
    std::vector<Value> v;
    std::string d = fmt::format("SEED_VALUE_DATA_{}", std::rand());
[ ... ]
}
```

`src/crypto/test/CryptoTests.cpp`:
```
TEST_CASE("StrKey tests", "[crypto]")
{
    std::regex b32("^([A-Z2-7])+$");
    std::regex b32Pad("^([A-Z2-7])+(=|===|====|======)?$");

    autocheck::generator<std::vector<uint8_t>> input;

    auto randomB32 = []() {
        char res;
        char d = static_cast<char>(std::rand() % 32);
[ ... ]
}
```

`src/bucket/test/BucketTests.cpp`:
```
        SECTION("random live entries overwrite live entries in any order")
        {
[ ... ]
            std::random_shuffle(live.begin(), live.end());
```
(`std::random_shuffle()` uses `std::rand()` internally.)

PR #2742, opened 2020-10-02
Explicitly select RNG type for gRandomEngine

### LINKED ISSUE #2736: Implementation-dependent choice of `std::default_random_engine` introduces test non-determinism across configurations (Version: [14.1])
While investigating issue #2730 using a branch containing the fix for #2663 , @sisuresh and I discovered a further source of non-determinism.  Issue #2730 always reproduced for me with `--rng-seed=6` (and never with previous seeds).  That parameter, however, did not reproduce the bug for Siddharth.

He discovered, however, that `--rng-seed=1` _did_ repeatably reproduce the bug for him, which it did _not_ for me.

Eventually we discovered that both of our C++ standard libraries had the same definitions of two linear congruential engines called `minstd_rand` and `minstd_rand0` -- but his library defined `std::default_random_engine` to be the former, but mine defined it to be the latter.

So perhaps to remove another source of non-determinism -- this time not within a single build, but across builds using different libraries -- we should replace our uses of `std::default_random_engine` with something like a `stellar::default_random_engine` that explicitly selects either `minstd_rand` or `minstd_rand0` (or something else, such as a Mersenne twister).

rokopt:
# Description
 
Resolves #2736

One of a series of sources of unit test non-determinism (in some cases within a single build, in other cases only across builds against different libraries -- this bug is one of the latter kind) that @sisuresh and I discovered is that `std::default_random_engine` is not defined to the same RNG by all C++ libraries.  In particular, the one that Siddharth was using `typedef`s it to `std::minstd_rand`, whereas mine uses `std::minstd_rand0`.

The fix that I am proposing here makes our `stellar::gRandomEngine` explicitly into a `std::minstd_rand`, rather than being whatever `std::default_random_device` turns out to be.

Note that even given this fix and the fix for issue #2663 , we still have non-determinism across different libraries from other sources:  the two I know of are #2739 and #2741 .

# Checklist
- [x] Compiles
- [x] Ran all tests

### DIFF
--- a/src/crypto/SecretKey.cpp
+++ b/src/crypto/SecretKey.cpp
@@ -155,7 +155,7 @@ SecretKey::random()
 
 #ifdef BUILD_TESTS
 static SecretKey
-pseudoRandomForTestingFromPRNG(std::default_random_engine& engine)
+pseudoRandomForTestingFromPRNG(stellar_default_random_engine& engine)
 {
     std::vector<uint8_t> bytes;
     for (size_t i = 0; i < crypto_sign_SEEDBYTES; ++i)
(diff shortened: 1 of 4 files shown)


MonsieurNicolas:
needs to be rebased - it looks like it contains a bunch of random commits from somewhere else

rokopt:
Rebased.

MonsieurNicolas:
r+ 9634041248316d8b62f9e1c14eb2734ad6dbd3a3

PR #2743, opened 2020-10-02
Add stale content hash detector for TransactionFrame

MonsieurNicolas:
Allows to systematically catch illegal use of `TransactionFrame`.

In prod code we (should) never modify envelopes, but in test it's a common pattern that can produce bad tests, sequences such as:

1. create a tx
2. sign tx
3. modify tx (like change the sequence number is a typical one)
4. check tx's validity -- this step uses the stale hash from step 2, and signatures are considered valid (which is incorrect)

I enabled this under `_DEBUG`
*  it's a typical macro that people use to enable "code that may be a lot slower",  but maybe we could use a (safer?) macro name that allows for more granular activation. It's "friend" `NDEBUG` is typically *not* set and not setting it typically has minimal performance implications (but still causes `assert` to crash the app).
* we can also enable it with "extrachecks" so that checked builds have this safeguard enabled as well.

### DIFF
--- a/src/transactions/TransactionFrame.cpp
+++ b/src/transactions/TransactionFrame.cpp
@@ -66,6 +66,12 @@ TransactionFrame::getFullHash() const
 Hash const&
 TransactionFrame::getContentsHash() const
 {
+#ifdef _DEBUG
+    // force recompute
+    Hash oldHash;
+    std::swap(mContentsHash, oldHash);
+#endif
+
     if (isZero(mContentsHash))
     {
         if (mEnvelope.type() == ENVELOPE_TYPE_TX_V0)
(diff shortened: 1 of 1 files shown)


jonjove:
I'm happy to merge this as-is, since it would not be enabled by default. I have a slight preference for using a more descriptive define for it rather than the catch-all `_DEBUG`. This would allow us to pick and choose which debugging facilities are enabled if we wanted, and we could always add a configure flag `--enable-all-debugging` which just turns them all on. I'm not very picky about this, so just tell me which way you want to go and I'll merge it.

MonsieurNicolas:
I think that we can split what is under `_DEBUG` later if we end up having to distinguish them.

Ideally we would have this enabled by "extra checks" and not have to worry about yet another compile flags, but I think this one just has too much perf impact.

Actually maybe the superior option is to just add a new global variable behind `BUILD_TESTS` that performs those additional checks: that way we never perform those checks on "real" nodes, but have all checks enabled when running tests (like an invariant).

A third possibility would be to make this fast enough that it can be enabled by "extra checks", for this I think we could use short hash for that logic (so the hot code that hit the cache would only compute SHA256 once).

jonjove:
r+ 8837615

Issue #2560, opened 2020-06-01
subcommands that take file name as input can't be used to parse standard input


MonsieurNicolas:
It looks like some subcommands require a file name to be passed in:

```
echo 'sdkjsjksjd' | /usr/bin/stellar-core print-xdr
FILE-NAME argument is required

pretty-print one XDR envelope, then quit

usage:
  stellar-core print-xdr <FILE-NAME> options

where options are:
  -?, -h, --help            display usage information
  --filetype <FILE-TYPE>    [auto|ledgerheader|meta|result|resultpair|tx
                            |txfee]
  --base64                  use base64
```

but there is no way to consume from standard input.

A few options:
* make the file name optional, defaulting to standard input (probably the easiest)
* make it possible to pass in `-` as file name. This is a bit more work because `-` is recognized as a bad option, and trying to invoke `stellar-core print-xdr --base64 -- -` doesn't seem to be supported by our parser (`--` is the standard for "stop parsing options from this point").

MonsieurNicolas:
Actually, this is much worst than I originally thought, we use `-` in a lot of places (grep for `"-"`), not just as main argument but also as argument for certain options, like `--conf` (when doing things like `cat stellar.cfg | ./stellar-core --conf - offline-info`).

Fixing the "clara" library (that is now [Lyra](https://github.com/bfgroup/Lyra)) seems to be pretty involved.

The alternative is to use a different string for defaulting to "stdin":
* centralize where that special string is defined, right now it's all over the place (so we'd be left with just docs and that central place for code)
* pick a new string instead of `"-"`. `"/dev/stdin"` may be a good candidate as it happens to be the path to stdin on many systems (and on Windows it's fine to use)

hidenori-shinohara:
For future reference: https://github.com/catchorg/Clara/issues/57

MonsieurNicolas:
@hidenori-shinohara as I mentioned above, that fork is not maintained anymore, so we'd have to switch to "Lyra" *and* fix the bug there (if it's not fixed), but either task doesn't seem worth doing.

I was proposing using the special name `/dev/stdin` as an alternative instead of `-` - which is weird but seems like a lot less work. What do you think (maybe others have a different opinion?)?

hidenori-shinohara:
[...]

PR #2744, opened 2020-10-02
Use "/dev/stdin" as a special name for `stdin` in command line arguments

hidenori-shinohara:
# Description

When "/dev/stdin" is given as a special name instead of a file name, the program reads stdin instead of a file.

Resolves https://github.com/stellar/stellar-core/issues/2560

# Checklist
- [x] Compiles
- [x] Ran all tests

### DIFF
--- a/src/main/Config.h
+++ b/src/main/Config.h
@@ -413,5 +413,9 @@ class Config : public std::enable_shared_from_this<Config>
 
     // function to stringify a quorum set
     std::string toString(SCPQuorumSet const& qset);
+
+    // A special name to be used for stdin in stead of a file name in command
+    // line arguments.
+    static std::string const STDIN_SPECIAL_NAME;
 };
 }
--- a/docs/software/commands.md
+++ b/docs/software/commands.md
@@ -8,7 +8,7 @@ stellar-core can be controlled via the following commands.
 ## Common options
 Common options can be placed at any place in the command line.
 
-* **--conf <FILE-NAME>**: Specify a config file to use. You can use '-' and
+* **--conf <FILE-NAME>**: Specify a config file to use. You can use '/dev/stdin' and
   provide the config file via STDIN. *default 'stellar-core.cfg'*
 * **--ll <LEVEL>**: Set the log level. It is redundant with `http-command ll`
   but we need this form if you want to change the log level during test runs.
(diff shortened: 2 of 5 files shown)


src/main/CommandLineSpecialNames.h
```diff
@@ -0,0 +1,12 @@
+#pragma once
```
MonsieurNicolas, line 1: Maybe we can avoid adding a new file and use `Config.h` for this? We already have a few constants defined there like `MAXIMUM_QUORUM_NESTING_LEVEL`

MonsieurNicolas (changes requested):
looks pretty close. It was much simpler than I thought :)

src/main/CommandLineSpecialNames.h
hidenori-shinohara: Makes sense! I switched to using `Config.h`.

src/main/Config.h
```diff
@@ -414,4 +414,7 @@ class Config : public std::enable_shared_from_this<Config>
     // function to stringify a quorum set
     std::string toString(SCPQuorumSet const& qset);
 };
+// A special name to be used for stdin in stead of a file name in command line
+// arguments.
+const std::string stdinSpecialName = "/dev/stdin";
```
MonsieurNicolas, line 419: define this in the `Config` namespace - see `POSSIBLY_PREFERRED_EXTRA` ; also not sure, but you may run into a similar link error that I did in https://github.com/stellar/stellar-core/pull/2745 where I had to assign the value in the cpp file (link error only on some versions of clang it looks like)

MonsieurNicolas: also, our naming convention for constants would make this look more like `STDIN_SPECIAL_NAME`

MonsieurNicolas (changes requested):

src/main/Config.h
hidenori-shinohara: Updated the constant name to `STDIN_SPECIAL_NAME` and moved it inside the `Config` namespace. Just to be safe, I also assigned the value in the cpp file.

MonsieurNicolas:
r+ acba3c3b9296287102c8b8844cd332a8d17872db

PR #2745, opened 2020-10-05
reduce the number of historical SCP messages we ask from peers

MonsieurNicolas:
This PR changes how many historical SCP messages (that a node doesn't directly need) get fetched from new peers.

Before this change, we would ask `MAX_SLOTS_TO_REMEMBER` older slots which can be arbitrary large, instead:
* we only ask for a few ledgers before lcl. basic assumptions here are that
   * either the local node has been in sync for a while, and as it receives SCP messages to externalize (so the extra data is unlikely to help).
   * or the local node is behind/was restarted recently
       * in that case other nodes that need those messages are likely connected to other peers that have not been restarted recently or are in sync (so the extra data again doesn't help them).
       * we're left with an edge case where many nodes get restarted in a short period of time, and for this we just need to have a reasonable estimate of how far behind those nodes are behind the quorum.
* we also avoid asking for data for ledgers that are too old
   * the old code made no effort to avoid asking for data that is clearly out of bound. This is mitigated by the smaller window, but I've added an extra guard in there. Note: we can probably still ask for out of bound messages as there is an inherent race where we can potentially close ledgers faster than the roundtrip lag to retrieve historical data from peers, but at least this should reduce the noise.

### DIFF
--- a/src/herder/Herder.h
+++ b/src/herder/Herder.h
@@ -121,6 +121,9 @@ class Herder
     // and local state
     virtual uint32_t getCurrentLedgerSeq() const = 0;
 
+    // return the smallest ledger number we need messages for when asking peers
+    virtual uint32 getMinLedgerSeqToAskPeers() const = 0;
+
     // Return the maximum sequence number for any tx (or 0 if none) from a given
     // sender in the pending or recent tx sets.
     virtual SequenceNumber getMaxSeqInPendingTxs(AccountID const&) = 0;
--- a/src/herder/Herder.h
+++ b/src/herder/Herder.h
@@ -58,6 +58,10 @@ class Herder
 
     static std::unique_ptr<Herder> create(Application& app);
 
+    // number of additional ledgers we retrieve from peers before our own lcl,
+    // this is to help recover potential missing SCP messages for other nodes
+    static uint32 const SCP_EXTRA_LOOKBACK_LEDGERS;
+
     enum State
     {
         HERDER_SYNCING_STATE,
--- a/src/herder/HerderImpl.cpp
+++ b/src/herder/HerderImpl.cpp
@@ -471,7 +471,7 @@ HerderImpl::checkCloseTime(SCPEnvelope const& envelope, bool enforceRecent)
 }
 
 uint32_t
-HerderImpl::getMinLedgerSeqToRemember()
+HerderImpl::getMinLedgerSeqToRemember() const
 {
     auto maxSlotsToRemember = mApp.getConfig().MAX_SLOTS_TO_REMEMBER;
     auto currSlot = getCurrentLedgerSeq();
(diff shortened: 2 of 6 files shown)


src/overlay/Peer.cpp
```diff
@@ -1216,20 +1216,25 @@ Peer::recvAuth(StellarMessage const& msg)
...
-    // possible) we need to ask for older slots in case other peers need
-    // messages we didn't need ourself
+    // this requests SCP data for all slots after `low`
+    // we ask for a few older slots before what our own lcl, in case other peers
+    // need messages we didn't need ourself
     auto low = mApp.getLedgerManager().getLastClosedLedgerNum() + 1;
-    auto maxSlots = mApp.getConfig().MAX_SLOTS_TO_REMEMBER;
+    auto maxSlots = std::min<uint32>(mApp.getConfig().MAX_SLOTS_TO_REMEMBER,
+                                     Config::SCP_EXTRA_LOOKBACK_LEDGERS);
     if (low > maxSlots)
```
marta-lokhova, line 1225: we have this exact same logic in `HerderImpl::getMoreSCPState`, which is called when a node goes out of sync. Should the new calculation be used there as well? (and if so, can we refactor it to avoid duplication?)

marta-lokhova (commented):
Re: edge case 
> we're left with an edge case where many nodes get restarted in a short period of time, and for this we just need to have a reasonable estimate of how far behind those nodes are behind the quorum.

I understand the edge case, but I'm not sure I understand the mitigation. How are we using this estimate?

src/overlay/Peer.cpp
MonsieurNicolas: ah good catch, I didn't see we had this duplicated

src/herder/Herder.h
```diff
@@ -121,6 +121,9 @@ class Herder
     // and local state
     virtual uint32_t getCurrentLedgerSeq() const = 0;
 
+    // return the smallest ledger number we need messages for when asking peers
+    virtual uint32 getLowNeedsSCPLedgerSeq() const = 0;
```
marta-lokhova, line 129: The name is a bit confusing, how about something like `getMinLedgerSeqToAskPeers`?

src/herder/HerderImpl.cpp
```diff
@@ -892,6 +892,27 @@ HerderImpl::getCurrentLedgerSeq() const
     return res;
 }
 
+uint32
+HerderImpl::getLowNeedsSCPLedgerSeq() const
+{
+    // computes the smallest ledger for which we *think* we need more SCP
+    // messages
+    // we ask for messages older than the latest ledger in case they have SCP
```
marta-lokhova, line 901: can we replace "latest ledger" with "lcl" for clarity?

marta-lokhova (approved):

src/herder/Herder.h
MonsieurNicolas: done

src/herder/HerderImpl.cpp
MonsieurNicolas: done

marta-lokhova:
r+ d189c36081ba64cedde9d1aae42bed8c0f604dad

marta-lokhova:
r+ 14229bf8c78542e29d3c471c5acaf98bf1260608

marta-lokhova:
r+ 181af3c290f225cf048c13f2c79f0514a8f16486

PR #2746, opened 2020-10-06
Update maxCommit for possible quorum intersection check performance improvements 

hidenori-shinohara:
# Description

This PR will reduce maxCommit by 1 in the quorum intersection check algorithm.

## Correctness
Given a network with `N` nodes, the quorum intersection check algorithm essentially enumerates all the subsets of size `<= (N/2) + 1` and checks

* if the subset is a quorum, and
* if the complement contains a quorum.

It actually suffices to enumerate all the subsets of size `<= N / 2` because if there's a pair of disjoint quorums `A, B` with `|A| <= |B|`, then `|A| <= floor(N/2)`. (Otherwise, `|B| >= |A| >= floor(N/2) + 1` which implies `|A| + |B| >= 2 * (floor(N/2) + 1) = 2 * floor(N/2) + 2 > N`, which is a contradiction.)

## Possible performance improvements

This change will not make the program slower as we are reducing the search space.

Given a network of `N` nodes, our search space consists of

* `choose(N, 1)` subsets with 1 element,
* `choose(N, 2)` subsets with 2 elements,
* `choose(N, 3)` subsets with 3 elements,
* ...
* `choose(N, N/2 + 1)` subsets with `N/2 + 1` elements.

This change will get rid of the necessity of checking `choose(N, N/2 + 1)`. This could potentially be huge. For instance, when `N = 30`, the search space contains 759,852,346 sets, and this change would remove 145,422,675 sets, which is roughly 19%.

Furthermore, I did a quick verification in https://github.com/stellar/stellar-core/commit/94bbeebee2aca906bbeaeff58f00c16054966030, which shows that in one of the existing tests, we explore 40,310 sets, but this change would remove 17,722(=43.9643%) of them. (This number doesn't exactly match with the number above for any `N` because we have other early exits.)

# Checklist
- [x] Compiles
- [x] Ran all tests

### DIFF
--- a/src/herder/QuorumIntersectionCheckerImpl.cpp
+++ b/src/herder/QuorumIntersectionCheckerImpl.cpp
@@ -182,7 +182,7 @@ MinQuorumEnumerator::pickSplitNode() const
 size_t
 MinQuorumEnumerator::maxCommit() const
 {
-    return (mScanSCC.count() / 2) + 1;
+    return mScanSCC.count() / 2;
 }
 
 MinQuorumEnumerator::MinQuorumEnumerator(
--- a/src/herder/QuorumIntersectionCheckerImpl.h
+++ b/src/herder/QuorumIntersectionCheckerImpl.h
@@ -172,7 +172,7 @@
 // ======================================================
 //
 // The first early exit is easy: just stop expanding when you get half-way
-// through the space (or precisely: at sets larger than MAXSZ = (#N/2) + 1)
+// through the space (or precisely: at sets larger than MAXSZ = #N/2)
 // because the problem is symmetric: any potential failing subset C with size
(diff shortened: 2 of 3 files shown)


hidenori-shinohara:
I did a quick performance test using `quorum intersection scaling test` in `src/herder/test/QuorumIntersectionTests.cpp` by running `time (src/stellar-core test 'quorum intersection scaling test')`.

I ran it 5 times on my local machine and

* Master branch => Average = 3.20 seconds, Median = 3.02 seconds.
* With this change =>  Average = 1.76 seconds, Median = 1.78 seconds.

which means that this change reduced the run time by around 40-45% in this particular test case.

MonsieurNicolas:
r+ f00ad3e8b767a9c19eff51f34bf51abb8dba6533

PR #2747, opened 2020-10-07
overlay: properly connect to preferred peers

MonsieurNicolas:
The change in b11afcc51e56d76ca63032445a46b67da156359c ended up breaking preferred peer logic:
in that change, preferred peers would only get connected to as replacement for existing connections.

This change re-introduces the proper behavior (and I think this version is easier to understand than the original one).

This was caught by our acceptance test: all known peers were set as "preferred" in those tests.


marta-lokhova:
r+ 5ffdb58d7d043ca024be07d8fdde67e69463837b

PR #2754, opened 2020-10-10
properly quote test spec for make check

MonsieurNicolas:
This fixes an issue where if you try to specify a test spec like
`export TEST_SPEC='[history][catchup][acceptance]'`

then `make check` doesn't find any tests to run.


jonjove:
r+ 3c95f54

PR #2756, opened 2020-10-13
Out of sync timer cleanup

MonsieurNicolas:
# Description

This PR changes the behavior related to re-broadcasting of SCP messages and asking peers for potential missing data:
* the main change is to associate this behavior to the "out of sync" event that is triggered when a node does not reach consensus
    * previously nodes would constantly rebroadcast SCP messages even though this should not be necessary
* a more aggressive change we can consider later on would be to remove this code entirely, but for now I left it as it can help in situations where nodes start up all at once. "pull mode" may render this obsolete anyways.

Overall this should reduce the number of duplicate SCP messages being sent in "steady state" (when the network is "in sync").


marta-lokhova (approved):

MonsieurNicolas:
r+ 1505d621c52aaa45aba14f6aeae70b4ffe10c02d

Issue #2749, opened 2020-10-08
Unnecessary Nesting when Printing XDR


jonjove:
I used print-xdr to print a transaction envelope containing a `MANAGE_DATA` operation. The output contained
```
"dataName": {
    "value0": "TestName"
},
```
where I would have expected
```
"dataName": "TestName",
```

jonjove:
The same thing happens with `MEMO_TEXT`. For example, if you use this very old (from ledger 3!) transaction https://horizon.stellar.org/transactions/3389e9f0f1a65f19736cacf544c2e825313e8447f569233bb8db39aa607c8889 you get
```
                "memo": {
                    "type": "MEMO_TEXT",
                    "text": {
                        "value0": "hello world"
                    }
                },
```

PR #2758, opened 2020-10-14
Fix some formatting issues when printing XDR

hidenori-shinohara:
# Description

Resolves https://github.com/stellar/stellar-core/issues/2749

This PR fixes some issues with printing XDR as described in https://github.com/stellar/stellar-core/issues/2749 by implementing `cereal_override` for `xstring` and `Memo`.

`AAAAAgAAAAAfpR7UY4+Podv9E4/IgUImlA8y6OKEQi8lJ8rRR8wMiAAAAGQAAAAAAodXoAAAAAEAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAEAAAAAAAAACgAAAA50aGlzaXNkYXRhbmFtZQAAAAAAAQAAAA90aGlzaXNkYXRhdmFsdWUAAAAAAAAAAAA=` contains `xstring` and this will be printed as `"dataName": "thisisdataname",`.

Similarly, `AAAAAAGUcmKO5465JxTSLQOQljwk2SfqAJmZSG6JH6wtqpwhAAABLAAAAAAAAAABAAAAAAAAAAEAAAALaGVsbG8gd29ybGQAAAAAAwAAAAAAAAAAAAAAABbxCy3mLg3hiTqX4VUEEp60pFOrJNxYM1JtxXTwXhY2AAAAAAvrwgAAAAAAAAAAAQAAAAAW8Qst5i4N4Yk6l+FVBBKetKRTqyTcWDNSbcV08F4WNgAAAAAN4Lazj4x61AAAAAAAAAAFAAAAAAAAAAAAAAAAAAAAAQAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAABLaqcIQAAAEBKwqWy3TaOxoGnfm9eUjfTRBvPf34dvDA0Nf+B8z4zBob90UXtuCqmQqwMCyH+okOI3c05br3khkH0yP4kCwcE` contains `Memo` and it will be printed as `"text": "hello world"`.

 

# Checklist
- [x] Compiles
- [x] Ran all tests

### DIFF
--- a/src/util/XDRCereal.h
+++ b/src/util/XDRCereal.h
@@ -19,6 +19,14 @@ using namespace std::placeholders;
 
 template <uint32_t N>
 void
+cereal_override(cereal::JSONOutputArchive& ar, const xdr::xstring<N>& s,
+                const char* field)
+{
+    xdr::archive(ar, static_cast<std::string const&>(s), field);
+}
+
+template <uint32_t N>
+void
 cereal_override(cereal::JSONOutputArchive& ar, const xdr::opaque_array<N>& s,
                 const char* field)
 {


src/util/XDRCereal.h
```diff
@@ -17,6 +17,16 @@
 
 using namespace std::placeholders;
 
+template <uint32_t N>
+void
+cereal_override(cereal::JSONOutputArchive& ar, const xdr::xstring<N>& s,
+                const char* field)
+{
+    std::stringstream out;
```
MonsieurNicolas, line 25: I don't think you need the extra buffer as all you want is to use the `std::string` flavor.

I think you just need this:

```
template <uint32_t N>
void
cereal_override(cereal::JSONOutputArchive& ar, const xdr::xstring<N>& s,
                const char* field)
{
    xdr::archive(ar, static_cast<std::string const&>(s), field);
}
```

src/util/XDRCereal.cpp
```diff
@@ -42,3 +45,51 @@ cereal_override(cereal::JSONOutputArchive& ar, const stellar::Asset& s,
 {
     xdr::archive(ar, stellar::assetToString(s), field);
 }
+
+void
+cereal_override(cereal::JSONOutputArchive& ar, const stellar::Memo& m,
```
MonsieurNicolas, line 50: I don't think you need this override:
`Memo` contains `xstring` so the other change should take care of it.

MonsieurNicolas (changes requested):

src/util/XDRCereal.h
hidenori-shinohara: That's much simpler! Fixed.

src/util/XDRCereal.cpp
hidenori-shinohara: Yes, you're right. I started with this change, so I didn't notice that the `xstring` stuff would take care of this! Fixed.

MonsieurNicolas:
r+ 55e9962ed36e05ee93985caf2b71dc2478d2c895

Issue #2757, opened 2020-10-14
deadlock in `WriteVerifiedCheckpointHashesWork::yieldMoreWork()`


MonsieurNicolas:
Looks like the code to generate trusted checkpoints doesn't work anymore.

Here is what I tried:

version: 74bccaa6cf778affd3086390c2b67227640290e9

`stellar-core --conf testnet.cfg verify-checkpoints --output-file /tmp/trusted`

it deadlocks in `WriteVerifiedCheckpointHashesWork::yieldMoreWork()` while constructing `VerifyLedgerChainWork::VerifyLedgerChainWork` because it tries to resolve a future that is not done yet (but that future will be completed by the main thread later on).

Not sure if I was doing something unexpected - if so, we should add some guards to fast fail

PR #2760, opened 2020-10-15
Fix deadlock in verify-checkpoints command

graydon:
# Description

The deadlock in bug #2757 was just mistakenly treating `.valid()` as meaning "future is ready" when guarding an assert.

Existing testcase didn't trap this because while I _thought_ it did 5 inner works because it did 5 checkpoints (and thus exercised the inter-work propagation code) I forgot that I also did batching for parallelism inside the outer work (to keep the network saturated with downloads) and so all 5 checkpoints were verified by a single inner work, this actually wasn't testing the inter-work propagation code on the inner works at all. So I changed the outer work to allow changing the batching factor (to avoid having to make the test run a long time) and then fixed the test to run 5 * the batching factor.

Once the test is corrected, the fix is straightforward: just add a helper called `futureIsReady` that does the (honesty somewhat silly) wait-for-a-nanosecond thing one has to do to check future readiness, and use it instead. Also replace other places in the code we do the same silly wait.

### DIFF
--- a/src/catchup/VerifyLedgerChainWork.cpp
+++ b/src/catchup/VerifyLedgerChainWork.cpp
@@ -9,6 +9,8 @@
 #include "main/Application.h"
 #include "main/ErrorMessages.h"
 #include "util/FileSystemException.h"
+#include "util/GlobalChecks.h"
+#include "util/Thread.h"
 #include "util/XDRStream.h"
 #include "util/types.h"
 #include <Tracy.hpp>
@@ -107,11 +109,6 @@ VerifyLedgerChainWork::VerifyLedgerChainWork(
     , mVerifyLedgerChainFailure(app.getMetrics().NewMeter(
           {"history", "verify-ledger-chain", "failure"}, "event"))
 {
-    if (trustedMaxLedger.valid())
-    {
-        assert(range.mCount == 0 ||
-               range.last() == (trustedMaxLedger.get()).first);
-    }
     // LCL should be at-or-after genesis and we should have a hash.
     assert(lastClosedLedger.first >= LedgerManager::GENESIS_LEDGER_SEQ);
     assert(lastClosedLedger.second);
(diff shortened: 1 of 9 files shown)


src/historywork/test/HistoryWorkTests.cpp
```diff
@@ -38,7 +44,12 @@ TEST_CASE("write verified checkpoint hashes", "[historywork]")
         catchupSimulation.getClock().crank();
     }
 
-    Hash h = WriteVerifiedCheckpointHashesWork::loadHashFromJsonOutput(
-        pair.first, file);
-    REQUIRE(h == *pair.second);
+    for (auto const& p : pairs)
+    {
+        LOG(INFO) << "Verified " << p.first << " with hash "
```
MonsieurNicolas, line 49: avoid default logging when running tests (this overwhelms Travis) : either use `DEBUG` level or use `INFO("Verified " << p.first << " with hash " << hexAbbrev(*p.second))` (as to give context in case of failure)

src/catchup/VerifyLedgerChainWork.cpp
```diff
@@ -107,7 +108,7 @@ VerifyLedgerChainWork::VerifyLedgerChainWork(
     , mVerifyLedgerChainFailure(app.getMetrics().NewMeter(
           {"history", "verify-ledger-chain", "failure"}, "event"))
 {
-    if (trustedMaxLedger.valid())
+    if (futureIsReady(trustedMaxLedger))
```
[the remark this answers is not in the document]
graydon: Fair enough. Fixed.

src/historywork/test/HistoryWorkTests.cpp
graydon: Fixed.

graydon:
Pushed update with fixes to comments.

MonsieurNicolas:
r+ b31a33498ac7f26db729565712060e478c373875

PR #2762, opened 2020-10-19
Sync filesystem changes after subprocess exit

graydon:
This changes the behaviour of the process manager to call `sync(2)` when a subprocess exits.

There are two purposes to this:

  - First -- the motivating purpose -- is to reduce spikes in dirty pages (still being written to disk) that happen when running large-ish test workloads on kubernetes; these spikes seem to provoke oom-kill events from the underlying kernel at unfortunate frequency.
  - Second, it may also close a small node-local data corruption gap should there be a power loss or node reboot during such a period of transient data. This is unlikely (we only usually run a slew of `get` commands once during catchup) but it's a possible failure mode and it's always nice to remove a few more of those.

As with the fsync-on-xdr-stream-close change, this change is accompanied by a config variable you can use to disable the behaviour should you need to (eg. on a super-slow spinning disk) but it seems even less likely to ever be needed here.

### DIFF
--- a/src/process/ProcessManagerImpl.cpp
+++ b/src/process/ProcessManagerImpl.cpp
@@ -259,7 +259,6 @@ ProcessManagerImpl::tryProcessShutdown(std::shared_ptr<ProcessExitEvent> pe)
 
 ProcessManagerImpl::ProcessManagerImpl(Application& app)
     : mMaxProcesses(app.getConfig().MAX_CONCURRENT_SUBPROCESSES)
-    , mSyncFilesystemOnProcessExit(!app.getConfig().DISABLE_SUBPROCESS_SYNC)
     , mIOContext(app.getClock().getIOContext())
     , mSigChild(mIOContext)
     , mTmpDir(
@@ -516,7 +515,6 @@ ProcessManagerImpl::forceShutdown(ProcessExitEvent& pe)
 
 ProcessManagerImpl::ProcessManagerImpl(Application& app)
     : mMaxProcesses(app.getConfig().MAX_CONCURRENT_SUBPROCESSES)
-    , mSyncFilesystemOnProcessExit(!app.getConfig().DISABLE_SUBPROCESS_SYNC)
     , mIOContext(app.getClock().getIOContext())
     , mSigChild(mIOContext, SIGCHLD)
     , mTmpDir(
@@ -566,11 +564,6 @@ ProcessManagerImpl::handleSignalWait()
             const int pid = std::get<0>(pidStatus);
             const int status = std::get<1>(pidStatus);
             handleProcessTermination(pid, status);
-        }
-        if (mSyncFilesystemOnProcessExit)
-        {
-            // Sync filesystem changes caused by processes.
-            sync();
         }
     }
     startSignalWait();
--- a/src/process/ProcessManagerImpl.cpp
+++ b/src/process/ProcessManagerImpl.cpp
@@ -34,6 +34,7 @@
 #else
 #include <errno.h>
 #include <fcntl.h>
+#include <unistd.h>
 #endif
 
 #if defined(__APPLE__) || defined(__FreeBSD__)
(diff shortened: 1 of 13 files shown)


src/process/ProcessManagerImpl.cpp
```diff
@@ -564,6 +567,11 @@ ProcessManagerImpl::handleSignalWait()
             const int status = std::get<1>(pidStatus);
             handleProcessTermination(pid, status);
         }
+        if (mSyncFilesystemOnProcessExit)
```
MonsieurNicolas, line 570: Should this be done before `handleProcessTermination` ?

Also, this seems too aggressive and unsafe:

* I *think* this gets executed off the main thread, and will stall it for arbitrary long
* even if we can push this to the worker thread, the main thread's simple writes (like logging) will get stalled behind all pending I/O.

I don't know if it's possible, but as we know the output filename, can't we invoke the SubCommand followed by something like (executed in a thread/or some clever async fsync):

```
fd = open(fn);
fsync(fd);
close(fd);
```

before claiming that the process completed and calling the `handleProcessTermination`

Obviously the potential issue here is that this doesn't work as `fd` doesn't get any new writes but maybe there is a way to trick the kernel in doing it, like append 1 bytes + truncate 1 byte or something like that.

MonsieurNicolas (changes requested):
We *could* merge this as is (as it has a safe default), but I suspect we'll want the real fix anyways as this may create a new class of problems when running under kubernetes.

src/process/ProcessManagerImpl.cpp
MonsieurNicolas: Ah found it, there is `syncfs` that does what we want. It's what happens when invoking from a shell the command
```
sync -f somefile
```

graydon: It does indeed get executed on the main thread -- that's the whole point, to block the main thread from doing more allocation / launching more processes / causing more VM pressure until existing dirty pages are written.

I'm not entirely sure what you're concerned about here: flushing too many filesystems besides the one we're intending to? Or flushing files on that filesystem that aren't necessary to flush?

fsync on a file flushes changes to that file -- we certainly could _just_ fsync the file that was transferred by (say) each get() command, that might be sufficient to relieve the pressure we're encountering, might not. In theory it should be but it's hard to tell.

I _think_ (trying to read between the lines of the man page) that opening a recently-written file -- even one written by a different process -- then calling `fsync` and closing it will in fact flush dirty pages associated with that file to the disk. I mean, I can try it and see anyways. If you prefer to do this narrower single-file version, I will try that.

`syncfs` is the same as `sync` but on a single filesystem. So it's a bit in-between `sync` (all filesystems) and `fsync` (one file). I'm not sure it's worth bothering to split that hair, since we're probably running on a single-filesystem node anyways (or only have one filesystem with much dirty data). Also `syncfs` isn't portable to macos or BSD.

graydon:
I've pushed a variant here that does what you suggested -- calls fsync on a file it reopened -- and have run it through the k8s cluster job in question. It does not appear to work: same dirty-page spikes we were seeing before. I guess fsync is only able to act on pages dirtied by a specific file handle not "any file handle pointing to the same file", whatever "same file" means at that level (inode? copy-of-file-in-memory? who knows).

I'll keep testing, it's possible I have a bug in my implementation here, but for the time being sync() is the only thing that works.

src/process/ProcessManagerImpl.cpp
MonsieurNicolas: no, what I am worried about is that the current behavior with `RunCommand` is that all work happens in a different process and the "join" with the main thread is virtually free as it happens when the child process finishes.

With that flag enabled, the sub process creates a bunch of debt that is paid back by the main thread. This can block for arbitrary long and will cause all sorts of havok in our single threaded app (like kill connections etc).

Then there is using system wide sync:
* some commands that don't even write to disk will potentially hang for a long time (for example, publish), as it may block on unrelated IO like a bucket merge or random log aggregator like the one that we have running on our boxes.
* this may negatively impact other processes running in the same container (or maybe it's even broader than this, like the whole host VM). I don't know how bad it can get for postgres or Horizon (captive core for sure is running in the same container).

MonsieurNicolas:
Yeah you are right it looks like https://stackoverflow.com/questions/37288453/calling-fsync2-after-close2 .

I would imagine that the trick I mentioned earlier (append one byte + flush + truncate) would work around that particular problem?

graydon:
Closing. Going to solve this via fiddling with `/proc/sys/vm/dirty_bytes`

PR #2779, opened 2020-10-29
k-means clustering: make weights at least 1

marta-lokhova:


src/util/Math.cpp
```diff
@@ -77,7 +77,8 @@ k_means(std::vector<double> const& points, uint32_t k)
         for (auto const& p : points)
         {
             auto closest = closest_cluster(p, centroids);
-            weights.emplace_back(std::pow(std::fabs(closest - p), 2));
+            auto weight = std::max<double>(1.0, std::pow(std::fabs(closest - p), 2));
```
MonsieurNicolas, line 80: That seems arbitrary... Which algorithm are we using? I *think* that this section is supposed to be the  implementing "kmeans++" right?

Also, sidenote: for computing `d2 = std::pow(std::fabs(closest - p), 2)`, it's just better to write it as `d2 = (closest - p); d2 *= d2;`

MonsieurNicolas:
I opened a different PR that implements k-means++ that fixes this and eliminates the "max iteration" on k-means++ (there is no need for this when implementing stock k-means++ as it's pretty much guaranteed to make progress or exhaust all points)

MonsieurNicolas:
https://github.com/stellar/stellar-core/pull/2785

PR #2780, opened 2020-10-29
Limit LogSlowExecution to one message per second (plus overflow summary).

graydon:
When there's a slow execution there's often a gigantic _storm_ of slow executions, producing thousands of log messages that then can be a performance problem of their own. This changes the slow-execution tracking machinery to (rather crudely) just throttle messages that occur faster than once a second, compressing them down to a summary of `Dropped <N> slow-execution warning messages` if they occur any faster.

(Ideally we'd never have such messages _and_ we'd have a fast enough logging subsystem not to care, but we're not quite living in that world yet. One step at a time.)

### DIFF
--- a/src/util/LogSlowExecution.cpp
+++ b/src/util/LogSlowExecution.cpp
@@ -1,0 +1,82 @@
+// Copyright 2020 Stellar Development Foundation and contributors. Licensed
+// under the Apache License, Version 2.0. See the COPYING file at the root
+// of this distribution or at http://www.apache.org/licenses/LICENSE-2.0
+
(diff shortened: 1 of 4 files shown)


src/util/LogSlowExecution.cpp
```diff
@@ -0,0 +1,82 @@
+// Copyright 2020 Stellar Development Foundation and contributors. Licensed
```
MonsieurNicolas, line 1: Visual C++ project file needs to be updated

MonsieurNicolas:
r+ f8df0fd61e1e56217739bc91d445f09910f33813

PR #2781, opened 2020-10-29
Split scheduler queues to be per-peer again.

graydon:
This just reverts a decision that snuck in at the end of the new action-scheduler work (back in 5085ad0b7) to merge the per-peer queues into per-message-type queues. I don't actually remember the rationale for that, and speculate that some of the performance hiccups we're seeing currently suggest that it was a mistake to merge them. At least worth trying.

### DIFF
--- a/src/overlay/Peer.cpp
+++ b/src/overlay/Peer.cpp
@@ -618,29 +618,30 @@ Peer::recvMessage(StellarMessage const& stellarMsg)
         fmt::format("Error RecvMessage T:{} cat:{} {} @{}", stellarMsg.type(),
                     cat, toString(), mApp.getConfig().PEER_PORT);
 
-    mApp.postOnMainThread([ err, weak, sm = StellarMessage(stellarMsg) ]() {
-        auto self = weak.lock();
-        if (self)
-        {
-            try
+    mApp.postOnMainThread(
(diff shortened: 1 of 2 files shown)


MonsieurNicolas:
I don't remember specifics as we were experimenting on pubnet and we don't actually have models to try this on (afaik).

I think that when we had per peer queues like this we ended up with higher latencies in general for item fetcher related traffic in some situations when we're overloading the entire system:
* I imagine that for a single peer generating a spike, this may help as we properly identify the queue for that peer as "overloaded" (with some downsides, see below)
* when many connections just have a lot of traffic, as decisions on dropping items is done on a per peer basis, we *delay* detection of the overloaded state because we end up with a highly fragmented view of the work which is a function of the number of connections, and end up having just larger queues for longer across the board.

The thing is, we have to make sure that we're not promoting bad patterns:
* in the case of a spike from one peer, if we don't throttle I/O on a per peer basis, we end up with most of the inbound network bandwidth going to one peer (which slows down all peers).
    * so maybe the real change is to try to throttle reads?
* in the case of evenly distributed traffic, it seems that the current strategy is better if the goal is to shed work

graydon:
Given that we took #2788 I'm going to close this for now. We can reopen if we want to revisit but I think you're right.

PR #2785, opened 2020-11-05
math: fix implementation of k-means++

MonsieurNicolas:


hidenori-shinohara:
1-dimensional k-means clustering can be solved deterministically using dynamic programming in O(n^2k) time and O(nk) space. For instance, https://www.ncbi.nlm.nih.gov/pmc/articles/PMC5148156/pdf/nihms833138.pdf describes the algorithm.

And implementing this isn't so bad https://github.com/hidenori-shinohara/stellar-core/commit/738468ecbbbd592df03ba0a7b9124809191ff390. This implementation passes all the [kmeans] tests. I'm not familiar with how we use k-means, so this algorithm may not work for our purpose, but I figured I'd mention just in case :)

MonsieurNicolas:
cool @hidenori-shinohara - right now we use it for ~ 100 samples, but that's good to know if we ever want to use it in other contexts.

marta-lokhova (approved):

MonsieurNicolas:
r+ a96897d680d014c451547a0d36a69905bc793cf1

PR #2786, opened 2020-11-06
Rework slot tracking

MonsieurNicolas:
This PR fixes issues related to how we keep track of slots, in particular it cleans up how we reason about slot related information between systems (bringing consistency) and erases old slots more aggressively when out of sync.


marta-lokhova:
r+ 504097f84b528e1a9b051a134cbb75ca78c41276

PR #2788, opened 2020-11-08
Split reads per peer

graydon:
Variant (as requested) of #2781 where we split the issuing of async reads into per-peer actions and run them through the scheduler for balance, but do not split the received message-type queues to be per-peer.

Currently doing perf evaluation / comparison of the two approaches.


MonsieurNicolas:
r+ 9dd0f33856d6d1b374b9211b6f8570fdc9df08ed

PR #2789, opened 2020-11-09
Lower scheduler latency window / overload signal threshold from 5s -> 1s

graydon:
This makes the Action Scheduler a little more aggressive about deciding it's overloaded and dropping messages.


MonsieurNicolas:
closing: I don't think this makes sense anymore

PR #2792, opened 2020-11-10
Lower crank time slice from 500ms -> 10ms.

graydon:
This is intended to help balance load more fairly between peers and tighten tails on queuing delays when under load.


MonsieurNicolas:
closing: I don't think this makes sense anymore

PR #2793, opened 2020-11-10
Minor cpu-usage shaving

graydon:
This is a grab bag of unobtrusive CPU-usage fixes, especially in the hot paths of networking. None is responsible for a lot, but every little bit helps.

### DIFF
--- a/src/overlay/PeerBareAddress.h
+++ b/src/overlay/PeerBareAddress.h
@@ -53,7 +53,7 @@ class PeerBareAddress
         return mPort;
     }
 
-    std::string toString() const;
+    std::string const& toString() const;
 
     bool isPrivate() const;
     bool isLocalhost() const;
@@ -66,5 +66,6 @@ class PeerBareAddress
     Type mType;
     std::string mIP;
     unsigned short mPort;
+    std::string mStringValue;
 };
 }
--- a/src/crypto/SHA.cpp
+++ b/src/crypto/SHA.cpp
@@ -58,7 +58,8 @@ uint256
 SHA256::finish()
 {
     uint256 out;
-    assert(out.size() == crypto_hash_sha256_BYTES);
+    static_assert(sizeof(out) == crypto_hash_sha256_BYTES,
+                  "unexpected crypto_hash_sha256_BYTES");
     if (mFinished)
     {
         throw std::runtime_error("finishing already-finished SHA256");
--- a/src/overlay/Peer.cpp
+++ b/src/overlay/Peer.cpp
@@ -570,7 +570,7 @@ Peer::recvMessage(StellarMessage const& stellarMsg)
         return;
     }
 
-    std::string cat;
+    char const* cat = nullptr;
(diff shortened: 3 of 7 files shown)


src/overlay/PeerBareAddress.h
```diff
@@ -66,5 +66,6 @@ class PeerBareAddress
     Type mType;
```
MonsieurNicolas, line 66: all members should be const now

src/main/ApplicationImpl.cpp
```diff
@@ -1026,11 +1026,12 @@ ApplicationImpl::postOnMainThread(std::function<void()>&& f, std::string&& name,
 {
     LogSlowExecution isSlow{name, LogSlowExecution::Mode::MANUAL,
                             "executed after"};
-    mVirtualClock.postAction([ this, f = std::move(f), isSlow ]() {
-        mPostOnMainThreadDelay.Update(isSlow.checkElapsedTime());
-        f();
-    },
-                             std::move(name), type);
+    mVirtualClock.postAction(
+        [ this, f = std::move(f), isSlow = std::move(isSlow) ]() {
```
MonsieurNicolas, line 1030: c++ trivia.

I think you're actually making a copy here:
as `LogSlowExecution` has a destructor, it does not have a "good" default move constructor, so instead this will just create a new one.

src/overlay/TCPPeer.cpp
```diff
@@ -441,8 +441,11 @@ TCPPeer::startRead()
 
     mIncomingHeader.clear();
 
-    CLOG(DEBUG, "Overlay") << "TCPPeer::startRead " << mSocket->in_avail()
-                           << " from " << toString();
+    if (Logging::logDebug("Overlay"))
+    {
+        CLOG(DEBUG, "Overlay") << "TCPPeer::startRead " << mSocket->in_avail()
```
MonsieurNicolas, line 446: we should just move to a proper logging library for which we don't need to do this all over the place

MonsieurNicolas (changes requested):

src/overlay/PeerBareAddress.h
graydon: Nope. Tried. We use it in too many places where a value of type PeerBareAddress gets assigned. C++ const is mostly useless in fields of value-like objects, only non-copied / non-moved objects.

src/overlay/TCPPeer.cpp
graydon: Well, yes, but I'm going to start on that _next_ month. This is for the meantime.

src/main/ApplicationImpl.cpp
graydon: Great. More trivia for the "why does C++ even try" bucket. I'll remove the change.

MonsieurNicolas:
r+ 9b80c3d75bbb07e9f13a702eb08e6fb1eed89646

MonsieurNicolas:
@latobarita: retry

MonsieurNicolas:
@latobarita: retry

PR #2794, opened 2020-11-10
Remove old load manager

graydon:
This removes the old LoadManager and all its supporting machinery. This is a mix of dead code and should-be-dead code: nothing that it measures currently is a very good proxy for "amount of load" -- database operations are all batched now and on-CPU time is almost all deferred to the run-side of the Action Scheduler which happens outside the TimeScopes in here -- and any decisions we're making based on it are likely to be as good as random; but conversely it consumes a fair amount of on-CPU time just to _record_ the metrics that it's recording. So I think removing it at this point is a better idea.

We can add new/better logic in the future based on tagging Actions flowing through the Action Scheduler and/or cost estimates in the Quorum Tracker.


MonsieurNicolas:
looks good - only thing you forgot is to update Visual Studio files

graydon:
(updated, after hours..)

MonsieurNicolas:
@graydon needs to be rebased

MonsieurNicolas:
r+ a3e678c5f280bf289c67fa0b1bcd13a44140b45b

PR #2795, opened 2020-11-10
Change internal strong hashes to BLAKE2

graydon:
This switches two of the strong internal hashes (one on floodgate, one on signature checking cache, neither exposed as part of any protocol) from SHA256 to [BLAKE2b](https://www.blake2.net/), which runs about twice as fast and has equal-or-superior security properties. This also happens to be the default "generic" strong hash in libsodium, so not especially hard to hook up to.

Along the way it also changes a bunch of `sha256(xdr_to_opaque(...))` calls to the non-allocating (and very slightly faster) `xdrSha256` and `xdrBlake2` variants

### DIFF
--- a/src/crypto/BLAKE2.cpp
+++ b/src/crypto/BLAKE2.cpp
@@ -1,0 +1,75 @@
+// Copyright 2020 Stellar Development Foundation and contributors. Licensed
+// under the Apache License, Version 2.0. See the COPYING file at the root
+// of this distribution or at http://www.apache.org/licenses/LICENSE-2.0
+
+#include "crypto/BLAKE2.h"
+#include "crypto/ByteSlice.h"
+#include "crypto/CryptoError.h"
+#include "util/NonCopyable.h"
+#include <Tracy.hpp>
+#include <sodium.h>
+
+namespace stellar
+{
+
+// Plain BLAKE2 (a.k.a. BLAKE2b)
+uint256
+blake2(ByteSlice const& bin)
+{
+    static_assert(crypto_generichash_BYTES == sizeof(uint256),
+                  "unexpected crypto_generichash_BYTES");
+    ZoneScoped;
+    uint256 out;
+    if (crypto_generichash(out.data(), out.size(), bin.data(), bin.size(),
+                           nullptr, 0) != 0)
+    {
+        throw CryptoError("error from crypto_generichash");
+    }
+    return out;
+}
+
+BLAKE2::BLAKE2()
+{
+    reset();
+}
+
+void
+BLAKE2::reset()
+{
+    if (crypto_generichash_init(&mState, nullptr, 0,
+                                crypto_generichash_BYTES) != 0)
+    {
+        throw CryptoError("error from crypto_generichash_init");
+    }
+    mFinished = false;
+}
+
+void
+BLAKE2::add(ByteSlice const& bin)
+{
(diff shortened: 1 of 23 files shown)


src/crypto/BLAKE2.cpp
```diff
@@ -0,0 +1,75 @@
+// Copyright 2020 Stellar Development Foundation and contributors. Licensed
```
MonsieurNicolas, line 1: need to update Visual Studio files

src/overlay/Peer.cpp
```diff
@@ -915,7 +915,7 @@ void
 Peer::recvSCPQuorumSet(StellarMessage const& msg)
 {
     ZoneScoped;
-    Hash hash = sha256(xdr::xdr_to_opaque(msg.qSet()));
+    Hash hash = xdrSha256(msg.qSet());
```
MonsieurNicolas, line 918: this belongs to the previous commit

src/overlay/Floodgate.cpp
```diff
@@ -58,7 +58,7 @@ bool
 Floodgate::addRecord(StellarMessage const& msg, Peer::pointer peer, Hash& index)
 {
     ZoneScoped;
-    index = sha256(xdr::xdr_to_opaque(msg));
+    index = xdrBlake2(msg);
```
MonsieurNicolas, line 61: you forgot to update `Tracker::listen` (so this commit breaks item fetcher)

MonsieurNicolas (changes requested):

src/crypto/BLAKE2.cpp
graydon: fixed

src/overlay/Floodgate.cpp
graydon: I .. don't think so! The item fetcher and tracker deals in public sha256 hashes -- the shared identities of txsets and qsets baked into the public protocol. We're not changing its hashes.

The floodmap is an entirely internal usage, the hashes that key it never leave it. They're private book-keeping, which is why it's safe to change them here (and changing them is unrelated to the hashes used elsewhere).

src/overlay/Peer.cpp
graydon: fixed

src/overlay/Floodgate.cpp
MonsieurNicolas: no: look at how `getPeersKnows` (which now uses Blake2 hashes) is used: it's passing what was computed in `Tracker::listen`, which is computed as:

```
    // NB: hash here is of StellarMessage
    mWaitingEnvelopes.push_back(
        std::make_pair(sha256(xdr::xdr_to_opaque(m)), env));
```

Those keys are only used in connection to the flood map from what I can tell. The entire code base, except for the actual download code only uses the values.

This should not be confused with *what* item fetcher is trying to download (which is also a hash but of the item that we're trying to get).

graydon: Oh my goodness. Nice catch! Fixed (and added a comment).

MonsieurNicolas:
r+ e18b39572e2a5ce705b129c8203c3eef32ca0347

katopz:
@graydon  What about [BLAKE3](https://github.com/BLAKE3-team/BLAKE3)? 😬

image

PR #2796, opened 2020-11-10
Overlay AEAD

graydon:
This changes the overlay transport system from plaintext-with-SHA256-HMAC-auth to ChaCha20-Poly1305-AEAD -- authenticated encryption.

This is a cheaper overall construct, a little more than twice as fast, provides stronger authentication, and happens to provide secrecy along with the authentication (the Poly1305 authenticator needs to crank a stream cipher primitive forward to provide a cryptographically-strong nonce stream, and it's evidently not really any more expensive to encrypt while doing so).

The approach here upgrades to the new transport mode when talking to new-protocol peers, but continues with the old mode when talking to old-protocol peers, and uses the same certified-ECDH exchange to negotiate a shared-key in either case (as luck has it both authenticators work with 32-byte shared keys, so that part can be reused).

There's a _little_ subtlety in the initialization sequence. AEAD-setup requires conveying an IV-like (public) initial nonce from source to sink. We _could_ do this in the `AUTH` message we send at the end of normal key negotiation (which is a no-content message to check that authentication has activated) but that message happens to be non-versioned and have no room for carrying an IV. Rather than adding a new `AUTH`-variant that has a version-union within it, we use the nearest container that _has_ a version-union, the `AuthenticatedMessage` containing it. We add _two_ versions to its `v()` union: version `1` for the IV-carrying message and version `2` for all the remaining messages in the stream.


MonsieurNicolas (changes requested):
Change is bit more complicated than I thought it would be:
* there is the flow that changed a bit - this may be fine if temporary, but...
* if we want to remove support for old style HMAC, we'll still have to support v0 type messages for `HELLO` "forever after". it would have been nicer to be able to clean migrate (so a new AUTH2 message was not as bad)
* could we avoid this by just not doing any encryption maybe?
    * I don't think we really care about having encrypted traffic - it's actually pretty nice to be able to analyze traffic with libpcap
    * On that front, I am quite puzzled by the perf gains... this seems to actually do more work than just hashing (there is always an extra copy involved in the encrypted case).
          * you made changes in a different PR that could have big perf impacts, like the commit "Change a bunch of sha256(xdr_to_opaque(...)) calls to xdrSha256(...)" in https://github.com/stellar/stellar-core/pull/2795/commits - maybe this kind of change is the real gain?
          * note that we're being a bit dumb when computing hmacs so we can squeeze more perf - we keep converting xdr into opaque (that we throw away). We could do something like `xdrSha256` here too (maybe this change would also benefit the new approach btw).
        * in libsodium's docs they claim that https://doc.libsodium.org/advanced/hmac-sha2#hmac-sha-512-256 may be faster than HMAC-SHA256 (probably on 64 bit machines as sha-512 is designed for that). I wonder if the crypto that you picked in this PR happens to be optimized for 64 bits (doing some googling around, 64 bit vs 32 bit algorithms are typically 50% faster).

PR #2797, opened 2020-11-10
Drop messages in the write queues as well when overloaded.

graydon:
This reproduces the load-shedding behaviour of the Action Scheduler at the lower level per-peer write queues. It uses the same time threshold and only drops messages enqueued during the execution of an action itself marked as droppable.

### DIFF
--- a/src/overlay/OverlayMetrics.cpp
+++ b/src/overlay/OverlayMetrics.cpp
@@ -13,6 +13,8 @@ OverlayMetrics::OverlayMetrics(Application& app)
           app.getMetrics().NewMeter({"overlay", "message", "read"}, "message"))
     , mMessageWrite(
           app.getMetrics().NewMeter({"overlay", "message", "write"}, "message"))
+    , mMessageDrop(
+          app.getMetrics().NewMeter({"overlay", "message", "drop"}, "message"))
     , mAsyncRead(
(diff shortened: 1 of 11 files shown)


src/overlay/OverlayMetrics.cpp
```diff
@@ -13,6 +13,8 @@ OverlayMetrics::OverlayMetrics(Application& app)
           app.getMetrics().NewMeter({"overlay", "message", "read"}, "message"))
     , mMessageWrite(
           app.getMetrics().NewMeter({"overlay", "message", "write"}, "message"))
+    , mMessageDrop(
+          app.getMetrics().NewMeter({"overlay", "message", "drop"}, "message"))
```
MonsieurNicolas, line 17: needs to be added to the metrics.md docs

MonsieurNicolas (changes requested):
Only a minor miss (docs), otherwise this seems good to me

src/overlay/OverlayMetrics.cpp
graydon: Fixed (and rebased to resolve conflict).

MonsieurNicolas:
r+ a7f0341d66f1644a797aec9839c682bd7d1a606f

PR #2798, opened 2020-11-10
Split broadcasts into multiple Scheduler Actions.

graydon:
This modifies the floodgate broadcast function to enqueue each message-send in the broadcast as a separate scheduler action (on separate, per-receiver broadcast queues). The idea here is to reduce maximum latency from broadcasts -- one broadcast can take a while when sending to a lot of peers, separately encoding and adding authentication material and writing to each.

### DIFF
--- a/src/overlay/Floodgate.cpp
+++ b/src/overlay/Floodgate.cpp
@@ -14,6 +14,7 @@
 #include "util/XDROperators.h"
 #include "xdrpp/marshal.h"
 #include <Tracy.hpp>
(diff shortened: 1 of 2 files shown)


MonsieurNicolas:
r+ 205c6bbe500216838d213826ebed00dd57a7f2f5

graydon:
(ah, minor test break. fixing and resubmitting)

graydon:
r+ c9d8f67

PR #2801, opened 2020-11-12
Subprocess shutdown fixes

graydon:
Third, slightly wider but still narrow attempt at this PR. I've attempted to integrate all review comments so far. There's one fairly large chunk of inter-dependent refactoring in the last commit; unfortunately it proved hard to separate out further while rebasing, but I _think_ it's still fairly comprehensible.

There should be no contract changes here, only improved internal book-keeping and interaction with external processes shutting down. There are two main things happening:

  - The motivating change: making all the SIGTERM and SIGKILL'ed processes route into the same child-reaping path that we use for normal (non-killing) process exit, so we actually reap them, and making sure we do this on shutdown and destruction. This is to avoid producing orphan zombies if we're running in a container. Containers don't always have init processes and so they can accumulate orphan zombies if we don't do this. (see horizon bug https://github.com/stellar/go/issues/3217). This mostly happens in the second-last change 04e0c26 though there's some setup before it.
  - Refactoring and cleanup:
    - Merging and reusing more code where possible, making some platform-specific stuff platform-generic.
    - Fixing minor and probably-never-exercised bugs in error-routing and locking.
    - Reporting errors with strerror.
    - Renaming things that were misnamed or inconsistently named.
    - Adding comments.
    - Adding an invariant-checking helper.

### DIFF
--- a/src/process/ProcessManagerImpl.cpp
+++ b/src/process/ProcessManagerImpl.cpp
@@ -440,7 +440,11 @@ ProcessExitEvent::Impl::run()
 
         if (ec)
         {
-            *(sf->mOuterEc) = ec;
+            // We can only transmit one ec to the callback, so if we
+            // already have a nonzero (failure) ec from the waitable
+            // process handle, don't overwrite it with a new one from
+            // handleProcessTermination.
+            manager->handleProcessTermination(sf->mProcessId, 1);
         }
         else
         {
@@ -451,9 +455,9 @@ ProcessExitEvent::Impl::run()
             {
                 exitCode = 1;
             }
-            ec = asio::error_code(exitCode, asio::system_category());
+            ec = manager->handleProcessTermination(sf->mProcessId,
+                                                   static_cast<int>(exitCode));
         }
-        ec = manager->handleProcessTermination(sf->mProcessId, ec.value());
         sf->cancel(ec);
     });
     mRunning = true;
--- a/src/process/ProcessManagerImpl.cpp
+++ b/src/process/ProcessManagerImpl.cpp
@@ -876,14 +876,20 @@ ProcessManagerImpl::maybeRunPendingProcesses()
         {
             CLOG_DEBUG(Process, "Running: {}", i->mImpl->mCmdLine);
 
-            if (fs::exists(i->mImpl->mOutFile))
+            if (!i->mImpl->mOutFile.empty() && fs::exists(i->mImpl->mOutFile))
             {
                 throw std::runtime_error(fmt::format(
                     "output file {} already exists", i->mImpl->mOutFile));
             }
 
             i->mImpl->run();
-            mProcesses[i->mImpl->getProcessId()] = i;
+            auto pid = i->mImpl->getProcessId();
+            if (mProcesses.find(pid) != mProcesses.end())
+            {
+                throw std::runtime_error(
+                    fmt::format("process {} already exists", pid));
+            }
+            mProcesses[pid] = i;
         }
         catch (std::runtime_error& e)
         {
--- a/src/process/ProcessManagerImpl.cpp
+++ b/src/process/ProcessManagerImpl.cpp
@@ -136,21 +136,38 @@ ProcessManagerImpl::getNumRunningOrShuttingDownProcesses()
 
 ProcessManagerImpl::~ProcessManagerImpl()
 {
-    const auto killProcess = [&](ProcessExitEvent& pe) {
-        pe.mImpl->cancel(ABORT_ERROR_CODE);
-        forceShutdown(pe);
-    };
-    // Use SIGKILL on any processes we already used SIGINT on
-    while (!mKillable.empty())
+// First unregister our SIGCHLD handler; it has a raw pointer to us and
+// anyways we won't be able to bounce off ASIO for the work we're about
+// to do below.
+#ifndef _WIN32
+    auto ec = ABORT_ERROR_CODE;
+    mSigChild.cancel(ec);
+#endif
+
+    // Then trigger shutdown, if we haven't yet (it's idempotent). This will ask
+    // every running process to shut down politely and cancel all events,
+    // pending and running.
+    shutdown();
+
+    for (int i = 0; i < 3 && !mProcesses.empty(); i++)
     {
-        auto impl = std::move(mKillable.front());
-        mKillable.pop_front();
-        killProcess(*impl);
-    }
-    // Use SIGKILL on any processes we haven't politely asked to exit yet
-    for (auto& pair : mProcesses)
-    {
-        killProcess(*pair.second);
+        // At this point we're purely racing against process exits; we don't
+        // want to block or spin in a dtor by trying to trap SIGCHLD (and we
+        // might have dropped a SIGCHLD after cancelling the handler above
+        // anyways) but we can at least try to lose the race by sleeping briefly
+        // before calling waitpid below with WNOHANG in the reapChildren() call.
+        std::this_thread::sleep_for(std::chrono::milliseconds(10));
+
+        // Reap anything that politely shut down in response to last iteration.
+        reapChildren();
+
+        // Stop early if we're done.
+        if (mProcesses.empty())
+            break;
+
+        // Re-trigger progressively-more-forcible shutdown of everything
+        // surviving.
+        tryProcessShutdownAll();
     }
 }
 
@@ -163,33 +180,19 @@ ProcessManagerImpl::isShutdown() const
 void
 ProcessManagerImpl::shutdown()
 {
+    std::lock_guard<std::recursive_mutex> guard(mProcessesMutex);
     if (!mIsShutdown)
     {
         mIsShutdown = true;
-        auto ec = ABORT_ERROR_CODE;
 
         // Cancel all pending.
-        std::lock_guard<std::recursive_mutex> guard(mProcessesMutex);
         for (auto& pending : mPending)
         {
-            pending->mImpl->cancel(ec);
+            pending->mImpl->cancel(ABORT_ERROR_CODE);
         }
         mPending.clear();
 
-        // Cancel all running.
-        for (auto& pair : mProcesses)
-        {
-            // Mark it as "ready to be killed"
-            mKillable.push_back(pair.second);
-            // Cancel any pending events and shut down the process cleanly
-            pair.second->mImpl->cancel(ec);
-            cleanShutdown(*pair.second);
-        }
-        mProcesses.clear();
-        gNumProcessesActive = 0;
-#ifndef _WIN32
-        mSigChild.cancel(ec);
-#endif
+        tryProcessShutdownAll();
     }
 }
 
--- a/src/process/ProcessManagerImpl.cpp
+++ b/src/process/ProcessManagerImpl.cpp
@@ -15,6 +15,7 @@
 #include "process/ProcessManager.h"
 #include "process/ProcessManagerImpl.h"
 #include "util/Fs.h"
+#include "util/GlobalChecks.h"
 #include "util/Logging.h"
 #include "util/Timer.h"
 #include <Tracy.hpp>
@@ -40,19 +41,88 @@
 extern char** environ;
 #endif
 
+namespace
+{
+enum ProcessLifecycle
+{
+    PENDING = 0,
+    RUNNING = 1,
+    TRIED_POLITE_SHUTDOWN = 2,
+    TRIED_FORCED_SHUTDOWN = 3,
+    TERMINATED = 5
+};
+}
+
 namespace stellar
 {
 
 static const asio::error_code ABORT_ERROR_CODE(asio::error::operation_aborted,
                                                asio::system_category());
 
+static asio::error_code
+mapExitStatusToErrorCode(std::string const& cmdLine, int pid, int status)
+{
+// On windows, an exit status is just an exit status. On unix it's
+// got some flags incorporated into it.
+#ifdef _WIN32
+    return asio::error_code(status, asio::system_category());
+#else
+    if (WIFEXITED(status))
+    {
+        if (WEXITSTATUS(status) == 0)
+        {
+            CLOG_DEBUG(Process, "process {} exited {}: {}", pid,
+                       WEXITSTATUS(status), cmdLine);
+        }
+        else
+        {
+            CLOG_WARNING(Process, "process {} exited {}: {}", pid,
+                         WEXITSTATUS(status), cmdLine);
+        }
+#ifdef __linux__
+        // Linux posix_spawnp does not fault on file-not-found in the
+        // parent process at the point of invocation, as BSD does; so
+        // rather than a fatal error / throw we get an ambiguous and
+        // easily-overlooked shell-like 'exit 127' on waitpid.
+        if (WEXITSTATUS(status) == 127)
+        {
+            CLOG_WARNING(Process, "");
+            CLOG_WARNING(Process, "************");
+            CLOG_WARNING(Process, "");
+            CLOG_WARNING(Process, "  likely 'missing command':");
+            CLOG_WARNING(Process, "");
+            CLOG_WARNING(Process, "    {}", cmdLine);
+            CLOG_WARNING(Process, "");
+            CLOG_WARNING(Process, "************");
+            CLOG_WARNING(Process, "");
+        }
+#endif
+        // FIXME: this doesn't _quite_ do the right thing; it conveys
+        // the exit status back to the caller but it puts it in "system
+        // category" which on POSIX means if you call .message() on it
+        // you'll get perror(value()), which is not correct. Errno has
+        // nothing to do with process exit values. We could make a new
+        // error_category to tighten this up, but it's a bunch of work
+        // just to convey the meaningless string "exited" to the user.
+        return asio::error_code(WEXITSTATUS(status), asio::system_category());
+    }
+    else
+    {
+        // FIXME: for now we also collapse all non-WIFEXITED exits on
+        // posix into a single "exit 1" error_code. This is enough
+        // for most callers; we can enrich it if anyone really wants
+        // to differentiate various signals that might have killed
+        // the child.
+        return asio::error_code(1, asio::system_category());
+    }
+#endif
+}
+
 std::shared_ptr<ProcessManager>
 ProcessManager::create(Application& app)
 {
     return std::make_shared<ProcessManagerImpl>(app);
 }
-
-std::atomic<size_t> ProcessManagerImpl::gNumProcessesActive{0};
 
 class ProcessExitEvent::Impl
     : public std::enable_shared_from_this<ProcessExitEvent::Impl>
@@ -63,7 +133,7 @@ class ProcessExitEvent::Impl
     std::string const mCmdLine;
     std::string const mOutFile;
     std::string const mTempFile;
-    bool mRunning{false};
+    ProcessLifecycle mLifecycle{ProcessLifecycle::PENDING};
 #ifdef _WIN32
     asio::windows::object_handle mProcessHandle;
 #endif
@@ -90,6 +160,7 @@ class ProcessExitEvent::Impl
     finish()
     {
         ZoneScoped;
+        releaseAssertOrThrow(mLifecycle == ProcessLifecycle::TERMINATED);
         if (!mOutFile.empty())
         {
             if (fs::exists(mOutFile))
@@ -97,8 +168,7 @@ class ProcessExitEvent::Impl
                 CLOG_WARNING(Process, "Outfile already exists: {}", mOutFile);
                 return false;
             }
-            else if (mRunning &&
-                     std::rename(mTempFile.c_str(), mOutFile.c_str()))
+            else if (std::rename(mTempFile.c_str(), mOutFile.c_str()))
             {
                 CLOG_ERROR(Process, "{} -> {} rename failed, error: {}",
                            mTempFile, mOutFile, errno);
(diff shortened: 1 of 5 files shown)


src/process/ProcessManagerImpl.cpp
```diff
@@ -210,8 +210,7 @@ ProcessManagerImpl::tryProcessShutdown(std::shared_ptr<ProcessExitEvent> pe)
     if (killableIt != mKillable.end())
     {
         CLOG(DEBUG, "Process") << "Shutting down (forced): " << impl->mCmdLine;
-        forceShutdown(*pe);
-        return true;
+        return forceShutdown(*pe);
```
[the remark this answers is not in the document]
graydon: This comment led to a much-needed overhaul of the PR! Revising description presently.

src/process/ProcessManagerImpl.cpp
```diff
@@ -210,8 +210,7 @@ ProcessManagerImpl::tryProcessShutdown(std::shared_ptr<ProcessExitEvent> pe)
     if (killableIt != mKillable.end())
     {
         CLOG(DEBUG, "Process") << "Shutting down (forced): " << impl->mCmdLine;
-        forceShutdown(*pe);
```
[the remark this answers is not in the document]
graydon: Doing so in the reaping path now.

src/historywork/RunCommandWork.h
```diff
@@ -33,6 +33,17 @@ class RunCommandWork : public BasicWork
                    size_t maxRetries = BasicWork::RETRY_A_FEW);
     ~RunCommandWork() = default;
 
+    bool
+    isDone() const
```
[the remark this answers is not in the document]
marta-lokhova: Also, I just realized that RunCommandWork is not exactly doing the right thing. When aborted, right now it is still possible for the callback to update Work's member variables (which is what you added in WorkTests), but we shouldn't do this. When work is aborted, its state is reset, so a callback making changes to the state is wrong. So I think instead in WorkTests we could just crank for a bit, then check if the number of remaining processes is as expected. 
(regarding the callback firing post reset, I can fix that separately as it's outside of the scope of this PR)

src/process/ProcessManagerImpl.cpp
```diff
@@ -235,13 +233,18 @@ ProcessManagerImpl::tryProcessShutdown(std::shared_ptr<ProcessExitEvent> pe)
...
+
+            // Send a SIGTERM / Ctrl-C
+            cleanShutdown(*runningIt->second);
+
+            // We initiated a clean shutdown here but the process isn't dead
+            // yet; caller needs to try again if they want to be sure.
+            //
+            // SIGCHLD and waitpid will be handled up above in
+            // handleSignalWait / handleProcessTermination.
+            return false;
```
MonsieurNicolas, line 277: I think this commit broke clean shutdown for `RunCommand`?

before: it would return `true` not because the process was terminated, but because the call `kill(pid, SIGINT)` succeeds. I am not sure when force kill was called in this particular situation (I think when the main process really stops)... but most commands that we call would not need to be force killed.

now: it returns false, causing `onAbort` to retry in the next `crank` call, which is likely to happen very fast, and when it gets called again, it does a force kill.

This can have bad effects as "force killing" scripts (like what can be done with history archives get/put) that spawn subprocesses would not be able to cleanup properly. This increases the chances of corruptions in those scripts/sub-processes.

src/process/ProcessManagerImpl.cpp
```diff
@@ -235,13 +233,18 @@ ProcessManagerImpl::tryProcessShutdown(std::shared_ptr<ProcessExitEvent> pe)
         auto runningIt = mProcesses.find(pid);
         if (runningIt != mProcesses.end())
         {
-            // Mark it as "ready to be killed"
+            // Mark it as "ready to be killed" if caller retries.
```
MonsieurNicolas, line 266: note: `mKillable` (and `mPending`) do enforce uniqueness of elements, maybe we should just make them a map instead of a deque

src/process/ProcessManagerImpl.cpp
```diff
@@ -133,26 +133,39 @@ ProcessManagerImpl::getNumRunningProcesses()
...
+    // dropped a SIGCHLD after cancelling the handler above anyways) but we can
+    // at least try to lose the race by sleeping briefly before calling waitpid
+    // below with WNOHANG in the handleSignalWait() call.
+    std::this_thread::sleep_for(std::chrono::milliseconds(10));
+
+    // Reap anything that politely shut down in response.
+    reapChildren();
+
+    // Then trigger forceful shutdown of everything that survived.
+    for (auto k : mKillable)
```
MonsieurNicolas, line 160: `auto const&`

src/process/ProcessManagerImpl.cpp
```diff
@@ -133,26 +133,39 @@ ProcessManagerImpl::getNumRunningProcesses()
...
     {
-        killProcess(*pair.second);
+        forceShutdown(*k);
     }
+
+    // And reap all the processes that died after that, too, again trying to
+    // lose the race against such exits if we can.
+    std::this_thread::sleep_for(std::chrono::milliseconds(10));
+    reapChildren();
```
MonsieurNicolas, line 168: I'm not 100% sure this sequence works as intended on Windows as process handling is done on the completion thread, but I think it might be "good enough" (force killing everything).
I did a few smoke tests, and things seems OK

src/process/ProcessManagerImpl.cpp
```diff
@@ -194,7 +194,12 @@ ProcessManagerImpl::shutdown()
         for (auto& pair : mProcesses)
         {
             // Mark it as "ready to be killed"
-            mKillable.push_back(pair.second);
+            auto killableIt =
```
MonsieurNicolas, line 197: see my note on using a map instead: both `mKillable` and `mPending` want the same thing (and it's error prone)

MonsieurNicolas (changes requested):

src/process/ProcessManagerImpl.cpp
graydon: I agree this is the effect of it but I'm not sure what the right thing to do is instead. Maybe have some sort of shutdown/grace period? We should force-kill _eventually_ if TERM isn't effective.

src/process/ProcessManagerImpl.cpp
graydon: Yeah, I thought of moving to a map/set instead but held off. The deque is (as far as I know) there to enforce ordering on operations -- so that we kill processes (and thereby start subsequent processes) in as-close-to-ordered as possible, rather than random pointer order.

I'm not sure this _really_ helps though since processes are likely to complete and exit in an arbitrary (or at least not-totally-predictable) order. And since we ultimately respond to SIGCHLDs as they come in, maybe we'll be pulling stuff out in effectively-random order anyway.

src/process/ProcessManagerImpl.cpp
graydon: Yeah, I'm not sure it works _great_ anywhere. Ultimately we're in a bit of a bind as far as not wanting to block or fail, just do best-effort kills.

I think the great majority of our users (especially those running core subprocesses and killing them at random, eg. captive subprocesses of horizon) are going to be running in unix containers. And you have to be running a process when someone kills core to even hit this path. I expect this path is a no-op for most users most of the time.

src/process/ProcessManagerImpl.cpp
MonsieurNicolas: There are probably a couple issues:
* we should transition from `ABORTING` to `ABORTED` only when we have confirmation that the process is actually dead
* we already "kinda have" a grace period between `ApplicationImpl::gracefulStop()` and `shutdownMainIOContext` (and then the actual "application/process manager destructor" that will force kill everything)
   * but I don't know if having some work in the `ABORTING` state actually stops the main loop from exiting

The cleanest thing to do might be to add a grace period (shorter than the one used in the application) to `RunCommand` - but I think that for that to work, we need a new work state `ABORTING_WAITING` (that avoids busy loops).

src/historywork/RunCommandWork.h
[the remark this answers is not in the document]
graydon: I guess I didn't want to just spin on checking that the process eventually exits because it's a process that just sleeps -- it will exit of its own accord regardless. We really want to check that it got _terminated_.

I also meant to ask about this, I guess: the existing code considers the work aborted once it sends a kill, without confirming process exit. We're preserving (and even enhancing) that behaviour a bit in this PR, with the event-handlers being cancelled in advance of the process exit happening. I wanted to check with you that that seems like the right thing in your mind.

It means core might _think_ a process has exited (or at least any part that's just judging from the cancelled event) before the process has actually exited; but on the other hand it means that same listener is not likely to get stuck waiting for an unkillable process, and can make progress on its own knowing that the process manager has at least tried its best to kill the process and it's now up to the OS to deal with. Not sure what's best.

src/process/ProcessManagerImpl.cpp
```diff
@@ -146,6 +146,10 @@ ProcessManagerImpl::~ProcessManagerImpl()
     }
     // Use SIGKILL on any processes we haven't politely asked to exit yet
     for (auto& pair : mProcesses)
+#ifndef _WIN32
```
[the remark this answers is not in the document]
graydon: Oh yeah weird that's supposed to be above the loop. Whoops.

src/historywork/RunCommandWork.h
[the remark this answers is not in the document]
marta-lokhova: Yeah, I think back then I thought that as long as Work tried its best to kill the process, it can transition into the "aborted" state. Thinking about it now, not sure what else we can do here.. I mean, at some point we'll have to just let the OS figure it out. So I'm fine preserving this behavior, mostly because I don't have a better alternative to offer.

re: tests. Yeah, it's annoying :/ I understand your intent, it's just that we're breaking the Work contract here by allowing mDone to be modified after the abort is done.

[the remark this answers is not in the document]
graydon: pushed variant that does not alter public API contracts.

src/process/ProcessManagerImpl.cpp
```diff
@@ -520,24 +555,31 @@ ProcessManagerImpl::ProcessManagerImpl(Application& app)
...
     if (isShutdown())
     {
         return;
     }
+    reapChildren();
+    startWaitingForSignalChild();
+}
+
+void
+ProcessManagerImpl::reapChildren()
```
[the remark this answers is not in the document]
graydon: added

src/process/ProcessManagerImpl.cpp
graydon: decided to back off the "guaranteed kill" approach: maintaining the old/existing API contract which is "report true when we successfully _signalled_ the process to exit (and cancelled the client's completion event), let ProcessManager itself mop up the actual exit".

I don't love this contract, but it works and it seems like replacing it just shifts the possible failure modes around without really fixing them. Best to leave as-is.

src/process/ProcessManagerImpl.cpp
graydon: Pushed version that uses a set (and in fact two sets -- one for "we sent SIGTERM" and one for "we sent SIGKILL"). Kept the pending deque a deque because it _is_ relevant which order we run processes in, even if they exit in a random order.

src/process/ProcessManagerImpl.cpp
graydon: yeah, see above re: map

src/process/ProcessManagerImpl.cpp
graydon: fixed

src/process/ProcessManagerImpl.cpp
```diff
@@ -126,27 +126,45 @@ class ProcessExitEvent::Impl
 size_t
 ProcessManagerImpl::getNumRunningProcesses()
 {
-    return gNumProcessesActive;
+    return gNumProcessesActive -
```
marta-lokhova, line 129: Now that the meaning of `gNumProcessesActive` has changed (i.e. it includes processes that are "being killed"), I wanted to double check if its usage in `ProcessManagerImpl::maybeRunPendingProcesses` is right. `gNumProcessesActive` is used to throttle pending processes:
https://github.com/stellar/stellar-core/blob/f3e45d1c3f1f5ea7be221e784d7ab4f292ac2dd0/src/process/ProcessManagerImpl.cpp#L842
I think it would be safer to use `getNumRunningProcesses` in that function?

src/process/ProcessManagerImpl.cpp
```diff
@@ -243,10 +241,7 @@ ProcessManagerImpl::tryProcessShutdown(std::shared_ptr<ProcessExitEvent> pe)
         auto runningIt = mProcesses.find(pid);
         if (runningIt != mProcesses.end())
         {
-            // Mark it as "ready to be killed"
-            mKillable.push_back(runningIt->second);
-            auto res = cleanShutdown(*runningIt->second);
-            return res;
+            return politelyShutdown(runningIt->second);
```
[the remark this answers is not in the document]
MonsieurNicolas: Yeah we should fall through to force kill if `politelyShutdown` returns false

MonsieurNicolas (changes requested):

src/process/ProcessManagerImpl.cpp
```diff
@@ -500,6 +497,8 @@ ProcessManagerImpl::cleanShutdown(ProcessExitEvent& pe)
 {
     ZoneScoped;
     if (!GenerateConsoleCtrlEvent(CTRL_C_EVENT, pe.mImpl->getProcessId()))
+    auto ec = ABORT_ERROR_CODE;
```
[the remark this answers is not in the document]
graydon: fixed.

src/process/ProcessManagerImpl.cpp
```diff
@@ -680,6 +679,8 @@ ProcessManagerImpl::cleanShutdown(ProcessExitEvent& pe)
 {
     ZoneScoped;
     const int pid = pe.mImpl->getProcessId();
+    auto ec = ABORT_ERROR_CODE;
+    pe->mImpl->cancel(ec);
```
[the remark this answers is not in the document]
graydon: fixed.

src/process/ProcessManagerImpl.cpp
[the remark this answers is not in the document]
graydon: Fixed.

src/process/ProcessManagerImpl.h
```diff
@@ -34,14 +34,16 @@ class ProcessManagerImpl : public ProcessManager
     uint64_t mTempFileCount{0};
 
     std::deque<std::shared_ptr<ProcessExitEvent>> mPending;
-    std::deque<std::shared_ptr<ProcessExitEvent>> mKillable;
+    std::set<std::shared_ptr<ProcessExitEvent>> mHavePolitelyShutdown;
```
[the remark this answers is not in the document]
graydon: Reworked to address this confusion.

src/process/ProcessManagerImpl.cpp
graydon: Fixed.

src/process/ProcessManagerImpl.cpp
```diff
@@ -160,11 +178,11 @@ ProcessManagerImpl::shutdown()
 {
     if (!mIsShutdown)
     {
+        std::lock_guard<std::recursive_mutex> guard(mProcessesMutex);
```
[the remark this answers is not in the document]
graydon: Fixed.

src/process/ProcessManagerImpl.cpp
```diff
@@ -126,27 +126,45 @@ class ProcessExitEvent::Impl
...
+// to do below.
+#ifndef _WIN32
+    auto ec = ABORT_ERROR_CODE;
+    mSigChild.cancel(ec);
+#endif
+
+    // Then trigger shutdown, if we haven't yet (it's idempotent). This will ask
+    // every running process to shut down politely and cancel all events,
+    // pending and running.
+    shutdown();
```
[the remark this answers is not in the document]
graydon: Implemented more-or-less as written. Seems to work!

MonsieurNicolas:
Something is going on wrt termination (at least on Windows):
the test "subprocess redirect to file" silently fails to delete the file at the end because the file is still in use, this in turn causes subsequent runs to fail because we check for the existence of the output file before starting the process.

A couple things about this test:
* we should call change that remove to `REQUIRE(std::remove(filename.c_str()) == 0)`
* we should use random file names in tests (and/or use the temp folder that we create)
* we should probably split that test into a positive "file doesn't exist" test (where the file does not exist) and a test that tests that indeed we're fast failing if the file already exist

Now as for the actual failure: I think it would be better if we invoke the callback when the process is actually done. I've never seen this fail before, so it seems that the new code is calling that callback earlier than in master.

src/process/ProcessManagerImpl.cpp
```diff
@@ -775,12 +861,14 @@ void
...
     ZoneScoped;
+    checkInvariants();
     if (mIsShutdown)
     {
         return;
     }
     std::lock_guard<std::recursive_mutex> guard(mProcessesMutex);
-    while (!mPending.empty() && gNumProcessesActive < mMaxProcesses)
+    while (!mPending.empty() &&
+           getNumRunningOrShuttingDownProcesses() < mMaxProcesses)
```
MonsieurNicolas, line 871: nit: `if (fs::exists(i->mImpl->mOutFile))` below should probably be `if (!i->mImpl->mOutFile.empty() && fs::exists(i->mImpl->mOutFile))` : that variant of `fs::exists` can throw and per standard I can't tell what the behavior is on empty filename as it's not a valid path.

src/process/ProcessManagerImpl.cpp
```diff
@@ -831,7 +884,6 @@ ProcessManagerImpl::maybeRunPendingProcesses()
 
             i->mImpl->run();
             mProcesses[i->mImpl->getProcessId()] = i;
-            ++gNumProcessesActive;
```
MonsieurNicolas, line 799: Dumb question: how likely process IDs get recycled? Is it possible to have a race where we blindly clobber a process that exited before with `mProcesses[i->mImpl->getProcessId()] = i` (before calling the callback)?

MonsieurNicolas (commented):
This version (with state tracking) makes a lot more sense to me. I can't see anything wrong with it, even though that test failing may point to some new issue caused by how we acknowledge process exit

src/process/ProcessManagerImpl.cpp
graydon: Good catch, fixed.

src/process/ProcessManagerImpl.cpp
graydon: Unlikely but fair. I'll add a guard and consider it a failed startup on the new process (which there's plenty of upper machinery to retry in any cases that matter).

src/process/ProcessManagerImpl.cpp
```diff
@@ -40,20 +41,89 @@
...
+            CLOG_WARNING(Process, "");
+            CLOG_WARNING(Process, "  likely 'missing command':");
+            CLOG_WARNING(Process, "");
+            CLOG_WARNING(Process, "    {}", cmdLine);
+            CLOG_WARNING(Process, "");
+            CLOG_WARNING(Process, "************");
+            CLOG_WARNING(Process, "");
+        }
+#endif
+        // FIXME: this doesn't _quite_ do the right thing; it conveys
```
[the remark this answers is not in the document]
graydon: No, this is the standard exit path (`WIFEXITED`) which gets both error and non-error exits. Non-error is zero which we need to pass through to `asio::error_code` to represent a "successful" exit in callers. Likewise with nonzero (and for the sake of thoroughness we pass the whole nonzero exit value through so users can fish it out with `ec.value`).

The only confusing / awkward part of this is that `asio::error_code` is assumed to transport a `std::error_category`-coded number rather than a process exit code, and so helper methods like `message` do the wrong thing. But there's no other place to _put_ the process exit code. Asio wasn't built assuming it would be conveying process-exits at all. We have to put it somewhere.

graydon:
> I think it would be better if we invoke the callback when the process is actually done.

@MonsieurNicolas the callback is fired when the process is actually done, if it exits of its own accord (if you think about it, this is the only time we _can_ call the callback when a process exits of its own accord -- we only learn it's exiting when it exits). The callback is fired _early_ if the user attempts to shut a process down explicitly, using one of the force-shutdown calls, intentionally: this was the contract before and it's preserved in this change.

The code before this change, on windows, left the file `hostname.txt` in the working directory you run the test from, when you run it: if you run the test twice, it fails the second time. The code with this change, on windows, does exactly the same thing: passes the first run, fails the second. If you didn't see it happen before, you got lucky (eg. ran the tests from a tempdir you cleaned yourself?) -- it definitely did this before the change too. I just checked.

I'm happy to change it / make the test fixes you propose to the test (this looks like a fairly old test file) but that's "fixing a test", not changing the implementation of process shutdown.

src/process/ProcessManagerImpl.cpp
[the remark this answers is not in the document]
MonsieurNicolas: ah ok - I thought you were saying that we have code paths that call `perror` (or equivalent), so would produce bogus log messages. If that's not the case, it seems fine then.

MonsieurNicolas:
@graydon indeed the test is failing the same way in master... I am not sure when things broke (can be scheduler changes or older), so not this PR - if you can make it that at least we don't fail because of old test artifacts left behind, it would be great.

graydon:
> @graydon indeed the test is failing the same way in master... I am not sure when things broke (can be scheduler changes or older), so not this PR - if you can make it that at least we don't fail because of old test artifacts left behind, it would be great.

@MonsieurNicolas already did so in the followup commit I posted last night. It now uses (and cleans up) a random tmpdir as requested.

MonsieurNicolas:
r+ dcd5237a395571ea42a664bf1782e5247662fa87

PR #2803, opened 2020-11-16
cancel stale GA workflows when PRs/branches get updated

MonsieurNicolas:
uses https://github.com/marketplace/actions/cancel-workflow-action


graydon:
r+ 860aad2

MonsieurNicolas:
@latobarita: retry

PR #2804, opened 2020-11-17
Cleanup peer type update

### LINKED ISSUE #2775: Overlay: TypeUpdate is confusing, needs cleanup
Currently, we use `TypeUpdate` enum to update peer's type in the database:
```
    enum class TypeUpdate
    {
        SET_OUTBOUND,
        SET_PREFERRED,
        REMOVE_PREFERRED,
        UPDATE_TO_OUTBOUND
    }
```

It's not clear what the difference is between `SET_OUTBOUND` and `UPDATE_TO_OUTBOUND`. Both names convey the same meaning, and there's only one update direction they both represent (that is, inbound --> outbound; preferred --> outbound is already covered by `REMOVE_PREFERRED`).

The code would be cleaner and easier to follow if `SET_OUTBOUND` and `UPDATE_TO_OUTBOUND` were consolidated.

marta-lokhova:
Resolves #2775 

minor cleanup. right now peer type update is really confusing as UPDATE_TO_OUTBOUND would only conditionally update the type. This is not exactly right: in the situation where we're using this, what we really want is to make sure that we're not "downgrading" preferred peers. This change aims to make this a little clearer. The downside of this is that we do an extra load per peer, but since this only happens on startup and once every 10 minutes (when we resolve peers), that doesn't seem like a significant overhead.

### DIFF
--- a/src/overlay/OverlayManagerImpl.cpp
+++ b/src/overlay/OverlayManagerImpl.cpp
@@ -346,9 +346,7 @@ OverlayManagerImpl::storePeerList(std::vector<PeerBareAddress> const& addresses,
                                   bool setPreferred, bool startup)
 {
     ZoneScoped;
-    auto typeUpgrade = setPreferred
-                           ? PeerManager::TypeUpdate::SET_PREFERRED
-                           : PeerManager::TypeUpdate::UPDATE_TO_OUTBOUND;
+    auto type = setPreferred ? PeerType::PREFERRED : PeerType::OUTBOUND;
     if (setPreferred)
     {
         mConfigurationPreferredPeers.clear();
@@ -363,7 +361,8 @@ OverlayManagerImpl::storePeerList(std::vector<PeerBareAddress> const& addresses,
 
         if (startup)
         {
-            getPeerManager().update(peer, typeUpgrade,
+            getPeerManager().update(peer, type,
+                                    /* preferredTypeKnown */ false,
                                     PeerManager::BackOffUpdate::HARD_RESET);
         }
         else
@@ -371,7 +370,8 @@ OverlayManagerImpl::storePeerList(std::vector<PeerBareAddress> const& addresses,
             // If address is present in the DB, `update` will ensure
             // type is correctly updated. Otherwise, a new entry is created.
             // Note that this won't downgrade preferred peers back to outbound.
-            getPeerManager().update(peer, typeUpgrade);
+            getPeerManager().update(peer, type,
+                                    /* preferredTypeKnown */ false);
         }
     }
 }
--- a/src/overlay/PeerManager.cpp
+++ b/src/overlay/PeerManager.cpp
@@ -324,9 +324,12 @@ PeerManager::update(PeerRecord& peer, TypeUpdate type)
 {
     switch (type)
     {
-    case TypeUpdate::SET_OUTBOUND:
+    case TypeUpdate::ENSURE_OUTBOUND:
     {
-        peer.mType = static_cast<int>(PeerType::OUTBOUND);
+        if (peer.mType == static_cast<int>(PeerType::INBOUND))
+        {
+            peer.mType = static_cast<int>(PeerType::OUTBOUND);
+        }
         break;
     }
     case TypeUpdate::SET_PREFERRED:
@@ -334,17 +337,9 @@ PeerManager::update(PeerRecord& peer, TypeUpdate type)
         peer.mType = static_cast<int>(PeerType::PREFERRED);
         break;
     }
-    case TypeUpdate::REMOVE_PREFERRED:
+    case TypeUpdate::ENSURE_NOT_PREFERRED:
     {
         if (peer.mType == static_cast<int>(PeerType::PREFERRED))
-        {
-            peer.mType = static_cast<int>(PeerType::OUTBOUND);
-        }
-        break;
-    }
-    case TypeUpdate::UPDATE_TO_OUTBOUND:
-    {
-        if (peer.mType == static_cast<int>(PeerType::INBOUND))
         {
             peer.mType = static_cast<int>(PeerType::OUTBOUND);
         }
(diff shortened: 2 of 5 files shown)


src/overlay/OverlayManagerImpl.cpp
```diff
@@ -346,32 +346,41 @@ OverlayManagerImpl::storePeerList(std::vector<PeerBareAddress> const& addresses,
                                   bool setPreferred, bool startup)
 {
     ZoneScoped;
-    auto typeUpgrade = setPreferred
-                           ? PeerManager::TypeUpdate::SET_PREFERRED
-                           : PeerManager::TypeUpdate::UPDATE_TO_OUTBOUND;
+    auto typeUpgrade = setPreferred ? PeerManager::TypeUpdate::SET_PREFERRED
```
MonsieurNicolas, line 349: What the old interface was trying (and failed) to do was to have `PeerManager` be the one responsible for setting the right type in the database so that callers don't have to care about this.

I think what you did here (remove `UPDATE_TO_OUTBOUND`) but I don't think the other changes in this method are right:
`SET_OUTBOUND`  should really behave like `UPDATE_TO_OUTBOUND` as the mechanism to "downgrade" a preferred peer is to use `REMOVE_PREFERRED`.

That said the names are just wrong as they don't really convey what they're doing:
what the interface is really about is describing transitions based on claims not states.

So maybe this PR should really be about deleting `SET_OUTBOUND` and rename the other one to something like `IS_PREFERRED_OR_OUTBOUND` (as a preferred peer is "better than outbound").

While we're at it, we could also rename the other ones to `IS_PREFERRED` and `IS_NOT_PREFERRED`, and I think we'll have something that makes sense?

MonsieurNicolas (changes requested):

src/overlay/OverlayManagerImpl.cpp
marta-lokhova: I don't think what you propose works. `SET_OUTBOUND` behaving like `UPDATE_TO_OUTBOUND` was where I started. However, I do not think you can let PeerManager be solely responsible for figuring out the type, because we're using this API in two different places with different contracts. First one: update DB peers based on config, but do not downgrade any known preferred peers. Second: set a peer outbound if we tried connecting to it, regardless of its current type. https://github.com/stellar/stellar-core/blob/adfe86c5281432e6f653271b4e313e852b0099ff/src/overlay/Peer.cpp#L1031

From what I understand, the latter case has to be unconditional. Specifically, if the existing peer in the DB is preferred, but actually it is not, we have to downgrade it to outbound (I guess this can happen if we previously learned it's preferred, then re-started the node with updated preferred peers, removing that peer). So, basically, at the time of the function call you won't know what upgrade type you need until you load the peer from the database, and `SET_OUTBOUND` sort of handles it in a hacky way. Overall, this API is very confusing, and probably needs a much bigger cleanup, but I thought I'd start somewhere.

MonsieurNicolas: Oh I missed that there was a nested conditional operator in `Peer::updatePeerRecordAfterEcho()` - this is evil :)

I still think that having more logic that deals with peer's state outside of "peer manager" is not a step forward as we're creating parallel logic to update state.
* If we wanted to update the state directly in code, we'd need to use the "peer" object (that represents a peer, and syncs with the database) - which would be a departure from the `PeerBareAddress` contracts.
    * Probably the right longer term move as it's more in line with how a data layer typically works, but a larger change compared to what we do now.
* Like you noted those transitions depend on the previous value, and this is exactly what `PeerManager::update` is supposed to do.
    * I think the more conservative change is to do what I was saying, which is "express facts" from the call site, and let "peer manager" to whatever it needs to do.

So, I think we should just change the interface of `PeerManager::update` to take better parameters:
* we can probably express all those intents with different methods instead of a generic "update"
* the path in "config" is really just trying to ensure that the entry is in the database
* the other path knows if the peer is preferred, so it should just be a parameter that peer manager can use to derive the actual values

marta-lokhova: @MonsieurNicolas updated the diff. The updated API has users provide the desired peer type (so no direction needed, which I think makes it easier to reason about), and a flag for whether we definitely know that a peer is preferred.

src/overlay/PeerManager.h
```diff
@@ -87,9 +86,13 @@ class PeerManager
...
 
     /**
-     * Update type of peer associated with given address.
+     * Update type of peer associated with given address. This function takes
+     * desired peer type, and whether the preferred type is definitely known (in
+     * some cases it is unknown whether a peer is preferred or not).
+     * Depending on the peer type stored in the DB, a new type is determined.
      */
-    void update(PeerBareAddress const& address, TypeUpdate type);
+    void update(PeerBareAddress const& address, PeerType desiredType,
```
MonsieurNicolas, line 94: `desiredType` doesn't seem accurate, I think it's more an `observedType`

src/overlay/PeerManager.cpp
```diff
@@ -326,30 +326,25 @@ PeerManager::update(PeerRecord& peer, TypeUpdate type)
     {
     case TypeUpdate::SET_OUTBOUND:
     {
-        peer.mType = static_cast<int>(PeerType::OUTBOUND);
+        if (peer.mType == static_cast<int>(PeerType::INBOUND))
```
MonsieurNicolas, line 329: `SET_OUTBOUND` and `SET_PREFERRED` should be called `ENSURE_OUTBOUND` and `ENSURE_NOT_PREFERRED`

MonsieurNicolas (changes requested):
This is going in the right direction: minor naming nits and we can merge.

I suspect we'll simplify this `PeerManager` interface more later but I am not sure when.

src/overlay/PeerManager.h
marta-lokhova: done

src/overlay/PeerManager.cpp
marta-lokhova: Done

MonsieurNicolas:
r+ 112f9c4c1e98fd6e863d1a71cc853bad9207cb01

Issue #2800, opened 2020-11-11
The <= operator for BitSet is a bit confusing


hidenori-shinohara:
From the discussion on https://github.com/stellar/stellar-core/pull/2771... (The PR is not directly related)

The `BitSet` class implements the `<=` operator which shows whether a set is a subset of the other. This seems to be intentional as the test `src/util/test/BitSetTests.cpp` indeed checks the behavior of `<=`, and this is used at at least one place in the code (https://github.com/stellar/stellar-core/blob/master/src/herder/QuorumIntersectionCheckerImpl.cpp#L275)

However, this seems a bit confusing because
- Not all sets are comparable (e.g., `{1, 2}` and `{3, 4}` are neither `<=` nor `>=`.)
- It's not immediately clear what `b1 <= b2` is whereas `b1.isSubsetEq(b2)` is.

graydon:
Hmm .. I dunno, I mean on the one hand I can't question the experience of feeling confusion on encountering it; on the other hand I'm not really sure it's out of place -- it's the standard partial order on the powerset lattice (of which sets are elements).

I can't decide if it meets (or if its associated `operator<` would meet) the C++ requirements for being a strict weak order. I think it would? But I guess if it's confusing we might as well remove it, I'm no particular fan of operator overloading anyways. I suspect I just added it to make the code read more like the paper.

jonjove:
I'm not committed either way, but I think we should remove this definition of `operator<=`. My main argument for removing it is that there isn't an obvious definition of what `b1 <= b2` means. For example, comparison operators on `std::set` mean lexicographical comparison.  Meanwhile,`std::bitset` doesn't define comparison at all, as far as I can tell.

rokopt:
> I can't decide if it meets (or if its associated `operator<` would meet) the C++ requirements for being a strict weak order.

I believe that it would fail those requirements because incomparability is not transitive?  For example {1} and {2} are incomparable and {2} and {1,3} are incomparable, but {1} and {1,3} are comparable.

MonsieurNicolas:
The responses here confirm that we should remove it :)

And yes, @rokopt is right which implies it's not even close to something usable with the standard library.

PR #2807, opened 2020-11-17
Remove the <= and bool operators from the BitSet class

hidenori-shinohara:
# Description

Resolves https://github.com/stellar/stellar-core/issues/2800

This PR removes the `<=` and `bool` operators from the `BitSet` class.

* The issue https://github.com/stellar/stellar-core/issues/2800 has some discussion on why we might consider removing `<=`.
* If we remove the `<=` operator but keep the `bool` operator, then `bitset1 <= bitset2` still compiles and runs (I think C++ implicitly converts each bitset into a boolean value and compare.) I felt that that might create some more confusion in the future (also the `bool` operator is defined to be `nonempty`, which may not be obvious to some people), so I decided that we might consider removing it, also.

# Checklist
- [x] Compiles
- [x] Ran all tests

### DIFF
--- a/src/herder/QuorumIntersectionCheckerImpl.cpp
+++ b/src/herder/QuorumIntersectionCheckerImpl.cpp
@@ -240,7 +240,8 @@ MinQuorumEnumerator::anyMinQuorumHasDisjointQuorum()
     {
         CLOG(TRACE, "SCP") << "checking for quorum in committed=" << mCommitted;
     }
-    if (auto committedQuorum = mQic.contractToMaximalQuorum(mCommitted))
+    auto committedQuorum = mQic.contractToMaximalQuorum(mCommitted);
+    if (!committedQuorum.empty())
     {
         if (mQic.isMinimalQuorum(committedQuorum))
         {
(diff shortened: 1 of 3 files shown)


src/herder/QuorumIntersectionCheckerImpl.cpp
```diff
@@ -270,9 +271,10 @@ MinQuorumEnumerator::anyMinQuorumHasDisjointQuorum()
     {
         CLOG(TRACE, "SCP") << "checking for quorum in perimeter=" << mPerimeter;
     }
-    if (auto extensionQuorum = mQic.contractToMaximalQuorum(mPerimeter))
+    auto extensionQuorum = mQic.contractToMaximalQuorum(mPerimeter);
+    if (!extensionQuorum.empty())
     {
-        if (!(mCommitted <= extensionQuorum))
+        if (!(mCommitted.isSubsetEq(extensionQuorum)))
```
MonsieurNicolas, line 277: remove extra `()`

MonsieurNicolas (changes requested):
minor nit

src/herder/QuorumIntersectionCheckerImpl.cpp
hidenori-shinohara: Fixed!

MonsieurNicolas:
r+ c487f50f612244d8f5d3a07e656f8409a19f8fd3

PR #2808, opened 2020-11-17
fix flaky herder test

MonsieurNicolas:
The test had a dependency on the hash function / std::unordered_map

### DIFF
--- a/src/herder/test/HerderTests.cpp
+++ b/src/herder/test/HerderTests.cpp
@@ -1386,6 +1386,11 @@ testSCPDriver(uint32 protocolVersion, uint32_t maxTxSize, size_t expectedOps,
                            : tx->getEnvelope().v1().tx.timeBounds.activate();
             tb.minTime = minTime;
             tb.maxTime = maxTime;
+            auto& sig = tx->getEnvelope().type() == ENVELOPE_TYPE_TX_V0
+                            ? tx->getEnvelope().v0().signatures
+                            : tx->getEnvelope().v1().signatures;
+            sig.clear();
+            tx->addSignature(root.getSecretKey());
             auto txSet = std::make_shared<TxSetFrame>(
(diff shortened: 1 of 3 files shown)


src/herder/test/HerderTests.cpp
```diff
@@ -471,7 +473,9 @@ testTxSetWithFeeBumps(uint32 protocolVersion)
                 feeBump(*app, account2, tx2, minBalance2 - minBalance0 - 199);
             txSet->add(fb2);
 
-            checkTrimCheck({fb1, fb2});
+            std::vector<TransactionFrameBasePtr> txs{fb1, fb2};
+            std::sort(txs.begin(), txs.end());
```
jonjove, line 477: I don't think this is ideal. While I agree it solves the problem, it leaves us open to having this problem again if we ever modify these tests. You should make this change in `checkTrimCheck` so that anyone who uses this function will have automatically accounted for it.

jonjove: Also it would probably be better if you sorted by hash-order rather than pointer-order (both here and on line 459) since the pointers passed in to `checkTrimCheck` aren't guaranteed to be the same ones that are actually in the `TransactionSet` (although in practice I think this is true today).

MonsieurNicolas: Agree with moving this into `checkTrimCheck` - fixed

re: hash order. This doesn't make sense here: tests check that those vectors (of shared pointers) are equal, and performs pointer comparison, we don't have a viable comparison function for `TransactionFrameBase`.

jonjove: Oh good point about the hash order. Someone definitely could misuse this function then, but if that does happen then I'm fairly certain that they would find out deterministically so it isn't too bad. Fixing this would be more effort than it's worth.

jonjove:
r+ 79154d2

PR #2811, opened 2020-11-19
Github Actions: work around stale cache

MonsieurNicolas:
This PR changes how build caches work:
* it actually makes caching work by working around a limitation of `actions/cache@v2` that does not update the cache it used (there is a PR open in their repo)
* it changes our cache size
   * there is a limit of 5GB for caches **project wide** - and the previous limit was 900MB, which means that with a few PRs, we would trash important caches like the one for the `auto` branch
   * we reduce the limit to 300MB - a build only needs about 200MB, so this allows to go slightly over that
* it changes the policy for caching
   * we switch back to using the default behavior for compiler signature `Hash(mtime + size)`
   * we force purge caches that are more than 1 month old. This acts as a safety net in case something else (standard library headers, etc) gets modified but is not tracked by ccache.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -12,12 +12,26 @@ jobs:
       matrix:
         toolchain: [ "gcc", "clang"]
     steps:
+      - name: Compute cache key
+        # this step works around a limitation in actions/cache
+        # that does not allow updating a cache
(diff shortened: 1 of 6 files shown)


graydon:
r+ 0d26a88

PR #2812, opened 2020-11-20
Use a randomized hasher in unordered sets/maps

### LINKED ISSUE #2809: Consider using a randomized hasher for checked builds
Every few months we run into issues (usually in tests but that can happen in production code), where code makes hidden assumptions on the iteration order for `unordered_map`.

Most recently in #2808 (in the past we had comparisons between two different `unordered_map` that would "usually" work).

I prototyped the change to increase the chance of detecting this kind of problem:
https://github.com/MonsieurNicolas/stellar-core/tree/randomHasher

I don't know if there is something better we can do; opening this issue for discussing for now

MonsieurNicolas:
# Description

Resolves #2809

The only real change is the first commit, rest is search/replace.

PR for now has this always enabled.
We could enable this on builds with `--enable-extrachecks` enabled, but having this always on gives better confidence that mitigations are in place in production environments.

### DIFF
--- a/Builds/VisualStudio/stellar-core.vcxproj
+++ b/Builds/VisualStudio/stellar-core.vcxproj
@@ -611,7 +611,6 @@ exit /b 0
     <ClCompile Include="..\..\src\work\Work.cpp" />
     <ClCompile Include="..\..\src\work\WorkScheduler.cpp" />
     <ClCompile Include="..\..\src\work\WorkSequence.cpp" />
-    <ClCompile Include="Source.cpp" />
   </ItemGroup>
   <ItemGroup>
     <ClInclude Include="..\..\lib\catch.hpp" />
--- a/Builds/VisualStudio/stellar-core.vcxproj
+++ b/Builds/VisualStudio/stellar-core.vcxproj
@@ -611,6 +611,7 @@ exit /b 0
     <ClCompile Include="..\..\src\work\Work.cpp" />
     <ClCompile Include="..\..\src\work\WorkScheduler.cpp" />
     <ClCompile Include="..\..\src\work\WorkSequence.cpp" />
+    <ClCompile Include="Source.cpp" />
   </ItemGroup>
   <ItemGroup>
     <ClInclude Include="..\..\lib\catch.hpp" />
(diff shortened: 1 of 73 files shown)


MonsieurNicolas:
Looking at the failure of *one* of the test cases (clang only), it seems, if I am not mistaken that this approach is actually not viable.

I started to read https://en.cppreference.com/w/cpp/container/unordered_set/operator_cmp and noticed the wording
> The behavior is also undefined if hash_function() and key_eq() do (until C++20)key_eq() does (since C++20) not have the same behavior on lhs and rhs or if operator== for Key is not a refinement of the partition into equivalent-key groups introduced by key_eq() (that is, if two elements that compare equal using operator== fall into different partitions)

the way I read this, it seems they may be using the hash function associated with `this` on *both* containers.

Even if this was not the case, we'd still need to fix the wrappers, as right now I am doing something like
```c++
template <class T, class Hasher = std::hash<T>> class RandHasher
{
    size_t mMixer{0};

  public:
    RandHasher()
    {
        mMixer =
            stellar::rand_uniform<size_t>(std::numeric_limits<size_t>::min(),
                                          std::numeric_limits<size_t>::max());
    }
    size_t
    operator()(T const& t) const
    {
        return Hasher()(t) ^ mMixer;
    }
};
```

which is unsafe as it does not satisfy the "stability requirement" for [std::hash](https://en.cppreference.com/w/cpp/utility/hash):
> For two parameters k1 and k2 that are equal, `std::hash<Key>()(k1) == std::hash<Key>()(k2)`

(note the use of default constructor here)

We could *maybe* ignore this and enforce that there is no way to "default construct" an instance of `RandHash` but I suspect we'd be potentially fighting with more bugs: we'd have to enforce constructing std::unordered_set/map with an actual hash object (so explicitly define all constructors).

MonsieurNicolas:
I am reducing the scope of that PR that use a seed per set/map type instead of per instance, not as good but this is definitely something we can include all the time

Builds/VisualStudio/stellar-core.vcxproj
```diff
@@ -611,6 +611,7 @@ exit /b 0
     <ClCompile Include="..\..\src\work\Work.cpp" />
     <ClCompile Include="..\..\src\work\WorkScheduler.cpp" />
     <ClCompile Include="..\..\src\work\WorkSequence.cpp" />
+    <ClCompile Include="Source.cpp" />
```
graydon: Is this intentional?

Builds/VisualStudio/stellar-core.vcxproj.filters
```diff
@@ -1137,6 +1137,9 @@
     <ClCompile Include="..\..\src\ledger\InternalLedgerEntry.cpp">
       <Filter>ledger</Filter>
     </ClCompile>
+    <ClCompile Include="Source.cpp">
```
graydon: Likewise this? Seems a typo

src/util/RandHasher.h
```diff
@@ -0,0 +1,24 @@
...
+
+namespace std
+{
+template <class T, class Hasher = std::hash<T>> class RandHasher
+{
+  public:
+    size_t
+    operator()(T const& t) const
+    {
+        static size_t mixer =
```
graydon, line 18: I'm not sure it's ok to be using a static local for this -- you could race with another thread on initialization and get a hashtable that uses a half-written `mixer` value or a `mixer` that's initialized twice.

graydon (changes requested):
Looks ok.

graydon (approved):

graydon:
r+ 91fa519

PR #2821, opened 2020-12-02
Fix close time cutoff when it is too small

marta-lokhova:
the bug makes close time validation too aggressive, as it does not respect `MAXIMUM_LEDGER_CLOSETIME_DRIFT`


MonsieurNicolas:
r+ 546497f0f17c1aec18854feb4db10155bc921d44

PR #2825, opened 2020-12-03
mark build as failed if there is nothing to do

MonsieurNicolas:
This fixes an issue where the transient build triggered by our CI robot is kicked off at the same time than the other build, and as it "succeeds in doing nothing", it creates a conflict with the build cache (we use the timestamp to uniquely identify a build for a given branch).

As added benefit, this avoids using cache capacity (currently 5GB for the entire project) for no good reason.


graydon:
r+ f41a306

PR #2840, opened 2020-12-14
Escape shell variable in Makefile.am

rokopt:
# Description

`make fuzz` uses a shell script, inlined in `Makefile.am`, which loops over a shell variable `$i`.  It's used as `fuzz$i.xdr` and is intended to generate numbered filenames.  But it should be (and used to be) `fuzz$$i.xdr`; as it is the `$i` is being treated as an unset Makefile variable, so instead of generating 10,000 files, the script generates `fuzz.xdr` and then overwrites it 9,999 times.

# Checklist
- [x] Compiles
- [ ] Ran all tests


jonjove:
r+ 3a07445

PR #2843, opened 2020-12-16
Improve HAS validation

marta-lokhova:
Additional checks based on bucket versions in HAS files. Specifically, 
* Bucket versions from oldest to newest are non-decreasing 
* Protocol 13 and higher bucketlists contain buckets of version 12 or higher

### DIFF
--- a/src/history/HistoryArchive.cpp
+++ b/src/history/HistoryArchive.cpp
@@ -17,6 +17,7 @@
 #include "main/StellarCoreVersion.h"
 #include "process/ProcessManager.h"
 #include "util/Fs.h"
+#include "util/GlobalChecks.h"
 #include "util/Logging.h"
 #include <Tracy.hpp>
 #include <fmt/format.h>
@@ -284,24 +285,78 @@ HistoryArchiveState::containsValidBuckets(Application& app) const
 {
     ZoneScoped;
     // This function assumes presence of required buckets to verify state
-    // Level 0 future buckets are always clear
-    assert(currentBuckets[0].next.isClear());
+    uint32_t minBucketVersion = 0;
+    uint32_t oldestBucketVersion = 0;
+    bool nonEmptySeen = false;
+    Hash const emptyHash;
 
-    for (uint32_t i = 1; i < BucketList::kNumLevels; i++)
-    {
-        auto& level = currentBuckets[i];
-        auto& prev = currentBuckets[i - 1];
-        Hash const emptyHash;
-
-        auto snap =
-            app.getBucketManager().getBucketByHash(hexToBin256(prev.snap));
-        assert(snap);
-        if (snap->getHash() == emptyHash)
+    auto validateBucketVersion = [&](uint32_t bucketVersion) {
+        if (bucketVersion < minBucketVersion)
         {
+            CLOG_ERROR(History,
+                       "Incompatible bucket versions: expected version "
+                       "{} or higher, got {}",
+                       minBucketVersion, bucketVersion);
+            return false;
+        }
+        minBucketVersion = bucketVersion;
+        return true;
+    };
+
+    // Process bucket, return version
+    auto processBucket = [&](std::string const& bucketHash) {
+        auto bucket =
+            app.getBucketManager().getBucketByHash(hexToBin256(bucketHash));
+        releaseAssert(bucket);
+        int32_t version = 0;
+        if (bucket->getHash() != emptyHash)
+        {
+            version = Bucket::getBucketVersion(bucket);
+            if (!nonEmptySeen)
+            {
+                nonEmptySeen = true;
+                oldestBucketVersion = version;
+            }
+        }
+        return version;
+    };
+
+    // Iterate bottom-up, from oldest to newest buckets
+    for (uint32_t i = BucketList::kNumLevels - 1; i >= 0; i--)
+    {
+        auto const& level = currentBuckets[i];
+
+        // Note: snap is always older than curr, and therefore must be processed
+        // first
+        if (!validateBucketVersion(processBucket(level.snap)) ||
+            !validateBucketVersion(processBucket(level.curr)))
+        {
+            return false;
+        }
+
+        // Level 0 future buckets are always clear
+        if (i == 0)
+        {
+            if (!level.next.isClear())
+            {
+                CLOG_ERROR(History,
+                           "Invalid HAS: next must be clear at level 0");
+                return false;
+            }
+            break;
+        }
+
+        // Validate "next" field
+        // Use previous level snap to determine "next" validity
+        auto const& prev = currentBuckets[i - 1];
+        uint32_t prevSnapVersion = processBucket(prev.snap);
+
+        if (!nonEmptySeen)
+        {
+            // No real buckets seen yet, move on
             continue;
         }
-        else if (Bucket::getBucketVersion(snap) >=
-                 Bucket::FIRST_PROTOCOL_SHADOWS_REMOVED)
+        else if (prevSnapVersion >= Bucket::FIRST_PROTOCOL_SHADOWS_REMOVED)
         {
             if (!level.next.isClear())
             {
@@ -316,6 +371,19 @@ HistoryArchiveState::containsValidBuckets(Application& app) const
             return false;
         }
     }
+
+    // By protocol "Bucket::FIRST_PROTOCOL_SHADOWS_REMOVED + 1", all buckets
+    // must be at least of version "Bucket::FIRST_PROTOCOL_SHADOWS_REMOVED"
+    if (minBucketVersion >= Bucket::FIRST_PROTOCOL_SHADOWS_REMOVED + 1 &&
+        oldestBucketVersion < Bucket::FIRST_PROTOCOL_SHADOWS_REMOVED)
+    {
+        CLOG_ERROR(History,
+                   "Invalid HAS: bucketlist contains bucket version {} that is "
+                   "too old",
+                   oldestBucketVersion);
+        return false;
+    }
+
     return true;
 }
 
--- a/docs/software/admin.md
+++ b/docs/software/admin.md
@@ -1074,6 +1074,10 @@ This section contains information that is useful to know but that should not sto
 [testnet.md](./testnet.md) is a short tutorial demonstrating how to
   configure and run a short-lived, isolated test network.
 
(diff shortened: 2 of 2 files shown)


src/history/HistoryArchive.cpp
```diff
@@ -288,22 +288,93 @@ HistoryArchiveState::containsValidBuckets(Application& app) const
...
+            return false;
+        }
+        return true;
+    };
+
+    // Process bucket, return version
+    auto getVersion = [&](std::string const& bucketHash) {
+        auto bucket =
+            app.getBucketManager().getBucketByHash(hexToBin256(bucketHash));
+        assert(bucket);
```
graydon, line 312: maybe prefer one of the releaseAssert* macros? Impervious to NDEBUG and now give you backtraces (on some platforms)

src/history/HistoryArchive.cpp
```diff
@@ -288,22 +288,93 @@ HistoryArchiveState::containsValidBuckets(Application& app) const
...
+            {
+                nonEmptySeen = true;
+                oldestBucketVersion = version;
+            }
+        }
+        return version;
+    };
+
+    // Iterate bottom-up, from oldest to newest buckets
+    for (uint32_t i = BucketList::kNumLevels - 1; i > 0; i--)
```
MonsieurNicolas, line 327: I think this loop can be changed to something easier to understand:
you already have a function to process a bucket `getVersion` (that should really be called `processBucket` as it has side effects), that also does `validateBucketVersion`, updates `minBucketVersion` and perform the check against `Bucket::FIRST_PROTOCOL_SHADOWS_REMOVED + 1`.

Then you just end up with a loop:
```
for (uint32_t i = BucketList::kNumLevels - 1; i > 0; i--)
{
   auto const& level = currentBuckets[i];
   if (!processBucket(level.snap) || !processBucket(level.cur))
   {
        return false;
   }
}
```
Note that you could also use `std::accumulate` here and pass around version information but I am not sure it's that much easier to read.

MonsieurNicolas, line 327: The current code doesn't check anything for `level[0].cur` ?

MonsieurNicolas (changes requested):

src/history/HistoryArchive.cpp
marta-lokhova: good catch! added an extra check.

marta-lokhova: Updated the first half of the function to use the loop like you suggested. Note that the check against `Bucket::FIRST_PROTOCOL_SHADOWS_REMOVED + 1` is done in the second half during `next` validation, so I kept it there.

marta-lokhova:
addressed your feedback @MonsieurNicolas, didn't squash commits yet so you can see the updates

src/history/HistoryArchive.cpp
MonsieurNicolas: A few questions.

Why is this true?
 >     // By protocol "Bucket::FIRST_PROTOCOL_SHADOWS_REMOVED + 1", all
 >     // buckets must be at least of version "Bucket::FIRST_PROTOCOL_SHADOWS_REMOVED"

Is it that it just happens to be true in pubnet? Wondering about testnet resets...

If we can keep that check, should it just be done after the loop as `minBucketVersion` covers everything?

```
        // By protocol "Bucket::FIRST_PROTOCOL_SHADOWS_REMOVED + 1", all
        // buckets must be at least of version "Bucket::FIRST_PROTOCOL_SHADOWS_REMOVED"
        if (minBucketVersion >= Bucket::FIRST_PROTOCOL_SHADOWS_REMOVED + 1 &&        
                oldestBucketVersion < Bucket::FIRST_PROTOCOL_SHADOWS_REMOVED)
        {
            CLOG_ERROR(History,
                        "Invalid HAS: bucket version {} is too old",
                        oldestBucketVersion);
            return false;
        }
```

The previous code was skipping checks when `snap` was the empty bucket, the new code computes `prevSnapVersion` in that case that returns `0`.
When this happens, are we sure that `level.next.hasOutputHash()` is true? Is it possible that we have instead `level.next.isClear()` ?

marta-lokhova: > Is it that it just happens to be true in pubnet? Wondering about testnet resets...

This check intended to validate older buckets (specifically, version 12 and 13). During testnet reset, we start generating new buckets with the latest protocol version. For example, right now the bucketlist at ledger 1000 on testnet has some version 15 buckets, and the rest are empty buckets (satisfying this check). This code would be invalid for new networks that start at some older protocol (for example, 10), then do multiple protocol upgrades within a short period of time. I assumed that was not a valid scenario. What do you think? 

> If we can keep that check, should it just be done after the loop as minBucketVersion covers everything?

I placed it there to fail fast in case the version is bad, to avoid unnecessary file reads from disk (that's how we get the bucket version).

> The previous code was skipping checks when snap was the empty bucket, the new code computes prevSnapVersion in that case that returns 0.
When this happens, are we sure that level.next.hasOutputHash() is true? Is it possible that we have instead level.next.isClear() ?

I _believe_ so. Prior to protocol 12, we always resolved merges-in-progress. Note that in case both inputs are empty buckets, an empty output bucket is still produced. This can be verified with some really old pubnet ledgers, as there were lots of empty buckets then. We've started actively clearing `next` in protocol 12. The only other scenario where we'd end up with pre-protocol-12 bucket levels that have `next` cleared is when the network just started closing ledgers, and some levels are brand new and haven't been filled yet. This is guarded by the `nonEmptySeen` check.

But there can never be too much of bucketlist paranoia! :) So maybe @graydon can take a look at my reasoning above as well?  ^^

MonsieurNicolas: > This check intended to validate older buckets (specifically, version 12 and 13). During testnet reset, we start generating new buckets with the latest protocol version

For this to be true we'd have to guarantee that no buckets are generated at protocol 0, right? I think this should be fine. I am going to double check the script that we use.
This would potentially bust private networks though? I think we should add a note somewhere about this.
Let's add a note at the end of the admin guide: a new section "Considerations when bootstrapping a new network", where we can explain that people need to upgrade right away to the latest version of the protocol before accepting any transactions, and provide the command to run to do it.

> I placed it there to fail fast in case the version is bad, to avoid unnecessary file reads from disk (that's how we get the bucket version).

I don't think fast failing here is that important, we're talking a few ms difference. I think the code is easier to understand when broken up: when I moved it in my local copy, I had then more questions on the main validation code that then fits easily on the screen.

graydon:
Ok, I _think_ I agree with your reasoning about the blank post-p12 and non-blank pre-p12 `next` fields. I think the old code was checking that too, and you're just preserving functionality here? It probably wouldn't hurt to add a comment about this since none of us seem to have 100% certainty.

(Really I'd like to test this against existing HASs to make sure we're not missing an edge case. I'd like to not get this wrong.)

A few further comments, 2 naming and 2 actual bugs (I think!):

  - I'd like to rename `minBucketVersion` to `newestBucketVersion`. It's not a minimum, it's actually a maximum if anything (over the bucketlist). But I don't think calling it min or max is helpful either way: it's the _newest_ version we've scanned in each iteration as we scan from old-to-new, and the fact that it gets updated in a non-decreasing fashion is a property to be checked, not its inherent definition.
  - I'd also like to rename `prev` to `newerLevel`. It was only called `prev` before because it was the level in the previous iteration fo the loop. Now that the loop is iterating in the opposite order, it's even less-good of a name!
  - As far as I can tell, now that the loop (after your changes to accommodate earlier review) counts _down to zero inclusive_, it actually underflows / wraps around to max uint32_t, and would run forever if it didn't have an early return inside it at level 0. I think it never processes the final check after the loop. This should .. probably be fixed? Unless I'm misreading.
  - I'm not sure I agree with the final check anyway. Like .. it might be true on pubnet, but it all depends on whether there is/was a 2 million ledger / 4 month delay between a network's p11->p12 and p12->p13 upgrades, right? That might not be true / have been true on some private network out there. Isn't it possible that there's a network that has a HAS that has p11, p12 and p13 buckets in it all at once?

Sorry for the lateness of this feedback.

MonsieurNicolas:
> (Really I'd like to test this against existing HASs to make sure we're not missing an edge case. I'd like to not get this wrong.)

Yeah you're probably right. Before we can merge this change, we should really double check that all HAS from pubnet pass this new validation code there might be conditions that we don't know about because code changed over the years.

> I'm not sure I agree with the final check anyway. Like .. it might be true on pubnet, but it all depends on whether there is/was a 2 million ledger / 4 month delay between a network's p11->p12 and p12->p13 upgrades, right? That might not be true / have been true on some private network out there. Isn't it possible that there's a network that has a HAS that has p11, p12 and p13 buckets in it all at once?

Private networks can just remove that check if they have to. I'd rather have checks as strict as possible on pubnet, and really optimize for the latest protocol version.

marta-lokhova:
> I think the old code was checking that too, and you're just preserving functionality here?

The old code was skipping empty buckets, I'm making this check a bit stricter. Specifically, as soon as we encounter a non-empty buckets of version earlier than 12, we expect the outputs to be resolved. 

> As far as I can tell, now that the loop (after your changes to accommodate earlier review) counts down to zero inclusive, it actually underflows / wraps around to max uint32_t, and would run forever if it didn't have an early return inside it at level 0. I think it never processes the final check after the loop. This should .. probably be fixed? Unless I'm misreading.

I don't think it underflows, as the level 0 check guards against that (otherwise, none of the tests for this would pass!). I think you are right thought about the final loop being skipped, `return` should have become `break` when the final check got moved.

marta-lokhova:
re: tests, here are HAS variants I tested:
* version 0 buckets, some levels empty
* all version 0 buckets, no empty levels
* mix of version 0 and version 11 buckets
* all version 11 buckets
* mix of version 11 and 12 buckets
* all version 12 buckets
* mix of version 12 and 13 buckets
* all version 13 buckets
* "current": mix of version 13 and 15 buckets

are there any other variants we want to test? I didn't test this against all pubnet HAS files, because it's time-consuming and I'm not sure it's even useful -- if the checks are passing against a particular version of HAS, but some other HAS file _of the same version_ doesn't pass them, that should point to the fact that the HAS file itself is corrupt.

MonsieurNicolas:
r+ 2eea0a9ca7b9c6630d8662b70935ddddb6fa4621

PR #2846, opened 2020-12-17
Dev container fixes

MonsieurNicolas:
This PR fixes a few issues in the dev container, in particular it switches to `mirror` which in addition to being more reliable, makes spinning up the image a lot faster.

Tested on my dev box, I could saturate my home network, where as the previous version was downloading very slowly (or just failing to find some packages).


MonsieurNicolas:
@latobarita: retry

PR #2848, opened 2020-12-17
Remove easylogging

graydon:
This just completes the "new logging" work by removing easylogging and setting spdlog to the default. It also removes the vestigial `el::` namespace and mops up some mistakes / missed things on windows.


Builds/VisualStudio/stellar-core.vcxproj
```diff
@@ -137,8 +138,9 @@ exit /b 0
...
-      <AdditionalIncludeDirectories>src;../../src;../../lib;../../lib/tracy;../../lib/libmedida/src;../../lib/soci/src/core;../../lib/autocheck/include;../../lib/cereal/include;../../lib/asio/asio/include;../../lib/xdrpp;../../lib/libsodium/src/libsodium/include;../../lib/fmt/include;../..;src/generated;../../lib/sqlite;c:\Program Files\PostgreSQL\9.5\include;%(AdditionalIncludeDirectories)</AdditionalIncludeDirectories>
+      <PreprocessorDefinitions>USE_SPDLOG;FMT_HEADER_ONLY=1;BUILD_TESTS;WIN32_LEAN_AND_MEAN;NOMINMAX;ASIO_STANDALONE;_WINSOCK_DEPRECATED_NO_WARNINGS;SODIUM_STATIC;ASIO_SEPARATE_COMPILATION;ASIO_ERROR_CATEGORY_NOEXCEPT=noexcept;TRACY_ENABLE;TRACY_ON_DEMAND;TRACY_NO_BROADCAST;TRACY_ONLY_LOCALHOST;USE_TRACY;_CRT_SECURE_NO_WARNINGS;_WIN32_WINNT=0x0601;WIN32;_MBCS;_CRT_NONSTDC_NO_DEPRECATE;%(PreprocessorDefinitions)</PreprocessorDefinitions>
+      <ForcedIncludeFiles>../../src/util/SpdlogTweaks.h</ForcedIncludeFiles>
```
MonsieurNicolas, line 142: mmm, I also missed that in the previous PR.

Why do we need to use this force include (I saw this in the Makefile as well)?
We never include spdlog directly, so the only place where we need to include it from is from `Logging.h` ?

MonsieurNicolas: If this is to build the various cpp files from spdlog, the easiest is to `#include` those cpp files directly from out own `spdlog.cpp` file (this is how we do it for asio - see `lib/asio.cpp`).
I am a bit nervous of injecting includes like this.

MonsieurNicolas (changes requested):
Looks good, only problem is that I think the previous PR broke console only logging

src/main/Config.cpp
```diff
@@ -129,12 +129,8 @@ Config::Config() : NODE_SEED(SecretKey::random())
     MAX_SLOTS_TO_REMEMBER = 12;
     METADATA_OUTPUT_STREAM = "";
 
-#ifdef USE_EASYLOGGING
-    LOG_FILE_PATH = "stellar-core.%datetime{%Y.%M.%d-%H:%m:%s}.log";
-#else
+    // Empty LOG_FILE_PATH causes a timestamp-based default in Logging.cpp
```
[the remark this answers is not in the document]
graydon: Ok. I can either invent a sentinel string or use like .. an option here? or some other discriminated type?

In earlier versions of this PR I did the timestamp-name calculation once, statically, in the default-setting here in Config.cpp, but to preserve behavioural parity with easylogging we need something that happens lazily, re-computes a new timestamp-name every time the log is rotated, to indicate the rotation-time.

Builds/VisualStudio/stellar-core.vcxproj
graydon: Yes, it needs to be included both in the .cpp files from spdlog that comprise the "compiled part" of spdlog _and_ in the .h files that comprise the public interface.

I can generate a lib/asio.cpp-alike thing if you prefer. Don't have a strong preference myself.

src/main/Config.cpp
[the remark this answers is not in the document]
MonsieurNicolas: Why not just
* changing this to `LOG_FILE_PATH="stellar-core.{datetime:%Y.%m.%d-%H:%M:%S}.log"`
* rename `mLastFilename` to `mLastFileNamePattern` (in Logging.cpp/h)
* always derive `filename` in `Logging::init` with
```
filename = fmt::format(mLastFileNamePattern, fmt::arg("datetime", fmt::localtime(time)));
```
 Note the use of `datetime` here as named parameter, that way we can potentially add new ones later, like a counter or whatever).
If the pattern doesn't specify any placeholder, parameters are ignored.

Builds/VisualStudio/stellar-core.vcxproj
MonsieurNicolas: please :)

src/main/Config.cpp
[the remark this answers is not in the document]
graydon: Mm, not a bad idea. I will give it a try.

graydon:
pushed fixed version.

src/main/Config.cpp
```diff
@@ -129,12 +129,11 @@ Config::Config() : NODE_SEED(SecretKey::random())
     MAX_SLOTS_TO_REMEMBER = 12;
     METADATA_OUTPUT_STREAM = "";
 
-#ifdef USE_EASYLOGGING
-    LOG_FILE_PATH = "stellar-core.%datetime{%Y.%M.%d-%H:%m:%s}.log";
+#ifdef _WIN32
```
MonsieurNicolas, line 132: I think we should just make the default the one with dashes or underscores and call it a day: our old default is the problem, Easylogging new this already as they use `kDefaultDateTimeFormatInFilename` as default that is `"%Y-%M-%d_%H-%m"`.

Builds/VisualStudio/stellar-core.vcxproj.filters
```diff
@@ -1143,11 +1143,32 @@
     <ClCompile Include="..\..\src\util\Backtrace.cpp">
       <Filter>util</Filter>
     </ClCompile>
+    <ClCompile Include="..\..\lib\spdlog\src\stdout_sinks.cpp">
```
MonsieurNicolas, line 1146: I think that those files have to be marked as "no compile" somehow, or we just don't include them in the project file.
I think only the `..\..\lib\spdlog.cpp` is needed.

MonsieurNicolas (changes requested):

Builds/VisualStudio/stellar-core.vcxproj.filters
graydon: ok. I'll remove them from the project file.

src/main/Config.cpp
graydon: Ok. If it breaks someone's ops logfile regex I .. guess it can be fixed there.

graydon:
Pushed change with fixes.

MonsieurNicolas:
r+ 7cc4763adc8b4c786c7ce0fc993d39ed76d53758

PR #2851, opened 2020-12-19
Tell libsodium not to download code from savannah.gnu.org in autogen.sh.

graydon:
Libsodium's autogen.sh step automatically downloads a fresh config.guess and config.sub from savannah.gnu.org, which .. is not something we really want it to be doing. At very least because savannah.gnu.org isn't always online.

Unfortunately we'll have to change this code slightly when 1.0.19 comes out, but thems the breaks.

### DIFF
--- a/autogen.sh
+++ b/autogen.sh
@@ -20,6 +20,14 @@ case "$1" in
 	;;
 esac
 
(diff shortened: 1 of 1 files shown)


MonsieurNicolas:
r+ 87fd89cc60d426c08199e4174dd046be7a2d19b9

PR #2858, opened 2021-01-05
Reduce default fuzz-testing account balances

rokopt:
# Description

Part of #1376

- Reduce default account balances while fuzzing to try to reduce the space that the fuzzer needs to explore.
- Correspondingly, reduce fees and reserves.
- Log some debug information when setup transactions fail.
- Don't crash when _non_-setup transactions fail -- we expect (or at least, hope) that the fuzzer will find some invalid-but-interesting inputs.

I have not given much thought yet to what really good numbers would be; I just shrank them by "a lot" and added some symbolic constants to try to clarify where some numbers were coming from (as well as fixing a bug that would make the fuzzer think that it had found a crash whenever it submitted an invalid transaction, such as one with a "too early" timestamp).  So I'm not sure whether we should accept this as is before I think more about the numbers we _really_ want, or whether we should wait to merge until after I have.  But I'm creating a PR at least to invite commentary on the structural changes / bug fix, as well as any suggestions about how to zero in on what numbers we should choose.

# Checklist
- [x] Compiles
- [x] Ran all tests

### DIFF
--- a/src/test/FuzzerImpl.cpp
+++ b/src/test/FuzzerImpl.cpp
@@ -4,6 +4,7 @@
 
 #include "test/FuzzerImpl.h"
 #include "ledger/LedgerTxn.h"
+#include "ledger/TrustLineWrapper.h"
 #include "main/Application.h"
 #include "main/Config.h"
 #include "overlay/OverlayManager.h"
@@ -32,11 +33,17 @@ namespace stellar
 namespace FuzzUtils
 {
 auto constexpr FUZZER_MAX_OPERATIONS = 5;
-auto constexpr INITIAL_LUMEN_AND_ASSET_BALANCE = 100000LL;
-auto constexpr NUMBER_OF_ASSETS_TO_ISSUE = 4;
+auto constexpr INITIAL_ACCOUNT_BALANCE = 1'000'000LL;    // reduced after setup
+auto constexpr INITIAL_ASSET_DISTRIBUTION = 1'000'000LL; // reduced after setup
+auto constexpr NUMBER_OF_ASSETS_TO_ISSUE = 4U;
+auto constexpr FUZZING_FEE = 1;
+auto constexpr FUZZING_RESERVE = 4;
+auto constexpr INITIAL_TRUST_LINE_LIMIT = 5 * INITIAL_ASSET_DISTRIBUTION;
+auto constexpr DEFAULT_NUM_TRANSACTIONS_TO_RESERVE_FEES_FOR = 10;
+auto constexpr DEFAULT_ASSET_AVAILABLE_FOR_TEST_ACTIVITY = 256;
 
 // must be strictly less than 255
-uint8_t constexpr NUMBER_OF_PREGENERATED_ACCOUNTS = 16;
+uint8_t constexpr NUMBER_OF_PREGENERATED_ACCOUNTS = 16U;
 
 void
 setShortKey(uint256& ed25519, int i)
@@ -405,6 +412,8 @@ getFuzzConfig(int instanceNumber)
     cfg.QUORUM_INTERSECTION_CHECKER = false;
     cfg.PREFERRED_PEERS_ONLY = false;
     cfg.RUN_STANDALONE = true;
+    cfg.TESTING_UPGRADE_DESIRED_FEE = FuzzUtils::FUZZING_FEE;
+    cfg.TESTING_UPGRADE_RESERVE = FuzzUtils::FUZZING_RESERVE;
 
     return cfg;
 }
@@ -567,6 +576,7 @@ TransactionFuzzer::initialize()
 {
     resetRandomSeed();
     mApp = createTestApplication(mClock, getFuzzConfig(0));
+    auto root = TestAccount::createRoot(*mApp);
 
     resetTxInternalState(*mApp);
     LedgerTxn ltxOuter(mApp->getLedgerTxnRoot());
@@ -578,7 +588,6 @@ TransactionFuzzer::initialize()
         // or enough to fill the first few bits such that we have a pregenerated
         // account for the last few bits of the 32nd byte of a public key, thus
         // account creation is over a deterministic range of public keys
-        auto root = TestAccount::createRoot(*mApp);
         xdr::xvector<Operation> ops;
         for (uint8_t i = 0; i < FuzzUtils::NUMBER_OF_PREGENERATED_ACCOUNTS; ++i)
         {
@@ -586,16 +595,9 @@ TransactionFuzzer::initialize()
             FuzzUtils::setShortKey(publicKey, i);
 
             auto createOp = txtest::createAccount(
-                publicKey,
-                FuzzUtils::INITIAL_LUMEN_AND_ASSET_BALANCE * 10000000);
+                publicKey, FuzzUtils::INITIAL_ACCOUNT_BALANCE);
 
             ops.emplace_back(createOp);
-
-            if (ops.size() == MAX_OPS_PER_TX)
-            {
-                attemptToApplyOps(ltx, root.getPublicKey(), ops, *mApp);
-                ops.clear();
-            }
 
             // select the first pregenerated account to be the hard coded source
             // account for all transactions
@@ -605,10 +607,7 @@ TransactionFuzzer::initialize()
             }
         }
 
-        if (!ops.empty())
-        {
-            attemptToApplyOps(ltx, root.getPublicKey(), ops, *mApp);
-        }
+        attemptToApplyOps(ltx, root.getPublicKey(), ops, *mApp);
 
         ltx.commit();
     }
@@ -637,16 +636,17 @@ TransactionFuzzer::initialize()
                 if (i != j)
                 {
                     // trust asset issuer
-                    auto trustOp = txtest::changeTrust(asset, INT64_MAX);
+                    auto trustOp = txtest::changeTrust(
+                        asset, FuzzUtils::INITIAL_TRUST_LINE_LIMIT);
                     trustOp.sourceAccount.activate() = toMuxedAccount(account);
                     ops.emplace_back(trustOp);
 
-                    // distribute asset
-                    auto distributeOp = txtest::payment(
-                        account, asset,
-                        FuzzUtils::INITIAL_LUMEN_AND_ASSET_BALANCE);
                     PublicKey issuer;
                     FuzzUtils::setShortKey(issuer, j);
+
+                    // distribute asset
+                    auto distributeOp = txtest::payment(
+                        account, asset, FuzzUtils::INITIAL_ASSET_DISTRIBUTION);
                     distributeOp.sourceAccount.activate() =
                         toMuxedAccount(issuer);
                     ops.emplace_back(distributeOp);
@@ -673,12 +673,12 @@ TransactionFuzzer::initialize()
         // +------------+-----+------+--------+------------------------------+
         // |  Account   | Bid | Sell | Amount | Price (in terms of Sell/Bid) |
         // +------------+-----+------+--------+------------------------------+
-        // | 0          | A   | B    |  1,000 | 3/2                          |
-        // | 1 (issuer) | A   | B    |  5,000 | 3/2                          |
-        // | 2          | A   | B    | 10,000 | 1/1                          |
-        // | 3 (issuer) | B   | A    |  1,000 | 10/9                         |
-        // | 4          | B   | A    |  5,000 | 10/9                         |
-        // | 0          | B   | A    | 10,000 | 22/7                         |
+        // | 0          | A   | B    |     10 | 3/2                          |
+        // | 1 (issuer) | A   | B    |     50 | 3/2                          |
+        // | 2          | A   | B    |    100 | 1/1                          |
+        // | 3 (issuer) | B   | A    |     10 | 10/9                         |
+        // | 4          | B   | A    |     50 | 10/9                         |
+        // | 0          | B   | A    |    100 | 22/7                         |
         // +------------+-----+------+--------+------------------------------+
         //
         // This gives us a good mixture of buy and sell, issuers with offers,
@@ -699,23 +699,23 @@ TransactionFuzzer::initialize()
                 ops.emplace_back(op);
             };
 
-            // A -> B acc0         : 1000A (3B/2A)
-            addOffer(A, B, pks[0], Price{3, 2}, 1000);
+            // A -> B acc0         : 1A (3B/2A)
+            addOffer(A, B, pks[0], Price{3, 2}, 10);
 
-            // A -> B acc1 (ISSUER): 5000A (3B/2A)
-            addOffer(A, B, pks[1], Price{3, 2}, 5000);
+            // A -> B acc1 (ISSUER): 5A (3B/2A)
+            addOffer(A, B, pks[1], Price{3, 2}, 50);
 
-            // A -> B acc2         : 10000A (1B/1A)
-            addOffer(A, B, pks[2], Price{1, 1}, 10000);
+            // A -> B acc2         : 10A (1B/1A)
+            addOffer(A, B, pks[2], Price{1, 1}, 100);
 
-            // B -> A acc3 (ISSUER): 1000B (10A/9B)
-            addOffer(B, A, pks[3], Price{10, 9}, 1000);
+            // B -> A acc3 (ISSUER): 1B (10A/9B)
+            addOffer(B, A, pks[3], Price{10, 9}, 10);
 
-            // B -> A acc4         : 5000B (10A/9B)
-            addOffer(B, A, pks[4], Price{10, 9}, 5000);
+            // B -> A acc4         : 5B (10A/9B)
+            addOffer(B, A, pks[4], Price{10, 9}, 50);
 
-            // B -> A acc0         : 10000B (22A/7B)
-            addOffer(B, A, pks[0], Price{22, 7}, 10000);
+            // B -> A acc0         : 10B (22A/7B)
+            addOffer(B, A, pks[0], Price{22, 7}, 100);
         };
 
         auto const XLM = txtest::makeNativeAsset();
@@ -729,16 +729,74 @@ TransactionFuzzer::initialize()
         genOffersForPair(ASSET_B, ASSET_C, {13, 2, 14, 3, 15}, ltx);
         genOffersForPair(ASSET_C, ASSET_D, {6, 3, 7, 4, 8}, ltx);
 
-        // construct transaction
-        auto tx = createFuzzTransactionFrame(mSourceAccountID, ops,
-                                             mApp->getNetworkID());
+        attemptToApplyOps(ltx, mSourceAccountID, ops, *mApp);
 
-        tx->attemptApplication(*mApp, ltx);
+        ltx.commit();
+    }
 
-        if (tx->getResultCode() != txSUCCESS)
+    {
+        LedgerTxn ltx(ltxOuter);
+
+        xdr::xvector<Operation> ops;
+
+        // Reduce account balances so that fuzzing has a better chance
+        // of exercising limits.
+        for (uint8_t i = 1; i < FuzzUtils::NUMBER_OF_PREGENERATED_ACCOUNTS; ++i)
         {
-            throw std::runtime_error("Error while initializing ledger state");
+            PublicKey account;
+            FuzzUtils::setShortKey(account, i);
+
+            // Reduce "account"'s native balance by paying the root.
+            auto ae = stellar::loadAccount(ltx, account);
+            auto const availableBalance =
+                getAvailableBalance(ltx.loadHeader(), ae);
+            auto const targetAvailableBalance =
+                FuzzUtils::DEFAULT_ASSET_AVAILABLE_FOR_TEST_ACTIVITY +
+                FuzzUtils::FUZZING_FEE *
+                    FuzzUtils::DEFAULT_NUM_TRANSACTIONS_TO_RESERVE_FEES_FOR;
+
+            if (availableBalance > targetAvailableBalance)
+            {
+                auto reduceNativeBalanceOp = txtest::payment(
+                    root, availableBalance - targetAvailableBalance);
+                reduceNativeBalanceOp.sourceAccount.activate() =
+                    toMuxedAccount(account);
+                ops.emplace_back(reduceNativeBalanceOp);
+            }
+
+            for (uint8_t j = 1; j <= FuzzUtils::NUMBER_OF_ASSETS_TO_ISSUE; ++j)
+            {
+                auto const asset = FuzzUtils::makeAsset(j);
+                if (i != j)
+                {
+                    PublicKey issuer;
+                    FuzzUtils::setShortKey(issuer, j);
+
+                    // Reduce "account"'s balance of asset "j" by paying the
+                    // issuer.
+                    auto tle = stellar::loadTrustLine(ltx, account, asset);
+                    auto const availableBalance =
+                        tle.getAvailableBalance(ltx.loadHeader());
+                    auto const targetAvailableBalance =
+                        FuzzUtils::DEFAULT_ASSET_AVAILABLE_FOR_TEST_ACTIVITY;
+
+                    if (availableBalance > targetAvailableBalance)
+                    {
+                        auto reduceNonNativeBalanceOp = txtest::payment(
+                            issuer, asset,
+                            availableBalance - targetAvailableBalance);
+                        reduceNonNativeBalanceOp.sourceAccount.activate() =
+                            toMuxedAccount(account);
+                        ops.emplace_back(reduceNonNativeBalanceOp);
+                    }
+                    // Here we could also reduce account "i"'s trust line limit
+                    // for asset "j" (testing the buying liabilities to
+                    // determine by how much).
+                }
+            }
         }
+
+        attemptToApplyOps(ltx, mSourceAccountID, ops, *mApp);
 
         ltx.commit();
     }
--- a/src/test/FuzzerImpl.cpp
+++ b/src/test/FuzzerImpl.cpp
@@ -16,9 +16,14 @@
 #include "transactions/OperationFrame.h"
 #include "transactions/SignatureChecker.h"
 #include "transactions/TransactionUtils.h"
+#include "util/Logging.h"
 #include "util/Math.h"
 #include "util/XDRCereal.h"
+#include "util/types.h"
+#include "xdr/Stellar-ledger-entries.h"
+#include "xdr/Stellar-transaction.h"
 
+#include <exception>
 #include <fmt/format.h>
 #include <xdrpp/autocheck.h>
 
@@ -501,19 +506,59 @@ isBadOverlayFuzzerInput(StellarMessage const& m)
     return m.type() == AUTH || m.type() == ERROR_MSG || m.type() == HELLO;
 }
 
+// Empties "ops" as operations are applied.  If `throwIfTxFails` is `false`,
+// ignores failures and might apply further transactions after failures.
 static void
 attemptToApplyOps(LedgerTxn& ltx, PublicKey const& sourceAccount,
-                  xdr::xvector<Operation> const& ops, Application& app)
+                  xdr::xvector<Operation>& ops, Application& app,
+                  bool const throwIfTxFails = true)
 {
-    auto txFramePtr =
-        createFuzzTransactionFrame(sourceAccount, ops, app.getNetworkID());
-    txFramePtr->attemptApplication(app, ltx);
-
-    if (txFramePtr->getResultCode() != txSUCCESS)
+    while (!ops.empty())
     {
-        throw std::runtime_error(fmt::format(
-            FMT_STRING("Error {} while applying operations while fuzzing"),
-            txFramePtr->getResultCode()));
+        auto endOpsInThisTx = ops.size() <= MAX_OPS_PER_TX
+                                  ? ops.end()
+                                  : ops.begin() + MAX_OPS_PER_TX;
+        auto txFramePtr = createFuzzTransactionFrame(
+            sourceAccount, xdr::xvector<Operation>(ops.begin(), endOpsInThisTx),
+            app.getNetworkID());
+        ops.erase(ops.begin(), endOpsInThisTx);
+        txFramePtr->attemptApplication(app, ltx);
+
+        if (throwIfTxFails)
+        {
+            if (txFramePtr->getResultCode() != txSUCCESS)
+            {
+                auto const msg = fmt::format(
+                    FMT_STRING("Error {} while applying operations "
+                               "while fuzzing: transaction result {}"),
+                    txFramePtr->getResultCode(),
+                    xdr_to_string(txFramePtr->getResult()));
+                LOG_FATAL(DEFAULT_LOG, "{}", msg);
+                throw std::runtime_error(msg);
+            }
+
+            for (auto const& opFrame : txFramePtr->getOperations())
+            {
+                auto const& op = opFrame->getOperation();
+                auto const& tr = opFrame->getResult().tr();
+                auto const opType = op.body.type();
+
+                if ((opType == MANAGE_BUY_OFFER &&
+                     tr.manageBuyOfferResult().success().offer.effect() ==
+                         MANAGE_OFFER_DELETED) ||
+                    (opType == MANAGE_SELL_OFFER &&
+                     tr.manageSellOfferResult().success().offer.effect() ==
+                         MANAGE_OFFER_DELETED))
+                {
+                    auto const msg = fmt::format(
+                        FMT_STRING("MANAGE_OFFER_DELETED while setting "
+                                   "up fuzzing -- operation is {}"),
+                        xdr_to_string(op));
+                    LOG_FATAL(DEFAULT_LOG, "{}", msg);
+                    throw std::runtime_error(msg);
+                }
+            }
+        }
     }
 }
 
@@ -609,11 +654,7 @@ TransactionFuzzer::initialize()
             }
         }
 
-        // construct transaction
-        auto txFramePtr = createFuzzTransactionFrame(mSourceAccountID, ops,
-                                                     mApp->getNetworkID());
-
-        txFramePtr->attemptApplication(*mApp, ltx);
+        attemptToApplyOps(ltx, mSourceAccountID, ops, *mApp);
 
         ltx.commit();
     }
--- a/src/test/FuzzerImpl.cpp
+++ b/src/test/FuzzerImpl.cpp
@@ -780,33 +780,26 @@ TransactionFuzzer::inject(std::string const& filename)
     {
         xdr::xdr_from_fuzzer_opaque(bins, ops);
     }
-    catch (...)
+    catch (std::exception const& e)
     {
         // in case of fuzzer creating an ill-formed xdr, generate an
         // xdr that will trigger a non-execution path so that the fuzzer
         // realizes it has hit an uninteresting case
+        LOG_TRACE(DEFAULT_LOG,
+                  "xdr::xdr_from_fuzzer_opaque() threw exception {}", e.what());
         return;
     }
     // limit operations per transaction to limit size of fuzzed input
     if (ops.size() < 1 || ops.size() > FuzzUtils::FUZZER_MAX_OPERATIONS)
     {
+        LOG_TRACE(DEFAULT_LOG, "invalid ops.size() {}", ops.size());
         return;
     }
 
-    // construct transaction
-    auto txFramePtr =
-        createFuzzTransactionFrame(mSourceAccountID, ops, mApp->getNetworkID());
-
-    if (Logging::logDebug("Ledger"))
-    {
-        LOG(DEBUG) << xdr_to_string(txFramePtr->getEnvelope());
-    }
-
     resetTxInternalState(*mApp);
+    LOG_TRACE(DEFAULT_LOG, "Fuzz ops ({}): {}", ops.size(), xdr_to_string(ops));
     LedgerTxn ltx(mApp->getLedgerTxnRoot());
-
-    // attempt to apply transaction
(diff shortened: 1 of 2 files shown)


src/test/FuzzerImpl.cpp
```diff
@@ -503,17 +514,27 @@ isBadOverlayFuzzerInput(StellarMessage const& m)
 
 static void
 attemptToApplyOps(LedgerTxn& ltx, PublicKey const& sourceAccount,
-                  xdr::xvector<Operation> const& ops, Application& app)
+                  xdr::xvector<Operation> const& ops, Application& app,
+                  bool const& throwIfTxFails = true)
```
MonsieurNicolas, line 518: pass by value.. this is a bool

src/test/FuzzerImpl.cpp
```diff
@@ -752,20 +759,9 @@ TransactionFuzzer::inject(std::string const& filename)
...
-    {
-        LOG(DEBUG) << xdr_to_string(txFramePtr->getEnvelope());
-    }
-
     resetTxInternalState(*mApp);
     LedgerTxn ltx(mApp->getLedgerTxnRoot());
-
-    // attempt to apply transaction
-    txFramePtr->attemptApplication(*mApp, ltx);
+    attemptToApplyOps(ltx, mSourceAccountID, ops, *mApp, false);
```
MonsieurNicolas, line 764: `attemptApplication` seems to do exactly what you wanted?

src/test/FuzzerImpl.cpp
MonsieurNicolas: not sure you need this parameter though as `attemptToApplyOps` is only called when performing setup and we always want to fail?

MonsieurNicolas (changes requested):

src/test/FuzzerImpl.cpp
rokopt: > pass by value.. this is a bool

Fixed.

> not sure you need this parameter though as attemptToApplyOps is only called when performing setup and we always want to fail?

It's also called by `TransactionFuzzer::inject()`, which is what we call when running fuzz tests, isn't it?

src/test/FuzzerImpl.cpp
rokopt: `attemptApplication()` is a method of `FuzzTransactionFrame`, which we were creating in several places -- `attemptToApplyOps()` factors that out.  It also factors out the `throw` when application fails, which we want during setup (though not while running generated test cases).  Is there a better way of avoiding that duplication?

src/test/FuzzerImpl.cpp
```diff
@@ -27,12 +28,20 @@ namespace stellar
...
 auto constexpr NUMBER_OF_ASSETS_TO_ISSUE = 4;
+auto constexpr FUZZING_FEE = 1;
+auto constexpr FUZZING_RESERVE = 4;
 
 // must be strictly less than 255
 uint8_t constexpr NUMBER_OF_PREGENERATED_ACCOUNTS = 16;
 
+// As a starting point, twice what we need for setup.
+auto constexpr INITIAL_STROOP_BALANCE =
+    2LL * (2 * FUZZING_RESERVE + (FUZZING_FEE + INITIAL_ASSET_DISTRIBUTION) *
```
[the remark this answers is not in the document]
rokopt: I like this idea.  I've added a revision which factors out the setup parameters from the execution of the operations that perform the setup, so that the configuration can be chosen in one place, and account balances in particular can be computed with knowledge of all relevant configuration parameters.  I've used something like your formula for that balance computation.

I still don't think that the current parameters will be the "final" (inevitably eternally provisional) ones, but I think that this reorganization should make it easy to tweak individual parameters and have any others that need to adapt do so automatically.  I think it will also be extensible to other enhancements, such as starting to use some of the trust line flags.

src/test/FuzzerImpl.cpp
```diff
@@ -644,9 +647,10 @@ TransactionFuzzer::initialize()
                                                          std::vector<int> pks) {
             auto addOfferParameters = [&accountParameters, &offerParameters](
                                           Asset bid, Asset sell, int pk,
-                                          Price price, int64 amount) {
-                offerParameters.emplace_back(
-                    FuzzUtils::OfferParameters(bid, sell, pk, price, amount));
+                                          Price price, int64 amount,
+                                          bool passive = false) {
```
jonjove, line 651: Nitpick: don't use a default parameter here. It's overkill for a function that you are going to call a handful of times, all within this one scope.

src/test/FuzzerImpl.cpp
```diff
@@ -800,7 +805,11 @@ TransactionFuzzer::initialize()
 
         for (auto const& offer : offerParameters)
         {
-            auto op = txtest::manageOffer(0, offer.mBid, offer.mSell,
+            auto op =
+                offer.mPassive
+                    ? txtest::createPassiveOffer(offer.mSell, offer.mBid,
+                                                 offer.mPrice, offer.mAmount)
+                    : txtest::manageOffer(0, offer.mBid, offer.mSell,
```
jonjove, line 812: These lines can't both be right. The assets come in the same order in both functions.

src/test/FuzzerImpl.cpp
```diff
@@ -33,15 +33,62 @@ auto constexpr NUMBER_OF_ASSETS_TO_ISSUE = 4;
...
+        , mNumAssetsToDistribute(NUMBER_OF_ASSETS_TO_ISSUE)
+        , mAvailableForTestActivity(DEFAULT_STROOPS_AVAILABLE_FOR_TEST_ACTIVITY)
+    {
+    }
+};
+
+class OfferParameters
+{
+  public:
+    stellar::Asset mBid;
```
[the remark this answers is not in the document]
jonjove: Also these names don't match our normal convention. One should be `mBuying` and the other `mSelling`.

src/test/FuzzerImpl.cpp
```diff
@@ -602,6 +606,11 @@ TransactionFuzzer::initialize()
     // We don't issue assets from account 0.
     accountParameters[0].mNumAssetsToDistribute = 0;
 
+    // Make some issuing accounts require authorization.
+    accountParameters[1].mAccountFlags = AUTH_REQUIRED_FLAG;
+    accountParameters[2].mAccountFlags =
+        AUTH_REQUIRED_FLAG | AUTH_REVOCABLE_FLAG;
```
jonjove, line 614: You also probably want some accounts with `AUTH_IMMUTABLE_FLAG`.

src/test/FuzzerImpl.cpp
```diff
@@ -745,11 +765,22 @@ TransactionFuzzer::initialize()
                     trustOp.sourceAccount.activate() = toMuxedAccount(account);
                     ops.emplace_back(trustOp);
 
+                    PublicKey issuer;
+                    FuzzUtils::setShortKey(issuer, j);
+
+                    if ((accountParameters[j].mAccountFlags &
```
jonjove, line 776: Probably want some to be authorized and some not to be authorized, right? What about authorized to maintain liabilities?

src/test/FuzzerImpl.cpp
```diff
@@ -19,6 +19,8 @@
 #include "util/Logging.h"
 #include "util/Math.h"
 #include "util/XDRCereal.h"
+#include "util/optional.h"
```
jonjove, line 22: Nitpick: I don't think this is used.

src/test/FuzzerImpl.cpp
```diff
@@ -27,12 +30,72 @@ namespace stellar
...
+auto constexpr FUZZING_FEE = 1;
+auto constexpr FUZZING_RESERVE = 4;
+auto constexpr INITIAL_TRUST_LINE_LIMIT = 3 * INITIAL_ASSET_DISTRIBUTION;
+auto constexpr DEFAULT_NUM_TRANSACTIONS_TO_RESERVE_FEES_FOR = 10;
+auto constexpr DEFAULT_STROOPS_AVAILABLE_FOR_TEST_ACTIVITY = 32LL;
 
 // must be strictly less than 255
 uint8_t constexpr NUMBER_OF_PREGENERATED_ACCOUNTS = 16;
 
+class AccountParameters
```
MonsieurNicolas, line 44: do we really need classes etc for this? Ideally all of this would be constexpr so that we don't have any overhead related to this at runtime

src/test/FuzzerImpl.cpp
MonsieurNicolas: I think this belongs to a different PR: I think this should be covered by
`Trustlines and orders seem "uninteresting". Setup is fully symetrical. No variations on issuer/trustline flags.`

src/test/FuzzerImpl.cpp
MonsieurNicolas: see my other comment, balances, limits and offers are also symmetrical. I am not sure a PR is the place to brainstorm on how we should mix things up.

src/test/FuzzerImpl.cpp
```diff
@@ -775,6 +884,76 @@ TransactionFuzzer::xdrSizeLimit()
...
 }
 
+// Given an Operation generated with autocheck through the XDR generator, try in
+// operation-type-specific ways to constrain the ranges of parameters into
+// spaces that will make it practical for the fuzzer to explore, while
+// preserving as much of the randomness as we can within the constrained space,
+// and occasionally exploring outside the constrained space to search for bugs
+// that would only be triggered by larger parameter values.
+void
+TransactionFuzzer::constrainOperation(Operation& op)
```
MonsieurNicolas, line 894: This change is out of scope for this PR. This seems to be loosely related to `The data generated to seed the fuzzer via the gen-fuzz is very low quality` which is much further down in the list of things to do.

MonsieurNicolas (changes requested):

src/test/FuzzerImpl.cpp
```diff
@@ -548,6 +595,85 @@ TransactionFuzzer::initialize()
...
+        auto addAssetPairParameters = [&accountParameters,
+                                       &offerParameters](Asset A, Asset B,
+                                                         std::vector<int> pks) {
+            auto addOfferParameters = [&accountParameters, &offerParameters](
+                                          Asset bid, Asset sell, int pk,
+                                          Price price, int64 amount) {
+                offerParameters.emplace_back(
+                    FuzzUtils::OfferParameters(bid, sell, pk, price, amount));
+                ++accountParameters[pk].mNumOffers;
+                accountParameters[pk].mOfferLiabilities +=
```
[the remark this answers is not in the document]
rokopt: > You shouldn't be computing this directly. The actual underlying computations are complicated so this may not work in general. Use getOfferSellingLiabilities or getOfferBuyingLiabilities.

Thanks -- I've pushed a commit to use `getOfferBuyingLiabilities()`.

> Also this calculation doesn't make sense to me. You aren't distinguishing between liabilities for different assets.

Currently the only thing that I'm using `AccountParameters::mOfferLiabilities` for is to help compute the total balance that we're going to give an account when it's created during fuzzer initialization, which I thought would only depend upon the total liabilities for all assets.  Does it already need to be per-asset for some reason I haven't thought of?  (I can imagine that I might need to make it that way in the future to support more specifically-targeted setup parameters -- for example, perhaps, to try to guide the fuzzer towards test cases that would try to exceed trustline limits without exceeding account balances?)

src/test/FuzzerImpl.cpp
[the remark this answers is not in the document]
rokopt: > Nitpick: no need for fully-qualified names here. You're in the stellar namespace.

> Also these names don't match our normal convention. One should be mBuying and the other mSelling.

Both fixed (in one commit).

src/test/FuzzerImpl.cpp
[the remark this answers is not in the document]
jonjove: This is a fundamental misunderstanding of liabilities. I strongly encourage you to read [CAP-0003](https://github.com/stellar/stellar-protocol/blob/master/core/cap-0003.md) to get a more thorough understanding. Liabilities are always tied to the asset. You should think of them as having units equal to the asset, so adding them for different assets can never make sense. Specifically, the account balance should only be influenced by the native selling liabilities (eg. the total amount of native asset that would be sold if all offers selling native executed simultaneously).

src/test/FuzzerImpl.cpp
rokopt: > These lines can't both be right. The assets come in the same order in both functions.

Good catch.  No idea how I managed to do that.  Fixed.

src/test/FuzzerImpl.cpp
rokopt: > Nitpick: don't use a default parameter here. It's overkill for a function that you are going to call a handful of times, all within this one scope.

Done.

src/test/FuzzerImpl.cpp
rokopt: > Nitpick: I don't think this is used.

Good catch.  (I'd used it in a previous iteration of the code, then changed it not to use `optional`.)  Removed.

src/test/FuzzerImpl.cpp
rokopt: > I think this belongs to a different PR: I think this should be covered by
Trustlines and orders seem "uninteresting". Setup is fully symetrical. No variations on issuer/trustline flags.

Yeah, that's what I was working on there.  I've removed that commit from this PR (and put it in a new branch).

src/test/FuzzerImpl.cpp
rokopt: > see my other comment, balances, limits and offers are also symmetrical. I am not sure a PR is the place to brainstorm on how we should mix things up.

Moved to a separate PR.  We can discuss whatever further variations we want elsewhere (issue #1376?).

src/test/FuzzerImpl.cpp
rokopt: > This change is out of scope for this PR. This seems to be loosely related to The data generated to seed the fuzzer via the gen-fuzz is very low quality which is much further down in the list of things to do.

Yes, that's what I was trying to start on.  I've moved that commit out of this PR and into a new branch.

src/test/FuzzerImpl.cpp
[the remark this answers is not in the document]
rokopt: Okay, now that you say this I realize I'm adding meters to electron-volts.

> Specifically, the account balance should only be influenced by the native selling liabilities (eg. the total amount of native asset that would be sold if all offers selling native executed simultaneously).

Done.

src/test/FuzzerImpl.cpp
```diff
@@ -644,14 +644,17 @@ TransactionFuzzer::initialize()
...
+                if (sell.type() == ASSET_TYPE_NATIVE)
+                {
+                    LedgerEntry offerEntry;
+                    offerEntry.data.type(OFFER);
+                    offerEntry.data.offer().buying = bid;
+                    offerEntry.data.offer().selling = sell;
+                    offerEntry.data.offer().price = price;
+                    offerEntry.data.offer().amount = amount;
+                    accountParameters[pk].mNativeOfferLiabilities +=
+                        getOfferBuyingLiabilities(header, offerEntry);
```
jonjove, line 656: I think you want selling liabilities? The buying liabilities relate to the amount you would receive if all of the offers were crossed simultaneously. Think about it this way, if you are selling native then you are _not_ buying native so once again these all have different units.

rokopt: Indeed!  Fixed.

src/test/FuzzerImpl.cpp
```diff
@@ -33,15 +33,62 @@ auto constexpr NUMBER_OF_ASSETS_TO_ISSUE = 4;
...
+        , mPrice(price)
+        , mAmount(amount)
+    {
+    }
+};
+
+int64_t
+initialAccountBalance(AccountParameters const& parameters)
+{
+    return (2 + parameters.mNumOffers) * FUZZING_RESERVE +
```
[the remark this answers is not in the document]
rokopt: > You aren't accounting for the number of trust lines.

Good point.  Added.

> Why does this depend on the number of assets to distribute? Those assets don't have a native balance so that shouldn't matter at all here.

That code derived from the same misunderstanding of liabilities that you pointed out in other comments.  I've removed it.

src/test/FuzzerImpl.cpp
rokopt: The classes are for gathering the configuration parameter calculations into one place preceding all of the transaction application, the idea being to make it easy to see where to modify parameters, and dealing with cases where a single operation affects multiple parameters, such as how a single offer can induce reserve requirements and selling liabilities.

I will attempt to measure performance impacts.  But with classes or without, I'm not sure how we'd correctly pre-compute required balances as `constexpr`s: as @jonjove pointed out, we should be (and now are) using `getOfferSellingLiabilities()`, which would require changes outside the fuzzer to make usable as a `constexpr`.

I'm open to ideas about that, of course.

rokopt:
Many of the changes that I've made responding to review comments have been churning code (often more than once) from previous commits, and I've just finished a round of responding to comments, so I've squashed the commits to date.

rokopt:
I've pushed a new commit to check for `MANAGE_OFFER_DELETED` effects during setup, since if we get any of those, then we're not actually creating all the offers that we're intending to create.

src/test/FuzzerImpl.cpp
```diff
@@ -537,21 +714,21 @@ TransactionFuzzer::initialize()
...
+                ops.clear();
+            }
+
             PublicKey publicKey;
             FuzzUtils::setShortKey(publicKey, i);
 
             auto createOp = txtest::createAccount(
                 publicKey,
-                FuzzUtils::INITIAL_LUMEN_AND_ASSET_BALANCE * 10000000);
+                FuzzUtils::initialAccountBalance(accountParameters[i]));
```
jonjove, line 786: I can't help but feel like you are doing a lot of work here that you don't need to do. It's hard for me to even review this PR as it stands because I think that if we were to simplify what we are doing here, most of this PR wouldn't even exist.

Here's my core observation: you aren't computing the initial trust line balance, so why are you computing the initial account balance?

I think that there is a pretty elegant way to back into the desired outcome. For each account A:
1. Create A with an unnecessarily high balance (eg. 1M native, even though reserves and liabilities are in the range 0-100).
2. Create trust lines, offers, other ledger entries for A
3. Use `getMinBalance` or `getAvailableBalance` to determine the amount of excess funds
4. Pay most of the excess funds back to the root, leaving `DEFAULT_STROOPS_AVAILABLE_FOR_TEST_ACTIVITY` behind

The advantages of such an approach are primarily:
1. You can do whatever set-up you want without much risk of failure; if it does fail, you'll find out immediately when you try to run it anyway
2. You don't need to guess the balance in advance which is requiring you to do a lot of tracking and run-time computation
3. You are outsourcing all computation to the internals of stellar-core, so you don't need to depend on explicitly computing liabilities or minimum balances. This will make the solution more robust to future protocol changes

Would this work? Or am I missing something here?

rokopt: It sounds plausible to me.  Besides paying back excess funds after setup, should we also start with high trust line limits and reduce them after setup, for symmetrical reasons?

jonjove: Keep it simple first. Let's see what happens if you do this for accounts, and we can decide if we want to do it for trust lines too.

src/test/FuzzerImpl.cpp
```diff
@@ -503,17 +517,82 @@ isBadOverlayFuzzerInput(StellarMessage const& m)
...
+                         .offer.effect() == MANAGE_OFFER_DELETED))
+            {
+                auto const msg =
+                    fmt::format(FMT_STRING("MANAGE_OFFER_DELETED while setting "
+                                           "up fuzzing -- op type {}"),
+                                op->getOperation().body.type());
+                LOG_FATAL(DEFAULT_LOG, "{}", msg);
+                if (op->getOperation().body.type() == MANAGE_SELL_OFFER)
+                {
+                    LOG_FATAL(
```
jonjove, line 566: A few comments here:
- This would have been way nicer if you had taken a reference to `op->getOperation().body.manageSellOfferOp()` and use it in the logging. Analogous comment applies on 581,591.
- This code is essentially a duplicate of the code on 581,591. You can easily refactor this into a static (this compilation unit only) template function. You could also use the lambda auto argument functionality to achieve the same.

jonjove: But more importantly, why even write this log message? You could do something like
```
auto const msg = fmt::format(FMT_STRING("MANAGE_OFFER_DELETED while setting up fuzzing. Operation is {}"), xdr_to_string(op));
```
Always go for the simple solution. Try to use as much existing functionality as possible. If you see yourself writing a lot of code that you've never seen anywhere else in stellar-core, you should probably be suspicious that there is an easier way.

src/test/FuzzerImpl.cpp
```diff
@@ -503,17 +517,82 @@ isBadOverlayFuzzerInput(StellarMessage const& m)
...
+            {
+                LOG_ERROR(DEFAULT_LOG, "op {}: result code {}",
+                          op->getOperation().body._xdr_discriminant(),
+                          op->getResultCode());
+            }
+            throw std::runtime_error(msg);
+        }
+        for (auto const& op : opFrames)
+        {
+            if ((op->getOperation().body.type() == MANAGE_BUY_OFFER &&
```
jonjove, line 546: This is pretty unpleasant to look at. Use references (similar to what I was saying in the comment on 566) to `op->getOperation().body` and `op->getResult().tr()` to trim this down.

src/test/FuzzerImpl.cpp
```diff
@@ -688,17 +763,87 @@ TransactionFuzzer::initialize()
...
+        // Reduce account balances so that fuzzing has a better change
+        // of exercising limits.
+        for (uint8_t i = 0; i < FuzzUtils::NUMBER_OF_PREGENERATED_ACCOUNTS; ++i)
         {
-            throw std::runtime_error("Error while initializing ledger state");
+            PublicKey account;
+            FuzzUtils::setShortKey(account, i);
+
+            // Reduce "account"'s native balance by paying the root.
+            LedgerKey accountKey;
```
jonjove, line 784: Look around the transaction subsystem to see how we do things. Be idiomatic: there are helpers for everything that we do often (mostly in transaction/TransactionUtils.cpp). In this case, `loadAccount` does all the work in 784,787.

src/test/FuzzerImpl.cpp
```diff
@@ -688,17 +763,87 @@ TransactionFuzzer::initialize()
...
+                    PublicKey issuer;
+                    FuzzUtils::setShortKey(issuer, j);
+
+                    // Reduce "account"'s balance of asset "j" by paying the
+                    // issuer.
+                    LedgerKey trustLineKey;
+                    trustLineKey.type(TRUSTLINE);
+                    trustLineKey.trustLine().accountID = account;
+                    trustLineKey.trustLine().asset = asset;
+                    LedgerEntry trustLineEntry =
```
jonjove, line 818: Same thing as above: `loadTrustLine`.

src/test/FuzzerImpl.cpp
```diff
@@ -688,17 +763,87 @@ TransactionFuzzer::initialize()
...
+            PublicKey account;
+            FuzzUtils::setShortKey(account, i);
+
+            // Reduce "account"'s native balance by paying the root.
+            LedgerKey accountKey;
+            accountKey.type(ACCOUNT);
+            accountKey.account().accountID = account;
+            LedgerEntry accountEntry = ltx.load(accountKey).current();
+            auto minBalance = getMinBalance(
+                mApp->getLedgerManager().getLastClosedLedgerHeader().header,
```
jonjove, line 789: Never do this. Always use the ledger header from `LedgerTxn` (see `loadHeader`).

src/test/FuzzerImpl.cpp
```diff
@@ -688,17 +763,87 @@ TransactionFuzzer::initialize()
...
         {
-            throw std::runtime_error("Error while initializing ledger state");
+            PublicKey account;
+            FuzzUtils::setShortKey(account, i);
+
+            // Reduce "account"'s native balance by paying the root.
+            LedgerKey accountKey;
+            accountKey.type(ACCOUNT);
+            accountKey.account().accountID = account;
+            LedgerEntry accountEntry = ltx.load(accountKey).current();
```
jonjove, line 787: This is quite unidiomatic for a few reasons:
1. `accountEntry` does not have type `AccountEntry` so the name is misleading
2. You copied out of a `LedgerTxnEntry` instead of just using `current` when you need it. This is bypassing some of the `LedgerTxn` safety guarantees. Don't take references to or copy from `current`, unless there is some deep necessity.

src/test/FuzzerImpl.cpp
```diff
@@ -688,17 +763,87 @@ TransactionFuzzer::initialize()
...
+            auto minBalance = getMinBalance(
+                mApp->getLedgerManager().getLastClosedLedgerHeader().header,
+                accountEntry.data.account());
+            auto remainingBalance =
+                minBalance +
+                FuzzUtils::DEFAULT_ASSET_AVAILABLE_FOR_TEST_ACTIVITY +
+                FuzzUtils::FUZZING_FEE *
+                    FuzzUtils::DEFAULT_NUM_TRANSACTIONS_TO_RESERVE_FEES_FOR;
+            assert(FuzzUtils::INITIAL_ASSET_DISTRIBUTION >= remainingBalance);
+            auto reduceNativeBalanceOp = txtest::payment(
```
jonjove, line 797: I don't see how this is going to produce the right balance for account 0, because it has been paying fees for all of these set-up transactions. You probably shouldn't be using `INITIAL_ASSET_DISTRIBUTION` here.

Conceptually, why are we using one of the generated source accounts for the set-up transactions instead of using the root account, which we aren't interested in?

jonjove: `getAvailableBalance` will probably help you get this section right.

src/test/FuzzerImpl.cpp
```diff
@@ -688,17 +763,87 @@ TransactionFuzzer::initialize()
...
+                    // Reduce "account"'s balance of asset "j" by paying the
+                    // issuer.
+                    LedgerKey trustLineKey;
+                    trustLineKey.type(TRUSTLINE);
+                    trustLineKey.trustLine().accountID = account;
+                    trustLineKey.trustLine().asset = asset;
+                    LedgerEntry trustLineEntry =
+                        ltx.load(trustLineKey).current();
+                    auto extVersion = trustLineEntry.data.trustLine().ext.v();
+                    auto sellingLiabilities =
```
jonjove, line 821: You're working too hard. Use `getAvailableBalance`.

src/test/FuzzerImpl.cpp
```diff
@@ -503,17 +517,82 @@ isBadOverlayFuzzerInput(StellarMessage const& m)
...
-    if (txFramePtr->getResultCode() != txSUCCESS)
+    if (throwIfTxFails)
     {
-        throw std::runtime_error(fmt::format(
-            FMT_STRING("Error {} while applying operations while fuzzing"),
-            txFramePtr->getResultCode()));
+        auto const opFrames = txFramePtr->getOperations();
+        if (txFramePtr->getResultCode() != txSUCCESS)
+        {
+            auto const msg = fmt::format(
```
jonjove, line 532: Another example of working too hard. Just log the entire transaction result using `xdr_to_string`. This isn't user-facing, it's developer facing.

jonjove (changes requested):
This is going to be pretty clear from the comments, but I really want you to focus on producing simple solutions. If you reuse the appropriate existing functionality, it will guide you toward solutions that behave correctly.

src/test/FuzzerImpl.cpp
rokopt: > If you see yourself writing a lot of code that you've never seen anywhere else in stellar-core, you should probably be suspicious that there is an easier way.

Well put.

I've cleaned up the logging in `attemptToApplyOps()`.

src/test/FuzzerImpl.cpp
rokopt: Fixed as in previous comment.

src/test/FuzzerImpl.cpp
rokopt: (This has been changed in accordance with Jon's suggestion to start out with large balances and reduce them after the rest of setup.)

src/test/FuzzerImpl.cpp
rokopt: Now using `loadAccount()`.

src/test/FuzzerImpl.cpp
rokopt: Also used now.

src/test/FuzzerImpl.cpp
rokopt: Done (along with using `loadAccount()`/`loadTrustLine()`).

src/test/FuzzerImpl.cpp
rokopt: Done (along with using `loadAccount()`).

src/test/FuzzerImpl.cpp
rokopt: Done.

src/test/FuzzerImpl.cpp
rokopt: Done.

src/test/FuzzerImpl.cpp
rokopt: > You probably shouldn't be using INITIAL_ASSET_DISTRIBUTION here.

> getAvailableBalance will probably help you get this section right.

Indeed, using `getAvailableBalance()` got rid of the use of `INITIAL_ASSET_DISTRIBUTION`.

rokopt: > Conceptually, why are we using one of the generated source accounts for the set-up transactions instead of using the root account, which we aren't interested in?

I don't know; I've wondered myself.  It's been that way since day one.  I'd be more than happy to change it, but I feel like that would be confusing to do together with the other changes in this PR; I'd rather make a separate PR for it once this one's in.  What do you think?

jonjove: Better to do it in a separate PR.

rokopt: > I don't see how this is going to produce the right balance for account 0, because it has been paying fees for all of these set-up transactions.

In fact, as long as we're still using account 0 as the source, it's meaningless to reduce account 0's balance at all -- it's just sending a payment to itself.  So in this PR, I'll just start the balance-reducing loop with account `1`.

In the PR that changes the source account to `root`, we can stop special-casing account `0`.

Does that make sense?

rokopt:
I believe that I've answered all pending review comments, and I've tried to create a coherent commit history for further review.

src/test/FuzzerImpl.cpp
```diff
@@ -729,17 +730,76 @@ TransactionFuzzer::initialize()
...
+                                    availableBalance - targetAvailableBalance);
+                reduceNativeBalanceOp.sourceAccount.activate() =
+                    toMuxedAccount(account);
+                ops.emplace_back(reduceNativeBalanceOp);
+            }
+
+            for (int j = 1; j <= FuzzUtils::NUMBER_OF_ASSETS_TO_ISSUE; ++j)
+            {
+                auto const asset = FuzzUtils::makeAsset(j);
+                if ((i != 0) && (i != j))
```
jonjove, line 772: `i == 0` is impossible here. Also you're doing a signed/unsigned comparison `i != j` here which is generally unsafe.

src/test/FuzzerImpl.cpp
```diff
@@ -3,7 +3,9 @@
 // of this distribution or at http://www.apache.org/licenses/LICENSE-2.0
 
 #include "test/FuzzerImpl.h"
+#include "ledger/InternalLedgerEntry.h"
```
jonjove, line 6: Nitpick: I think this is unused.

src/test/FuzzerImpl.cpp
```diff
@@ -729,17 +730,76 @@ TransactionFuzzer::initialize()
...
+                getAvailableBalance(ltx.loadHeader(), ae);
+            auto const targetAvailableBalance =
+                FuzzUtils::DEFAULT_ASSET_AVAILABLE_FOR_TEST_ACTIVITY +
+                FuzzUtils::FUZZING_FEE *
+                    FuzzUtils::DEFAULT_NUM_TRANSACTIONS_TO_RESERVE_FEES_FOR;
+
+            if (availableBalance > targetAvailableBalance)
+            {
+                auto reduceNativeBalanceOp =
+                    txtest::payment(mSourceAccountID, txtest::makeNativeAsset(),
```
jonjove, line 762: Nitpick: you can use `payment(PublicKey const& to, int64_t amount)` instead.

jonjove (commented):
This is getting close now, but I had a few more comments that are mostly about style.

src/test/FuzzerImpl.cpp
```diff
@@ -501,19 +506,59 @@ isBadOverlayFuzzerInput(StellarMessage const& m)
...
-    if (txFramePtr->getResultCode() != txSUCCESS)
-    {
-        throw std::runtime_error(fmt::format(
-            FMT_STRING("Error {} while applying operations while fuzzing"),
-            txFramePtr->getResultCode()));
+            for (auto const& opFrame : txFramePtr->getOperations())
+            {
+                auto const& op = opFrame->getOperation();
+                auto const& tr = opFrame->getResult().tr();
+                auto const& opType = op.body.type();
```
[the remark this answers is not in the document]
rokopt: Fixed.

src/test/FuzzerImpl.cpp
```diff
@@ -501,19 +506,59 @@ isBadOverlayFuzzerInput(StellarMessage const& m)
     return m.type() == AUTH || m.type() == ERROR_MSG || m.type() == HELLO;
 }
 
+// Empties "ops" as operations are applied.  If `throwIfTxFails` is `false`,
+// ignores failures and might apply further transactions after failures.
 static void
 attemptToApplyOps(LedgerTxn& ltx, PublicKey const& sourceAccount,
-                  xdr::xvector<Operation> const& ops, Application& app)
+                  xdr::xvector<Operation>& ops, Application& app,
```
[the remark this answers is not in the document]
rokopt: Done.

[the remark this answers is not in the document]
rokopt: (I passed them all the way through to `createFuzzTransactionFrame()`, in fact.)

src/test/FuzzerImpl.cpp
```diff
@@ -501,19 +506,59 @@ isBadOverlayFuzzerInput(StellarMessage const& m)
...
-    txFramePtr->attemptApplication(app, ltx);
+    while (!ops.empty())
+    {
+        auto endOpsInThisTx = ops.size() <= MAX_OPS_PER_TX
+                                  ? ops.end()
+                                  : ops.begin() + MAX_OPS_PER_TX;
+        auto txFramePtr = createFuzzTransactionFrame(
+            sourceAccount, xdr::xvector<Operation>(ops.begin(), endOpsInThisTx),
+            app.getNetworkID());
+        ops.erase(ops.begin(), endOpsInThisTx);
```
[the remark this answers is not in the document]
rokopt: Agreed.  It now uses iterators, and no erasing.  (The clients didn't have any particular need for the vector to have been emptied.)

src/test/FuzzerImpl.cpp
```diff
@@ -780,33 +780,27 @@ TransactionFuzzer::inject(std::string const& filename)
...
-    }
-
     resetTxInternalState(*mApp);
+    LOG_TRACE(DEFAULT_LOG, "Fuzz ops ({}): {}", ops.size(), xdr_to_string(ops));
     LedgerTxn ltx(mApp->getLedgerTxnRoot());
-
-    // attempt to apply transaction
-    txFramePtr->attemptApplication(*mApp, ltx);
+    attemptToApplyOps(ltx, mSourceAccountID, ops, *mApp, false);
+    ltx.rollback();
```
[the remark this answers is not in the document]
rokopt: Removed.

src/test/FuzzerImpl.cpp
```diff
@@ -729,17 +730,76 @@ TransactionFuzzer::initialize()
...
+        ltx.commit();
+    }
 
-        if (tx->getResultCode() != txSUCCESS)
+    {
+        LedgerTxn ltx(ltxOuter);
+
+        xdr::xvector<Operation> ops;
+
+        // Reduce account balances so that fuzzing has a better change
```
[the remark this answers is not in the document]
rokopt: Fixed.

src/test/FuzzerImpl.cpp
```diff
@@ -729,17 +730,76 @@ TransactionFuzzer::initialize()
...
+        // of exercising limits.
+        for (uint8_t i = 1; i < FuzzUtils::NUMBER_OF_PREGENERATED_ACCOUNTS; ++i)
         {
-            throw std::runtime_error("Error while initializing ledger state");
+            PublicKey account;
+            FuzzUtils::setShortKey(account, i);
+
+            // Reduce "account"'s native balance by paying the root.
+            auto ae = stellar::loadAccount(ltx, account);
```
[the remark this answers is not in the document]
rokopt: I agree that there's no reason not to pay the root, so I've changed the code rather than the comment.

src/test/FuzzerImpl.cpp
rokopt: Fixed.

src/test/FuzzerImpl.cpp
rokopt: Removed.

src/test/FuzzerImpl.cpp
rokopt: Done.

jonjove:
Changes look good. Please squash and rebase, then I'll merge it.

rokopt:
> Changes look good. Please squash and rebase, then I'll merge it.

Thanks!  Done.

jonjove:
r+ 26e72fb

PR #2859, opened 2021-01-06
StellarValue cleanup

### LINKED ISSUE #2646: sweep code base for STELLAR_VALUE_BASIC
Post protocol 14, we should remove all handling of STELLAR_VALUE_BASIC values and update all tests to use signed values

The only place where `STELLAR_VALUE_BASIC` will still be referenced will be in the XDR.

marta-lokhova:
Resolves #2646

### DIFF
--- a/src/herder/test/HerderTests.cpp
+++ b/src/herder/test/HerderTests.cpp
@@ -1075,8 +1075,7 @@ TEST_CASE("surge pricing", "[herder][txset]")
 }
 
 static void
-testSCPDriver(uint32 protocolVersion, uint32_t maxTxSize, size_t expectedOps,
-              bool const expectTxSetCloseTimeAffinity)
+testSCPDriver(uint32 protocolVersion, uint32_t maxTxSize, size_t expectedOps)
 {
     using SVUpgrades = decltype(StellarValue::upgrades);
 
@@ -1101,22 +1100,17 @@ testSCPDriver(uint32 protocolVersion, uint32_t maxTxSize, size_t expectedOps,
 
     using TxPair = std::pair<Value, TxSetFramePtr>;
     auto makeTxUpgradePair = [&](HerderImpl& herder, TxSetFramePtr txSet,
-                                 uint64_t closeTime, SVUpgrades const& upgrades,
-                                 bool sig) {
+                                 uint64_t closeTime,
+                                 SVUpgrades const& upgrades) {
         txSet->sortForHash();
-        auto sv = StellarValue(txSet->getContentsHash(), closeTime, upgrades,
-                               STELLAR_VALUE_BASIC);
-        if (sig)
-        {
-            herder.signStellarValue(root.getSecretKey(), sv);
-        }
+        StellarValue sv = herder.makeStellarValue(
+            txSet->getContentsHash(), closeTime, upgrades, root.getSecretKey());
         auto v = xdr::xdr_to_opaque(sv);
         return TxPair{v, txSet};
     };
     auto makeTxPair = [&](HerderImpl& herder, TxSetFramePtr txSet,
-                          uint64_t closeTime, bool sig) {
-        return makeTxUpgradePair(herder, txSet, closeTime, emptyUpgradeSteps,
-                                 sig);
+                          uint64_t closeTime) {
+        return makeTxUpgradePair(herder, txSet, closeTime, emptyUpgradeSteps);
     };
     auto makeEnvelope = [&s](HerderImpl& herder, TxPair const& p, Hash qSetHash,
                              uint64_t slotIndex, bool nomination) {
@@ -1217,13 +1211,12 @@ testSCPDriver(uint32 protocolVersion, uint32_t maxTxSize, size_t expectedOps,
                 Value upgrade(xdr::xdr_to_opaque(ledgerUpgrade));
                 SVUpgrades upgrades;
                 upgrades.emplace_back(upgrade.begin(), upgrade.end());
-                addToCandidates(makeTxUpgradePair(herder, txSet, spec.closeTime,
-                                                  upgrades, true));
+                addToCandidates(
+                    makeTxUpgradePair(herder, txSet, spec.closeTime, upgrades));
             }
             else
             {
-                addToCandidates(
-                    makeTxPair(herder, txSet, spec.closeTime, true));
+                addToCandidates(makeTxPair(herder, txSet, spec.closeTime));
             }
 
             // Compute the expected transaction set, close time, and upgrade
@@ -1233,10 +1226,7 @@ testSCPDriver(uint32 protocolVersion, uint32_t maxTxSize, size_t expectedOps,
                 std::max_element(txSetSizes.begin(), txSetSizes.end()));
             REQUIRE(txSetSizes.size() == closeTimes.size());
             auto const expectedHash = txSetHashes[bestTxSetIndex];
-            auto const expectedCloseTime =
-                expectTxSetCloseTimeAffinity
-                    ? closeTimes[bestTxSetIndex]
-                    : *std::max_element(closeTimes.begin(), closeTimes.end());
+            auto const expectedCloseTime = closeTimes[bestTxSetIndex];
             SVUpgrades expectedUpgradeVector;
             if (!baseFees.empty())
             {
@@ -1259,8 +1249,7 @@ testSCPDriver(uint32 protocolVersion, uint32_t maxTxSize, size_t expectedOps,
 
             // Compare the returned StellarValue's contents with the
             // expected ones that we computed above.
-            REQUIRE(sv.ext.v() ==
-                    herder.getHerderSCPDriver().compositeValueType());
+            REQUIRE(sv.ext.v() == STELLAR_VALUE_SIGNED);
             REQUIRE(sv.txSetHash == expectedHash);
             REQUIRE(sv.closeTime == expectedCloseTime);
             REQUIRE(sv.upgrades == expectedUpgradeVector);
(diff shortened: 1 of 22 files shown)


marta-lokhova:
build failure because of the randomized tx application in "tx envelope" (using signed value changed ledger hash which affects application order). I might revert back to just using basic value for this test.

marta-lokhova:
couple of updates here: updated protocol-based close time behavior in addition to BASIC value cleanup, plus added a small feature to tx tests to allow executing transactions in the exact order they were added (note that I removed `skipValid` flag, as it was used in 1 place, and removing it didn't make any difference, test still passed). CI is passing now. @MonsieurNicolas

src/test/TxTests.cpp
```diff
@@ -395,26 +395,49 @@ validateTxResults(TransactionFramePtr const& tx, Application& app,
...
 TxSetResultMeta
 closeLedgerOn(Application& app, uint32 ledgerSeq, time_t closeTime,
-              std::vector<TransactionFrameBasePtr> const& txs, bool skipValid)
+              std::vector<TransactionFrameBasePtr> const& txs, bool strictOrder)
 {
-    auto txSet = std::make_shared<TxSetFrame>(
-        app.getLedgerManager().getLastClosedLedgerHeader().hash);
+    std::shared_ptr<TxSetFrame> txSet;
+    auto lclHash = app.getLedgerManager().getLastClosedLedgerHeader().hash;
+    if (strictOrder)
```
jonjove, line 425: I'm not sure if we want to do this. Transaction order is defined at the protocol level, we really want to know if this is changing.

MonsieurNicolas: Some tests want to test a specific sequence of transactions. I don't think they're testing that the apply order is the same one than in previous implementations - that's what historical replays are for?

There are a few other ways to do it with different tradeoffs, what do you suggest we do in that case?

Alternative to this approach are:
* not close ledgers in that case - just use some other test wrapper (so not testing apply order, so seems the same)
* control the apply order from the test. This can be done by hard coding some memo in one of the transactions. We'd need to compute the value of that memo somehow.

jonjove: In general we've done this approach "control the apply order from the test. This can be done by hard coding some memo in one of the transactions. We'd need to compute the value of that memo somehow." It's not particularly difficult if the number of transactions is small.

These tests have alerted me in the past to inadvertent changes to transaction order, although I agree historical replay is the ultimate source of truth on this.

MonsieurNicolas: mmm, maybe what we should do then is have `REQUIRE` statements in `closeLedgerOn` that enforce that the order passed in is the same than the apply order.

As far as pattern to control the apply order, I thought of a third possibility that might be the least invasive: we can change it by changing the hash of the previous ledger, so changing anything about the previous ledger should do it. It can be close time etc - or even modify something in the genesis ledger like `TESTING_UPGRADE_MAX_TX_SET_SIZE`.

marta-lokhova: I'm open to discussion on this, but... the "modify txs to ensure correct order" approach seems backwards to me. If we want to catch ordering bugs, why not have a separate test that focuses on that? The test that I modified expects a specific order. If it fails due to the wrong order, it's not immediately obvious, because we're checking result codes, and nothing explicitly order-related (whereas a designated test could provide a clear error message).
Besides the fact that it's extra time spent guessing and maintaining the right hash (that works for all protocol versions), I also didn't like the fact that it makes the test more confusing, since we have to mess with seemingly unrelated stuff (memo, prev hash, or whatever we pick).

MonsieurNicolas: I think you are right @marta-lokhova : we should just add a canary test with some non trivial (multiple transactions) transaction set, and have a hardcoded check that verifies that the hash of that tx set is what we expect.
That way we avoid mixing test purpose.

marta-lokhova: This was updated with an extra apply order test.

src/herder/test/HerderTests.cpp
```diff
@@ -1104,11 +1103,16 @@ testSCPDriver(uint32 protocolVersion, uint32_t maxTxSize, size_t expectedOps,
...
+        StellarValue sv;
         if (sig)
         {
-            herder.signStellarValue(root.getSecretKey(), sv);
+            sv = herder.makeStellarValue(txSet->getContentsHash(), closeTime,
+                                         upgrades, root.getSecretKey());
+        }
+        else
+        {
+            sv = StellarValue(txSet->getContentsHash(), closeTime, upgrades,
```
MonsieurNicolas, line 1114: actually we should remove this: the only place where we want to check that `BASIC` values are invalid is in the `SECTION("invalid")` where we should add a new mutation ("basic value")

src/herder/test/HerderTests.cpp
```diff
@@ -1332,7 +1326,7 @@ testSCPDriver(uint32 protocolVersion, uint32_t maxTxSize, size_t expectedOps,
             REQUIRE(scp.validateValue(seq, nomV.first, true) ==
                     SCPDriver::kInvalidValue);
 
-            auto balV = makeTxPair(herder, txSet0, ct, !signedBallots);
+            auto balV = makeTxPair(herder, txSet0, ct, false);
```
MonsieurNicolas, line 1329: `nomV` and `balV` are not needed anymore:
* they are testing the same thing
* it's simpler to just add a new case below for `BASIC`

src/herder/test/HerderTests.cpp
```diff
@@ -1412,9 +1406,7 @@ testSCPDriver(uint32 protocolVersion, uint32_t maxTxSize, size_t expectedOps,
             // Confirm that trimInvalid() as used by
             // HerderImpl::triggerNextLedger() trims the transaction if and only
             // if we expect it to be invalid.
-            auto closeTimeOffset = expectTxSetCloseTimeAffinity
-                                       ? (nextCloseTime - lclCloseTime)
-                                       : 0;
+            auto closeTimeOffset = nextCloseTime - lclCloseTime;
```
MonsieurNicolas, line 1409: wrong commit (this belongs to the next commit that removes support for not having ct affinity)

src/herder/test/HerderTests.cpp
```diff
@@ -1429,12 +1421,12 @@ testSCPDriver(uint32 protocolVersion, uint32_t maxTxSize, size_t expectedOps,
...
         SECTION("expired (invalid maxTime) with txSet/closeTime affinity")
         {
-            testTxBounds(0, t1, t2, !expectTxSetCloseTimeAffinity);
+            testTxBounds(0, t1, t2, false);
         }
 
         SECTION("premature (invalid minTime) without txSet/closeTime affinity")
         {
-            testTxBounds(t1, 0, t1, expectTxSetCloseTimeAffinity);
+            testTxBounds(t1, 0, t1, true);
```
MonsieurNicolas, line 1429: (next commit) section names here are now confusing ("all protocols", "with/ without affinity")

src/herder/test/HerderTests.cpp
```diff
@@ -1607,13 +1599,9 @@ testSCPDriver(uint32 protocolVersion, uint32_t maxTxSize, size_t expectedOps,
...
 TEST_CASE("SCP Driver", "[herder][acceptance]")
 {
-    SECTION("protocol 13")
-    {
-        testSCPDriver(13, 1000, 15, false);
-    }
     SECTION("protocol current")
     {
-        testSCPDriver(Config::CURRENT_LEDGER_PROTOCOL_VERSION, 1000, 15, true);
+        testSCPDriver(Config::CURRENT_LEDGER_PROTOCOL_VERSION, 1000, 15);
```
MonsieurNicolas, line 1604: belongs to the next commit

src/history/test/HistoryTestsUtils.cpp
```diff
@@ -452,6 +452,8 @@ void
 CatchupSimulation::generateRandomLedger(uint32_t version)
 {
     auto& lm = mApp.getLedgerManager();
+    auto& herder = static_cast<HerderImpl&>(mApp.getHerder());
```
MonsieurNicolas, line 455: why not just add `makeStellarValue` to `Herder` (the public contract)? I don't see a downside and it avoids having to expose the "impl" version everywhere

MonsieurNicolas (changes requested):
looks good overall.
A few minor things

marta-lokhova:
@MonsieurNicolas I believe I addressed all of your comments, the PR is rebased now as well (I ended up putting close time and BASIC value changes in one commit as it was simpler)

src/herder/test/HerderTests.cpp
```diff
@@ -1318,49 +1301,54 @@ testSCPDriver(uint32 protocolVersion, uint32_t maxTxSize, size_t expectedOps,
...
-            auto balV = makeTxPair(herder, txSet0, ct, !signedBallots);
-            REQUIRE(scp.validateValue(seq, balV.first, false) ==
-                    SCPDriver::kInvalidValue);
-
-            auto p = makeTxPair(herder, txSet0, ct, true);
-            StellarValue sv;
-            xdr::xdr_from_opaque(p.first, sv);
 
-            auto checkInvalid = [&](StellarValue const& sv) {
+            auto checkInvalid = [&](StellarValue const& sv,
```
MonsieurNicolas, line 1315: Actually why not wrap this whole thing and test both variations (I don't see a reason to only check for bad signatures etc only for the ballot protocol case)

```
            SECTION("Nomination")
            {
                TestInvalid(true);
            }
            SECTION("Ballot")
            {
                TestInvalid(false);
            }
```

MonsieurNicolas (changes requested):
looks good! Only one suggestion in a test case and we'll merge it

src/herder/test/HerderTests.cpp
marta-lokhova: done!

MonsieurNicolas:
r+ fbdad24d6ef7fc7104566a2486a881ec59dd4290

marta-lokhova:
updated the PR @MonsieurNicolas, there was some non-determinism in the ordering test because tests use random seed when generating NODE_SEED. So I hardcoded the NODE_SEED value (and moved that particular test into a separate test case)
https://github.com/stellar/stellar-core/pull/2859/commits/fc1fe2ac4136aa6d97a3ff19234f787ca0c7cb42#diff-1bee61d99b026d9e89a65a5eb493bc03388c3013f7541140df8612ee61352f7fR48

MonsieurNicolas:
r+ 8b2d28302d4a2261bf7f256a06438c38c06b5995

PR #2862, opened 2021-01-06
RunCommandWork: avoid modifying work's state after completion

marta-lokhova:
The issue came up during the review of process shutdown change: https://github.com/stellar/stellar-core/pull/2801#discussion_r524841404

Note that `basic_waitable_timer::async_wait` fires the callback regardless of whether the timer has expired or it has been cancelled, so we have to handle this case within the callback itself.


MonsieurNicolas:
r+ 443a88d860648d42a75b6650dd384ff0c361234b

PR #2865, opened 2021-01-07
Don't generate operation meta if an operation has already failed

jonjove:
# Description
This is a slight optimization to the way we currently generate meta. I think the new version is also easier to understand. This originally caught my attention because I was wondering why we were generating meta differently prior to protocol version 14.

# Checklist
- [x] Compiles
- [x] Ran all tests


MonsieurNicolas:
r+ 0b34c8c58d9638da0bdefa748cc5df0f4d15d993

PR #2870, opened 2021-01-08
Only truncate log on logrotate, not startup. Fixes #2868.

### LINKED ISSUE #2868: logging truncates log files
When we reopen log files (on startup or logrotate), we truncate the file.

This is bad because if core crashes or gets restarted, it basically deletes its log file.

We should be appending instead.

It looks to be a 1 line change as we're passing `/*truncate=*/true` when we create the file sink.

graydon:
Resolves #2868 -- only truncate log file when we're rotating (as easylogging did) but not when starting up.

### DIFF
--- a/src/util/Logging.cpp
+++ b/src/util/Logging.cpp
@@ -80,7 +80,7 @@ convert_loglevel(LogLevel level)
 #endif
 
 void
-Logging::init()
+Logging::init(bool truncate)
 {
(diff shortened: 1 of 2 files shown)


MonsieurNicolas:
r+ 4f1e2d30a24467ec33ff5d27c5885b1af848a667

PR #2872, opened 2021-01-09
Truncate log files manually rather than via spdlog. Fixes #2867.

### LINKED ISSUE #2867: logrotate trashes files
It looks like logrotate (the tool, not the endpoint) causes unexpected behavior with spdlog.

When a file gets rotated, it looks like the file is partially reset and appended to:
The log file after log rotation contains a lot of `0` bytes at the beginning of the file (exactly "log file size" before rotation), followed by the actual log content.

I am not sure what is going on - it looks like spdlog is fseeking to some cached offset, causing the 0 bytes to be added?

Note that this is using the `copytruncate` option of logrotate - that is supposed to be the "universal" log rotation.

### System information

Ubuntu

log rotate config file : `/etc/logrotate.d/stellar-core`
```
# Keep 7 days worth of logs
/var/log/stellar/stellar-core.log {
  hourly
  missingok
  rotate 72
  compress
  notifempty
  copytruncate
}
```

graydon:
This changes the behaviour of our logging system to truncate log files manually when rotating (or otherwise asking to truncate), rather than passing the truncate option through to spdlog. The explanation is contained in the comments. Change itself is fairly trivial. I'll also open an issue upstream on spdlog to discuss there, they might consider this a bug on their end.

Resolves #2867

### DIFF
--- a/src/util/Logging.cpp
+++ b/src/util/Logging.cpp
@@ -9,6 +9,7 @@
 #include "util/Timer.h"
 #include <chrono>
 #include <fmt/chrono.h>
+#include <fstream>
 #include <spdlog/sinks/basic_file_sink.h>
 #include <spdlog/sinks/stdout_color_sinks.h>
 #include <spdlog/sinks/stdout_sinks.h>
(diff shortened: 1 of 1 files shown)


MonsieurNicolas:
r+ 658ca575226365ae06aa3018ffc4b7bc312bd2b6

Issue #2852, opened 2020-12-19
consider replacing `assert` in our codebase


MonsieurNicolas:
We have plenty of calls to `assert` that should really be `releaseAssert`.

I noticed that we have them in archive/bucket related code.

seekaddo:
@MonsieurNicolas  Hi Happy New Year.  😄 

I will like to work on this issue if nobody is currently working on it.

MonsieurNicolas:
yeah go for it @seekaddo - before opening a PR check the [contributing guide](https://github.com/stellar/stellar-core/blob/master/CONTRIBUTING.md) in particular running tests, format and filling the license agreement

seekaddo:
@MonsieurNicolas Thanks for the response and pointing me to the first steps.
1. I have already signed the "Contributor License Agreement."

2. I have setup the project and all testcases for the package bucket  and others are working after the build.

3. Concerning the issue `Improve HAS validation #2843` are the cleanup already done or planned or Can I just add them to my upcoming PR.

Package to work on:
src/bucket --> (I hope that is what you mean with comments above `archive/bucket` )
src/catchup
src/scp
src/herder
src/histroy
src/invariant
src/ledger
src/overlay
etc .....

MonsieurNicolas:
Yeah ignore other PRs: what will happen is that you'll have to rebase anyways when we're ready to merge. So just take notes on which search/replace you had to do. As we're getting close to a release, I think that we should be able to get a bunch of PRs out of the way before this one.

seekaddo:
Ok that sounds like a good plan. I will apply the changes to the remaining packages.

PR #2876, opened 2021-01-13
Fix the rendering issues of some XDR fields

### LINKED ISSUE #2869: some xdr fields are not rendered properly in JSON
There are a few issues:

 * there is an extra `value0` field at the root. We should do a code sweep, I think this is not the first time we're seeing this.
 * `assetCode` is displayed as hex. I think we should try to display as a string, and only degrade back to hex if the string is not a valid asset code.

```
[Tx TRACE] Operation: {
    "value0": {
        "sourceAccount": "...",
        "body": {
            "type": "PATH_PAYMENT_STRICT_RECEIVE",
            "pathPaymentStrictReceiveOp": {
                "sendAsset": "XLM",
                "destAsset": "XLM",
                "path": [
                    {
                        "type": "ASSET_TYPE_CREDIT_ALPHANUM12",
                        "alphaNum12": {
                            "assetCode": "43454e545553000000000000",
                        }
                    },
                    {
                        "type": "ASSET_TYPE_CREDIT_ALPHANUM4",
                        "alphaNum4": {
                            "assetCode": "55534400",
                        }
                    }
                ]
            }
        }
    }
}
```

hidenori-shinohara:
# Description

Resolves #2869

- The "value0" issue. This seems like a feature rather than a bug (for instance, see this [cereal issue](https://github.com/USCiLab/cereal/issues/303) ) Long story short, it's there to support multi-entry outputs such as  
```
{          
    "value0": {
        "x": 1,
        "y": 2,
        "z": 3
    },    
    "value1": {           
        "x": 4,                
        "y": 5,
        "z": 6
    },
    "value2": {
        "x": 7,
        "y": 8,
        "z": 9
    }
}
```
- `xdr_to_string` has an optional field name argument. When that's not passed, Cereal uses the word "value0". This PR will make the argument mandatory, so we will print a descriptive name given by a caller, instead of "value0". We already use this in multiple places, and this PR updates the usage of `xdr_to_string` when the field name argument is not passed.
- I used the following XDR to make sure that the asset list is rendered correctly. With this PR, the XDR will be rendered as following:

```
...                             "destAsset": "XLM",
                                "destAmount": 2000000000,
                                "path": [
                                    "CENTUS",
                                    "USD"
                                ]
...
```
instead of
```
...
                                "path": [
                                    {
                                        "type": "ASSET_TYPE_CREDIT_ALPHANUM12",
                                        "alphaNum12": {
                                            "assetCode": "43454e545553000000000000",
                                            "issuer": "GCD5VCGHNLC46D5O2SAZCFFP3ZF7JRBQZGSFINXWNY5HCEOE42PH3XH5"
                                        }
                                    },
                                    {
                                        "type": "ASSET_TYPE_CREDIT_ALPHANUM4",
                                        "alphaNum4": {
                                            "assetCode": "55534400",
                                            "issuer": "GDJAG52BJUFI6RBKFTNUEJBDZODJPZBLIHN3SXMR7QIXNRAOJK3XL6P3"
                                        }
                                    }
                                ]
...
```

`AAAAAgAAAACJCjEa69Jmm5xInlCSLmzEAf8u1q/AeXUbhoZQFHOsoQAAAGQAAAAcvpkaFAAAAAEAAAAAAAAAAAAAAAAAAAAAAAAAAQAAABx0aGlzIGlzIGEgbWVtbyBmb3IgZGVidWdnaW5nAAAAAQAAAAAAAAACAAAAAAAAAAA7msoAAAAAANhWYalldqwqmK2G0p/IIaS+64sMI9nfPeBe5RlS2NDCAAAAAAAAAAB3NZQAAAAAAgAAAAJDRU5UVVMAAAAAAAAAAAAAh9qIx2rFzw+u1IGRFK/eS/TEMMmkVDb2bjpxEcTmnn0AAAABVVNEAAAAAADSA3dBTQqPRCos20IkI8uGl+QrQdu5XZH8EXbEDkq3dQAAAAAAAAAA`

# Checklist
- [x] Compiles
- [x] Ran all tests

### DIFF
--- a/src/catchup/simulation/TxSimApplyTransactionsWork.cpp
+++ b/src/catchup/simulation/TxSimApplyTransactionsWork.cpp
@@ -66,9 +66,9 @@ checkOperationResults(xdr::xvector<OperationResult> const& expected,
     {
         if (expected[i].code() != actual[i].code())
         {
-            CLOG_ERROR(History, "Expected operation result {} but got {}",
-                       xdr_to_string(expected[i].code()),
-                       xdr_to_string(actual[i].code()));
+            CLOG_ERROR(History, "Expected {} but got {}",
+                       xdr_to_string(expected[i].code(), "OperationResultCode"),
+                       xdr_to_string(actual[i].code(), "OperationResultCode"));
             continue;
         }
 
@@ -154,10 +154,10 @@ checkOperationResults(xdr::xvector<OperationResult> const& expected,
 
         if (!match)
         {
-            CLOG_ERROR(History, "Expected operation result: {}",
-                       xdr_to_string(expectedOpRes));
-            CLOG_ERROR(History, "Actual operation result: {}",
-                       xdr_to_string(actualOpRes));
+            CLOG_ERROR(History, "Expected {}",
+                       xdr_to_string(expectedOpRes, "OperationResult"));
+            CLOG_ERROR(History, "Actual {}",
+                       xdr_to_string(actualOpRes, "OperationResult"));
         }
     }
 }
@@ -179,9 +179,9 @@ checkResults(Application& app, uint32_t ledger,
         if (dbRes.code() != archiveRes.code())
         {
             CLOG_ERROR(
-                History,
-                "Expected result code {} does not agree with {} for tx {}",
-                xdr_to_string(archiveRes.code()), xdr_to_string(dbRes.code()),
+                History, "Expected {} does not agree with {} for tx {}",
+                xdr_to_string(archiveRes.code(), "TransactionResultCode"),
+                xdr_to_string(dbRes.code(), "TransactionResultCode"),
                 binToHex(results[i].transactionHash));
         }
         else if (dbRes.code() == txFEE_BUMP_INNER_FAILED ||
@@ -193,11 +193,13 @@ checkResults(Application& app, uint32_t ledger,
             {
                 CLOG_ERROR(
                     History,
-                    "Expected result code {} does not agree with {} for "
+                    "Expected {} does not agree with {} for "
                     "fee-bump inner tx {}",
                     xdr_to_string(
-                        archiveRes.innerResultPair().result.result.code()),
-                    xdr_to_string(dbRes.innerResultPair().result.result.code()),
+                        archiveRes.innerResultPair().result.result.code(),
+                        "TransactionResultCode"),
+                    xdr_to_string(dbRes.innerResultPair().result.result.code(),
+                                  "TransactionResultCode"),
                     binToHex(archiveRes.innerResultPair().transactionHash));
             }
             else if (dbRes.innerResultPair().result.result.code() == txFAILED ||
--- a/src/invariant/LiabilitiesMatchOffers.cpp
+++ b/src/invariant/LiabilitiesMatchOffers.cpp
@@ -100,7 +100,7 @@ checkAuthorized(LedgerEntry const* current, LedgerEntry const* previous)
                 {
                     return fmt::format(
                         "Liabilities increased on unauthorized trust line {}",
-                        xdr_to_string(trust));
+                        xdr_to_string(trust, "TrustLineEntry"));
                 }
             }
             else
@@ -110,7 +110,7 @@ checkAuthorized(LedgerEntry const* current, LedgerEntry const* previous)
                 {
                     return fmt::format(
                         "Unauthorized trust line has liabilities {}",
-                        xdr_to_string(trust));
+                        xdr_to_string(trust, "TrustLineEntry"));
                 }
             }
         }
@@ -256,8 +256,8 @@ checkBalanceAndLimit(LedgerHeader const& header, LedgerEntry const* current,
                 (INT64_MAX - account.balance < liabilities.buying))
             {
                 return fmt::format(
-                    "Balance not compatible with liabilities for account {}",
-                    xdr_to_string(account));
+                    "Balance not compatible with liabilities for {}",
+                    xdr_to_string(account, "AccountEntry"));
             }
         }
     }
@@ -273,9 +273,8 @@ checkBalanceAndLimit(LedgerHeader const& header, LedgerEntry const* current,
         if ((trust.balance < liabilities.selling) ||
             (trust.limit - trust.balance < liabilities.buying))
         {
-            return fmt::format(
-                "Balance not compatible with liabilities for trustline {}",
-                xdr_to_string(trust));
+            return fmt::format("Balance not compatible with liabilities for {}",
+                               xdr_to_string(trust, "TrustLineEntry"));
         }
     }
     return {};
@@ -346,20 +345,20 @@ LiabilitiesMatchOffers::checkOnOperationApply(Operation const& operation,
                     return fmt::format(
                         "Change in buying liabilities differed from "
                         "change in total buying liabilities of "
-                        "offers by {} for account {} in asset {}",
+                        "offers by {} for {} in {}",
                         assetLiabilities.second.buying,
-                        xdr_to_string(accLiabilities.first),
-                        xdr_to_string(assetLiabilities.first));
+                        xdr_to_string(accLiabilities.first, "account"),
+                        xdr_to_string(assetLiabilities.first, "asset"));
                 }
                 else if (assetLiabilities.second.selling != 0)
                 {
                     return fmt::format(
                         "Change in selling liabilities differed from "
                         "change in total selling liabilities of "
-                        "offers by {} for account {} in asset {}",
+                        "offers by {} for {} in {}",
                         assetLiabilities.second.selling,
-                        xdr_to_string(accLiabilities.first),
-                        xdr_to_string(assetLiabilities.first));
+                        xdr_to_string(accLiabilities.first, "account"),
+                        xdr_to_string(assetLiabilities.first, "asset"));
                 }
             }
         }
--- a/src/test/FuzzerImpl.cpp
+++ b/src/test/FuzzerImpl.cpp
@@ -789,11 +789,11 @@ applySetupOperations(LedgerTxn& ltx, PublicKey const& sourceAccount,
 
         if (txFramePtr->getResultCode() != txSUCCESS)
         {
-            auto const msg =
-                fmt::format(FMT_STRING("Error {} while setting up fuzzing -- "
-                                       "transaction result {}"),
-                            txFramePtr->getResultCode(),
-                            xdr_to_string(txFramePtr->getResult()));
+            auto const msg = fmt::format(
+                FMT_STRING("Error {} while setting up fuzzing -- "
+                           "{}"),
+                txFramePtr->getResultCode(),
+                xdr_to_string(txFramePtr->getResult(), "TransactionResult"));
             LOG_FATAL(DEFAULT_LOG, "{}", msg);
             throw std::runtime_error(msg);
         }
@@ -816,8 +816,9 @@ applySetupOperations(LedgerTxn& ltx, PublicKey const& sourceAccount,
             {
                 auto const msg = fmt::format(
                     FMT_STRING("Manage offer result {} while setting "
-                               "up fuzzing -- operation is {}"),
-                    xdr_to_string(tr), xdr_to_string(op));
+                               "up fuzzing -- {}"),
+                    xdr_to_string(tr, "Operation"),
+                    xdr_to_string(op, "Operation"));
                 LOG_FATAL(DEFAULT_LOG, "{}", msg);
                 throw std::runtime_error(msg);
             }
@@ -1280,7 +1281,9 @@ TransactionFuzzer::inject(std::string const& filename)
     }
 
     resetTxInternalState(*mApp);
-    LOG_TRACE(DEFAULT_LOG, "Fuzz ops ({}): {}", ops.size(), xdr_to_string(ops));
+    LOG_TRACE(DEFAULT_LOG, "{}",
+              xdr_to_string(ops, fmt::format("Fuzz ops ({})", ops.size())));
+
     LedgerTxn ltx(mApp->getLedgerTxnRoot());
     applyFuzzOperations(ltx, mSourceAccountID, ops.begin(), ops.end(), *mApp);
 }
--- a/src/transactions/OperationFrame.cpp
+++ b/src/transactions/OperationFrame.cpp
@@ -122,7 +122,7 @@ OperationFrame::apply(SignatureChecker& signatureChecker,
     bool res;
     if (Logging::logTrace("Tx"))
     {
-        CLOG_TRACE(Tx, "Operation: {}", xdr_to_string(mOperation));
+        CLOG_TRACE(Tx, "{}", xdr_to_string(mOperation, "Operation"));
     }
     res = checkValid(signatureChecker, ltx, true);
     if (res)
@@ -130,7 +130,7 @@ OperationFrame::apply(SignatureChecker& signatureChecker,
         res = doApply(ltx);
         if (Logging::logTrace("Tx"))
         {
-            CLOG_TRACE(Tx, "Operation result: {}", xdr_to_string(mResult));
+            CLOG_TRACE(Tx, "{}", xdr_to_string(mResult, "OperationResult"));
         }
     }
 
--- a/src/util/XDRCereal.cpp
+++ b/src/util/XDRCereal.cpp
@@ -37,8 +37,19 @@ cereal_override(cereal::JSONOutputArchive& ar,
 }
 
 void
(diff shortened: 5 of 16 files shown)


jonjove:
Regarding the second part of your change: generally I would want to see both the asset code and the issuer.

hidenori-shinohara:
Update the way `Assets` is handled, and now it'll be printed as following:

```
                            "pathPaymentStrictReceiveOp": {
                                "sendAsset": {   
                                    "assetCode": "XLM",           
                                    "issuer": "SDF"                                                                                                
                                },                                                                                                                 
                                "sendMax": 1000000000,          
                                "destination": "GDMFMYNJMV3KYKUYVWDNFH6IEGSL524LBQR5TXZ54BPOKGKS3DIMFIAB",
                                "destAsset": {                                                                                                     
                                    "assetCode": "XLM",
                                    "issuer": "SDF"
                                },
                                "destAmount": 2000000000,
                                "path": [
                                    {   
                                        "assetCode": "CENTUS",
                                        "issuer": "GCD5VCGHNLC46D5O2SAZCFFP3ZF7JRBQZGSFINXWNY5HCEOE42PH3XH5"
                                    },                                  
                                    {            
                                        "assetCode": "USD",    
                                        "issuer": "GDJAG52BJUFI6RBKFTNUEJBDZODJPZBLIHN3SXMR7QIXNRAOJK3XL6P3"
                                    }
                                ]       
                            }   

```

src/catchup/simulation/TxSimApplyTransactionsWork.cpp
```diff
@@ -67,8 +67,8 @@ checkOperationResults(xdr::xvector<OperationResult> const& expected,
         if (expected[i].code() != actual[i].code())
         {
             CLOG_ERROR(History, "Expected operation result {} but got {}",
-                       xdr_to_string(expected[i].code()),
-                       xdr_to_string(actual[i].code()));
+                       xdr_to_string(expected[i].code(), "OperationResultCode"),
```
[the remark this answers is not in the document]
hidenori-shinohara: fixed!

src/catchup/simulation/TxSimApplyTransactionsWork.cpp
```diff
@@ -181,7 +182,8 @@ checkResults(Application& app, uint32_t ledger,
             CLOG_ERROR(
                 History,
                 "Expected result code {} does not agree with {} for tx {}",
-                xdr_to_string(archiveRes.code()), xdr_to_string(dbRes.code()),
+                xdr_to_string(archiveRes.code(), "TransactionResultCode"),
```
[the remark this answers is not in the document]
hidenori-shinohara: fixed!

src/catchup/simulation/TxSimApplyTransactionsWork.cpp
```diff
@@ -154,10 +154,11 @@ checkOperationResults(xdr::xvector<OperationResult> const& expected,
 
         if (!match)
         {
-            CLOG_ERROR(History, "Expected operation result: {}",
-                       xdr_to_string(expectedOpRes));
-            CLOG_ERROR(History, "Actual operation result: {}",
-                       xdr_to_string(actualOpRes));
+            CLOG_ERROR(
+                History, "{}",
+                xdr_to_string(expectedOpRes, "Expected operation result"));
```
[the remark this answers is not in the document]
hidenori-shinohara: makes sense!

src/catchup/simulation/TxSimApplyTransactionsWork.cpp
```diff
@@ -196,8 +198,10 @@ checkResults(Application& app, uint32_t ledger,
                     "Expected result code {} does not agree with {} for "
```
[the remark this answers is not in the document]
hidenori-shinohara: removed!

src/util/XDRCereal.h
```diff
@@ -118,14 +138,7 @@ xdr_to_string(const T& t, std::string const& name = "", bool compact = false)
...
                         : cereal::JSONOutputArchive::Options::Default());
-        if (!name.empty())
-        {
-            ar(cereal::make_nvp(name, t));
-        }
-        else
-        {
-            ar(t);
-        }
+        ar(cereal::make_nvp(name, t));
```
jonjove, line 141: This doesn't seem quite right to me (and I don't think it was right before either). I think this should be done with `xdr::archive` to make sure it uses the appropriate archive adapter, thereby presenting access to `cereal_override`.

src/util/XDRCereal.h
```diff
@@ -51,6 +51,27 @@ cereal_override(cereal::JSONOutputArchive& ar,
     xdr::archive(ar, tmp, field);
 }
 
+template <uint32_t N>
+void
+cereal_override(cereal::JSONOutputArchive& ar,
```
jonjove, line 56: I don't think this is necessary once https://github.com/xdrpp/xdrpp/pull/25 is merged.

src/util/XDRCereal.h
hidenori-shinohara: Fixed! Thanks for pointing this out. I never noticed it.

src/util/XDRCereal.h
hidenori-shinohara: After several attempts, I think I'll give up fixing this in a more general way such as a new `save` function in XDRPP or `cereal_override` for `xvector`. While I definitely think that it's suboptimal that we need to implement something like this whenever we have an `xvector` containing elements with `cereal_override`, it seems that this is our second time, so I hope that this is not that bad. Let me know what you think!

hidenori-shinohara:
Updated the change using `clang-format-8`

src/util/XDRCereal.cpp
```diff
@@ -37,8 +37,17 @@ cereal_override(cereal::JSONOutputArchive& ar,
...
 
 void
-cereal_override(cereal::JSONOutputArchive& ar, const stellar::Asset& s,
+cereal_override(cereal::JSONOutputArchive& ar, const stellar::Asset& asset,
                 const char* field)
 {
-    xdr::archive(ar, stellar::assetToString(s), field);
+    xdr::archive(
+        ar,
+        std::make_tuple(
```
jonjove: This should be changed to render native asset as "native" (no sub-object).

src/util/XDRCereal.cpp
```diff
@@ -37,17 +37,19 @@ cereal_override(cereal::JSONOutputArchive& ar,
...
-        ar,
-        std::make_tuple(
-            cereal::make_nvp("assetCode", stellar::assetToString(asset)),
-            cereal::make_nvp(
-                "issuer",
-                asset.type() == stellar::ASSET_TYPE_NATIVE
-                    ? "SDF"
-                    : stellar::KeyUtils::toStrKey(stellar::getIssuer(asset)))),
-        field);
+    return std::make_tuple(
```
jonjove, line 49: Same thing as mentioned in the previous commit about rendering native asset.

src/util/XDRCereal.cpp
hidenori-shinohara: I only know one way to print an `xvector` of `asset` and that is to create a vector where each element is an NVP of the corresponding element of `xvector`. In other words, I'm not sure how I can render native asset without a sub-object and other assets as a sub-object. I made a change to make it clear that native asset is a native asset and there is no `assetCode` and `issuer` like following:

```
"pathPaymentStrictReceiveOp": {    
    "sendAsset": {                                
        "assetType": "native",                                
        "assetCode/issuer": "N/A"                                
    },
...
```

Let me know what you think!

src/util/XDRCereal.h
```diff
@@ -51,6 +51,22 @@ cereal_override(cereal::JSONOutputArchive& ar,
     xdr::archive(ar, tmp, field);
 }
 
+typedef cereal::NameValuePair<std::string> CerealStringNVP;
+typedef std::tuple<CerealStringNVP, CerealStringNVP> NVPTuple;
```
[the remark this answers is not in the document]
hidenori-shinohara: Makes sense. Since `NVPTuple` is only used a few times and it's unclear if we'll use it much more often, I got rid of the `typedef`.

src/invariant/LiabilitiesMatchOffers.cpp
```diff
@@ -256,8 +256,8 @@ checkBalanceAndLimit(LedgerHeader const& header, LedgerEntry const* current,
                 (INT64_MAX - account.balance < liabilities.buying))
             {
                 return fmt::format(
-                    "Balance not compatible with liabilities for account {}",
-                    xdr_to_string(account));
+                    "Balance not compatible with liabilities for {}",
+                    xdr_to_string(account, "account"));
```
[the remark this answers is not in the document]
hidenori-shinohara: Fixed!

src/test/FuzzerImpl.cpp
```diff
@@ -540,11 +540,12 @@ attemptToApplyOps(LedgerTxn& ltx, PublicKey const& sourceAccount,
...
-                    FMT_STRING("Error {} while applying operations "
-                               "while fuzzing: transaction result {}"),
-                    txFramePtr->getResultCode(),
-                    xdr_to_string(txFramePtr->getResult()));
+                auto const msg =
+                    fmt::format(FMT_STRING("Error {} while applying operations "
+                                           "while fuzzing: {}"),
+                                txFramePtr->getResultCode(),
+                                xdr_to_string(txFramePtr->getResult(),
+                                              "transaction result"));
```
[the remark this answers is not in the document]
hidenori-shinohara: fixed!

src/transactions/OperationFrame.cpp
```diff
@@ -122,15 +122,15 @@ OperationFrame::apply(SignatureChecker& signatureChecker,
...
+        CLOG_TRACE(Tx, "{}", xdr_to_string(mOperation, "Operation"));
     }
     res = checkValid(signatureChecker, ltx, true);
     if (res)
     {
         res = doApply(ltx);
         if (Logging::logTrace("Tx"))
         {
-            CLOG_TRACE(Tx, "Operation result: {}", xdr_to_string(mResult));
+            CLOG_TRACE(Tx, "{}", xdr_to_string(mResult, "Operation result"));
```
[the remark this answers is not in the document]
hidenori-shinohara: Done!

src/util/XDRCereal.cpp
hidenori-shinohara: Thank you for https://github.com/jonjove/stellar-core/commit/d7472a00f05788f1165722054174d0426cba13f5! I would have never come up with this idea.

I updated the PR with the change. I also tested nested containers (which we don't have now, but for future reference & out of curiosity), and it seems to work exactly as we'd expect. https://github.com/hidenori-shinohara/xdrpp/commit/abe8047a86fa7054d3ed4469d38bc53683b06094

src/util/XDRCereal.cpp
hidenori-shinohara: Fixed!

hidenori-shinohara:
I updated the PR with the `cereal_override` for all containers, and I also addressed all the comments.

src/util/XDRCereal.cpp
hidenori-shinohara: This is how asset information is printed with this change:
```
                        "body": {
                            "type": "PATH_PAYMENT_STRICT_RECEIVE",
                            "pathPaymentStrictReceiveOp": {
                                "sendAsset": "NATIVE",
                                "sendMax": 1000000000,
                                "destination": "GDMFMYNJMV3KYKUYVWDNFH6IEGSL524LBQR5TXZ54BPOKGKS3DIMFIAB",
                                "destAsset": "NATIVE",
                                "destAmount": 2000000000,
                                "path": [
                                    {
                                        "assetCode": "CENTUS",
                                        "issuer": "GCD5VCGHNLC46D5O2SAZCFFP3ZF7JRBQZGSFINXWNY5HCEOE42PH3XH5"
                                    },
                                    {
                                        "assetCode": "USD",
                                        "issuer": "GDJAG52BJUFI6RBKFTNUEJBDZODJPZBLIHN3SXMR7QIXNRAOJK3XL6P3"
                                    }
                                ]
                            }
                        }
```

src/util/XDRCereal.h
```diff
@@ -124,7 +124,7 @@ xdr_to_string(const T& t, std::string const& name, bool compact = false)
         cereal::JSONOutputArchive ar(
             os, compact ? cereal::JSONOutputArchive::Options::NoIndent()
                         : cereal::JSONOutputArchive::Options::Default());
-        ar(cereal::make_nvp(name, t));
+        xdr::archive(ar, t, name.c_str());
```
jonjove, line 127: I am pretty sure it's okay, but I just want to confirm that this does the right thing if `name.empty()` right?

src/util/XDRCereal.h
```diff
@@ -34,21 +34,28 @@ cereal_override(cereal::JSONOutputArchive& ar, const xdr::opaque_array<N>& s,
...
-void
-cereal_override(cereal::JSONOutputArchive& ar,
-                const xdr::xarray<stellar::Hash, N>& s, const char* field)
+template <typename T>
+std::enable_if_t<xdr::xdr_traits<T>::is_container>
+cereal_override(cereal::JSONOutputArchive& ar, T const& t, const char* field)
 {
-    std::vector<std::string> tmp;
-    for (auto const& h : s)
+    // setNextName, startNode, and finishNode create a new sub-object
```
jonjove, line 41: You might add some details about the fact that `setNextName, startNode, ..., finishNode` is basically what `ar(cereal::make_nvp(....))` does.

src/util/XDRCereal.h
```diff
@@ -34,21 +34,28 @@ cereal_override(cereal::JSONOutputArchive& ar, const xdr::opaque_array<N>& s,
...
+    // > SizeTags are strictly ignored for JSON, they just indicate
+    // > that the current node should be made into an array
+    //
+    // in include/cereal/archives/json.hpp
+    ar(cereal::make_size_tag(0));
+    for (auto const& element : t)
     {
-        tmp.emplace_back(
-            stellar::binToHex(stellar::ByteSlice(h.data(), h.size())));
+        xdr::archive(ar, element);
```
jonjove, line 56: You might mention that `xdr::archive` is required instead of `ar(...)` to actually access the nested cereal overrides.

jonjove (commented):
A couple small comments here about making the comments a bit more helpful to future maintainers. Generally, I'm happy with this. @graydon, how do you feel about the documentation level here?

src/util/XDRCereal.h
hidenori-shinohara: Thanks for pointing that out! I didn't think about that possibility. I checked the code and it just prints something like
```
"": {
 ...
}
```
which I doubt will be very useful. I decided to require that the input string be non-empty.

Initially, I felt that it might be an over-kill, but the original issue was that we have the word `value0`, so replacing it with the empty string will not really be much of an improvement, so I figured that I'd require it. Let me know what you think!

src/util/XDRCereal.h
hidenori-shinohara: Updated the comment. As I was updating it, I started to think that explicitly calling prologue and epilogue might be easier for future developers to understand since it'll be more similar to `ar()`, so I went ahead and replaced them. Let me know what you think!

src/util/XDRCereal.h
hidenori-shinohara: Added it!

src/util/XDRCereal.h
graydon: I think this is ok, but you might want to do a ReleaseAssertOrThrow (to ensure the assert remains inside release builds) and of course make sure tests all pass. maybe also run them at trace log-level in case we missed any cases.

graydon (commented):
I agree with the current structure for printing containers. Conversation with @jonjove the other day made sense, and this code matches my current recollection of that conversation.

graydon (approved):

src/util/XDRCereal.h
hidenori-shinohara: @graydon 
- I didn't know about `ReleaseAssert`! Although after some thinking, I started to wonder that if it'd be better if the `assert` does NOT remain inside release builds. I'm currently thinking that this issue of printing `"": {` may not be important enough to warrant an abort or throw in production.
- I just confirmed that all the tests do pass.

Let me know what you think!

hidenori-shinohara: After more thoughts & learning about https://github.com/stellar/stellar-core/issues/2852, I started thinking that it may be better to not have any assertion statements. Printing `"": {` doesn't seem bad enough to warrant an abort or throw in production.

graydon:
r+ 8df71d8

PR #2890, opened 2021-01-26
Add the command to simplify the master graph in OverlaySurvey.py

hidenori-shinohara:
# Description

The `simplify` command takes the "master" graph with all the information, and simplifies it and outputs it in a JSON file.

# Checklist
- [x] Compiles
- [x] Ran all tests

### DIFF
--- a/scripts/OverlaySurvey.py
+++ b/scripts/OverlaySurvey.py
@@ -232,6 +232,22 @@ def run_survey(args):
         json.dump(merged_results, outfile)
 
 
+def flatten(args):
+    output_graph = []
+    graph = nx.read_graphml(args.graphmlInput).to_undirected()
+    for node, attr in graph.nodes(data=True):
+        new_attr = {"publicKey": node, "peers": list(map(str, graph.adj[node]))}
+        for key in attr:
+            try:
+                new_attr[key] = json.loads(attr[key])
+            except (json.JSONDecodeError, TypeError):
+                new_attr[key] = attr[key]
+        output_graph.append(new_attr)
+    with open(args.jsonOutput, 'w') as output_file:
+        json.dump(output_graph, output_file)
+    sys.exit(0)
+
+
 def main():
     # construct the argument parse and parse the arguments
     argument_parser = argparse.ArgumentParser()
@@ -284,6 +300,19 @@ def main():
                                 required=True,
                                 help="output file for the augmented graph")
     parser_augment.set_defaults(func=augment)
+
+    parser_flatten = subparsers.add_parser("flatten",
+                                            help="Flatten a directed graph into "
+                                            "an undirected graph in JSON")
+    parser_flatten.add_argument("-gmli",
+                                 "--graphmlInput",
+                                 required=True,
+                                 help="input file containing a directed graph")
+    parser_flatten.add_argument("-json",
+                                 "--jsonOutput",
+                                 required=True,
+                                 help="output JSON file for the flattened graph")
+    parser_flatten.set_defaults(func=flatten)
 
     args = argument_parser.parse_args()
     args.func(args)
--- a/scripts/README.md
+++ b/scripts/README.md
@@ -21,6 +21,9 @@ This folder is for storing any scripts that may be helpful for using stellar-cor
     - sub command `augment` - analyze an existing graph
         - `-gmli GRAPHMLINPUT` - input graphml file
         - `-gmlo GRAPHMLOUTPUT` - output graphml file
+    - sub command `flatten` - Take a graphml file containing a bidrectional graph (possibly augmented with StellarBeat data) and flatten it into an undirected graph in JSON.
+        - `-gmli GRAPHMLINPUT` - input graphml file
+        - `-json JSONOUTPUT` - output json file
 
 ## Style guide
 We follow [PEP-0008](https://www.python.org/dev/peps/pep-0008/).


scripts/OverlaySurvey.py
```diff
@@ -232,6 +237,39 @@ def run_survey(args):
         json.dump(merged_results, outfile)
 
 
+def simplify(args):
+    simplified_graph = []
+    graph = nx.read_graphml(args.graphmlInput)
+    bidirectional_adj = {}
+    for v in graph.nodes():
```
marta-lokhova, line 244: Why do we need to do this? I think preferred peers should reflect _outbound_ peers of a node, so you need to keep the direction as-is.

scripts/OverlaySurvey.py
```diff
@@ -232,6 +237,39 @@ def run_survey(args):
...
+        appears_on_sb = False
+        for key in attr:
+            if key.startswith("sb_"):
+                new_key = key[3:]
+                appears_on_sb = True
+            else:
+                new_key = key
+            try:
+                new_attr[new_key] = json.loads(attr[key])
+            except Exception:
```
marta-lokhova, line 263: you should catch a specific exception type that `json.loads` throws, `JSONDecodeError` from what I see

scripts/OverlaySurvey.py
```diff
@@ -285,6 +323,18 @@ def main():
                                 help="output file for the augmented graph")
     parser_augment.set_defaults(func=augment)
 
+    parser_simplify = subparsers.add_parser('simplify',
+                                            help="simplify the master graph "
+                                            "and output it as json file")
+    parser_simplify.add_argument("-gmli",
```
marta-lokhova, line 329: Can we clarify the usage here? Specifically, the input graph here should be augmented? It looks like you should also specify `required=True`

scripts/OverlaySurvey.py
```diff
@@ -285,6 +323,18 @@ def main():
                                 help="output file for the augmented graph")
     parser_augment.set_defaults(func=augment)
 
+    parser_simplify = subparsers.add_parser('simplify',
+                                            help="simplify the master graph "
```
marta-lokhova, line 327: We need to explain this command in more detail, right now it's pretty vague. What exactly are we simplifying? Maybe `simplify` isn't the best name for it - we're basically setting up a new graph for the simulation purposes, so perhaps that's what we should name it: `adjustForSimulation` or something like that.

scripts/OverlaySurvey.py
hidenori-shinohara: I added this part since I was thinking that the direction of an edge in the survey results tells us who sets the connection up initially, but once established all traffic flows both ways? I'm not 100% sure, though.

This also seems to make the simulation simpler. For instance, by making the graph bidirectional, we can ensure that all nodes have been connected to all the nodes it's supposed to be connected to by checking if `the # of nodes connected to node v` equals `the length of preferredPeers list`. If we don't make it bidirectional, we probably have to check something like `the preferredPeers list` ⊂ `the nodes connected to v`.

scripts/OverlaySurvey.py
hidenori-shinohara: Updated the comment, and added `required=True`!

scripts/OverlaySurvey.py
hidenori-shinohara: Updated the command name, and I also added more explanation to `README.md`.

scripts/OverlaySurvey.py
```diff
@@ -285,6 +323,20 @@ def main():
...
 
+    parser_simplify = subparsers.add_parser('adjustForSimulation',
+                                            help="adjust the master graph"
+                                            "for simulation purposes")
+    parser_simplify.add_argument("-gmli",
+                                 "--graphmlInput",
+                                 required=True,
+                                 help="input master graph"
+                                 "(i.e., output of the augment command)")
+    parser_simplify.add_argument("-json",
```
MonsieurNicolas, line 334: This command is not right: it actually performs 2 distinct things.

The paradigm in this script is to have various "transforms" that operate on graph data, and removing nodes or connections would be such a transform that would be run 

* It simplifies the graph, by removing nodes based on some criteria.
* It converts a graph into json.

I'm pretty sure we need that second one  (we can call it `flatten`), that takes a graph and "flattens" it in JSON form (undirected).

scripts/OverlaySurvey.py
MonsieurNicolas: nx can import graphs as undirected - so you don't need to do this manual tracking.
By using the right data structure, you will fix a bug: `peers` contains nodes that you deleted.

scripts/OverlaySurvey.py
```diff
@@ -232,6 +237,39 @@ def run_survey(args):
...
+            if key.startswith("sb_"):
+                new_key = key[3:]
+                appears_on_sb = True
+            else:
+                new_key = key
+            try:
+                new_attr[new_key] = json.loads(attr[key])
+            except (json.JSONDecodeError, TypeError):
+                new_attr[new_key] = attr[key]
+        new_attr["preferredPeers"] = list(bidirectional_adj[node])
```
MonsieurNicolas, line 265: nit: they are just `peers` in this context, not `preferredPeers`

scripts/OverlaySurvey.py
```diff
@@ -115,15 +115,20 @@ def augment(args):
...
-                                  "index",
-                                  "isp",
-                                  "ip"]
+            desired_properties = [
+                "quorumSet",
+                "geoData",
+                "isValidating",
+                "name",
+                "homeDomain",
+                "organizationId",
```
MonsieurNicolas, line 124: I don't see why we'd need any of those new properties (taking this offline)

MonsieurNicolas (changes requested):
I think the `flatten` part makes sense to me, the other one (that removes nodes)  does not. Let's talk offline

scripts/OverlaySurvey.py
marta-lokhova: ok clarification on my previous comment: @MonsieurNicolas  let me know that we can use both inbound and outbound peers (as opposed to just outbound) as preferred peers so that we can use the "preferred peers only" flag (the flag is useful to reduce potential churn when nodes need to reconnect, as it rejects any connections that are not preferred peers).

scripts/OverlaySurvey.py
hidenori-shinohara: Fixed!

scripts/OverlaySurvey.py
hidenori-shinohara: Replaced the command with the `flatten` command.

scripts/OverlaySurvey.py
hidenori-shinohara: Updated it to `peers`.

scripts/OverlaySurvey.py
hidenori-shinohara: Removed.

MonsieurNicolas:
r+ c5678c264b9383f8d554a2ad75d271e5bc7f4815

PR #2891, opened 2021-01-26
Use some authorization flags in fuzzer setup

rokopt:
# Use some authorization flags in fuzzer setup

Part of #1376 

Set authorization flags on a couple of issuing accounts during fuzzer setup, and later issue `ALLOW_TRUST` operations from them for the accounts to which they will be distributing assets.

There are more variations we could add, but I figured I'd start simple and to validate an approach.

# Checklist
- [x] Compiles
- [x] Ran all tests

### DIFF
--- a/src/test/FuzzerImpl.cpp
+++ b/src/test/FuzzerImpl.cpp
@@ -602,6 +602,20 @@ TransactionFuzzer::initialize()
 
             ops.emplace_back(createOp);
 
+            // Make some issuing accounts require authorization.
+            auto const accountFlags =
+                (i == 1)
+                    ? AUTH_REQUIRED_FLAG
+                    : (i == 2) ? AUTH_REQUIRED_FLAG | AUTH_REVOCABLE_FLAG : 0;
+
+            if (accountFlags != 0)
+            {
+                auto optionsOp =
+                    txtest::setOptions(txtest::setFlags(accountFlags));
+                optionsOp.sourceAccount.activate() = toMuxedAccount(publicKey);
+                ops.emplace_back(optionsOp);
+            }
+
             // select the first pregenerated account to be the hard coded source
             // account for all transactions
             if (i == 0)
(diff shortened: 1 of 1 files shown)


MonsieurNicolas:
As we add more and more special cases... it seems that we're just going to eventually end up with really complicated code with many (sometimes nested) conditions.

Should we just instead migrate to listing out explicitly the state for each test account and order book?

If we do this right, a combination of data sets and generators should allow us to actually list all combinations in a way that is human readable (with comments on which properties we're shooting for every time we add a new block of data).

rokopt:
> Should we just instead migrate to listing out explicitly the state for each test account and order book?

I'd be happy to -- I did write an earlier iteration of fuzzer setup which separated the declaration of an explicit state from performing the setup operations using that state to take parameters from.  I will look for a better way of doing something similar.

Unless you feel otherwise, though, I'd rather do that in a separate, follow-on PR which only changed the mechanism and didn't change the semantics or parameters.  I suspect it'd be easier to write and to review that way, and would have the advantage of having correct uses of the needed operations already in place and tested, so that any test failures in the development of the follow-on PR could immediately be identified as having accidentally changed semantics, rather than having chosen an invalid setup or misused any of the test interfaces that create the operations.

MonsieurNicolas:
I don't think you got the gist of what I was saying:
I understand the code change, I don't understand if it's the right one (as in - it's pretty much impossible to understand its impact), so I don't know if we're making some weird tradeoff where we make the 99% case harder to analyze by a factor of 2.

The previous implementation of the setup code did not achieve that goal either, it was basically the same thing but with more abstractions getting in the way.

From what it looks like that's because we're not actually reasoning about higher level constructs like "order books", "market makers" and "takers" and the various properties we want to inject in each (which may or not depend on asset types - as in issuer flags).

For this PR for example, it's actually very hard to understand just a basic fact about the setup: how many order books are being effected, and of course what we'd really care about is understanding how many of the existing links (in terms of potential combinations of interactions) in the setup are being effected.

rokopt:
> I don't think you got the gist of what I was saying:
> I understand the code change, I don't understand if it's the right one (as in - it's pretty much impossible to understand its impact), so I don't know if we're making some weird tradeoff where we make the 99% case harder to analyze by a factor of 2.
> 
> The previous implementation of the setup code did not achieve that goal either, it was basically the same thing but with more abstractions getting in the way.
> 
> From what it looks like that's because we're not actually reasoning about higher level constructs like "order books", "market makers" and "takers" and the various properties we want to inject in each (which may or not depend on asset types - as in issuer flags).
> 
> For this PR for example, it's actually very hard to understand just a basic fact about the setup: how many order books are being effected, and of course what we'd really care about is understanding how many of the existing links (in terms of potential combinations of interactions) in the setup are being effected.

I've tried again in PR #2897 .  Is that along the lines of what you envision?

rokopt:
Now that PR #2897 has been merged, this should be done in the same style as that PR.  I've done that in PR #2927 , so this one is superseded.

PR #2892, opened 2021-01-26
Create passive offers during fuzzer setup

rokopt:
# Create passive offers during fuzzer setup

Part of #1376 

Create a couple of passive offers during fuzzer setup.

We will probably look for many other interesting combinations of offers to add in the future, but I figured I'd start simple and try to validate an approach.

# Checklist
- [x] Compiles
- [x] Ran all tests

### DIFF
--- a/src/test/FuzzerImpl.cpp
+++ b/src/test/FuzzerImpl.cpp
@@ -560,7 +560,11 @@ attemptToApplyOps(LedgerTxn& ltx, PublicKey const& sourceAccount,
                          MANAGE_OFFER_DELETED) ||
                     (opType == MANAGE_SELL_OFFER &&
                      tr.manageSellOfferResult().success().offer.effect() ==
-                         MANAGE_OFFER_DELETED))
+                         MANAGE_OFFER_DELETED) ||
+                    (opType == CREATE_PASSIVE_SELL_OFFER &&
+                     tr.createPassiveSellOfferResult()
+                             .success()
+                             .offer.effect() == MANAGE_OFFER_DELETED))
                 {
                     auto const msg = fmt::format(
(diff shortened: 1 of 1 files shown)


MonsieurNicolas:
It seems to me that we're continuing to move forward with highly symmetrical setup - 0 spread seems to be quite special for an order book that maybe we should only have that for one pair?

rokopt:
> It seems to me that we're continuing to move forward with highly symmetrical setup - 0 spread seems to be quite special for an order book that maybe we should only have that for one pair?

I think that the ability to create orders with a 0 spread is what's distinct about a passive offer -- that's why I'm creating this new one with a 0 spread.  But maybe I'm misunderstanding that.

Agreed that we should change the spreads in the existing orders, but maybe in a separate PR?

MonsieurNicolas:
> I think that the ability to create orders with a 0 spread is what's distinct about a passive offer -- that's why I'm creating this new one with a 0 spread. But maybe I'm misunderstanding that.

This is correct, I am asking if we should have that many 0 spread order books. I don't understand why we need so many of them.

rokopt:
> > I think that the ability to create orders with a 0 spread is what's distinct about a passive offer -- that's why I'm creating this new one with a 0 spread. But maybe I'm misunderstanding that.
> 
> This is correct, I am asking if we should have that many 0 spread order books. I don't understand why we need so many of them.

I don't know of any (good) reason; it's one of the many further enhancements to setup that I think we both agree we should make.  It just wasn't my goal in this specific PR.

rokopt:
Now that PR #2897 has been merged, this should be done in the same style as that PR.  I've done that in PR #2927 , so this one is superseded.

PR #2894, opened 2021-01-27
Reduce trustline limits after fuzzer offer setup

rokopt:
# Reduce trustline limits after fuzzer offer setup

Part of #1376 

As a follow-on to PR #2858 , which reduced account balances after setting up offers for fuzzing, reduce trust line limits similarly.  The idea is to make it plausible that the fuzzer will sometimes attempt operations which would exceed limits.

# Checklist
- [x] Reviewed the [contributing](https://github.com/stellar/stellar-core/blob/master/CONTRIBUTING.md#submitting-changes) document
- [x] Rebased on top of master (no merge commits)
- [x] Ran `clang-format` v5.0.0 (via `make format` or the Visual Studio extension)
- [x] Compiles
- [x] Ran all tests
- [ ] If change impacts performance, include supporting evidence per the [performance document](https://github.com/stellar/stellar-core/blob/master/performance-eval/performance-eval.md)

PR #2897, opened 2021-01-29
Separate fuzzer setup parameters from operations

rokopt:
# Separate fuzzer setup parameters from operations

Part of #1376 

Could be the beginning of a foundation for #2891 , #2892 , #2894 

Factor out the specifics of which offers to create into a
single structure, which can be examined at a glance and
changed in isolation, from the operations which establish
that order book structure during fuzzer setup.

We plan to set up the order book to be more varied and less symmetrical than it currently is, but I think that we should separate working out how to factor out the parameters to make that kind of change simpler to perform and understand from actually making those changes.  So in this commit, I'm not changing the initial order book, only how the setup is specified (this commit makes it declarative, and `constexpr`).

# Checklist
- [x] Compiles
- [x] Ran all tests

### DIFF
--- a/src/test/FuzzerImpl.cpp
+++ b/src/test/FuzzerImpl.cpp
@@ -574,6 +574,102 @@ attemptToApplyOps(LedgerTxn& ltx, PublicKey const& sourceAccount,
     }
 }
 
+// Unlike Asset, this can be a constexpr.
+struct AssetID
+{
+    constexpr AssetID() : mIsNative(true), mId(0)
+    {
+    }
+
+    constexpr AssetID(int id) : mIsNative(false), mId(id)
+    {
+    }
+
+    Asset
+    toAsset() const
+    {
+        return mIsNative ? txtest::makeNativeAsset()
+                         : FuzzUtils::makeAsset(mId);
+    }
+
+    bool const mIsNative;
+    int const mId; // meaningful only if !isNative
+};
+
+struct OfferParameters
+{
+    constexpr OfferParameters(int publicKey, AssetID const bid,
+                              AssetID const sell, int64_t amount,
+                              int32_t priceNumerator, int32_t priceDenominator)
+        : mPublicKey(publicKey)
+        , mBid(bid)
+        , mSell(sell)
+        , mAmount(amount)
+        , mNumerator(priceNumerator)
+        , mDenominator(priceDenominator)
+    {
+    }
+
+    int const mPublicKey;
+    AssetID const mBid;
+    AssetID const mSell;
+    int64_t const mAmount;
+    int32_t const mNumerator;
+    int32_t const mDenominator;
+};
+
+// The current order book setup generates identical configurations for the
+// following asset pairs:
+//      XLM - A
+//      A   - B
+//      B   - C
+//      C   - D
+//
+// For any asset A and asset B, the generic order book setup will be as
+// follows:
+//
+// +------------+-----+------+--------+------------------------------+
+// |  Account   | Bid | Sell | Amount | Price (in terms of Sell/Bid) |
+// +------------+-----+------+--------+------------------------------+
+// | 0          | A   | B    |     10 | 3/2                          |
+// | 1 (issuer) | A   | B    |     50 | 3/2                          |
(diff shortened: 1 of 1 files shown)


src/test/FuzzerImpl.cpp
```diff
@@ -574,6 +574,100 @@ attemptToApplyOps(LedgerTxn& ltx, PublicKey const& sourceAccount,
...
+// | 3 (issuer) | B   | A    |     10 | 10/9                         |
+// | 4          | B   | A    |     50 | 10/9                         |
+// | 0          | B   | A    |    100 | 22/7                         |
+// +------------+-----+------+--------+------------------------------+
+//
+// This gives us a good mixture of buy and sell, issuers with offers,
+// and an account with a bid and a sell. It also gives us an initial
+// order book that is in a legal state.
+std::array<OfferParameters, 24> constexpr orderBookParameters{
+    {OfferParameters{AssetID(), AssetID(1), 13, 3, 2, 10}, // asset pair 0
```
MonsieurNicolas, line 646: I think we should try to  make this section look as much as possible like the comment above (and that comment should be removed as it's wrong anyways):
* use an empty line between order book pairs
* price should be the last argument, and price should be sorted so that we can clearly see how the order book is structured (note that because price is in terms of sell/bid, the sort order flips when switching buy and sell)
* I think you can get rid of the `OfferParameters`? (or maybe this is C++ 17)

MonsieurNicolas: Each section we should have some comment that explains which property we're trying to achieve.

To do this, we may have to do the same for accounts, that right now are setup pretty much all the same (instead, we probably want to mix things up a bit in terms of limits on balance entries and assets trusted).

Note: as we're trying to match the coverage from master, we may need less accounts and offers variations than what we're doing now.

MonsieurNicolas (changes requested):
looks to be a step in the right direction

src/test/FuzzerImpl.cpp
rokopt: > and that comment should be removed as it's wrong anyways

I left the part that documents the initial state of the order book, but removed the stuff about it being a "good mixture".

rokopt: > use an empty line between order book pairs

Done.

rokopt: > price should be the last argument, and price should be sorted so that we can clearly see how the order book is structured (note that because price is in terms of sell/bid, the sort order flips when switching buy and sell)

I've made Price into the last argument (and otherwise brought the argument order into line with the comment).  I think it was already in best-offer order?  I might be misunderstanding that.

rokopt: > I think you can get rid of the OfferParameters? (or maybe this is C++ 17)

Oh yes, you're right.  I thought it was C++17, but `clang` at least is allowing me to do it with `std=c++14`.

rokopt: > Each section we should have some comment that explains which property we're trying to achieve.
>
> To do this, we may have to do the same for accounts, that right now are setup pretty much all the same (instead, we probably want to mix things up a bit in terms of limits on balance entries and assets trusted).
>
> Note: as we're trying to match the coverage from master, we may need less accounts and offers variations than what we're doing now.

Agreed on all points -- I wanted to start with just the order book so that we could get consensus on how the parameter-refactoring should look, and then once that was merged, I'd do the same for accounts, and for the other structures that I'm adding to setup.  I also definitely plan to change the setup parameters themselves, but I wanted to separate the mechanism changes from the policy changes.

rokopt:
I think I've handled all code-review comments; how does the change look now?

src/test/FuzzerImpl.cpp
MonsieurNicolas: I think this is good now. I'm going to merge it tomorrow

rokopt: > I think this is good now. I'm going to merge it tomorrow

Thanks.  I've rebased onto master.

MonsieurNicolas:
r+ 9ef9dbceb3e9d9863d7a805cfef13474209f589a

Issue #2888, opened 2021-01-22
crash when running new-db command


MonsieurNicolas:
(copy/pasted from the dev-mailing list - I didn't try it)

I installed stellar-core from stable repository. Stellar-core version 15.2.0. OS Ubuntu 18.04. I run command:
`stellar-core new-db --conf /etc/stellar/stellar-core.cfg`

Stellar-core returned error:

```
2021-01-21T11:47:46.011 [default INFO] Config from /etc/stellar/stellar-core.cfg
2021-01-21T11:47:46.012 [default INFO] Using QUORUM_SET: {
   "t" : 3,
   "v" : [
      "sdf_watcher1",
      "sdf_watcher2",
      "sdf_watcher3",
      {
         "t" : 3,
         "v" : [ "stronghold1", "eno", "tempo.eu.com", "satoshipay" ]
      }
   ]
}

Segmentation fault (core dumped)
```

Config file use this: https://github.com/stellar/docs/blob/master/other/stellar-core-validator-example.cfg
What is my error?

marta-lokhova:
this seems to be related to PostgreSQL access rights. Specifically, this soci code appears to be causing issues: https://github.com/stellar/stellar-core/blob/9d94156182056aca560628758b47430a43e06252/src/database/Database.cpp#L195 Switching to a user with the right privileges fixes it for me

MonsieurNicolas:
Good find. Yeah we're probably missing some logging when there is an exception?

marta-lokhova:
Looked into this more closely, and I think I blamed postgres prematurely! The stack trace actually points to spdlog. 
```
Thread 0 Crashed:: Dispatch queue: com.apple.main-thread
0   libsystem_c.dylib             	0x00007fff6b988a5d flockfile + 18
1   libsystem_c.dylib             	0x00007fff6b98bb04 fwrite + 66
2   stellar-core                  	0x0000000102e7cd87 spdlog::details::file_helper::write(fmt::v7::basic_memory_buffer<char, 250ul, std::__1::allocator<char> > const&) + 39 (file_helper-inl.h:76)
3   stellar-core                  	0x0000000102e7d523 spdlog::sinks::basic_file_sink<std::__1::mutex>::sink_it_(spdlog::details::log_msg const&) + 115 (basic_file_sink-inl.h:34)
[...]

PR #2906, opened 2021-02-05
Fix legacy logging

marta-lokhova:
Accidentally discovered that we can't build core with `--disable-spdlog` flag while looking into #2888


graydon:
r+ 9d128f8

graydon (approved):
Nice catch, thanks.

PR #2908, opened 2021-02-06
Do not enforce bucket oldest versions check

marta-lokhova:
Introduced by #2843, this check breaks private networks that do not upgrade fast enough


MonsieurNicolas:
r+ b1dbad492f746ca719186ffcdc894087fc9a36b1

PR #2909, opened 2021-02-09
logging:do not use reserved names (Windows)

MonsieurNicolas:
`ERROR` is `#define`'d in some headers and conflicts with the change from #2906


marta-lokhova:
r+ e2d6aec5e720708941defbe2c667490a4dc77a36

PR #2910, opened 2021-02-09
Switch to llvm-8 and gcc-7 (focal updates)

MonsieurNicolas:
# Description

5.0.0 is not available anymore (cannot be downloaded)


MonsieurNicolas:
this PR updates CI to use llvm-8, and switches to clang-format-8

MonsieurNicolas:
Turns out the issue was the switch to focal. So this PR just makes everything build under focal

graydon:
r+ d50695a

PR #2919, opened 2021-02-12
Reject peers early if no pending slots are available

### LINKED ISSUE #2915: Overlay: dead peers logic is wrong
Each known peer we keep track of has `num failures` attribute, which is used for outbound connections. When we attempt connecting to a peer, we increase the back-off. When the outbound peer successfully authenticates, we reset `num failures` back to 0 to indicate that the peer is responsive and can successfully connect.

We increase the back-off right before we attempt the connection, and reset it after we receive a valid AUTH message. The problem is that in between these steps, it is possible that stellar-core runs into conditions that would make it drop the new peer. So far, I saw two, but there might be more:
* Peer promotion. When there is an authenticated inbound connection, and we attempt a new outbound connection, it will be rejected because the peer is already connected
* We make an outbound connection even when no pending slots are available (initiate a connection, send hello, then drop peer)

In both cases, back-off is increased (since the connection is initiated) and not reset (as we never get to complete authentication), even though it's not the peer's fault that we couldn't connect to it. Peers with failures are periodically purged from the DB, so it is possible that good peers will be purged (although the failure threshold right now is 120, so I think it's less severe). We should make it such that only actual connection failures are counted.

marta-lokhova:


src/overlay/OverlayManagerImpl.cpp
```diff
@@ -697,8 +706,7 @@ OverlayManagerImpl::addOutboundConnection(Peer::pointer peer)
     assert(peer->getRole() == Peer::WE_CALLED_REMOTE);
     mOutboundPeers.mConnectionsAttempted.Mark();
 
-    if (mShuttingDown || mOutboundPeers.mPending.size() >=
-                             mApp.getConfig().MAX_OUTBOUND_PENDING_CONNECTIONS)
+    if (mShuttingDown || availableOutboundPendingSlots() <= 0)
```
MonsieurNicolas, line 709: thanks for cleaning this up

MonsieurNicolas (changes requested):

MonsieurNicolas:
r+ 3c0d2270041ddac4ebbfcf6d635a4aedd79a0191

PR #2927, opened 2021-02-18
Separate more fuzzer setup parameters into constexpr arrays

rokopt:
# Separate more fuzzer setup parameters into constexpr arrays

Part of  #1376 

Supersedes #2891 
Supersedes #2892 
Supersedes #2894 

Follow-on from #2897 

In the same style as was done for some order book parameters in #2897 , introduce `constexpr` arrays with fuzzer setup parameters determining authorization-related flags, post-setup trustline limits, and which offers are passive.  Also factor out some existing parameters, such as post-setup account balances, in the same way.

# Checklist
- [x] Compiles
- [x] Ran all tests

### DIFF
--- a/src/test/FuzzerImpl.cpp
+++ b/src/test/FuzzerImpl.cpp
@@ -755,9 +755,10 @@ TransactionFuzzer::initialize()
 
             for (int j = 0; j < FuzzUtils::NUMBER_OF_PREGENERATED_ACCOUNTS; ++j)
             {
-                auto const asset = FuzzUtils::makeAsset(j);
                 if (i != j && accountParameters[j].mIssueAsset)
                 {
+                    auto const asset = FuzzUtils::makeAsset(j);
+
                     // trust asset issuer
                     auto trustOp = txtest::changeTrust(
                         asset, FuzzUtils::INITIAL_TRUST_LINE_LIMIT);
@@ -837,9 +838,10 @@ TransactionFuzzer::initialize()
             for (uint8_t j = 0; j < FuzzUtils::NUMBER_OF_PREGENERATED_ACCOUNTS;
                  ++j)
             {
-                auto const asset = FuzzUtils::makeAsset(j);
                 if (i != j && accountParameters[j].mIssueAsset)
                 {
+                    auto const asset = FuzzUtils::makeAsset(j);
+
                     PublicKey issuer;
                     FuzzUtils::setShortKey(issuer, j);
 
--- a/src/test/FuzzerImpl.cpp
+++ b/src/test/FuzzerImpl.cpp
@@ -40,7 +40,6 @@ auto constexpr FUZZING_FEE = 1;
 auto constexpr FUZZING_RESERVE = 4;
 auto constexpr INITIAL_TRUST_LINE_LIMIT = 5 * INITIAL_ASSET_DISTRIBUTION;
 auto constexpr DEFAULT_NUM_TRANSACTIONS_TO_RESERVE_FEES_FOR = 10;
-auto constexpr DEFAULT_ASSET_AVAILABLE_FOR_TEST_ACTIVITY = 256;
 
 // must be strictly less than 255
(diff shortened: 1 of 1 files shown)


src/test/FuzzerImpl.cpp
```diff
@@ -768,6 +786,19 @@ TransactionFuzzer::initialize()
                     PublicKey issuer;
                     FuzzUtils::setShortKey(issuer, j);
 
+                    // Allow trust for any accounts whose parameters require it.
```
jonjove, line 818: This change doesn't make a lot of sense to me for a variety of reasons:
1. If `issuerAccount.flags == 0` then `trustLine.flags != 0` when the trust line is created (check how `ChangeTrust` works), so perhaps you actually want to set them to 0.
2. This is implicitly assuming that all the trust lines for a single asset should have the same flags.

src/test/FuzzerImpl.cpp
```diff
@@ -893,9 +910,20 @@ TransactionFuzzer::initialize()
                             toMuxedAccount(account);
                         ops.emplace_back(reduceNonNativeBalanceOp);
                     }
-                    // Here we could also reduce account "i"'s trust line limit
-                    // for asset "j" (testing the buying liabilities to
-                    // determine by how much).
+
+                    // Reduce "account"'s trustline limit for asset "j".
```
jonjove, line 932: I'm not sure about this for the same reason I'm unsure about the trust line flags in an earlier commit. In both places, you are implicitly assuming uniformity and the implementation doesn't permit anything else. Given that I think uniformity is unlikely what we desire (although in what ways it shouldn't be uniform will be for you to determine) I'm not sure how much value there is in these implementations.

jonjove (commented):
Mostly looks good to me, but there are two changes that might be a bit premature given we don't know what functionality is actually required.

src/test/FuzzerImpl.cpp
rokopt: Yeah, I thought it might be more readable if I did that step first and then after getting that in replaced the single limits with per-asset arrays in `AccountParameters`.  If that's a sensible approach, I could either remove this step from the PR altogether and save it all for a future PR, update the commit to do all that in one step in this PR, or add a further commit to this one to take that further step.  (Even after that, it would be possible to take another step afterwards to make the issuance of `Asset`s more flexible -- currently the number and codes are the same for each issuer.)  Do you have a preference?

src/test/FuzzerImpl.cpp
rokopt: > This is implicitly assuming that all the trust lines for a single asset should have the same flags.

Same question as below -- I'd been planning to take that further step toward increased variety in a follow-up PR as I thought it would be clearer, but I might be misjudging that.

rokopt: > If issuerAccount.flags == 0 then trustLine.flags != 0 when the trust line is created (check how ChangeTrust works), so perhaps you actually want to set them to 0.

Aha -- I think I'm gather that `AUTHORIZED_FLAG` might already be set (if the issuer doesn't have `AUTH_REQUIRED` set) and `TRUSTLINE_CLAWBACK_ENABLED_FLAG` might already be set (if the issuer has `AUTH_CLAWBACK_ENABLED` set).

One option would be to make `mAllowTrustFlags` `optional`, being set only if we want the trust line to have flags other than the ones with which it was created as a consequence of the issuer's account flags.  That'd make the initialization of `accountParameters` more verbose, but it would also make it easy to eyeball exactly which trustline flags _weren't_ going to be in the same state after setup that they were in immediately after their creations.  I think that leads me to favor doing that, but am open to other preferences.

rokopt:
Since we're not sure of the shape that we want for the representation of the auth flags or trustline limit reductions yet, I've rebased them out of this pull request (and stashed them in another branch).

jonjove:
This looks good to me, I'll merge it after the 15.3.0 release.

jonjove:
r+ afad87e

PR #2931, opened 2021-02-22
Add --disable-libunwind support to configure.ac

graydon:
This lets users override the autodetect of libunwind and forcibly turn it off.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -343,37 +343,42 @@ AC_ARG_ENABLE(next-protocol-version-unsafe-for-production,
 AM_CONDITIONAL(ENABLE_NEXT_PROTOCOL_VERSION_UNSAFE_FOR_PRODUCTION,
     [test x$enable_next_protocol_version_unsafe_for_production = xyes])
 
-case "${host_os}" in
-    *darwin*)
-        # libunwind comes standard with the command-line tools on macos
-        AC_DEFINE([HAVE_LIBUNWIND], [1],
-            [Define to 1 if you have the <libunwind.h> header file])
-        ;;
-    *)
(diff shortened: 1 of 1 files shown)


configure.ac
```diff
@@ -343,37 +343,42 @@ AC_ARG_ENABLE(next-protocol-version-unsafe-for-production,
...
+            # default of libcxx and a requirement for libstdc++). I think this is
+            # roughly caused by both libgcc and libunwind providing the C++ EH ABI
+            # symbols but, evidently, interacting with clang-compiled code slightly
+            # differently when ELF-interposed with one another, in the same binary.
+            #
+            # Haven't been able to make a reduced testcase that works. You can check
+            # that this is still a problem by running the stellar-core Catch2-based
+            # unit test suite: many of the tests do REQUIRE_THROWS_AS(...) and
+            # that crashes with libunwind+clang.  Haven't been able to figure out a
+            # workaround either.
```
Geod24, line 368: Just wondering if that's because LLVM [has its own `libunwind`](https://github.com/llvm/llvm-project/tree/main/libunwind) which defines the same symbols as the upstream one but have a different ABI. For example, Alpine Linux has [a separate package for it](https://pkgs.alpinelinux.org/package/edge/community/x86_64/llvm-libunwind).

MonsieurNicolas:
r+ 55fa30c02117ed5667c5ebce73e056f6b89a1e02

configure.ac
graydon: Something like this, yes. I think it might be possible to get around if you compile your own libcxx and explicitly disable their libunwind / rely on the libgcc one, but I couldn't make it work (and anyway we wouldn't be able to convince users that this is a reasonable step).

PR #2938, opened 2021-03-01
Make stored fuzzer ledger keys deterministic

rokopt:
# Make stored fuzzer ledger keys deterministic

Part of #1376 
Follow-on to #2904 

I realize that the array of stored ledger keys created by #2904 in fuzzer setup is non-deterministic (it can vary from one `stellar-core fuzz` execution to the next) for two reasons:

- We now use `autocheck`, to generate the mix of validated and unvalidated ledger keys to fill in the room we have that's not used by setup, and `autocheck` has its own PRNG.  So we should reseed that in the same place as we reseed `gRandomEngine` (and call `srand()`).
- `getAllEntries()` is not guaranteed to return entries in the same order with each call, since it uses an `unordered_map`, which does not make any such guarantee.  So we should sort the returned entries.

# Checklist
- [x] Compiles
- [x] Ran all tests

### DIFF
--- a/src/test/FuzzerImpl.cpp
+++ b/src/test/FuzzerImpl.cpp
@@ -673,6 +673,7 @@ resetRandomSeed(unsigned int seed)
     // seed randomness
     srand(seed);
     gRandomEngine.seed(seed);
+    autocheck::rng().seed(seed);
 }
 
 static void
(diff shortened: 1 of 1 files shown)


jonjove:
r+ ec94ff0

PR #2940, opened 2021-03-02
Fix several signed overflows in tests

rokopt:
# Fix several signed overflows in tests

`ubsan` reveals three signed overflows in our tests, all of which I think admit small fixes:

- `TestAccount::nextSequenceNumber()` does not account for the possibility that `mSn` might already be `INT64_MAX`.  This makes it simply remain at `INT64_MAX` in that case . (The `"large bump"` test in `BumpSequenceTests.cpp`  is the one which exposes the UB.)
- The `"ExchangeV10"` test case in `ExchangeTests.cpp` has a potential overflow in a `REQUIRE()` when multiplying 64-bit numbers.  There are several other tests in the same file which avoid this by casting to `uint128_t`, so this PR does the same.  (It also adds `REQUIRE()`s that confirm that it's not casting any negative values to `uint128_t`.)
- The `"Account balances unchanged without inflation"` test in `ConservationOfLumens.cpp` increases `LedgerHeader::totalCoins` by the sum of the balances of the accounts that it's going to delete:
    ```
        int64_t balanceToDelete = getTotalBalance(toDelete);
        auto entries4 = updateBalances(entries3, *app, balanceToDelete);
    ```
  But nothing ensures that that increase can't overflow.  What this PR proposes is to check whether it would overflow and, if so, shorten the `toDelete` list until it _wouldn't_ overflow:
    ```
        while (getTotalBalance(std::vector<LedgerEntry>(
                   keepEnd, entries2.end())) > maxBalanceIncrease(*app))
        {
            ++keepEnd;
        }
    ```
  In my tests, during the 100 times through that loop, `keepEnd` had to be bumped during only 3 of the loops, so I don't think that this change is reducing the effectiveness of the test significantly.  There might be a simpler/better fix, though; I don't think I fully understand the aims of this test.

# Checklist
- [x] Compiles
- [x] Ran all tests

### DIFF
--- a/src/invariant/test/ConservationOfLumensTests.cpp
+++ b/src/invariant/test/ConservationOfLumensTests.cpp
@@ -43,7 +43,7 @@ getCoinsAboveReserve(std::vector<LedgerEntry> const& entries, Application& app)
 
 std::vector<LedgerEntry>
 updateBalances(std::vector<LedgerEntry> entries, Application& app,
-               int64_t netChange)
+               int64_t netChange, bool updateTotalCoins)
 {
     int64_t initialCoins = getTotalBalance(entries);
     int64_t pool = netChange + getCoinsAboveReserve(entries, app);
@@ -79,9 +79,18 @@ updateBalances(std::vector<LedgerEntry> entries, Application& app,
     auto finalCoins = getTotalBalance(entries);
     REQUIRE(initialCoins + netChange == finalCoins);
 
-    LedgerTxn ltx(app.getLedgerTxnRoot());
-    ltx.loadHeader().current().totalCoins += netChange;
-    ltx.commit();
+    if (updateTotalCoins)
+    {
+        LedgerTxn ltx(app.getLedgerTxnRoot());
+        auto& current = ltx.loadHeader().current();
+        REQUIRE(current.totalCoins >= 0);
+        if (netChange > 0)
+        {
+            REQUIRE(current.totalCoins <= INT64_MAX - netChange);
+        }
+        current.totalCoins += netChange;
+        ltx.commit();
+    }
     return entries;
 }
 
@@ -99,7 +108,7 @@ updateBalances(std::vector<LedgerEntry> const& entries, Application& app)
     std::uniform_int_distribution<int64_t> dist(totalCoins - coinsAboveReserve,
                                                 INT64_MAX);
     int64_t newTotalCoins = dist(gRandomEngine);
-    return updateBalances(entries, app, newTotalCoins - totalCoins);
+    return updateBalances(entries, app, newTotalCoins - totalCoins, true);
 }
 
 TEST_CASE("Total coins change without inflation",
@@ -195,7 +204,7 @@ TEST_CASE("Account balances unchanged without inflation",
             REQUIRE(!store(*app, updates));
         }
 
-        auto entries2 = updateBalances(entries1, *app, 0);
+        auto entries2 = updateBalances(entries1, *app, 0, false);
         {
             auto updates = makeUpdateList(entries2, entries1);
             REQUIRE(store(*app, updates));
@@ -205,7 +214,7 @@ TEST_CASE("Account balances unchanged without inflation",
         std::vector<LedgerEntry> entries3(entries2.begin(), keepEnd);
         std::vector<LedgerEntry> toDelete(keepEnd, entries2.end());
         int64_t balanceToDelete = getTotalBalance(toDelete);
-        auto entries4 = updateBalances(entries3, *app, balanceToDelete);
+        auto entries4 = updateBalances(entries3, *app, balanceToDelete, false);
         {
             auto updates = makeUpdateList(entries4, entries3);
             auto updates2 = makeUpdateList(nullptr, toDelete);
@@ -273,7 +282,7 @@ TEST_CASE("Inflation changes are consistent",
                               InvariantDoesNotHold);
         }
 
-        auto entries2 = updateBalances(entries1, *app, inflationAmount);
+        auto entries2 = updateBalances(entries1, *app, inflationAmount, true);
         {
             LedgerTxn ltx(app->getLedgerTxnRoot());
             ltx.loadHeader().current().feePool += deltaFeePool;
--- a/src/test/TestAccount.h
+++ b/src/test/TestAccount.h
@@ -132,6 +132,11 @@ class TestAccount
     nextSequenceNumber()
     {
         updateSequenceNumber();
+        if (mSn == std::numeric_limits<SequenceNumber>::max())
+        {
+            throw std::runtime_error(
+                "Sequence number overflow in test account");
+        }
         return ++mSn;
     }
     SequenceNumber loadSequenceNumber();
--- a/src/transactions/test/BumpSequenceTests.cpp
+++ b/src/transactions/test/BumpSequenceTests.cpp
@@ -51,7 +51,12 @@ TEST_CASE("bump sequence", "[tx][bumpsequence]")
                 REQUIRE(a.loadSequenceNumber() == newSeq);
                 SECTION("no more tx when INT64_MAX is reached")
                 {
-                    REQUIRE_THROWS_AS(a.pay(root, 1), ex_txBAD_SEQ);
+                    REQUIRE_THROWS_AS(
+                        applyTx(
+                            {a.tx({payment(root, 1)},
+                                  std::numeric_limits<SequenceNumber>::min())},
+                            *app),
+                        ex_txBAD_SEQ);
                 }
             }
             SECTION("backward jump (no-op)")
--- a/src/transactions/test/ExchangeTests.cpp
+++ b/src/transactions/test/ExchangeTests.cpp
@@ -745,8 +745,9 @@ TEST_CASE("ExchangeV10", "[exchange]")
             else
             {
                 REQUIRE(res.wheatStays ==
-                        (maxWheatSend * p.n >
-                         std::min(maxSheepSend * p.d, maxWheatReceive * p.n)));
+                        bigMultiply(maxWheatSend, p.n) >
+                            std::min(bigMultiply(maxSheepSend, p.d),
+                                     bigMultiply(maxWheatReceive, p.n)));
             }
             REQUIRE(res.numWheatReceived == wheatReceive);
             REQUIRE(res.numSheepSend == sheepSend);


src/test/TestAccount.h
```diff
@@ -132,6 +132,10 @@ class TestAccount
     nextSequenceNumber()
     {
         updateSequenceNumber();
+        if (mSn == std::numeric_limits<SequenceNumber>::max())
```
MonsieurNicolas, line 135: why not throw instead? This seems to hide a bug in whatever test does this

src/invariant/test/ConservationOfLumensTests.cpp
```diff
@@ -202,6 +217,17 @@ TEST_CASE("Account balances unchanged without inflation",
         }
 
         auto keepEnd = entries2.begin() + N / 2;
+
+        // We're going to increase LedgeHeader::totalCoins by the total balance
+        // of the entries starting at `keepEnd` when we call `updateBalances()`
+        // below, so check whether that would overflow, and if so, bump
+        // `keepEnd` until it wouldn't overflow.
+        while (getTotalBalance(std::vector<LedgerEntry>(
```
MonsieurNicolas, line 225: This is a "garbage in, garbage out" situation. The problem is that we feed bad data to the tests in the first place.

Code like `std::generate_n(std::back_inserter(entries1), N, std::bind(generateRandomAccount, 2))` is not good enough in this context:
we have a lot of places where we assume that the sum of all native assets is less than `INT64_MAX`, as a consequence it's just better to generate data that at least meets that constraint.

Maybe pick balances sequentially over the "max left" (starting with `1..INT64_max`)?

src/invariant/test/ConservationOfLumensTests.cpp
```diff
@@ -80,11 +80,26 @@ updateBalances(std::vector<LedgerEntry> entries, Application& app,
     REQUIRE(initialCoins + netChange == finalCoins);
 
     LedgerTxn ltx(app.getLedgerTxnRoot());
-    ltx.loadHeader().current().totalCoins += netChange;
+    auto& current = ltx.loadHeader().current();
```
MonsieurNicolas, line 83: see my other comment: if we have UB here that means the overall data is just bad.
`totalCoins < 0` indicates that there is something really wrong

MonsieurNicolas (changes requested):

src/invariant/test/ConservationOfLumensTests.cpp
jonjove, line 225: I think this works, but I don't think it fundamentally solves the problem. The root cause of this bug is that we are misusing `getTotalBalance`.

What the test is trying to do is delete approximately half the entries, transferring the balance randomly to other entries that are not being deleted. This doesn't actually change `totalCoins`. But `getTotalBalance` does change `totalCoins`.

jonjove (changes requested):

src/test/TestAccount.h
rokopt: It's the "large bump" BumpSequence test, as mentioned in the description.  This is what it's expecting:

```
            SECTION("large bump")
            {
                auto newSeq = INT64_MAX;
                a.bumpSequence(newSeq);
                REQUIRE(a.loadSequenceNumber() == newSeq);
                SECTION("no more tx when INT64_MAX is reached")
                {
                    REQUIRE_THROWS_AS(a.pay(root, 1), ex_txBAD_SEQ);
                }
            }
```

So I can make that pass by throwing `txtest::ex_txBAD_SEQ{}` in the code above, although it would mean that the test wasn't testing what it was trying to anymore -- I think it wants to go through `TestAccount::pay()` -> `applyTx()` -> `throwIf()` to confirm that the transaction containing the payment operation fails with `txBAD_SEQ`.  Specifically I think it's trying ultimately to exercise this:

```
bool
TransactionFrame::isBadSeq(int64_t seqNum) const
{
    return seqNum == INT64_MAX || seqNum + 1 != getSeqNum();
}
```

If we `throw` in `nextSequenceNumber()` then I think this test will just be testing test code.

`TestAccount::tx()` does have an optional `SequenceNumber` parameter; maybe we should just use it to create a `TransactionFrame` directly in the `"large bump"` and call `applyTx()` on it ourselves?

src/invariant/test/ConservationOfLumensTests.cpp
```diff
@@ -80,11 +80,26 @@ updateBalances(std::vector<LedgerEntry> entries, Application& app,
...
+    if ((current.totalCoins < 0 && netChange < 0) ||
+        (current.totalCoins > 0 && netChange > 0))
+    {
+        CAPTURE(current.totalCoins);
```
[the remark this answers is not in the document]
rokopt: I did, in case there were other tests failing in the same way -- `CAPTURE()` only reports those values if the test fails.  However, I'm not attached to them; I can delete them while doing the other simplification deriving from `totalCoins` being non-negative.

src/invariant/test/ConservationOfLumensTests.cpp
rokopt: > Code like std::generate_n(std::back_inserter(entries1), N, std::bind(generateRandomAccount, 2)) is not good enough in this context:
we have a lot of places where we assume that the sum of all native assets is less than INT64_MAX, as a consequence it's just better to generate data that at least meets that constraint.

I think the accounts do meet that constraint -- `generateRandomAccount()`, despite the name, always sets the account balance to `0`, and `updateBalances()` checks `totalCoins` to limit the amount of asset that it can distribute to the accounts in total.

I think Jon nailed the root cause below:

> The root cause of this bug is that we are misusing getTotalBalance.
>
> What the test is trying to do is delete approximately half the entries, transferring the balance randomly to other entries that are not being deleted. This doesn't actually change totalCoins. But getTotalBalance does change totalCoins.

rokopt: > The root cause of this bug is that we are misusing getTotalBalance.
>
> What the test is trying to do is delete approximately half the entries, transferring the balance randomly to other entries that are not being deleted. This doesn't actually change totalCoins. But getTotalBalance does change totalCoins.

I think you mean `updateBalances`, not `getTotalBalance`.

Maybe we should just give the three-argument version of `updateBalances()` a fourth, `bool`, argument telling it whether to update `totalCoins` or not?  The other callers _do_ mean to be updating `totalCoins`.  Crude, maybe, but simple.

src/test/TestAccount.h
rokopt: > TestAccount::tx() does have an optional SequenceNumber parameter; maybe we should just use it to create a TransactionFrame directly in the "large bump" and call applyTx() on it ourselves?

This looks reasonable to me and fixes the test, so I've done it.

src/invariant/test/ConservationOfLumensTests.cpp
rokopt: (Also see below for the fix.)

src/invariant/test/ConservationOfLumensTests.cpp
rokopt: I've added a `REQUIRE()` to that effect and simplified the rest of the test correspondingly.

src/transactions/test/ExchangeTests.cpp
```diff
@@ -744,9 +744,16 @@ TEST_CASE("ExchangeV10", "[exchange]")
             }
             else
             {
-                REQUIRE(res.wheatStays ==
-                        (maxWheatSend * p.n >
-                         std::min(maxSheepSend * p.d, maxWheatReceive * p.n)));
+                REQUIRE(maxWheatSend >= 0);
```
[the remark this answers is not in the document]
rokopt: Ah, much nicer.  Thanks.  I'm using `bigMultiply` now.

src/invariant/test/ConservationOfLumensTests.cpp
```diff
@@ -80,11 +80,26 @@ updateBalances(std::vector<LedgerEntry> entries, Application& app,
     REQUIRE(initialCoins + netChange == finalCoins);
 
     LedgerTxn ltx(app.getLedgerTxnRoot());
-    ltx.loadHeader().current().totalCoins += netChange;
+    auto& current = ltx.loadHeader().current();
+    if ((current.totalCoins < 0 && netChange < 0) ||
```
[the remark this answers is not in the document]
rokopt: Done.

src/invariant/test/ConservationOfLumensTests.cpp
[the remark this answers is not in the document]
rokopt: I've removed them along with the other simplifications.

src/invariant/test/ConservationOfLumensTests.cpp
rokopt: > Maybe we should just give the three-argument version of updateBalances() a fourth, bool, argument telling it whether to update totalCoins or not? The other callers do mean to be updating totalCoins. Crude, maybe, but simple.

I think this works and it doesn't look completely appalling, so I've tried it.

jonjove: Yes, I definitely meant `updateBalances`!

jonjove:
This looks fine to me. @MonsieurNicolas feel free to merge if your comments are resolved to your satisfaction.

src/transactions/test/BumpSequenceTests.cpp
```diff
@@ -51,7 +51,9 @@ TEST_CASE("bump sequence", "[tx][bumpsequence]")
                 REQUIRE(a.loadSequenceNumber() == newSeq);
                 SECTION("no more tx when INT64_MAX is reached")
                 {
-                    REQUIRE_THROWS_AS(a.pay(root, 1), ex_txBAD_SEQ);
+                    REQUIRE_THROWS_AS(
+                        applyTx({a.tx({payment(root, 1)}, newSeq)}, *app),
```
MonsieurNicolas, line 55: This is not right: passing the same sequence number will never work.

The test is just misleading. What we want to do here is ensure that we don't "wrap" sequence numbers (we used to use unsigned ints for sequence numbers).
So a reasonable candidate for this test is `std::numeric_limits<SequenceNumber>::min()` - assuming 2s complements etc which is the only thing we can really do.

src/test/TestAccount.h
MonsieurNicolas: yes fixing the test is the way to go

MonsieurNicolas (changes requested):
looks good - minor change needed in a test

src/transactions/test/BumpSequenceTests.cpp
MonsieurNicolas: ah nm, passing `newSeq` is the right thing to do. It was the old behavior that was incorrect

MonsieurNicolas: nope checking again, this is testing that we're doing the first part of this condition (ie that core doesn't wrap around), so we have to do what I posted above.

```
bool
TransactionFrame::isBadSeq(int64_t seqNum) const
{
    return seqNum == INT64_MAX || seqNum + 1 != getSeqNum();
}
```

rokopt: > The test is just misleading. What we want to do here is ensure that we don't "wrap" sequence numbers (we used to use unsigned ints for sequence numbers).
>So a reasonable candidate for this test is std::numeric_limits<SequenceNumber>::min() - assuming 2s complements etc which is the only thing we can really do.

It has certainly misled me!  Thanks for the suggestion.  I've made that change.

MonsieurNicolas:
r+ adecddad268adc6adea8cdc15b3406359c1e87d4

PR #2941, opened 2021-03-02
Improve naming of externalize metrics

### LINKED ISSUE #2864: Improve naming of externalize metrics
We have several metrics with similar names that are easy to mix-up. Specifically, 
* `scp.timing.externalize-lag` - local node's lag compared to the first node that externalized (could be the same node)
* `scp.timing.externalize-delay` - Other nodes' lag compared to the the local node's externalized time
* `qset_lag` reported by the quorum endpoint - Other nodes' lag compared to the first node that externalized (could be the same as `scp.timing.externalize-delay` if the local node is consistently externalizing first)

We should rename these for clarity.

marta-lokhova:
Resolves #2864

### DIFF
--- a/docs/metrics.md
+++ b/docs/metrics.md
@@ -126,8 +126,8 @@ scp.timeout.nominate                     | meter     | timeouts in nomination
 scp.timeout.prepare                      | meter     | timeouts in ballot protocol
 scp.timing.nominated                     | timer     | time spent in nomination
 scp.timing.externalized                  | timer     | time spent in ballot protocol
-scp.timing.externalize-lag               | timer     | delay between first externalize message and local node externalizing
-scp.timing.externalize-delay             | timer     | delay between local node externalizing and later externalize messages
+scp.timing.first-to-self-externalize-lag | timer     | delay between first externalize message and local node externalizing
+scp.timing.self-to-others-externalize-lag| timer     | delay between local node externalizing and later externalize messages from other nodes
 scp.value.invalid                        | meter     | SCP value is invalid
 scp.value.valid                          | meter     | SCP value is valid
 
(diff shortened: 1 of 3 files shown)


docs/metrics.md
```diff
@@ -126,8 +126,8 @@ scp.timeout.nominate                     | meter     | timeouts in nomination
 scp.timeout.prepare                      | meter     | timeouts in ballot protocol
 scp.timing.nominated                     | timer     | time spent in nomination
 scp.timing.externalized                  | timer     | time spent in ballot protocol
-scp.timing.externalize-lag               | timer     | delay between first externalize message and local node externalizing
-scp.timing.externalize-delay             | timer     | delay between local node externalizing and later externalize messages
+scp.timing.self-to-first-externalize-lag | timer     | delay between first externalize message and local node externalizing
```
MonsieurNicolas, line 129: `self-to-` sounds like self is *before*. So maybe this should be `self-from-first-externalize-lag` ?

docs/metrics.md
```diff
@@ -126,8 +126,8 @@ scp.timeout.nominate                     | meter     | timeouts in nomination
...
+scp.timing.rest-to-self-externalize-lag  | timer     | delay between local node externalizing and later externalize messages from other nodes
```
MonsieurNicolas, line 130: this one also is not super clear. Maybe `self-to-other-externalize-lag` *or* `other-from-self-externalize-lag`

MonsieurNicolas (changes requested):

docs/metrics.md
MonsieurNicolas: actually using the "from" variant is bad, we should just use "to" so in this case `first-to-self-externalize-lag`

docs/metrics.md
MonsieurNicolas: same than earlier "to" reads more naturally I think

marta-lokhova:
updated

MonsieurNicolas:
r+ 5636e8d1d135985b65b1d1ea0633f89dee7a7b38

PR #2944, opened 2021-03-04
Best-offer debugging

rokopt:
# Best-offer debugging

Updates and integrates some formerly-private debugging for `getBestOffer()` which cross-checks its return values against a much slower, simpler algorithm.

Because the debug algorithm is so slow, it is compiled in by default only when tests are compiled in (or when explicitly requested with a `configure` option), and if is compiled in, it runs by default only under the `test` and `fuzz` commands (or when explicitly requested with a command-line option; it can also be disabled for `test` and `fuzz` with a command-line option).

# Checklist
- [x] Compiles
- [x] Ran all tests

### DIFF
--- a/src/ledger/InMemoryLedgerTxnRoot.cpp
+++ b/src/ledger/InMemoryLedgerTxnRoot.cpp
@@ -147,4 +147,21 @@ InMemoryLedgerTxnRoot::resetForFuzzer()
     abort();
 }
 #endif // FUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION
+
+#ifdef BEST_OFFER_DEBUGGING
+bool
+InMemoryLedgerTxnRoot::bestOfferDebuggingEnabled() const
+{
+    return mBestOfferDebuggingEnabled;
+}
+
+std::shared_ptr<LedgerEntry const>
+InMemoryLedgerTxnRoot::getBestOfferSlow(Asset const& buying,
+                                        Asset const& selling,
+                                        OfferDescriptor const* worseThan,
+                                        std::unordered_set<int64_t>& exclude)
+{
+    return nullptr;
+}
+#endif
 }
--- a/src/ledger/InMemoryLedgerTxnRoot.cpp
+++ b/src/ledger/InMemoryLedgerTxnRoot.cpp
@@ -10,8 +10,15 @@
 namespace stellar
 {
 
-InMemoryLedgerTxnRoot::InMemoryLedgerTxnRoot()
+InMemoryLedgerTxnRoot::InMemoryLedgerTxnRoot(
+#ifdef BEST_OFFER_DEBUGGING
+    bool bestOfferDebuggingEnabled
+#endif
+    )
     : mHeader(std::make_unique<LedgerHeader>())
+#ifdef BEST_OFFER_DEBUGGING
+    , mBestOfferDebuggingEnabled(bestOfferDebuggingEnabled)
+#endif
 {
 }
 
--- a/src/ledger/LedgerTxn.cpp
+++ b/src/ledger/LedgerTxn.cpp
@@ -80,6 +80,29 @@ AssetPairHash::operator()(AssetPair const& key) const
     return hashAsset(key.buying) ^ (hashAsset(key.selling) << 1);
 }
 
+#ifdef BEST_OFFER_DEBUGGING
+void
+compareOffers(std::shared_ptr<LedgerEntry const> debugBest,
+              std::shared_ptr<LedgerEntry const> best)
+{
+    if ((bool)debugBest ^ (bool)best)
+    {
+        if (best)
+        {
+            printErrorAndAbort("best offer not null when it should be");
+        }
+        else
+        {
+            printErrorAndAbort("best offer null when it should not be");
+        }
+    }
+    if (best && *best != *debugBest)
+    {
+        printErrorAndAbort("best offer mismatch");
+    }
+}
+#endif
+
 // Implementation of AbstractLedgerTxnParent --------------------------------
 AbstractLedgerTxnParent::~AbstractLedgerTxnParent()
 {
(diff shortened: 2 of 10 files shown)


src/test/FuzzerImpl.cpp
```diff
@@ -982,7 +982,11 @@ void
 TransactionFuzzer::initialize()
 {
     resetRandomSeed(1);
-    mApp = createTestApplication(mClock, getFuzzConfig(0));
+    auto cfg = getFuzzConfig(0);
+#ifdef BEST_OFFER_DEBUGGING
+    cfg.BEST_OFFER_DEBUGGING_ENABLED = mBestOfferDebuggingEnabled;
```
jonjove, line 987: This should always be `true` if `BEST_OFFER_DEBUGGING` is defined, no need to actually support toggling it from the command line.

src/test/test.cpp
```diff
@@ -191,6 +195,10 @@ getTestConfig(int instanceNumber, Config::TestDbMode mode)
         // only spin up a small number of worker threads
         thisConfig.WORKER_THREADS = 2;
         thisConfig.QUORUM_INTERSECTION_CHECKER = false;
+
+#ifdef BEST_OFFER_DEBUGGING
+        thisConfig.BEST_OFFER_DEBUGGING_ENABLED = !gBestOfferDebuggingDisabled;
```
jonjove, line 200: If we're running tests this should always be `true`. No need for command line configuration.

src/ledger/LedgerTxn.cpp
```diff
@@ -1980,13 +1980,16 @@ LedgerTxn::Impl::WorstBestOfferIteratorImpl::clone() const
 size_t const LedgerTxnRoot::Impl::MIN_BEST_OFFERS_BATCH_SIZE = 5;
 
 LedgerTxnRoot::LedgerTxnRoot(Database& db, size_t entryCacheSize,
-                             size_t prefetchBatchSize)
-    : mImpl(std::make_unique<Impl>(db, entryCacheSize, prefetchBatchSize))
+                             size_t prefetchBatchSize,
+                             bool bestOfferDebuggingEnabled)
```
jonjove, line 2106: Nitpick: there should be no mention of best offer debugging if it isn't compiled in. That will make it obvious that the behavior is unchanged if the flag isn't set, and prevent warnings about unused variables.

src/main/CommandLine.cpp
```diff
@@ -1179,6 +1210,9 @@ runGenerateOrSimulateTxs(CommandLineArgs const& args, bool generate)
     // Default to no simulated transactions, just real data
     uint32_t multiplier = 1;
     bool verifyResults = false;
+#ifdef BEST_OFFER_DEBUGGING
```
jonjove, line 1221: No need for runtime configuration in this path either.

src/test/FuzzerImpl.cpp
rokopt: Done.

src/test/test.cpp
rokopt: Done.

src/main/CommandLine.cpp
rokopt: Done.

src/ledger/LedgerTxn.cpp
rokopt: Done.

rokopt:
To try to get some idea of the impact of this on test speed, I ran `./src/stellar-core test --all-versions --ll info -w NoTests -a -r simple "[tx]"` (I'm guessing the transaction tests would include the most impacted) with and without best-offer debugging compiled in:

`--enable-extrachecks --enable-asan --disable-best-offer-debugging`: 7m55s
`--enable-extrachecks --enable-asan --enable-best-offer-debugging`: 7m57s

I'm guessing that's well within the variation (and even if it weren't, it's probably an irrelevant difference for a unit test), so I think we can indeed comfortably afford not to provide a `--disable-best-offer-debugging` for `stellar-core test`.

src/ledger/LedgerTxn.cpp
```diff
@@ -646,10 +646,130 @@ LedgerTxn::Impl::getAllOffers()
...
+        if (kv.second)
+        {
+            auto const& le = kv.second->ledgerEntry();
+            if (!(le.data.offer().buying == buying &&
+                  le.data.offer().selling == selling))
+            {
+                continue;
+            }
+
+            if (worseThan && !isBetterOffer(*worseThan, le))
```
sisuresh, line 703: Nit: Would this be easier to read as `isBetterOffer(le, *worseThan)`? Same for `LedgerTxnRoot::Impl::
getBestOfferSlow`.

sisuresh (commented):
I reviewed the last commit and the logic looks good to me. Just included a few minor comments.

src/ledger/LedgerTxn.cpp
jonjove: I don't think these are equivalent. If `*worseThan == le` then 
```
!isBetterOffer(*worseThan, le) == !isBetterOffer(le, *worseThan) != isBetterOffer(le, *worseThan)
```
I didn't look at all the context, so if that condition isn't possible in this scenario then the change might be acceptable.

src/ledger/LedgerTxn.cpp
```diff
@@ -2484,17 +2624,106 @@ LedgerTxnRoot::Impl::getAllOffers()
...
+        }
+        else
+        {
+            printErrorAndAbort("best offer null when it should not be");
+        }
+    }
+    if (best && *best != *debugBest)
+    {
+        printErrorAndAbort("best offer mismatch");
+    }
```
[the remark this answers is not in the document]
rokopt: Good idea.  Done.

src/ledger/LedgerTxnImpl.h
```diff
@@ -586,6 +586,20 @@ class LedgerTxn::Impl
...
+    bool bestOfferDebuggingEnabled() const;
+
+    std::shared_ptr<LedgerEntry const>
+    getBestOfferSlow(Asset const& buying, Asset const& selling,
+                     OfferDescriptor const* worseThan,
+                     std::unordered_set<int64_t>& exclude);
+#endif
+
+    std::shared_ptr<LedgerEntry const>
+    checkBestOffer(Asset const& buying, Asset const& selling,
```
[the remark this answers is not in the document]
rokopt: I like that too.  Done.

src/ledger/LedgerTxn.cpp
rokopt: It's not obvious to me why that wouldn't be possible, so I'm inclined to leave it as it is unless someone else has a clear demonstration that it isn't possible.

sisuresh: Ah yeah nevermind. I was wrong.

rokopt:
> To try to get some idea of the impact of this on test speed, I ran `./src/stellar-core test --all-versions --ll info -w NoTests -a -r simple "[tx]"` (I'm guessing the transaction tests would include the most impacted) with and without best-offer debugging compiled in:
> 
> `--enable-extrachecks --enable-asan --disable-best-offer-debugging`: 7m55s
> `--enable-extrachecks --enable-asan --enable-best-offer-debugging`: 7m57s
> 
> I'm guessing that's well within the variation (and even if it weren't, it's probably an irrelevant difference for a unit test), so I think we can indeed comfortably afford not to provide a `--disable-best-offer-debugging` for `stellar-core test`.

I also tried it on a catchup.  The range I chose took ~5.5 minutes without best-offer debugging, and ~45 minutes with it.

MonsieurNicolas:
Is the configure option needed now that you have the config? Why not just enable it if people enable extra-checks?

rokopt:
> Is the configure option needed now that you have the config? Why not just enable it if people enable extra-checks?

Extrachecks are off unless `--enable-extrachecks` is explicitly specified, even if tests are being built.  I think we were intending that tests would build and run with best-offer debugging unless someone explicitly requested otherwise.  (Just keying best-offer debugging off `BUILD_TESTS` is an option.)

rokopt:
Nicolas requested the removal of the separate `configure` option, `--enable-best-offer-debugging`, in favor of using `--enable-extrachecks` to determine whether to set `-DBEST_OFFER_DEBUGGING`.

He also requested removing the `--enable-best-offer-debugging` option from `run` and `catchup`, as if anyone needs such detailed debugging for those commands, they'll probably be modifying code and rebuilding anyway.

I've pushed commits to do each of those two things.

src/ledger/InMemoryLedgerTxnRoot.cpp
```diff
@@ -10,8 +10,15 @@
 namespace stellar
 {
 
-InMemoryLedgerTxnRoot::InMemoryLedgerTxnRoot()
+InMemoryLedgerTxnRoot::InMemoryLedgerTxnRoot(
+#if BEST_OFFER_DEBUGGING
```
MonsieurNicolas, line 14: do not mix `#if BEST_OFFER_DEBUGGING` and `#ifdef BEST_OFFER_DEBUGGING`
What you want everywhere is `#ifdef BEST_OFFER_DEBUGGING` given how you set it in the configure script

src/test/FuzzerImpl.cpp
```diff
@@ -983,7 +983,11 @@ void
 TransactionFuzzer::initialize()
 {
     resetRandomSeed(1);
-    mApp = createTestApplication(mClock, getFuzzConfig(0));
+    auto cfg = getFuzzConfig(0);
+#ifdef BEST_OFFER_DEBUGGING
```
MonsieurNicolas, line 987: you don't need this: as you're using a test configuration, it's already set

MonsieurNicolas (changes requested):
cool. you should squash those last 2 commits into the previous commits as they're (mostly) cancelling things that this PR added

rokopt:
> cool. you should squash those last 2 commits into the previous commits as they're (mostly) cancelling things that this PR added

Squashed.  (The previous two were also squashable.)

src/test/FuzzerImpl.cpp
rokopt: Good point.  Removed.

src/ledger/InMemoryLedgerTxnRoot.cpp
rokopt: Yikes, thanks for catching that. Fixed.

src/ledger/LedgerTxn.cpp
```diff
@@ -646,10 +669,120 @@ LedgerTxn::Impl::getAllOffers()
...
+        return isBetterOffer(selfBest->ledgerEntry(), *parentBest)
+                   ? std::make_shared<LedgerEntry const>(
+                         selfBest->ledgerEntry())
+                   : std::make_shared<LedgerEntry const>(*parentBest);
+    }
+    return nullptr;
+}
+#endif
+
+#ifdef BEST_OFFER_DEBUGGING
```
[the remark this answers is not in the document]
rokopt: Done.

rokopt:
Fuzz metrics from current master without this PR:

master-without-order-book-not-crossed-invariant-for-20-min

And with this PR:

best-offer-debugging-after-20-minutes

So the impact of this on exec speed is very substantial, although the benefits in terms of noticing bugs that fuzzing might find could also be large.  (That all applies to https://github.com/stellar/stellar-core/pull/2959 as well.)

MonsieurNicolas:
r+ ae3dfa09115d650a18566aaa405a9a42dbdb7e2e

PR #2959, opened 2021-03-11
Introduce "order book not crossed" invariant (for fuzzing only)

### LINKED ISSUE #2417: cleanup and merge orderbook not crossed invariant
https://github.com/stellar/stellar-core/pull/2210

rokopt:
# Introduce "order book not crossed" invariant

Resolves #2417 

This PR introduces a new invariant which keeps track of the state of the order book as operations are applied, and asserts that it is not crossed.

There is currently no mechanism by which an invariant can be informed that an operation application for which its `checkOnOperationApply()` method had been called is being rolled back.  Hence, we can not in general support an invariant such as this which maintains state across method calls.  But we can when fuzzing, where setup operations are never rolled back and where we can explicitly tell invariants when to snapshot state (that is, after setup completes) and when to reset it to that state (after each fuzz test).

Therefore, this invariant is compiled in when fuzzing is configured, and not otherwise.

This PR updates #2210 to current code, adapting to some changes such as the introduction of `InternalLedgerEntry` and the rearranging of the fuzzing code, and adds support for passive offers.

# Checklist
- [x] Compiles
- [x] Ran all tests

### DIFF
--- a/src/invariant/OrderBookIsNotCrossed.cpp
+++ b/src/invariant/OrderBookIsNotCrossed.cpp
@@ -1,0 +1,230 @@
+#ifdef BUILD_TESTS
+// Copyright 2021 Stellar Development Foundation and contributors. Licensed
+// under the Apache License, Version 2.0. See the COPYING file at the root
+// of this distribution or at http://www.apache.org/licenses/LICENSE-2.0
+
+#include "invariant/OrderBookIsNotCrossed.h"
+#include "invariant/InvariantManager.h"
+#include "ledger/InternalLedgerEntry.h"
+#include "ledger/LedgerTxn.h"
+#include "main/Application.h"
+#include "util/Logging.h"
+#include "util/XDROperators.h"
+#include "xdr/Stellar-ledger-entries.h"
+#include <fmt/format.h>
+#include <functional>
+#include <xdrpp/printer.h>
+
+namespace stellar
+{
+namespace
+{
+double
+priceAsDouble(Price const& price)
+{
+    return static_cast<double>(price.n) / static_cast<double>(price.d);
+}
+
+OrderBookIsNotCrossed::AssetPairSet
+extractAssetPairs(LedgerTxnDelta const& ltxd)
+{
+    OrderBookIsNotCrossed::AssetPairSet assets;
+    for (auto const& entry : ltxd.entry)
+    {
+        if (entry.first.type() == InternalLedgerEntryType::LEDGER_ENTRY &&
+            entry.first.ledgerKey().type() == OFFER && entry.second.current)
+        {
+            auto const& offer =
+                entry.second.current->ledgerEntry().data.offer();
+            auto const assetPair =
+                offer.selling < offer.buying
+                    ? std::make_pair(offer.selling, offer.buying)
+                    : std::make_pair(offer.buying, offer.selling);
+
+            assets.insert(assetPair);
+        }
+    }
+
+    return assets;
+}
+
+std::string
+checkCrossed(Asset const& a, Asset const& b,
+             OrderBookIsNotCrossed::OrderBook const& orderBook)
+{
+    // If either side of the order book for this asset pair is empty or does not
+    // yet exist, then the order book cannot be crossed.
+    auto const iterAB = orderBook.find({a, b});
+    auto const iterBA = orderBook.find({b, a});
+    if (iterAB == orderBook.end() || iterBA == orderBook.end())
+    {
+        return {};
+    }
+
+    auto const& asks = iterAB->second;
+    auto const& bids = iterBA->second;
+    if (asks.empty() || bids.empty())
+    {
+        return {};
+    }
+
+    // check if crossed
+    auto lowestAsk = priceAsDouble(asks.cbegin()->price);
+    auto highestBidInverse = bids.cbegin()->price;
+    auto highestBid =
+        priceAsDouble(Price{highestBidInverse.d, highestBidInverse.n});
+
+    if (lowestAsk <= highestBid)
+    {
+        if (lowestAsk == highestBid)
+        {
+            // We ordered non-passive offers before passive offers so that if
+            // the first offer at this price in the list is passive, then we
+            // know that all offers at this price in the list are passive.
+            if (((asks.cbegin()->flags & PASSIVE_FLAG) != 0) ||
+                ((bids.cbegin()->flags & PASSIVE_FLAG) != 0))
+            {
+                // In at least one of the lists, all the offers at this price
+                // are passive, so the equal-price crossing does not represent a
+                // bug.
+                return {};
+            }
+
+            // There is at least one non-passive offer in each list, so at least
+            // one of them should have crossed with an offer on the other list,
+            // so we fall through to the invariant failure.
+        }
+
+        return fmt::format(
+            FMT_STRING(
+                "Order book is in a crossed state for {} - {} "
+                "asset pair.\nTop of the book is:\n\tAsk price: {}\n\tBid "
+                "price: {}\n\nWhere {} >= {}!"),
+            xdr::xdr_to_string(a), xdr::xdr_to_string(b), lowestAsk, highestBid,
+            highestBid, lowestAsk);
+    }
+    return {};
+}
+}
+
+bool
+OrderBookIsNotCrossed::OfferEntryCmp::operator()(OfferEntry const& a,
+                                                 OfferEntry const& b) const
+{
+    auto const priceA = priceAsDouble(a.price);
+    auto const priceB = priceAsDouble(b.price);
+
+    if (priceA < priceB)
+    {
+        return true;
+    }
+    if (priceA == priceB)
+    {
+        // We order non-passive offers before passive offers so that we
+        // can find any non-passive offer at a given price at the front
+        // of the list of offers at that price.
+        if ((a.flags & PASSIVE_FLAG) != (b.flags & PASSIVE_FLAG))
+        {
+            return (a.flags & PASSIVE_FLAG) == 0;
+        }
+        return a.offerID < b.offerID;
+    }
+    return false;
+}
+
+void
+OrderBookIsNotCrossed::updateOrderBook(LedgerTxnDelta const& ltxd)
+{
+    for (auto const& entry : ltxd.entry)
+    {
+        // there are three possible "deltas" for an offer:
+        //      CREATED:  (nil)     -> LedgerKey
+        //      MODIFIED: LedgerKey -> LedgerKey
+        //      DELETED:  LedgerKey -> (nil)
+        if (entry.first.type() == InternalLedgerEntryType::LEDGER_ENTRY &&
+            entry.first.ledgerKey().type() == OFFER)
+        {
+            if (mRestoreBeforeNextUpdate)
+            {
+                mOrderBook = mOrderBookSnapshot;
+                mRestoreBeforeNextUpdate = false;
+            }
+            if (entry.second.previous)
+            {
+                auto const& oe =
+                    entry.second.previous->ledgerEntry().data.offer();
+                mOrderBook[{oe.selling, oe.buying}].erase(oe);
+            }
+            if (entry.second.current)
+            {
+                auto const& oe =
+                    entry.second.current->ledgerEntry().data.offer();
+                mOrderBook[{oe.selling, oe.buying}].emplace(oe);
+            }
+        }
+    }
+}
+
+std::string
+OrderBookIsNotCrossed::check(AssetPairSet const& assetPairs)
+{
+    assert(!mRestoreBeforeNextUpdate);
+    for (auto const& assetPair : assetPairs)
+    {
+        auto checkCrossedResult =
+            checkCrossed(assetPair.first, assetPair.second, mOrderBook);
+        if (!checkCrossedResult.empty())
+        {
+            return checkCrossedResult;
+        }
+    }
+
+    return std::string{};
+}
+
+std::shared_ptr<Invariant>
+OrderBookIsNotCrossed::registerAndEnableInvariant(Application& app)
+{
+    auto invariant =
+        app.getInvariantManager().registerInvariant<OrderBookIsNotCrossed>();
+    app.getInvariantManager().enableInvariant(invariant->getName());
+    return invariant;
+}
+
+std::string
+OrderBookIsNotCrossed::getName() const
+{
+    return "OrderBookIsNotCrossed";
+}
+
+std::string
+OrderBookIsNotCrossed::checkOnOperationApply(Operation const& operation,
+                                             OperationResult const& result,
+                                             LedgerTxnDelta const& ltxDelta)
+{
+    updateOrderBook(ltxDelta);
+    auto assetPairs = extractAssetPairs(ltxDelta);
+    return assetPairs.size() > 0 ? check(assetPairs) : std::string{};
+}
+
+void
+OrderBookIsNotCrossed::snapshotForFuzzer()
+{
+    mOrderBookSnapshot = mOrderBook;
+}
+
+void
+OrderBookIsNotCrossed::resetForFuzzer()
+{
+    // This call indicates that the fuzzer has completed one test and its next
+    // call to `checkOnOperationApply` will be in the context of the order book
+    // state immediately after the completion of setup (`mOrderBookSnapshot`).
+    // However, it's possible that the next one or more fuzz tests won't
+    // generate any operations involving orders, in which case it wouldn't
+    // matter if `mOrderBook` were stale.  Therefore, we defer copying the map
+    // until we need it to be up to date.
+    mOrderBook.clear();
+    mRestoreBeforeNextUpdate = true;
+}
+}
+#endif // BUILD_TESTS
--- a/src/invariant/OrderBookIsNotCrossed.h
+++ b/src/invariant/OrderBookIsNotCrossed.h
@@ -1,0 +1,79 @@
+#ifdef BUILD_TESTS
+#pragma once
+
+// Copyright 2021 Stellar Development Foundation and contributors. Licensed
+// under the Apache License, Version 2.0. See the COPYING file at the root
+// of this distribution or at http://www.apache.org/licenses/LICENSE-2.0
+
+#include "invariant/Invariant.h"
+#include "ledger/LedgerHashUtils.h"
+#include "ledger/LedgerTxn.h"
+#include "main/Application.h"
+#include "xdr/Stellar-ledger-entries.h"
+#include "xdr/Stellar-ledger.h"
+
+#include <set>
+#include <unordered_map>
+
+namespace stellar
+{
+// compare two OfferEntry's by price
+class OrderBookIsNotCrossed : public Invariant
+{
+  public:
+    struct OfferEntryCmp
+    {
+        bool operator()(OfferEntry const& a, OfferEntry const& b) const;
+    };
+
+    using Orders = std::set<OfferEntry, OfferEntryCmp>;
+    using OrderBook = std::unordered_map<AssetPair, Orders, AssetPairHash>;
+    using AssetPairSet = std::set<std::pair<Asset, Asset>>;
+
+    explicit OrderBookIsNotCrossed()
+        : Invariant(true), mRestoreBeforeNextUpdate(false)
+    {
+    }
+
+    // The order-book-not-crossed invariant relies on maintaining state across
+    // checkOnOperationApply() calls, and there is no general mechanism in the
+    // InvariantManager to notify an invariant if an operation was later rolled
+    // back.  Therefore, we can reliably check the order-book-not-crossed
+    // invariant only when either the calling code uses the
+    // snapshotForFuzzer()/resetForFuzzer() mechanism, which currently only the
+    // fuzzer does, or in a test specifically of the OrderBookIsNotCrossed
+    // invariant, which is aware of its maintaining state.  Therefore, we
+    // register and enable the invariant only explicitly, from those places:
+    // fuzzing and the invariant's own tests.
+    static std::shared_ptr<Invariant>
+    registerAndEnableInvariant(Application& app);
+
+    virtual std::string getName() const override;
+
+    virtual std::string
+    checkOnOperationApply(Operation const& operation,
+                          OperationResult const& result,
+                          LedgerTxnDelta const& ltxDelta) override;
+
+    OrderBook const&
+    getOrderBook() const
+    {
+        return mOrderBook;
+    }
+
+    void snapshotForFuzzer() override;
+    void resetForFuzzer() override;
+
+  private:
+    // as of right now, since this is only used in fuzzing, the mOrderBook will
+    // be empty to start. If used in production, since invariants are
+    // configurable, it is likely that this invariant will be enabled with
+    // pre-existing ledger and thus the mOrderBook will need a way to sync
+    OrderBook mOrderBook;
+    OrderBook mOrderBookSnapshot;
+    bool mRestoreBeforeNextUpdate;
+    void updateOrderBook(LedgerTxnDelta const& ltxd);
+    std::string check(AssetPairSet const& assetPairs);
+};
+}
+#endif // BUILD_TESTS
--- a/src/invariant/test/InvariantTestUtils.cpp
+++ b/src/invariant/test/InvariantTestUtils.cpp
@@ -34,6 +34,27 @@ generateRandomAccount(uint32_t ledgerSeq)
     return le;
 }
 
+LedgerEntry
+generateOffer(Asset const& selling, Asset const& buying, int64_t amount,
+              Price price)
+{
+    REQUIRE(!(selling == buying));
+    REQUIRE(amount >= 1);
+
+    LedgerEntry le;
+    le.lastModifiedLedgerSeq = 2;
+    le.data.type(OFFER);
+
+    auto offer = LedgerTestUtils::generateValidOfferEntry();
+    offer.amount = amount;
+    offer.price = price;
+    offer.selling = selling;
+    offer.buying = buying;
+
+    le.data.offer() = offer;
+    return le;
+}
+
 bool
 store(Application& app, UpdateList const& apply, AbstractLedgerTxn* ltxPtr,
       OperationResult const* resPtr)
--- a/src/invariant/test/OrderBookIsNotCrossedTests.cpp
+++ b/src/invariant/test/OrderBookIsNotCrossedTests.cpp
@@ -1,0 +1,376 @@
+#ifdef BUILD_TESTS
+// Copyright 2021 Stellar Development Foundation and contributors. Licensed
+// under the Apache License, Version 2.0. See the COPYING file at the root
+// of this distribution or at http://www.apache.org/licenses/LICENSE-2.0
+
+#include "invariant/OrderBookIsNotCrossed.h"
+#include "invariant/test/InvariantTestUtils.h"
+#include "ledger/LedgerTxn.h"
+#include "lib/catch.hpp"
+#include "test/TestAccount.h"
+#include "test/TestUtils.h"
+#include "test/TxTests.h"
+#include "test/test.h"
+
+namespace stellar
+{
+
+namespace
+{
+void
+applyCheck(Application& app, std::vector<LedgerEntry> const& current,
+           std::vector<LedgerEntry> const& previous, bool shouldPass)
+{
+    stellar::InvariantTestUtils::UpdateList updates;
+
+    if (previous.empty())
+    {
+        updates = InvariantTestUtils::makeUpdateList(current, nullptr);
+    }
+    else if (current.empty())
+    {
+        updates = InvariantTestUtils::makeUpdateList(nullptr, previous);
+    }
+    else
+    {
+        std::vector<LedgerKey> curKeys;
+        std::vector<LedgerKey> prevKeys;
+        std::transform(current.begin(), current.end(),
+                       std::back_inserter(curKeys), LedgerEntryKey);
+        std::transform(previous.begin(), previous.end(),
+                       std::back_inserter(prevKeys), LedgerEntryKey);
+        REQUIRE(curKeys == prevKeys);
+        updates = InvariantTestUtils::makeUpdateList(current, previous);
+    }
+
+    REQUIRE(InvariantTestUtils::store(app, updates, nullptr, nullptr) ==
+            shouldPass);
+}
+
+LedgerEntry
+createOffer(Asset const& ask, Asset const& bid, int64 amount,
+            Price const& price, bool passive = false)
+{
+    auto offer = InvariantTestUtils::generateOffer(ask, bid, amount, price);
+    offer.data.offer().flags = passive ? PASSIVE_FLAG : 0;
+    return offer;
+}
+
+LedgerEntry
+modifyOffer(LedgerEntry offer, bool passive)
+{
+    offer.data.offer().flags = passive ? PASSIVE_FLAG : 0;
+    return offer;
+}
+
+LedgerEntry
+modifyOffer(LedgerEntry offer, int64 amount, Price const& price)
+{
+    offer.data.offer().amount = amount;
+    offer.data.offer().price = price;
+    return offer;
+}
+
+LedgerEntry
+modifyOffer(LedgerEntry offer, Asset const& ask, Asset const& bid, int64 amount,
+            Price const& price)
+{
+    offer.data.offer().selling = ask;
+    offer.data.offer().buying = bid;
+    return modifyOffer(offer, amount, price);
+}
+}
+
+TEST_CASE("Comparison of offers meets ordering requirements",
+          "[invariant][OrderBookIsNotCrossed]")
+{
+    OrderBookIsNotCrossed::OfferEntryCmp cmp;
+    OfferEntry l, r;
+
+    l.price = Price{1, 1};
+    l.flags = PASSIVE_FLAG;
+    r.price = Price{2, 1};
+    r.flags = 0;
+    REQUIRE(cmp(l, r));
+    REQUIRE(!cmp(r, l));
+
+    l.price = Price{1, 1};
+    l.flags = 0;
+    l.offerID = 2;
+    r.price = Price{1, 1};
+    r.flags = PASSIVE_FLAG;
+    r.offerID = 1;
+    REQUIRE(cmp(l, r));
+    REQUIRE(!cmp(r, l));
+}
+
+TEST_CASE("OrderBookIsNotCrossed in-memory order book is consistent with "
+          "application behaviour",
+          "[invariant][OrderBookIsNotCrossed]")
+{
+    VirtualClock clock;
+    auto cfg = getTestConfig(0);
+    // When testing the order book not crossed invariant, enable it and no other
+    // invariants (these tests do things which violate other invariants).
+    cfg.INVARIANT_CHECKS = {};
+    auto app = createTestApplication(clock, cfg);
+    OrderBookIsNotCrossed::registerAndEnableInvariant(*app);
+    app->start();
+
+    auto root = TestAccount::createRoot(*app);
+
+    auto const cur1 = root.asset("CUR1");
+    auto const cur2 = root.asset("CUR2");
+
+    LedgerTxn ltxOuter{app->getLedgerTxnRoot()};
+
+    auto const invariant = std::make_shared<OrderBookIsNotCrossed>();
+    auto offer = InvariantTestUtils::generateOffer(cur1, cur2, 3, Price{3, 2});
+
+    // create
+    {
+        LedgerTxn ltx{ltxOuter};
+        ltx.create(offer);
+
+        invariant->checkOnOperationApply({}, OperationResult{}, ltx.getDelta());
+        auto const& orders = invariant->getOrderBook().at({cur1, cur2});
+
+        REQUIRE(orders.size() == 1);
+
+        auto const& offerToCheck = *orders.cbegin();
+
+        REQUIRE(offerToCheck.amount == 3);
+        REQUIRE(offerToCheck.price.n == 3);
+        REQUIRE(offerToCheck.price.d == 2);
+
+        ltx.commit();
+    }
+
+    // modify
+    {
+        LedgerTxn ltx{ltxOuter};
+
+        offer.data.offer().amount = 2;
+        offer.data.offer().price = Price{5, 3};
+        auto entry = ltx.load(LedgerEntryKey(offer));
+        entry.current() = offer;
+
+        invariant->checkOnOperationApply({}, OperationResult{}, ltx.getDelta());
+        auto const& orders = invariant->getOrderBook().at({cur1, cur2});
+
+        REQUIRE(orders.size() == 1);
+
+        auto const& offerToCheck = *orders.cbegin();
+
+        REQUIRE(offerToCheck.amount == 2);
+        REQUIRE(offerToCheck.price.n == 5);
+
+        ltx.commit();
+    }
+
+    // delete
+    {
+        LedgerTxn ltx{ltxOuter};
+
+        auto entry = ltx.load(LedgerEntryKey(offer));
+        entry.erase();
+
+        invariant->checkOnOperationApply({}, OperationResult{}, ltx.getDelta());
+
+        REQUIRE(invariant->getOrderBook().at({cur1, cur2}).size() == 0);
+    }
+}
+
+TEST_CASE("OrderBookIsNotCrossed properly throws if order book is crossed",
+          "[invariant][OrderBookIsNotCrossed]")
+{
+
+    VirtualClock clock;
+    auto cfg = getTestConfig(0);
+    // When testing the order book not crossed invariant, enable it and no other
+    // invariants (these tests do things which violate other invariants).
+    cfg.INVARIANT_CHECKS = {};
+    auto app = createTestApplication(clock, cfg);
+    OrderBookIsNotCrossed::registerAndEnableInvariant(*app);
+    app->start();
+
+    auto root = TestAccount::createRoot(*app);
+
+    auto const cur1 = root.asset("CUR1");
+    auto const cur2 = root.asset("CUR2");
+    auto const cur3 = root.asset("CUR3");
+    auto const cur4 = root.asset("CUR4");
+    auto const cur5 = root.asset("CUR5");
+
+    // the initial set up for the order book follows where:
+    // A = cur1, B = cur2, and C = cur3
+    // A against B
+    //     offer1: 3 A @  3/2 (B/A)
+    //     offer2: 2 A @  5/3 (B/A)
+    auto const offer1 = createOffer(cur1, cur2, 3, Price{3, 2});
+    auto const offer2 = createOffer(cur1, cur2, 2, Price{5, 3});
+    // B against A
+    //     offer3: 4 B @  3/4 (A/B)
+    //     offer4: 1 B @  4/5 (A/B)
+    auto const offer3 = createOffer(cur2, cur1, 4, Price{3, 4});
+    auto const offer4 = createOffer(cur2, cur1, 1, Price{4, 5});
+    // B against C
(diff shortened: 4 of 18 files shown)


rokopt:
This invariant _does_ appear to have a major effect on fuzzer execution speed.  Metrics after 20 minutes in current code without this PR:

master-without-order-book-not-crossed-invariant-for-20-min

And with this PR:

order-book-not-crossed-invariant-present-for-20-min

src/invariant/OrderBookIsNotCrossed.cpp
```diff
@@ -0,0 +1,205 @@
...
+            xdr::xdr_to_string(a), xdr::xdr_to_string(b), lowestAsk, highestBid,
+            highestBid, lowestAsk);
+    }
+    return {};
+}
+
+void
+OrderBookIsNotCrossed::deleteFromOrderBook(OfferEntry const& oe)
+{
+    mOrderBook[oe.selling][oe.buying].erase(oe);
```
MonsieurNicolas, line 131: Why isn't `mOrderbook` an `std::unordered_map<std::pair<Asset, Asset>, Orders>` ? All places I see that we're always looking things up with both assets fixed so we're just making things slower for no reason by nesting them.

src/invariant/test/OrderBookIsNotCrossedTests.cpp
```diff
@@ -0,0 +1,333 @@
+#ifdef FUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION
```
MonsieurNicolas, line 1: see me other comment: I'm pretty sure we can always test that the invariant works, regardless of fuzzing enabled or not

MonsieurNicolas: (otherwise this test case will never be run anywhere)

src/main/ApplicationImpl.cpp
```diff
@@ -177,6 +181,9 @@ ApplicationImpl::initialize(bool createNewDB)
     LedgerEntryIsValid::registerInvariant(*this);
     LiabilitiesMatchOffers::registerInvariant(*this);
     SponsorshipCountIsValid::registerInvariant(*this);
+#ifdef FUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION
+    OrderBookIsNotCrossed::registerInvariant(*this);
```
MonsieurNicolas, line 185: as this invariant does not work in the general case (from what I understand), we should probably register it only where we need it

src/invariant/test/OrderBookIsNotCrossedTests.cpp
```diff
@@ -0,0 +1,333 @@
...
+#include "test/TestUtils.h"
+#include "test/TxTests.h"
+#include "test/test.h"
+
+using namespace stellar;
+using namespace stellar::InvariantTestUtils;
+
+// convert a single offer into an offer operation
+Operation
+opFromLedgerEntries(LedgerEntry const& le)
```
MonsieurNicolas, line 19: I don't think those two function are needed: invariants don't care about operations

src/invariant/test/OrderBookIsNotCrossedTests.cpp
```diff
@@ -0,0 +1,333 @@
...
+    {
+        updates = makeUpdateList(current, nullptr);
+    }
+    else if (current.empty())
+    {
+        updates = makeUpdateList(nullptr, previous);
+    }
+    else
+    {
+        updates = makeUpdateList(current, previous);
```
MonsieurNicolas, line 65: we should probably add a `REQUIRE` inside that version of `makeUpdateList` that ensures that keys are matching 1:1 (I know that code was already there, but it's super error prone especially in this context)

src/invariant/test/OrderBookIsNotCrossedTests.cpp
```diff
@@ -0,0 +1,333 @@
...
+
+#include "invariant/OrderBookIsNotCrossed.h"
+#include "invariant/test/InvariantTestUtils.h"
+#include "ledger/LedgerTxn.h"
+#include "lib/catch.hpp"
+#include "test/TestUtils.h"
+#include "test/TxTests.h"
+#include "test/test.h"
+
+using namespace stellar;
```
MonsieurNicolas, line 14: we normally prefer wrapping the whole block with `namespace stellar {`

src/invariant/test/OrderBookIsNotCrossedTests.cpp
```diff
@@ -0,0 +1,333 @@
...
+    {
+        path.emplace_back(les[i].data.offer().selling);
+    }
+
+    // public key and amounts are not important
+    return txtest::pathPayment(PublicKey{}, send, 1, dest, 1, path);
+}
+
+void
+applyCheck(Application& app, std::vector<LedgerEntry> const& current,
```
MonsieurNicolas, line 49: all those helper functions should be in the anonymous namespace as they're likely going to clash

src/invariant/test/OrderBookIsNotCrossedTests.cpp
```diff
@@ -0,0 +1,333 @@
...
+
+    Asset cur2;
+    cur2.type(ASSET_TYPE_CREDIT_ALPHANUM4);
+    strToAssetCode(cur2.alphaNum4().assetCode, "CUR2");
+
+    VirtualClock clock;
+    auto app = createTestApplication(clock, getTestConfig(0));
+    LedgerTxn ltxroot{app->getLedgerTxnRoot()};
+
+    auto const& invariant = std::make_shared<OrderBookIsNotCrossed>();
```
MonsieurNicolas, line 104: reference?

src/invariant/test/OrderBookIsNotCrossedTests.cpp
```diff
@@ -0,0 +1,333 @@
...
+    Asset cur3;
+    cur3.type(ASSET_TYPE_CREDIT_ALPHANUM4);
+    strToAssetCode(cur3.alphaNum4().assetCode, "CUR3");
+
+    // the initial set up for the order book follows where:
+    // A = cur1, B = cur2, and C = cur3
+    // A against B
+    //     offer1: 3 A @  3/2 (B/A)
+    //     offer2: 2 A @  5/3 (B/A)
+    auto const& offer1 = createOffer(cur1, cur2, 3, Price{3, 2});
```
MonsieurNicolas, line 187: reference again?

MonsieurNicolas: this file is filled with bogus refs to temp variable, please fix

MonsieurNicolas (changes requested):

src/invariant/OrderBookIsNotCrossed.h
```diff
@@ -0,0 +1,84 @@
+#ifdef FUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION
+#pragma once
+
+// Copyright 2019 Stellar Development Foundation and contributors. Licensed
```
sisuresh, line 4: Year should be 2021. This comment applies to the two other files you added as well.

src/invariant/test/OrderBookIsNotCrossedTests.cpp
rokopt: You're right, they don't -- the only loss from removing this (and the new `store()` parameter mentioned above) is that the `InvariantManager` won't report any interesting operation from an order-book-not-crossed-invariant test failure, and we'll be able to tell what the operation was by which test failed anyway, so there doesn't seem much point in having those functions.

rokopt:
Alright, I've gone over the inherited files here and made a bunch of changes:

- The part I found trickiest was just getting the code to build and run in as many situations as possible!  We want it to build any time we have `BUILD_TESTS` configured, and to be able to run the invariant's own unit tests in that case, to minimize code rot.  And we want to build and use it when fuzzing is configured.  But we don't want the invariant to be active during any _other_ unit tests, because it maintains state across calls to `checkOnOperationApply()` and the code base in general does not support invariants that do that.  What I ended up doing was not registering the invariant in general because test configs enable `.*`, so if the invariant were registered, then it would be enabled during tests by default.  Instead, I exported a `registerAndEnableInvariant()`, unlike most invariants which export a `registerInvariant()` and let the config manager decide which to enable.  Then only those callers that explicitly know how to handle the order-book invariant's maintaining state can call `registerAndEnableInvariant()`.  And two callers do so:  the fuzzer and the invariant's own unit tests.  The invariant's own unit tests also have to disable _other_ invariants, because they modify ledger entries directly, which can break some other invariants.  (That part is something that several other invariant tests do as well -- they set `INVARIANT_CHECKS` to just themselves.)
- Rebased on master, which in particular includes the best-offer debugging code, with which there were a couple small conflicts.
- Improved the passive-offer-crossing-detection algorithm to break early if possible.
- Removed some weird C++ such as the references to temporaries and unnecessary `at()`s.
- Removed `using namespace` and added conventional namespace use.
- Added an assertion to `makeUpdateList()`.
- Deferred copying `mRolledBackOrderBook` as late as we can.
- Made `extractAssets()` clearer (I hope).
- Removed some changes which aren't necessary anymore after the above changes.

src/invariant/OrderBookIsNotCrossed.h
rokopt: Done.

src/invariant/test/OrderBookIsNotCrossedTests.cpp
rokopt: Wow, those were weird.  I've gotten rid of all the ones I spotted.

src/invariant/test/OrderBookIsNotCrossedTests.cpp
rokopt: Fixed.

src/invariant/test/OrderBookIsNotCrossedTests.cpp
rokopt: Agreed, done.

src/invariant/test/OrderBookIsNotCrossedTests.cpp
rokopt: Done.

src/invariant/test/OrderBookIsNotCrossedTests.cpp
rokopt: Added.

src/invariant/test/OrderBookIsNotCrossedTests.cpp
rokopt: I've removed them.

src/main/ApplicationImpl.cpp
rokopt: I think I've managed that with the `registerAndEnableInvariant()` mechanism.

src/invariant/OrderBookIsNotCrossed.cpp
rokopt: You're right -- that made several passes shorter.

src/invariant/OrderBookIsNotCrossed.cpp
```diff
@@ -0,0 +1,205 @@
...
+    // offers is passive.
+    if (lowestAsk == highestBid)
+    {
+        auto findEqualPriceNonPassive = [](auto const targetPrice,
+                                           auto const& offer) {
+            return price(offer) == targetPrice &&
+                   (offer.flags & PASSIVE_FLAG) == 0;
+        };
+
+        auto nonPassiveAskAtSamePrice =
```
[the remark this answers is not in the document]
rokopt: Done.

src/invariant/OrderBookIsNotCrossed.cpp
```diff
@@ -0,0 +1,205 @@
...
+            std::find_if(asks.cbegin(), asks.cend(),
+                         std::bind(findEqualPriceNonPassive, lowestAsk,
+                                   std::placeholders::_1));
+        auto nonPassiveBidAtSamePrice =
+            std::find_if(bids.cbegin(), bids.cend(),
+                         std::bind(findEqualPriceNonPassive, 1 / lowestAsk,
```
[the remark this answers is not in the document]
rokopt: Actually not -- this is the case where `lowestAsk==highestBid`, so it had better not be a requirement that `1/lowestAsk==highestBid` as well, or else we've seriously narrowed down what `lowestAsk` is allowed to be.  ;-)

src/invariant/OrderBookIsNotCrossed.cpp
```diff
@@ -0,0 +1,205 @@
...
+static std::set<std::pair<Asset, Asset>>
+extractAssetPairs(LedgerTxnDelta const& ltxd)
+{
+    std::set<std::pair<Asset, Asset>> assets;
+    for (auto const& entry : ltxd.entry)
+    {
+        if (entry.first.type() == InternalLedgerEntryType::LEDGER_ENTRY &&
+            entry.first.ledgerKey().type() == OFFER && entry.second.current)
+        {
+            auto const& offer =
```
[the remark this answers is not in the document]
rokopt: I hope I've made it more readable now by ordering the pairs as you suggest.

src/invariant/OrderBookIsNotCrossed.cpp
```diff
@@ -201,5 +201,17 @@ OrderBookIsNotCrossed::checkOnOperationApply(Operation const& operation,
...
+void
+OrderBookIsNotCrossed::snapshotForFuzzer()
+{
+    mRolledBackOrderBook = mOrderBook;
+}
+
+void
+OrderBookIsNotCrossed::resetForFuzzer()
+{
+    mOrderBook = mRolledBackOrderBook;
```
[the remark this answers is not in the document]
rokopt: Good idea, done.

src/test/FuzzerImpl.cpp
```diff
@@ -743,6 +743,7 @@ resetTxInternalState(Application& app)
 // reset caches to clear persistent state
 #ifdef FUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION
     app.getLedgerTxnRoot().resetForFuzzer();
+    app.getInvariantManager().resetForFuzzer();
```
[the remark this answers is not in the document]
rokopt: Done.

src/invariant/test/InvariantTestUtils.cpp
```diff
@@ -57,7 +57,7 @@ generateOffer(Asset const& selling, Asset const& buying, int64_t amount,
 
 bool
 store(Application& app, UpdateList const& apply, AbstractLedgerTxn* ltxPtr,
-      OperationResult const* resPtr)
+      OperationResult const* resPtr, Operation const* opPtr)
```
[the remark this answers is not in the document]
rokopt: It had been used in the previous PR, but other changes in this one have allowed it to be removed.

src/invariant/OrderBookIsNotCrossed.cpp
```diff
@@ -0,0 +1,225 @@
...
+                return false;
+            };
+
+        OfferEntry ask, bid;
+        auto nonPassiveAskAtSamePrice =
+            existsEqualPriceNonPassive(lowestAsk, asks, ask);
+        auto nonPassiveBidAtSamePrice =
+            existsEqualPriceNonPassive(1 / lowestAsk, bids, bid);
+
+        if (nonPassiveAskAtSamePrice && nonPassiveBidAtSamePrice)
```
jonjove, line 102: Based on all of my comments here, I believe that there is actually an O(1) way to perform this check. Your current implementation is O(n) so this would be a big improvement.

src/invariant/OrderBookIsNotCrossed.cpp
```diff
@@ -0,0 +1,225 @@
...
+void
+OrderBookIsNotCrossed::snapshotForFuzzer()
+{
+    mRolledBackOrderBook = mOrderBook;
+}
+
+void
+OrderBookIsNotCrossed::resetForFuzzer()
+{
+    mRollBackBeforeNextCheck = true;
```
jonjove, line 222: I'm not 100% sure I understand the reasoning for this implementation. You could do `mOrderBook = mRolledBackOrderBook` here, right? I suppose I can imagine some reasons for not doing that. What was your reasoning? You probably want to put it in a comment if it's important.

src/invariant/OrderBookIsNotCrossed.cpp
```diff
@@ -0,0 +1,225 @@
...
+                "price: {}\n\nWhere {} >= {}!"),
+            xdr::xdr_to_string(a), xdr::xdr_to_string(b), lowestAsk, highestBid,
+            highestBid, lowestAsk);
+    }
+    return {};
+}
+}
+
+void
+OrderBookIsNotCrossed::deleteFromOrderBook(OfferEntry const& oe)
```
jonjove, line 130: Nitpick: are these functions really adding any value? They're both one liners that are easy to understand, and they have a single caller.

src/invariant/OrderBookIsNotCrossed.h
```diff
@@ -0,0 +1,99 @@
...
+        else
+        {
+            return false;
+        }
+    }
+};
+
+using Orders = std::set<OfferEntry, OfferEntryCmp>;
+using OrderBook = std::unordered_map<AssetPair, Orders, AssetPairHash>;
+using AssetPairSet = std::set<std::pair<Asset, Asset>>;
```
jonjove, line 46: Why not `unordered_set`?

jonjove (changes requested):
I didn't review the tests yet because I think the implementation is going to change. You should add tests for some of the interesting floating-point edge cases I pointed out.

src/invariant/OrderBookIsNotCrossed.cpp
rokopt: That was Nicolas's idea, and, I think, a good one: https://github.com/stellar/stellar-core/pull/2959#discussion_r592848513

The point is that some fuzz operation sets won't involve orders, and in those cases we call `OrderBookIsNotCrossed::checkOnOperationApply()` but not `OrderBookIsNotCrossed::check()`.  In such a case, we can skip the copy and allow the invariant's representation of the order book to be stale until we actually use it, in `OrderBookIsNotCrossed::check()`.  I'll add a comment to that effect.

jonjove: That's the same reason I came up with.

src/invariant/OrderBookIsNotCrossed.cpp
```diff
@@ -0,0 +1,225 @@
...
+                        found = offer;
+                        return true;
+                    }
+                }
+                return false;
+            };
+
+        OfferEntry ask, bid;
+        auto nonPassiveAskAtSamePrice =
+            existsEqualPriceNonPassive(lowestAsk, asks, ask);
```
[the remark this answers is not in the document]
rokopt: Here's what I think is true: there is an ask with price `lowestAsk`, but it could be passive, and if it is, then it doesn't indicate by itself that the order book is crossed (or perhaps it's more accurate to say, crossed in a way that we don't intend?  We intend, as I understand it, for a passive ask/bid to be able to be placed into the book even if there is an existing bid/ask at the same price).  However, if there is any other ask at the same price which isn't passive, and a bid at the same price which isn't passive, then we know that some non-passive ask/bid arrived at a time when another bid/ask at the same price was already present, so the new, non-passive offer should take taken the existing one; we must have a bug.  Is that right?

src/test/FuzzerImpl.cpp
```diff
@@ -1327,6 +1328,10 @@ TransactionFuzzer::initialize()
     // commit this to the ledger so that we have a starting, persistent
     // state to fuzz test against
     ltxOuter.commit();
+
+#ifdef BUILD_TESTS
+    mApp->getInvariantManager().snapshotForFuzzer();
```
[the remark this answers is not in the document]
rokopt: The fuzzer already uses `txtest`, and `getTestConfig()`, and `TestAccount`, so I think we committed to that a while ago.  We could revisit that if fuzzing without testing infrastructure has significant value, of course, but it's not clear to me that it does, and there would be a lot to change besides `snapshotForFuzzer()`.

src/invariant/OrderBookIsNotCrossed.cpp
```diff
@@ -0,0 +1,225 @@
...
+                           "asset pair.\nEqual price {} for non-passive offers "
+                           "{} and {}."),
+                xdr::xdr_to_string(a), xdr::xdr_to_string(b), lowestAsk,
+                xdr::xdr_to_string(ask), xdr::xdr_to_string(bid));
+        }
+
+        return {};
+    }
+
+    if (lowestAsk < highestBid)
```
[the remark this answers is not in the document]
rokopt: `lowestAsk == highestBid` and `lowestAsk < highestBid` are mutually exclusive, so how would we ever end up doing the (higher-complexity check) in the former path in a case that we could have avoided by doing the latter check first?

src/invariant/OrderBookIsNotCrossed.cpp
```diff
@@ -0,0 +1,225 @@
...
+    auto lowestAsk = price(*asks.cbegin());
+    auto highestBid = 1 / price(*bids.cbegin());
+
+    // Allow highestBid to equal (not exceed) lowestAsk if at least one of the
+    // offers is passive.
+    if (lowestAsk == highestBid)
+    {
+        auto existsEqualPriceNonPassive =
+            [](auto const targetPrice, Orders const& offers, auto& found) {
+                for (auto const& offer : offers)
```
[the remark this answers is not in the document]
rokopt: The loop breaks as soon as it hits any offer whose price differs from that of the first offer in the list (`lowestAsk`), so it is taking advantage of the price-ordering, isn't it?

Within the same price, they're ordered by offer ID, i.e. the order in which they were created, but it's not apparent to me how to take advantage of that; if there were a bug that allowed some non-passive order to sneak into the midst of a long list of passive offers in the same `checkOnOperationApply()` call, all at the same price as the non-passive order, which should have crossed exactly equally with some pre-existing offer, then as far as I can tell we have to look at all the offers at that price to catch the bug.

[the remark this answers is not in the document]
jonjove: My mistake, this does take advantage of the sort order.

src/invariant/OrderBookIsNotCrossed.cpp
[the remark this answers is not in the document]
jonjove: Yeah sorry. I was jumping around when I reviewed this and forgot that the previous check was gated on `lowestAsk == highestBid`.

src/invariant/OrderBookIsNotCrossed.cpp
rokopt: My comment above (https://github.com/stellar/stellar-core/pull/2959#discussion_r595418524) explains why I currently do not see how we can avoid searching a list of offers at a price that exactly equally crosses some pre-existing offer to find out whether any of them is not passive, if we are to catch a bug where there is such an offer.  (Unless we were to check and update state with each individual `Operation` in `checkOnOperationApply`, rather than extracting all the asset pairs and batching the updates and then doing a single check, but I would expect that to be slowing down more common cases to speed up this rarer one.)  I might be misreading the semantics of passive offers, though!

src/invariant/OrderBookIsNotCrossed.cpp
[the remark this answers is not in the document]
jonjove: Okay this works. Comment highly recommended because that is very strange looking.

src/invariant/OrderBookIsNotCrossed.cpp
jonjove: Despite your clarifications about how this implementation works, I still believe this can be done with O(1) complexity.

src/invariant/OrderBookIsNotCrossed.cpp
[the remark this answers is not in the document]
rokopt: In light of discussion below, the implementation can be made faster.  The replacement will also need comments since it's a bit tricky too.  :)

src/invariant/OrderBookIsNotCrossed.cpp
rokopt: I eventually caught on to Jon's clever idea here:  extend `OfferEntryCmp` to sort non-passive orders before passive orders at the same price.  Then we know that if there's _any_ non-passive order in the list of offers at the same price, then one of the non-passive orders will be at the front.  That allows us to check only the fronts of each list.

src/invariant/OrderBookIsNotCrossed.h
rokopt: I have actually tried it at a few different stages of development, but so far I haven't found a point where fuzzing was clearly faster with `unordered_set`, and at times it's been slower.  There's a lot of variation on my laptop so it's hard to be sure, and I was surprised at first, but it's actually not implausible that in this case `set` might be faster:  this is a fuzzing-only invariant and when fuzzing there are only ever a very small number of orders for a given asset pair, so we might be doing one or zero comparisons typically with `set`, and with `unordered_set` we always have to do one hash.

If we ever make this invariant workable on production-size order books, then I would guess that `unordered_set` would become clearly better, but the current version of this invariant isn't.  :)

src/invariant/OrderBookIsNotCrossed.cpp
rokopt: Agreed, I'll inline those.

src/invariant/OrderBookIsNotCrossed.cpp
```diff
@@ -0,0 +1,225 @@
...
+static double
+price(OfferEntry const& offer)
+{
+    return double(offer.price.n) / double(offer.price.d);
+}
+
+static std::string
+checkCrossed(Asset const& a, Asset const& b, OrderBook const& orderBook)
+{
+    // if either side of order book for asset pair empty or does not yet exist,
```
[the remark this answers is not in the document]
rokopt: Will fix.

rokopt:
I've pushed changes to address all comments except on the floating-point reciprocal issue; these changes are small so I figure I may as well make them available for review while I think about the floating-point.

src/invariant/OrderBookIsNotCrossed.h
```diff
@@ -30,6 +30,14 @@ struct OfferEntryCmp
         {
             return true;
         }
+        else if (((a.flags & PASSIVE_FLAG) == 0) &&
```
jonjove, line 33: This makes `OfferEntryCmp` violate the asymmetry requirement of https://en.cppreference.com/w/cpp/named_req/Compare.

Let `x < y` denote `OfferEntryCmp()(x, y)` for any `OfferEntry x, y`. If
```
L: Price = 1/1, Flags = PASSIVE_FLAG (offerID doesn't matter in this example)
R: Price = 2/1, Flags = 0 (offerID doesn't matter in this example)
```
Then

- `L < R == true` because `price(L) < price(R) == true`
- `R < L == true` because `price(R) < price(L) == false` and `(R.flags & PASSIVE_FLAG == 0) && (L.flags & PASSIVE_FLAG != 0) == true`

jonjove (changes requested):

src/invariant/OrderBookIsNotCrossed.h
rokopt: Thanks -- that test is supposed to be inside the prices-equal case, of course.  Sigh.  Fixed.

src/invariant/OrderBookIsNotCrossed.cpp
```diff
@@ -0,0 +1,225 @@
...
+    auto const& asks = iterAB->second;
+    auto const& bids = iterBA->second;
+    if (asks.empty() || bids.empty())
+    {
+        return {};
+    }
+
+    // check if crossed
+    auto lowestAsk = price(*asks.cbegin());
+    auto highestBid = 1 / price(*bids.cbegin());
```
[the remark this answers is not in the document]
rokopt: One thing I am trying to confirm to myself -- or find a counterexample to -- as I consider this is whether we really can always resolve the question of whether the book is crossed by looking at the first offers in the per-asset-pair ask/bid lists.  I see this comment to `exchangeV10()`:

```
// exchangeV10 is a system for crossing offers that provides guarantees
// regarding the direction and magnitude of rounding errors:
// - When considering two crossing offers subject to a variety of limits,
//   exchangeV10 has a consistent approach to determining which offer is larger.
//   The smaller offer is always removed from the book.
// - When two offers cross, the rounding error will favor the offer that remains
//   in the book.
// - The rounding error will not favor either party by more than 1% (except in
//   the case of path payment, where the rounding error can favor the offer in
//   the book by an arbitrary amount). If the rounding error would exceed 1%,
//   no trade occurs and the smaller offer is removed.
```

In particular I think that the "smaller offer is removed" part means that rounding error should not allow the book to cross, but rather, if rounding error would prevent an exchange required to prevent the book from crossing, then one of the offers is simply removed altogether.  And I think that that means in turn that we indeed don't need to search beyond the first element:  I don't think there is a case where we are _expected_ to leave crossed offers in the book because of rounding errors preventing them from being taken.  But I have by no means absorbed all the rules yet.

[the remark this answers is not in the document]
rokopt: > but suffice it to say that x = 1/49 is such an example

Bingo!  If I write this test:

```
    SECTION("Not crossed where highest bid = lowest ask but only because of "
            "passive offer(s), where double-precision reciprocal of price is "
            "not involutory")
    {
        auto const offer9 = createOffer(cur4, cur5, 1, Price{49, 1}, true);
        auto const offer10 = createOffer(cur5, cur4, 1, Price{1, 49}, true);
        applyCheck(*app, {offer9, offer10}, {}, true);
    }
```

Then I get:

```
2021-03-16T15:08:19.365 <test> [default FATAL] Order book is in a crossed state for { type = ASSET_TYPE_CREDIT_ALPHANUM4,
  alphaNum4 = {
    assetCode = 43555234,
    issuer = {
      type = PUBLIC_KEY_TYPE_ED25519,
      ed25519 = 0000000000000000000000000000000000000000000000000000000000000000
    }
  } }
 - { type = ASSET_TYPE_CREDIT_ALPHANUM4,
  alphaNum4 = {
    assetCode = 43555235,
    issuer = {
      type = PUBLIC_KEY_TYPE_ED25519,
      ed25519 = 0000000000000000000000000000000000000000000000000000000000000000
    }
  } }
 asset pair.
Top of the book is:
        Ask price: 49.0
        Bid price: 49.00000000000001

Where 49.00000000000001 >= 49.0!
```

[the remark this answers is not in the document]
rokopt: As for fixing it...

`stellar-core` does use `double` when comparing prices, but as far as I can tell, it's in `isBetterOffer` where we're comparing prices with the same ask and bid assets, not with the reversed ones.  When we do a reciprocal, we do it in the rational form:

```
static Price
getInversePrice(Price const& price)
{
    return Price{price.d, price.n};
}
```

And when checking for crossing, we're using the `Price` comparison operators from `util/types.h`:

```
bool
ManageOfferOpFrameBase::doApply(AbstractLedgerTxn& ltxOuter)
{
[ ... ]
                if ((passive && (o.price >= maxWheatPrice)) ||
                    (o.price > maxWheatPrice))
                {
                    return OfferFilterResult::eStop;
                }
[ ... ]
}
```

Those operators use large rational arithmetic:

```
bool
operator>(Price const& a, Price const& b)
{
    uint128_t l(a.n);
    uint128_t r(a.d);
    l *= b.d;
    r *= b.n;
    return l > r;
}
```

So it's not obvious to me why we should be using `double` in the `OrderBookIsNotCrossed` invariant at all.  Indeed, if I switch to using the `Price` comparison operator, then the test above, and a similar one which should throw but doesn't when using `double`, both pass.  So I've committed those tests, and I've also committed changes to remove the use of `double` from the invariant module.

However, there are a couple things which I might want to change but that I'm not sure what the best way would be, so I'll ask here first:

- `types.h` only defines `>`, `>=`, and `==`; it doesn't define `<` or `<=`.  For now I just flipped the `<=` so that I wouldn't have to touch `types.h` in this PR, but I could add it if you prefer.
-  `types.h`'s version of `==` uses structural rather than rational equality:
```
bool
operator==(Price const& a, Price const& b)
{
    return (a.n == b.n) && (a.d == b.d);
}
```
So it's unsuitable for the uses in the invariant module.  In the committed code, I wrote something with the spelled-out name `pricesEqualAsRationals`, which works analogously to `>=` and `>`.  I'm guessing that some things are relying on the current semantics of `types.h`'s `==`, or else it wouldn't be that way, so I won't touch it unless someone can confirm otherwise.  I'm not sure whether I should do something more concise than `pricesEqualAsRationals` or whether the wordiness is actually good for clarity in this case.

src/invariant/OrderBookIsNotCrossed.h
```diff
@@ -30,16 +30,16 @@ struct OfferEntryCmp
...
-                 ((b.flags & PASSIVE_FLAG) != 0))
-        {
-            // We order non-passive offers before passive offers so that we can
-            // find any non-passive offer at a given price at the front of the
-            // list of offers at that price.
-            return true;
-        }
         else if (priceA == priceB)
         {
+            if (((a.flags & PASSIVE_FLAG) == 0) &&
```
jonjove, line 36: This still violates the asymmetry requirement.
```
L: Price = 1/1, Flags = 0, OfferID = 2
R: Price = 1/1, Flags = PASSIVE_FLAG, OfferID = 1
```

Then

- `L < R == true` because `price(L) == price(R)` and `(L.flags & PASSIVE_FLAG == 0) && (R.flags & PASSIVE_FLAG != 0) == true`
- `R < L == true` because `price(R) == price(L)` and `(R.flags & PASSIVE_FLAG == 0) && (L.flags & PASSIVE_FLAG != 0) == false` and `id(R) < id(L) == true`

src/invariant/OrderBookIsNotCrossed.h
jonjove: That's a good reason.

src/invariant/OrderBookIsNotCrossed.h
rokopt: Good catch -- I should be checking the reverse condition on flags as well.  I've pushed a commit to do that.

While staring at it for a while to see what _else_ I might have messed up, I also decided that I like it better without the else-after-returns.  So I removed them in a separate commit.  There appears to me to be some legitimate dispute over else-after-return, but `clang-tidy`, for example, is unambiguous (https://clang.llvm.org/extra/clang-tidy/checks/readability-else-after-return.html), as is Chromium (https://chromium.googlesource.com/chromium/src/+/refs/tags/73.0.3627.0/styleguide/c++/c++.md), and those two strike me as the most authoritative opinions that I've found so far.

rokopt: I've also now added your two examples in a new `TEST_CASE` specifically for `OfferEntryCmp`.

jonjove:
@rokopt can you squash+rebase at this point? We've gone through a lot of iterations here, with considerable back-and-forth on changes, and I need to get a clean look at it.

rokopt:
> @rokopt can you squash+rebase at this point? We've gone through a lot of iterations here, with considerable back-and-forth on changes, and I need to get a clean look at it.

Sure, done.

src/invariant/test/OrderBookIsNotCrossedTests.cpp
```diff
@@ -0,0 +1,361 @@
...
+        auto entry = ltx.load(LedgerEntryKey(offer));
+        entry.erase();
+
+        invariant->checkOnOperationApply({}, OperationResult{}, ltx.getDelta());
+
+        REQUIRE(invariant->getOrderBook().at({cur1, cur2}).size() == 0);
+    }
+}
+
+TEST_CASE("OrderBookIsNotCrossed properly throws if order book is crossed",
```
jonjove, line 170: One interesting group of cases that you aren't testing: changing the assets on an offer (this is permitted, you can even swap them). You might want to add some tests for this in both this case and the previous one.

src/invariant/OrderBookIsNotCrossed.h
```diff
@@ -0,0 +1,112 @@
...
+        auto const priceA = priceAsDouble(a.price);
+        auto const priceB = priceAsDouble(b.price);
+
+        if (priceA < priceB)
+        {
+            return true;
+        }
+        if (priceA == priceB)
+        {
+            if (((a.flags & PASSIVE_FLAG) == 0) &&
```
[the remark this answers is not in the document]
rokopt: Ah, that is nice.  Done.

src/invariant/OrderBookIsNotCrossed.cpp
```diff
@@ -0,0 +1,202 @@
...
+#include "util/Logging.h"
+#include "util/XDROperators.h"
+#include "xdr/Stellar-ledger-entries.h"
+#include <fmt/format.h>
+#include <functional>
+#include <xdrpp/printer.h>
+
+namespace stellar
+{
+double
```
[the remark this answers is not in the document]
rokopt: I've removed uses of `static` in favor of either the anonymous namespace or of membership in `OrderBookNotCrossed`, depending on whether any members of the latter need the types/functions.

src/invariant/OrderBookIsNotCrossed.cpp
```diff
@@ -0,0 +1,202 @@
...
+priceAsDouble(Price const& price)
+{
+    return static_cast<double>(price.n) / static_cast<double>(price.d);
+}
+
+namespace
+{
+static AssetPairSet
```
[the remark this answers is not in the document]
rokopt: `static` removed.

src/invariant/OrderBookIsNotCrossed.h
```diff
@@ -0,0 +1,112 @@
...
+
+namespace stellar
+{
+double priceAsDouble(Price const& price);
+
+// compare two OfferEntry's by price
+struct OfferEntryCmp
+{
+    bool
+    operator()(OfferEntry const& a, OfferEntry const& b) const
```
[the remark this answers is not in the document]
rokopt: Agreed, I moved that into the cpp file.

src/invariant/test/OrderBookIsNotCrossedTests.cpp
```diff
@@ -0,0 +1,361 @@
...
+    r.offerID = 1;
+    REQUIRE(cmp(l, r));
+    REQUIRE(!cmp(r, l));
+}
+
+TEST_CASE("OrderBookIsNotCrossed in-memory order book is consistent with "
+          "application behaviour",
+          "[invariant][OrderBookIsNotCrossed]")
+{
+    Asset cur1;
```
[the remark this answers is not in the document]
rokopt: I factored that out to use `TestAccount::asset()`.

src/invariant/test/OrderBookIsNotCrossedTests.cpp
rokopt: That _is_ interesting.  I added several tests.  It also made me realize that you could cross the book by making a passive offer non-passive, so I added a test for that too.

rokopt:
Note:  in addition to responding to the latest code review comments, I've also pushed a commit to remove the `mRollbackBeforeNextCheck` optimization, because I discovered that it causes occasional crashes during long fuzzing runs.  I don't understand why yet.  We can either delay this PR until I've found a fix, or merge it and file an issue to perform that optimization (correctly) afterwards, as you prefer.

src/invariant/OrderBookIsNotCrossed.cpp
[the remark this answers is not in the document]
MonsieurNicolas: I see that you reverted your change. From the looks of it, it was just not correct.
you need to always `clear` in `resetForFuzzer` (for stability) and CoW

src/invariant/OrderBookIsNotCrossed.cpp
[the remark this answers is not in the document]
MonsieurNicolas: yes, thank you!

src/invariant/test/OrderBookIsNotCrossedTests.cpp
MonsieurNicolas: wow that was a jar jar binks comment from my part.
Meesa resolve this comment now!

src/invariant/test/OrderBookIsNotCrossedTests.cpp
```diff
@@ -1,3 +1,4 @@
+#include "test/TestAccount.h"
```
jonjove, line 1: Nitpick: this looks like it's in the wrong spot?

rokopt: Indeed.  And odd.  Fixed.

rokopt:
> Note: in addition to responding to the latest code review comments, I've also pushed a commit to remove the `mRollbackBeforeNextCheck` optimization, because I discovered that it causes occasional crashes during long fuzzing runs. I don't understand why yet. We can either delay this PR until I've found a fix, or merge it and file an issue to perform that optimization (correctly) afterwards, as you prefer.

This optimization is restored now, in a corrected form.

MonsieurNicolas:
Looks like it's basically ready to go? Needs to be rebased and we can merge. @jonjove what do you think?

As a sanity check I didn't see an update on the fuzzer stability front (only one I see here is the one from very early on). So maybe @rokopt can post this in parallel?

jonjove:
@MonsieurNicolas I just went through and resolved all open comments from me. Should be ready for rebase/squash/merge. Agreed that I'd like to see the fuzzer statistics though.

rokopt:
I've squashed and rebased.  Stability didn't change noticeably in this PR; it's still at the ~92% where it's been since the ledger-key mapping work:

afl-stats-with-squashed-order-book-not-crossed-pr

MonsieurNicolas:
mmm, I do have a question though. Is the "exec speed" really that slow compared to both master and the previous iteration of that PR?

I would not expect such a difference as the vast majority of fuzzing should not involve this invariant (it is, after all, unlikely that the fuzzer explores branches with offers).

src/invariant/OrderBookIsNotCrossed.cpp
```diff
@@ -0,0 +1,230 @@
...
+}
+
+std::string
+OrderBookIsNotCrossed::checkOnOperationApply(Operation const& operation,
+                                             OperationResult const& result,
+                                             LedgerTxnDelta const& ltxDelta)
+{
+    updateOrderBook(ltxDelta);
+    auto assetPairs = extractAssetPairs(ltxDelta);
+    return assetPairs.size() > 0 ? check(assetPairs) : std::string{};
```
MonsieurNicolas, line 207: I *think* that the reason things are super slow is that `assetPairs.size() > 0` is always true even when there are no changes to the order book from this operation. In fact we should not perform any work in `extractAssetPairs` if the operation didn't touch the order book.

MonsieurNicolas:
I think I found the bug, let me know if this makes sense @rokopt

src/invariant/OrderBookIsNotCrossed.cpp
rokopt: Hmm, why must `assetPairs.size()` be `> 0`?  It looks to me as though if there are no `OFFER`s in `LedgerTxnDelta`, then it will be `0`, and I put in some logs and they say that it is often `0`.

One thing I mentioned to Jon about performance which I should bring up is that exec speed of fuzzing seems highly variable on my laptop, even from run to run without code changes (unlike stability, for example, which is generally within a fraction of a percent across many runs).  Fuzzing does make my laptop fan roar, and I have a suspicion (though no confirmation) that there might be highly variable amounts of CPU-throttling going on depending on I-don't-know-what thermal factors.

MonsieurNicolas: meh and I thought my head was straight in the morning :)

yeah this is extracting only from the current txn object so not it

MonsieurNicolas:
I ran the version with the invariant enabled and without based on the same initial corpus (so no variation because of that) and they performed the same way, so it was probably a fluke with your computer

MonsieurNicolas (approved):

MonsieurNicolas:
we'll merge as soon as master reopens

PR #2962, opened 2021-03-11
Update dev container to focal

MonsieurNicolas:
Was using xenial, which is EOL (and we're using focal in CI anyways)


graydon:
r+ f8d6a90

leighmcculloch (approved):

PR #2965, opened 2021-03-18
Check for memcpy() of null in signHashX

rokopt:
# Check for memcpy() of null in signHashX

`ubsan` complains about a `memcpy` on `NULL` (of 0 bytes) in `signHashX()`.  I don't think we can be getting any bad behavior in practice, but it's technically UB even to copy 0 bytes from `NULL`, and we should silence `ubsan`.

# Checklist
- [x] Compiles
- [x] Ran all tests

### DIFF
--- a/src/transactions/SignatureUtils.cpp
+++ b/src/transactions/SignatureUtils.cpp
@@ -51,7 +51,14 @@ signHashX(const ByteSlice& x)
     DecoratedSignature result;
     Signature out(0, 0);
     out.resize(static_cast<uint32_t>(x.size()));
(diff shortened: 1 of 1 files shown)


MonsieurNicolas:
r+ cc5561bb243ecb28feadbb5d2e612db6ae506ae8

rokopt:
The failures here look like some connectivity problem?

> fatal: could not read Username for 'https://github.com': No such device or address

MonsieurNicolas:
@latobarita: retry

PR #2966, opened 2021-03-18
Fix fuzzer offer setup parameters and assertion

rokopt:
# Fix two bugs in fuzzer setup configuration

While working on making fuzzer setup more varied, I noticed two bugs in the existing code:

- There's an assertion that tries to confirm that setup has done exactly what the configuration was intended to mean, which states that the root account is the only modified ledger entry in the setup transaction (the others should all be newly-created).  That's true for almost any configuration -- but not for the _empty_ configuration, in which case there are _no_ modified ledger entries.  We are not planning for the configuration to be empty, but technically the assertion is wrong; an empty configuration is not inherently erroneous, and nothing else in the code depends on the configuration being non-empty.
- More important, the ask and bid parameters to `txtest::manageOffer()` have been reversed ever since the first version of the order book setup code!  This hasn't mattered either because the setup has been symmetrical, but we're going to make it _not_ so, and it would cause complete confusion then.

# Checklist
- [x] Compiles
- [x] Ran all tests

### DIFF
--- a/src/test/FuzzerImpl.cpp
+++ b/src/test/FuzzerImpl.cpp
@@ -1344,12 +1344,18 @@ TransactionFuzzer::storeSetupLedgerKeys(AbstractLedgerTxn& ltx)
     std::sort(init.begin(), init.end());
 
     // Setup should only create entries; there should be no dead entries, and
-    // only one "live" (modified) one:  the root account.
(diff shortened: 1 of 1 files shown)


jonjove:
r+ 715c78f

PR #2970, opened 2021-03-22
Make apt mirror configurable in devcontainer

leighmcculloch:
# Description

Reduce the build time of the devcontainer from 5min30sec to 1min20sec for users in San Francisco and AWS's California location by making the apt mirrors configurable.

The currently configured mirrors are insane slow from several locations I tested, compared to the default apt repositories. It's not one slow mirror being selected either, apt is selecting a different mirror for each package install.

A build arg is added that is the apt mirror that should be used. When not set the default apt repos are used. A developer may specify the mirror using a variety of options:
- `mirror://mirrors.ubuntu.com/mirrors.txt` to use the full list of all mirrors
- `mirror://mirrors.ubuntu.com/<country-code>.txt` to use country specific mirror lists directly
- `http://<country-code>.archive.ubuntu.com/ubuntu/` to use the country specific mirrors via Canonical
- `http://<aws-region>.ec2.archive.ubuntu.com/ubuntu` to use the mirrors hosted in the same AWS region

# Checklist
- [x] Compiles
- [x] Ran all tests

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -19,10 +19,21 @@ ARG USER_GID=$USER_UID
 RUN apt-get update \
     && apt-get -y install --no-install-recommends apt-utils dialog ca-certificates 2>&1
 
-# use mirrors instead of hardcoded location for performance
-RUN sed -i -e 's/http:\/\/us.archive/mirror:\/\/mirrors/' -e 's/\/ubuntu\//\/mirrors.txt/' /etc/apt/sources.list \
-    && sed -i -e 's/http:\/\/security/mirror:\/\/mirrors/' -e 's/\/ubuntu\//\/mirrors.txt/' /etc/apt/sources.list \
-    && sed -i -e 's/http:\/\/archive/mirror:\/\/mirrors/' -e 's/\/ubuntu\//\/mirrors.txt/' /etc/apt/sources.list
+# use apt mirror instead of default archives if specified
+# to use, specify the build arg or as an env var on the host machine
+# e.g.:
+#   mirror://mirrors.ubuntu.com/mirrors.txt
+#   mirror://mirrors.ubuntu.com/<country-code>.txt
+#   http://<country-code>.archive.ubuntu.com/ubuntu/
+#   http://<aws-region>.ec2.archive.ubuntu.com/ubuntu
+ARG APT_MIRROR=
+RUN if [ ! -z "${APT_MIRROR}" ]; then \
+    sed -i \
+        -e "s|http://archive.ubuntu.com/ubuntu/|${APT_MIRROR}|" \
+        -e "s|http://security.ubuntu.com/ubuntu/|${APT_MIRROR}|" \
+        /etc/apt/sources.list \
+    ; fi \
+    ; grep "^[^#;]" /etc/apt/sources.list
 
 # install base container packages and prep for VSCode
 RUN apt-get update \
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -3,6 +3,11 @@
 {
 	"name": "Ubuntu 20.04 & Git",
 	"dockerFile": "Dockerfile",
+	"build": {
+		"args": {
+			"APT_MIRROR": "${localEnv:APT_MIRROR}"
+		}
+	},
 	// The optional 'runArgs' property can be used to specify additional runtime arguments.
 	"runArgs": [
 		// Uncomment the line if you will use a ptrace-based debugger like C++, Go, and Rust.


MonsieurNicolas:
I think you forgot to clean docker caches between runs, which made your timing results unreliable.

I opened a different PR to fix mirroring (that was broken when I switched over to focal)

leighmcculloch:
>I think you forgot to clean docker caches between runs, which made your timing results unreliable.

I'm skipping cache by using the `--no-cache` option when building. Docker will rebuild every layer when that option is in use regardless of what docker layers are cached. My timing results appear to be consistent, I ran them repeatedly, and reran them now and they're still consistent. I've tested them from San Francisco, and from AWS us-west-1, and from both locations it takes 4m to 6m, averaging about 5m20s.

I tried your fix for mirrors in PR #2974 and it is still really slow from both locations.

leighmcculloch:
@MonsieurNicolas I've updated this PR so that the use of mirrors is optional - the idea you shared in Keybase (keybase://chat/stellar.public#core/3106). Turns out this is doable by setting an environment variable on the host machine and that can be forwarded as a build argument. What do you think?

.devcontainer/Dockerfile
```diff
@@ -17,12 +17,15 @@ ARG USER_GID=$USER_UID
...
 
 # use mirrors instead of hardcoded location for performance
-RUN sed -i -e 's/http:\/\/us.archive/mirror:\/\/mirrors/' -e 's/\/ubuntu\//\/mirrors.txt/' /etc/apt/sources.list \
+ARG USE_APT_MIRRORS=0
+RUN if [ "${USE_APT_MIRRORS}" = "1" ]; then \
+    sed -i -e 's/http:\/\/us.archive/mirror:\/\/mirrors/' -e 's/\/ubuntu\//\/mirrors.txt/' /etc/apt/sources.list \
     && sed -i -e 's/http:\/\/security/mirror:\/\/mirrors/' -e 's/\/ubuntu\//\/mirrors.txt/' /etc/apt/sources.list \
-    && sed -i -e 's/http:\/\/archive/mirror:\/\/mirrors/' -e 's/\/ubuntu\//\/mirrors.txt/' /etc/apt/sources.list
+    && sed -i -e 's/http:\/\/archive/mirror:\/\/mirrors/' -e 's/\/ubuntu\//\/mirrors.txt/' /etc/apt/sources.list \
+    ; fi
```
leighmcculloch, line 28: Another idea, that might be better, would be for the build-arg to provide the mirror URL that should be used, that way people can choose their mirror.

MonsieurNicolas: this seems fine as it's the recommended way that automatically picks the right one based on location

MonsieurNicolas: or are you talking about the case when  `USE_APT_MIRRORS` is false? That may make sense: right now configs use `archive.ubuntu.com` but I noticed (on my machine) that if I force it to `us.archive.ubuntu.com` instead it's about twice as fast (but still slower than other mirrors).

This may make a big difference for people not in the US.

.devcontainer/Dockerfile
```diff
@@ -45,24 +48,23 @@ RUN apt-get update \
...
-RUN apt-get -y install git build-essential pkg-config autoconf automake libtool bison flex libpq-dev parallel libunwind-dev
-
-# Update compiler tools
-RUN apt-get -y install libstdc++-7-dev clang-format-8 ccache
-
-# gcc
-RUN apt-get -y install cpp-7 gcc-7 g++-7
-# clang
-RUN apt-get -y install clang-8 llvm-8
+# Install development tools
```
MonsieurNicolas, line 51: this change is actually not a good change:
running multiple `apt-get install` has no real performance benefits, on the other hand, crafting things like this forces a full image rebuild (when changing compilers) as there are no layers and as a consequence slows down a lot future updates.

Now that you figured out how to pass environment variables, I am actually tempted to switch to user selectable compiler toolchains in the devcontainer.json file (and having proper layers becomes more important)

MonsieurNicolas (changes requested):

.devcontainer/Dockerfile
leighmcculloch: If the goal is to be able to run the separate apt-get installs in isolation later on, then this Dockerfile requires some modifications, because every isolated apt install requires an apt update otherwise rerunning only one layer later can error. I'll undo this part of the change and will open another PR fixing the missing apt updates.

.devcontainer/Dockerfile
leighmcculloch: I was meaning change the var to something like `APT_MIRROR` which you can configure to `http://mirrors.ubuntu.com/mirrors.txt`, or you could manually specify your country with `http://mirrors.ubuntu.com/AU.txt`, or you can select a specific mirror if for example you're on a University campus, or in AWS, where both typically have local network mirrors. This would be helpful on AWS specifically.

leighmcculloch:
I've changed the PR removing any non-mirror changes, and making the mirror configurable. A build arg is added that is the apt mirror that should be used. When not set the default apt repos are used.

A developer may specify the mirror using a variety of options:
 
 * `mirror://mirrors.ubuntu.com/mirrors.txt` to use the full list of all mirrors
 * `mirror://mirrors.ubuntu.com/<country-code>.txt` to use country specific mirror lists directly
 * `http://<country-code>.archive.ubuntu.com/ubuntu/` to use the country specific mirrors via Canonical
 * `http://<aws-region>.ec2.archive.ubuntu.com/ubuntu` to use the mirrors hosted in the same AWS region
 * Or a specific single mirror, like that of their university, or ISP.

leighmcculloch:
The build failed because of a launchpad.net timeout, and not because of the changes in this PR.

MonsieurNicolas (approved):
Looking good - we'll merge next week as soon as master reopens

MonsieurNicolas:
r+ da81455ae68d82a5a4f582a0736eb28b65928fa9

PR #2980, opened 2021-03-25
Captive core uses any available archive to catchup

### LINKED ISSUE #2977: Allow captive core to catchup from multiple archives
Currently, when running core with the `--in-memory` flag with `--start-at-ledger/hash`, we select a single archive to catchup from. The problem is that if the selected archive has issues (it's corrupt or stale or something else), then core ends up being stuck, unable to make progress. 
A better option would be to allow core to select from multiple available archives, similar to what we do during the normal catchup flow. Then in case of a failure, core can retry with a different archive. The fix for this seems very simple, as we need to remove the archive selection code in the in memory mode logic, and pass nullptr instead.

marta-lokhova:
Resolves #2977

### DIFF
--- a/src/main/ApplicationUtils.cpp
+++ b/src/main/ApplicationUtils.cpp
@@ -85,10 +85,8 @@ runWithConfig(Config cfg, optional<CatchupConfiguration> cc)
         if (cc)
         {
             Json::Value catchupInfo;
-            std::shared_ptr<HistoryArchive> archive;
-            auto const& ham = app->getHistoryArchiveManager();
-            archive = ham.selectRandomReadableHistoryArchive();
-            int res = catchup(app, *cc, catchupInfo, archive);
+            // Allow catchup from any readable archive
+            int res = catchup(app, *cc, catchupInfo, /* archive */ nullptr);
             if (res != 0)
             {
                 return res;


graydon (commented):
LGTM!

graydon:
(avoiding r+ since I think tree is closed presently? if it's open again, feel free)

MonsieurNicolas:
This fix works but in a way this seems a bit arbitrary: the separate `catchup` subcommand supports the `--archive` option.
So maybe the right fix is to refactor the code so that we actually share the code between the separate `catchup` sub command and the "catchup +run" that was introduced later for captive core?

MonsieurNicolas:
What do you think @marta-lokhova ?

marta-lokhova:
Not sure I agree @MonsieurNicolas. In-memory live node mode currently doesn't support the `--archive` option at all, and I don't really see a good reason to support it during the live node operation. If captive core really wants to stream meta based on a particular archive, why not do an offline catchup where `--archive` option is available? In live mode, core would end up doing online catchup later anyway in most cases, and that functionality doesn't support `--archive` either (see https://github.com/stellar/stellar-core/blob/ff56e7f207b918aae53710abcc001511ead7ece9/src/catchup/CatchupManagerImpl.cpp#L333)

All I'm trying to do in this change is to prevent the node from getting stuck, which is easy right now as long as there's one bad archive.

MonsieurNicolas:
Yes you are right @marta-lokhova ; I thought there were maybe other options but if this is the only one, then it doesn't make sense

MonsieurNicolas:
r+ 08bd95c3d968e88418e5a5725a25d712a13b9409

PR #2982, opened 2021-03-25
Introduce a mode to delay streaming meta by one ledger

### LINKED ISSUE #2925: add a new mode to mitigate effect of a local node corruption
Something identified as part of hardening analysis.

Currently, in the event of a 1 ledger fork (bad quorum set configuration, or more likely some bug due to local corruption of the core state), the core node will get stuck and notice that it's on a fork *after* externalizing the bad state as it won't be able to reach consensus with the network.
When that happens, the Horizon node (or any dependency), that consumes meta from that core instance will have ingested the bad data from a ledger that does not match the overall network state, and in turn other applications in the ecosystem may perform certain tasks with the expectation of finality of transactions - this is basically impossible to recover from in a sane way.

The solution is to adopt a new operating mode, driven by a configuration element `UNSAFE_EMIT_META` (with a default value of `true` for now), that limits the scope of that 1 ledger fork to the core node and does not allow the information to "leak out" to other parts of the system. The downside is to have a delay of 1 ledger in downstream systems (so perceived latency is longer).

We probably need to talk to the Horizon team a bit to make sure the transition to "safe meta" works as transparently as possible.

In a nutshell, when that mode is enabled:
* captive core (in memory mode) emits meta for the previous ledger when a ledger closes (after checking that hashes match) 
  * note that "command line" modes may have to change a bit as we cannot emit the meta for the last ledger in a range
      * in many cases, we can just "add 1" internally to the last element in the range (when replaying historical data, we can close an extra ledger)
      * for the latest ledger, a possibility to emit it anyways is to provide the hash of the following ledger (recovered offline, from a different core node for example) or just set `UNSAFE_EMIT_META=true`. I am not sure this is really needed though as trying to ingest the "latest" ledger outside of "online" mode probably doesn't make sense (but it may be needed for testing).
* for modes that rely on a persistent database the setting should be ignored/error out
   * for "old style" ingestion (previous Horizon ingestion) dependencies read directly from core's database (and should just delay ingest themselves)
   * for captive core configured with a database (if we ever allow this), we should just not support this for now (and bomb out, people must set "unsafe" in their config)

rokopt:
# Introduce a mode to delay streaming meta by one ledger

Resolves #2925

Introduce a config option affecting streaming metadata (characterized by a non-empty `METADATA_OUTPUT_STREAM`), `UNSAFE_EMIT_META`, which defaults to `true` for now, in which case `stellar-core`'s current behavior does not change.  If `UNSAFE_EMIT_META` is set to `false` (our intended long-term default), which may be done only in conjunction with `--in-memory` (in particular, we intend for this to be used by captive core watcher nodes that stream meta for Horizon), then, while streaming meta online, `stellar-core` defers streaming the meta for ledger `N` until ledger `N+1` closes.  The idea is that in case of a local fault which corrupts a ledger, incorrect meta will never be streamed, as the following ledger will never close on that node, since it will not be able to reach consensus with the network given its corrupt previous ledger.

When streaming meta from the replay of a history archive, `stellar-core` will still emit meta for the latest ledger replayed, as the archive will contain a hash against which that ledger can be validated.

# Checklist
- [x] Compiles
- [x] Ran all tests

### DIFF
--- a/docs/quick-reference.md
+++ b/docs/quick-reference.md
@@ -167,6 +167,22 @@ network's most recent checkpoint, but this behaviour can be further modified usi
 cause the node to start with a fast in-memory catchup to ledger `N` with hash `HEXHASH`, and then
 replay ledgers forward to the current state of the network.
 
+A stateless and meta-streaming node can additionally be configured with
+`UNSAFE_EMIT_META_EARLY=false` (if unspecified, the default is `true`).  If
+`UNSAFE_EMIT_META_EARLY` is `false`, then the node will defer emitting meta for
+a ledger `<N>` until the _next_ ledger, `<N+1>`, closes.  The idea is that if a
+node suffers local corruption in a ledger because of a software bug or hardware
+fault, it will be unable to close the _next_ ledger because it won't be able to
+reach consensus with other nodes. Therefore, the meta for the corrupted ledger
+will never be emitted.  With `UNSAFE_EMIT_META_EARLY` set to `true`, local
+corruption could cause a node to emit meta that were inconsistent with what the
+network reached consensus on.  Setting `UNSAFE_EMIT_META_EARLY` to `false` does
+have a cost, though:  clients waiting for the meta to determine the result of a
+transaction will have to wait for an extra ledger close duration.
+
+During catchup from history archives, a stateless node can emit meta for the
+latest ledger replayed even if `UNSAFE_EMIT_META_EARLY` is `false`, because it
+can validate the latest ledger's hash from the history archive.
 
 #### Publish backlog
 There is a command `publish` that allows to flush the publish backlog without starting
--- a/docs/quick-reference.md
+++ b/docs/quick-reference.md
@@ -168,21 +168,23 @@ cause the node to start with a fast in-memory catchup to ledger `N` with hash `H
 replay ledgers forward to the current state of the network.
 
 A stateless and meta-streaming node can additionally be configured with
-`UNSAFE_EMIT_META_EARLY=false` (if unspecified, the default is `true`).  If
-`UNSAFE_EMIT_META_EARLY` is `false`, then the node will defer emitting meta for
-a ledger `<N>` until the _next_ ledger, `<N+1>`, closes.  The idea is that if a
-node suffers local corruption in a ledger because of a software bug or hardware
-fault, it will be unable to close the _next_ ledger because it won't be able to
-reach consensus with other nodes. Therefore, the meta for the corrupted ledger
-will never be emitted.  With `UNSAFE_EMIT_META_EARLY` set to `true`, local
-corruption could cause a node to emit meta that were inconsistent with what the
-network reached consensus on.  Setting `UNSAFE_EMIT_META_EARLY` to `false` does
-have a cost, though:  clients waiting for the meta to determine the result of a
-transaction will have to wait for an extra ledger close duration.
+`EXPERIMENTAL_PRECAUTION_DELAY_META=true` (if unspecified, the default is
+`false`).  If `EXPERIMENTAL_PRECAUTION_DELAY_META` is `true`, then the node will
+delay emitting meta for a ledger `<N>` until the _next_ ledger, `<N+1>`, closes.
+The idea is that if a node suffers local corruption in a ledger because of a
+software bug or hardware fault, it will be unable to close the _next_ ledger
+because it won't be able to reach consensus with other nodes on the input state
+of the next ledger. Therefore, the meta for the corrupted ledger will never be
+emitted.  With `EXPERIMENTAL_PRECAUTION_DELAY_META` set to `false`, a local
+corruption bug could cause a node to emit meta that is inconsistent with that of
+other nodes on the network. Setting `EXPERIMENTAL_PRECAUTION_DELAY_META` to
+`true` does have a cost, though: clients waiting for the meta to determine the
+result of a transaction will have to wait for an extra ledger close duration.
 
-During catchup from history archives, a stateless node can emit meta for the
-latest ledger replayed even if `UNSAFE_EMIT_META_EARLY` is `false`, because it
-can validate the latest ledger's hash from the history archive.
+During catchup from history archives, a stateless node will emit meta for any
+historical ledger without delay, even if `EXPERIMENTAL_PRECAUTION_DELAY_META` is
+`true`, because the ledger's results are already part of the validated consensus
+history.
 
 #### Publish backlog
 There is a command `publish` that allows to flush the publish backlog without starting
--- a/src/catchup/ApplyCheckpointWork.cpp
+++ b/src/catchup/ApplyCheckpointWork.cpp
@@ -20,6 +20,7 @@
 #include <fmt/format.h>
 #include <medida/meter.h>
 #include <medida/metrics_registry.h>
+#include <optional>
 
 namespace stellar
 {
@@ -241,7 +242,7 @@ ApplyCheckpointWork::getNextLedgerCloseData()
 
     return std::make_shared<LedgerCloseData>(
         header.ledgerSeq, txset, header.scpValue,
-        make_optional<Hash>(mHeaderHistoryEntry.hash));
+        std::make_optional<Hash>(mHeaderHistoryEntry.hash));
 }
 
 BasicWork::State
--- a/src/catchup/ApplyCheckpointWork.cpp
+++ b/src/catchup/ApplyCheckpointWork.cpp
@@ -239,8 +239,9 @@ ApplyCheckpointWork::getNextLedgerCloseData()
     }
 #endif
 
-    return std::make_shared<LedgerCloseData>(header.ledgerSeq, txset,
-                                             header.scpValue);
+    return std::make_shared<LedgerCloseData>(
+        header.ledgerSeq, txset, header.scpValue,
+        make_optional<Hash>(mHeaderHistoryEntry.hash));
 }
 
 BasicWork::State
--- a/src/ledger/LedgerManagerImpl.cpp
+++ b/src/ledger/LedgerManagerImpl.cpp
@@ -589,6 +589,7 @@ LedgerManagerImpl::closeLedger(LedgerCloseData const& ledgerData)
     // LedgerHeader, we optionally collect an even-more-fine-grained record of
     // the ledger entries modified by each tx during tx processing in a
     // LedgerCloseMeta, for streaming to attached clients (typically: horizon).
+    std::unique_ptr<LedgerCloseMeta> ledgerCloseMeta;
     if (mMetaStream)
     {
         if (mNextMetaToEmit)
@@ -597,9 +598,13 @@ LedgerManagerImpl::closeLedger(LedgerCloseData const& ledgerData)
                           getLastClosedLedgerHeader().hash);
             emitNextMeta();
         }
-        mNextMetaToEmit = std::make_unique<LedgerCloseMeta>();
-        mNextMetaToEmit->v0().txProcessing.reserve(txSet->sizeTx());
-        txSet->toXDR(mNextMetaToEmit->v0().txSet);
+        releaseAssert(!mNextMetaToEmit);
+        // Write to a local variable rather than a member variable first: this
+        // enables us to discard incomplete meta and retry, should anything in
+        // this method throw.
+        ledgerCloseMeta = std::make_unique<LedgerCloseMeta>();
+        ledgerCloseMeta->v0().txProcessing.reserve(txSet->sizeTx());
+        txSet->toXDR(ledgerCloseMeta->v0().txSet);
     }
 
     // the transaction set that was agreed upon by consensus
@@ -610,11 +615,11 @@ LedgerManagerImpl::closeLedger(LedgerCloseData const& ledgerData)
     // first, prefetch source accounts fot txset, then charge fees
     prefetchTxSourceIds(txs);
     processFeesSeqNums(txs, ltx, txSet->getBaseFee(header.current()),
-                       mNextMetaToEmit);
+                       ledgerCloseMeta);
 
     TransactionResultSet txResultSet;
     txResultSet.results.reserve(txs.size());
-    applyTransactions(txs, ltx, txResultSet, mNextMetaToEmit);
+    applyTransactions(txs, ltx, txResultSet, ledgerCloseMeta);
 
     ltx.loadHeader().current().txSetResultHash = xdrSha256(txResultSet);
 
@@ -647,9 +652,9 @@ LedgerManagerImpl::closeLedger(LedgerCloseData const& ledgerData)
 
             auto ledgerSeq = ltxUpgrade.loadHeader().current().ledgerSeq;
             LedgerEntryChanges changes = ltxUpgrade.getChanges();
-            if (mNextMetaToEmit)
+            if (ledgerCloseMeta)
             {
-                auto& up = mNextMetaToEmit->v0().upgradesProcessing;
+                auto& up = ledgerCloseMeta->v0().upgradesProcessing;
                 up.emplace_back();
                 UpgradeEntryMeta& uem = up.back();
                 uem.upgrade = lupgrade;
@@ -685,8 +690,14 @@ LedgerManagerImpl::closeLedger(LedgerCloseData const& ledgerData)
 
     if (mMetaStream)
     {
-        releaseAssert(mNextMetaToEmit);
-        mNextMetaToEmit->v0().ledgerHeader = mLastClosedLedger;
+        releaseAssert(ledgerCloseMeta);
+        ledgerCloseMeta->v0().ledgerHeader = mLastClosedLedger;
+
+        // At this point we've got a complete meta and we can store it to the
+        // member variable: if we throw while committing below, we will at worst
+        // emit duplicate meta, when retrying.
+        mNextMetaToEmit = std::move(ledgerCloseMeta);
+
         // If the LedgerCloseData provided an expected hash, then we validated
         // it above.
         if (!mApp.getConfig().EXPERIMENTAL_PRECAUTION_DELAY_META ||
--- a/src/ledger/LedgerManagerImpl.cpp
+++ b/src/ledger/LedgerManagerImpl.cpp
@@ -504,6 +504,16 @@ LedgerManagerImpl::syncMetrics()
     mApp.syncOwnMetrics();
 }
 
+void
+LedgerManagerImpl::emitNextMeta()
+{
+    releaseAssert(mNextMetaToEmit);
+    releaseAssert(mMetaStream);
+    mMetaStream->writeOne(*mNextMetaToEmit);
+    mMetaStream->flush();
+    mNextMetaToEmit.reset();
+}
+
 /*
     This is the main method that closes the current ledger based on
 the close context that was computed by SCP or by the historical module
@@ -579,12 +589,17 @@ LedgerManagerImpl::closeLedger(LedgerCloseData const& ledgerData)
     // LedgerHeader, we optionally collect an even-more-fine-grained record of
     // the ledger entries modified by each tx during tx processing in a
     // LedgerCloseMeta, for streaming to attached clients (typically: horizon).
-    std::unique_ptr<LedgerCloseMeta> ledgerCloseMeta;
     if (mMetaStream)
     {
-        ledgerCloseMeta = std::make_unique<LedgerCloseMeta>();
-        ledgerCloseMeta->v0().txProcessing.reserve(txSet->sizeTx());
-        txSet->toXDR(ledgerCloseMeta->v0().txSet);
+        if (mNextMetaToEmit)
+        {
+            releaseAssert(mNextMetaToEmit->v0().ledgerHeader.hash ==
+                          getLastClosedLedgerHeader().hash);
+            emitNextMeta();
+        }
+        mNextMetaToEmit = std::make_unique<LedgerCloseMeta>();
+        mNextMetaToEmit->v0().txProcessing.reserve(txSet->sizeTx());
+        txSet->toXDR(mNextMetaToEmit->v0().txSet);
     }
 
     // the transaction set that was agreed upon by consensus
@@ -595,11 +610,11 @@ LedgerManagerImpl::closeLedger(LedgerCloseData const& ledgerData)
     // first, prefetch source accounts fot txset, then charge fees
     prefetchTxSourceIds(txs);
     processFeesSeqNums(txs, ltx, txSet->getBaseFee(header.current()),
-                       ledgerCloseMeta);
+                       mNextMetaToEmit);
 
     TransactionResultSet txResultSet;
     txResultSet.results.reserve(txs.size());
-    applyTransactions(txs, ltx, txResultSet, ledgerCloseMeta);
+    applyTransactions(txs, ltx, txResultSet, mNextMetaToEmit);
 
     ltx.loadHeader().current().txSetResultHash = xdrSha256(txResultSet);
 
@@ -632,9 +647,9 @@ LedgerManagerImpl::closeLedger(LedgerCloseData const& ledgerData)
 
             auto ledgerSeq = ltxUpgrade.loadHeader().current().ledgerSeq;
             LedgerEntryChanges changes = ltxUpgrade.getChanges();
-            if (ledgerCloseMeta)
+            if (mNextMetaToEmit)
             {
-                auto& up = ledgerCloseMeta->v0().upgradesProcessing;
+                auto& up = mNextMetaToEmit->v0().upgradesProcessing;
                 up.emplace_back();
                 UpgradeEntryMeta& uem = up.back();
                 uem.upgrade = lupgrade;
@@ -662,12 +677,23 @@ LedgerManagerImpl::closeLedger(LedgerCloseData const& ledgerData)
 
     ledgerClosed(ltx);
 
+    if (ledgerData.getExpectedHash() &&
+        *ledgerData.getExpectedHash() != mLastClosedLedger.hash)
+    {
+        throw std::runtime_error("Local node's ledger corrupted during close");
+    }
+
     if (mMetaStream)
     {
-        releaseAssert(ledgerCloseMeta);
-        ledgerCloseMeta->v0().ledgerHeader = mLastClosedLedger;
-        mMetaStream->writeOne(*ledgerCloseMeta);
-        mMetaStream->flush();
+        releaseAssert(mNextMetaToEmit);
+        mNextMetaToEmit->v0().ledgerHeader = mLastClosedLedger;
+        // If the LedgerCloseData provided an expected hash, then we validated
+        // it above.
+        if (mApp.getConfig().UNSAFE_EMIT_META_EARLY ||
+            ledgerData.getExpectedHash())
+        {
+            emitNextMeta();
+        }
     }
 
     // The next 4 steps happen in a relatively non-obvious, subtle order.
--- a/src/ledger/LedgerManagerImpl.cpp
+++ b/src/ledger/LedgerManagerImpl.cpp
@@ -689,7 +689,7 @@ LedgerManagerImpl::closeLedger(LedgerCloseData const& ledgerData)
         mNextMetaToEmit->v0().ledgerHeader = mLastClosedLedger;
         // If the LedgerCloseData provided an expected hash, then we validated
         // it above.
-        if (mApp.getConfig().UNSAFE_EMIT_META_EARLY ||
+        if (!mApp.getConfig().EXPERIMENTAL_PRECAUTION_DELAY_META ||
             ledgerData.getExpectedHash())
         {
             emitNextMeta();
--- a/src/ledger/LedgerManagerImpl.h
+++ b/src/ledger/LedgerManagerImpl.h
@@ -55,6 +55,8 @@ class LedgerManagerImpl : public LedgerManager
     std::unique_ptr<VirtualClock::time_point> mStartCatchup;
     medida::Timer& mCatchupDuration;
 
+    std::unique_ptr<LedgerCloseMeta> mNextMetaToEmit;
+
     void
     processFeesSeqNums(std::vector<TransactionFrameBasePtr>& txs,
                        AbstractLedgerTxn& ltxOuter, int64_t baseFee,
@@ -74,6 +76,8 @@ class LedgerManagerImpl : public LedgerManager
 
     State mState;
     void setState(State s);
+
+    void emitNextMeta();
 
   protected:
     virtual void transferLedgerEntriesToBucketList(AbstractLedgerTxn& ltx,
--- a/src/ledger/test/LedgerCloseMetaStreamTests.cpp
+++ b/src/ledger/test/LedgerCloseMetaStreamTests.cpp
@@ -28,7 +28,7 @@ TEST_CASE("LedgerCloseMetaStream file descriptor - LIVE_NODE",
     // validator and record metadata streams at the same time (to avoid the
     // unbounded-latency stream-write step): N nodes participating in consensus,
     // and two watching and streaming metadata -- the second one using
-    // !UNSAFE_EMIT_META_EARLY.
+    // EXPERIMENTAL_PRECAUTION_DELAY_META.
 
     Hash expectedLastUnsafeHash, expectedLastSafeHash;
     TmpDirManager tdm(std::string("streamtmp-") + binToHex(randomBytes(8)));
@@ -57,8 +57,11 @@ TEST_CASE("LedgerCloseMetaStream file descriptor - LIVE_NODE",
         SIMULATION_CREATE_NODE(Node2); // Validator
         SIMULATION_CREATE_NODE(Node3); // Validator
 
-        SIMULATION_CREATE_NODE(Node4); // Watcher, UNSAFE_EMIT_META_EARLY
-        SIMULATION_CREATE_NODE(Node5); // Watcher, !UNSAFE_EMIT_META_EARLY
+        // Watcher, !EXPERIMENTAL_PRECAUTION_DELAY_META
+        SIMULATION_CREATE_NODE(Node4);
+
+        // Watcher, EXPERIMENTAL_PRECAUTION_DELAY_META
+        SIMULATION_CREATE_NODE(Node5);
 
         SCPQuorumSet qSet;
         qSet.threshold = 3;
@@ -90,9 +93,9 @@ TEST_CASE("LedgerCloseMetaStream file descriptor - LIVE_NODE",
         cfg5.METADATA_OUTPUT_STREAM = fmt::format("fd:{}", fdSafe);
 #endif
 
-        cfg4.UNSAFE_EMIT_META_EARLY = true;
-        cfg5.UNSAFE_EMIT_META_EARLY = false;
-        cfg5.setInMemoryMode(); // required by !UNSAFE_EMIT_META_EARLY
+        cfg4.EXPERIMENTAL_PRECAUTION_DELAY_META = false;
+        cfg5.EXPERIMENTAL_PRECAUTION_DELAY_META = true;
+        cfg5.setInMemoryMode(); // needed by EXPERIMENTAL_PRECAUTION_DELAY_META
 
         // Step 3: Run simulation a few steps to stream metadata.
         auto app1 = simulation->addNode(vNode1SecretKey, qSet, &cfg1);
@@ -189,8 +192,9 @@ TEST_CASE("LedgerCloseMetaStream file descriptor - LIVE_NODE",
     // The "- 1" is because we don't stream meta for the genesis ledger.
     REQUIRE(lcms.size() == expectedLastWatcherLedger - 1);
     REQUIRE(lcms.back().v0().ledgerHeader.hash == expectedLastUnsafeHash);
-    // The node with !UNSAFE_EMIT_META_EARLY should not have streamed the meta
-    // for the latest ledger (or the latest ledger before the corrupt one) yet.
+    // The node with EXPERIMENTAL_PRECAUTION_DELAY_META should not have streamed
+    // the meta for the latest ledger (or the latest ledger before the corrupt
+    // one) yet.
     REQUIRE(lcmsSafe.size() == lcms.size() - 1);
     REQUIRE(lcmsSafe.back().v0().ledgerHeader.hash == expectedLastSafeHash);
     REQUIRE(lcmsSafe ==
@@ -241,7 +245,7 @@ TEST_CASE("LedgerCloseMetaStream file descriptor - REPLAY_IN_MEMORY",
     cfg1.METADATA_OUTPUT_STREAM = fmt::format("fd:{}", fd);
 #endif
 
-    bool const unsafeEmitMeta = GENERATE(true, false);
+    bool const delayMeta = GENERATE(true, false);
 
     // Step 3: pass it to an application and have it catch up to the generated
     // history, streaming ledgerCloseMeta to the file descriptor.
@@ -252,7 +256,7 @@ TEST_CASE("LedgerCloseMetaStream file descriptor - REPLAY_IN_MEMORY",
         cfg.FORCE_SCP = false;
         cfg.RUN_STANDALONE = true;
         cfg.setInMemoryMode();
-        cfg.UNSAFE_EMIT_META_EARLY = unsafeEmitMeta;
+        cfg.EXPERIMENTAL_PRECAUTION_DELAY_META = delayMeta;
         VirtualClock clock;
         auto app = createTestApplication(clock, cfg, /*newdb=*/false);
 
@@ -276,9 +280,9 @@ TEST_CASE("LedgerCloseMetaStream file descriptor - REPLAY_IN_MEMORY",
     // Step 4: reopen the file as an XDR stream and read back the LCMs
     // and check they have the expected content.
     //
-    // The !UNSAFE_EMIT_META_EARLY case should still have streamed the latest
-    // meta, because catchup should have validated that ledger's hash by
-    // validating a chain of hashes back from one obtained from consensus.
+    // The EXPERIMENTAL_PRECAUTION_DELAY_META case should still have streamed
+    // the latest meta, because catchup should have validated that ledger's hash
+    // by validating a chain of hashes back from one obtained from consensus.
     XDRInputFileStream stream;
     stream.open(path);
     LedgerCloseMeta lcm;
@@ -292,20 +296,19 @@ TEST_CASE("LedgerCloseMetaStream file descriptor - REPLAY_IN_MEMORY",
     REQUIRE(lcm.v0().ledgerHeader.hash == hash);
 }
 
-TEST_CASE("UNSAFE_EMIT_META_EARLY configuration",
+TEST_CASE("EXPERIMENTAL_PRECAUTION_DELAY_META configuration",
           "[ledgerclosemetastreamlive][ledgerclosemetastreamreplay]")
 {
     VirtualClock clock;
     Config cfg = getTestConfig();
 
-    SECTION("UNSAFE_EMIT_META_EARLY may take either value (which is ignored) "
-            "without "
-            "METADATA_OUTPUT_STREAM")
+    SECTION("EXPERIMENTAL_PRECAUTION_DELAY_META may take either value "
+            "(which is ignored) without METADATA_OUTPUT_STREAM")
     {
         cfg.METADATA_OUTPUT_STREAM = "";
-        auto const unsafeEmitMeta = GENERATE(false, true);
+        auto const delayMeta = GENERATE(false, true);
         auto const inMemory = GENERATE(false, true);
-        cfg.UNSAFE_EMIT_META_EARLY = unsafeEmitMeta;
+        cfg.EXPERIMENTAL_PRECAUTION_DELAY_META = delayMeta;
         if (inMemory)
         {
             cfg.setInMemoryMode();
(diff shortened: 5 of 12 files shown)


src/ledger/LedgerManagerImpl.cpp
```diff
@@ -582,6 +595,7 @@ LedgerManagerImpl::closeLedger(LedgerCloseData const& ledgerData)
     std::unique_ptr<LedgerCloseMeta> ledgerCloseMeta;
     if (mMetaStream)
     {
+        emitAnyStoredSafeMetaAfterLedgerValidation();
```
jonjove, line 598: How does this work on the first ledger after restart? Imagine that ledger 73 is closed, then the node restarts before ledger 74 is closed. What happens to the meta for ledger 73? I don't think you are serializing it anywhere, so it seems to be gone. Am I missing something? @MonsieurNicolas initially pointed this out as a potential pitfall in https://github.com/stellar/stellar-core/issues/2925#issuecomment-805398657 so I want to make sure we're handling it.

rokopt: This is what I think happens: https://github.com/stellar/stellar-core/issues/2925#issuecomment-807575702 , but I'll re-check.

src/main/ApplicationImpl.cpp
```diff
@@ -431,6 +431,14 @@ ApplicationImpl::start()
             "RUN_STANDALONE is not set");
     }
 
+    if (mConfig.METADATA_OUTPUT_STREAM != "" && !mConfig.UNSAFE_EMIT_META &&
```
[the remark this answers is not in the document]
rokopt: Sure, I'll add a comment.  This is from Nicolas's description (which sounds right to me) in https://github.com/stellar/stellar-core/issues/2925#issue-811442363 :

> - for modes that rely on a persistent database the setting should be ignored/error out
>   - for "old style" ingestion (previous Horizon ingestion) dependencies read directly from core's database (and should just delay ingest themselves)
>   - for captive core configured with a database (if we ever allow this), we should just not support this for now (and bomb out, people must set "unsafe" in their config)

src/ledger/LedgerManagerImpl.cpp
```diff
@@ -504,6 +504,19 @@ LedgerManagerImpl::syncMetrics()
     mApp.syncOwnMetrics();
 }
 
+void
+LedgerManagerImpl::emitAnyStoredSafeMetaAfterLedgerValidation()
```
graydon, line 508: I'm a little iffy on this name -- maybe `emitDelayedMetaFromLastLedger`?

src/ledger/LedgerManagerImpl.cpp
```diff
@@ -504,6 +504,19 @@ LedgerManagerImpl::syncMetrics()
...
+{
+    if (mLastLedgerMeta)
+    {
+        assert(!mApp.getConfig().UNSAFE_EMIT_META);
```
graydon, line 512: `releaseAssert` on both of these `assert`s

src/ledger/LedgerManagerImpl.h
```diff
@@ -55,6 +55,8 @@ class LedgerManagerImpl : public LedgerManager
     std::unique_ptr<VirtualClock::time_point> mStartCatchup;
     medida::Timer& mCatchupDuration;
 
+    std::unique_ptr<LedgerCloseMeta> mLastLedgerMeta;
```
graydon, line 58: I'd rename this to `mDelayedMetaFromLedger` (to match method name and clarify that it's delayed), and also add a comment relating this to the method and to the config flag.

src/ledger/test/LedgerCloseMetaStreamTests.cpp
```diff
@@ -181,6 +220,10 @@ TEST_CASE("LedgerCloseMetaStream file descriptor - REPLAY_IN_MEMORY",
 
     // Step 4: reopen the file as an XDR stream and read back the LCMs
     // and check they have the expected content.
+    //
+    // The !UNSAFE_EMIT_META case should still have streamed the latest meta,
+    // because catchup should have validated that ledger's hash against the
+    // history archive's version of it.
```
graydon, line 226: Change comment: we don't validate against history archive, we validate history archive contents against user-provided or network-observed trusted hash.

It is also not clear to me that in OFFLINE_COMPLETE catchup, we will actually have streamed the last meta. I don't think we will, given my reading of the code. Not that it matters to users who are passing in `--start-ledger` and `--start-hash`, but in this _particular_ test, I think it should be failing when `unsafeEmitMeta==false`. Maybe double check?

docs/quick-reference.md
```diff
@@ -167,6 +167,22 @@ network's most recent checkpoint, but this behaviour can be further modified usi
 cause the node to start with a fast in-memory catchup to ledger `N` with hash `HEXHASH`, and then
 replay ledgers forward to the current state of the network.
 
+A stateless and meta-streaming node can additionally be configured with
+`UNSAFE_EMIT_META=false` (if unspecified, the default is `true`).  If
```
graydon, line 171: I would prefer not to leave the reader guessing about what exact kind of unsafety there is here (and honestly I prefer the non-double-negative approach also, but nicolas gets to choose that I guess). Can we do maybe `UNSAFE_EMIT_META_EARLY`? Then at least it's clear(er) that the emitting itself isn't unsafe, just doing so earlier than the delayed (safe) case.

graydon (changes requested):
Overall looks pretty good! A few cosmetic fixes plus a couple more-serious testing things:

  - I think one of the tests has a surprising outcome to consider (i.e. if it works I am surprised and I think it probably shouldn't). Noted in the test.
  - Related, in general I'm a bit unclear on the set of conditions under which it's safe to be dropping-on-the-ground the meta for the last ledger in a given operation. I think it might be worth considering which operations generate discrete _ranges_ (specifically those with final points) of ledger meta, and being sure that if we're dropping the last meta in that range, it'll be picked up by the "next" operation instead. For example I think the horizon "full parallel ingestion" mode carves up history into [K, K+100000] ranges to ingest in parallel, and it seems like we might be feeding them only [K, K+99999] in that case.
  - The key scenario -- making sure we don't emit meta for ledger N if if is the result of a one-ledger fork between N-1 and N -- isn't being tested here yet. I realise this may be fairly annoying to test, but I think it's probably worth trying since it's the overarching motivation for the change.

src/ledger/test/LedgerCloseMetaStreamTests.cpp
rokopt: > Change comment: we don't validate against history archive, we validate history archive contents against user-provided or network-observed trusted hash.

Thanks, will do.

> It is also not clear to me that in OFFLINE_COMPLETE catchup, we will actually have streamed the last meta. I don't think we will, given my reading of the code. Not that it matters to users who are passing in --start-ledger and --start-hash, but in this particular test, I think it should be failing when unsafeEmitMeta==false. Maybe double check?

It still passes with `unsafeEmitMeta` forced to `false`, as opposed to using `GENERATE`.  What makes it fail is commenting out that `emitAnyStoredSafeMetaAfterLedgerValidation` (still using iffy name for the moment, will change!) call in `ApplyCheckpointWork::onRun()`:

```cpp
-                lm.emitAnyStoredSafeMetaAfterLedgerValidation();
+                LOG_INFO(DEFAULT_LOG,
+                         "ApplyCheckpointWork::onRun would be calling "
+                         "emitAnyStoredSafeMetaAfterLedgerValidation for "
+                         "ledger {} if it weren't commented out",
+                         lm.getLastClosedLedgerNum());
+                // lm.emitAnyStoredSafeMetaAfterLedgerValidation();
```

If I add another couple log messages, and run the test (which now fails with `emitAnyStoredSafeMetaAfterLedgerValidation` commented out):

```cpp
@@ -509,12 +509,20 @@ LedgerManagerImpl::emitAnyStoredSafeMetaAfterLedgerValidation()
 {
     if (mLedgerCloseMeta)
     {
+        LOG_INFO(DEFAULT_LOG, "Emitting stored metadata for ledger {}",
+                 mLedgerCloseMeta->v0().ledgerHeader.header.ledgerSeq);
         assert(!mApp.getConfig().UNSAFE_EMIT_META);
         assert(mMetaStream);
         mMetaStream->writeOne(*mLedgerCloseMeta);
         mMetaStream->flush();
         mLedgerCloseMeta.reset();
     }
+    else
+    {
+        LOG_INFO(DEFAULT_LOG,
+                 "No stored metadata to emit; last closed ledger {}",
+                 getLastClosedLedgerNum());
+    }
 }
```

```cpp
@@ -594,6 +602,10 @@ LedgerManagerImpl::closeLedger(LedgerCloseData const& ledgerData)
     // LedgerCloseMeta, for streaming to attached clients (typically: horizon).
     if (mMetaStream)
     {
+        LOG_INFO(DEFAULT_LOG,
+                 "closeLedger on ledger {} calling "
+                 "emitAnyStoredSafeMetaAfterLedgerValidation",
+                 header.current().ledgerSeq);
         emitAnyStoredSafeMetaAfterLedgerValidation();
```

Then at the tail of the log I see:

```
2021-03-25T17:56:54.283 <test> [default INFO] closeLedger on ledger 318 calling emitAnyStoredSafeMetaAfterLedgerValidation
2021-03-25T17:56:54.283 <test> [default INFO] Emitting stored metadata for ledger 317
2021-03-25T17:56:54.285 <test> [default INFO] ApplyCheckpointWork::onRun would be calling emitAnyStoredSafeMetaAfterLedgerValidation for ledger 318 if it weren't commented out
2021-03-25T17:56:54.285 <test> [default INFO] closeLedger on ledger 319 calling emitAnyStoredSafeMetaAfterLedgerValidation
2021-03-25T17:56:54.285 <test> [default INFO] Emitting stored metadata for ledger 318
2021-03-25T17:56:54.286 <test> [default INFO] ApplyCheckpointWork::onRun would be calling emitAnyStoredSafeMetaAfterLedgerValidation for ledger 319 if it weren't commented out
2021-03-25T17:56:54.287 <test> [History INFO] Catching up to ledger 319: Download & apply checkpoints: num checkpoints left to apply:0 (100% done)
2021-03-25T17:56:54.287 <test> [History INFO] Catching up to ledger 319: Succeeded: catchup-seq
2021-03-25T17:56:54.287 <test> [History INFO] Catchup finished

[ ... ]

ledger/test/LedgerCloseMetaStreamTests.cpp:236: FAILED:
  REQUIRE( nLcm == 0x13f )
with expansion:
  318 (0x13e) == 319 (0x13f)
```

So the catchup to ledger 319 makes that call (when it's not commented out) for all ledgers up to and including 319.  Not making that call means that the latest meta we've emitted were in `closeLedger` for ledger 319, when we emitted meta for ledger 318.  Here's the log with the real code, with `emitAnyStoredSafeMetaAfterLedgerValidation` restored:

```
2021-03-25T18:01:53.196 <test> [default INFO] closeLedger on ledger 318 calling emitAnyStoredSafeMetaAfterLedgerValidation
2021-03-25T18:01:53.196 <test> [default INFO] No stored metadata to emit; last closed ledger 317
2021-03-25T18:01:53.198 <test> [default INFO] ApplyCheckpointWork::onRun is really calling emitAnyStoredSafeMetaAfterLedgerValidation for ledger 318 (it's not commented out anymore)
2021-03-25T18:01:53.198 <test> [default INFO] Emitting stored metadata for ledger 318
2021-03-25T18:01:53.198 <test> [default INFO] closeLedger on ledger 319 calling emitAnyStoredSafeMetaAfterLedgerValidation
2021-03-25T18:01:53.198 <test> [default INFO] No stored metadata to emit; last closed ledger 318
2021-03-25T18:01:53.200 <test> [default INFO] ApplyCheckpointWork::onRun is really calling emitAnyStoredSafeMetaAfterLedgerValidation for ledger 319 (it's not commented out anymore)
2021-03-25T18:01:53.200 <test> [default INFO] Emitting stored metadata for ledger 319
2021-03-25T18:01:53.201 <test> [History INFO] Catching up to ledger 319: Download & apply checkpoints: num checkpoints left to apply:0 (100% done)
2021-03-25T18:01:53.201 <test> [History INFO] Catching up to ledger 319: Succeeded: catchup-seq
2021-03-25T18:01:53.201 <test> [History INFO] Catchup finished

[ ... ]
All tests passed (8 assertions in 1 test case)
```

So during this test, `ApplyCheckpointWork::onRun` is going through the path that calls `emitAnyStoredSafeMetaAfterLedgerValidation` for each ledger that it closes, including the last.  That causes us still to emit the metadata for the last ledger in this case.  It's never `closeLedger` that emits the meta here, but always `ApplyCheckpointWork`.

src/ledger/LedgerManagerImpl.cpp
rokopt: > > How does this work on the first ledger after restart? Imagine that ledger 73 is closed, then the node restarts before ledger 74 is closed. What happens to the meta for ledger 73? I don't think you are serializing it anywhere, so it seems to be gone. Am I missing something?

> This is what I think happens: #2925 (comment) , but I'll re-check.

Although actually, I don't think Horizon's specific current policy is really relevant.  Captive Core's stateless (hence the requirement for the in-memory DB in the config); what happens after a restart is that Horizon has to ask `stellar-core` to `--start-from` at least as early as the first one that Horizon hasn't already ingested, regardless of what ledger `stellar-core` had last closed at the time of the restart, or what ledger the network is on when the node comes back up.  Even before this change, it's possible for the core to have closed later ledgers when Horizon hasn't ingested earlier ones -- Core's just dumping stuff to a pipe; nothing guarantees that Horizon has consumed it and gotten it all persisted in its own database before Core closes more ledgers.  When there's a restart, Horizon has to look at its database, see what ledger it last ingested, and ask Core to start streaming with the next one, which Core will do.  The difference with this change is that if Horizon asks it to start streaming ledger `N`, then Core will give it the meta from ledger `N` when ledger `N+1` closes, not when ledger `N` closes.  But that doesn't mean it will _skip_ `N`; it just delivers its meta a ledger late, as with all ledgers that it closes.

rokopt:
> * Related, in general I'm a bit unclear on the set of conditions under which it's safe to be dropping-on-the-ground the meta for the last ledger in a given operation. I think it might be worth considering which operations generate discrete _ranges_ (specifically those with final points) of ledger meta, and being sure that if we're dropping the last meta in that range, it'll be picked up by the "next" operation instead. For example I think the horizon "full parallel ingestion" mode carves up history into [K, K+100000] ranges to ingest in parallel, and it seems like we might be feeding them only [K, K+99999] in that case.

It occurs to me that as long as I still have the silly log messages from https://github.com/stellar/stellar-core/pull/2982#discussion_r601939872 , I can also use them to illustrate/explore this case.

If you were going to do a two-threaded ingestion of [192,319], I think the following command lines would express that -- is this right?

```
./stellar-core catchup 255/64 --conf stellar-core-meta-streamer.cfg --in-memory
./stellar-core catchup 319/64 --conf stellar-core-meta-streamer.cfg --in-memory
```

The first one's log:
```
2021-03-25T21:39:30.088 GBYAB [History INFO] Catching up to ledger 255: Download & apply checkpoints: num checkpoints left to apply:1 (0% done)
2021-03-25T21:39:30.156 GBYAB [default INFO] closeLedger on ledger 192 calling emitAnyStoredSafeMetaAfterLedgerValidation
2021-03-25T21:39:30.156 GBYAB [default INFO] No stored metadata to emit; last closed ledger 191
2021-03-25T21:39:30.160 GBYAB [default INFO] ApplyCheckpointWork::onRun is really calling emitAnyStoredSafeMetaAfterLedgerValidation for ledger 192 (it's not commented out anymore)
2021-03-25T21:39:30.160 GBYAB [default INFO] Emitting stored metadata for ledger 192
2021-03-25T21:39:30.160 GBYAB [default INFO] closeLedger on ledger 193 calling emitAnyStoredSafeMetaAfterLedgerValidation
2021-03-25T21:39:30.160 GBYAB [default INFO] No stored metadata to emit; last closed ledger 192
2021-03-25T21:39:30.162 GBYAB [default INFO] ApplyCheckpointWork::onRun is really calling emitAnyStoredSafeMetaAfterLedgerValidation for ledger 193 (it's not commented out anymore)
2021-03-25T21:39:30.162 GBYAB [default INFO] Emitting stored metadata for ledger 193
[ ... ]
2021-03-25T21:39:30.280 GBYAB [default INFO] ApplyCheckpointWork::onRun is really calling emitAnyStoredSafeMetaAfterLedgerValidation for ledger 254 (it's not commented out anymore)
2021-03-25T21:39:30.280 GBYAB [default INFO] Emitting stored metadata for ledger 254
2021-03-25T21:39:30.280 GBYAB [default INFO] closeLedger on ledger 255 calling emitAnyStoredSafeMetaAfterLedgerValidation
2021-03-25T21:39:30.280 GBYAB [default INFO] No stored metadata to emit; last closed ledger 254
2021-03-25T21:39:30.283 GBYAB [default INFO] ApplyCheckpointWork::onRun is really calling emitAnyStoredSafeMetaAfterLedgerValidation for ledger 255 (it's not commented out anymore)
2021-03-25T21:39:30.283 GBYAB [default INFO] Emitting stored metadata for ledger 255
2021-03-25T21:39:30.283 GBYAB [History INFO] Catching up to ledger 255: Download & apply checkpoints: num checkpoints left to apply:0 (100% done)
2021-03-25T21:39:30.283 GBYAB [History INFO] Catching up to ledger 255: Succeeded: catchup-seq
2021-03-25T21:39:30.283 GBYAB [History INFO] Catchup finished
```

The second one's:
```
2021-03-25T21:40:07.521 GBYAB [History INFO] Catching up to ledger 319: Download & apply checkpoints: num checkpoints left to apply:1 (0% done)
2021-03-25T21:40:07.582 GBYAB [default INFO] closeLedger on ledger 256 calling emitAnyStoredSafeMetaAfterLedgerValidation
2021-03-25T21:40:07.582 GBYAB [default INFO] No stored metadata to emit; last closed ledger 255
2021-03-25T21:40:07.587 GBYAB [default INFO] ApplyCheckpointWork::onRun is really calling emitAnyStoredSafeMetaAfterLedgerValidation for ledger 256 (it's not commented out anymore)
2021-03-25T21:40:07.587 GBYAB [default INFO] Emitting stored metadata for ledger 256
2021-03-25T21:40:07.587 GBYAB [default INFO] closeLedger on ledger 257 calling emitAnyStoredSafeMetaAfterLedgerValidation
2021-03-25T21:40:07.587 GBYAB [default INFO] No stored metadata to emit; last closed ledger 256
2021-03-25T21:40:07.588 GBYAB [default INFO] ApplyCheckpointWork::onRun is really calling emitAnyStoredSafeMetaAfterLedgerValidation for ledger 257 (it's not commented out anymore)
2021-03-25T21:40:07.589 GBYAB [default INFO] Emitting stored metadata for ledger 257
[ ... ]
2021-03-25T21:40:07.708 GBYAB [default INFO] closeLedger on ledger 318 calling emitAnyStoredSafeMetaAfterLedgerValidation
2021-03-25T21:40:07.708 GBYAB [default INFO] No stored metadata to emit; last closed ledger 317
2021-03-25T21:40:07.710 GBYAB [default INFO] ApplyCheckpointWork::onRun is really calling emitAnyStoredSafeMetaAfterLedgerValidation for ledger 318 (it's not commented out anymore)
2021-03-25T21:40:07.710 GBYAB [default INFO] Emitting stored metadata for ledger 318
2021-03-25T21:40:07.710 GBYAB [default INFO] closeLedger on ledger 319 calling emitAnyStoredSafeMetaAfterLedgerValidation
2021-03-25T21:40:07.710 GBYAB [default INFO] No stored metadata to emit; last closed ledger 318
2021-03-25T21:40:07.712 GBYAB [default INFO] ApplyCheckpointWork::onRun is really calling emitAnyStoredSafeMetaAfterLedgerValidation for ledger 319 (it's not commented out anymore)
2021-03-25T21:40:07.712 GBYAB [default INFO] Emitting stored metadata for ledger 319
2021-03-25T21:40:07.712 GBYAB [History INFO] Catching up to ledger 319: Download & apply checkpoints: num checkpoints left to apply:0 (100% done)
2021-03-25T21:40:07.712 GBYAB [History INFO] Catching up to ledger 319: Succeeded: catchup-seq
2021-03-25T21:40:07.712 GBYAB [History INFO] Catchup finished
```

So the first command did emit meta for ledger 255, for the same reason as in the unit test -- `ApplyCheckpointWork::onRun` called `emitAnyStoredSafeMetaAfterLedgerValidation`.  And the second command did emit meta for ledger 256.  So I think they're stitching together right, unless I'm misinterpreting what the command lines are supposed to be.  (Avoiding checkpoint boundaries makes it easier to avoid off-by-one errors in the sense that you can get overlap, not gaps.  It is possible to emit some meta twice if you divide up the work on non-checkpoint boundaries, or to emit meta that Horizon already knows about after a restart because we start emitting after a checkpoint, but those things, as far as I can tell, are harmless and nothing new.)

docs/quick-reference.md
rokopt: `UNSAFE_EMIT_META_EARLY` is perfectly fine with me.  (I could probably also be convinced of `EMIT_IRRATIONALLY_OPTIMISTIC_METADATA`.)

src/ledger/LedgerManagerImpl.h
rokopt: That's fine with me, although taking Jon's suggestion in https://github.com/stellar/stellar-core/pull/2982#discussion_r601832293 would change is meaning (it would spend some of its life as the meta from the ledger currently being closed, and might never be delayed as it would be used regardless of whether the config flag was set) and hence its choice of names.

src/ledger/LedgerManagerImpl.cpp
rokopt: Will do.

src/ledger/LedgerManagerImpl.cpp
rokopt: Sure, I'd be happy with that.  (I'm certainly not attached to `emitAnyStoredSafeMetaAfterLedgerValidation`!)

src/ledger/LedgerManagerImpl.cpp
```diff
@@ -582,6 +595,7 @@ LedgerManagerImpl::closeLedger(LedgerCloseData const& ledgerData)
     std::unique_ptr<LedgerCloseMeta> ledgerCloseMeta;
```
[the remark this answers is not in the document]
rokopt: Sure, it's possible -- they have disjoint lifetimes.  And it is shorter without the local.  I'm not convinced it's clearer, though; I find it easier to recognize at a glance what's happening when the names and scopes distinguish "stuff we're in the process of building" from "stuff we've stashed for later".

rokopt:
> * The key scenario -- making sure we don't emit meta for ledger N if if is the result of a one-ledger fork between N-1 and N -- isn't being tested here yet. I realise this may be fairly annoying to test, but I think it's probably worth trying since it's the overarching motivation for the change.

Actively corrupting a node in a `Simulation` is a good idea.  It might actually not be that hard to adapt the test which I've currently made confirm that the "safe" node's meta is always one ledger behind to also overwrite that node's ledger hash at some point and confirm that it never emits any meta for that ledger.  I'll start by trying that approach.

src/ledger/test/LedgerCloseMetaStreamTests.cpp
graydon: Hm. Can I rephrase this from my understanding and you can confirm I'm getting the point?

  - When calling `closeLedger` during live externalize events on the network, the sequence of events is:
    - Externalizing ledger N calls `closeLedger`
    - `closeLedger` calls `emitAnyStored...` which flushes the meta that was pending from N-1
    - `closeLedger` buffers the meta for N
    - Externalizing ledger N+1 calls `closeLedger`
    - `closeLedger` calls `emitAnyStored...` which flushes the meta that was pending from N
    - ...
  
  - Whereas when calling `closeLedger` from `ApplyCheckpointWork::onRun`, the sequence of events is:
    - `ApplyCheckpointWork::onRun` N calls `closeLedger`
    - `closeLedger` calls `emitAnyStored...` which has no pending meta to flush
    - `closeLedger` buffers the meta for N
    - `ApplyCheckpointWork::onRun` calls `emitAnyStored...` _again_ which now flushes the meta for N
    - `ApplyCheckpointWork::onRun` N+1 calls `closeLedger`
    - ...
  
Is that correct? If so I .. can see that it works, in the sense that all the replay scenarios get run end-to-end (because we already trust their end hash value) and all the live scenarios are delayed by 1 ledger.

But: it's fairly subtle, involving two different subsystems (live networking and catchup) having a mental model of a state variable that's being set-and-flushed over successive calls to `closeLedger` in `LedgerManager`. It's subtle enough I completely missed this aspect of the design on first glance over the code! 

I would like to request one of two changes to accompany it:

At minimum a comment very clearly spelling out those two different patterns of calling-and-buffering written in the body of `emitAnyStoredSafeMetaAfterLedgerValidation` (or whatever it's renamed to).

More appealing to me would be an enum-typed flag passed to `closeLedger` that indicates the caller's disposition towards meta, eg. `EMIT_META_IMMEDIATELY` or `DELAY_META_BY_ONE_LEDGER` or something, and keep the state variable private to the `LedgerManager` (i.e. remove `emitAnyStoredSafeMetaAfterLedgerValidation` from the public interface of `LedgerManager`). There's a little complexity in here about how you mediate the relationship between that flag and the config flag (eg. whether the caller looks at config or the callee) but I think it keeps the `closeLedger` API from growing a new dimension of statefulness that a caller could corrupt.

Does that make sense? Am I even understanding the situation correctly?

rokopt: > Is that correct? If so I .. can see that it works, in the sense that all the replay scenarios get run end-to-end (because we already trust their end hash value) and all the live scenarios are delayed by 1 ledger.

I think you have those whole sequences exactly right.  As for how to express it, a mental model requiring the subtlety that you cited...

> But: it's fairly subtle, involving two different subsystems (live networking and catchup) having a mental model of a state variable that's being set-and-flushed over successive calls to `closeLedger` in `LedgerManager`. 

...is what I'm hoping to avoid.  Let me see if I can explain, and you can tell me I'm talking nonsense.  :)

I haven't been thinking about this change (which I found shockingly small at first, but have been gradually convincing myself makes sense, though perhaps wrongly!) _operationally_, as in, this happens and then that happens.  That _does_ require coupling between subsystems and a lot of thought about the state at the time that you're in any given code path.

I've been trying to think about it _declaratively_, as a way of satisfying an invariant.  I want to be able to reason about it purely locally, meaning that it should be possible to:

- Look at any given piece of code and immediately know whether it is a place where we should emit the metadata for any particular ledger or not
- Search the whole code base for some single symbol and know that the only places where it could possibly be necessary to emit metadata must be among the places that use that symbol

And I believe that that _is_ possible, and that:

- The symbol you can search for (at least, one of the symbols – there might be others with the same characteristics?) is `LedgerHeaderHistoryEntry::hash`; in particular, this leads us to `LedgerManagerImpl::closeLedger` and `ApplyCheckpointWork::onRun`
- We should emit metadata if and only if the code is confirming that the last closed ledger's hash matches one which has been validated by a hash chain leading back from a `StellarValue` on which the network reached consensus at some point (Graydon greatly simplified this condition for me by explaining that this is what is happening in the `mApplyLedgerSuccess` case of `ApplyCheckpointWork::onRun`)

The "only if" case is the soundness condition:  that our execution of a transaction set has produced a ledger header with a hash that matches one verified by a hash chain leading back from consensus is what tells us that we executed that transaction set correctly.

The "if" case is the completeness condition, and here, I believe, is why:  the condition that we've verified the last closed ledger's hash against a hash chain leading back from consensus is the same condition that we must meet in order to return success from an operation that is trying to catch up to that ledger.  That is what I think ensures that we don't have off-by-one "missed one ledger's meta" errors here:  if a catchup operation ever returned success from a request to catch up to ledger `N` without having checked the hash resulting from closing ledger `N` against a hash chain, then the catchup operation itself is implemented wrongly.  Therefore, if we emit meta at the place in the code where we've just done that check (and it has passed!), then we will always have emitted meta for ledger `N` before catchup gets a chance to tell its caller that it has successfully caught up to ledger `N`.

Broadening the model beyond the question of when to emit meta, I am thinking of the root cause of the problem that we emit incorrect meta after a one-ledger fork as the existing code's conflating two conditions which are and must be separate:  consensus on the _contents_ of a ledger (represented by a `StellarValue`) and consensus on the _results_ of applying that ledger (of which the meta are one representation, a particularly verbose one for the benefit of Horizon).  It is inevitable that those conditions must be met at different times, because nodes can't even start computing the results of applying a ledger until consensus has been reached on the contents of that ledger.  By emitting meta as soon as we've applied a ledger, we're treating the results as a deterministic, pure function of the contents.  The case of a one-ledger fork caused by a local corruption is the case where that does not hold.

One solution, therefore, would be to have another round of consensus on the _results_ of applying each ledger.  If that phase had short timeouts, it might be the quickest way of reaching a point where we could safely emit meta.  But it would increase network traffic, and also it would take me a hundred years and a thousand bugs to implement.  Fortunately, the `StellarValue` for ledger `N+1` certifies that consensus has been reached not only on the contents of ledger `N+1`, but also on the _results_ of applying ledger `N`, and, via the chain formed by the headers, the results of applying each ledger in `[1..N]`.  That's why the one-ledger-delay implementation provides safety.

So I might have this all wrong, but my conception of this problem, and the proposed solution, is that we're switching to emitting meta for ledger `N` when we reach consensus on the _results_ of applying ledger `N`, instead of when we've only reached consensus on the _contents_ of ledger `N`, as currently written.  That we do so by (when we're in sync with a network) implicitly recognizing that consensus on the contents of ledger `N+1` implies consensus on the results of applying ledger `N` is an implementation detail; we could achieve the same with an explicit consensus round on the results.  But, if I really can get away with it, I would much rather do so just by factoring out one eight-line function and calling it in one new place.  ;)

If I _don't_ have it all wrong and the above is actually true, then I think the next question is, what names/APIs would best express it?

graydon: I can see what you're saying, and I think mostly agree with the clarification to the mental model of consensus you're applying here (though it's maybe worth mentioning that not all forks are caused by local corruption / nondeterminism -- they can also be caused by quorum misconfiguration).

My concern is less to do with mental model of consensus and more to do with more basic code modularity:
  - Callers need to think about a new implicit state variable in `LedgerManager`.
  - The two different types of caller require two different patterns of call (close, close, close vs. close, flush, close, flush).
  - If callers call the flush function at an unexpected time or in the wrong order, they corrupt the state variable.

I'm asking if it's possible to localize the maintenance of _the state variable_ to `LedgerManager`, so callers don't have to know about it / call things in the right order to maintain its invariants. I'd like to keep its invariants private to `LedgerManager` if we can.

Following your comments on mental model, I think this can even be extended to callers lacking awareness of the concept of metadata flushing altogether: maybe the caller only needs to express whether they're providing a ledger that's had its hash confirmed by a subsequent consensus ledger or not. That leaves it up to the `LedgerManager` to interpret that as a sign to flush or delay the metadata emission, and (as in the current code) the `LedgerManager` is the only module that really has any idea there's a metadata stream at all.

Does that make sense?

rokopt: > I think this can even be extended to callers lacking awareness of the concept of metadata flushing altogether: maybe the caller only needs to express whether they're providing a ledger that's had its hash (and txSetResultHash) confirmed by a subsequent consensus ledger or not. That leaves it up to the LedgerManager to interpret that as a sign to flush or delay the metadata emission, and (as in the current code) the LedgerManager is the only module that really has any idea there's a metadata stream at all.

> Does that make sense?

I think so!  Is your idea that `LedgerCloseData` could be given something like a `std::optional<Hash>` which, if it has a value, represents the expected `txResultsHash`, and then `closeLedger` itself could (without callers knowing that this was exactly what it was doing with the information) emit metadata right away if there is an expected `txResultsHash` and if it matches at the end of `closeLedger`, abort right away if there is an expected `txResultsHash` and it _doesn't_ match, and defer emitting metadata if there is no expected `txResultsHash`?  And then `ApplyCheckpointWork` would provide an expected `txResultsHash` from the archive, whereas `valueExternalized` would not provide one?

src/catchup/ApplyCheckpointWork.cpp
```diff
@@ -274,6 +274,8 @@ ApplyCheckpointWork::onRun()
                             lm.getLastClosedLedgerHeader())));
                 }
 
+                lm.emitAnyStoredSafeMetaAfterLedgerValidation();
```
MonsieurNicolas, line 277: I think that the contract with `emitAnyStoredSafeMetaAfterLedgerValidation` is not great:
by splitting (in code) the actual work (`closeLedger`) and how we emit the meta, we provide no guarantees that we're not going to emit some random meta that was stored by a previous call to `closeLedger`.

It looks to me that the way to "idiot proof" this is to instead make it that there is a variant of `LedgerManagerImpl::closeLedger` that exists that takes the expected lcl hash as a parameter, and if present (and matches) we can emit meta right away (sharing the code path with the `UNSAFE_EMIT_META==true`).

src/ledger/LedgerManagerImpl.cpp
MonsieurNicolas, line 598: should there be some invariant check here that the meta that is emitted corresponds to what we check?

src/ledger/LedgerManagerImpl.cpp
```diff
@@ -666,8 +680,15 @@ LedgerManagerImpl::closeLedger(LedgerCloseData const& ledgerData)
     {
         releaseAssert(ledgerCloseMeta);
         ledgerCloseMeta->v0().ledgerHeader = mLastClosedLedger;
-        mMetaStream->writeOne(*ledgerCloseMeta);
-        mMetaStream->flush();
+        if (!mApp.getConfig().UNSAFE_EMIT_META)
```
MonsieurNicolas, line 683: as you're handling both conditions, there is no reason to use the negative here?

MonsieurNicolas (changes requested):

src/ledger/test/LedgerCloseMetaStreamTests.cpp
graydon: Maybe? I mean, again, the `txSetResultHash` matching only traps forks caused by local corruption / nondeterminism, whereas the set of things-that-diverge due to a quorum misconfiguration is a bit wider than that (the other fields in the ledger header -- `baseFee`, `maxTxSetSize`, `ledgerVersion` and so forth). But it might be that we treat passing an expected `txSetResultHash` as implying "we confirmed everything about this ledger, including this `txSetResultHash`, in a subsequent ledger".

If it's too fussy to fish out the `txSetResultHash` from the existing code when doing replay (which I think it might be), I don't think it'd necessarily be any worse to just pass a 2-state flag indicating whether the caller does or doesn't have later-confirmation-confidence in the ledger being closed. Passing a `txSetResultHash` does give a slightly-earlier abort _during_ close if there's a mismatch, but absent that any mismatch from nondeterminism will trap on the next ledger anyway (which will happen immediately during replay), so I don't know that it's worth too much contortion to catch early.

rokopt: > Maybe? I mean, again, the txSetResultHash matching only traps forks caused by local corruption / nondeterminism, whereas the set of things-that-can-diverge due to a quorum misconfiguration is a bit wider than that (the other fields in the ledger header -- baseFee, maxTxSetSize, ledgerVersion and so forth). But it might be that we treat passing an expected txSetResultHash as implying "we confirmed everything about this ledger, including this txSetResultHash, in a subsequent ledger".

Right, thanks -- I should have said the expected ledger header hash; I think that should detect non-determinism from any source -- is that true?

> If it's too fussy to fish out the `txSetResultHash` from the existing code when doing replay (which I think it might be), I don't think it'd necessarily be any worse to just pass a 2-state flag indicating whether the caller does or doesn't have later-confirmation-confidence in the ledger being closed. Passing a `txSetResultHash` does give a slightly-earlier abort _during_ close if there's a mismatch, but absent that any mismatch from nondeterminism will trap on the next ledger anyway (which will happen immediately during replay), so I don't know that it's worth too much contortion to catch early.

The reason I thought it'd have to be a `std::optional<Hash>` is that I'm not following how `LedgerManager` would be able to determine from a flag alone whether it should emit metadata for the ledger that it's closing -- if it knows that the caller knows what the results of closing the ledger should be, but doesn't know what those results _are_, then if it did choose to emit metadata, then even though incorrect results would trap as soon as `closeLedger` returned, wouldn't it already be too late?

src/catchup/ApplyCheckpointWork.cpp
rokopt: Oh, I see now that you'd already suggested something equivalent (I think) to a possibility I asked Graydon about above (https://github.com/stellar/stellar-core/pull/2982#discussion_r602505269) before I'd read your comment :)

src/ledger/LedgerManagerImpl.cpp
rokopt: Yeah, that'd be good -- and I think that this suggestion of yours would give us the parameter that would allow us to do that, as well as making the interface harder to misuse:

> It looks to me that the way to "idiot proof" this is to instead make it that there is a variant of LedgerManagerImpl::closeLedger that exists that takes the expected lcl hash as a parameter, and if present (and matches) we can emit meta right away (sharing the code path with the UNSAFE_EMIT_META==true).

src/ledger/test/LedgerCloseMetaStreamTests.cpp
graydon: > if it knows that the caller knows what the results of closing the ledger should be, but doesn't know what those results are, then if it did choose to emit metadata, then even though incorrect results would trap as soon as closeLedger returned, wouldn't it already be too late?

Oh, yeah, I guess so -- fair point! I hadn't noticed that the flush in the `onRun` caller happens after an expected-hash-check by the caller, thought it was also going to delay noticing until next iteration anyway.

src/ledger/LedgerManagerImpl.cpp
rokopt: True.  Depending on which interfaces from among the current (or future) suggestions we pick, that path may or may not end up with two branches; if it does I'll put the positive branch first.

src/ledger/test/LedgerCloseMetaStreamTests.cpp
rokopt: Okay, I really like the idea of callers giving `closeLedger` an optional expected ledger hash; it makes it explicit that there are two different things we might have consensus on already:  the contents (represented by a `StellarValue`), which we _definitely_ have consensus on or we wouldn't be calling `closeLedger`, and the results (represented by the value that, if present, we expect the ledger hash to have after the close is complete), which we might or might not have consensus on already, depending on the context in which we're being called.  `closeLedger` becomes "externalize contents of ledger `N`, and _maybe_ externalize results of ledger N".  Also, `closeLedger(N)` always implies "externalize results of ledgers `[1..N-1]`".  And what `LedgerManagerImpl` privately chooses to do with "externalize contents" is apply transactions and upgrades; what it does with "externalize results" is stream meta.

My first thought was to stick an `optional<Hash>` into `LedgerCloseData`; @MonsieurNicolas 's was to provide an overload of `closeLedger` with a `Hash` parameter.  Is there a reason to prefer one or the other clearly?

src/main/ApplicationImpl.cpp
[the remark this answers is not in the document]
rokopt: Comment added.

src/ledger/LedgerManagerImpl.cpp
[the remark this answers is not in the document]
rokopt: I have taken this suggestion.

src/ledger/LedgerManagerImpl.cpp
rokopt: (This method is private now, and can now be called to emit either delayed or immediate metadata, so I've just given it the name `emitNextMeta`.)

src/ledger/LedgerManagerImpl.cpp
rokopt: Done.

src/ledger/LedgerManagerImpl.h
rokopt: I did take Jon's suggestion, so that variable can now refer to current as well as delayed meta.  So I'm calling it `mNextMetaToEmit`.

docs/quick-reference.md
rokopt: Changed to add `_EARLY`.

rokopt:
> > * The key scenario -- making sure we don't emit meta for ledger N if if is the result of a one-ledger fork between N-1 and N -- isn't being tested here yet. I realise this may be fairly annoying to test, but I think it's probably worth trying since it's the overarching motivation for the change.
> 
> Actively corrupting a node in a `Simulation` is a good idea. It might actually not be that hard to adapt the test which I've currently made confirm that the "safe" node's meta is always one ledger behind to also overwrite that node's ledger hash at some point and confirm that it never emits any meta for that ledger. I'll start by trying that approach.

I think I managed to test this now:  I've extended the livestream test so that it has a case in which it calls a spurious `closeLedger` on the watchers, and confirms that only the `UNSAFE_EMIT_META_EARLY` node emits meta for that spuriously-closed ledger.  (Both of them get stuck after that, of course.)

src/catchup/ApplyCheckpointWork.cpp
rokopt: I've made this API change now (I put the new hash parameter into `LedgerCloseData`).

src/ledger/LedgerManagerImpl.cpp
rokopt: I've added a `releaseAssert` there.

src/ledger/LedgerManagerImpl.cpp
rokopt: (In light of other changes, there's now only one branch anyway.)

rokopt:
I've squashed and rebased because the total volume of code is small and all the API names and signatures changed, so I figured it would be easier to review from scratch than from the first set of changes.

graydon:
Rebased and adapted to recent churn in C++17 and in-memory mode.

Going to sleep and let tests cycle now, will give it a final pass of review with fresh eyes tomorrow. @MonsieurNicolas or @jonjove might also be good for you to give it another pass if you have the time.

graydon (approved):

graydon:
I've re-reviewed all the comments on this PR (and marked all mine as resolved), and read the tests and am fairly confident they're doing what they need to do. I believe all the @jonjove and @MonsieurNicolas comments are also addressed though I've not marked them as resolved. Could y'all mark your comments as resolved / approve if so?

Of course there's still the possibility that somehow this will confuse horizon, or we've changed the contract on the command line args somehow despite all signs suggesting not; but I'm happy to land this now. We also need to decide if it's ok to.

src/ledger/LedgerManagerImpl.cpp
```diff
@@ -579,12 +589,17 @@ LedgerManagerImpl::closeLedger(LedgerCloseData const& ledgerData)
...
-        ledgerCloseMeta = std::make_unique<LedgerCloseMeta>();
-        ledgerCloseMeta->v0().txProcessing.reserve(txSet->sizeTx());
-        txSet->toXDR(ledgerCloseMeta->v0().txSet);
+        if (mNextMetaToEmit)
+        {
+            releaseAssert(mNextMetaToEmit->v0().ledgerHeader.hash ==
+                          getLastClosedLedgerHeader().hash);
+            emitNextMeta();
+        }
+        mNextMetaToEmit = std::make_unique<LedgerCloseMeta>();
```
MonsieurNicolas, line 600: This is not exception safe:
it's entirely possible that `LedgerManagerImpl::closeLedger` gets called and throws after this point - I think if this happens the caller will just retry (if this is not possible right now, writing the code this way is a clear footgun if we ever want to change that).

So what we want to do here is consume `mNextMetaToEmit` like it is now (consensus was reached so we can emit the meta from the previous ledger), and continue to use a local variable `ledgerCloseMeta` like before, and only at the end move it into `mNextMetaToEmit` when it is safe to do so.

src/ledger/LedgerManagerImpl.cpp
```diff
@@ -595,11 +610,11 @@ LedgerManagerImpl::closeLedger(LedgerCloseData const& ledgerData)
     // first, prefetch source accounts fot txset, then charge fees
     prefetchTxSourceIds(txs);
     processFeesSeqNums(txs, ltx, txSet->getBaseFee(header.current()),
-                       ledgerCloseMeta);
+                       mNextMetaToEmit);
```
MonsieurNicolas, line 613: see my other comment: it's better to continue to use a local variable here

src/ledger/LedgerManagerImpl.cpp
```diff
@@ -662,12 +677,23 @@ LedgerManagerImpl::closeLedger(LedgerCloseData const& ledgerData)
...
-        mMetaStream->flush();
+        releaseAssert(mNextMetaToEmit);
+        mNextMetaToEmit->v0().ledgerHeader = mLastClosedLedger;
+        // If the LedgerCloseData provided an expected hash, then we validated
+        // it above.
+        if (mApp.getConfig().UNSAFE_EMIT_META_EARLY ||
+            ledgerData.getExpectedHash())
+        {
+            emitNextMeta();
+        }
```
MonsieurNicolas, line 696: I would change the code to set `mNextMetaToEmit` in the else here and use `ledgerCloseMeta` before

MonsieurNicolas (changes requested):

src/ledger/LedgerManagerImpl.cpp
graydon: Fixed.

src/ledger/LedgerManagerImpl.cpp
graydon: (also fixed)

src/ledger/LedgerManagerImpl.cpp
graydon: (also fixed)

MonsieurNicolas:
r+ 152683e54d12bdf7de83fb340489ba4001eadb43

PR #2987, opened 2021-03-30
fix confusing error message when NODE_HOME_DOMAIN is not set

MonsieurNicolas:


NickJH:
Is this correct? From my testing both `HOME_DOMAIN` and `NODE_HOME_DOMAIN` must be set **and** the same or you cannot do a new-db It seems like the waste of a parameter.

MonsieurNicolas:
yes this is correct.

> It seems like the waste of a parameter

this is because `HOME_DOMAINS` is how validators in general are configured, not just the local one

NickJH:
There is a similar problem with the sample config at https://github.com/stellar/packages/blob/master/docs/examples/pubnet-validator-full/stellar-core.cfg. It shows `NODE_HOME_DOMAIN="example.com"` which is effectively a prompt to set your `NODE_HOME_DOMAIN`, but the sample config still fails unless you set a `HOME_DOMAIN` in a `[[HOME_DOMAINS]]` section.

MonsieurNicolas:
Good catch @NickJH - I opened a PR https://github.com/stellar/packages/pull/143 to fix this in the other repo

I actually tweaked this PR to instead not require `NODE_HOME_DOMAIN` when setting up a single validator; it's only needed when building out "full public validators" (that need to be redundant)

MonsieurNicolas:
Updated again.

Current `master` requires `NODE_HOME_DOMAIN` and to have a corresponding `[[HOME_DOMAINS]]` section - the problem is that the error message is super confusing (bombs out with `Must specify a matching HOME_DOMAINS for self`).

At first I thought I would just make it optional, but remembered (thanks to @marta-lokhova ) that I made it mandatory in the first place to avoid a foot-gun: most validators are actually built with redundancy in mind, and if they forget to set `NODE_HOME_DOMAIN`, it would be equivalent to a "silent failure" or sorts where they would have registered their nodes as separate entities instead of a single entity (this was actually one of the reasons we had a network halt in 2019).

So at the end, what I did was just add better error messages:
* when not set, force people to set it
* explain that there has to be a matching `HOME_DOMAINS` section

src/main/Config.cpp
```diff
@@ -649,20 +649,28 @@ Config::addSelfToValidators(
     std::vector<ValidatorEntry>& validators,
     UnorderedMap<std::string, ValidatorQuality> const& domainQualityMap)
 {
-    auto it = domainQualityMap.find(NODE_HOME_DOMAIN);
+    if (NODE_HOME_DOMAIN.empty())
+    {
+        throw std::invalid_argument("NODE_HOME_DOMAIN must be set");
```
marta-lokhova, line 654: Commit name is wrong now

MonsieurNicolas:
@marta-lokhova updated! Thank you

marta-lokhova:
r+ 59feabecf8c85a0e9a21b9f21b0158927eed0398

PR #2989, opened 2021-03-31
Migrate to C++17

graydon:
This is just a collection of changes to move us to C++17. Hopefully nothing too controversial.


src/util/optional.h
```diff
@@ -6,23 +6,26 @@
...
-template <typename T> using optional = std::shared_ptr<T>;
 
-template <typename T, class... Args>
-optional<T>
-make_optional(Args&&... args)
+template <class T> using optional = std::optional<T>;
+
+template <typename T>
+constexpr optional<T>
+make_optional(T const& arg)
```
MonsieurNicolas, line 19: why do we need our own `make_optional` ?

src/herder/test/HerderTests.cpp
```diff
@@ -1259,7 +1259,7 @@ testSCPDriver(uint32 protocolVersion, uint32_t maxTxSize, size_t expectedOps)
         // combineCandidates() and the one we compute at each step.
 
         std::vector<CandidateSpec> const specs{
-            {0, 1, 100, 10, make_optional<uint32>()},
+            {0, 1, 100, 10, std::optional<uint32>()},
```
MonsieurNicolas, line 1262: this change looks odd
* should change all to `std::make_optional`
* `std::make_optional<uint32>()` should just work

src/util/optional.h
```diff
@@ -6,23 +6,25 @@
 
 #include "cereal/cereal.hpp"
 #include <memory>
+#include <optional>
```
MonsieurNicolas, line 9: any reason to keep this file? This was our poor man's implementation of `std::optional`

m4/ax_fresh_compiler.m4
```diff
@@ -47,7 +47,6 @@ AC_DEFUN([AX_FRESH_COMPILER],
                 auto a = sizeof(std::max_align_t);
                 auto b = std::get_new_handler();
                 auto c = std::get_terminate();
-                auto d = std::get_unexpected();
```
MonsieurNicolas, line 50: did you just remove this or picked up a more recent version of this file?

lib/soci/src/core/session.h
```diff
@@ -77,8 +77,8 @@ class SOCI_DECL session
     template <typename T>
     void set_query_transformation(T callback)
     {
-        std::auto_ptr<details::query_transformation_function> qtf(new details::query_transformation<T>(callback));
-        set_query_transformation_(qtf);
+        std::unique_ptr<details::query_transformation_function> qtf(new details::query_transformation<T>(callback));
```
MonsieurNicolas, line 80: the c++ way is `make_unique`

MonsieurNicolas (changes requested):
I reviewed based on what I can see in the diff, which is probably close enough

src/util/optional.h
graydon: I'll remove this and the residual `util/optional.h` file

src/util/optional.h
graydon: removing

src/herder/test/HerderTests.cpp
graydon: (yes, it does)

m4/ax_fresh_compiler.m4
graydon: We wrote this file! But std::get_unexpected was removed in C++17.

lib/soci/src/core/session.h
graydon: in fairness: this hunk is [closer to upstream](https://github.com/SOCI/soci/blob/master/include/soci/session.h#L89), which somewhat astoundingly provides a `using` alias [for their own `cxx_details::auto_ptr`](https://github.com/SOCI/soci/blob/master/include/soci/soci-platform.h#L120-L125) that, of course, _changes semantics_ between `auto_ptr` and `unique_ptr` depending on whether it's post-C++11. WHAT COULD GO WRONG?

Anyway I think for the sake of smaller diffs against upstream, we ought to leave this hunk alone.

src/util/optional.h
MonsieurNicolas: thanks!

lib/soci/src/core/session.h
MonsieurNicolas: ok, I guess that library is doomed anyways if constructors throw for whatever reason

lib/autocheck/include/autocheck/apply.hpp
```diff
@@ -23,14 +23,6 @@ namespace autocheck {
     typedef typename std::result_of<F(Args&...)>::type type;
   };
 
-  template <typename F, typename Tuple>
```
MonsieurNicolas, line 26: why is this not needed? I don't see this change in upstream... and I actually don't understand how it works without this.

Also, we should avoid deviating from upstream if we can.

rokopt: In C++17 there's a `std::apply` which looks to me as though it does the same thing (in different words):

https://en.cppreference.com/w/cpp/utility/apply

So with the `autocheck` `apply` deleted, we just use the STL one, whereas with it present, we get:

```
In file included from crypto/test/CryptoTests.cpp:17:
In file included from ../lib/autocheck/include/autocheck/autocheck.hpp:4:
../lib/autocheck/include/autocheck/check.hpp:48:12: error: call to 'apply' is ambiguous
      if (!apply(prop, args.ref())) {
           ^~~~~
../lib/autocheck/include/autocheck/check.hpp:59:5: note: in instantiation of function template specialization 'autocheck::check<std::vector<unsigned char>, (lambda at crypto/test/CryptoTests.cpp:62:9), autocheck::arbitrary<autocheck::generator<std::vector<unsigned char>>>>' requested here
    check<Args...>(std::move(prop),
    ^
crypto/test/CryptoTests.cpp:61:16: note: in instantiation of function template specialization 'autocheck::check<std::vector<unsigned char>, (lambda at crypto/test/CryptoTests.cpp:62:9)>' requested here
    autocheck::check<std::vector<uint8_t>>(
               ^
/usr/bin/../lib/gcc/x86_64-linux-gnu/10/../../../../include/c++/10/tuple:1729:5: note: candidate function [with _Fn = (lambda at crypto/test/CryptoTests.cpp:62:9) &, _Tuple = std::tuple<std::vector<unsigned char>> &]
    apply(_Fn&& __f, _Tuple&& __t)
    ^
../lib/autocheck/include/autocheck/apply.hpp:28:3: note: candidate function [with F = (lambda at crypto/test/CryptoTests.cpp:62:9), Tuple = std::tuple<std::vector<unsigned char>> &]
  apply(F f, Tuple&& args) {
  ^
```

Other options besides just deleting that include updating to upstream en masse and isolating just whatever they did to fix this and pulling that change.

MonsieurNicolas: Yeah we should first try upstream, and if it doesn't work look into rolling our own.

It looks to me that this entire file should go away (as everything in here looks to exist as building blocks for `apply`), potentially more.

graydon: Upstream has not altered this file since 2015. Unfortunately it's fairly dormant. I don't think it compiles on C++17 by default.

MonsieurNicolas: OK, so I guess we can just make it work for C++ 17. From what I see it's basically deleting this file and use `std::apply` instead

graydon: @MonsieurNicolas done

MonsieurNicolas (approved):
0 warnings in VC++, the way I like it :)

MonsieurNicolas:
r+ ad0163dcf90aa4f6deae3fc3518615f13a94068b

PR #2990, opened 2021-03-31
Fix logging issues

### LINKED ISSUE #2752: perf issues in easylogging
As I was debugging a test case that uses the `Simulation` class (runs multiple nodes), I added some logging code to change the log context to log which node in the simulation is doing work.

With this patch, simulation tests are extremely slow, and a quick profiling shows that tests spend 80% of their time in the function that updates sinks in easylogging (looks like some broken parsing code to me).

Patch follows (as of 5626a895038055c7d1be30522e9e79881fd38792 ):

```
diff --git a/src/simulation/Simulation.cpp b/src/simulation/Simulation.cpp
index 7f4f998a..4f34103c 100644
--- a/src/simulation/Simulation.cpp
+++ b/src/simulation/Simulation.cpp
@@ -378,6 +378,8 @@ Simulation::crankAllNodes(int nbTicks)
 
     VirtualTimer mainQuantumTimer(*mIdleApp);
 
+    bool debugFmt = Logging::logDebug("Process");
+
     int i = 0;
     do
     {
@@ -390,6 +392,10 @@ Simulation::crankAllNodes(int nbTicks)
 
         if (mClock.getIOContext().stopped())
         {
+            if (debugFmt)
+            {
+                Logging::setFmt("<test>");
+            }
             return 0;
         }
 
@@ -436,6 +442,13 @@ Simulation::crankAllNodes(int nbTicks)
                         continue;
                     }
                 }
+
+                if (debugFmt)
+                {
+                    Logging::setFmt(fmt::format(
+                        "<test-{}>", p.second.mApp->getConfig().PEER_PORT));
+                }
+
                 crankNode(p.first, nextTime);
             }
         } while (appBehind);
@@ -454,6 +467,12 @@ Simulation::crankAllNodes(int nbTicks)
             i++;
         }
     } while (i < nbTicks);
+
+    if (debugFmt)
+    {
+        // restore the default context
+        Logging::setFmt("<test>");
+    }
+
     return count;
 }
```

MonsieurNicolas:
# Description

Resolves #2752

This PR makes logging easier to follow when running simulation by setting the log format string based off  the node in the simulation.

I also fixed a couple bugs in `logDebug` and `logTrace` (were checking conditions backward and were broken when setting log levels in partitions) as well as removed using them (this was made obsolete with the switch to spdlog)

### DIFF
--- a/src/herder/HerderImpl.cpp
+++ b/src/herder/HerderImpl.cpp
@@ -159,12 +159,8 @@ HerderImpl::processExternalized(uint64 slotIndex, StellarValue const& value)
     ZoneScoped;
     bool validated = getSCP().isSlotFullyValidated(slotIndex);
 
-    if (Logging::logDebug("Herder"))
-    {
-        CLOG_DEBUG(Herder,
-                   "HerderSCPDriver::valueExternalized index: {} txSet: {}",
-                   slotIndex, hexAbbrev(value.txSetHash));
-    }
+    CLOG_DEBUG(Herder, "HerderSCPDriver::valueExternalized index: {} txSet: {}",
+               slotIndex, hexAbbrev(value.txSetHash));
 
     if (getSCP().isValidator() && !validated)
     {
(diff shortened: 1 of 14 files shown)


graydon (approved):
LGTM!

graydon:
r+ d6f6df1

PR #2994, opened 2021-04-02
Captive core fast startup

### LINKED ISSUE #2960: Consider restoring previous captive core state on startup instead of catchup
As suggested by @bartekn, consider skipping the download of a particular bucket if it's already present in the buckets folder. This would be useful, as captive core has to re-download all the buckets upon re-start. Note that we should still verify buckets, to ensure that we are not processing invalid buckets.

### LINKED ISSUE #2993: Captive core should remember peers in between restarts
Captive core is “all in memory”, including its overlay state. As a consequence, every restart of captive core requires rediscovering good quality peers. This in turn may cause arbitrary issues during operations. We may want to consider allowing captive core to remember its peers in between restarts. This seems inline with the changes we're introducing in #2960

marta-lokhova:
Implementation based on the discussion in #2960 

Resolves #2960 
Resolves #2993

### DIFF
--- a/src/bucket/BucketManagerImpl.cpp
+++ b/src/bucket/BucketManagerImpl.cpp
@@ -107,7 +107,10 @@ BucketManagerImpl::BucketManagerImpl(Application& app)
     , mBucketSnapMerge(app.getMetrics().NewTimer({"bucket", "snap", "merge"}))
     , mSharedBucketsSize(
           app.getMetrics().NewCounter({"bucket", "memory", "shared"}))
-    , mDeleteEntireBucketDirInDtor(app.getConfig().isInMemoryMode())
+    // Minimal DB is stored in the buckets dir, so delete it only when
+    // mode does not use minimal DB
+    , mDeleteEntireBucketDirInDtor(
+          app.getConfig().isInMemoryModeWithoutMinimalDB())
 {
 }
 
--- a/src/herder/test/HerderTests.cpp
+++ b/src/herder/test/HerderTests.cpp
@@ -2586,7 +2586,7 @@ TEST_CASE("slot herder policy", "[herder]")
     Config cfg(getTestConfig());
 
     // start in sync
-    cfg.FORCE_SCP = true;
+    cfg.FORCE_SCP = false;
     cfg.MANUAL_CLOSE = false;
     cfg.NODE_SEED = v0SecretKey;
     cfg.MAX_SLOTS_TO_REMEMBER = 5;
@@ -2599,6 +2599,7 @@ TEST_CASE("slot herder policy", "[herder]")
 
     VirtualClock clock;
     Application::pointer app = createTestApplication(clock, cfg);
+    app->start();
 
     auto& herder = static_cast<HerderImpl&>(app->getHerder());
 
--- a/src/main/ApplicationImpl.cpp
+++ b/src/main/ApplicationImpl.cpp
@@ -40,6 +40,7 @@
 #include "medida/timer.h"
 #include "overlay/BanManager.h"
 #include "overlay/OverlayManager.h"
+#include "overlay/OverlayManagerImpl.h"
 #include "process/ProcessManager.h"
 #include "scp/LocalNode.h"
 #include "scp/QuorumSetUtils.h"
@@ -398,6 +399,18 @@ ApplicationImpl::~ApplicationImpl()
     LOG_INFO(DEFAULT_LOG, "Application destroyed");
 }
 
+void
+ApplicationImpl::resetDBForInMemoryMode()
+{
+    // Load the peer information and reinitialize the DB
+    auto& pm = getOverlayManager().getPeerManager();
+    auto peerData = pm.loadAllPeers();
+    newDB();
+    pm.storePeers(peerData);
+
+    LOG_INFO(DEFAULT_LOG, "In-memory state is reset back to genesis");
+}
+
 uint64_t
 ApplicationImpl::timeNow()
 {
@@ -449,9 +462,7 @@ ApplicationImpl::validateAndLogConfig()
         {
             throw std::invalid_argument(
                 "Core is not configured to store history, but "
-                "some history archives are writable (see "
-                "MODE_STORES_HISTORY_MISC "
-                "and MODE_STORES_HISTORY_LEDGERHEADERS)");
+                "some history archives are writable");
         }
     }
 
--- a/src/main/ApplicationImpl.cpp
+++ b/src/main/ApplicationImpl.cpp
@@ -401,21 +401,8 @@ ApplicationImpl::timeNow()
 }
 
 void
-ApplicationImpl::start()
+ApplicationImpl::validateAndLogConfig()
 {
-    if (mStarted)
-    {
-        CLOG_INFO(Ledger, "Skipping application start up");
-        return;
-    }
-    CLOG_INFO(Ledger, "Starting up application");
-    mStarted = true;
-
-    if (mConfig.TESTING_UPGRADE_DATETIME.time_since_epoch().count() != 0)
-    {
-        mHerder->setUpgrades(mConfig);
-    }
-
     if (mConfig.FORCE_SCP && !mConfig.NODE_IS_VALIDATOR)
     {
         throw std::invalid_argument(
@@ -467,6 +454,23 @@ ApplicationImpl::start()
     }
 
     mConfig.logBasicInfo();
+}
+
+void
+ApplicationImpl::start()
+{
+    if (mStarted)
+    {
+        CLOG_INFO(Ledger, "Skipping application start up");
+        return;
+    }
+    CLOG_INFO(Ledger, "Starting up application");
+    mStarted = true;
+
+    if (mConfig.TESTING_UPGRADE_DATETIME.time_since_epoch().count() != 0)
+    {
+        mHerder->setUpgrades(mConfig);
+    }
 
     bool done = false;
     mLedgerManager->loadLastKnownLedger([this,
--- a/src/main/ApplicationImpl.cpp
+++ b/src/main/ApplicationImpl.cpp
@@ -445,10 +445,13 @@ ApplicationImpl::validateAndLogConfig()
 
     if (getHistoryArchiveManager().hasAnyWritableHistoryArchive())
     {
-        if (!mConfig.MODE_STORES_HISTORY)
+        if (!mConfig.modeStoresAllHistory())
         {
-            throw std::invalid_argument("MODE_STORES_HISTORY is not set, but "
-                                        "some history archives are writable");
+            throw std::invalid_argument(
+                "Core is not configured to store history, but "
+                "some history archives are writable (see "
+                "MODE_STORES_HISTORY_MISC "
+                "and MODE_STORES_HISTORY_LEDGERHEADERS)");
         }
     }
 
--- a/src/main/ApplicationImpl.cpp
+++ b/src/main/ApplicationImpl.cpp
@@ -135,13 +135,24 @@ ApplicationImpl::ApplicationImpl(VirtualClock& clock, Config const& cfg)
 void
 ApplicationImpl::initialize(bool createNewDB)
 {
+    // Subtle: initialize the bucket manager first before initializing the
+    // database. This is needed as some modes in core (such as in-memory) use a
+    // small database inside the bucket directory.
+    mBucketManager = BucketManager::create(*this);
+
+    bool initNewDB =
+        createNewDB || mConfig.DATABASE.value == "sqlite3://:memory:";
+    if (initNewDB)
+    {
+        mBucketManager->dropAll();
+    }
+
     mDatabase = createDatabase();
     mPersistentState = std::make_unique<PersistentState>(*this);
     mOverlayManager = createOverlayManager();
     mLedgerManager = createLedgerManager();
     mHerder = createHerder();
     mHerderPersistence = HerderPersistence::create(*this);
-    mBucketManager = BucketManager::create(*this);
     mCatchupManager = CatchupManager::create(*this);
     mHistoryArchiveManager = std::make_unique<HistoryArchiveManager>(*this);
     mHistoryManager = HistoryManager::create(*this);
@@ -192,13 +203,13 @@ ApplicationImpl::initialize(bool createNewDB)
     SponsorshipCountIsValid::registerInvariant(*this);
     enableInvariantsFromConfig();
 
-    if (createNewDB || mConfig.DATABASE.value == "sqlite3://:memory:")
+    if (initNewDB)
     {
         newDB();
     }
     else
     {
-        upgradeDB();
+        mDatabase->upgradeToCurrentSchema();
     }
 
     // Subtle: process manager should come to existence _after_ BucketManager
@@ -213,14 +224,7 @@ ApplicationImpl::newDB()
 {
     mDatabase->initialize();
     mDatabase->upgradeToCurrentSchema();
-    mBucketManager->dropAll();
     mLedgerManager->startNewLedger();
-}
-
-void
-ApplicationImpl::upgradeDB()
-{
-    mDatabase->upgradeToCurrentSchema();
 }
 
 void
--- a/src/main/ApplicationUtils.cpp
+++ b/src/main/ApplicationUtils.cpp
@@ -7,12 +7,14 @@
 #include "bucket/BucketManager.h"
 #include "catchup/ApplyBucketsWork.h"
 #include "catchup/CatchupConfiguration.h"
+#include "crypto/Hex.h"
 #include "database/Database.h"
 #include "herder/Herder.h"
 #include "history/HistoryArchive.h"
 #include "history/HistoryArchiveManager.h"
 #include "history/HistoryArchiveReportWork.h"
 #include "historywork/GetHistoryArchiveStateWork.h"
+#include "ledger/LedgerHeaderUtils.h"
 #include "ledger/LedgerManager.h"
 #include "main/ErrorMessages.h"
 #include "main/ExternalQueue.h"
@@ -20,6 +22,7 @@
 #include "main/PersistentState.h"
 #include "main/StellarCoreVersion.h"
 #include "overlay/OverlayManager.h"
+#include "util/GlobalChecks.h"
 #include "util/Logging.h"
 #include "work/WorkScheduler.h"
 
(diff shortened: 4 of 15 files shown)


src/main/CommandLine.cpp
```diff
@@ -281,10 +281,17 @@ maybeSetMetadataOutputStream(Config& cfg, std::string const& stream)
...
+{
+    return fmt::format("sqlite3://{}/partial.db", cfg.BUCKET_DIR_PATH);
+}
+
 optional<CatchupConfiguration>
 maybeEnableInMemoryLedgerMode(Config& config, bool inMemory,
                               uint32_t startAtLedger,
-                              std::string const& startAtHash)
+                              std::string const& startAtHash,
+                              bool persistPartially = false)
```
MonsieurNicolas, line 294: do not use default value here

src/main/CommandLine.cpp
```diff
@@ -916,11 +939,26 @@ int
...
 {
     CommandLine::ConfigOption configOption;
+    bool resetInMemoryState = false;
 
-    return runWithHelp(args, {configurationParser(configOption)}, [&] {
-        initializeDatabase(configOption.getConfig());
-        return 0;
-    });
+    auto resetParser = [](bool& resetInMemoryState) {
+        return clara::Opt{resetInMemoryState}["--reset-for-in-memory-mode"](
```
MonsieurNicolas, line 945: should use `inMemoryParser`: it was added so that we use the same option across subcommands

src/main/CommandLine.cpp
```diff
@@ -299,14 +306,30 @@ maybeEnableInMemoryLedgerMode(Config& config, bool inMemory,
...
+    if (persistPartially)
+    {
+        config.DATABASE = SecretValue{partialDBForInMemoryMode(config)};
+        config.MODE_STORES_HISTORY = true;
+        // Since this mode still stores historical data (needed to restore
+        // ledger state in certain scenarios), set maintenance to run
+        // aggressively so that we only store a few ledgers worth of data
+        config.AUTOMATIC_MAINTENANCE_PERIOD = std::chrono::seconds(30);
+        config.AUTOMATIC_MAINTENANCE_COUNT = 100;
+        config.DISABLE_XDR_FSYNC = false;
```
MonsieurNicolas, line 321: you should not touch `DISABLE_XDR_FSYNC` here

MonsieurNicolas (changes requested):
a few suggestions to make the code flow easier to understand

MonsieurNicolas:
@graydon may have better insight than me

src/herder/test/HerderTests.cpp
```diff
@@ -2552,7 +2552,7 @@ TEST_CASE("slot herder policy", "[herder]")
     Config cfg(getTestConfig());
 
     // start in sync
-    cfg.FORCE_SCP = true;
+    cfg.FORCE_SCP = false;
```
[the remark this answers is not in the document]
marta-lokhova: because it was wrong, look at how the node is configured - it was `NODE_IS_VALIDATOR=false`, but `FORCE_SCP=true`

src/herder/test/HerderTests.cpp
```diff
@@ -2565,6 +2565,7 @@ TEST_CASE("slot herder policy", "[herder]")
 
     VirtualClock clock;
     Application::pointer app = createTestApplication(clock, cfg);
+    app->start();
```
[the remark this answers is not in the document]
marta-lokhova: because it was wrong, we usually start the application in other tests (see the rest of this file).

src/main/ApplicationUtils.cpp
```diff
@@ -131,6 +155,63 @@ runWithConfig(Config cfg, optional<CatchupConfiguration> cc)
...
+        throw std::runtime_error("Could not load ledger from database");
+    }
+
+    // Ensure all buckets are present on disk before application begins
+    for (auto const& b : has.currentBuckets)
+    {
+        auto checkBucket = [&](std::string bucketStr) {
+            auto bucketHash = hexToBin256(bucketStr);
+            auto bucket = app.getBucketManager().getBucketByHash(bucketHash);
+            if (!bucket || (isZero(bucket->getHash()) && !isZero(bucketHash)))
```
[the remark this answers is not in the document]
marta-lokhova: Did you look at the implementation of `getBucketByHash`? It creates an empty bucket pointer if it can't find the bucket, so checking `!bucket` is not enough.

That being said, looking at the last known ledger loading code, we already check for missing buckets there, so I think I don't even need this check.

src/main/CommandLine.cpp
marta-lokhova: The call `new-db --in-memory` is confusing and contradicts your other comment about calling the database "an in-memory database". The flag is called "in-memory", however the command sets a persistent state, that the next commands also depend on. I'd like to be as clear to the users as possible.

src/main/ApplicationUtils.cpp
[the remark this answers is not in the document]
MonsieurNicolas: > Did you look at the implementation of `getBucketByHash`? It creates an empty bucket pointer if it can't find the bucket, so checking !bucket is not enough.

yes I did, when it doesn't find the file, it does `return std::shared_ptr<Bucket>();` which is `nullptr` not an empty bucket.

other places in the code are only checking for `nullptr`

But yeah maybe we don't care anyways

src/main/ApplicationImpl.cpp
```diff
@@ -134,13 +134,24 @@ ApplicationImpl::ApplicationImpl(VirtualClock& clock, Config const& cfg)
 void
 ApplicationImpl::initialize(bool createNewDB)
 {
+    // Subtle: initialize the bucket manager first before initializing the
+    // database. This is needed as some modes in core (such as in-memory) use a
+    // small database inside the bucket directory.
+    mBucketManager = BucketManager::create(*this);
```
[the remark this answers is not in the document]
marta-lokhova: we don't delete the buckets folder in this case, but I agree it's safer to ensure the order of destruction is consistent with initialization order. Done.

src/main/ApplicationUtils.cpp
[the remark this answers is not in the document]
marta-lokhova: my bad, for some reason I was convinced that function uses `std::make_shared`. Sorry about that. But yeah, I removed this code now.

src/main/CommandLine.cpp
marta-lokhova: Picked up the latest master, which already addresses this problem but not forcing `DISABLE_XDR_FSYNC`, so I believe this is resolved now.

src/main/CommandLine.cpp
```diff
@@ -957,11 +970,26 @@ int
...
+    bool resetInMemoryState = false;
 
-    return runWithHelp(args, {configurationParser(configOption)}, [&] {
-        initializeDatabase(configOption.getConfig());
-        return 0;
-    });
+    auto resetParser = [](bool& resetInMemoryState) {
+        return clara::Opt{resetInMemoryState}["--reset-for-in-memory-mode"](
+            "Reset the special database used only for in-memory mode (see "
+            "--replay-in-memory flag");
```
graydon, line 978: the `--replay-in-memory` option is itself marked as deprecated; should _reference_ `--in-memory` here.

src/main/ApplicationUtils.cpp
```diff
@@ -30,75 +32,125 @@
...
+    // ledger and LCL is within this window.
+    uint32_t const RESTORE_STATE_LEDGER_WINDOW = 10;
+    // Do not rebuild genesis state
+    bool isGenesis = lcl == LedgerManager::GENESIS_LEDGER_SEQ;
+    return !isGenesis && startAtLedger >= lcl &&
+           startAtLedger - lcl <= RESTORE_STATE_LEDGER_WINDOW;
+}
+
+void
+setupMinimalDB(Config const& cfg, uint32_t startAtLedger)
```
graydon, line 49: I like the term "minimal DB" you're using here -- I would repeat / reuse / surface this term to the user. It is clear and can be extended into the phrase "minimal DB for in-memory mode".

src/main/CommandLine.cpp
```diff
@@ -282,11 +282,17 @@ maybeSetMetadataOutputStream(Config& cfg, std::string const& stream)
     }
 }
 
-std::optional<CatchupConfiguration>
-maybeEnableInMemoryLedgerMode(Config& config, bool inMemory,
-                              uint32_t startAtLedger,
-                              std::string const& startAtHash)
+std::string
+partialDBForInMemoryMode(Config const& cfg)
```
graydon, line 286: rename "partial" here to "minimal" for consistency.

src/main/CommandLine.cpp
```diff
@@ -282,11 +282,17 @@ maybeSetMetadataOutputStream(Config& cfg, std::string const& stream)
...
+{
+    return fmt::format("sqlite3://{}/partial.db", cfg.BUCKET_DIR_PATH);
```
graydon, line 288: and `partial.db` here to `minimal.db`

src/main/CommandLine.cpp
graydon: I suggest reusing the term you're using above: minimal. I.e. `stellar-core new-db --minimal-for-in-memory-mode`

graydon: (I was especially concerned by the word "reset" here because this suggests a non-minimal DB can be changed / reset into a minimal one, and I don't _think_ that's possible?)

src/main/CommandLine.cpp
```diff
@@ -957,11 +970,26 @@ int
 runNewDB(CommandLineArgs const& args)
 {
     CommandLine::ConfigOption configOption;
+    bool resetInMemoryState = false;
```
graydon, line 973: rename variable: `minimalForInMemoryMode`

src/main/ApplicationUtils.cpp
```diff
@@ -130,6 +182,35 @@ runWithConfig(Config cfg, std::optional<CatchupConfiguration> cc)
     return 0;
 }
 
+bool
+rebuildLatestState(Application& app)
```
graydon, line 186: rename function for specificity: `applyBucketsOfLastClosedLedger`

src/main/ApplicationUtils.cpp
```diff
@@ -130,6 +182,35 @@ runWithConfig(Config cfg, std::optional<CatchupConfiguration> cc)
...
+{
+    auto has = app.getLedgerManager().getLastClosedLedgerHAS();
+    auto lclHash =
+        app.getPersistentState().getState(PersistentState::kLastClosedLedger);
+
```
graydon, line 191: Add a `releaseAssertOrThrow` here that the ledger tables are all empty (using `LedgerTxn::countObjects`). We never want this code to run on a nonempty ledger.

src/main/ApplicationImpl.cpp
```diff
@@ -392,6 +392,24 @@ ApplicationImpl::~ApplicationImpl()
     LOG_INFO(DEFAULT_LOG, "Application destroyed");
 }
 
+void
+ApplicationImpl::resetDBForInMemoryMode()
```
graydon, line 396: This worries me / reminds me a bit of a "schema migration", or just .. an alternative path for creating a DB (albeit a minimal one) that does not _necessarily_ arrive at the same place we'd arrive if we reinitialized. I gather you're trying to preserve the overlay state here (but nothing else); would it be possible to do that explicitly? That is:

  1. load all the overlay state you want to preserve into a temporary data structure in memory.
  2. reinitialize the database using the normal code path.
  3. reinsert just the overlay state you wanted to preserve (data which we're a bit lax about anyways)

src/main/ApplicationUtils.cpp
```diff
@@ -30,75 +32,125 @@
 namespace stellar
 {
 
-int
-runWithConfig(Config cfg, std::optional<CatchupConfiguration> cc)
+bool
+canRebuildFromBuckets(uint32_t startAtLedger, uint32_t lcl)
```
graydon, line 36: rename for specificity: `canRestartInMemoryLedgerFromOldBuckets`

src/main/ApplicationUtils.cpp
graydon: (Maybe also extend the name of this function so any reader immediately knows it's about in-memory mode: `setupMinimalDBForInMemoryMode`

src/main/ApplicationUtils.cpp
```diff
@@ -30,75 +32,125 @@
...
+    // Do not rebuild genesis state
+    bool isGenesis = lcl == LedgerManager::GENESIS_LEDGER_SEQ;
+    return !isGenesis && startAtLedger >= lcl &&
+           startAtLedger - lcl <= RESTORE_STATE_LEDGER_WINDOW;
+}
+
+void
+setupMinimalDB(Config const& cfg, uint32_t startAtLedger)
+{
+    VirtualClock clock;
```
graydon, line 51: Add a `releaseAssert` that we're in in-memory mode.

src/main/ApplicationUtils.cpp
```diff
@@ -30,75 +32,125 @@
...
+            app->resetDBForInMemoryMode();
         }
     }
+}
+
+Application::pointer
+setupApp(Config& cfg, VirtualClock& clock, bool inMemory,
+         uint32_t startAtLedger, std::string const& startAtHash)
+{
+    if (inMemory)
```
graydon, line 96: separate flag here seems redundant -- isn't this just `cfg.isInMemoryMode()`?

src/bucket/BucketManagerImpl.cpp
```diff
@@ -107,7 +107,9 @@ BucketManagerImpl::BucketManagerImpl(Application& app)
     , mBucketSnapMerge(app.getMetrics().NewTimer({"bucket", "snap", "merge"}))
     , mSharedBucketsSize(
           app.getMetrics().NewCounter({"bucket", "memory", "shared"}))
-    , mDeleteEntireBucketDirInDtor(app.getConfig().isInMemoryMode())
+    , mDeleteEntireBucketDirInDtor(
+          app.getConfig().isInMemoryMode() &&
+          !app.getConfig().MODE_STORES_HISTORY_LEDGERHEADERS)
```
graydon, line 112: A comment would be nice here explaining why `!MODE_STORES_HISTORY_LEDGERHEADERS` relates to the deletion of entries in the bucketlist; or perhaps a helper method like `bool Config::isInMemoryModeWithoutMinimalDB() const { return isInMemoryMode() && !MODE_STORES_HISTORY_LEDGERHEADERS; }` or something.

src/main/Config.h
```diff
@@ -212,6 +212,9 @@ class Config : public std::enable_shared_from_this<Config>
     // fees, and scp history in the database
     bool MODE_STORES_HISTORY;
 
+    // A config parameter that stores ledger headers in the database
```
graydon, line 215: Extend this comment to include the fact that this exists in order to support a sub-mode of in-memory mode, where `MODE_STORES_HISTORY` is false but `MODE_STORES_HISTORY_LEDGERHEADERS` remains true. That is: that there is no legal configuration where `MODE_STORES_HISTORY_LEDGERHEADERS` is false but `MODE_STORES_HISTORY` is true. Or in yet other words, that only 3 of the 4 states of these 2 flags are legal.

(Also option to discuss here: I know we've gone back and forth on orthogonal flags vs. enumerated modes, but maybe this is as good as any to ask whether to reconsider merging `MODE_USES_IN_MEMORY_LEDGER`, `MODE_STORES_HISTORY` and `MODE_STORES_HISTORY_LEDGERHEADERS` back into a single enum. I'm not sure which of the _8_ possible combinations is legal off the top of my head, which is not a great sign.)

graydon (changes requested):
Overall looks promising! Fairly subtle changes though -- I've marked places where I think things could maybe be made a little more obvious for our future selves when we revisit these treacherous paths (as we surely will). Mostly just cosmetic though, I think there's only one structural / logic change around the question of how to most-safely "reset" vs. "reinitialize + reinsert" a minimal DB that can't be reused.

src/main/Config.h
MonsieurNicolas: mmm, I don't know @graydon : we used to have those "uber flags" used in the code base and it was actually very hard to understand which parts of the code had to change when touching things.

What we may want to do in this commit is probably change `MODE_STORES_HISTORY` to something that doesn't overlap with `MODE_STORES_HISTORY_LEDGERHEADERS` (right now it's hard to tell how they relate to each other: are they mutually exclusive?!).

So if we rename `MODE_STORES_HISTORY` to something like `MODE_STORES_HISTORY_MISC` (and introduce `MODE_STORES_HISTORY_LEDGERHEADERS` before) we can make it clear that they complement each other.

We can add helper functions to help check if *any* historical data is enabled (that seems to be the common case), or maybe to enable/disable "persist all historical data" (to flip both at the same time).

src/bucket/BucketManagerImpl.cpp
MonsieurNicolas: +1 on adding a helper function for clarity

src/main/ApplicationImpl.cpp
```diff
@@ -439,10 +439,13 @@ ApplicationImpl::validateAndLogConfig()
 
     if (getHistoryArchiveManager().hasAnyWritableHistoryArchive())
     {
-        if (!mConfig.MODE_STORES_HISTORY)
+        if (!mConfig.MODE_STORES_HISTORY ||
```
MonsieurNicolas, line 460: see my other comment: what we're looking for here is to check that if publish is enabled we better store all historical data. So a helper function here would help

src/main/CommandHandler.cpp
```diff
@@ -75,7 +75,8 @@ CommandHandler::CommandHandler(Application& app) : mApp(app)
 
     mServer->add404(std::bind(&CommandHandler::fileNotFound, this, _1, _2));
 
-    if (mApp.getConfig().MODE_STORES_HISTORY)
+    if (mApp.getConfig().MODE_STORES_HISTORY &&
+        mApp.getConfig().MODE_STORES_HISTORY_LEDGERHEADERS)
```
MonsieurNicolas, line 79: this should be an `||` as `maintenance` related endpoints should work regardless of which historical data is in there

src/main/CommandLine.cpp
```diff
@@ -311,17 +313,27 @@ maybeEnableInMemoryLedgerMode(Config& config, bool inMemory,
...
+    // Set database to a small sqlite database used to store minimal data needed
+    // to restore the ledger state
+    if (persistMinimalData)
+    {
+        config.DATABASE = SecretValue{partialDBForInMemoryMode(config)};
+        config.MODE_STORES_HISTORY_LEDGERHEADERS = true;
+        // Since this mode stores historical data (needed to restore
+        // ledger state in certain scenarios), set maintenance to run
+        // aggressively so that we only store a few ledgers worth of data
+        config.AUTOMATIC_MAINTENANCE_PERIOD = std::chrono::seconds(30);
```
MonsieurNicolas, line 334: any reason to only use `100` here? I think it's possible to replay more than 100 ledgers if captive core is started with a large catchup range (we use `1000000` in the catchup code a little later in the code, maybe we should use the same constant?)

MonsieurNicolas (changes requested):

src/main/ApplicationUtils.cpp
marta-lokhova: this sounds like a useful check, but the problem is that calling `countObjects` is not allowed on non-root LedgerTxn, which is what we use in in-memory mode (that is, the never committing LedgerTxn). There doesn't seem to be any other machinery to verify LedgerTxn's state (we could add some, but I don't think we should at this time). I will add a comment about this though.

src/main/Config.h
marta-lokhova: I like the idea to rename `MODE_STORES_HISTORY`. I ended up implementing `modeStoresAllHistory` and `modeStoresAnyHistory`, and that seems to improve clarity quite a bit.

marta-lokhova:
@graydon @MonsieurNicolas thanks for the feedback! I believe I addressed your comments in several new commits (I haven't squashed the changes yet, so that it's easy to review).

src/main/ApplicationUtils.cpp
MonsieurNicolas: I agree with Marta, I think it's fine as is for now

src/main/ApplicationUtils.cpp
```diff
@@ -130,6 +195,35 @@ runWithConfig(Config cfg, std::optional<CatchupConfiguration> cc)
...
+bool
+applyBucketsForLCL(Application& app)
+{
+    auto has = app.getLedgerManager().getLastClosedLedgerHAS();
+    auto lclHash =
+        app.getPersistentState().getState(PersistentState::kLastClosedLedger);
+
+    // As the local HAS might have merges in progress, let
+    // `prepareForPublish` convert it into a "valid historical HAS".
+    has.prepareForPublish(app);
```
MonsieurNicolas, line 207: Another "carry over".

Is this `has.prepareForPublish(app);` really needed?
The current ledger state should never depend on merges in progress (when we close a ledger we have already computed the current ledger) - I think the only thing that this call is going to do is potentially cause a stall when restarting because it will perform merges that it doesn't need to do.

src/main/CommandLine.cpp
```diff
@@ -1068,26 +1100,82 @@ run(CommandLineArgs const& args)
...
+                    }
+                    if (cfg.RUN_STANDALONE)
+                    {
+                        clockMode = VirtualClock::VIRTUAL_TIME;
+                        if (cfg.AUTOMATIC_MAINTENANCE_COUNT != 0 ||
+                            cfg.AUTOMATIC_MAINTENANCE_PERIOD.count() != 0)
+                        {
+                            LOG_WARNING(DEFAULT_LOG,
+                                        "Using MANUAL_CLOSE and RUN_STANDALONE "
+                                        "together "
```
MonsieurNicolas, line 1147: this string got chopped-formatted into too many bits. Maybe you need to merge it back into a single line and re-run clang-format?

src/main/Config.h
MonsieurNicolas: looking good!

src/overlay/PeerManager.cpp
```diff
@@ -570,6 +570,60 @@ PeerManager::dropAll(Database& db)
     db.getSession() << kSQLCreateStatement;
 }
 
+std::vector<std::pair<PeerBareAddress, PeerRecord>>
+PeerManager::loadAllPeers()
+{
+    ZoneScoped;
+    std::vector<std::pair<PeerBareAddress, PeerRecord>> result;
+    std::string sql = "SELECT * FROM peers";
```
MonsieurNicolas, line 578: avoid using `SELECT *` as it's a footgun, just list the column names that you expect

MonsieurNicolas (changes requested):
Looks very close to something we can merge. Good job. I added a couple suggestions/questions related to potentially bad things that we're carrying over from master

src/overlay/PeerManager.cpp
marta-lokhova: good call, updated.

src/main/ApplicationUtils.cpp
marta-lokhova: This is currently needed since ApplyBucketsWork is designed to apply buckets from an untrusted source, so it verifies validity of the HAS. That being said, I think the check could be moved out of ApplyBucketsWork into catchup. I opened https://github.com/stellar/stellar-core/issues/3044 to track this.

src/main/CommandLine.cpp
marta-lokhova: Ah yes, you're right, looks much better now.

marta-lokhova:
I believe this is ready now @MonsieurNicolas, I've squashed the commits as well.

MonsieurNicolas:
r+ 7cde80908478dd16ee5d6358b9a31b77feb38bbe

MonsieurNicolas:
@latobarita: retry

MonsieurNicolas:
@latobarita: retry

PR #3001, opened 2021-04-07
normalize meta when using printXDR subcommand

MonsieurNicolas:
# Description

This PR fixes a long annoyance when printing `TransactionMeta`:
it now normalizes changes, so that it's possible to diff meta captured during debugging sessions.

I added an optional `raw` mode so that we can still inspect data "as-is".


graydon:
r+ b36676d

PR #3008, opened 2021-04-08
doc: fixed both titles to be properly capitalised and visible

gammaled:
# Description

This PR addresses the titles in terms of proper capitalisation and uniform format. It looks nice and tidy now!

# Checklist
- [ ] Compiles
- [ ] Ran all tests

### DIFF
--- a/docs/architecture.md
+++ b/docs/architecture.md
@@ -1,6 +1,6 @@
----
-title: Process-level architecture
----
+
+Process-Level Architecture
+==========================
 
 Application owns a ledger-forming component, a p2p "overlay" component for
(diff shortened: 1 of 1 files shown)


MonsieurNicolas:
Thanks. The "title" section is actually used when pages are transformed into static html files, this PR would break that.
Also, the capitalization changes needed are much broader than this (so the PR should fix all of it to make sense). I'm not sure it's a good use of your time though, but if feel strongly about this feel free to update this PR with all titles updated.

MonsieurNicolas:
looks like this change was abandoned

Issue #2976, opened 2021-03-23
Catchup should sanitize archive names used in filenames


marta-lokhova:
Discovered while playing with the `--in-memory` mode. My config is taken directly from Stellarbeat:
```

[[VALIDATORS]]
NAME = "SDF 1"
PUBLIC_KEY = "GCGB2S2KGYARPVIA37HYZXVRM2YZUEXA6S33ZU5BUDC6THSB62LZSTYH"
ADDRESS = "3.90.255.55:11625"
HISTORY = "curl -sf http://history.stellar.org/prd/core-live/core_live_001/ -o {1}"
HOME_DOMAIN = "www.stellar.org"

[[VALIDATORS]]
NAME = "SDF 3"
PUBLIC_KEY = "GABMKJM6I25XI4K7U6XWMULOUQIQ27BCTMLS6BYYSOWKTBUXVRJSXHYQ"
ADDRESS = "3.238.240.236:11625"
HISTORY = "curl -sf http://history.stellar.org/prd/core-live/core_live_003/ -o {1}"
HOME_DOMAIN = "www.stellar.org"

[[VALIDATORS]]
NAME = "SDF 2"
PUBLIC_KEY = "GCM6QMP3DLRPTAZW2UZPCPX2LF3SXWXKPMP3GKFZBDSF3QZGV2G5QSTK"
ADDRESS = "52.207.236.180:11625"
HISTORY = "curl -sf http://history.stellar.org/prd/core-live/core_live_002/ -o {1}"
HOME_DOMAIN = "www.stellar.org"
```
The problem is that catchup uses archive names to construct filenames for downloaded data. If whitespaces are present, then catchup fails to download the files correctly:
```
2021-03-22T18:15:19.895 GD7GO [Process WARNING] process 92530 exited 6: curl -sf http://history.stellar.org/prd/core-live/core_live_003/history/02/0e/a8/history-020ea87f.json -o buckets/tmp/history-8b99e716d6d3d000/SDF 3-stellar-history.json
```

Removing any whitespaces from the archive names fixes the issue, so we should make sure that in the code archive names are properly sanitized before they are used in local filenames.

MonsieurNicolas:
[...]

PR #3012, opened 2021-04-12
Use random hashes for temp filenames

marta-lokhova:
Resolves https://github.com/stellar/stellar-core/issues/2976

This change improves the naming of local temp files during catchup and publish. Specifically, it avoids dependencies on node's config. These temp files are contained within the works that use it, and are not exposed outside. On retry, we re-use the same filename (which is what happens in master as well, although this is not critical for correctness).


graydon (approved):
LGTM, thanks!

graydon:
r+ e1177e6

PR #3013, opened 2021-04-13
C17 changes try again

graydon:
This attempts to merge the C++17 changes again, this time with a somewhat horribe (but I think safe) workaround for the integration test that failed previously.


src/herder/test/UpgradesTests.cpp
```diff
@@ -2319,3 +2319,33 @@ TEST_CASE("validate upgrade expiration logic", "[upgrades]")
         REQUIRE(upgrades.mBaseReserve);
     }
 }
+
+TEST_CASE("upgrade from cpp14 serialized data", "[upgrades]")
+{
+    std::string in = "{\n"
```
MonsieurNicolas, line 2325: please use a raw string litteral instead of escaping like this

MonsieurNicolas (changes requested):
minor suggestion to make it easier to maintain

src/herder/test/UpgradesTests.cpp
graydon: fixed

MonsieurNicolas:
r+ fcf6fe7df770ae402a910a62ff5ed05f617f0c88

src/herder/test/UpgradesTests.cpp
joko300370: joko300370 move and run

PR #3015, opened 2021-04-13
Log root account address when starting new ledger

leighmcculloch:
# Description
## What
Log the root account's public key in strkey format when starting a new
ledger.

## Why
When debugging and testing locally on standalone ledgers the developer needs
to know the root account seed to prefund test accounts. The seed is
conveniently outputted when running the `new-db` command when the new ledger
is being created, but the account public key / address is not. This is
inconvenient because the developer needs both. The developer can derive the
public key from the secret key using tools such as stc but it would be more
convenient if the address was logged alongside the seed.

# Checklist
- [x] Compiles
- [ ] Ran all tests


MonsieurNicolas:
good idea. Thanks!

MonsieurNicolas:
r+ f6332d479c1ed9295caa051331d9ace3c3156218

PR #3016, opened 2021-04-14
Use nullopt

graydon:
# Description

Mainly this change shifts us from using `std::optional()` to `std::nullopt`. I introduced a bunch of the former in #3013 and having read a bit more internet lore I now think it's better to do the latter.

I also included some warning fixups, one of which might have been serious but I asked @sisuresh and it's not -- tests already handle the missing enums earlier.

No functional change.

# Checklist
- [x] Compiles
- [x] Ran all tests


MonsieurNicolas:
r+ 68adb4f67609af6384439d43d6140054b8eef327

PR #3027, opened 2021-04-20
fix issues found with static analysis

MonsieurNicolas:
only real bug here is the use of the non initialized config setting `LOG_COLOR`


graydon:
r+ 519860c

MonsieurNicolas:
fixed a typo, rekicking

MonsieurNicolas:
r+ 4c273e31469373f07412061f775a9750a8429cd8

PR #3031, opened 2021-04-22
Rebuild ledger tables on schema upgrade

jonjove:
# Description
This PR changes the way that schema upgrades are performed. The existing approach directly modifies the ledger tables. The new approach is to reconstruct the ledger tables entirely by applying buckets. This is slower, but significantly reduces the risks associated with a schema upgrade.

# Checklist
- [x] Compiles
- [x] Ran all tests


src/database/Database.cpp
```diff
@@ -198,257 +199,76 @@ Database::Database(Application& app)
...
+        soci::transaction tx(mSession);
+        mApp.getLedgerTxnRoot().dropAccounts();
+        mApp.getLedgerTxnRoot().dropOffers();
+        mApp.getLedgerTxnRoot().dropTrustLines();
+        mApp.getLedgerTxnRoot().dropData();
+        mApp.getLedgerTxnRoot().dropClaimableBalances();
+        tx.commit();
+    }
+
+    // No transaction is needed. ApplyBucketsWork breaks the apply into many
```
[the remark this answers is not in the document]
jonjove: Done.

src/database/Database.cpp
```diff
@@ -198,257 +199,76 @@ Database::Database(Application& app)
...
-    case 13:
-        if (!mApp.getConfig().MODE_USES_IN_MEMORY_LEDGER)
+        HistoryArchiveState has;
+        has.fromString(hasStr);
+        has.prepareForPublish(mApp);
+
+        std::map<std::string, std::shared_ptr<Bucket>> localBuckets;
+        auto& ws = mApp.getWorkScheduler();
+        auto applyBucketsWork = ws.executeWork<ApplyBucketsWork>(
+            localBuckets, has, Config::CURRENT_LEDGER_PROTOCOL_VERSION);
```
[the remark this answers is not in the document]
marta-lokhova: The scenario I was wondering about is when the state got corrupted for whatever reason in between restarts. For buckets, we would have to wait for the application to complete before failing, for example. I don't feel super strongly about this though, as I think the chances of these specific corruptions are low, and eventually we do fail, so if you prefer to leave it as is, that's fine with me.

MonsieurNicolas (changes requested):
Looks pretty good overall.

I would make the following tweaks:
* I think we should keep the schema upgrade logic (just no-ops right now) so that if we want to make changes to tables that are not ledger related in the future, we don't have to re-introduce the `vers` parameter (`Database::applySchemaUpgrade(unsigned long vers)`) and the tracking of `MIN_SCHEMA_VERSION` to perform incremental schema upgrades.
* I think that we should organize the code to use a new persistent state "rebuild ledger state" that when set drops ledger related tables and rebuilds (acts as a "SQL tables are dirty" flag).
    * That way we can set that flag when we need to (during schema upgrades, like 13->14 would just set that flag) but also add a new command (can be command line or http command) that sets that flag so that on the next startup we rebuild from buckets, that way we can test it easily.
    * You would only clear that flag after rebuilding state from buckets, so it makes the whole thing safe even when not using transactions (for performance reason).
    * In the future we can even break up that flag into a per table flag, so that we can filter ledger entries when we apply buckets. We could even do this right away so that the 13->14 schema upgrade only rebuilds the `Offers` table (that said, I am not sure we need to rebuild right now?).

That last point would mitigate a concern I have with this change that we'd be taking a big performance hit with every schema upgrade: we've had people complain already that they can't take the downtime with captive core, full rebuilds backed by SQL are a lot slower (even with the SKUs that we use apply buckets takes close to 20 minutes right now).

src/bucket/BucketApplicator.cpp
```diff
@@ -17,9 +17,15 @@ namespace stellar
...
-                                   std::shared_ptr<const Bucket> bucket)
-    : mApp(app), mMaxProtocolVersion(maxProtocolVersion), mBucketIter(bucket)
+                                   std::shared_ptr<const Bucket> bucket,
+                                   std::vector<uint8_t> const& filter)
+    : mApp(app)
+    , mMaxProtocolVersion(maxProtocolVersion)
+    , mBucketIter(bucket)
+    , mEntryTypeFilter(filter)
 {
+    assert(mEntryTypeFilter.size() ==
```
graydon, line 27: releaseAssert

src/bucket/Bucket.cpp
```diff
@@ -82,19 +82,24 @@ Bucket::containsBucketIdentity(BucketEntry const& id) const
     return false;
 }
 
+#ifdef BUILD_TESTS
 void
 Bucket::apply(Application& app) const
 {
     ZoneScoped;
+
+    std::vector<uint8_t> filter(
```
graydon, line 91: Why uint8_t instead of bool?

src/main/ApplicationImpl.cpp
```diff
@@ -133,6 +135,80 @@ ApplicationImpl::ApplicationImpl(VirtualClock& clock, Config const& cfg)
...
+                abort();
+            }
+        }
+
+        tx.commit();
+    }
+
+    // No transaction is needed. ApplyBucketsWork breaks the apply into many
+    // small chunks, each of which has its own transaction. If it fails at some
+    // point in the middle, then rebuildledger will not be cleared so this will.
```
graydon, line 195: Incomplete sentence? "... so this will." I think this should read "... so this will re-run on next startup."

src/ledger/LedgerTxnAccountSQL.cpp
```diff
@@ -481,7 +481,9 @@ LedgerTxnRoot::Impl::dropAccounts()
            "thresholds         TEXT         NOT NULL,"
            "flags              INT          NOT NULL,"
            "signers            TEXT,"
-           "lastmodified       INT          NOT NULL"
+           "lastmodified       INT          NOT NULL,"
+           "extension          TEXT,"
```
graydon, line 485: I'm unclear on why `extension` remains nullable here when `ledgerext` is not. They both get routed into `decodeOpaqueXDR` (subject to a SOCI indicator variable). Probably I am missing some obvious reason here but could you spell it out? Why are the two columns treated differently?

graydon:
For unclear reasons github has removed the review-in-progress and the review button, perhaps I've been de-authorized or something .. anyway I am mostly ok with this, just a few nits above plus a question concerning residual nullability.

graydon (changes requested):
mostly ok with this, just a few nits above plus a question concerning residual nullability

src/bucket/BucketApplicator.cpp
```diff
@@ -58,23 +80,22 @@ BucketApplicator::advance(BucketApplicator::Counters& counters)
         BucketEntry const& e = *mBucketIter;
         Bucket::checkProtocolLegality(e, mMaxProtocolVersion);
         counters.mark(e);
-        if (e.type() == LIVEENTRY || e.type() == INITENTRY)
-        {
-            ltx.createOrUpdateWithoutLoading(e.liveEntry());
-        }
-        else
+
+        if (shouldApplyEntry(mEntryTypeFilter, e))
```
MonsieurNicolas, line 84: do you need to move the `counters.mark()` inside this if ?

src/ledger/LedgerTxnAccountSQL.cpp
MonsieurNicolas: This is because this PR has to preserve the table schemas from schema v13 (the problem here is that it's actually not easy to see that this change is just the squash from one off 12->13 upgrades into the main "create table" flow).
We can only change schema in this PR for the offer table (as it's the only one getting upgraded as part of schema 14)

src/ledger/LedgerTxnOfferSQL.cpp
```diff
@@ -671,6 +671,8 @@ LedgerTxnRoot::Impl::dropOffers()
            "price            DOUBLE PRECISION NOT NULL,"
            "flags            INT              NOT NULL,"
            "lastmodified     INT              NOT NULL,"
+           "extension        TEXT,"
```
MonsieurNicolas, line 674: should we fix this table? as this is the one table we're going to rebuild for sure, we can do anything we want wrt schema and indexes
at least make this `NOT NULL` as we always set it to something.

src/database/Database.cpp
```diff
@@ -208,32 +208,17 @@ Database::applySchemaUpgrade(unsigned long vers)
...
-            // base64-encoded empty v0 XDR extensions, so that, as with the
-            // other LedgerEntry extensions, we'll be able to add such
-            // extensions in the future without bumping the database schema
-            // version, writing any upgrade code, or changing the SQL that reads
-            // and writes those tables.
-            addTextColumn("offers", "extension");
-            addTextColumn("accountdata", "extension");
-
-            mApp.getLedgerTxnRoot().dropClaimableBalances();
+            std::vector<LedgerEntryType> let{ACCOUNT, TRUSTLINE, OFFER, DATA,
```
MonsieurNicolas, line 211: we can actually delete this code: schema version 13 has been around for almost a year, so there is no need to support 12->13 (so we can bump `MIN_SCHEMA_VERSION` to 13)

src/main/ApplicationImpl.cpp
```diff
@@ -236,6 +312,7 @@ ApplicationImpl::newDB()
 {
     mDatabase->initialize();
     mDatabase->upgradeToCurrentSchema();
+    maybeRebuildLedger(*this, false);
```
MonsieurNicolas, line 315: grep for `upgradeToCurrentSchema`: there are two other call sites.
I *think* we should factor this into its own method so that we never forget to call `maybeRebuildLedger` after upgrading the schema.

MonsieurNicolas (changes requested):
I like how this turned out!

src/bucket/BucketApplicator.cpp
jonjove: Completely gone now.

src/bucket/Bucket.cpp
jonjove: I agree that predicates is much cleaner but probably not more efficient (I did the `vector` because of the very efficient look-up). But the reality is the interface is so much cleaner that it's worth it, and we could always wrap the vector approach in a function if profiling indicates that would be valuable. Done.

src/database/Database.cpp
jonjove: Good idea, done.

src/main/ApplicationImpl.cpp
jonjove: Fixed, thanks.

src/ledger/LedgerTxnOfferSQL.cpp
jonjove: Fixed.

src/main/ApplicationImpl.cpp
jonjove: My implementation used to look like this, but something went wrong in the rebase (lot of changes to this code lately). Good catch, fixed by:
- creating a method that bundles schema upgrade and ledger rebuild
- eliminating the third call site, rerouting that to the other code paths

MonsieurNicolas (approved):
looking good!

src/main/ApplicationImpl.cpp
```diff
@@ -133,8 +135,86 @@ ApplicationImpl::ApplicationImpl(VirtualClock& clock, Config const& cfg)
     }
 }
 
+static void
+maybeRebuildLedger(Application& app, bool applyBuckets)
+{
+    std::set<LedgerEntryType> toRebuild;
+    auto& ps = app.getPersistentState();
+    for (auto let : xdr::xdr_traits<LedgerEntryType>::enum_values())
```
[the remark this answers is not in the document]
jonjove: Good idea: https://github.com/xdrpp/xdrpp/issues/28

src/ledger/LedgerTxnAccountSQL.cpp
graydon: Ok so this is short term and we'll eventually squeeze out the nulls after enough single-table upgrades. good enough!

graydon (approved):

MonsieurNicolas:
r+ f3c5b4da73cfc840c830936b769e5f45a9d7f2bd

PR #3033, opened 2021-04-23
Enable in-memory LedgerTxn for in-memory mode

marta-lokhova:


src/transactions/TransactionFrame.cpp
```diff
@@ -795,6 +797,31 @@ TransactionFrame::applyOperations(SignatureChecker& signatureChecker,
                 return false;
             }
 
+            if (gIsProductionNetwork && ledgerSeq == 34793621)
```
jonjove, line 800: I like this approach, but I don't think this is in the right place. Specifically, this won't generate the correct meta when replaying this ledger. The meta should show that this sponsorship was added to the offer, in the specific transaction where that actually happened and no others. Probably the best thing to do here is push this into `convertWithOffers`, since at that point you can see if this particular offer was actually crossed.

src/main/Config.cpp
```diff
@@ -1558,11 +1558,6 @@ Config::getExpectedLedgerCloseTime() const
 void
 Config::setInMemoryMode()
 {
-    // For the time being, --in-memory mode on the command line does not enable
-    // MODE_USES_IN_MEMORY_LEDGER; rather it uses in-memory sqlite. This change
-    // happened when mitigating the protocol 16 issue, funneling all database
-    // access through the sqlite layer; it may be reverted once a different
-    // workaround for that issue is in place.
```
jonjove, line 1566: Nitpick: might considering squashing this into the previous commit. I added a comment saying this was stale, only to see it fixed in the next commit.

marta-lokhova: done.

src/transactions/TransactionFrame.cpp
marta-lokhova: fixed.

jonjove (approved):
The transactions and LedgerTxn part of this look good to me. I'm not an expert on the in-memory configuration (although it looks reasonable to me), so if someone else is happy with that part then feel free to merge this.

MonsieurNicolas:
r+ c0ee493d3e058def370c1160a32839a8f9bc602e

PR #3034, opened 2021-04-23
Update simulate-apply-per-op 

hidenori-shinohara:
# Description

Update simulate-apply-per-op which is a constant value applied to every operation in order to provide different durations and how often we want each value.

This PR deprecates the `simulate-apply-per-op` command line argument, and adds `OP_APPLY_SLEEP_TIME_DURATION_FOR_TESTING` and `OP_APPLY_SLEEP_TIME_WEIGHT_FOR_TESTING`. Each of them is a list such that their `i`-th elements specify how often (in percentage) and how long we want to sleep.

# Checklist
- [x] Compiles
- [x] Ran all tests

### DIFF
--- a/src/ledger/LedgerManagerImpl.cpp
+++ b/src/ledger/LedgerManagerImpl.cpp
@@ -740,17 +740,24 @@ LedgerManagerImpl::closeLedger(LedgerCloseData const& ledgerData)
     // step 4
     mApp.getBucketManager().forgetUnreferencedBuckets();
 
-    // Maybe sleep for parameterized amount of time in simulation mode
-    auto sleepFor = std::chrono::microseconds{
-        mApp.getConfig().OP_APPLY_SLEEP_TIME_FOR_TESTING * txSet->sizeOp()};
-    std::chrono::microseconds applicationTime =
-        closeLedgerTime.checkElapsedTime();
-    if (applicationTime < sleepFor)
+    if (!mApp.getConfig().getOpApplySleepTimeForTesting().empty())
     {
-        sleepFor -= applicationTime;
-        CLOG_DEBUG(Perf, "Simulate application: sleep for {} microseconds",
-                   sleepFor.count());
-        std::this_thread::sleep_for(sleepFor);
+        // Sleep for a parameterized amount of time in simulation mode
+        std::chrono::microseconds sleepFor{0};
+        for (int i = 0; i < txSet->sizeOp(); i++)
+        {
+            sleepFor +=
+                rand_element(mApp.getConfig().getOpApplySleepTimeForTesting());
+        }
+        std::chrono::microseconds applicationTime =
+            closeLedgerTime.checkElapsedTime();
+        if (applicationTime < sleepFor)
+        {
+            sleepFor -= applicationTime;
+            CLOG_DEBUG(Perf, "Simulate application: sleep for {} microseconds",
+                       sleepFor.count());
+            std::this_thread::sleep_for(sleepFor);
+        }
     }
 
     std::chrono::duration<double> ledgerTimeSeconds = ledgerTime.Stop();
--- a/src/main/CommandLine.cpp
+++ b/src/main/CommandLine.cpp
@@ -1078,25 +1078,18 @@ run(CommandLineArgs const& args)
 {
     CommandLine::ConfigOption configOption;
     auto disableBucketGC = false;
-    uint32_t simulateSleepPerOp = 0;
     std::string stream;
     bool inMemory = false;
     bool waitForConsensus = false;
     uint32_t startAtLedger = 0;
     std::string startAtHash;
 
-    auto simulateParser = [](uint32_t& simulateSleepPerOp) {
-        return clara::Opt{simulateSleepPerOp,
-                          "MICROSECONDS"}["--simulate-apply-per-op"](
-            "simulate application time per operation");
-    };
-
     return runWithHelp(
         args,
         {configurationParser(configOption),
          disableBucketGCParser(disableBucketGC),
-         simulateParser(simulateSleepPerOp), metadataOutputStreamParser(stream),
-         inMemoryParser(inMemory), waitForConsensusParser(waitForConsensus),
+         metadataOutputStreamParser(stream), inMemoryParser(inMemory),
+         waitForConsensusParser(waitForConsensus),
          startAtLedgerParser(startAtLedger), startAtHashParser(startAtHash)},
         [&] {
             Config cfg;
@@ -1109,10 +1102,11 @@ run(CommandLineArgs const& args)
                 // First, craft and validate the configuration
                 cfg = configOption.getConfig();
                 cfg.DISABLE_BUCKET_GC = disableBucketGC;
-                if (simulateSleepPerOp > 0)
+
+                if (!cfg.OP_APPLY_SLEEP_TIME_DURATION_FOR_TESTING.empty() ||
+                    !cfg.OP_APPLY_SLEEP_TIME_WEIGHT_FOR_TESTING.empty())
                 {
                     cfg.DATABASE = SecretValue{"sqlite3://:memory:"};
-                    cfg.OP_APPLY_SLEEP_TIME_FOR_TESTING = simulateSleepPerOp;
                     cfg.MODE_STORES_HISTORY_MISC = false;
                     cfg.MODE_USES_IN_MEMORY_LEDGER = false;
                     cfg.MODE_ENABLES_BUCKETLIST = false;
--- a/src/main/Config.cpp
+++ b/src/main/Config.cpp
@@ -43,7 +43,8 @@ static const std::unordered_set<std::string> TESTING_ONLY_OPTIONS = {
     "ARTIFICIALLY_ACCELERATE_TIME_FOR_TESTING",
     "ARTIFICIALLY_SET_CLOSE_TIME_FOR_TESTING",
     "ARTIFICIALLY_REPLAY_WITH_NEWEST_BUCKET_LOGIC_FOR_TESTING",
-    "OP_APPLY_SLEEP_TIME_FOR_TESTING"};
+    "OP_APPLY_SLEEP_TIME_DURATION_FOR_TESTING",
+    "OP_APPLY_SLEEP_TIME_WEIGHT_FOR_TESTING"};
 
 // Options that should only be used for testing
 static const std::unordered_set<std::string> TESTING_SUGGESTED_OPTIONS = {
@@ -102,7 +103,9 @@ Config::Config() : NODE_SEED(SecretKey::random())
     MODE_STORES_HISTORY_LEDGERHEADERS = true;
     MODE_DOES_CATCHUP = true;
     MODE_AUTO_STARTS_OVERLAY = true;
-    OP_APPLY_SLEEP_TIME_FOR_TESTING = 0;
+    OP_APPLY_SLEEP_TIME_DURATION_FOR_TESTING =
+        std::vector<std::chrono::microseconds>();
+    OP_APPLY_SLEEP_TIME_WEIGHT_FOR_TESTING = std::vector<unsigned short>();
 
     FORCE_SCP = false;
     LEDGER_PROTOCOL_VERSION = CURRENT_LEDGER_PROTOCOL_VERSION;
@@ -239,10 +242,11 @@ readString(ConfigItem const& item)
     return item.second->as<std::string>()->get();
 }
 
-std::vector<std::string>
-readStringArray(ConfigItem const& item)
+template <typename T>
+std::vector<T>
+readArray(ConfigItem const& item)
 {
-    auto result = std::vector<std::string>{};
+    auto result = std::vector<T>{};
     if (!item.second->is_array())
     {
         throw std::invalid_argument(
@@ -250,12 +254,12 @@ readStringArray(ConfigItem const& item)
     }
     for (auto v : item.second->as_array()->get())
     {
-        if (!v->as<std::string>())
+        if (!v->as<T>())
         {
             throw std::invalid_argument(
                 fmt::format("invalid element of '{}'", item.first));
         }
-        result.push_back(v->as<std::string>()->get());
+        result.push_back(v->as<T>()->get());
     }
     return result;
 }
@@ -350,7 +354,7 @@ Config::loadQset(std::shared_ptr<cpptoml::table> group, SCPQuorumSet& qset,
         }
         else if (item.first == "VALIDATORS")
         {
-            auto values = readStringArray(item);
+            auto values = readArray<std::string>(item);
             for (auto v : values)
             {
                 PublicKey nodeID;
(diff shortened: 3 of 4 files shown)


src/main/CommandLine.cpp
```diff
@@ -1041,30 +1041,109 @@ runReportLastHistoryCheckpoint(CommandLineArgs const& args)
...
     };
 
     return runWithHelp(
         args,
         {configurationParser(configOption),
          disableBucketGCParser(disableBucketGC),
-         simulateParser(simulateSleepPerOp), metadataOutputStreamParser(stream),
-         inMemoryParser(inMemory), waitForConsensusParser(waitForConsensus),
+         simulateParser(simulateApplyPerOp),
+         simulateParserGenerator("fast")(simulateApplyPerOpFast),
```
MonsieurNicolas, line 1142: there is no need to hardcode slow/medium/high:
`clara::Opt` can take a vector as a parameter, so in this case you would add a parser like
```
std::vector<std::string> perOpPairs;
clara::Opt{confs, "PERCENT/MICROSECOND"}["--simulate-apply-per-op-dist"]("percent followed by time in usecond (specified several times, sum of percent must be 100)")}
```

src/ledger/LedgerManagerImpl.cpp
```diff
@@ -739,17 +739,35 @@ LedgerManagerImpl::closeLedger(LedgerCloseData const& ledgerData)
...
-    if (applicationTime < sleepFor)
+    if (!mApp.getConfig().OP_APPLY_SLEEP_TIME_FOR_TESTING.empty())
     {
-        sleepFor -= applicationTime;
-        CLOG_DEBUG(Perf, "Simulate application: sleep for {} microseconds",
-                   sleepFor.count());
-        std::this_thread::sleep_for(sleepFor);
+        // Sleep for a parameterized amount of time in simulation mode
+        int sample = rand_uniform(1, 100);
+        uint32_t chosenDuration = 0;
```
MonsieurNicolas, line 746: extract in a separate function, also, why not just have `OP_APPLY_SLEEP_TIME_FOR_TESTING` be exactly 100 elements, that way there is no need for a loop here and just do a random pick

src/ledger/LedgerManagerImpl.cpp
```diff
@@ -739,17 +739,35 @@ LedgerManagerImpl::closeLedger(LedgerCloseData const& ledgerData)
...
+        {
+            sample -= p.first;
+            if (sample <= 0)
+            {
+                chosenDuration = p.second;
+                break;
+            }
+        }
+        auto sleepFor =
+            std::chrono::microseconds{chosenDuration * txSet->sizeOp()};
```
MonsieurNicolas, line 761: do we want this or should it pick a random duration for each op separately?

src/main/CommandLine.cpp
```diff
@@ -1041,30 +1041,109 @@ runReportLastHistoryCheckpoint(CommandLineArgs const& args)
...
+parseSimulateSleepPerOp(std::vector<std::string> const& inputs)
+{
+    // The input must be of the form percentage/duration
+    // where
+    //     * 0 <= percentage <= 100
+    //     * 0 <= duration
+    //     * percentage for slow + percentage for medium + percentage for fast =
+    //     100.
+    //     * duration for slow >= duration for medium >= duration for fast.
+    if (std::any_of(inputs.begin(), inputs.end(),
```
MonsieurNicolas, line 1054: I don't think this function should care about slow/medium/fast: it's just a vector of `percentage/duration` that needs to be transformed

src/main/CommandLine.cpp
```diff
@@ -1041,30 +1041,109 @@ runReportLastHistoryCheckpoint(CommandLineArgs const& args)
...
+                    [](std::string const& s) { return s.empty(); }))
+    {
+        throw std::runtime_error(
+            "All three of simulate-sleep-per-op-{slow, medium, fast} must be "
+            "defined if at least one is defined");
+    }
+    auto parser = [](std::string const& input, uint32_t& percentage,
+                     uint32_t& duration) {
+        auto separatorIndex = input.find("/");
```
MonsieurNicolas, line 1063: do not roll your own parser: you have several bugs here (doesn't handle trailing characters for example).

Best is to use a regex in this kind of situation and only feed proper data to the integer conversion functions.

src/main/CommandLine.cpp
```diff
@@ -1041,30 +1041,109 @@ runReportLastHistoryCheckpoint(CommandLineArgs const& args)
...
+    };
+
+    uint32_t percentage, duration, totalPercentage{0};
+    std::vector<std::pair<uint32_t, uint32_t>> ret;
+    for (auto const& input : inputs)
+    {
+        parser(input, percentage, duration);
+        if (!ret.empty() && ret.back().second < duration)
+        {
+            throw std::runtime_error(
```
MonsieurNicolas, line 1085: If we make this generalized, there should not be any requirement on order

src/main/CommandLine.cpp
```diff
@@ -1041,30 +1041,109 @@ runReportLastHistoryCheckpoint(CommandLineArgs const& args)
...
+        auto separatorIndex = input.find("/");
+        try
+        {
+            auto pos = std::size_t{0};
+            percentage = std::stoul(input.substr(0, separatorIndex), &pos);
+            duration = std::stoul(input.substr(separatorIndex + 1), &pos);
+        }
+        catch (std::exception&)
+        {
+            throw std::runtime_error(
```
MonsieurNicolas, line 1072: exception type is argument error not a runtime error

src/main/CommandLine.cpp
```diff
@@ -1041,30 +1041,109 @@ runReportLastHistoryCheckpoint(CommandLineArgs const& args)
         });
 }
 
+std::vector<std::pair<uint32_t, uint32_t>>
```
MonsieurNicolas, line 1044: we should be returning `std::chrono::microseconds`

MonsieurNicolas (changes requested):

src/main/CommandLine.cpp
hidenori-shinohara: Updated it. Thanks for this idea! This is much cleaner :)

src/ledger/LedgerManagerImpl.cpp
hidenori-shinohara: Put the logic in `chooseOpApplySleepTimeForTesting` and turned `OP_APPLY_SLEEP_TIME_FOR_TESTING` into a vector of 100 elements

src/ledger/LedgerManagerImpl.cpp
hidenori-shinohara: I updated it to pick a random duration for each op since I thought that it'll be closer to what the real network is likely doing.

But then after updating the code I remembered that it may not matter since loadgen currently includes only one Payment operation in a transaction. I'm not quite sure what would be the best approach here. Let me know what you think.

src/main/CommandLine.cpp
hidenori-shinohara: Removed the slow/medium/fast stuff.

src/main/CommandLine.cpp
hidenori-shinohara: Updated it with regex.

src/main/CommandLine.cpp
hidenori-shinohara: Removed this check.

src/main/CommandLine.cpp
hidenori-shinohara: Fixed.

src/main/CommandLine.cpp
hidenori-shinohara: Updated it.

src/main/Config.h
```diff
@@ -190,9 +194,10 @@ class Config : public std::enable_shared_from_this<Config>
     bool ARTIFICIALLY_REPLAY_WITH_NEWEST_BUCKET_LOGIC_FOR_TESTING;
 
     // A config parameter that forces transaction application during ledger
-    // close to sleep for a given number of microseconds. This option is only
-    // for consensus and overlay simulation testing.
-    uint32_t OP_APPLY_SLEEP_TIME_FOR_TESTING;
+    // close to sleep. This option is only for consensus and overlay simulation
+    // testing. This vector contains 100 elements each of which is a duration.
+    // Some durations repeat themselves to achieve a desired distribution.
+    std::vector<std::chrono::microseconds> OP_APPLY_SLEEP_TIME_FOR_TESTING;
```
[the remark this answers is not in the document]
marta-lokhova: correction: the member var should probably live in the Config class, as it's not really LedgerManager's responsibility to maintain this config.

src/ledger/LedgerManagerImpl.cpp
```diff
@@ -1,4 +1,5 @@
 // Copyright 2014 Stellar Development Foundation and contributors. Licensed
+// heheh
```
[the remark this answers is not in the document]
hidenori-shinohara: Yes lol

src/main/Config.cpp
```diff
@@ -1074,6 +1134,14 @@ Config::processConfig(std::shared_ptr<cpptoml::table> t)
                 EXCLUDE_TRANSACTIONS_CONTAINING_OPERATION_TYPE =
                     readXdrEnumArray<OperationType>(item);
             }
+            else if (item.first == "OP_APPLY_SLEEP_TIME_DURATION")
```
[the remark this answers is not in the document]
hidenori-shinohara: Done!

src/main/Config.cpp
```diff
@@ -1082,6 +1150,17 @@ Config::processConfig(std::shared_ptr<cpptoml::table> t)
                 throw std::invalid_argument(err);
             }
         }
+        if (!opApplySleepTimePercent.empty() ||
```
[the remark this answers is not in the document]
hidenori-shinohara: Done!

src/main/Config.h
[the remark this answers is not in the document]
hidenori-shinohara: After discussing this offline & some more thinking, I updated this such that:

- `processOpApplySleepTimeForTestingConfigs` is a public method that takes the weight and duration lists and produces a vector of 100 elements. (I made it public so I can run it from `CommandLine::run`)
- `opApplySleepTimeForTesting` is a public member that contains 100 elements. (I made it public and removed the `choose` function because `rand_element` makes it simple and also it can be used to check if we need to simulate op apply.)

Let me know what you think!

src/main/Config.cpp
```diff
@@ -1689,4 +1768,23 @@ Config::toString(SCPQuorumSet const& qset)
 }
 
 std::string const Config::STDIN_SPECIAL_NAME = "/dev/stdin";
+
+std::chrono::microseconds
+Config::chooseOpApplySleepTimeForTesting() const
+{
+    if (OP_APPLY_SLEEP_TIME_FOR_TESTING.size() != 100)
```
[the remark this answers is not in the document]
hidenori-shinohara: Removed!

src/main/Config.cpp
```diff
@@ -1689,4 +1768,23 @@ Config::toString(SCPQuorumSet const& qset)
...
+    {
+        throw std::runtime_error(fmt::format(
+            "chooseOpApplySleepTimeForTesting should be called only if "
+            "OP_APPLY_SLEEP_TIME_FOR_TESTING contains exactly 100 elements "
+            "each of which represents one percent, but it contains {} elements",
+            OP_APPLY_SLEEP_TIME_FOR_TESTING.size()));
+    }
+    else
+    {
+        int sample = rand_uniform(0, 99);
```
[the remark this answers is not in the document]
hidenori-shinohara: Thanks for the idea!

src/main/Config.cpp
```diff
@@ -22,6 +22,7 @@
 #include <fmt/chrono.h>
 #include <fmt/format.h>
 #include <functional>
+#include <regex>
```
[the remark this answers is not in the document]
hidenori-shinohara: Yup, removed.

src/main/Config.cpp
```diff
@@ -260,6 +261,28 @@ readStringArray(ConfigItem const& item)
     return result;
 }
 
+template <typename T>
+std::vector<T>
+readIntArray(ConfigItem const& item)
```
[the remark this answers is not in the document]
hidenori-shinohara: I created `readArray`

src/main/Config.cpp
```diff
@@ -707,6 +730,41 @@ Config::verifyHistoryValidatorsBlocking(
     }
 }
 
+std::vector<std::chrono::microseconds>
+Config::parseOpApplySleepTimeForTesting(std::vector<int> const& percent,
```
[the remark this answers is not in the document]
hidenori-shinohara: Renamed.

src/main/Config.cpp
```diff
@@ -707,6 +730,41 @@ Config::verifyHistoryValidatorsBlocking(
...
+            "OP_APPLY_SLEEP_TIME_DURATION and OP_APPLY_SLEEP_TIME_PERCENT must "
+            "be defined together");
+    }
+    if (percent.size() != duration.size())
+    {
+        throw std::invalid_argument(
+            "OP_APPLY_SLEEP_TIME_DURATION and OP_APPLY_SLEEP_TIME_PERCENT must "
+            "have the same size");
+    }
+    if (std::accumulate(percent.begin(), percent.end(), 0) != 100)
```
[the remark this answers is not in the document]
hidenori-shinohara: Thanks for the idea!

src/main/Config.cpp
```diff
@@ -707,6 +730,41 @@ Config::verifyHistoryValidatorsBlocking(
...
+    {
+        throw std::invalid_argument(
+            "The sum of the percentages in "
+            "OP_APPLY_SLEEP_TIME_PERCENT must equal 100");
+    }
+    std::vector<std::chrono::microseconds> ret(100);
```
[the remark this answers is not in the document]
hidenori-shinohara: Yes, you're right. I always confuse them because `ArrayList` in Java uses it as the capacity, and `vector` in C++ uses it as the size. 😕  Updated it!

marta-lokhova (commented):
getting close! A few suggestions

src/main/Config.h
```diff
@@ -430,5 +438,12 @@ class Config : public std::enable_shared_from_this<Config>
     // A special name to be used for stdin in stead of a file name in command
     // line arguments.
     static std::string const STDIN_SPECIAL_NAME;
+
+    std::vector<std::chrono::microseconds> opApplySleepTimeForTesting;
```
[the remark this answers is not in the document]
hidenori-shinohara: Done!

src/main/CommandLine.cpp
```diff
@@ -1073,10 +1066,15 @@ run(CommandLineArgs const& args)
...
                 cfg = configOption.getConfig();
                 cfg.DISABLE_BUCKET_GC = disableBucketGC;
-                if (simulateSleepPerOp > 0)
+
+                if (!cfg.OP_APPLY_SLEEP_TIME_DURATION_FOR_TESTING.empty() ||
+                    !cfg.OP_APPLY_SLEEP_TIME_WEIGHT_FOR_TESTING.empty())
                 {
                     cfg.DATABASE = SecretValue{"sqlite3://:memory:"};
-                    cfg.OP_APPLY_SLEEP_TIME_FOR_TESTING = simulateSleepPerOp;
+                    cfg.opApplySleepTimeForTesting =
```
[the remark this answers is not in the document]
hidenori-shinohara: That makes sense. Moved it there.

src/main/Config.cpp
```diff
@@ -1679,5 +1744,12 @@ Config::toString(SCPQuorumSet const& qset)
     return fw.write(json);
 }
 
+const std::vector<std::chrono::microseconds>
```
MonsieurNicolas, line 1747: I think you ment to return a const reference

MonsieurNicolas (changes requested):
Only one minor issue from what I see before we can merge.

src/main/Config.cpp
hidenori-shinohara: Yes. Fixed it!

hidenori-shinohara:
@MonsieurNicolas I updated the const-reference issue, and I also squashed the commits

marta-lokhova:
r+ 51d8f8347affd0efcf58b831eae262c5f14e1546

MonsieurNicolas:
@latobarita: retry

PR #3035, opened 2021-04-24
Refresh dependencies April 2021

MonsieurNicolas:
Avoid drifting too much from latest stable.

Notable update is the fmt update that includes large performance improvement and bug fixes with gcc 8.

I didn't pick up sqlite's 3.35.x release as it's still churning


jonjove:
r+ 1e9f4cf

PR #3040, opened 2021-05-05
Bump to gcc 8 and llvm10

MonsieurNicolas:
(as Xenial support was dropped we're moving to latest supported by Ubuntu 18)


graydon (approved):

graydon:
r+ e99339a

PR #3042, opened 2021-05-12
Photesthesis - preliminary

graydon:
This is a very preliminary sketch of wiring up the [photesthesis](github.com/graydon/photesthesis) testing library to the path payment tests of stellar-core. It's to show roughly how it might work and explore ergonomic and comprehensibility issues. 

Feedback very welcome.

The current PR here is definitely a draft -- it has much of the path-payments file commented out! -- and contains only a single parametric test. That test can do some successful payments using multiple assets, paths and market makers. Running it over 25,000 random test plans from the grammar produces about 600 distinct trajectories, which is .. probably uncomfortably many to manually review still, but not many-orders-of-magnitude too many. My next planned step is to adapti the harness to filter out pointless cases that can't obviously be filtered by the grammar context system.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -23,3 +23,6 @@
 	path = lib/spdlog
 	url = https://github.com/gabime/spdlog
 	ignore = dirty
+[submodule "lib/photesthesis"]
+	path = lib/photesthesis
+	url = https://github.com/graydon/photesthesis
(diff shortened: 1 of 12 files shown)


MonsieurNicolas:
closing: stale and unlikely to land anytime soon

PR #3043, opened 2021-05-12
Add official Dockerfile and Makefile

jacekn:
# Add official Dockerfile and Makefile

This PR adds package-based Dockerfile which we'll use to publish the `stellar/stellar-core` docker image. It also includes convenience Makefile and small documentation updates

# Checklist
- [ ] ~~Ran `clang-format` v8.0.0 (via `make format` or the Visual Studio extension)~~
- [ ] ~~Compiles~~
- [ ] ~~Ran all tests~~
- [ ] ~~If change impacts performance, include supporting evidence per the [performance document](https://github.com/stellar/stellar-core/blob/master/performance-eval/performance-eval.md)~~


MonsieurNicolas:
cool. Maybe we should move all docker related things into a `Docker` folder as some files like `setup` and `Makefile` are a bit too generic looking without context

jacekn:
> cool. Maybe we should move all docker related things into a `Docker` folder as some files like `setup` and `Makefile` are a bit too generic looking without context

Done. In many other projects we use `docker` directory so I used that for consistency. If instead you'd like to capitalize it let me know and I can do that.
There is also top level `.devcontainer` directory which should probably be migrated to `docker`, I can add that if it's safe to do without affecting developers.

MonsieurNicolas:
Cool thanks, looking good

>  There is also top level .devcontainer directory which should probably be migrated to docker, I can add that if it's safe to do without affecting developers.

This is used by Visual Studio code, and it looks for it at the top level (at least it's a hidden folder)

PR #3030, opened 2021-04-22
Diagnostic tools

graydon:
This PR contains a set of enhancements to `stellar-core self-check` to do more thorough analysis of local state:
   - Comparing the expected bucketlist to the buckets found in the bucket directory
   - Checking bucket hashes
   - Decoding all bucket contents and comparing to all database ledger contents
   - Benchmarking the IO speeds, DB latency and cryptography speeds of the node
   - Checking that all history archives have a ledger header that is identical with the local database LCL header
 
These enhancements are divided into a small set that run online, during startup or when triggered by the HTTP endpoint (those that probe the history archive) and a larger set that run offline when the `self-check` command-line is run (all the IO intensive and database/bucket tests).

PR #3049, opened 2021-05-14
Make ApplyBucketsWork independent of merges in progress

### LINKED ISSUE #3044: Allow ApplyBucketsWork to use local HAS
Right now ApplyBucketsWork assumes it receives the HAS file from the history archives, so it performs `containsValidBuckets` check. The problem is that in a few places in the code we actually use local HAS files which are valid and trusted, but might not meet the archive validity criteria. So we should move `containsValidBuckets` check outside of the ApplyBucketsWork, and into catchup instead.

marta-lokhova:
Resolves #3044

This change removes the dependency on future buckets state from ApplyBucketsWork. Previously, ApplyBucketsWork would expect the future buckets in HAS to be either resolved or clear, as it expected HAS from an untrusted archive. This has changed, since we now use the local HAS in ApplyBucketsWork as well (in captive core startup, for example), which may have merges in progress. So this change moves the HAS validation into DownloadBucketsWork instead.

Note that it also fixes the race condition that impacts older (before protocol 12) buckets: the HAS validation expects some merges to be resolved, however `prepareForPublish` restarts the merges in progress, but does not necessarily wait for them to finish.

### DIFF
--- a/src/work/WorkWithCallback.cpp
+++ b/src/work/WorkWithCallback.cpp
@@ -1,0 +1,48 @@
+// Copyright 2021 Stellar Development Foundation and contributors. Licensed
+// under the Apache License, Version 2.0. See the COPYING file at the root
+// of this distribution or at http://www.apache.org/licenses/LICENSE-2.0
+
+#include "work/WorkWithCallback.h"
+#include "util/Logging.h"
+
+namespace stellar
+{
+
+WorkWithCallback::WorkWithCallback(
+    Application& app, std::string name,
+    std::function<bool(Application& app)> callback)
+    : BasicWork(app, std::move(name), BasicWork::RETRY_NEVER)
+    , mCallback(callback)
+{
+    if (!mCallback)
+    {
+        throw std::invalid_argument("Must provide a valid function. Use "
+                                    "BasicWork if a callback is not needed.");
+    }
+}
+
+BasicWork::State
+WorkWithCallback::onRun()
+{
+    bool res;
+    try
+    {
+        res = mCallback(mApp);
+    }
+    catch (std::runtime_error const& er)
+    {
+        CLOG_ERROR(Work,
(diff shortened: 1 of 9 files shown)


src/historywork/DownloadBucketsWork.cpp
```diff
@@ -17,8 +17,8 @@ namespace stellar
 DownloadBucketsWork::DownloadBucketsWork(
     Application& app, std::map<std::string, std::shared_ptr<Bucket>>& buckets,
     std::vector<std::string> hashes, TmpDir const& downloadDir,
-    std::shared_ptr<HistoryArchive> archive)
-    : BatchWork{app, "download-verify-buckets"}
+    std::shared_ptr<HistoryArchive> archive, std::function<bool()> verifyCb)
```
MonsieurNicolas, line 20: I think it's time to actually do https://github.com/stellar/stellar-core/pull/3030#discussion_r628649644 what do you think?

marta-lokhova: I added a new small work class `WorkWithCallback` to address this, let me know what you think.

src/work/WorkWithCallback.cpp
```diff
@@ -0,0 +1,48 @@
+// Copyright 2021 Stellar Development Foundation and contributors. Licensed
```
MonsieurNicolas, line 1: need to update Visual Studio files

MonsieurNicolas (changes requested):
this looks good. Just forgot about VC++ :)

src/work/WorkWithCallback.cpp
marta-lokhova: Fixed.

MonsieurNicolas:
r+ 55e5ddd60eed626ad23d1d3784d68c3f9572c618

PR #3057, opened 2021-05-20
Harden bucket apply path

MonsieurNicolas:
I noticed while investigating something that there could be situations during catchup where we do not properly clear the in memory state (this only applies to captive core).

This PR fixes that by ensuring that we rebuild the ledger every time we apply buckets.

### DIFF
--- a/src/catchup/ApplyBucketsWork.cpp
+++ b/src/catchup/ApplyBucketsWork.cpp
@@ -81,8 +81,7 @@ ApplyBucketsWork::onReset()
         // apply
         if (!mApp.getConfig().MODE_USES_IN_MEMORY_LEDGER)
         {
-            auto& lsRoot = mApp.getLedgerTxnRoot();
-            lsRoot.deleteObjectsModifiedOnOrAfterLedger(0);
+            mApp.resetLedgerState();
         }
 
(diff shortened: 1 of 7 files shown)


src/catchup/ApplyBucketsWork.cpp
```diff
@@ -77,6 +77,14 @@ ApplyBucketsWork::onReset()
 
     if (!isAborting())
     {
+        // clear ledgerTxn state of all ledger entries in preparation of bucket
+        // apply
+        if (!mApp.getConfig().MODE_USES_IN_MEMORY_LEDGER)
+        {
+            auto& lsRoot = mApp.getLedgerTxnRoot();
+            lsRoot.deleteObjectsModifiedOnOrAfterLedger(0);
```
marta-lokhova: I don't think we want to do this if the work fails?

MonsieurNicolas: yes we want to always reset on failure: otherwise we leave the ledger in a partially applied state (which is non obvious when debugging, when this could happen)

marta-lokhova:
r+ 1e0f0d1dbfe80bec37fd6074e610c194b77ef6d1

PR #3060, opened 2021-05-21
Warning police

graydon:
While diagnosing failures in #3059 I found gcc 9 was emitting a lot of warnings. Decided to mop them up.

### DIFF
--- a/src/main/Config.cpp
+++ b/src/main/Config.cpp
@@ -24,6 +24,7 @@
 #include <functional>
 #include <numeric>
 #include <sstream>
+#include <type_traits>
 #include <unordered_set>
 
 namespace stellar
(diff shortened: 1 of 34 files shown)


src/main/Config.cpp
```diff
@@ -275,10 +276,24 @@ readInt(ConfigItem const& item, T min = std::numeric_limits<T>::min(),
         throw std::invalid_argument(fmt::format("invalid '{}'", item.first));
     }
     int64_t v = item.second->as<int64_t>()->get();
-    if (v < min || v > max)
+    if (std::is_signed_v<T>)
```
MonsieurNicolas, line 279: C++ 17 makes this kind of things so nice

src/test/FuzzerImpl.h
```diff
@@ -22,7 +22,7 @@ struct Operation;
 namespace FuzzUtils
 {
 size_t static constexpr NUM_STORED_LEDGER_KEYS = 0x100U;
-size_t static constexpr NUM_UNVALIDATED_LEDGER_KEYS = 0x40U;
+ssize_t static constexpr NUM_UNVALIDATED_LEDGER_KEYS = 0x40;
```
MonsieurNicolas, line 25: `ssize_t` is not what you want (this is a C standard type not C++). I imagine that you're trying to silence the warning in `generateStoredLedgerKeys`?

If so the type should be `StoredLedgerKeys::iterator::difference_type` (on my machine this is `ptrdiff_t` but I don't know if this is true across platforms)

MonsieurNicolas (changes requested):

src/main/Config.cpp
graydon: Quite!

src/test/FuzzerImpl.h
graydon: Fixed (it's just `StoredLedgerKeys::difference_type`)

MonsieurNicolas:
r+ 0a6a8ca55f15945978a5bd710267e442bd181e87

PR #3064, opened 2021-05-26
Remove deprecated --reply-in-memory flag.

### LINKED ISSUE #2996: remove deprecated `--replay-in-memory`
while reviewing https://github.com/stellar/stellar-core/pull/2994 I realized that we have too many flag combinations

graydon:
# Description

Resolves #2996 -- removes a deprecated `--replay-in-memory` flag in favour of the newer general-purpose `--in-memory` flag.

Interestingly, populating the `replayInMemory` command-line flag was already disabled, but _using_ it was still enabled on one codepath (see the diff hunk where it was being passed to `Application::create`). That use winds up as the `createNewDB` argument to `ApplicationImpl`, but it's "supposed to be true" only in the in-memory-replay scenarios that also set the database to `sqlite3://:memory:` which then effectively interprets the `createNewDB` flag as true anyways. So the failure to pass a command-line-populated value for the `createNewDB` flag didn't really matter.

(In any case, this change restores it to actually pass the `inMemory` flag, which is populated properly).

# Checklist
- [x] Compiles
- [x] Ran all tests

### DIFF
--- a/src/main/CommandLine.cpp
+++ b/src/main/CommandLine.cpp
@@ -603,7 +603,6 @@ runCatchup(CommandLineArgs const& args)
     std::string archive;
     std::string trustedCheckpointHashesFile;
     bool completeValidation = false;
-    bool replayInMemory = false;
     bool inMemory = false;
     bool forceBack = false;
     uint32_t startAtLedger = 0;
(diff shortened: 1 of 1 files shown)


marta-lokhova (approved):
LGTM, thanks!

marta-lokhova:
r+ c3b27b60c59ac301602d562bb5be99db321e2e71

PR #3065, opened 2021-06-01
SCP: Slightly reduce dependency on crypto/SecretKey.h

Geod24:
Since #2048 it is much easier to use a different signing algorithm (and key type altogether) with SCP.
However, when isolating SCP into a self-contained library, one still has a dependency on `crypto/SecretKey.h`.
There are two reasons for this: First, there was a lingering `#include` in the `.h`, which has been moved to the `.cpp`.
Second, the `#include` is actually used for the functions `toStrKey` / `toShortString`. By first using `NodeID` for them instead of `PublicKey`, we can reduce downstream diff, as only the typedef will need to be changed (and the traits to be added).

# Checklist
- [x] Compiles
- [x] Ran all tests


MonsieurNicolas:
makes sense

PR #3068, opened 2021-06-03
Herder tracking cleanup and tests

marta-lokhova:
First set of changes in Herder:
* Improvements to the notion of "tracking" in Herder, plus tracking vs LCL invariants
* More tests for various externalize paths and ledger trigger scenario. I anticipate that the ledger trigger test suite will become more extensive once we introduce additional changes for the trigger path.

### DIFF
--- a/src/herder/Herder.h
+++ b/src/herder/Herder.h
@@ -67,8 +67,14 @@ class Herder
 
     enum State
     {
+        // Starting up, no state is known
+        HERDER_BOOTING_STATE,
+        // Never been in sync with the network, track LCL
+        HERDER_TRACKING_LCL_STATE,
+        // Fell out of sync, resyncing
         HERDER_SYNCING_STATE,
-        HERDER_TRACKING_STATE,
+        // Fully in sync with the network
+        HERDER_TRACKING_NETWORK_STATE,
         HERDER_NUM_STATE
     };
 
--- a/src/herder/Herder.h
+++ b/src/herder/Herder.h
@@ -96,7 +96,7 @@ class Herder
     };
 
     virtual State getState() const = 0;
-    virtual std::string getStateHuman() const = 0;
+    virtual std::string getStateHuman(State st) const = 0;
 
     // Ensure any metrics that are "current state" gauge-like counters reflect
     // the current reality as best as possible.
--- a/src/herder/Herder.h
+++ b/src/herder/Herder.h
@@ -69,8 +69,6 @@ class Herder
     {
         // Starting up, no state is known
         HERDER_BOOTING_STATE,
-        // Never been in sync with the network, track LCL
-        HERDER_TRACKING_LCL_STATE,
         // Fell out of sync, resyncing
         HERDER_SYNCING_STATE,
         // Fully in sync with the network
--- a/src/herder/Herder.h
+++ b/src/herder/Herder.h
@@ -122,11 +122,17 @@ class Herder
     // We are learning about a new envelope.
     virtual EnvelopeStatus recvSCPEnvelope(SCPEnvelope const& envelope) = 0;
 
+#ifdef BUILD_TESTS
     // We are learning about a new fully-fetched envelope.
     virtual EnvelopeStatus recvSCPEnvelope(SCPEnvelope const& envelope,
                                            const SCPQuorumSet& qset,
                                            TxSetFrame txset) = 0;
 
+    virtual void
+    externalizeValue(std::shared_ptr<TxSetFrame> txSet, uint32_t ledgerSeq,
+                     uint64_t closeTime,
+                     xdr::xvector<UpgradeType, 6> const& upgrades) = 0;
+#endif
     // a peer needs our SCP state
     virtual void sendSCPStateToPeer(uint32 ledgerSeq, Peer::pointer peer) = 0;
 
--- a/src/herder/HerderImpl.cpp
+++ b/src/herder/HerderImpl.cpp
@@ -1607,15 +1607,17 @@ HerderImpl::restoreSCPState()
     ZoneScoped;
     // setup a sufficient state that we can participate in consensus
     auto const& lcl = mLedgerManager.getLastClosedLedgerHeader();
+
+    setTrackingSCPState(lcl.header.ledgerSeq, lcl.header.scpValue, false);
+
     if (!mApp.getConfig().FORCE_SCP &&
         lcl.header.ledgerSeq == LedgerManager::GENESIS_LEDGER_SEQ)
     {
         // if we're on genesis ledger, there is no point in claiming
         // that we're "in sync"
+        lostSync();
         return;
     }
-
-    setTrackingSCPState(lcl.header.ledgerSeq, lcl.header.scpValue, false);
 
     trackingHeartBeat();
 
--- a/src/herder/HerderImpl.cpp
+++ b/src/herder/HerderImpl.cpp
@@ -654,6 +654,8 @@ HerderImpl::recvSCPEnvelope(SCPEnvelope const& envelope)
     return status;
 }
 
+#ifdef BUILD_TESTS
+
 Herder::EnvelopeStatus
 HerderImpl::recvSCPEnvelope(SCPEnvelope const& envelope,
                             const SCPQuorumSet& qset, TxSetFrame txset)
@@ -665,6 +667,19 @@ HerderImpl::recvSCPEnvelope(SCPEnvelope const& envelope,
     mPendingEnvelopes.addSCPQuorumSet(xdrSha256(qset), qset);
     return recvSCPEnvelope(envelope);
 }
+
+void
+HerderImpl::externalizeValue(std::shared_ptr<TxSetFrame> txSet,
+                             uint32_t ledgerSeq, uint64_t closeTime,
+                             xdr::xvector<UpgradeType, 6> const& upgrades)
+{
+    getPendingEnvelopes().putTxSet(txSet->getContentsHash(), ledgerSeq, txSet);
+    StellarValue sv = makeStellarValue(txSet->getContentsHash(), closeTime,
+                                       upgrades, mApp.getConfig().NODE_SEED);
+    getHerderSCPDriver().valueExternalized(ledgerSeq, xdr::xdr_to_opaque(sv));
+}
+
+#endif
 
 void
 HerderImpl::sendSCPStateToPeer(uint32 ledgerSeq, Peer::pointer peer)
--- a/src/herder/HerderImpl.cpp
+++ b/src/herder/HerderImpl.cpp
@@ -130,6 +130,15 @@ HerderImpl::trackingConsensusCloseTime() const
 void
 HerderImpl::setState(State st)
 {
+    bool initState =
+        st == HERDER_BOOTING_STATE || st == HERDER_TRACKING_LCL_STATE;
+    if (initState && (mState == HERDER_TRACKING_NETWORK_STATE ||
+                      mState == HERDER_SYNCING_STATE))
+    {
+        throw std::runtime_error(
+            fmt::format("Invalid state transition in Herder: {} -> {}",
+                        getStateHuman(mState), getStateHuman(st)));
+    }
     mState = st;
 }
 
@@ -155,12 +164,12 @@ HerderImpl::syncMetrics()
 }
 
 std::string
-HerderImpl::getStateHuman() const
+HerderImpl::getStateHuman(State st) const
 {
     static std::array<const char*, HERDER_NUM_STATE> stateStrings = {
         "HERDER_BOOTING_STATE", "HERDER_TRACKING_LCL_STATE",
         "HERDER_SYNCING_STATE", "HERDER_TRACKING_NETWORK_STATE"};
-    return std::string(stateStrings[getState()]);
+    return std::string(stateStrings[st]);
 }
 
 void
--- a/src/herder/HerderImpl.cpp
+++ b/src/herder/HerderImpl.cpp
@@ -531,10 +531,8 @@ HerderImpl::recvSCPEnvelope(SCPEnvelope const& envelope)
         maxLedgerSeq = mHerderSCPDriver.nextConsensusLedgerIndex() +
                        LEDGER_VALIDITY_BRACKET;
     }
-    else if (!checkCloseTime(
-                 envelope,
-                 (mHerderSCPDriver.lastTrackingSCP() == nullptr) &&
-                     (minLedgerSeq <= LedgerManager::GENESIS_LEDGER_SEQ)))
+    else if (!checkCloseTime(envelope, getCurrentLedgerSeq() <=
+                                           LedgerManager::GENESIS_LEDGER_SEQ))
     {
         // if we've never been in sync, we can be more aggressive in how we
         // filter messages: we can ignore messages that are unlikely to be
--- a/src/herder/HerderImpl.cpp
+++ b/src/herder/HerderImpl.cpp
@@ -104,12 +104,10 @@ HerderImpl::getState() const
 }
 
 void
-HerderImpl::setTrackingSCPState(uint64_t index, StellarValue const& value,
-                                bool isTrackingNetwork)
+HerderImpl::setTrackingSCPState(uint64_t index, StellarValue const& value)
 {
     mTrackingSCP = ConsensusData{index, value.closeTime};
-    setState(isTrackingNetwork ? Herder::HERDER_TRACKING_NETWORK_STATE
-                               : Herder::HERDER_TRACKING_LCL_STATE);
+    setState(Herder::HERDER_TRACKING_NETWORK_STATE);
 }
 
 uint32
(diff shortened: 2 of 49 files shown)


MonsieurNicolas (changes requested):

src/herder/HerderImpl.cpp
```diff
@@ -1577,14 +1587,8 @@ HerderImpl::restoreSCPState()
...
     auto const& lcl = mLedgerManager.getLastClosedLedgerHeader();
-    if (!mApp.getConfig().FORCE_SCP &&
-        lcl.header.ledgerSeq == LedgerManager::GENESIS_LEDGER_SEQ)
-    {
-        // if we're on genesis ledger, there is no point in claiming
-        // that we're "in sync"
-        return;
-    }
 
+    // TODO: is there a way to avoid waiting to go out of sync on genesis
```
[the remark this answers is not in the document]
marta-lokhova: Yeah, hence the TODO to avoid the delay. I've thought about going out of sync right away as well, and added this to my local changes.

src/herder/HerderSCPDriver.cpp
```diff
@@ -1132,4 +1139,26 @@ HerderSCPDriver::wrapStellarValue(StellarValue const& sv)
...
+void
+HerderSCPDriver::checkInvariants() const
+{
+    auto lcl = mApp.getLedgerManager().getLastClosedLedgerNum();
+    // Either tracking or last tracking must be set
+    // Tracking is ahead of or equal to LCL
+    if (mTrackingSCP)
+    {
+        releaseAssert(!mLastTrackingSCP);
+        releaseAssert(mTrackingSCP->mConsensusIndex >= lcl);
```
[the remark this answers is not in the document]
marta-lokhova: this isn't true because captive core defers overlay, and resets Herder _after_ the initial catchup is done, so tracking ends up being the same as LCL. That being said, as per our discussion elsewhere, the general approach we might want to take here is to introduce a "booting" state in Herder, and only check the invariants when Herder is properly tracking the network.

src/herder/Herder.h
```diff
@@ -67,8 +67,14 @@ class Herder
 
     enum State
     {
+        // Starting up, no state is known
```
MonsieurNicolas, line 70: you added new states but forgot to update the string array - I opened PR https://github.com/stellar/stellar-core/pull/3078 to catch this at compile time as this can crash the app

src/herder/Herder.h
```diff
@@ -67,8 +67,14 @@ class Herder
...
+        HERDER_BOOTING_STATE,
+        // Fell out of sync, resyncing
         HERDER_SYNCING_STATE,
-        HERDER_TRACKING_STATE,
+        // Never been in sync with the network, track LCL
+        HERDER_TRACKING_LCL_STATE,
```
MonsieurNicolas, line 75: should `HERDER_TRACKING_LCL_STATE` be defined before `HERDER_SYNCING_STATE` ?

I imagine that later on, the only valid transition is to bounce between `HERDER_TRACKING_LCL_STATE` and `HERDER_TRACKING_NETWORK_STATE`

MonsieurNicolas: Also, I find it confusing that `HERDER_TRACKING_LCL_STATE` is called "tracking" - I'd prefer to call it `HERDER_BOOTING_LCL_STATE` as it should only be in that state until it sees consensus.
That way the node can then bounce between `HERDER_SYNCING_STATE` and `HERDER_TRACKING_NETWORK_STATE`

src/herder/HerderImpl.cpp
```diff
@@ -98,7 +100,44 @@ HerderImpl::~HerderImpl()
...
+HerderImpl::trackingConsensusCloseTime() const
+{
+    releaseAssert(getState() != Herder::State::HERDER_BOOTING_STATE);
+    return mTrackingSCP.mConsensusCloseTime;
+}
+
+void
+HerderImpl::setState(State st)
+{
+    mState = st;
```
MonsieurNicolas, line 133: can we enforce that we don't go back to booting state? (
and related to this, I think it's invalid to go back to `HERDER_TRACKING_LCL_STATE`)

src/herder/HerderImpl.cpp
```diff
@@ -898,18 +934,8 @@ HerderImpl::getCurrentLedgerSeq() const
...
-        res = static_cast<uint32_t>(
-            mHerderSCPDriver.trackingSCP()->mConsensusIndex);
-    }
-    if (mHerderSCPDriver.lastTrackingSCP() &&
-        res < mHerderSCPDriver.lastTrackingSCP()->mConsensusIndex)
-    {
-        res = static_cast<uint32_t>(
-            mHerderSCPDriver.lastTrackingSCP()->mConsensusIndex);
-    }
+    res = std::max(res, trackingConsensusLedgerIndex());
```
MonsieurNicolas, line 937: this is not equivalent: old code was checking that tracking was set, new code would crash (as it tries to access index while it's not tracking)

src/herder/HerderImpl.cpp
```diff
@@ -399,16 +440,12 @@ HerderImpl::checkCloseTime(SCPEnvelope const& envelope, bool enforceRecent)
...
-        if (cd && envLedgerIndex >= cd->mConsensusIndex &&
-            cd->mConsensusIndex > lastCloseIndex)
-        {
-            lastCloseIndex = static_cast<uint32>(cd->mConsensusIndex);
-            lastCloseTime = cd->mConsensusValue.closeTime;
-        }
-    };
-    adjustLastCloseTime(mHerderSCPDriver.trackingSCP());
-    adjustLastCloseTime(mHerderSCPDriver.lastTrackingSCP());
+    auto trackingIndex = trackingConsensusLedgerIndex();
```
MonsieurNicolas, line 443: this is not equivalent: old code did not assume that it was tracking, this version would crash

src/herder/HerderImpl.cpp
MonsieurNicolas: (I expect a refactor commit to not make functional changes)

src/herder/HerderImpl.cpp
```diff
@@ -1132,7 +1158,7 @@ void
 HerderImpl::forceSCPStateIntoSyncWithLastClosedLedger()
 {
     auto const& header = mLedgerManager.getLastClosedLedgerHeader().header;
-    mHerderSCPDriver.restoreSCPState(header.ledgerSeq, header.scpValue);
+    setTrackingSCPState(header.ledgerSeq, header.scpValue, false);
```
MonsieurNicolas, line 1161: this should be `true`?
This function is used in "manual close" mode and I don't see a reason to distinguish it from the normal state "we're in sync, everything is fine"

src/herder/HerderImpl.cpp
```diff
@@ -1575,16 +1601,18 @@ HerderImpl::restoreSCPState()
     ZoneScoped;
     // setup a sufficient state that we can participate in consensus
     auto const& lcl = mLedgerManager.getLastClosedLedgerHeader();
+
+    setTrackingSCPState(lcl.header.ledgerSeq, lcl.header.scpValue, false);
```
MonsieurNicolas, line 1605: in this commit I would not expect to see this:
old code would just leave the app in `BOOTING` state?

MonsieurNicolas: after reviewing the commit as a whole, I am not sure we need (in this commit) `HERDER_TRACKING_LCL_STATE`. If we need it, it should be introduced in a later commit (otherwise we have too many things happening in the same commit, making things hard to review).

From this commit point of view I'd expect to
1. in the "genesis" case below, just leave the state untouched (so booting?)
2. after that state is just `HERDER_TRACKING_NETWORK_STATE`

MonsieurNicolas (changes requested):

src/herder/HerderImpl.cpp
marta-lokhova: > this is not equivalent: old code was checking that tracking was set, new code would crash (as it tries to access index while it's not tracking)

I did not mean this to be equivalent. In this PR, I am making the "tracking" invariant stronger, i.e. it is always set on startup (see the change to `restoreSCPState`). With this guarantee, it becomes invalid to call these functions while in booting state (as we saw in some tests that wouldn't start the app properly, for example). Like I said before, I had to squash commits because the rebase turned out to be pretty painful (there used to be individual commits with functional/non-functional changes), but I guess I can attempt it again to make it easier to review.

src/herder/HerderImpl.cpp
marta-lokhova: same thing here: I do not think this is true in the current state of the proposal as per my comment above.

src/herder/Herder.h
marta-lokhova: Good catch, this is fixed now

src/herder/Herder.h
marta-lokhova: > Also, I find it confusing that HERDER_TRACKING_LCL_STATE is called "tracking" - I'd prefer to call it HERDER_BOOTING_LCL_STATE as it should only be in that state until it sees consensus.
> That way the node can then bounce between HERDER_SYNCING_STATE and HERDER_TRACKING_NETWORK_STATE

This one is less clear to me. From the node's perspective, "tracking" means being in sync, i.e. ready to validate slots and participate in consensus. It seems cleaner to me to keep calling the state restored from DB as "tracking", not "booting", because as soon as node reaches that state, it's ready to vote (and it doesn't necessarily need to hear from the network to do so). This naming conversion also improves clarity and readability of helper functions such as `isTracking` (if we rename this state, this function becomes "HERDER_TRACKING_NETWORK_STATE || HERDER_BOOTING_LCL_STATE", which is pretty confusing). 

As I'm writing this, I'm wondering - maybe the issue is that both states should indicate that the node's Herder is synced? Maybe the state names should be something like "SYNCED_LOCAL_LCL" and "SYNCED_TRACKING_NETWORK", and `isTracking` can be renamed to `isSynced`. what do you think?

src/herder/HerderImpl.cpp
```diff
@@ -1608,16 +1608,18 @@ HerderImpl::restoreSCPState()
...
+
     if (!mApp.getConfig().FORCE_SCP &&
         lcl.header.ledgerSeq == LedgerManager::GENESIS_LEDGER_SEQ)
     {
         // if we're on genesis ledger, there is no point in claiming
         // that we're "in sync"
+        lostSync();
```
MonsieurNicolas, line 1628: I think we can do better than this in this commit:
looking at how the various states are used at this point, we can *remove* the `HERDER_TRACKING_LCL_STATE` state.

That way we get rid of the `bool` used by `setTrackingSCPState`.

Only difference will be:
when we startup we're "in sync" (tracking), unless we're on genesis ledger (out of sync, per this commit).

Other functional change is that `ApplicationImpl::getState()` can:
* if herder is "booting" --> application state is `APP_CREATED_STATE` (right now it lies with `APP_ACQUIRING_CONSENSUS_STATE`)
* if herder is not tracking --> application state is `APP_ACQUIRING_CONSENSUS_STATE`
* if herder is tracking --> use lm's state to determine state

difference with master is that on startup if "tracking lcl" but didn't see consensus yet, master claims "joining SCP" where as now it will say "Connected" (which is actually better). Comment above `APP_CONNECTED_STANDBY_STATE` should be updated to say "tracking" (as it doesn't mean "in sync").

That way after this commit we'll only have 2 modes for herder: tracking and "out of sync".

What do you think?

src/herder/test/PendingEnvelopesTests.cpp
```diff
@@ -28,6 +28,7 @@ TEST_CASE("PendingEnvelopes recvSCPEnvelope", "[herder]")
     auto& pk = s.getPublicKey();
     cfg.QUORUM_SET.validators.emplace_back(s.getPublicKey());
     Application::pointer app = createTestApplication(clock, cfg);
+    app->start();
```
MonsieurNicolas, line 31: Why do we need to start the application?
Should this be done by `createTestApplication` - it sounds like a bug to use the app without starting it (I guess there might be a few places where we want to do this, but maybe those places should be the ones passing a `start=false` parameter to `createTestApplication`?

src/transactions/test/TxEnvelopeTests.cpp
```diff
@@ -1532,8 +1533,11 @@ TEST_CASE("txenvelope", "[tx][envelope]")
             StellarValue sv = app->getHerder().makeStellarValue(
                 txSet->getContentsHash(), 1, emptyUpgradeSteps,
                 app->getConfig().NODE_SEED);
-            LedgerCloseData ledgerData(1, txSet, sv);
-            app->getLedgerManager().closeLedger(ledgerData);
+            auto& herder = static_cast<HerderImpl&>(app->getHerder());
+            herder.getPendingEnvelopes().putTxSet(txSet->getContentsHash(), 2,
```
MonsieurNicolas, line 1537: should this be a test only method exposed by `Herder` (this avoids importing `HerderImpl` and duplicate code) -- this is similar to the 3 param `recvSCPEnvelope(SCPEnvelope const& envelope, const SCPQuorumSet& qset, TxSetFrame txset)`

src/herder/HerderImpl.cpp
```diff
@@ -130,6 +130,15 @@ HerderImpl::trackingConsensusCloseTime() const
 void
 HerderImpl::setState(State st)
 {
+    bool initState =
+        st == HERDER_BOOTING_STATE || st == HERDER_TRACKING_LCL_STATE;
```
MonsieurNicolas, line 134: There might be a problem when the app runs "twice" in the `!MODE_AUTO_STARTS_OVERLAY` mode:
```
runApp(Application::pointer app):
if (!app->getConfig().MODE_AUTO_STARTS_OVERLAY)
        {
            app->getHerder().restoreState();


HerderImpl::restoreState() -> HerderImpl::restoreSCPState() -> setTrackingSCPState(lcl.header.ledgerSeq, lcl.header.scpValue, false);

```

How can we guarantee that the app didn't move to `HERDER_TRACKING_NETWORK_STATE` or `HERDER_SYNCING_STATE`  when `ApplicationImpl::start()` invoked `mHerder->restoreState();` when the app gets started the first time?

MonsieurNicolas (changes requested):

src/herder/HerderImpl.cpp
```diff
@@ -898,18 +938,11 @@ HerderImpl::getCurrentLedgerSeq() const
...
-        res = static_cast<uint32_t>(
-            mHerderSCPDriver.trackingSCP()->mConsensusIndex);
-    }
-    if (mHerderSCPDriver.lastTrackingSCP() &&
-        res < mHerderSCPDriver.lastTrackingSCP()->mConsensusIndex)
-    {
-        res = static_cast<uint32_t>(
-            mHerderSCPDriver.lastTrackingSCP()->mConsensusIndex);
-    }
+    uint32_t tracking =
```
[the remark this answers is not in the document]
marta-lokhova: this is not equivalent, we need to query consensus ledger index if Herder is "out of sync" (i.e. in `HERDER_SYNCING_STATE` state)

src/herder/HerderImpl.cpp
marta-lokhova: > difference with master is that on startup if "tracking lcl" but didn't see consensus yet, master claims "joining SCP" where as now it will say "Connected" (which is actually better).

this is the case when FORCE_SCP is set to false. If it's set to true, LM is set to "synced" state, so the app will return "Synced!" before it hears from the network. On one hand, logically it makes sense to me, as starting the app with FORCE_SCP implies that it is in sync. On the other hand, we might have external dependencies on app's "synced" state (for example, in simulations). Besides tests, does anything else depend on it?

marta-lokhova: @MonsieurNicolas so what do you think about "synced" being herder synced + lm synced + closed at least one ledger since startup? I know you also proposed using ledger age, but I think it's less accurate, because as soon as we start the app, ledger age will be pretty small, causing the app to go in sync.

src/transactions/test/TxEnvelopeTests.cpp
marta-lokhova: sure, I added a new test method to externalize a value.

src/herder/test/PendingEnvelopesTests.cpp
marta-lokhova: good idea, updated

src/herder/HerderImpl.cpp
MonsieurNicolas: per our offline chat, I think we can just consider the node "in sync" on startup in this case. Maybe we can improve that signal later on when we improve how we trigger (don't set LM as synced on startup: the only way LM gets to sync state is by closing a ledger)

src/herder/HerderImpl.cpp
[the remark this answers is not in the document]
MonsieurNicolas: Oh I see - isn't that code just checking that we're not in "BOOTING" state then?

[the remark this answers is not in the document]
marta-lokhova: yes, it's equivalent to using 0 when BOOTING. I updated the condition to check for BOOTING state explicitly though, seems a bit simpler this way.

marta-lokhova:
bad rebase, will fix (sorry!)

MonsieurNicolas:
r+ 3eb1588d3811e2c23e4b20725df6f6e24f634640

Issue #3067, opened 2021-06-03
Stellar Core missing config.h in for windows builds. (Version: 17.0.0)


wu-ziching:
## Issue Description

stellar-core/src/util/Backtrace.cpp contains an include:

`#include "config.h"`

This file is not generated during the build process for Windows, which means it must be created or copied over somehow in order to complete the build process outlined in the INSTALL-Windows.md instructions.

## Steps to Reproduce

Follow the INSTALL-Windows.md build instructions. Attempt to a Release build on the x64 architecture.

## Expected Result

Successful build.

## Actual Result

Visual Studio 2019 reports E1696: Cannot open source file: "config.h" in Backtrace.cpp

## Your Environment and Setup

Windows x64
Visual Studio Code 2019
PostgreSQL 13 (Project updated to reflect different include and linker input location)
Cygwin, installed Flex, Bison, curl, gcc-core, sed

### stellar-core Version
Master, v17.0.0

### Environment

Windows 10 x64
Visual Studio 2019
Release, x64 Build
Project updated to point to PostgreSQL 13 include and linker directory

## Additional context
Build works for v12.0.0, prior to the introduction of Backtrace.cpp.

MonsieurNicolas:
Thanks for the report. Let me try this out

MonsieurNicolas:
this was made irrelevant by the linked PR: we don't include that header anymore

PR #3069, opened 2021-06-04
Visual C++ fixes and warnings

MonsieurNicolas:
This PR works around a problem when building release builds with the latest version of Visual C++ (unclear if this was introduced recently or not).

It also fixes a few warnings that were introduced in a recent commit (mismatch between signed and unsigned types).

### DIFF
--- a/src/work/test/WorkTests.cpp
+++ b/src/work/test/WorkTests.cpp
@@ -770,10 +770,10 @@ TEST_CASE("WorkSequence test", "[work]")
 class TestBatchWork : public BatchWork
 {
     bool mShouldFail;
-    int mTotalWorks;
+    size_t mTotalWorks;
 
   public:
-    int mCount{0};
+    size_t mCount{0};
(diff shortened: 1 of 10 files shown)


MonsieurNicolas:
resolves #3067

src/work/test/WorkTests.cpp
```diff
@@ -770,10 +770,10 @@ TEST_CASE("WorkSequence test", "[work]")
 class TestBatchWork : public BatchWork
 {
     bool mShouldFail;
-    int mTotalWorks;
+    size_t mTotalWorks;
 
   public:
-    int mCount{0};
+    size_t mCount{0};
```
marta-lokhova, line 776: what was the warning here? Looks like we're using ints across the board from what I see

graydon:
r+ c8ea04e

src/work/test/WorkTests.cpp
MonsieurNicolas: `mTotalWorks` getting initialized with a `size_t`

PR #3071, opened 2021-06-04
SCP: Remove unused includes of crypto/SHA.h

Geod24:
SCP was made mostly independent of the underlying crypto implementation,
so those includes are just leftovers.


MonsieurNicolas:
r+ b554ce7a06e07446b1556ef5616d231049334061

PR #3078, opened 2021-06-12
make static arrays safe (old style doesn't enforce proper size at compile time)

MonsieurNicolas:
This compiles, but we really don't want it to be the case:
```
   std::array<int, 2> foo = { 100 }; // 1 element initialized!
   int bar[2] = { 100 }; // 1 element initialized!
```

We can force size checking by using C++17's type deduction that lets the compiler figure out the parameters to `std::array` .


marta-lokhova:
r+ 703817bdd73481400719438877ca098726b3c838

PR #3079, opened 2021-06-12
Initialize CommandHandler last

### LINKED ISSUE #3076: race conditions (crash) on startup
Current master sometimes crashes on startup. This became more visible because of the time it takes to rebuild state after schema upgrades.

From what I see the races are in `ApplicationImpl::initialize`:

* `mCommandHandler = std::make_unique<CommandHandler>(*this);` is initialized fairly early - this exposes the HTTP port (allowing people to call commands before the app is fully initialized)
*  `mProcessManager = ProcessManager::create(*this);` is called after `newDB` and `upgradeToCurrentSchemaAndMaybeRebuildLedger` - the later ends up calling `applyBucketsForLCL` which calls `crank` (which allows http commands to be invoked)

The crash I ran into is that the command `metrics` is called by prometheus probes which tries to update metrics for all subsystems, in particular `mProcessManager` (which is not initialized).

marta-lokhova:
This change fixes #3076 by initializing CommandHandler last, so the user can't issue HTTP commands until all of the Application components are fully initialized.

### DIFF
--- a/src/main/ApplicationImpl.cpp
+++ b/src/main/ApplicationImpl.cpp
@@ -239,7 +239,6 @@ ApplicationImpl::initialize(bool createNewDB, bool forceRebuild)
     mHistoryManager = HistoryManager::create(*this);
     mInvariantManager = createInvariantManager();
     mMaintainer = std::make_unique<Maintainer>(*this);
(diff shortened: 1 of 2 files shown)


MonsieurNicolas:
r+ 0f618e510de717f2de327a96edd52c4664377345

PR #3082, opened 2021-06-15
Herder: process old slots correctly

marta-lokhova:
This change fixes a bug in Herder, where older slots sometimes are processed as latest slots. The bug occurs if a node externalizes a ledger that is already closed but doesn't have messages from the network: for example, if the node is restarted with LCL=N, and receives enough messages to externalize N again.

The solution is to avoid reasoning based on `getCurrentLedgerSeq`, which is very confusing anyway, and explicitly pass `isLatestSlot` instead.


MonsieurNicolas:
r+ 0a2fedf8e47eba36fd4e0f02f8dc0a3da4c0624e

Issue #3054, opened 2021-05-18
Improve lifetime dependencies of core components on shutdown


marta-lokhova:
It looks like changes in https://github.com/stellar/stellar-core/pull/2994 pointed to some undesired lifetime dependencies between objects. Specifically, now the BucketManager is initialized first (and destructed last), but during destruction is relies on LedgerManager and HistoryManager being alive to perform garbage collection. The problem is that if the application throws _after_ it initializes BM but _before_ it initializes LM or HM, core will crash while trying to destruct BM. 

We could try to carefully evaluate the dependencies and somehow re-arrange initialization and destruction code, but I think a better approach here would be to re-think the dependencies in the first place. Specifically, does BucketManager really need to do garbage collection on destruction? We perform GC on ledger close, during graceful shutdown and on re-start, which seems  sufficient. If we don't GC on shutdown, this means that occasionally we might end up with extra stale buckets, but that doesn't seem to be a problem, since those will be cleaned up on restart anyway.

graydon:
I agree that it might be possible to avoid GC here, but I _think_ there were reasons for insisting on it. Possibly these can be dealt with another way, eg. by explicitly running GC at the end of certain testcases. I'd be willing to explore this possibility.

marta-lokhova:
Digging up some history, it seems like GC on shutdown was added as a preventative measure (as per this PR: https://github.com/stellar/stellar-core/pull/1345), so just removing it _might_ be ok (I'm currently skething/testing this change, so will report back).

MonsieurNicolas:
I am not even sure which scenario this catches at this point: afaik we GC on startup (in case we crashed) and when we close a ledger

PR #3084, opened 2021-06-15
BucketManager: move garbage collection out of shutdown

marta-lokhova:
Resolves https://github.com/stellar/stellar-core/issues/3054

This change removes GC call from shutdown, making the app explicitly call it in `gracefulStop` instead. This prevents issues when destructing the app that is not fully initialized. 

Note: we still need to call shutdown in graceful stop and in the destructor, as we use BucketManager's state to determine if we need to stop merges in-progress.


MonsieurNicolas:
r+ b17d4de7137be08904c02add8779757999842058

PR #3088, opened 2021-06-18
Add unit tests for application setup

### LINKED ISSUE #3061: Add captive core unit tests
Changes we made in 17.1.0 showed that we have some gaps in how we test captive core. We should add some basic unit tests to stellar-core. For example, given that captive core startup and setup is more involved now, we should verify that one the app is ready to run, it's in-memory state matches its bucketlist.

marta-lokhova:
Resolves #3061

### DIFF
--- a/src/main/test/ApplicationUtilsTests.cpp
+++ b/src/main/test/ApplicationUtilsTests.cpp
@@ -2,13 +2,17 @@
 // under the Apache License, Version 2.0. See the COPYING file at the root
 // of this distribution or at http://www.apache.org/licenses/LICENSE-2.0
 
+#include "crypto/Random.h"
+#include "history/HistoryArchiveManager.h"
 #include "history/test/HistoryTestsUtils.h"
+#include "invariant/BucketListIsConsistentWithDatabase.h"
 #include "ledger/LedgerTxn.h"
 #include "lib/catch.hpp"
 #include "main/Application.h"
 #include "main/ApplicationUtils.h"
 #include "main/CommandHandler.h"
 #include "main/Config.h"
+#include "simulation/Simulation.h"
 #include "test/TestUtils.h"
 #include "test/TxTests.h"
 #include "test/test.h"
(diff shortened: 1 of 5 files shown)


src/main/test/ApplicationUtilsTests.cpp
```diff
@@ -146,3 +150,197 @@ TEST_CASE("offline self-check works", "[applicationutils][selfcheck]")
...
+                    selectedLedger = lcl.header.ledgerSeq;
+                    selectedHash = binToHex(lcl.hash);
+                }
+
+                // Validator should publish at least two checkpoints
+                return loadDone &&
+                       simulation->haveAllExternalized(3 * checkpoint, 5);
+            },
+            50 * Herder::EXP_LEDGER_TIMESPAN_SECONDS, false);
+
```
graydon, line 261: Maybe require selectedLedger and selectedHash got populated here?

src/main/test/ApplicationUtilsTests.cpp
```diff
@@ -146,3 +150,197 @@ TEST_CASE("offline self-check works", "[applicationutils][selfcheck]")
...
+        }
+        SECTION("rebuild state")
+        {
+            SECTION("from buckets")
+            {
+                selectedLedger = lcl.header.ledgerSeq;
+                selectedHash = binToHex(lcl.hash);
+
+                auto app = setupApp(cfg2, clock, selectedLedger, selectedHash);
+                REQUIRE(app);
```
graydon, line 298: Maybe require no-replay happened here?

src/main/test/ApplicationUtilsTests.cpp
```diff
@@ -146,3 +150,197 @@ TEST_CASE("offline self-check works", "[applicationutils][selfcheck]")
...
+                REQUIRE(checkState(*app));
+                REQUIRE(app->getLedgerManager().getLastClosedLedgerNum() ==
+                        selectedLedger);
+            }
+            SECTION("via catchup")
+            {
+                // startAtLedger is behind LCL, reset to genesis and catchup
+                auto app = setupApp(cfg2, clock, selectedLedger, selectedHash);
+                REQUIRE(app);
+                REQUIRE(checkState(*app));
```
graydon, line 308: Maybe require replay happened here?

src/main/test/ApplicationUtilsTests.cpp
```diff
@@ -146,3 +150,197 @@ TEST_CASE("offline self-check works", "[applicationutils][selfcheck]")
...
+                REQUIRE(app->getLedgerManager().getLastClosedLedgerNum() ==
+                        selectedLedger);
+            }
+        }
+        SECTION("set meta stream")
+        {
```
graydon, line 314: There is a `readLcms` lambda over in src/ledger/tests/LedgerCloseMetaStreamTests.cpp that open a meta stream and extracts its content -- might hoist that out to a common test utility routine and use it here to check that the stream has what you expect.

graydon (changes requested):
In general this is awesome! Very glad to see tests here. Just a few suggested additions.

src/main/test/ApplicationUtilsTests.cpp
marta-lokhova: Good idea, done.

src/main/test/ApplicationUtilsTests.cpp
marta-lokhova: Added a combination of checks: `canRebuildInMemoryLedgerFromBuckets` and the state of LedgerManager (if catchup indeed happened, LM is in LM_CATCHING_UP_STATE state at the end)

src/main/test/ApplicationUtilsTests.cpp
marta-lokhova: Mmm, we do not write to the stream here though, at this is just to test setup (i.e. spinning up multiple apps on startup shouldn't impact the stream, a bug we had before).

marta-lokhova:
addressed your feedback @graydon, and followed up on some comments. Thanks for reviewing!

src/main/test/ApplicationUtilsTests.cpp
graydon: Ok. Thought it might be nice to check it produced the right thing, but I guess that's already being checked in the LedgerCloseMetaStreamTests themselves.

graydon:
r+ 89242fc

graydon (approved):

PR #3090, opened 2021-06-21
SCP / Quorum utils: Minor fixes to reduce dependencies

Geod24:
Another PR which reduces SCP dependencies. Not much to say here, I think the diff is pretty self-explanatory, and if it isn't, there's a commit message.
Best reviewed commit-by-commit. If needed, I can break it into smaller PRs.

### DIFF
--- a/src/scp/LocalNode.cpp
+++ b/src/scp/LocalNode.cpp
@@ -20,18 +20,17 @@
 namespace stellar
 {
 LocalNode::LocalNode(NodeID const& nodeID, bool isValidator,
-                     SCPQuorumSet const& qSet, SCP* scp)
-    : mNodeID(nodeID), mIsValidator(isValidator), mQSet(qSet), mSCP(scp)
+                     SCPQuorumSet const& qSet, SCPDriver& driver)
+    : mNodeID(nodeID), mIsValidator(isValidator), mQSet(qSet), mDriver(driver)
 {
     normalizeQSet(mQSet);
-    auto const& scpDriver = mSCP->getDriver();
-    mQSetHash = scpDriver.getHashOf({xdr::xdr_to_opaque(mQSet)});
+    mQSetHash = driver.getHashOf({xdr::xdr_to_opaque(mQSet)});
 
     CLOG_INFO(SCP, "LocalNode::LocalNode@{} qSet: {}",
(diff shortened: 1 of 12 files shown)


Geod24:
Added a tiny commit to make the Github CI more readable. It was slightly annoying that the most important information (the matrix variable) were barely readable.

Before:
Screen Shot 2021-06-21 at 15 58 08

After:
Screen Shot 2021-06-21 at 15 58 20

LMK if wanted / should be in another PR.

Geod24:
Ping

src/scp/LocalNode.cpp
```diff
@@ -20,18 +20,17 @@
 namespace stellar
 {
 LocalNode::LocalNode(NodeID const& nodeID, bool isValidator,
-                     SCPQuorumSet const& qSet, SCP* scp)
-    : mNodeID(nodeID), mIsValidator(isValidator), mQSet(qSet), mSCP(scp)
+                     SCPQuorumSet const& qSet, SCPDriver& driver)
```
MonsieurNicolas, line 22: nice

src/scp/LocalNode.cpp
```diff
@@ -381,8 +380,8 @@ LocalNode::findClosestVBlocking(SCPQuorumSet const& qset,
 Json::Value
 LocalNode::toJson(SCPQuorumSet const& qSet, bool fullKeys) const
 {
-    return toJson(qSet, [&](NodeID const& k)
-                  { return mDriver.toStrKey(k, fullKeys); });
+    return toJson(
+        qSet, [&](NodeID const& k) { return mDriver.toStrKey(k, fullKeys); });
```
MonsieurNicolas, line 384: weird that it decided to change format here?

MonsieurNicolas (approved):
Looks good. That was an easy one to review :)

MonsieurNicolas:
r+ 3cc9d1c686582e408899b19e64995de910f5cd87

src/scp/LocalNode.cpp
Geod24: I don't really know what is going on with `clang-format`. Every version does things differently 🤷 
I *guess* the logic here is, put all arguments on the same line if possible, but if not, the body of a lambda is a nice cutting point. Since the change reduced the length of the line, it probably went back to putting all args on the same line.

PR #3093, opened 2021-06-22
In memory mode wants soci transactions

graydon:
This shows two possible fixes for an atomicity bug on non-ledger auxiliary tables in in-memory mode. Neither is particularly pretty -- one is simpler and more local, one is uglier and more invasive but more systematic. We should only keep one.

### DIFF
--- a/src/ledger/InMemoryLedgerTxn.cpp
+++ b/src/ledger/InMemoryLedgerTxn.cpp
@@ -1,0 +1,87 @@
+// Copyright 2021 Stellar Development Foundation and contributors. Licensed
+// under the Apache License, Version 2.0. See the COPYING file at the root
+// of this distribution or at http://www.apache.org/licenses/LICENSE-2.0
+
+#include "ledger/InMemoryLedgerTxn.h"
+#include "ledger/LedgerTxnImpl.h"
+#include "util/GlobalChecks.h"
+
+namespace stellar
+{
+
+InMemoryLedgerTxn::InMemoryLedgerTxn(InMemoryLedgerTxnRoot& parent,
+                                     Database& db)
+    : LedgerTxn(parent), mDb(db)
+{
+}
+
+InMemoryLedgerTxn::~InMemoryLedgerTxn()
+{
+}
+
+void
+InMemoryLedgerTxn::addChild(AbstractLedgerTxn& child)
+{
+    if (mTransaction)
+    {
+        throw std::runtime_error(
+            "Adding child to already-open InMemoryLedgerTxn");
+    }
+    LedgerTxn::addChild(child);
+    mTransaction = std::make_unique<soci::transaction>(mDb.getSession());
+}
+
+void
+InMemoryLedgerTxn::commitChild(EntryIterator iter, LedgerTxnConsistency cons)
+{
+    if (!mTransaction)
+    {
+        throw std::runtime_error(
+            "Committing child to non-open InMemoryLedgerTxn");
+    }
+    try
+    {
+        LedgerTxn::commitChild(iter, cons);
+        mTransaction->commit();
+        mTransaction.reset();
+    }
+    catch (std::exception& e)
+    {
+        printErrorAndAbort("fatal error during commit to InMemoryLedgerTxn: ",
+                           e.what());
+    }
+    catch (...)
+    {
+        printErrorAndAbort(
+            "unknown fatal error during commit to InMemoryLedgerTxn");
+    }
+}
+
+void
+InMemoryLedgerTxn::rollbackChild()
+{
+    if (!mTransaction)
+    {
+        throw std::runtime_error(
+            "Rolling back child on non-open InMemoryLedgerTxn");
+    }
+    try
+    {
+        LedgerTxn::rollbackChild();
+        mTransaction->rollback();
+        mTransaction.reset();
+    }
+    catch (std::exception& e)
+    {
+        printErrorAndAbort(
+            "fatal error when rolling back child of InMemoryLedgerTxn: ",
+            e.what());
+    }
+    catch (...)
+    {
+        printErrorAndAbort(
+            "unknown fatal error when rolling back child of InMemoryLedgerTxn");
+    }
+}
+
+}
--- a/src/ledger/InMemoryLedgerTxn.h
+++ b/src/ledger/InMemoryLedgerTxn.h
@@ -1,0 +1,56 @@
+#pragma once
+
+// Copyright 2021 Stellar Development Foundation and contributors. Licensed
+// under the Apache License, Version 2.0. See the COPYING file at the root
+// of this distribution or at http://www.apache.org/licenses/LICENSE-2.0
+
+#include "database/Database.h"
+#include "ledger/InMemoryLedgerTxnRoot.h"
+#include "ledger/LedgerTxn.h"
+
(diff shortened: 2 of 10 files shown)


src/ledger/InMemoryLedgerTxn.cpp
```diff
@@ -0,0 +1,59 @@
+// Copyright 2021 Stellar Development Foundation and contributors. Licensed
+// under the Apache License, Version 2.0. See the COPYING file at the root
+// of this distribution or at http://www.apache.org/licenses/LICENSE-2.0
+
+#include "ledger/InMemoryLedgerTxn.h"
+#include "ledger/LedgerTxnImpl.h"
```
jonjove, line 6: Why do you actually need this include?

src/ledger/InMemoryLedgerTxn.cpp
```diff
@@ -0,0 +1,59 @@
...
+    {
+        throw std::runtime_error(
+            "Adding child to already-open InMemoryLedgerTxn");
+    }
+    LedgerTxn::addChild(child);
+    mTransaction = std::make_unique<soci::transaction>(mDb.getSession());
+}
+
+void
+InMemoryLedgerTxn::commitChild(EntryIterator iter, LedgerTxnConsistency cons)
```
jonjove, line 34: It's probably worthwhile to be more careful here and in `rollbackChild` about exception safety. For example, if `LedgerTxn::commitChild(...)` succeeds but `mTransaction->commit()` throws then this will reach an inconsistent state. Looking at `LedgerTxnRoot` might give you a model to work from.

jonjove (changes requested):
I think the second approach is reasonable and satisfies the principle of "easy to use correctly, hard to use incorrectly".

jonjove:
One downside of the second approach is that I don't think it will allow solving #3095 in the easiest/cleanest way possible (not that the first approach helps with this issue either). But moving in the direction of a "pure" in-memory root instead of a two-level one would probably help in that regard. Either way, we could always do this and then move further when we actually do #3095.

src/ledger/InMemoryLedgerTxn.cpp
```diff
@@ -0,0 +1,59 @@
+// Copyright 2021 Stellar Development Foundation and contributors. Licensed
```
MonsieurNicolas, line 1: missing windows changes

src/ledger/InMemoryLedgerTxn.cpp
MonsieurNicolas: and by model, what we end up doing is just `printErrorAndAbort` if this happens - but I agree it's better than throwing exceptions

MonsieurNicolas (changes requested):
approach from the first commit should be removed entirely: it's super error prone (so much so that it missed the other places where we may "throw at the wrong time" like when generating the genesis ledger, which may leave the database in a bad state).

Second approach looks to be fine; a potentially cleaner version would have been to use dedicated fields for lcl (we already have `mHeader`), has (missing right now) and meta (interestingly, we could use `mNextMetaToEmit` or something similar to it, to write into historical tables so we could move all this code in the "commit" path).

That way, only `LedgerTxnRoot` and `InMemoryLedgerTxn` have code actually writing to the database (in the `commitChild` function).

src/ledger/InMemoryLedgerTxn.cpp
graydon: For the definition of `stellar::EntryIterator::AbstractImpl`. It's only forward-declared in `LedgerTxn.h`

src/ledger/InMemoryLedgerTxn.cpp
graydon: Fixed.

graydon:
@jonjove I'm not sure #3095 would be difficult here -- we'd just have `InMemoryLedgerTxn` override `LedgerTxn::countObjects` and actually count its objects, no?

src/ledger/InMemoryLedgerTxn.cpp
graydon: Fixed.

graydon:
@MonsieurNicolas Fixed issues marked in review & pushed update.

As for the "cleaner" approach you're suggesting, I could see possibly pursuing it, but I think it would be a fair bit more invasive and involve reorganizing a fair bit more code.

MonsieurNicolas:
Fair point @graydon - I opened https://github.com/stellar/stellar-core/issues/3117 to track potentially making those type of changes

src/ledger/InMemoryLedgerTxn.cpp
jonjove: Ah that's unfortunate. Opened https://github.com/stellar/stellar-core/issues/3125 to make this cleaner at some future date.

src/ledger/InMemoryLedgerTxn.h
```diff
@@ -0,0 +1,55 @@
...
+namespace stellar
+{
+
+class InMemoryLedgerTxn : public LedgerTxn
+{
+    Database& mDb;
+    std::unique_ptr<soci::transaction> mTransaction;
+
+  public:
+    InMemoryLedgerTxn(AbstractLedgerTxnParent& parent, Database& db);
```
jonjove, line 47: I was responding to https://github.com/stellar/stellar-core/pull/3093#issuecomment-871912113, emphasizing that one of my big concerns was that `InMemoryLedgerTxn` currently can be used anywhere that `LedgerTxn` can be used. But calling `InMemoryLedgerTxn::countObjects` on a nested one would be an error (with the simple implementation, rather than the complete one which could just be done in `LedgerTxn` instead). In other words, it wouldn't be an interface that is hard to use incorrectly. But this lead me to a realization.

It occurred to me that this is currently the case but it doesn't _have_ to be the case. If this constructor had the signature `InMemoryLedgerTxn(InMemoryLedgerTxnRoot& parent, Database& db)` then the above error would no longer be possible. Even without the `countObjects` issue, this is still safer because the compiler will enforce where this can be used.

src/ledger/InMemoryLedgerTxn.h
```diff
@@ -0,0 +1,55 @@
...
+// mode. In-memory mode only holds the _ledger_ contents in memory; it still has
+// a "small" SQL database storing some additional tables, and we still want to
+// have transactional atomicity on those tables in regions of code we have a
+// LedgerTxn open. So that's the _purpose_.
+//
+// On to messy implementation details: in-memory mode is implemented by
+// replacing the normal LedgerTxnRoot with a stub class InMemoryLedgerTxnRoot
+// that never issues _any_ SQL, and then substituting a subclass of LedgerTxn as
+// a fake root that stores LEs in memory (like any other LedgerTxn) but that we
+// never commit to its parent at all -- only commit children _to_. This calss is
```
jonjove, line 20: Nitpick: typo "calss" -> "class"

graydon: fixed.

src/ledger/InMemoryLedgerTxn.h
graydon: Fixed (with a slight change to ApplicationImpl -- it needs 2 possible root members now since they have different types and you can't dynamically downcast a unique_ptr)

jonjove:
r+ dc79e7c

PR #3097, opened 2021-06-25
Additional startup checks

marta-lokhova:
The goal of this change is to spot database inconsistencies as early as possible. I moved ledger header insert to be done last, and kept the startup code reasoning based on `kLastClosedLedger` (so if the application is killed, on restart it would complain that ledger header does not exist). The other route was getting the max ledger header known, and comparing it against the rest of the database. The problem is that I don't think we can have this consistency guarantee, as I see some areas in the code (in particular, the `--force-back` functionality) that don't enforce consistency between ledger headers and HAS.


MonsieurNicolas:
r+ b4c2eb80faa16cfe3209ff402ac2cfd86a45f59b

PR #3124, opened 2021-07-06
Improve invariance in Herder

marta-lokhova:
Depends on https://github.com/stellar/stellar-core/pull/3068

This change introduces stronger invariance in Herder. Specifically, querying "tracking" now requires Herder to be fully booted. Additionally, tracking must always be ahead of or equal to LCL. Note that this is not the case currently in master, as when we run offline/captive core catchup, Herder ends up behind. As part of this change, catchup now ensures that Herder state is properly setup before we start applying ledgers.

### DIFF
--- a/src/catchup/CatchupWork.cpp
+++ b/src/catchup/CatchupWork.cpp
@@ -11,6 +11,7 @@
 #include "catchup/CatchupRange.h"
 #include "catchup/DownloadApplyTxsWork.h"
 #include "catchup/VerifyLedgerChainWork.h"
+#include "herder/Herder.h"
 #include "history/FileTransferInfo.h"
 #include "history/HistoryManager.h"
 #include "historywork/BatchDownloadWork.h"
@@ -32,6 +33,45 @@ namespace stellar
 
 uint32_t const CatchupWork::PUBLISH_QUEUE_UNBLOCK_APPLICATION = 8;
 uint32_t const CatchupWork::PUBLISH_QUEUE_MAX_SIZE = 16;
+
+static std::shared_ptr<LedgerHeaderHistoryEntry>
+getHistoryEntryForLedger(uint32_t ledgerSeq, FileTransferInfo const& ft)
+{
+    XDRInputFileStream in;
+    in.open(ft.localPath_nogz());
+
+    auto lhhe = std::make_shared<LedgerHeaderHistoryEntry>();
+
+    while (in && in.readOne<LedgerHeaderHistoryEntry>(*lhhe))
+    {
+        if (lhhe->header.ledgerSeq == ledgerSeq)
+        {
+            return lhhe;
+        }
+    }
+
+    return nullptr;
+}
+
+static bool
+setHerderStateTo(FileTransferInfo const& ft, uint32_t ledger, Application& app)
+{
+    auto entry = getHistoryEntryForLedger(ledger, ft);
+    if (!entry)
+    {
+        CLOG_ERROR(History,
+                   "Malformed catchup file does not contain "
+                   "LedgerHeaderHistoryEntry for ledger {}",
+                   ledger);
+        return false;
+    }
+
+    app.getHerder().setTrackingSCPState(ledger, entry->header.scpValue,
+                                        /* isTrackingNetwork */ false);
+    CLOG_INFO(History, "Herder state is set! tracking={}, closeTime={}", ledger,
+              entry->header.scpValue.closeTime);
+    return true;
+}
 
 CatchupWork::CatchupWork(Application& app,
                          CatchupConfiguration catchupConfiguration,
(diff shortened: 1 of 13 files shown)


src/catchup/CatchupWork.cpp
```diff
@@ -300,6 +321,40 @@ CatchupWork::runCatchupStep()
...
+        if (!entry)
+        {
+            CLOG_ERROR(History,
+                       "Malformed catchup file does not contain "
+                       "LedgerHeaderHistoryEntry for ledger {}",
+                       catchupRange.last());
+            return BasicWork::State::WORK_FAILURE;
+        }
+
+        mApp.getHerder().setTrackingSCPState(catchupRange.last(),
```
MonsieurNicolas, line 352: Are you sure this is correct?

We only verify the ledger header (in particular if we use trusted hash) at step `// Step 4.1: Download and verify ledger chain` - so this seems to potentially call `setTrackingSCPState` with bad data?

I would imagine that we should call `setTrackingSCPState` as late as possible (but before we close ledgers).

src/catchup/CatchupWork.cpp
```diff
@@ -351,6 +351,8 @@ CatchupWork::runCatchupStep()
 
         mApp.getHerder().setTrackingSCPState(catchupRange.last(),
                                              entry->header.scpValue);
+        // As we're catching up, do not claim to be in sync
+        mApp.getHerder().lostSync();
```
MonsieurNicolas, line 355: Why do we need this? If catchup is triggered when "online" this is not necessary for sure.

Also, I am not sure we need to expose `lostSync` like this: I could see performing something like that in `setTrackingSCPState` if we force herder to "jump forward" for example.

marta-lokhova: Note that this code doesn't run during online catchup due to the if condition - we only move Herder forward if it's behind. In case of online catchup, tracking should always be ahead or up to date (that's how we trigger catchup in the first place)

I agree that it's not the most intuitive way to manually call this function in catchup. I added this because I discovered a bug in my initial implementation. Specifically, when force-setting Herder to "tracking", I wasn't setting up out of sync timer, so the node would never go out of sync to read the latest ledgers from the network. So one option was to set Herder to tracking and set timers correctly. The other option was to start out of sync and let the timers be set when we hear from the network (this is similar to the startup code in Herder::restoreSCPState). I figured that for this use case it's unlikely that we'll be in sync after finishing catchup (the use case being captive core run with --start-at-ledger), but if you think otherwise, let me know.

src/catchup/CatchupWork.cpp
marta-lokhova: yes, you're totally right, this should be done later. Will fix.

marta-lokhova: Moved this to after ledger verification success

src/catchup/CatchupWork.cpp
marta-lokhova: refactored this slightly: `setTrackingSCPState` can now set either tracking or syncing state, so external systems don't need to explicitly call lostSync. Note that this change made the startup path in `HerderImpl::start` a little cleaner too

src/catchup/StateConsistencyWork.cpp
```diff
@@ -0,0 +1,90 @@
...
+BasicWork::State
+EnsureHerderConsistencyWork::doWork()
+{
+    // No-op if Herder is already up to date
+    if (mApp.getHerder().trackingConsensusLedgerIndex() >= mLedger)
+    {
+        return State::WORK_SUCCESS;
+    }
+
+    auto checkpoint =
```
MonsieurNicolas, line 34: this doesn't seem right: I think we already downloaded and verified the last ledger in the range with `VerifyLedgerChainWork` (that also checks that it matches the trusted anchor if set), this new Work is downloading a potentially different ledger header that is not trusted (also, it doesn't deal with malicious checkpoint files at all).

MonsieurNicolas (changes requested):
aside from how we restore state for the last ledger in the range (offline mode), things look pretty good

src/catchup/StateConsistencyWork.cpp
marta-lokhova: you're right: this work doesn't need to download anything, we just need to reuse the file that's already verified (not sure why I left it there...). I updated it now.

MonsieurNicolas (changes requested):
looking very close - just one question

src/catchup/CatchupWork.cpp
```diff
@@ -403,6 +404,13 @@ CatchupWork::runCatchupStep()
             }
 
             std::vector<std::shared_ptr<BasicWork>> seq;
+            // Before replaying ledgers, ensure Herder state is consistent with
+            // LCL
+            auto consistencyWork =
+                std::make_shared<EnsureHerderConsistencyWork>(
```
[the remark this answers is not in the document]
marta-lokhova: Mostly for clarity: placing this work in a WorkSequence makes the order of events clear. I can switch to using WorkWithCallback (I was trying to avoid adding more stuff to CatchupWork as it's pretty huge and getting hard to follow)

MonsieurNicolas:
Looks good., We can merge it on Monday - I'm waiting to see a clean supercluster run.

MonsieurNicolas:
r+ e06705553c937f7421db5bd35cd1824e18f6b327

PR #3129, opened 2021-07-09
Fix the logging logic of LoadGenerator to take care of the division-by-0

hidenori-shinohara:
The logging logic of `LoadGenerator.cpp` doesn't handle the case when `applyTx.one_minute_rate() = 0` which leads to the division-by-0 error. This PR fixes that by replacing the `one_minute_rate()` with 1 when smaller than 1. It's used for calculating the ETA, and thus slight inaccuracy there doesn't seem like a big concern.

# Checklist
- [x] Compiles
- [x] Ran all tests


MonsieurNicolas:
r+ 8d8a6d6355c6f2b35513eea53ee77d07a727df28

PR #3130, opened 2021-07-12
Herder: cleanup trigger ledger mechanism

### LINKED ISSUE #3111: Herder: improve ledger trigger mechanism
Currently, Herder uses a combination of tracking and LedgerManager state checks to decide if it needs to trigger next ledger, but the code is very error-prone and hard to follow (and buggy in certain edge cases). It would be much easier to reason about this logic if instead we had a signal from LedgerManager to trigger next ledger when a new ledger is closed.

marta-lokhova:
Resolves #3111

### DIFF
--- a/src/herder/HerderImpl.cpp
+++ b/src/herder/HerderImpl.cpp
@@ -198,7 +198,7 @@ HerderImpl::bootstrap()
     mLedgerManager.moveToSynced();
     mHerderSCPDriver.bootstrap();
 
-    maybeTriggerNextLedger();
+    setupTriggerNextLedger();
     newSlotExternalized(
         true, mLedgerManager.getLastClosedLedgerHeader().header.scpValue);
 }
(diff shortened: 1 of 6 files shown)


src/herder/HerderImpl.cpp
```diff
@@ -918,15 +918,11 @@ HerderImpl::eraseBelow(uint32 ledgerSeq)
 }
 
 void
-HerderImpl::ledgerClosed()
+HerderImpl::cleanup()
```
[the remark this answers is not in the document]
MonsieurNicolas: Also, I don't really like that we're creating more mini functions that need to be called exactly right: I am not sure why there is drift between what happens in `bootstrap` and the "externalize" path.

I think we should be instead consolidating/reorder things a bit such that we can have 1 function that performs something like:
```
        // start timing next externalize from this point
        mLastExternalize = mApp.getClock().now();

        // perform cleanups
        TxSetFramePtr externalizedSet =
            mPendingEnvelopes.getTxSet(value.txSetHash);
        updateTransactionQueue(externalizedSet->mTransactions);

        cleanup(); // replace by actual content 
        safelyProcessSCPQueueUpToIndex(nextConsensusLedgerIndex(), false);  // maybe replace by actual content?
```

src/herder/test/HerderTests.cpp
```diff
@@ -2016,10 +2016,15 @@ TEST_CASE("herder externalizes values", "[herder]")
             if (i == ledgers.size() - 1)
             {
                 checkSynced(*(getC()));
+                // All the buffered ledgers are applied by now, so it's safe to
+                // trigger the next ledger
+                REQUIRE(herderC.mTriggerNextLedgerSeq == fourth + 1);
```
MonsieurNicolas, line 2021: this check is incomplete: what you mean here is that there is a "trigger timer" setup in the future and that it's "the right ledger" (same applies to other places where you check `mTriggerNextLedgerSeq` directly)

src/herder/HerderImpl.cpp
```diff
@@ -844,6 +836,17 @@ HerderImpl::safelyProcessSCPQueueUpToIndex(uint64 nextIndex, bool synchronous)
     }
 }
 
+void
+HerderImpl::lastClosedLedgerIncreased()
+{
+    releaseAssert(isTracking());
```
MonsieurNicolas, line 844: note that in the context of this PR this is fine, but those invariants may not hold in the future:
for example, if we want to be able to "trigger" at the end of catchup without having to wait for an extra ledger.

MonsieurNicolas (changes requested):
looks pretty close to me. I made a few suggestions to improve clarity

src/herder/test/HerderTests.cpp
marta-lokhova: I added a check for trigger timer setup. I left the section `older ledgers externalize and no-op` as-is, since it's supposed to skip triggering an older ledger, and the application may or may not have a trigger setup for the most recent ledger.

marta-lokhova:
addressed your comments @MonsieurNicolas in the most recent commit. I'll squash when we're ready to merge

MonsieurNicolas:
looks good. Please rebase @marta-lokhova

marta-lokhova:
@MonsieurNicolas done!

MonsieurNicolas:
r+ 21433d8b7320e3a0c4e5c57023cf3e7194c715ab

PR #3138, opened 2021-07-16
ApplyLedgerWork fails correctly

### LINKED ISSUE #3131: crash when failing to catchup
Looks like recent changes cause core to crash when catching up and there is a hash mismatch (offline catchup).

```
2021-07-14T22:55:48.298 GD7I3 [Tx INFO] applying ledger 247686 (txs:4, ops:6)
2021-07-14T22:55:48.323 GD7I3 [Tx INFO] applying ledger 247687 (txs:3, ops:9)
2021-07-14T22:55:48.364 GD7I3 [Tx INFO] applying ledger 247688 (txs:4, ops:6)
2021-07-14T22:55:48.398 GD7I3 [Tx INFO] applying ledger 247689 (txs:2, ops:5)
2021-07-14T22:55:48.421 GD7I3 [Tx INFO] applying ledger 247690 (txs:4, ops:9)
2021-07-14T22:55:48.476 GD7I3 [History ERROR] Replay failed: Local node's ledger corrupted during close
2021-07-14T22:55:48.476 GD7I3 [History ERROR] Archive core_testnet_002 maybe contains corrupt file transactions/00/03/c7/transactions-0003c7bf.xdr.gz
stellar-core: work/BasicWork.cpp:47: virtual stellar::BasicWork::~BasicWork(): Assertion `isDone() || mState == InternalState::PENDING' failed.

Aborted (core dumped)
```

I suspect this is related to how we shutdown (again).

marta-lokhova:
Resolves #3131

### DIFF
--- a/src/catchup/ApplyCheckpointWork.cpp
+++ b/src/catchup/ApplyCheckpointWork.cpp
@@ -250,96 +250,75 @@ BasicWork::State
 ApplyCheckpointWork::onRun()
 {
     ZoneScoped;
-    try
+    if (mConditionalWork)
     {
-        if (mConditionalWork)
-        {
-            mConditionalWork->crankWork();
+        mConditionalWork->crankWork();
 
-            if (mConditionalWork->getState() == State::WORK_SUCCESS)
+        if (mConditionalWork->getState() == State::WORK_SUCCESS)
+        {
+            auto& lm = mApp.getLedgerManager();
+
+            CLOG_DEBUG(History, "{}",
+                       xdr_to_string(lm.getLastClosedLedgerHeader(),
+                                     "LedgerManager LCL"));
+
+            CLOG_DEBUG(History, "{}",
+                       xdr_to_string(mHeaderHistoryEntry, "Replay header"));
+            if (lm.getLastClosedLedgerHeader().hash != mHeaderHistoryEntry.hash)
             {
-                auto& lm = mApp.getLedgerManager();
-
-                CLOG_DEBUG(History, "{}",
-                           xdr_to_string(lm.getLastClosedLedgerHeader(),
-                                         "LedgerManager LCL"));
-
-                CLOG_DEBUG(History, "{}",
-                           xdr_to_string(mHeaderHistoryEntry, "Replay header"));
-                if (lm.getLastClosedLedgerHeader().hash !=
-                    mHeaderHistoryEntry.hash)
-                {
-                    mApplyLedgerFailure.Mark();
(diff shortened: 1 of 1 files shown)


src/catchup/ApplyCheckpointWork.cpp
```diff
@@ -323,12 +323,6 @@ ApplyCheckpointWork::onRun()
         mConditionalWork->startWork(wakeSelfUpCallback());
         return State::WORK_RUNNING;
     }
-    catch (InvariantDoesNotHold&)
-    {
-        // already displayed e.what()
-        CLOG_ERROR(History, "Replay failed");
-        throw;
-    }
```
MonsieurNicolas, line 331: you can stop including the related header

MonsieurNicolas: that said... I think the intent of the throw here was to ensure that we get a core dump in this kind of situation

MonsieurNicolas: never mind: this exception cannot happen anymore as we queue up an `ApplyLedgerWork`, this was missed when we made the `ApplyLedgerWork` change

src/catchup/ApplyLedgerWork.cpp
```diff
@@ -23,7 +24,15 @@ BasicWork::State
 ApplyLedgerWork::onRun()
 {
     ZoneScoped;
-    mApp.getLedgerManager().closeLedger(mLedgerCloseData);
+    try
+    {
+        mApp.getLedgerManager().closeLedger(mLedgerCloseData);
+    }
+    catch (std::exception const& e)
```
MonsieurNicolas, line 31: I am not quite sure this is the right fix:
with this change, we stop crashing if there is some catastrophic problem when closing a ledger, and just retry etc

the reason we want to crash is that the core file may contain critical information that would allow to diagnose the issue.

I *think* that the original problem is not so much that we crash but that we crash with the wrong context (an assert in Work).

MonsieurNicolas: Looking at the kind of exceptions that `CloseLedger` can throw, I am not sure there are any situations where we want to fail the "work" as all situations indicate some severe problem with the local node
(I was trying to see if we should crash only for a subset of exceptions, like what was done before in `ApplyCheckpointWork`)

MonsieurNicolas (changes requested):

MonsieurNicolas:
r+ d3b9ddea5bcd393352a92148a1d93caa4b5a4e76

PR #3142, opened 2021-07-16
Remove herder bogus recovery timer

MonsieurNicolas:
I am seeing warnings on latest master. not a big deal, just confusing for people:
`HerderImpl::outOfSyncRecovery called when tracking`
until the node sees consensus from the network.

That's because we call `startOutOfSyncTimer` in `restoreSCPState`, even though it's not needed n(node is initialized as "tracking a few lines above").

We used to setup at that location the broadcast timer before I got rid of it in #2756 and I just forgot to remove it.

### DIFF
--- a/src/ledger/LedgerHashUtils.h
+++ b/src/ledger/LedgerHashUtils.h
@@ -52,10 +52,9 @@ getAssetHash(T const& asset)
     }
     case stellar::ASSET_TYPE_POOL_SHARE:
     {
-        size_t res = asset.type();
         res ^= stellar::shortHash::computeHash(
             stellar::ByteSlice(getLiquidityPoolID(asset).data(), 8));
-        return res;
+        break;
     }
     default:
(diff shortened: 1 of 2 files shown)


src/ledger/LedgerHashUtils.h
```diff
@@ -52,10 +52,9 @@ getAssetHash(T const& asset)
     }
     case stellar::ASSET_TYPE_POOL_SHARE:
     {
-        size_t res = asset.type();
```
graydon, line 55: This is a functional change:  you're removing the asset type from contributing to the hash. Not sure you want to do that.

graydon: Oh, never mind, I expanded the diff a bit further and saw it's set above as well.

marta-lokhova:
r+ 2d4bab287ef2e86522046d65cfdeaae3fcc505eb

PR #3143, opened 2021-07-20
devcontainer: add c/c++ extension

leighmcculloch:
# Description

## What
Add the Microsoft C/C++ extension to the VSCode devcontainer configuration that is specified for this repository. When using the devcontainer this will have the C/C++ extension installed during devcontainer setup.

## Why
The extension does a decent job of providing autocompletion and code navigation features to this code base. It would be helpful for new folks using the devcontainer or anyone picking it up if the extension was installed automatically so that new folks do not need to figure out what the best extensions to install are to get navigation in VSCode.

The extension is developed by Microsoft so seemingly not a security risk to have it installed automatically since Microsoft also develop VSCode itself.

# Checklist
- [x] Compiles
- [x] Ran all tests

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -32,5 +32,7 @@
 	// "postCreateCommand": "uname -a",
 
 	// Add the IDs of extensions you want installed when the container is created in the array below.
-	"extensions": []
+	"extensions": [
+		"ms-vscode.cpptools"
+	]
 }


MonsieurNicolas:
I don't see a reason to do this:
* for new folks, VS Code already offers to install that extension with a popup when opening any C++ file, and the default is to install the official Microsoft extension
* for others, it's actually not that useful. For example, for me:
   * in many cases I don't actually need extensions installed
   * when I need extensions, my workflow is to install in the container the same extensions that I have installed locally (one click), and this includes many more extensions

PR #3148, opened 2021-07-22
Resize ccache in ci-build.sh

jonjove:
# Description
ci-build.sh specifies a `CCACHE_MAXSIZE` that is too small. I did some testing about this locally and I can see that the ccache size is about 376.8 MB (without extrachecks, since I was testing this on Mac) using the ccache compression settings from ci-build.sh without the max size. Unfortunately, the max size is set to 300 MB and ccache evicts down to 20% below max by default to avoid repetitive clean up. So the cache is approximately a factor of 2 too small. From my testing, this makes caching much more effective and improves build time by more than 50%.

# Checklist
- [x] Compiles
- [x] Ran all tests

### DIFF
--- a/ci-build.sh
+++ b/ci-build.sh
@@ -115,7 +115,7 @@ export CCACHE_DIR=$HOME/.ccache
 export CCACHE_COMPRESS=true
 export CCACHE_COMPRESSLEVEL=9
 # cache size should be large enough for a full build
-export CCACHE_MAXSIZE=300M
+export CCACHE_MAXSIZE=500M
 export CCACHE_CPP2=true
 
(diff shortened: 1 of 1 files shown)


MonsieurNicolas:
This is probably fine: we'll merge as soon as master reopens.

The issue with caches is that there is a global limit per repository of 5 GB as per https://github.com/actions/cache#cache-limits with each build taking 1 GB, this means we have the capacity for 5 full builds, which includes PRs.

MonsieurNicolas:
r+ cf9bdaf4a18720130f8ad7e8cefa8b4c25d458e9

PR #3149, opened 2021-07-26
Add `maintainer::shutdown()` and stop maintenance after a crash

### LINKED ISSUE #3106: Crash on shutdown while performing maintenance
Needs to be investigated. The crash occurred when stellar-core was restarted (as part of v17.2.0rc2 -> v.17.2.0 upgrade):
```
2021-06-30T00:03:36.491 GCW5Q [Tx INFO] applying ledger 223875 (txs:2, ops:6)
2021-06-30T00:03:36.505 GCW5Q [Ledger INFO] Closed ledger: [seq=223875, hash=5231b0]
2021-06-30T00:03:40.193 GCW5Q [default INFO] got signal 15, shutting down
2021-06-30T00:03:40.526 GCW5Q [default INFO] Performing maintenance
2021-06-30T00:03:40.526 GCW5Q [History INFO] Trimming history <= ledger 150920 (rmin=150920, qmin=223875, lmin=223811)
2021-06-30T00:03:41.405 GBH7K [default INFO] Starting stellar-core stellar-core 17.2.0 (e47d483433ce4f7eaf0a0512e2556beb482c8851)
2021-06-30T00:03:41.407 GBH7K [Database INFO] Connecting to: postgresql://dbname=stellar user=stellar password=******** host=/var/run/postgresql
2021-06-30T00:03:41.412 GBH7K [SCP INFO] LocalNode::LocalNode@GBH7K qSet: 59d361
2021-06-30T00:03:41.414 GBH7K [Database INFO] DB schema is in current version
```

Note that maintenance ran _after_ core received the shutdown signal, so there might be an issue related to object lifetime during shutdown

hidenori-shinohara:
# Description

Resolves #3106 

The message `got signal 15, shutting down` mentioned in #3106 comes from `gracefulStop`. This PR adds the `shutdown()` method in `Maintainer` and call it from `gracefulStop` to prevent it from performing any maintenance after `shutdown`.

# Checklist
- [x] Compiles
- [x] Ran all tests

### DIFF
--- a/src/main/ApplicationImpl.cpp
+++ b/src/main/ApplicationImpl.cpp
@@ -741,6 +741,10 @@ ApplicationImpl::gracefulStop()
     {
         mHerder->shutdown();
     }
+    if (mMaintainer)
+    {
+        mMaintainer->shutdown();
+    }
 
     mStoppingTimer.expires_from_now(
         std::chrono::seconds(SHUTDOWN_DELAY_SECONDS));
--- a/src/main/Maintainer.cpp
+++ b/src/main/Maintainer.cpp
@@ -16,7 +16,8 @@
 namespace stellar
 {
 
-Maintainer::Maintainer(Application& app) : mApp{app}, mTimer{mApp}
+Maintainer::Maintainer(Application& app)
+    : mApp{app}, mTimer{mApp}, mShuttingDown{false}
 {
 }
 
@@ -64,6 +65,10 @@ void
 Maintainer::performMaintenance(uint32_t count)
 {
     ZoneScoped;
+    if (mShuttingDown)
+    {
+        return;
+    }
     LOG_INFO(DEFAULT_LOG, "Performing maintenance");
     auto logSlow = LogSlowExecution(
         "Performing maintenance", LogSlowExecution::Mode::AUTOMATIC_RAII,
@@ -72,5 +77,16 @@ Maintainer::performMaintenance(uint32_t count)
         std::chrono::seconds{2});
     ExternalQueue ps{mApp};
     ps.deleteOldEntries(count);
+}
+
+void
+Maintainer::shutdown()
+{
+    if (mShuttingDown)
+    {
+        return;
+    }
+    mShuttingDown = true;
+    mTimer.cancel();
 }
 }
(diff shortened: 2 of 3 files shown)


hidenori-shinohara:
Though there may be other ways this bug occurs, the following command reproduced the bug.

```
ps ax  | grep "stellar-core run" | head -n 1 | awk '{print $1;}' | xargs kill -15 && \
    stellar-core http-command --conf docs/stellar-core_example.cfg "maintenance?queue=true"
```

After the change in this PR, this command no longer leads to the issue.

MonsieurNicolas:
What was the actual crash?
I think this fix may hide the problem instead of fixing it:
maintenance is just an event scheduled by a timer, and we're supposed to kill of events when we shutdown.

What that tells me is that maybe the method `shutdownMainIOContext` is not called at the right time in some situation?

hidenori-shinohara:
I'll look into `shutdownMainIOContext` more.

Is the issue that the maintenance ran after core received the shutdown signal? Or is the issue that the core crashed during the restart as a part of v17.2.0rc2 -> v.17.2.0 upgrade?

I had the impression that it is the former, but I just want to make sure that I'm working on the right problem.

MonsieurNicolas:
I think the problem was that it crashed on shutdown (while running maintenance) yes

heifner:
`bool mShuttingDown;` is not initialized and therefore is undefined behavior on access.

marta-lokhova:
as discussed in https://github.com/stellar/stellar-core/issues/3106, I think the crash is actually unrelated to maintenance (the problem seems to be in `self-check`)

MonsieurNicolas:
`self-check` follows the same pattern: schedules a timer that is supposed to not reschedule itself when cancelled (as it's using `VirtualTimer::onFailureNoop`), but for whatever reason we're still rescheduling timers when we shutdown - so as per my previous comment I suspect there is some problem with the shutdown sequence (it's supposed to go through a period of time where it's cancelling everything and draining events).

marta-lokhova:
@MonsieurNicolas The events are shutdown as expected. The problem is that there is a delay between WorkScheduler shutdown and IO service shutdown, which gives the self-check timer an opportunity to get rescheduled. This is problematic because at that point WorkScheduler is already in ABORTED state. 
https://github.com/stellar/stellar-core/blob/cb833ebd1baca4bcf14c3c2296dda4705ca38a8f/src/main/ApplicationImpl.cpp#L745

hidenori-shinohara:
Closing this as the issue is unrelated to maintenance. See the discussion in #3106

PR #3152, opened 2021-07-27
replace use of "assert" by releaseAssert*

### LINKED ISSUE #2852: consider replacing `assert` in our codebase

MonsieurNicolas:
# Description

Resolves #2852

### DIFF
--- a/src/bucket/Bucket.cpp
+++ b/src/bucket/Bucket.cpp
@@ -20,12 +20,12 @@
 #include "main/Application.h"
 #include "medida/timer.h"
 #include "util/Fs.h"
+#include "util/GlobalChecks.h"
 #include "util/Logging.h"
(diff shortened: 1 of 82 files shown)


graydon:
@MonsieurNicolas I notice some of these but not all were changed to releaseAssertOrThrow -- is there a rationale for which changed which way?

MonsieurNicolas:
yes: I made asserts that can trigger as part of applying transactions "throw" instead of crash. This is to stay consistent with the rest of the code base.

graydon:
r+ 17174edb52f1ae782ed5fc9cd9960c551e14bdeb

PR #3164, opened 2021-08-10
Check test tx meta in GitHub CI

graydon:
This PR does a few organizational touchups to the TxMeta record-and-check machinery and then attempts to wire it in to github actions / CI.

The CI adds a new dimension to the test matrix, so it should run in parallel. This doubles the parallel job count to 8 but we should still not be hitting the github free-tier limit of 20 parallel jobs.

When checking TxMeta, we build a variant without the asan and extrachecks instrumentation. Those are handled in the other matrix entries. The TxMeta check is just checking that transaction semantics haven't changed.

With `--all-versions` the TxMeta baselines are somewhat large. I split them up into seprate baselines, one per test file, to help reduce the size (and the size of any diffs), as well as switching from sha256 to SIPHash (totally adequate for this purpose) and base64-encoding the result. At this point the baseline consumes 3MB. It gzips to 424KB but I left it as JSON in the interests of readable diffs.


src/test/test.cpp
```diff
@@ -600,69 +614,88 @@ recordOrCheckGlobalTestTxMetadata(TransactionMeta const& txMeta)
...
-    Json::Value root;
-    in >> root;
     size_t n = 0;
-    for (auto entry = root.begin(); entry != root.end(); ++entry)
+    for (auto const& dirent : stdfs::directory_iterator{dir})
     {
-        std::string name = entry.key().asString();
-        std::vector<Hash> hashes;
-        for (auto const& h : *entry)
+        auto path = dirent.path();
```
MonsieurNicolas, line 626: should filter out entries that are not files and files without the `json` extension?

src/test/test.cpp
```diff
@@ -600,69 +614,88 @@ recordOrCheckGlobalTestTxMetadata(TransactionMeta const& txMeta)
...
+        for (auto const& testCasePair : filePair.second)
+        {
+            Json::Value& hashes = fileRoot[testCasePair.first];
+            for (auto const& h : testCasePair.second)
+            {
+                hashes.append(binToHex(h));
+            }
+        }
+        stdfs::path path = dir / filePair.first;
+        path.replace_extension("json");
```
MonsieurNicolas, line 690: nice - almost makes me cry to see a standard way of replacing extensions

ci-build.sh
```diff
@@ -26,6 +28,11 @@ while [[ -n "$1" ]]; do
             export TEMP_POSTGRES=1
             echo Using temp database
             ;;
+    "--check-test-tx-meta")
+            # When testing against a TxMeta baseline, we turn off instrumentation
+            # to have a faster build & test cycle.
+            INSTRUMENTATION_CONFIG=""
+            CHECK_TEST_TX_META=1
```
MonsieurNicolas, line 35: missing `;;`

ci-build.sh
```diff
@@ -26,6 +28,11 @@ while [[ -n "$1" ]]; do
             export TEMP_POSTGRES=1
             echo Using temp database
             ;;
+    "--check-test-tx-meta")
+            # When testing against a TxMeta baseline, we turn off instrumentation
+            # to have a faster build & test cycle.
+            INSTRUMENTATION_CONFIG=""
```
MonsieurNicolas, line 34: re: `INSTRUMENTATION_CONFIG` approach.
Concern I have here is that we're doubling the size of caches (as we build asan and non asan builds) and we're going to hit the global Github limit between the `auto` branch and PRs.

I don't know if having that leg faster makes any difference as we're limited by the slowest build leg anyways - so maybe we should just use the same build flavors?

An alterrnative would be to do something like what we used to do in Travis: split the "build" step from "test".

* first we'd perform the "build" step, that would just build without running tests, and would populate the cache
* "test" steps would build again, but with caches the time spent building was very low

ci-build.sh
```diff
@@ -166,11 +173,20 @@ if [ $TEMP_POSTGRES -eq 0 ] ; then
...
 fi
 
-export ALL_VERSIONS=1
-export NUM_PARTITIONS=$((NPROCS*2))
-export RUN_PARTITIONS
 ulimit -n 256
-time make check
+
+if [ $CHECK_TEST_TX_META -eq 1 ] ; then
+    time src/stellar-core test \
```
MonsieurNicolas, line 179: This doesn't look right: you should be able to run in parallel.

All you need to do is override
`STELLAR_CORE_TEST_PARAMS` and `TEST_SPEC` before calling `make check`

MonsieurNicolas (changes requested):

src/test/test.cpp
graydon: Fixed.

src/test/test.cpp
graydon: Yes :)

ci-build.sh
graydon: Fixed.

ci-build.sh
graydon: Hmm. Looking into this: github says its cache limit is 5GB. If I do a clean build with the CFLAGS in ci-build.sh (`-g1 -O2`), ccache says I use 1GB.

We are currently doing 4 builds but the protocol dimension appears to only affect a single file (Config.cpp) so I suspect we really only have 1 dimension (toolchain) to work with and should eliminate the protocol dimension -- a cache miss on `Config.cpp` is harmless, so long as the presence of the additional preprocessor symbol `ENABLE_NEXT_PROTOCOL_VERSION_UNSAFE_FOR_PRODUCTION` doesn't cause ccache misses across the board (I will check on this and propose a patch to work around it if so).

Given that, I think we have room to have a 2-dimensional / 4-build cache: {gcc,clang} and {instrumentation,none}.

Another option I'd be happy to explore is turning off `-g1` from one or more of the builds. It's not clear to me that we get a lot of value out of the debuginfo _in CI_ (if there's a failure we tend to diagnose locally) and a no-debuginfo build only uses 400MB.

If you prefer to just press ahead on this and get it working as quickly as possible, I concur that recycling the existing cache / not disabling instrumentation on this dimension is the simplest change. Happy to explore any of the options listed above.

jonjove: > If I do a clean build with the CFLAGS in ci-build.sh (`-g1 -O2`), ccache says I use 1GB.

That number doesn't sound right to me. I was testing this recently with regard to setting the cache size correctly, and I recall that 1GB is approximately the size of the cache in the absence of any compression flags. But ci-build.sh does set some compression flags (`CCACHE_COMPRESS=true` and `CCACHE_COMPRESSLEVEL=9`), and in that case the result should be 300-400 MB based on my experience. The 300-400 MB number is also corroborated by the fact that the max cache size of 500 MB does actually work effectively.

Regardless of the above, it seems highly desirable if we can avoid caching on the protocol dimension so I'm interested in that proposal.

ci-build.sh
graydon: Ah, I didn't know those existed for setting parameters for make check. Fixed.

ci-build.sh
graydon: Followup: I did some testing with ccache and it appears that the change in command-line defines (that is: adding `-DENABLE_NEXT_PROTOCOL_VERSION_UNSAFE_FOR_PRODUCTION`) does not actually cause a cache miss.

It does cause a _direct mode_ cache miss (where ccache merely hashes the command line and inputs), but when that happens ccache falls back to _preprocessor mode_ where it runs `cpp` and hashes the resulting output. Since this symbol has no effect on the majority of the files, they still get a cache hit at this point.

However, I see that there is a 500MB cache limit (as Jon mentions) and I believe this has to do with the desire to have multiple caches in the github store, in order to support _updating_ the cache using hourly timestamps and cache-key fallbacks.

Based on these observations, I suggest (and will modify the PR to conform to) the following:

  - Remove the protocol dimension of the cache key
  - Remove the scenario dimension of the cache key
  - Remove the code that disables instrumentation in TxMeta-checking scenario

In other words, only a toolchain dimension will remain in the cache key, and all builds will have instrumentation enabled.

We may still want to remove debuginfo to give us more breathing room wrt. the cache limit, but for now I will leave it enabled. I measured with `CFLAGS=-g1 -O2` and `CCACHE_COMPRESSLEVEL=9` and it appears to only be 185MB -- compression helps a lot! -- so we're probably ok leaving it on.

src/test/test.cpp
```diff
@@ -675,7 +685,13 @@ saveTestTxMeta(stdfs::path const& dir)
             Json::Value& hashes = fileRoot[testCasePair.first];
             for (auto const& h : testCasePair.second)
             {
-                hashes.append(binToHex(h));
+                uint64_t tmp = h;
+                std::vector<uint8_t> buf;
+                for (size_t i = sizeof(uint64_t); i > 0; --i)
+                {
+                    buf.emplace_back(uint8_t(tmp >> (8 * (i - 1))));
```
MonsieurNicolas, line 711: pretty sure we need to & 0xFF here to avoid tripping verifiers (I know Visual C++ catches this by default, I imagine valgrind might as well)

lib/util/stdrandom.h
```diff
@@ -0,0 +1,369 @@
...
+struct __log2_imp<0, _Rp>
+{
+    static const size_t value = _Rp + 1;
+};
+
+template <class _UIntType, _UIntType _Xp>
+struct __log2
+{
+    static const size_t value = __log2_imp<_Xp,
+                                         sizeof(_UIntType) * __CHAR_BIT__ - 1>::value;
```
MonsieurNicolas, line 135: doesn't build for me on Windows: `__CHAR_BIT__` is not standard

src/util/RandHasher.cpp
```diff
@@ -0,0 +1,32 @@
...
+
+size_t gMixer{0};
+bool gHaveInitialized{false};
+static std::mutex gInitMutex;
+
+void
+initialize()
+{
+    std::lock_guard<std::mutex> guard(gInitMutex);
+    if (!gHaveInitialized)
```
MonsieurNicolas, line 23: you could also do what was done in the header (use a static "local" variable: this is perfectly defined behavior and thread safe per standard)

src/util/RandHasher.h
```diff
@@ -3,22 +3,27 @@
...
 {
   public:
     size_t
     operator()(T const& t) const
     {
-        static size_t mixer =
-            stellar::rand_uniform<size_t>(std::numeric_limits<size_t>::min(),
-                                          std::numeric_limits<size_t>::max());
-        return Hasher()(t) ^ mixer;
+        releaseAssert(randHash::gHaveInitialized);
```
MonsieurNicolas, line 25: I guess we now use the same random value per type, but this is probably fine

.github/workflows/build.yml
```diff
@@ -32,9 +33,9 @@ jobs:
         with:
           path: |
             /home/runner/.ccache
-          key: ${{ runner.os }}-${{ matrix.toolchain }}-${{ matrix.protocol }}-cacheID-${{ steps.cache_extra_id.outputs.id }}
+          key: ${{ runner.os }}-${{ matrix.toolchain }}-cacheID-${{ steps.cache_extra_id.outputs.id }}
```
MonsieurNicolas, line 36: maybe add a comment here that explains why we can omit `protocol` (very few files need to be recompiled) for future us that will probably forget about this

MonsieurNicolas (changes requested):

src/util/RandHasher.cpp
graydon: No, in this case we explicitly want to differentiate the case of "user initialized this variable by calling `initialize` from a known position in the code" (which means it'll always have the same relationship to gRandomEngine's seed) and "variable was initialized whenever it was first used" (which may change the point in gRandomEngine's trajectory it initializes from, depending on workload, eg. we can get a different initializer if we run with `[tx]` than we get if we run with `someSpecificTest`).

So we do not want a compiler-generated initialization flag, we want one we manually track so that we can treat failure-to-set-it as an error.

src/util/RandHasher.h
graydon: Yeah, it's possible there's some way to arrange it to still be per-type while supporting a separate initialization flag visible in each using-type; but I didn't think it worth bothering to figure out how.

src/test/test.cpp
graydon: Ok, fixed.

lib/util/stdrandom.h
graydon: Ah, missed that. There's a compat header in libc++ that defines it to `CHAR_BIT` (no underscores) which is in `<limits.h>`. Fixed by copying that hunk in.

.github/workflows/build.yml
graydon: Sure, fixed.

lib/util/stdrandom.h
MonsieurNicolas: still doesn't build for me. Looks like the fix broke intrinsics

lib/catch.hpp
```diff
@@ -13,6 +13,7 @@
 // start catch.hpp
 
 
+#include <stdrandom.h>
```
MonsieurNicolas, line 16: I think that we do not need those changes anymore:
* we should not be depending on any random functionality in catch
* I would prefer if we can avoid making changes to Catch as it makes picking up changes potentially harder. Also note that we have other dependencies on `<random>` (grep -r to see where those are, I am seeing libmedida and tracy potentially)

lib/autocheck/include/autocheck/generator.hpp
```diff
@@ -2,6 +2,7 @@
 #define AUTOCHECK_GENERATOR_HPP
 
 #include <random>
+#include <stdrandom.h>
```
MonsieurNicolas, line 5: you need to do `#include "lib/util/stdrandom.h` like in other places (does not build on Windows otherwise)

src/test/test.cpp
```diff
@@ -636,6 +636,36 @@ recordOrCheckGlobalTestTxMetadata(TransactionMeta const& txMetaIn)
...
+static char const* TESTKEY_PROTOCOL_VERSION = "!cfg protocol version";
+static char const* TESTKEY_RNG_SEED = "!rng seed";
+static char const* TESTKEY_ALL_VERSIONS = "!test all versions";
+static char const* TESTKEY_VERSIONS_TO_TEST = "!versions to test";
+
+template <typename T>
+void
+checkTestKeyVal(std::string const& k, T const& expected, T const& got,
+                std::string const& path)
+{
```
MonsieurNicolas, line 649: standard compliance issue: you're declaring `path` as an `std::string` but you're calling with a `std::filesystem::path` and there is no implicit way to convert a path into a string (for whatever reason clang is happy to do it but not VC++)

MonsieurNicolas (changes requested):

src/test/test.cpp
```diff
@@ -67,6 +67,7 @@ struct ReseedPRNGListener : Catch::TestEventListenerBase
         shortHash::seed(sCommandLineSeed);
         Catch::rng().seed(sCommandLineSeed);
         autocheck::rng().seed(sCommandLineSeed);
+        randHash::initialize();
```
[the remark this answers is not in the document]
graydon: fixed.

lib/catch.hpp
graydon: yeah, I have a PR open on medida actually waiting on your review already: https://github.com/stellar/medida/pull/23

I figured we could ignore tracy since it's not something we read back anywhere in our code and we frequently compile it out. if you want I can patch it too though.

I'll remove the catch change and install a `make check` step that greps code for use of these things in the future.

lib/autocheck/include/autocheck/generator.hpp
graydon: fixed.

lib/util/stdrandom.h
graydon: Fixed on windows.

src/test/test.cpp
graydon: fixed

MonsieurNicolas:
I merged the medida change so you can pick it up in this PR.

Looks like you caught an actual use of `std::uniform_int_distribution` that was added recently

graydon:
@MonsieurNicolas fixed test it hit, regenerated results, picked up medida change, rebased. Should be ready to go again.

MonsieurNicolas:
r+ d121eabd5264d509d334ddd5612dc5310e11989e

PR #3168, opened 2021-08-12
Windows compile fix

MonsieurNicolas:
Project file was referencing the same file twice and `IGNORE` is a macro defined in one of the Windows headers (and causes conflicts).


jonjove:
r+ c0e597e

PR #3174, opened 2021-08-16
Add CONFIGURE_FLAGS argument to Dockerfile.testing

2opremio:
For next-protocol testing with Horizon, we need to build core with
--enable-next-protocol-version-unsafe-for-production , for which it's
handy to be able to provide the argument to `docker build`.


MonsieurNicolas:
r+ 7713d5deff81d053e210c5dc3bf6497cebd951d1

PR #3177, opened 2021-08-20
Improve performance of transaction processing

marta-lokhova:


MonsieurNicolas (approved):
Just a minor thing and I think it's ready to merge

graydon:
Basic structure looks ok. I'm trying to think if there are any ways to maybe make it at all less accident-prone:

  - Use an enum with values `READ_ONLY_WITHOUT_SQL_TXN` and `READ_WRITE_WITH_SQL_TXN`
  - Assert that read-only mode is on main thread (any non-main thread definitely has to worry about concurrency)
  - Track the count of read and write ltxs in a pair of mutex-guarded global counters, and assert mutual exclusion between multiple concurrent readers and at most a single writer
 
What do you think?

src/ledger/LedgerTxn.h
```diff
@@ -635,13 +635,20 @@ class LedgerTxn : public AbstractLedgerTxn
     std::unique_ptr<Impl> const& getImpl() const;
 
   public:
+    // WARNING: use useTransaction flag with caution. It does not start a SQL
```
jonjove, line 638: How big is the performance difference between using a read-only transaction and no transaction at all? I'm asking because I want to make sure that the performance difference justifies the risk (and by association the giant warning).

src/ledger/LedgerTxn.h
```diff
@@ -635,13 +635,20 @@ class LedgerTxn : public AbstractLedgerTxn
...
+    // transaction, which uses the strongest SERIALIZABLE level isolation.
+    // Therefore, if you have concurrent transactions, you are risking getting
+    // inconsistent view of the database. Only use this mode for read-only
+    // transactions with no concurrent writers present.
     explicit LedgerTxn(AbstractLedgerTxnParent& parent,
-                       bool shouldUpdateLastModified = true);
-    explicit LedgerTxn(LedgerTxn& parent, bool shouldUpdateLastModified = true);
+                       bool shouldUpdateLastModified = true,
+                       bool useTransaction = true);
```
jonjove, line 645: I'm not sure that this interface satisfies the principle of easy to use correctly, hard to use incorrectly. Let me give you an example. At a certain moment, a maintainer believes that a certain `LedgerTxn` actually represents a read-only transaction with no concurrent writers. The maintainer sets `useTransaction = false`. Later the code evolves such that the parent of that `LedgerTxn` is not guaranteed to be a true root. The code still compiles, but there is a performance regression.

It would be great if it was only possible to compile the code in cases where the flag actually made sense.

jonjove (commented):
I agree with @graydon that anything we can do to make this less accident-prone is valuable. We should thoroughly explore that space.

src/ledger/LedgerTxn.h
marta-lokhova: I made the change to set the transaction as read-only and ran a few tests on my machine with Postgres. The results are similar to what I observed comparing this change to master, i.e. no transaction still yields better performance (in the image below yellow is this change, red is read-only transaction)
Screen Shot 2021-09-07 at 3 55 23 PM

jonjove: Interesting, thanks for testing! Slightly surprising to me but supported by https://stackoverflow.com/questions/33332706/does-postgresql-run-some-performance-optimizations-for-read-only-transactions.

src/ledger/LedgerTxn.h
marta-lokhova: Discussed with Jon that it is pretty annoying/invasive to make this change with the current structure of LedgerTxn roots. We can make the fix later, so I opened an issue to track this: https://github.com/stellar/stellar-core/issues/3197

src/ledger/LedgerTxn.cpp
```diff
@@ -264,13 +267,13 @@ LedgerTxn::getImpl() const
...
-LedgerTxn::addChild(AbstractLedgerTxn& child)
+LedgerTxn::addChild(AbstractLedgerTxn& child, bool useTransaction)
 {
-    getImpl()->addChild(child);
+    getImpl()->addChild(child, useTransaction);
 }
 
 void
-LedgerTxn::Impl::addChild(AbstractLedgerTxn& child)
+LedgerTxn::Impl::addChild(AbstractLedgerTxn& child, bool useTransaction)
```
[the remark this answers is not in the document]
marta-lokhova: updated

marta-lokhova:
@graydon I implemented the first two suggestions. I was looking into the last one, and realized that we actually should not have competing ltxs - each application uses a single "root", which allows a single child ltx (and therefore, one transaction) at a time. I think we'd need to track soci transactions instead, but that seems much more involved/invasive, so I am deferring for now.

graydon:
LGTM (or at least .. I mean .. I'm a little scared of the whole solution space, but I think it's as good as it can be)

MonsieurNicolas:
r+ 97c9afc268ba6809e116821c41b0373bd8e48558

PR #3179, opened 2021-08-24
Add --enable-tracy-csvexport

graydon:
This is a few pieces of tracy infrastructure:

  - Enables the `tracy-csvexport` tool, that converts a `tracy-capture` file to CSV
  - Adds `scripts/DiffTracyCSV.py` that does a high-level statistical filter-and-comparison between two CSV files
  - Bumps the tracy submodule to get a change I submitted upstream to allow time-limited runs of `tracy-capture`
  - Adds the `tracy-{capture,csvexport,gui}` programs to `bin_PROGRAMS` in `Makefile.am` which (I think) means they will be both automatically built and installed and, more importantly, included in any debian package that is built with them enabled.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -1,5 +1,6 @@
 ## Process this file with automake to produce Makefile.in
 
+bin_PROGRAMS =
 SUBDIRS = lib src
 AM_EXTRA_RECURSIVE_TARGETS = check-valgrind
 ACLOCAL_AMFLAGS = -I m4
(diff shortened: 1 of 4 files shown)


MonsieurNicolas:
r+ 11d08dd567c009fa90aec25846a1b165682fd833

graydon:
@MonsieurNicolas I think you pasted the revision number from some other PR here.

MonsieurNicolas:
r+ 1390e6854eba73ea7f45c1d77458c76d47c393c9

MonsieurNicolas:
Sorry there is an annoying bug in FF that sometimes it doesn't copy/paste properly

PR #3180, opened 2021-08-25
Improve performance of bucket apply for captive core

MonsieurNicolas:
# Description

This PR makes some performance improvements to captive core.

On my test machine (Intel Xeon 3.1GHz), the time to apply buckets when running the command
`stellar-core catchup --in-memory 36939003/0`
goes from 1 minute to 34 seconds with this PR.

List of changes:
* Adjusted hash function used with ledger entries to reduce chance of rehashing
* Do not use an intermediate `LedgerTxn` when using in memory mode
* Guesstimate the size of dictionaries ahead of time to reduce chance of rehashing
    * this one only yields a small (5 seconds) improvement, the main benefit is that it removes pauses caused by rehashing while applying buckets

### DIFF
--- a/src/bucket/BucketApplicator.cpp
+++ b/src/bucket/BucketApplicator.cpp
@@ -88,6 +88,7 @@ BucketApplicator::advance(BucketApplicator::Counters& counters)
     {
         innerLtx = std::make_unique<LedgerTxn>(root, false);
         ltx = innerLtx.get();
+        ltx->prepareNewObjects(LEDGER_ENTRY_BATCH_COMMIT_SIZE);
     }
 
     for (; mBucketIter; ++mBucketIter)
--- a/src/bucket/BucketApplicator.cpp
+++ b/src/bucket/BucketApplicator.cpp
@@ -73,7 +73,23 @@ BucketApplicator::advance(BucketApplicator::Counters& counters)
     size_t count = 0;
 
     auto& root = mApp.getLedgerTxnRoot();
-    LedgerTxn ltx(root, false);
+    AbstractLedgerTxn* ltx;
+    std::unique_ptr<LedgerTxn> innerLtx;
+
+    // when running in memory mode, make changes to the in memory ledger
+    // directly instead of creating a temporary inner LedgerTxn
+    // as "advance" commits changes during each step this does not introduce any
+    // new failure mode
+    if (mApp.getConfig().MODE_USES_IN_MEMORY_LEDGER)
+    {
+        ltx = static_cast<AbstractLedgerTxn*>(&root);
+    }
+    else
+    {
+        innerLtx = std::make_unique<LedgerTxn>(root, false);
+        ltx = innerLtx.get();
+    }
+
     for (; mBucketIter; ++mBucketIter)
     {
         BucketEntry const& e = *mBucketIter;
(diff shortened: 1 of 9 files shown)


src/catchup/ApplyBucketsWork.cpp
```diff
@@ -119,6 +119,13 @@ ApplyBucketsWork::onReset()
             addBucket(getBucket(hsb.snap));
             addBucket(getBucket(hsb.curr));
         }
+        // estimate the number of ledger entries contained in those buckets
+        // use accounts as a rought approximator as to overestimate a bit
```
jonjove, line 123: Nitpick: typo "rought" -> "rough"

src/catchup/ApplyBucketsWork.cpp
```diff
@@ -119,6 +119,13 @@ ApplyBucketsWork::onReset()
...
+        size_t constexpr estimatedLedgerEntrySize = sizeof(AccountEntry);
```
jonjove, line 124: This is going to be a very bad estimate because it won't account for the dynamic size of `AccountEntry`. If you want something a bit more reasonable, you might try constructing a typical `AccountEntry ae` and use `xdr_traits<AccountEntry>::serial_size(ae)`.

src/ledger/LedgerTxn.h
```diff
@@ -458,6 +458,9 @@ class AbstractLedgerTxnParent
     // than a (real or stub) root LedgerTxn.
     virtual uint32_t prefetch(UnorderedSet<LedgerKey> const& keys) = 0;
 
+    // prepares to increase the capacity of pending changes up to "s" changes
+    virtual void reserveObjects(size_t s) = 0;
```
jonjove, line 462: I'm not 100% sure about this interface. `LedgerTxn` doesn't provide a convenient way to know how much space is already used, so probably it would be better to have `reserveAdditionalObjects` instead.

src/catchup/ApplyBucketsWork.cpp
MonsieurNicolas: fixed

src/catchup/ApplyBucketsWork.cpp
MonsieurNicolas: This estimate (using a default constructed `AccountEntry`) is "good enough" for this purpose:
* we want to over-estimate the number of ledger entries, so under-estimating the size of a `BucketEntry` as the size of a default constructed `BucketEntry` is good enough as a default constructed `AccountEntry` already contains things like `AccountID` (fixed array).
* I agree that using `xdr_traits<>::serial_size()` is better as it doesn't make any assumption on how xvectors are stored in memory for example.

src/ledger/LedgerTxn.h
MonsieurNicolas: I agree. In my branch I had a get/set but when I reduced it, I kept only the "set" which should be made relative

MonsieurNicolas:
@jonjove updated

jonjove:
r+ 2621441

src/bucket/BucketApplicator.cpp
```diff
@@ -73,7 +73,24 @@ BucketApplicator::advance(BucketApplicator::Counters& counters)
     size_t count = 0;
 
     auto& root = mApp.getLedgerTxnRoot();
```
Diego77777777777, line 75: 75     auto& root = mApp.getLedgerTxnRoot{start):

src/bucket/BucketApplicator.cpp
```diff
@@ -73,7 +73,24 @@ BucketApplicator::advance(BucketApplicator::Counters& counters)
...
-    LedgerTxn ltx(root, false);
```
Diego77777777777, line 76: Fake and rial

src/bucket/BucketApplicator.cpp
Diego77777777777, line 75: 75     auto& root = mApp.getLedgerTxnRoot(mp3),

Diego77777777777 (approved):

PR #3183, opened 2021-08-26
Do not call bucket manager shutdown twice

marta-lokhova:


MonsieurNicolas:
Why not just remove the error message in `BucketManagerImpl::shutdown()` instead? That function doesn't do anything other than set a boolean.

marta-lokhova:
I did it this way to make it less error prone, as we generally don't expect to call it multiple times, so the log might help reveal a bug. I don't feel strongly which way we do it though, I just want to make sure we don't confuse the operators, so I updated the PR to just remove the log completely.

MonsieurNicolas:
r+ 13b991b6e617378eb8898eb6609a727e441b9f6c

PR #3190, opened 2021-09-01
Adjust automatic maintenance to run faster (every 6 minutes)

MonsieurNicolas:
this avoids putting too much pressure on slower disks


graydon:
r+ bbc2d3f

PR #3191, opened 2021-09-01
Add ARTIFICIAL_DATABASE_LATENCY_FOR_TESTING config param (BUILD_TESTS only)

graydon:
This is just a test utility to let us crudely simulate the effects of arbitrarily-slow databases without having to manually construct them (eg. spinning disks, slow network drives, etc.)

Compiled-out of release builds, just for testing.

### DIFF
--- a/src/database/Database.cpp
+++ b/src/database/Database.cpp
@@ -336,6 +336,16 @@ Database::getUpsertTimer(std::string const& entityName)
 }
 
 void
+Database::maybeArtificiallySleepForSimulatedLatency()
+{
+    auto usec = mApp.getConfig().ARTIFICIAL_DATABASE_LATENCY_FOR_TESTING;
+    if (usec.count() != 0)
+    {
+        std::this_thread::sleep_for(usec);
+    }
+}
+
+void
 Database::setCurrentTransactionReadOnly()
 {
     if (!isSqlite())
@@ -417,6 +427,7 @@ Database::getSession()
 {
     // global session can only be used from the main thread
     assertThreadIsMain();
+    maybeArtificiallySleepForSimulatedLatency();
     return mSession;
 }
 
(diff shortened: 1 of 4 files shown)


src/database/Database.cpp
```diff
@@ -335,6 +335,18 @@ Database::getUpsertTimer(std::string const& entityName)
         .TimeScope();
 }
 
+void
+Database::maybeArtificiallySleepForSimulatedLatency()
+{
+#ifdef BUILD_TESTS
```
MonsieurNicolas, line 341: I think we can have this enabled in all builds, we just need to add `ARTIFICIAL_DATABASE_LATENCY_FOR_TESTING` to `TESTING_ONLY_OPTIONS` in config.cpp

src/database/Database.cpp
```diff
@@ -335,6 +335,18 @@ Database::getUpsertTimer(std::string const& entityName)
...
+    auto usec = mApp.getConfig().ARTIFICIAL_DATABASE_LATENCY_FOR_TESTING;
+    if (usec.count() != 0)
+    {
+        std::this_thread::sleep_for(usec);
```
MonsieurNicolas, line 345: Is this "good enough" (like did you actually look at something like Tracy plots for database access)?
If the goal is to simulate slow disks, the latency hit is on a per iops basis, not on a per query basis (and this depends on a lot of things).
For example, I would imagine that there is some function of log(table size):database cache size ratio. Also, read vs write performance could matter (especially when there are indexes that need to be updated as 1 sql op -> many iops).

MonsieurNicolas (changes requested):

src/database/Database.cpp
graydon: Ah, ok. Fixed.

src/database/Database.cpp
graydon: I have not yet examined tracy traces for comparison, will do so.

In terms of expectations: I believe there are two latency hits:
  - Round trip through socket to database process
  - Time to perform 1 fsync on database WAL

Granted the size of data to fsync will differ depending on the query, but I do not think it will differ by a lot -- I think the cost to push 1k and 1mb through to the disk is fairly similar (especially on spinning disk).

I believe a crude sleep like this therefore over-approximates the slowdown (as it penalizes reads as well as writes) but for purposes of simulation I think it's plausible.

MonsieurNicolas:
closing: stale and unlikely to land anytime soon

PR #3196, opened 2021-09-04
Bump tracy to take change that disables building with -flto

graydon:
Update tracy to not use -flto. Going to self-approve because it's 8pm on a friday and this is a strictly "build fixing" change I'd be amazed if anyone dislikes. Will revert if there's any disagreement later.


graydon:
r+ 4aea0c0

PR #3200, opened 2021-09-10
Update docker image build timestamp format

jacekn:
Golang time parser is not compatible with format rendered by the GNU date,
even though both claim to be RFC3339. This change replaces space separator
between date and time with "T" which will allow golang to parse it.
It also switches from `--utc` to `-u` that's supported by both GNU and BSD
date commands.


tprobinson (approved):
LGTM

tprobinson (approved):
LGTM -- agh I keep forgetting that my approvals only work if I'm authorized to write to the repo :laughing:

MonsieurNicolas:
r+ 4e3fc8bbc78629949a2a9866720399466e6d292c

PR #3205, opened 2021-09-11
Uint128_t native

graydon:
This is an experiment in using compiler-native uint128_t types when they are available (on gcc and clang), as a possible performance improvement over the library type. Need to discuss a bit, but it seems to work.

### DIFF
--- a/src/herder/TransactionQueue.cpp
+++ b/src/herder/TransactionQueue.cpp
@@ -86,13 +86,14 @@ canReplaceByFee(TransactionFrameBasePtr tx, TransactionFrameBasePtr oldTx,
     //
     // FEE_MULTIPLIER * v2 does not overflow uint128_t because fees are bounded
     // by INT64_MAX, while number of operations and FEE_MULTIPLIER are small.
-    auto v1 = bigMultiply(newFee, oldNumOps);
-    auto v2 = bigMultiply(oldFee, newNumOps);
-    auto minFeeN = v2 * TransactionQueue::FEE_MULTIPLIER;
+    uint128_t v1 = bigMultiply(newFee, oldNumOps);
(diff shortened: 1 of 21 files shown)


graydon:
Updated to fix bug in `uint128_bits`

src/herder/TransactionQueue.cpp
```diff
@@ -710,10 +710,11 @@ TransactionQueue::getMaxOpsToFloodThisPeriod() const
     int64_t opsToFloodLedger = static_cast<int64_t>(opsToFloodLedgerDbl);
 
     int64_t opsToFlood;
-    opsToFlood = mBroadcastOpCarryover +
-                 bigDivide(opsToFloodLedger, cfg.FLOOD_TX_PERIOD_MS,
-                           cfg.getExpectedLedgerCloseTime().count() * 1000,
-                           Rounding::ROUND_UP);
+    opsToFlood =
+        mBroadcastOpCarryover +
+        bigDivide(opsToFloodLedger, int64_t(cfg.FLOOD_TX_PERIOD_MS),
```
MonsieurNicolas, line 715: I think it would be safer to just avoid function overloading (as it's way too easy to forget to cast like this), we can just rename the version that takes an uint128 into `bigDivide128` or something like that

graydon: fixed at some length -- wound up renaming all the overloads in numeric.h because I couldn't tell in a lot of cases which overload was even being called in which order. Easiest fix was to split them all apart and just check (and possibly rewrite) each callsite to call the version that corresponds to what I think was meant by the operands passed and surrounding code. This should be carefully reviewed, especially by @jonjove !

MonsieurNicolas:
I had to stop this while getting 18.0.3 out the door, rekicking

MonsieurNicolas:
@latobarita: retry

PR #3207, opened 2021-09-13
More CAP-38 path payment tests

jonjove:
# Description
Built on top of #3198, so that should be merged first. New code starts from "CAP-38: Fix bug with PATH_PAYMENT_STRICT_*_CROSS_SELF".

# Checklist
- [x] Compiles
- [x] Ran all tests

### DIFF
--- a/src/transactions/ManageOfferOpFrameBase.cpp
+++ b/src/transactions/ManageOfferOpFrameBase.cpp
@@ -345,17 +345,17 @@ ManageOfferOpFrameBase::doApply(AbstractLedgerTxn& ltxOuter)
                 if ((passive && (o.price >= maxWheatPrice)) ||
                     (o.price > maxWheatPrice))
                 {
-                    return OfferFilterResult::eStop;
+                    return OfferFilterResult::eStopBadPrice;
                 }
                 if (o.sellerID == getSourceID())
                 {
                     // we are crossing our own offer
-                    setResultCrossSelf();
-                    return OfferFilterResult::eStop;
+                    return OfferFilterResult::eStopCrossSelf;
                 }
                 return OfferFilterResult::eKeep;
             },
             offerTrail, maxOffersToCross);
+
         releaseAssertOrThrow(sheepSent >= 0);
 
         bool sheepStays;
@@ -367,13 +367,12 @@ ManageOfferOpFrameBase::doApply(AbstractLedgerTxn& ltxOuter)
         case ConvertResult::ePartial:
             sheepStays = true;
             break;
-        case ConvertResult::eFilterStop:
-            if (!isResultSuccess())
-            {
-                return false;
-            }
+        case ConvertResult::eFilterStopBadPrice:
             sheepStays = true;
             break;
+        case ConvertResult::eFilterStopCrossSelf:
+            setResultCrossSelf();
+            return false;
         case ConvertResult::eCrossedTooMany:
             mResult.code(opEXCEEDED_WORK_LIMIT);
             return false;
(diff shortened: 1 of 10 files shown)


src/transactions/PathPaymentOpFrameBase.cpp
```diff
@@ -81,13 +81,17 @@ PathPaymentOpFrameBase::convert(AbstractLedgerTxn& ltx,
...
             {
                 // we are crossing our own offer
-                setResultOfferCrossSelf();
                 return OfferFilterResult::eStop;
             }
             return OfferFilterResult::eKeep;
         },
         offerTrail, maxOffersToCross);
 
+    if (r == ConvertResult::eFilterStop)
```
MonsieurNicolas, line 90: should this just be moved to the switch below (line 103 in this commit)?

MonsieurNicolas (changes requested):

src/transactions/PathPaymentOpFrameBase.cpp
jonjove: I considered that but it's _technically_ a protocol change to do that. I'm comfortable with it if you are, so let me know.

MonsieurNicolas: I don't understand why this would be a protocol change: the only difference in this function is that in case of exception below the result would not be set - but in that case, we end up with an internal error anyways.

src/transactions/ManageOfferOpFrameBase.cpp
```diff
@@ -367,7 +372,8 @@ ManageOfferOpFrameBase::doApply(AbstractLedgerTxn& ltxOuter)
         case ConvertResult::ePartial:
             sheepStays = true;
             break;
-        case ConvertResult::eFilterStop:
+        case ConvertResult::eFilterStopBadPrice:
+        case ConvertResult::eFilterStopCrossSelf:
```
MonsieurNicolas, line 376: You should handle those two cases separately and remove the `if (!isResultSuccess())` (this was used to distinguish between `eFilterStopBadPrice` and `eFilterStopCrossSelf`), that way you can also move `setResultCrossSelf` into the `eFilterStopCrossSelf` case

src/transactions/PathPaymentOpFrameBase.cpp
```diff
@@ -99,7 +99,8 @@ PathPaymentOpFrameBase::convert(AbstractLedgerTxn& ltx,
 
     switch (r)
     {
-    case ConvertResult::eFilterStop:
+    case ConvertResult::eFilterStopBadPrice:
```
MonsieurNicolas, line 102: why handle `eFilterStopBadPrice` like this? I think it should just hit the `default` and crash as this should never happen

MonsieurNicolas (changes requested):
couple questions there

src/transactions/PathPaymentOpFrameBase.cpp
jonjove: Oh yes, you're right. I'll make this change.

MonsieurNicolas:
r+ ecfdeb1081742ee2b6b7bd8463b213f22c3be691

src/transactions/test/LiquidityPoolTradeTests.cpp
```diff
@@ -21,6 +21,957 @@
...
+                                 int64_t amountB, Price const& minPrice,
+                                 Price const& maxPrice) {
+        if (!(cur1 < cur2))
+        {
+            std::swap(amountA, amountB);
+        }
+        acc.liquidityPoolDeposit(pool12, amountA, amountB, minPrice, maxPrice);
+    };
+
+    SECTION("chooses best price")
```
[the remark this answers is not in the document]
jonjove: Done. It wasn't super easy because there are several things to handle, but I really like that it applies to all our tests globally.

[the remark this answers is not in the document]
jonjove: (To be clear, done in my new work. Not done in this PR.)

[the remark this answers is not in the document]
jonjove: There were some edge cases I couldn't sort out easily. It almost works, but I'll have to revisit it later.

PR #3215, opened 2021-09-17
OfferExchange: fast fail when hitting limits

MonsieurNicolas:
Improve error code to be easier to understand when limits have been reached

### DIFF
--- a/src/transactions/OfferExchange.cpp
+++ b/src/transactions/OfferExchange.cpp
@@ -1398,6 +1398,8 @@ exchangeWithPool(AbstractLedgerTxn& ltxOuter, Asset const& toPoolAsset,
     {
         // offerTrail is going to be too long after exchanging with the
         // liquidity pool
+        // note that this condition can only happen in path payment when
+        // performing subsequent hops
         return false;
     }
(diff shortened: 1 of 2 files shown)


src/transactions/OfferExchange.cpp
```diff
@@ -1491,6 +1491,13 @@ convertWithOffers(
     wheatReceived = 0;
 
     bool needMore = (maxWheatReceive > 0 && maxSheepSend > 0);
+    if (needMore && maxOffersToCross == 0 &&
+        ltxOuter.loadHeader().current().ledgerVersion >= 18)
+    {
+        // offerTrail is going to be too long, fast fail
```
jonjove, line 1497: Can you extend this comment to indicate that this can happen for path payment, but cannot happen for manage offer? I think this is important because if this _could_ happen for manage offer, then it would be incorrect.

MonsieurNicolas: good idea, done. I also added the same comment in the pool exchange function

jonjove:
r+ df5efb9

PR #3223, opened 2021-09-23
remove obsolete tx flooding code

MonsieurNicolas:
Delayed flooding code has been the default for a while. There is no need to maintain the legacy synchronous code at this point.


marta-lokhova:
r+ 5b2483d8bc14b8de153cf0906cb91893c9c321aa

PR #3227, opened 2021-09-28
Ensure docker builds don't use cached layers

jacekn:
In some circumstances docker will use cached layers during buidls. This can
cause problems as we depend on the external repository that may change at any
time. This change will disable layer cache.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -20,8 +20,7 @@ ARG STELLAR_CORE_VERSION
 #install stellar-core
 RUN wget -qO - https://apt.stellar.org/SDF.asc | apt-key add -
 RUN echo "deb https://apt.stellar.org ${DISTRO} unstable" | tee -a /etc/apt/sources.list.d/SDF-unstable.list
-RUN apt-get update
-RUN apt-get install -y stellar-core=${STELLAR_CORE_VERSION}
+RUN apt-get update && apt-get install -y stellar-core=${STELLAR_CORE_VERSION}
 
 WORKDIR "/etc/stellar"
 ENTRYPOINT ["/usr/bin/stellar-core"]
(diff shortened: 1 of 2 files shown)


leighmcculloch:
>we depend on the external repository that may change at any time

Could you go into more detail on this? It's unclear why the build cache is a problem. If we have reproducible builds the cache should be okay. If the cache is not okay it points to some other more fundamental issue I think.

jacekn:
> > we depend on the external repository that may change at any time
> 
> Could you go into more detail on this? It's unclear why the build cache is a problem. If we have reproducible builds the cache should be okay. If the cache is not okay it points to some other more fundamental issue I think.

In this particular case we use stellar public apt repository to install packages during the build. If we use cached layers then `apt-get update` is not executed and that might cause the build to use out of date repo metadata. If the metadata is stale apt might be unable to install recently published packages causing `docker build` failures.

leighmcculloch:
Should the build take a stellar-core version as an input that is fed into the apt install process, so it is reproducible and is building a specific version of core? That is what we do in some other builds that install core and horizon from apt.

jacekn:
> Should the build take a stellar-core version as an input that is fed into the apt install process, so it is reproducible and is building a specific version of core? That is what we do in some other builds that install core and horizon from apt.

We do that already but the `apt-get update` is a separate step from the install. But you have a good point, we could combine the two and that should ensure we don't cache that layer unless we try with the same package version.
I'll redo the PR

leighmcculloch:
> `apt-get update` is a separate step from the install. But you have a good point, we could combine the two

💯

MonsieurNicolas:
r+ 09229ae0edb43c74d94e045fdf20361bdef08c80

PR #3228, opened 2021-09-28
overlay: message processing metrics

marta-lokhova:
Initial implementation of the message processing metrics in overlay. The main two metrics are: how long each message type sits in the scheduler queue before getting dispatched, and how long it takes to completely process certain message types. "completely process" here includes actions like writing to the network, fetching values etc. This should give us a better picture of the consensus round impact.

This is a preliminary implementation, and as it touches a lot of areas in the codebase, it still requires more testing. There's also a TODO to ensure proper shutdown of TCPPeer given that we now store callbacks that update metrics once messages are popped from the write queue. But I wanted to push this change anyway so we can start reviewing it.

### DIFF
--- a/src/herder/Herder.h
+++ b/src/herder/Herder.h
@@ -124,13 +124,16 @@ class Herder
     virtual SCPQuorumSetPtr getQSet(Hash const& qSetHash) = 0;
 
     // We are learning about a new envelope.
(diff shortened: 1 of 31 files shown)


marta-lokhova:
this functionality has been incorporated into https://github.com/stellar/stellar-core/pull/3276

Issue #3221, opened 2021-09-21
Metrics: audit and remove obsolete or broken metrics


MonsieurNicolas:
According to some people, we are reporting 500 metrics in the `metrics` endpoint.

Some that I can think of that are basically useless:
* catchup related metrics
* Granular database reporting (this should be done with Tracy) - see #2222
* redundant things like "app state" vs "stellar core state"

MonsieurNicolas:
I wonder if what we need to do is something similar to "partitions" used for logging:
* define a set of "default" metrics that should be exported
* define alternate presets that are sets of metrics that can be enabled on demand

That way by default we would not have all metrics enabled (which avoids overloading systems that aggregate metrics), while at the same time we can enable a different subset of metrics when performing certain type of simulations (or we're just debugging something).

I could see the interface look something like `metrics?{enable=subsetname}&{disable=subsetname}`

So for debugging overlay and consensus related we could do something like this in config:

```
COMMANDS=[
"metrics?disable=default",
"metrics?enable=allscp",
"metrics?enable=alloverlay"
]
```

MonsieurNicolas:
I just realized that we can potentially also enable/disable individual metrics with the same interface: we would just pass a metric name instead of a subset name. For example `metrics?enable=scp.timing.self-to-others-externalize-lag`

Issue #2222, opened 2019-08-13
Bogus database related metrics reported in `metrics` endpoint


MonsieurNicolas:
v11.3.0 Linux, running Postgresql

For some reason, many database related metrics show up with values of "0".

Glancing at the code, this doesn't seem to make sense as those metrics should only be queried/updated when running with sqlite (this is accessing the `metrics` endpoint directly).

Here is the list:
```
"database.insert.trustline"
"database.insert.trust"
"database.insert.state"
"database.insert.peer"
"database.insert.offer"
"database.insert.ledger-header-history"
"database.insert.data"
"database.insert.ban"
"database.insert.account"
"database.delete.txhistory"
"database.delete.txfeehistory"
"database.delete.trust"
"database.delete.state"
"database.delete.scpquorums"
"database.delete.quoruminfo"
"database.delete.ledger-header-history"
"database.delete.ledger-header"
"database.delete.ban"
"app.post-on-main-thread-with-delay.delay"
"database.select.publishqueue"
"database.select.trustline"
"database.select.txfeehistory"
"database.update.account"
"database.update.ban"
"database.update.data"
"database.update.ledger-header"
"database.update.ledger-header-history"
"database.update.offer"
"database.update.publishqueue"
"database.update.quoruminfo"
"database.update.scphistory"
"database.update.scpquorums"
"database.update.trust"
"database.update.trustline"
"database.update.txfeehistory"
"database.update.txhistory"
```

MonsieurNicolas:
I think we should remove database related metrics that are that granular: the right way to track down performance issues with SQL is using SQL profiling, not with client side code (there are many tools that generate reports from SQL trace logs, for example https://pgbadger.darold.net/ )

graydon:
These are zero valued because we moved to mostly upserts with the bulk query work last year.

I would like to keep these metrics. End-to-end latency on a database query _from the client's perspective_ is important and we spent a lot of time tuning it, will likely have to go back to tuning it again someday. We bought enough performance room not to look at it in recent months, but that's not permanent.

MonsieurNicolas:
Yeah for this, I suspect what we want is not metrics though but high quality, low overhead, trace logging (a problem we have for other subsystems).

MonsieurNicolas:
Another problem with those bulk sql metrics is that we don't know how many entries are involved, so samples are quite misleading.
For example, recently we have some ledgers that load many claimable balances, so the `select_claimablebalance` metrics shot up.
[...]

PR #3230, opened 2021-10-01
Metric cleanup

marta-lokhova:
Cleanup unused metrics as per #3221. Most of these metrics fall into two categories:
* catchup metrics. we do not use these, as we usually report such events via logs. The only place where these were actually useful was in tests, so I wired up some accounting to CatchupManager to accomplish the same goal (I thought about simplifying tests and maybe not having them rely on metrics at all, but ultimately chose to keep it because in the past I found these checks very useful to catch bugs while working on catchup changes)
* granular database metrics. I kept the database metrics for ledger entries, tx history and overlay peers, and removed metrics for other tables. Please let me know if you would like to keep some of these


graydon (changes requested):
I'm still somewhat uncertain about the database changes, leaning towards not wanting them. I understand these are infrequent timers relative to the main ledger types, but some of them (especially peer-related and publishing-related) have shown up in profiles as significant stalls on the IO loop, and I'd somewhat prefer not to hide them from our attention. I understand they represent operational load on prometheus and noise on the dashboards, so I'm willing to consider alternatives .. just feel uneasy dropping them entirely.

In https://github.com/stellar/stellar-core/issues/2222 @MonsieurNicolas mentioned he'd be willing to split the difference with my preference and see these replaced with detailed tracing (I guess tracy zones). I notice there aren't any tracy annotations in the src/database directory at all currently -- what would you think about adding zones there so we can at least get a picture of the distribution of DB operation latencies (tagged with tables / action-types as userdata) in tracy?

The rest looks good!

marta-lokhova:
@graydon I went back and replaced the database timers with named tracy zones, so we should be able to see the same data in Tracy. what do you think?

PR #3239, opened 2021-10-09
fix: captive core crashes with all invariants enabled

MonsieurNicolas:
disable bucketlist check invariant that relies on SQL


marta-lokhova:
r+ 14a86866267fb314a0d66eff2e02b127cf5d2775

PR #3240, opened 2021-10-12
Windows: uint128 and bucket merge fix

MonsieurNicolas:
This PR makes a few Windows specific tweaks that help with performance investigation:
* it replaces the uint128 class by a class that can properly be inlined
* it adds Visual C++ intrinsic support to uint128
* it uses C style FILE handles to write buckets.
 * The `shouldUseRandomAccessHandle` path was a Windows only path anyways, so I just ended up removing support for that path on Linux/MacOS


marta-lokhova:
r+ dc55c1f3d2da2b90646f729011fc275a3f5e03e8

PR #3244, opened 2021-10-14
Some more perf tweaks

MonsieurNicolas:
I ended up salvaging a commit from the other PR.

This PR yields another 20% on top of master.

### DIFF
--- a/src/ledger/InternalLedgerEntry.cpp
+++ b/src/ledger/InternalLedgerEntry.cpp
@@ -3,6 +3,7 @@
 // of this distribution or at http://www.apache.org/licenses/LICENSE-2.0
 
 #include "ledger/InternalLedgerEntry.h"
+#include "ledger/LedgerHashUtils.h"
 #include "util/GlobalChecks.h"
 #include "util/XDRCereal.h"
 #include "util/types.h"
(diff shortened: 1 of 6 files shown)


src/ledger/InternalLedgerEntry.h
```diff
@@ -46,7 +46,16 @@ class InternalLedgerKey
 
     void checkDiscriminant(InternalLedgerEntryType expected) const;
 
+    void type(InternalLedgerEntryType t);
+    LedgerKey& ledgerKeyRef();
```
jonjove, line 50: These are only ever used for assignment. Should we change them to setters, for example
```c++
void setLedgerKey(LedgerKey const& key);
void setLedgerKey(LedgerKey&& key);
```
This would make it even harder for a future maintainer to accidentally invalidate the hash by using a reference improperly.

src/ledger/InternalLedgerEntry.cpp
```diff
@@ -131,10 +132,39 @@ InternalLedgerKey::~InternalLedgerKey()
...
 
+size_t
+InternalLedgerKey::hash() const
+{
+    if (mHash != 0)
+    {
+        return mHash;
+    }
+    size_t res;
+    switch (type())
```
jonjove, line 159: I know this is the hash function that we were already using, but it's not very good. It collides sponsorships and sponsorship counters.

src/ledger/LedgerTxn.cpp
```diff
@@ -1946,10 +1969,15 @@ LedgerTxn::Impl::updateEntry(InternalLedgerKey const& key,
...
 
-    auto iter = mEntry.find(key);
-    if (iter != mEntry.end() && iter->second)
+    EntryMap::iterator localIter;
+    if (!keyHint)
+    {
+        localIter = mEntry.find(key);
+        keyHint = &localIter;
+    }
+    if (keyHint && *keyHint != mEntry.end() && (*keyHint)->second)
```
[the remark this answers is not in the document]
MonsieurNicolas: fixed

src/ledger/LedgerTxn.cpp
```diff
@@ -1946,10 +1969,15 @@ LedgerTxn::Impl::updateEntry(InternalLedgerKey const& key,
         return;
     }
 
-    auto iter = mEntry.find(key);
-    if (iter != mEntry.end() && iter->second)
+    EntryMap::iterator localIter;
```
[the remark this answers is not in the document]
MonsieurNicolas: > I think you could simplify this function if you moved this to the top of this function (specifically, you wouldn't need to keep checking if keyHint is non-null).

not quite: in the "insert" case (not update/not offer), we end up calling `mEntry[key] = lePtr;` (that performs the entry lookup) and moving this code to the top would trigger an additional lookup 

For `localIter`: I renamed it and added a comment

src/ledger/InternalLedgerEntry.h
jonjove: That's fine for now. If we ever change the way those methods are used (unlikely) we can change to a safer idiom.

PR #3246, opened 2021-10-14
log when meta stream writes are slow

MonsieurNicolas:
I picked 100 ms for DEBUG logging -> 1 second WARNING level as 1 second will start to really interfere with core's operation


graydon:
r+ 45ff712

PR #3256, opened 2021-10-26
Log as much uncaught-exception info as we can, fixes #3155.

### LINKED ISSUE #3155: we should log exception information when hitting std::terminate
Right now we set a handler with `std::set_terminate` that doesn't actually report anything about the exception that caused the problem.

We should instead follow the pattern described in https://en.cppreference.com/w/cpp/error/current_exception that allows to rebuild the exception state that caused the problem.

Log should contain information on the exception and `REPORT_INTERNAL_BUG` before proceeding with dumping the stack trace.

We should also double check that exception information is indeed still available in a core dump after hitting a crash handled that way (I don't know if rethrowing the exception in the handler causes problems or not).

graydon:
This just adds some additional reporting in the case where we call `std::terminate` with an uncaught exception. Especially bad when backtraces aren't working (as they don't on many configs).

I also added a change that makes the libunwind-based backtracing machinery turn on a little more automatically when we are able to use it. But I re-tested various configs and it is still only able to work reliably on gcc-only builds (or possibly clangs that you build yourself with a specific choice of libunwind, but not stock / platform clangs).

### DIFF
--- a/src/main/main.cpp
+++ b/src/main/main.cpp
@@ -2,22 +2,134 @@
 // under the Apache License, Version 2.0. See the COPYING file at the root
 // of this distribution or at http://www.apache.org/licenses/LICENSE-2.0
 
+#include "crypto/CryptoError.h"
+#include "invariant/InvariantDoesNotHold.h"
+#include "ledger/NonSociRelatedException.h"
 #include "main/CommandLine.h"
 #include "util/Backtrace.h"
+#include "util/FileSystemException.h"
 #include "util/Logging.h"
 
 #include "crypto/ShortHash.h"
 #include "util/RandHasher.h"
 #include <cstdlib>
 #include <exception>
+#include <filesystem>
 #include <sodium/core.h>
+#include <system_error>
 #include <xdrpp/marshal.h>
 
 namespace stellar
 {
 static void
+printCurrentException()
+{
+    std::exception_ptr eptr = std::current_exception();
+    if (eptr)
+    {
+        try
+        {
+            std::rethrow_exception(eptr);
+        }
+        catch (NonSociRelatedException const& e)
+        {
+            fprintf(stderr,
+                    "current exception: NonSociRelatedException(\"%s\")\n",
+                    e.what());
+        }
+        catch (CryptoError const& e)
+        {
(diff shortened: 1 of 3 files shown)


src/main/main.cpp
```diff
@@ -2,22 +2,134 @@
...
+printCurrentException()
+{
+    std::exception_ptr eptr = std::current_exception();
+    if (eptr)
+    {
+        try
+        {
+            std::rethrow_exception(eptr);
+        }
+        catch (const NonSociRelatedException& e)
```
MonsieurNicolas, line 34: `X const&` not `const X&`

src/main/main.cpp
```diff
@@ -2,22 +2,134 @@
...
+        {
+            fprintf(stderr, "current exception: std::overflow_error(\"%s\")\n",
+                    e.what());
+        }
+        catch (const std::underflow_error& e)
+        {
+            fprintf(stderr, "current exception: std::underflow_error(\"%s\")\n",
+                    e.what());
+        }
+        catch (const std::exception& e)
```
MonsieurNicolas, line 116: we should remove the `try/catch` https://github.com/stellar/stellar-core/blob/d55d2bc98433c3440e3c5731049ecf31a1768d58/src/main/ApplicationUtils.cpp#L206-L210 as this ends up being redundant with this code (and by removing the `throw;` we actually capture the stack trace)

MonsieurNicolas (changes requested):

src/util/Backtrace.cpp
```diff
@@ -3,6 +3,7 @@
 // of this distribution or at http://www.apache.org/licenses/LICENSE-2.0
 
 #include "util/Backtrace.h"
+#include "config.h"
```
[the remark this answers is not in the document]
graydon: No idea how that wound up in there. Possibly an ambitious editor inclusion. Removed.

src/main/main.cpp
graydon: fixed

src/main/main.cpp
graydon: fixed

MonsieurNicolas:
r+ 496593374064ad6190443b04f5b7bbc6361e6061

PR #3260, opened 2021-10-29
Update the logic in `LogSlowExecution` to correctly count dropped messages

### LINKED ISSUE #3241: throttled logging is not working correctly
I noticed that we have messages in our logs like
```
2021-10-12T14:41:48.925 GBVQQ [Ledger INFO] Closed ledger: [seq=44, hash=1b7f77]                                                                                                                                                                                              
2021-10-12T14:41:56.849 GBVQQ [Perf WARNING] Dropped 1 slow-execution warning messages                                                                                                                                                                                        
2021-10-12T14:41:59.253 GBVQQ [Perf WARNING] Dropped 1 slow-execution warning messages                                                                                                                                                                                        
2021-10-12T14:41:59.746 GBVQQ [Herder INFO] Quorum information for 43 : {"agree":23,"cost":9371856,"delayed":0,"disagree":0,"fail_at":6,"hash":"e1a626","lag_ms":262,"ledger":43,"missing":0,"phase":"EXTERNALIZE","validated":true}
```

The thing is: there are no "dropped messages", just a message that messages got dropped.

The problem is that it throttles DEBUG and WARNING in aggregate, so it's very possible that DEBUG entries are "logged" (but no log line is emitted if log level is WARNING) and WARNING entries are dropped entirely because DEBUG entries took out all logging quota.

hidenori-shinohara:
# Description

Resolves #3241

`LogSlowExecution` counts both `DEBUG` and `WARNING` messages even when the log level doesn't allow `DEBUG` messages to be printed. This leads to an unexpected behavior.

Here's an example:

The log level is `INFO`.

1. `checkElapsedTime` is called. It has exceeded the threshold, but not by 10x.
1. 0.5 seconds later: `checkElapsedTime` is called for another `LogSlowExecution` object. It has exceeded the threshold by more than 10x.
1. 10 seconds later. `checkElapsedTime` is called again. It has exceeded the threshold, but not by 10x.

*Results*: The sequence of events above results in only one line of logs that just says `Dropped 1 slow-execution warning messages`. 1 and 3 are "logged" at the `DEBUG` level, and the log level in this example is `INFO` so users won't see them. 2 won't get logged since it's only half a second after 1.

*Expected*: We see Message 2. Message 1 and 3 should be ignored when calculating `gDroppedLogMessagesSinceLast` and `gLastLogMessage` since it wasn't going to be printed anyway due to the current logging level.

# Checklist
- [x] Compiles
- [x] Ran all tests

### DIFF
--- a/src/util/LogSlowExecution.cpp
+++ b/src/util/LogSlowExecution.cpp
@@ -10,8 +10,12 @@
 namespace
 {
 std::mutex gLogSlowExecMutex;
-std::chrono::system_clock::time_point gLastLogMessage;
-size_t gDroppedLogMessagesSinceLast{0};
+
+// For each log level, keep track of when the last message was and
+// the number of skipped messages.
+std::map<stellar::LogLevel,
+         std::pair<std::chrono::system_clock::time_point, size_t>>
+    gLastLogMessage;
 }
 
 namespace stellar
@@ -44,36 +48,48 @@ LogSlowExecution::checkElapsedTime() const
     {
         std::lock_guard<std::mutex> guard(gLogSlowExecMutex);
 
+        // Check whether we're 10 times over the threshold to decide the log
+        // level.
+        LogLevel ll = elapsed > mThreshold * 10 ? LogLevel::LVL_WARNING
+                                                : LogLevel::LVL_DEBUG;
+
+        auto& lastMsgTime = gLastLogMessage[ll].first;
+        auto& skippedCnt = gLastLogMessage[ll].second;
+
         // Only emit a new log message once-per-second (optionally preceded
         // by a summary of how many messages have been dropped since last time).
-        if ((finish - gLastLogMessage) > std::chrono::seconds(1))
+        if ((finish - lastMsgTime) > std::chrono::seconds(1))
         {
-            if (gDroppedLogMessagesSinceLast > 0)
+            std::string skipped;
+            if (skippedCnt > 0)
             {
-                CLOG_WARNING(Perf, "Dropped {} slow-execution warning messages",
-                             gDroppedLogMessagesSinceLast);
-                gDroppedLogMessagesSinceLast = 0;
+                skipped =
+                    fmt::format(FMT_STRING(" [skipped: {:d}]"), skippedCnt);
+                skippedCnt = 0;
             }
-            auto msg = fmt::format(FMT_STRING("'{}' {} {} s"), mName, mMessage,
-                                   static_cast<float>(elapsed.count()) / 1000);
-            // if we're 10 times over the threshold, log with INFO, otherwise
-            // use DEBUG
-            if (elapsed > mThreshold * 10)
+            auto msg = fmt::format(
+                FMT_STRING("'{:s}' {:s} {:f} s{:s}"), mName, mMessage,
+                static_cast<float>(elapsed.count()) / 1000, skipped);
+            switch (ll)
             {
+            case LogLevel::LVL_WARNING:
                 CLOG_WARNING(Perf, "{}", msg);
-            }
-            else
-            {
+                break;
+            case LogLevel::LVL_DEBUG:
                 CLOG_DEBUG(Perf, "{}", msg);
+                break;
+            default:
+                throw std::runtime_error(
+                    "Unexpected log level in LogSlowExecution");
             }
-            gLastLogMessage = finish;
+            lastMsgTime = finish;
         }
 
         // If we've emitted this second already, just bump a counter that
         // will be flushed in a summary message at the next second-boundary.
         else
         {
-            gDroppedLogMessagesSinceLast++;
+            skippedCnt++;
         }
     }
     return elapsed;


src/util/LogSlowExecution.cpp
```diff
@@ -44,36 +44,50 @@ LogSlowExecution::checkElapsedTime() const
...
-            // if we're 10 times over the threshold, log with INFO, otherwise
-            // use DEBUG
-            if (elapsed > mThreshold * 10)
+            // Only emit a new log message once-per-second (optionally preceded
+            // by a summary of how many messages have been dropped since last
+            // time).
+            if ((finish - gLastLogMessage) > std::chrono::seconds(1))
             {
-                CLOG_WARNING(Perf, "{}", msg);
+                if (gDroppedLogMessagesSinceLast > 0)
```
MonsieurNicolas, line 60: I think this implementation still suffers from the same issue than master: if a very slow execution is sandwiched between "regularly slow" ones, the actual entry will not be logged because things are logged with `CLOG_DEBUG`

hidenori-shinohara: In your example, is the user expected to _not_ see any messages about "regularly slow" ones (because of the logging level), but is the user expected to see a message about the very slow execution? If that's the case, I believe that this change addresses that.

https://github.com/stellar/stellar-core/blob/89a6bffcd94fc6f48eeb0848f6bc4d3f6328ba9e/src/util/LogSlowExecution.cpp#L53

If the current logging level prevents `CLOG_DEBUG` from actually being printed, then this method will not update `gLastLogMessage` or `gDroppedLogMessagesSinceLast`. `gLastLogMessage` is always the last time we actually printed something and the user saw it, not the time when we called `CLOG_*`. So, when a very slow execution (more than 10x) is sandwiched between slow ones (more than the threshold, less than 10x), then those slow ones won't make it past `if (ll <= Logging::getLogLevel("Perf"))`, and the very slow execution will get logged and the user will see it.

Please let me know if I'm misunderstanding something! @MonsieurNicolas

MonsieurNicolas: yeah this version is slightly better than master in that at `WARNING` level, it will only count logging events that correspond to `CLOG_WARNING`, *but* at `DEBUG` level can hide the `CLOG_WARNING` sandwiched like I wrote earlier

hidenori-shinohara: What would be the expected outcome when we have "regularly slow" -> "extremely slow" -> "regularly slow" in succession at the `DEBUG` level?

Would we want the user to see both the first "regularly slow" one _and_ the "extremely slow" one? I was under the impression that we'd only show the first "regularly slow" one.

Maybe what you're saying is that we should keep track of `gLastLogMessage` for `WARNING` (=extremely slow) and `DEBUG` (=regularly slow) separately? That way, we'd decide whether to print `WARNING` messages (=extremely slow) only based on when the last `WARNING` message was, and it wouldn't depend on any `DEBUG` messages. Is my understanding of this correct? @MonsieurNicolas

hidenori-shinohara: @MonsieurNicolas I updated the code so it keeps track of the last message for each logging level.

If there's a series of events "slow" -> "slow" -> "extremely slow" -> (2 seconds of nothing) -> "slow", then

- If the log level is `DEBUG`
    - The first "slow" is printed.
    - The second "slow" is ignored. (because we just printed a `DEBUG` message)
    - The "extremely slow" event is printed along with `dropped 1 message`
    - The last "slow" is printed.
- If the log level is `WARNING`
    - Only the "extremely slow" event is printed.
    - No message about dropping messages.

MonsieurNicolas: Would probably be better to just bundle `gLastLogMessage` and `gDroppedLogMessagesSinceLast` ? The map could be a map of `pair<std::chrono::system_clock::time_point, size_t>` ? That way the number of dropped messages is related to what is being logged (and we can fix the log level by logging a single line) and do something like:
```
            std::string skipped;
            if (droppedLogMessagesSinceLast > 0)
            {
                skipped = fmt::format(FMT_STRING(" [skipped: {}]"), droppedLogMessagesSinceLast);
                droppedLogMessagesSinceLast = 0;
            }
            auto msg = fmt::format(FMT_STRING("'{}' {} {} s{}"), mName, mMessage,
                                   static_cast<float>(elapsed.count()) / 1000, droppedLogMessagesSinceLast, skipped);
```

hidenori-shinohara: Makes sense! I updated the code.

src/util/LogSlowExecution.cpp
```diff
@@ -44,36 +45,45 @@ LogSlowExecution::checkElapsedTime() const
...
 
+        // Check whether we're 10 times over the threshold to decide the log
+        // level.
+        LogLevel ll = elapsed > mThreshold * 10 ? LogLevel::LVL_WARNING
+                                                : LogLevel::LVL_DEBUG;
+
         // Only emit a new log message once-per-second (optionally preceded
         // by a summary of how many messages have been dropped since last time).
-        if ((finish - gLastLogMessage) > std::chrono::seconds(1))
+        if ((finish - gLastLogMessage[ll].first) > std::chrono::seconds(1))
```
MonsieurNicolas, line 55: nit: just create a local ref to `gLastLogMessage[ll]` to avoid looking it up all the time/makes the code easier to follow

hidenori-shinohara: I added a ref for both elements in the pair for readability!

MonsieurNicolas:
r+ c7d980e9b7735c70f9533a3d4a27b5b9fe357508

PR #3272, opened 2021-11-12
properly convert some enum values when printing json

MonsieurNicolas:
# Description

Resolves issue introduced in #2703 : `xdr_to_string` properly handles this case that was missed when converting into json.


graydon:
r+ af51fdc

PR #3276, opened 2021-11-23
Overlay flow control

marta-lokhova:
Initial version of the overlay flow control change, wanted to get more eyes on it. Main changes are:
* Overlay now supports flow control via the new message type SEND_MORE
* Reading from the peers is throttled based on the available capacity
* Peers may opt-out of flow control via the config flag ENABLE_OVERLAY_FLOW_CONTROL (in which case they send SEND_MORE=0 during auth to notify the other peer that they don't want flow control)

Currently, the next things I'm working on are:
* Implementing emergency mechanisms to unblock the nodes in case there are bugs (deadlocks etc)
* Automatically calculating the "right" window size and pipeline parameters based on the connection quality
* Adding unit tests

### DIFF
--- a/src/herder/HerderImpl.cpp
+++ b/src/herder/HerderImpl.cpp
@@ -730,7 +730,8 @@ HerderImpl::sendSCPStateToPeer(uint32 ledgerSeq, Peer::pointer peer)
                 StellarMessage m;
                 m.type(SCP_MESSAGE);
                 m.envelope() = e;
-                peer->sendMessage(m, log);
+                auto mPtr = std::make_shared<StellarMessage const>(m);
+                peer->sendMessage(mPtr, log);
                 log = false;
                 slotHadData = true;
                 return true;
--- a/src/herder/HerderImpl.cpp
+++ b/src/herder/HerderImpl.cpp
@@ -1891,4 +1891,12 @@ HerderImpl::makeStellarValue(Hash const& txSetHash, uint64_t closeTime,
                                   sv.txSetHash, sv.closeTime));
     return sv;
 }
+
+bool
+HerderImpl::isNewerNominationOrBallotSt(SCPStatement const& oldSt,
+                                        SCPStatement const& newSt)
+{
+    return getSCP().isNewerNominationOrBallotSt(oldSt, newSt);
+}
+
 }
--- a/src/main/ApplicationImpl.cpp
+++ b/src/main/ApplicationImpl.cpp
@@ -763,13 +763,7 @@ ApplicationImpl::gracefulStop()
 void
 ApplicationImpl::shutdownMainIOContext()
 {
-    if (!mVirtualClock.getIOContext().stopped())
-    {
-        // Drain all events; things are shutting down.
-        while (mVirtualClock.cancelAllEvents())
-            ;
-        mVirtualClock.getIOContext().stop();
-    }
+    mVirtualClock.shutdown();
 }
 
 void
--- a/src/main/ApplicationImpl.cpp
+++ b/src/main/ApplicationImpl.cpp
@@ -1270,6 +1270,12 @@ ApplicationImpl::postOnMainThread(std::function<void()>&& f, std::string&& name,
     mVirtualClock.postAction(
         [this, f = std::move(f), isSlow]() {
             mPostOnMainThreadDelay.Update(isSlow.checkElapsedTime());
+            auto sleepFor =
+                this->getConfig().ARTIFICIALLY_SLEEP_MAIN_THREAD_FOR_TESTING;
+            if (sleepFor > std::chrono::microseconds::zero())
+            {
+                std::this_thread::sleep_for(sleepFor);
+            }
             f();
         },
         std::move(name), type);
--- a/src/overlay/OverlayManagerImpl.cpp
+++ b/src/overlay/OverlayManagerImpl.cpp
@@ -871,6 +871,12 @@ OverlayManagerImpl::isPreferred(Peer* peer) const
     return false;
 }
 
+bool
+OverlayManagerImpl::isFloodMessage(StellarMessage const& msg)
+{
+    return msg.type() == SCP_MESSAGE || msg.type() == TRANSACTION;
+}
+
 std::vector<Peer::pointer>
 OverlayManagerImpl::getRandomAuthenticatedPeers()
 {
@@ -1022,8 +1028,7 @@ OverlayManagerImpl::recordMessageMetric(StellarMessage const& stellarMsg,
     };
 
     bool flood = false;
-    if (stellarMsg.type() == TRANSACTION || stellarMsg.type() == SCP_MESSAGE ||
-        stellarMsg.type() == SURVEY_REQUEST ||
+    if (isFloodMessage(stellarMsg) || stellarMsg.type() == SURVEY_REQUEST ||
         stellarMsg.type() == SURVEY_RESPONSE)
     {
         flood = true;
--- a/src/overlay/Peer.cpp
+++ b/src/overlay/Peer.cpp
@@ -911,7 +911,6 @@ void
 Peer::recvGetSCPQuorumSet(StellarMessage const& msg)
 {
     ZoneScoped;
-    maybeProcessPingResponse(msg.qSetHash());
 
     SCPQuorumSetPtr qset = mApp.getHerder().getQSet(msg.qSetHash());
 
@@ -931,6 +930,7 @@ Peer::recvSCPQuorumSet(StellarMessage const& msg)
 {
     ZoneScoped;
     Hash hash = xdrSha256(msg.qSet());
+    maybeProcessPingResponse(hash);
     mApp.getHerder().recvSCPQuorumSet(hash, msg.qSet());
 }
 
--- a/src/overlay/Peer.cpp
+++ b/src/overlay/Peer.cpp
@@ -60,11 +60,78 @@ Peer::Peer(Application& app, PeerRole role)
     , mLastWrite(app.getClock().now())
     , mEnqueueTimeOfLastWrite(app.getClock().now())
     , mPeerMetrics(app.getClock().now())
+    , mFlowControlState(Peer::FlowControlState::DONT_KNOW)
+    , mCapacity{app.getConfig().PEER_FLOOD_READING_CAPACITY,
+                app.getConfig().PEER_READING_CAPACITY}
 {
     mPingSentTime = PING_NOT_SENT;
     mLastPing = std::chrono::hours(24); // some default very high value
     auto bytes = randomBytes(mSendNonce.size());
     std::copy(bytes.begin(), bytes.end(), mSendNonce.begin());
+}
+
+void
+Peer::beginMesssageProcessing(StellarMessage const& msg)
+{
+    // Check if flow control is enabled on the local node
+    if (flowControlEnabled() == Peer::FlowControlState::ENABLED)
+    {
+        releaseAssert(mCapacity.mTotalCapacity > 0);
+        mCapacity.mTotalCapacity--;
+
+        if (mApp.getOverlayManager().isFloodMessage(msg))
+        {
+            if (mCapacity.mFloodCapacity == 0)
+            {
+                drop("unexpected flood message, peer at capacity",
+                     Peer::DropDirection::WE_DROPPED_REMOTE,
+                     Peer::DropMode::IGNORE_WRITE_QUEUE);
+                return;
+            }
+
+            mCapacity.mFloodCapacity--;
+            if (mCapacity.mFloodCapacity == 0)
+            {
+                CLOG_TRACE(Overlay, "{}: no capacity for peer {}",
+                           mApp.getConfig().toShortString(
+                               mApp.getConfig().NODE_SEED.getPublicKey()),
+                           mApp.getConfig().toShortString(getPeerID()));
+            }
+        }
+    }
+}
+
+Peer::MsgCapacityTracker::MsgCapacityTracker(std::weak_ptr<Peer> peer,
+                                             StellarMessage const& msg)
+    : mWeakPeer(peer), mMsg(msg)
+{
+    auto self = mWeakPeer.lock();
+    if (!self)
+    {
+        throw std::runtime_error("Invalid peer");
+    }
+    self->beginMesssageProcessing(mMsg);
+}
+
+Peer::MsgCapacityTracker::~MsgCapacityTracker()
+{
+    auto self = mWeakPeer.lock();
+    if (self)
+    {
+        self->endMessageProcessing(mMsg);
+    }
+}
+
+StellarMessage const&
+Peer::MsgCapacityTracker::getMessage()
+{
+    return mMsg;
+}
+
+std::weak_ptr<Peer>
+Peer::MsgCapacityTracker::getPeer()
+{
+    return mWeakPeer;
 }
 
 void
@@ -85,7 +152,9 @@ Peer::sendHello()
     elo.peerID = cfg.NODE_SEED.getPublicKey();
     elo.cert = this->getAuthCert();
     elo.nonce = mSendNonce;
-    sendMessage(msg);
+
+    auto msgPtr = std::make_shared<StellarMessage const>(msg);
+    sendMessage(msgPtr);
 }
 
 AuthCert
@@ -190,7 +259,8 @@ Peer::sendAuth()
     ZoneScoped;
     StellarMessage msg;
     msg.type(AUTH);
-    sendMessage(msg);
+    auto msgPtr = std::make_shared<StellarMessage const>(msg);
+    sendMessage(msgPtr);
 }
 
 std::string const&
@@ -225,8 +295,8 @@ Peer::sendDontHave(MessageType type, uint256 const& itemID)
     msg.type(DONT_HAVE);
     msg.dontHave().reqHash = itemID;
     msg.dontHave().type = type;
-
-    sendMessage(msg);
+    auto msgPtr = std::make_shared<StellarMessage const>(msg);
+    sendMessage(msgPtr);
 }
 
 void
@@ -236,8 +306,8 @@ Peer::sendSCPQuorumSet(SCPQuorumSetPtr qSet)
     StellarMessage msg;
     msg.type(SCP_QUORUMSET);
     msg.qSet() = *qSet;
-
-    sendMessage(msg);
+    auto msgPtr = std::make_shared<StellarMessage const>(msg);
+    sendMessage(msgPtr);
 }
 
 void
@@ -248,7 +318,8 @@ Peer::sendGetTxSet(uint256 const& setID)
     newMsg.type(GET_TX_SET);
     newMsg.txSetHash() = setID;
 
-    sendMessage(newMsg);
+    auto msgPtr = std::make_shared<StellarMessage const>(newMsg);
+    sendMessage(msgPtr);
 }
 
 void
@@ -259,7 +330,8 @@ Peer::sendGetQuorumSet(uint256 const& setID)
     newMsg.type(GET_SCP_QUORUMSET);
     newMsg.qSetHash() = setID;
 
-    sendMessage(newMsg);
+    auto msgPtr = std::make_shared<StellarMessage const>(newMsg);
+    sendMessage(msgPtr);
 }
 
 void
@@ -268,8 +340,8 @@ Peer::sendGetPeers()
     ZoneScoped;
     StellarMessage newMsg;
     newMsg.type(GET_PEERS);
-
-    sendMessage(newMsg);
+    auto msgPtr = std::make_shared<StellarMessage const>(newMsg);
+    sendMessage(msgPtr);
 }
 
 void
@@ -279,8 +351,8 @@ Peer::sendGetScpState(uint32 ledgerSeq)
     StellarMessage newMsg;
     newMsg.type(GET_SCP_STATE);
     newMsg.getSCPLedgerSeq() = ledgerSeq;
-
-    sendMessage(newMsg);
+    auto msgPtr = std::make_shared<StellarMessage const>(newMsg);
+    sendMessage(msgPtr);
 }
 
 void
@@ -303,7 +375,8 @@ Peer::sendPeers()
         {
             newMsg.peers().push_back(toXdr(address));
         }
-        sendMessage(newMsg);
+        auto msgPtr = std::make_shared<StellarMessage const>(newMsg);
+        sendMessage(msgPtr);
     }
 }
 
@@ -315,7 +388,8 @@ Peer::sendError(ErrorCode error, std::string const& message)
     m.type(ERROR_MSG);
     m.error().code = error;
     m.error().msg = message;
-    sendMessage(m);
+    auto msgPtr = std::make_shared<StellarMessage const>(m);
+    sendMessage(msgPtr);
 }
 
 void
@@ -391,15 +465,17 @@ Peer::msgSummary(StellarMessage const& msg)
     case SURVEY_REQUEST:
     case SURVEY_RESPONSE:
         return SurveyManager::getMsgSummary(msg);
+    case SEND_MORE:
+        return "SENDMORE";
     }
     return "UNKNOWN";
 }
 
 void
-Peer::sendMessage(StellarMessage const& msg, bool log)
+Peer::sendMessage(std::shared_ptr<StellarMessage const> msg, bool log)
 {
     ZoneScoped;
-    CLOG_TRACE(Overlay, "send: {} to : {}", msgSummary(msg),
+    CLOG_TRACE(Overlay, "send: {} to : {}", msgSummary(*msg),
                mApp.getConfig().toShortString(mPeerID));
 
     // There are really _two_ layers of queues, one in Scheduler for actions and
@@ -417,7 +493,7 @@ Peer::sendMessage(StellarMessage const& msg, bool log)
         return;
     }
 
-    switch (msg.type())
+    switch (msg->type())
     {
     case ERROR_MSG:
         getOverlayMetrics().mSendErrorMeter.Mark();
@@ -464,8 +540,33 @@ Peer::sendMessage(StellarMessage const& msg, bool log)
     case SURVEY_RESPONSE:
         getOverlayMetrics().mSendSurveyResponseMeter.Mark();
         break;
+    case SEND_MORE:
+        getOverlayMetrics().mSendSendMoreMeter.Mark();
+        break;
     };
 
+    if (mApp.getOverlayManager().isFloodMessage(*msg))
+    {
+        auto flowControl = flowControlEnabled();
+        if (flowControl == Peer::FlowControlState::DONT_KNOW)
+        {
+            // Drop any flood messages while flow control is being set
+            return;
+        }
+        else if (flowControl == Peer::FlowControlState::ENABLED)
+        {
+            addMsgAndMaybeTrimQueue(msg);
+            maybeSendNextBatch();
+            return;
+        }
+    }
+
+    sendAuthenticatedMessage(*msg);
+}
+
+void
+Peer::sendAuthenticatedMessage(StellarMessage const& msg)
+{
     AuthenticatedMessage amsg;
     amsg.v0().message = msg;
     if (msg.type() != HELLO && msg.type() != ERROR_MSG)
@@ -585,10 +686,59 @@ Peer::recvMessage(StellarMessage const& stellarMsg)
     switch (stellarMsg.type())
     {
     // group messages used during handshake, process those synchronously
-    case MessageType::HELLO:
-    case MessageType::AUTH:
+    case HELLO:
+    case AUTH:
         Peer::recvRawMessage(stellarMsg);
         return;
+    case SEND_MORE:
+    {
+        if (mRemoteOverlayVersion < Peer::FIRST_VERSION_SUPPORTING_FLOW_CONTROL)
+        {
+            drop("Peer sent SEND_MORE that it doesn't support",
+                 Peer::DropDirection::WE_DROPPED_REMOTE,
+                 Peer::DropMode::IGNORE_WRITE_QUEUE);
+            return;
+        }
+
+        if (mApp.getConfig().OVERLAY_PROTOCOL_VERSION <
+            Peer::FIRST_VERSION_SUPPORTING_FLOW_CONTROL)
+        {
+            drop("does not support SEND_MORE",
+                 Peer::DropDirection::WE_DROPPED_REMOTE,
+                 Peer::DropMode::IGNORE_WRITE_QUEUE);
+            return;
+        }
+
+        cat = "CTRL";
+        // Set this once: peers aren't allowed to switch between modes
+        // Either set the flow control to "off" or continue processing the
+        // message and potentially send the next batch of items
+        if (mFlowControlState == Peer::FlowControlState::DONT_KNOW)
+        {
+            if (stellarMsg.sendMoreMessage().numMessages == 0)
+            {
+                mFlowControlState = Peer::FlowControlState::DISABLED;
+                // Done with the message
+                return;
+            }
+            else
+            {
+                mFlowControlState = Peer::FlowControlState::ENABLED;
+            }
+        }
+
+        if (stellarMsg.sendMoreMessage().numMessages >
+            UINT64_MAX - mOutboundCapacity)
+        {
+            drop("Peer capacity overflow",
+                 Peer::DropDirection::WE_DROPPED_REMOTE,
+                 Peer::DropMode::IGNORE_WRITE_QUEUE);
+            return;
+        }
+
+        mOutboundCapacity += stellarMsg.sendMoreMessage().numMessages;
+        break;
+    }
 
     // control messages
     case GET_PEERS:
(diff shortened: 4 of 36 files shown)


src/overlay/Peer.h
```diff
@@ -121,6 +145,32 @@ class Peer : public std::enable_shared_from_this<Peer>,
...
+        VirtualClock::time_point mTimeEmplaced;
+    };
+
+    struct ReadingCapacity
+    {
+        uint32_t mFloodCapacity;
+        uint32_t mTotalCapacity;
+    };
+
+    std::vector<OutboundSCPMessage> mOutboundSCPMessages;
```
graydon, line 161: This part kinda makes me nervous too: adding queues! A message already traverses an inbound buffered stream, the action queue, the tx queue, the action queue again, the outbound message-writing queue, and the outbound stream buffer.

I guess I feel (this is a bit abstract, please let me know why I'm wrong) like we _ought_ to be able to hook the new flow control events (stop sending / message being processed / message done / send more) into some places in the existing queues rather than adding new ones.

src/overlay/TCPPeer.cpp
```diff
@@ -487,6 +500,14 @@ TCPPeer::startRead()
                 }
                 noteFullyReadBody(length);
                 recvMessage();
+                if (!hasReadingCapacity())
+                {
+                    // Break and wait until more capacity frees up
+                    CLOG_TRACE(Overlay, "Throttle reading for peer {}!",
+                               mApp.getConfig().toShortString(getPeerID()));
+                    mShouldScheduleRead = false;
```
graydon, line 508: This part seems to be the key, and I am happy to see it! I am sorry the rest of the review seems so critical -- I am I guess somewhat curious to see if you can get away with something smaller that just uses this, the message trackers, and some adjustments / improvements to the write-queue load shedding on the sending side.

src/overlay/Tracker.h
```diff
@@ -39,6 +39,14 @@ using AskPeer = std::function<void(Peer::pointer, Hash)>;
 
 class Tracker
 {
+  public:
+    struct EnvelopeToTrack
+    {
+        Hash hash;
+        SCPEnvelope envelope;
+        Peer::MessageTrackerPtr cb;
```
graydon, line 47: calling this `cb` is a little obscure. Maybe like `fetchProvokingMessage`?

(Assuming that's what it is, and not the tracker for the fetch message itself)

graydon (changes requested):
Nice work! I left a scatter of comments, roughly in three themes. We should probably discuss:

  - The threshold broadcast to N-of-K issue
  - The question of whether we can do without our own flow control signalling, just layer on TCP
  - The question of whether we can do without another layer of queues, just use the existing write queue

I know that's a lot to go through, and I'm sorry to be bringing it up late like this. I'm generally excited to get better end-to-end backpressure -- just want to get there with the fewest new moving pieces we can.

src/overlay/Peer.cpp
```diff
@@ -42,6 +43,133 @@
 namespace stellar
 {
 
+// When flow control is enabled, peer asks for more data every time it processes
+// capacity / FLOW_CONTROL_SEND_MORE_PIPELINE_RATIO messages
+const uint32_t FLOW_CONTROL_SEND_MORE_PIPELINE_RATIO = 4;
```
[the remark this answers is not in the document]
marta-lokhova: chatted offline about this. Relying completely on TCP sliding window logic would require us to have multiple connections for different traffic types. This is because in order to process SCP messages completely, we need to fetch data from the network. So if we pause reading, we'd never complete fetching and free up capacity, creating a deadlock. 

In this implementation, we're handling two streams of data already ("flood" and "control" traffic), where control traffic is throttled via throttled reading, and flood traffic is throttled by explicitly communicating the load status to the sender (the latter also gives us some flexibility, meaning that in the future we could attach some relevant info for the sender to make more intelligent decision on what to flood next)

src/herder/TransactionQueue.cpp
```diff
@@ -736,8 +736,8 @@ TransactionQueue::broadcastTx(AccountState& state, TimestampedTx& tx)
 #endif
     tx.mBroadcasted = true;
     state.mBroadcastQueueOps -= tx.mTx->getNumOperations();
-    return mApp.getOverlayManager().broadcastMessage(
-        tx.mTx->toStellarMessage());
+    return mApp.getOverlayManager().broadcastMessage(tx.mTx->toStellarMessage(),
+                                                     nullptr);
```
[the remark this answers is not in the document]
marta-lokhova: Based on a discussion with @MonsieurNicolas, we decided to finish tracking early in this first version, because the transaction queue itself already has throttling/load shedding mechanisms. Also, since the transaction queue implements some prioritization policy as well, we might get very noisy signals if we track the transaction all the way to broadcasting phase. On the other hand, current definition of "amount of work per transaction" might be too optimistic, since we do not include the write time at all. So I think we can have a subsequent change after this PR that improves the accounting of work for transactions.

src/overlay/Peer.cpp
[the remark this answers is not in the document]
nano-o: Have you considered using the QUIC protocol? It supports multiple streams per connection, and each stream has its own, 
independent flow control. So you could have one stream per queue on a single connection.

src/overlay/Peer.cpp
```diff
@@ -623,38 +818,242 @@ Peer::recvMessage(StellarMessage const& stellarMsg)
...
+        dropped++;
+    }
+    if (dropped)
+    {
+        CLOG_INFO(Overlay, "Dropped {} txs to peers {}", dropped,
+                  mApp.getConfig().toShortString(getPeerID()));
+        dropped = 0;
+    }
+
+    while (itSCP != mOutboundSCPMessages.end() &&
```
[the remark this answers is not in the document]
marta-lokhova: Agreed. I updated the code to always send SCP messages first

src/overlay/Peer.cpp
```diff
@@ -42,6 +43,133 @@
...
+    if (self)
+    {
+        // Notify on completion, schedule more reading if necessary
+        self->messageProcessed(mMsg);
+    }
+}
+
+Peer::MessageTracker::~MessageTracker()
+{
+    done();
```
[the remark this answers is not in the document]
marta-lokhova: updated. It's a bit annoying that we have to call `done` manually everywhere (reason why I went with the destructor approach), but it does make the any-N-of-K logic and the shutdown logic simpler (for shutdown, we don't have to worry about the destructor calling methods on half-destroyed objects)

src/main/ApplicationImpl.cpp
```diff
@@ -568,6 +568,10 @@ ApplicationImpl::~ApplicationImpl()
     LOG_INFO(DEFAULT_LOG, "Application destructing");
     try
     {
+        if (mOverlayManager)
```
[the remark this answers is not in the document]
marta-lokhova: the application doesn't always shutdown via `gracefulStop` (for example, if there's an exception). I added this so that we properly reset trackers to not call half-destroyed objects in `done` during shutdown. However, now that we call `done` manually, we don't really this special handling, so I think it can be removed.

src/overlay/Tracker.h
marta-lokhova: I renamed this one to `SCPMsgTracker`: it seemed a bit more natural given that it tracks an SCP message from start to end (with or without fetching involved)

src/overlay/Peer.h
graydon: (this is too abstract to be actionable -- we discussed in chat and it seems like there'd be just as much complexity moving the prioritization and dropping conditions into the action queues, so marking this resolved)

src/overlay/Peer.cpp
```diff
@@ -42,6 +43,133 @@
 namespace stellar
 {
 
+// When flow control is enabled, peer asks for more data every time it processes
+// capacity / FLOW_CONTROL_SEND_MORE_PIPELINE_RATIO messages
+const uint32_t FLOW_CONTROL_SEND_MORE_PIPELINE_RATIO = 4;
+
+static bool
+isFloodMessage(StellarMessage const& msg)
```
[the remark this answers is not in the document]
marta-lokhova: done

src/overlay/Peer.cpp
```diff
@@ -42,6 +43,133 @@
...
+        CLOG_TRACE(Overlay, "Peer {}, begin processing message {}",
+                   mApp.getConfig().toShortString(peer.lock()->getPeerID()),
+                   binToHex(xdrBlake2(mMsg)));
+    }
+}
+
+void
+Peer::MessageTracker::done()
+{
+    auto self = mWeakPeer.lock();
```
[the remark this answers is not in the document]
marta-lokhova: renamed to `peerThatSentMsg`, seems a bit more intuitive

src/overlay/Peer.cpp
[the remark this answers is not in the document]
marta-lokhova: re: window size (re-sharing for posterity)
main thing is to make sure that the pipelining is sufficient, which is directly related to window size. meaning: if the receiver just processed a batch of size B, and asked for the next batch, it should have enough items left to process while it’s waiting for the next batch.

I'm thinking of the worst case scenario: when the sender always has something to send (i.e. it’s producing outbound traffic at a fast rate) and the connection latency is high. I was thinking that 1 round-trip time (outbound refill request, and inbound flood messages) should roughly be equal to `(window size - batch size) * average time it takes to process a task`. I think round trip is something like 30ms-400ms (i.e. nodes within the same state, vs inter-continental). 

I think we can set the window size based on the upper bounds: 400 ms latency, and sender producing messages at 1000 msgs/s (or 1 msg/ms). So setting the window size to something between 500 and 1000 should account for a conservative scenario. When I experimented with tier1 mission and window size 500, the metrics were very similar, and I didn’t see a significant drop in read rate and stuff like that (that's just one specific setup though)

src/herder/HerderImpl.cpp
```diff
@@ -402,7 +403,9 @@ HerderImpl::broadcast(SCPEnvelope const& e)
                    e.statement.slotIndex);
 
         mSCPMetrics.mEnvelopeEmit.Mark();
-        mApp.getOverlayManager().broadcastMessage(m, true);
+        // Broadcast initiated by progress in a consensus round, no need to
+        // track message completion, so pass `nullptr`
+        mApp.getOverlayManager().broadcastMessage(m, nullptr, true);
```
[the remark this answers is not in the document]
marta-lokhova: I think we shouldn't include the tracker here: we're already accounting for the time to process an SCP message (i.e. fetch and broadcast). Here the node advances in a consensus round, so I wouldn't count it towards a specific peer.

src/overlay/Peer.cpp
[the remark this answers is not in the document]
graydon: We discussed this in chat and I agree with this arithmetic. Marking resolved.

src/herder/TransactionQueue.cpp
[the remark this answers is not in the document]
graydon: Ok.

src/herder/HerderImpl.cpp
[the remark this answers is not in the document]
graydon: Ok. It's a little hard for me to judge what represents "work from a message" vs. "work spontaneously generated" given that it's an event driven system, but I haven't a really strong leg to stand on, was just a thought.

src/overlay/Peer.cpp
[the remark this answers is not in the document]
graydon: Great, thanks!

src/overlay/TCPPeer.cpp
```diff
@@ -265,6 +270,7 @@ TCPPeer::messageSender()
             break;
     }
 
+    // TODO: shed SCP messages
```
graydon, line 273: Is this TODO still relevant?

src/overlay/Peer.h
```diff
@@ -55,6 +55,26 @@ class Peer : public std::enable_shared_from_this<Peer>,
   public:
     typedef std::shared_ptr<Peer> pointer;
 
+    // Data structure that records the time it takes to process a message
```
graydon, line 58: I don't think the part of this comment about destructors is true anymore, right?

graydon (changes requested):
I resolved all the previous comments I think you've covered -- thanks! -- but unfortunately there were a few new ones. Some obsolete comments, a couple renaming questions, and really only one substantial question about the correctness of arithmetic in the herder code that updates the message tracker.

Very close now!

src/overlay/Peer.h
```diff
@@ -55,6 +55,26 @@ class Peer : public std::enable_shared_from_this<Peer>,
...
+    // Callers ensure that this struct is only destructed after the message is
+    // done
+    struct MessageTracker
+    {
+        Application& mApp;
+        StellarMessage const mMsg;
+        std::weak_ptr<Peer> mWeakPeer;
+        std::chrono::system_clock::time_point mStart;
+        uint32_t mNumCallsDone{1};
```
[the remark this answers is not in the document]
graydon: thanks for fixing!

src/overlay/Peer.h
```diff
@@ -53,8 +53,32 @@ class Peer : public std::enable_shared_from_this<Peer>,
 {
 
   public:
+    static constexpr uint32_t FIRST_VERSION_SUPPORTING_FLOW_CONTROL = 20;
+
     typedef std::shared_ptr<Peer> pointer;
 
+    // Data structure that records the time it takes to process a message and
+    // sends the signal back to the peer. Callers must manually call
+    // `finishProcessing`.
```
graydon, line 62: I'm so sorry to have noticed this, but: this comment and the two below still mention `finishProcessing` which is no longer the name of the method.

graydon (changes requested):
Tiniest, littlest, most trivial remaining change!

src/overlay/Peer.h
```diff
@@ -53,8 +53,32 @@ class Peer : public std::enable_shared_from_this<Peer>,
...
+    struct MessageTracker
```
MonsieurNicolas, line 63: `MessageTracker` is a bit misleading.

What this thing is doing really is:
* refcount usage
* invoke a callback with some context when that refcount gets to zero

It may be useful to have this tracker "own" some other data to add to the context for the callback etc, like a `StellarMessage` or a `shared_ptr<StellarMessage>` (depending on context), then it allows you to replace function calls like
foo(StellarMessage const& msg, MessageTracker cb) by `foo(TrackerPtr<StellarMessage> msgTracker)`, and access the store object with `msgTracker.getData()`

MonsieurNicolas (changes requested):
only did a partial review - it looks like the code is there just needs some polish to get it across the line

src/overlay/Peer.cpp
```diff
@@ -585,10 +646,29 @@ Peer::recvMessage(StellarMessage const& stellarMsg)
...
+            releaseAssert(!mRemotePeerWantsFlowControl);
+            mRemotePeerWantsFlowControl = std::make_shared<bool>(false);
+            // Done with the message
+            return;
+        }
+        else if (!mRemotePeerWantsFlowControl)
+        {
+            mRemotePeerWantsFlowControl = std::make_shared<bool>(true);
+        }
+        break;
```
MonsieurNicolas, line 670: I am quite not sure if this is right: it seems to me that we should update `mOutboundCapacity` right away and fall through so that `maybeShedLoadAndSendNextBatch` gets called later

src/overlay/Peer.cpp
```diff
@@ -630,38 +710,256 @@ Peer::recvMessage(StellarMessage const& stellarMsg)
...
+                    cat, self->toString(), app.getConfig().PEER_PORT);
+                CLOG_ERROR(Overlay, "Dropping connection with {}: {}", err,
+                           e.what());
+                self->drop("Bad crypto request",
+                           Peer::DropDirection::WE_DROPPED_REMOTE,
+                           Peer::DropMode::IGNORE_WRITE_QUEUE);
             }
+
+            // Notify on completion, schedule more reading if necessary
+            self->messageProcessed(sm);
```
MonsieurNicolas, line 775: I think that if the message gets dropped by the scheduler, this doesn't get called so you end up never releasing capacity. What you were doing before (using shared ptr) was handling this case correctly

src/overlay/Peer.cpp
```diff
@@ -630,38 +710,256 @@ Peer::recvMessage(StellarMessage const& stellarMsg)
...
-    std::weak_ptr<Peer> weak(static_pointer_cast<Peer>(shared_from_this()));
-    mApp.postOnMainThread(
-        [weak, sm = StellarMessage(stellarMsg), mtype = stellarMsg.type(), cat,
-         port = mApp.getConfig().PEER_PORT]() {
-            auto self = weak.lock();
-            if (self)
+    auto self = shared_from_this();
+    std::weak_ptr<Peer> weak(static_pointer_cast<Peer>(self));
+
+    if (mRemoteOverlayVersion >= Peer::FIRST_VERSION_SUPPORTING_FLOW_CONTROL &&
```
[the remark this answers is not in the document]
MonsieurNicolas: note that if you don't do accounting right away, you may end up with messages in flight while you activate flow control, and you don't want to credit capacity if this happens (so you need to only release capacity for messages that actually took capacity)

src/overlay/Peer.cpp
```diff
@@ -630,38 +710,256 @@ Peer::recvMessage(StellarMessage const& stellarMsg)
...
+    {
+        CLOG_TRACE(Overlay, "Dropped {} messages to peer {}", dropped,
+                   mApp.getConfig().toShortString(getPeerID()));
+    }
+}
+
+void
+Peer::maybeShedLoadAndSendNextBatch()
+{
+    releaseAssert(mRemoteOverlayVersion >=
```
MonsieurNicolas, line 895: this function should be reworked a bit:
it's called for every message added and every time we add capacity. This ends up using a lot of resources.

Maybe rate limit how often you trim?

src/overlay/Peer.cpp
```diff
@@ -630,38 +710,256 @@ Peer::recvMessage(StellarMessage const& stellarMsg)
...
+        }
+    };
+
+    int dropped = 0;
+    auto it = mFlowControlledQueue.cbegin();
+    while (it != mFlowControlledQueue.cend())
+    {
+        if (shouldTrim(*it))
+        {
+            getOutboundFloodTimer(mApp, it->mMessage)
```
MonsieurNicolas, line 875: why do you update this timer if the message gets dropped?

src/overlay/Peer.cpp
```diff
@@ -630,38 +710,256 @@ Peer::recvMessage(StellarMessage const& stellarMsg)
...
+
+    return app.getMetrics().NewTimer({"overlay", "outbound-queue", type});
+}
+
+void
+Peer::maybeTrimQueue()
+{
+    auto now = mApp.getClock().now();
+    auto shouldTrim = [&](QueuedOutboundMessage const& msg) -> bool {
+        if (msg.mMessage.type() == TRANSACTION)
```
MonsieurNicolas, line 843: we should just put an upper bound on the number of transactions that can sit there, that way you don't need to keep track of how long they wait and trim is a single call to "erase"

src/overlay/Peer.cpp
```diff
@@ -630,38 +710,256 @@ Peer::recvMessage(StellarMessage const& stellarMsg)
...
+            // Trim transactions that sit in the queue for too long
+            return (now - msg.mTimeEmplaced) >
+                   Peer::OUTBOUND_TRANSACTION_DROP_TIMEOUT;
+        }
+        else if (msg.mMessage.type() == SCP_MESSAGE)
+        {
+            auto slotIndex = msg.mMessage.envelope().statement.slotIndex;
+            auto minSlotToRemember =
+                mApp.getHerder().getMinLedgerSeqToRemember();
+            if (slotIndex >= minSlotToRemember &&
```
MonsieurNicolas, line 854: maybe you can avoid scanning all SCP messages for all slots if you just pass as a "hint" to this trim function the SCP message that you just added to the queue, that way you can focus on messages for a specific validator.

This hint could also be used to avoid trimming SCP messages when you send a transaction

MonsieurNicolas: This may actually be better than throttling as you will perform a lot less comparisons.

src/overlay/Peer.h
```diff
@@ -112,6 +116,18 @@ class Peer : public std::enable_shared_from_this<Peer>,
         xdr::msg_ptr mMessage;
     };
 
+    struct QueuedOutboundMessage
+    {
+        StellarMessage mMessage;
```
MonsieurNicolas, line 121: we should find a way to not copy messages - I suspect this adds quite a bit of overhead (or we need to move things properly?)

MonsieurNicolas (changes requested):
another round of review. Thank you

src/overlay/Peer.cpp
```diff
@@ -429,8 +453,14 @@ Peer::sendMessage(StellarMessage const& msg, bool log)
         getOverlayMetrics().mSendAuthMeter.Mark();
         break;
     case DONT_HAVE:
+    {
+        if (msg.dontHave().type == SCP_QUORUMSET)
+        {
```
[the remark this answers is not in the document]
marta-lokhova: hmm, ok. before you asked that if the node is accumulating a lot of scp messages, you want it to be reflected in ping, but I can update it (in the previous version, I was doing exactly that already).

src/overlay/Peer.cpp
marta-lokhova: not sure I understand what you're asking. Here we continue processing the message and are guaranteed to invoke `maybeShedLoadAndSendNextBatch` in case there's something in the queue to send. If you don't do this, `maybeShedLoadAndSendNextBatch` will be invoked next time we generate a new message, and I don't think that's better

src/overlay/Peer.cpp
```diff
@@ -585,10 +646,29 @@ Peer::recvMessage(StellarMessage const& stellarMsg)
...
-    case MessageType::HELLO:
-    case MessageType::AUTH:
+    case HELLO:
+    case AUTH:
         Peer::recvRawMessage(stellarMsg);
         return;
+    case SEND_MORE:
+    {
+        cat = "CTRL";
+        // Set this once: peers aren't allowed to switch between modes
```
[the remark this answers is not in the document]
marta-lokhova: it doesn't make sense to check `ENABLE_OVERLAY_FLOW_CONTROL` as we're processing a flow control request from another peer. `ENABLE_OVERLAY_FLOW_CONTROL` reflects flow control preference of the _local_ node. We can, however, check for `mRemoteOverlayVersion` (there's already a check like that in `recvSendMore`, but it could be moved higher up here)

src/overlay/Peer.cpp
```diff
@@ -464,8 +494,39 @@ Peer::sendMessage(StellarMessage const& msg, bool log)
...
+        // Flow control is not yet set, so don't send anything to the peer,
+        // until we receive the SEND_MORE message
+        if (!mRemotePeerWantsFlowControl)
+        {
+            return;
+        }
+
+        if (*mRemotePeerWantsFlowControl)
+        {
+            mFlowControlledQueue.insert(
```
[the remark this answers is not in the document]
marta-lokhova: * as discussed, Compare is correct here since it meets the strict weak ordering requirements
* yeah, i just wanted to replace my original two vectors with two deques, but ended up with a sorted data structure as per feedback elsewhere. I think it's actually causing some overhead, so I'll update it to `std::array<std::deque<>>`

src/overlay/Peer.cpp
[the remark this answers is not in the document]
MonsieurNicolas: We need both nodes to opt-in: we don't want experimental code to be enabled on production systems.

src/overlay/Peer.cpp
MonsieurNicolas: the way the code is written now, you update capacity much later via the scheduler. So any message in the scheduler that need to send messages will see the capacity as if this message was not received.

src/overlay/Peer.cpp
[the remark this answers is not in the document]
MonsieurNicolas: and even later, I don't think we want to allow connections to have flow control only in one direction: it's more things to test

src/overlay/Peer.cpp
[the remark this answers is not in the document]
MonsieurNicolas: see my other comment: you can't turn on local capacity accounting if the other side didn't enable flow control.

src/overlay/Peer.cpp
```diff
@@ -630,38 +842,233 @@ Peer::recvMessage(StellarMessage const& stellarMsg)
...
+        if (mCapacity.mTotalCapacity > 0)
+        {
+            mCapacity.mTotalCapacity--;
+        }
+
+        if (mApp.getOverlayManager().isFloodMessage(stellarMsg))
+        {
+            if (mCapacity.mFloodCapacity == 0)
+            {
+                // TODO: for now, only test well-behaved peers, so throw for
```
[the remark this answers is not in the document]
MonsieurNicolas: OK. you can leave for now, but this needs to be resolved before we merge this PR

MonsieurNicolas (changes requested):

src/overlay/test/OverlayTests.cpp
```diff
@@ -1139,6 +1139,139 @@ TEST_CASE("inbounds nodes can be promoted to ouboundvalid",
...
+    REQUIRE(it->mMessage.type() == TRANSACTION);
+    REQUIRE(it->mTimeEmplaced == time);
+
+    ++it;
+
+    REQUIRE(it->mMessage.type() == TRANSACTION);
+    REQUIRE(it->mTimeEmplaced == time1);
+}
+
+TEST_CASE("overlay flow control", "[overlay][flowcontrol]")
```
MonsieurNicolas, line 1197: missing tests:
* old vs new (new wants to enable flow control, but old doesn't support it)
* handshake tests
  * various combinations for "send_more" (peer disconnects when receiving it at the wrong time with illegal values like 0 or very large numbers)
  * check that nodes agree on flow control being activated or not
* tests that verify that we don't read too many things at once
* tests that check send_more contract respected (adjust flood capacity on the fly, drop peers that don't respect contract)

src/xdr/Stellar-overlay.x
```diff
@@ -22,6 +22,11 @@ struct Error
     string msg<100>;
 };
 
+struct SendMore
+{
+    uint64 numMessages;
```
MonsieurNicolas, line 27: this should be uint32 otherwise you can overflow `mOutboundCapacity`

src/overlay/Peer.cpp
```diff
@@ -630,38 +710,256 @@ Peer::recvMessage(StellarMessage const& stellarMsg)
...
+    {
+        mCapacity.mFloodCapacity++;
+        CLOG_TRACE(Overlay, "{}: got capacity {} for peer {}",
+                   mApp.getConfig().toShortString(
+                       mApp.getConfig().NODE_SEED.getPublicKey()),
+                   mCapacity.mFloodCapacity,
+                   mApp.getConfig().toShortString(getPeerID()));
+        mFloodMsgsProcessed++;
+
+        auto batchFraction = std::max<uint32_t>(
```
MonsieurNicolas, line 945: this should be called `batchSize`... actually this is the only place where `FLOW_CONTROL_SEND_MORE_PIPELINE_RATIO` is called, so maybe the setting should just be `FLOW_CONTROL_SEND_MORE_BATCH_SIZE` (as I am not even sure it's that meaningful anyways as a fraction)

src/overlay/Peer.h
```diff
@@ -121,6 +137,33 @@ class Peer : public std::enable_shared_from_this<Peer>,
...
+
+    // How many flood messages have we received and processed since sending
+    // SEND_MORE to this peer
+    uint64_t mFloodMsgsProcessed{0};
+
+    // How many flood messages can we send to this peer
+    uint64_t mOutboundCapacity{0};
+
+    // Does this peer want flow control enabled
+    std::shared_ptr<bool> mRemotePeerWantsFlowControl;
```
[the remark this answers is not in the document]
MonsieurNicolas: (see other place - I think we can just make this a `bool`)

src/overlay/Peer.cpp
```diff
@@ -630,38 +710,256 @@ Peer::recvMessage(StellarMessage const& stellarMsg)
...
+    std::weak_ptr<Peer> weak(static_pointer_cast<Peer>(self));
+
+    if (mRemoteOverlayVersion >= Peer::FIRST_VERSION_SUPPORTING_FLOW_CONTROL &&
+        mApp.getConfig().ENABLE_OVERLAY_FLOW_CONTROL)
+    {
+        assert(mCapacity.mTotalCapacity > 0);
+        if (mCapacity.mTotalCapacity > 0)
+        {
+            mCapacity.mTotalCapacity--;
+        }
```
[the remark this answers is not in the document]
marta-lokhova: that doesn't seem like the right behavior. Total capacity is fully controlled by the local peer (i.e. how often we read), so if it's 0 and we still read, that indicates that there's a bug (which is why there's an assert there)

src/overlay/Peer.cpp
```diff
@@ -585,10 +646,29 @@ Peer::recvMessage(StellarMessage const& stellarMsg)
...
         return;
+    case SEND_MORE:
+    {
+        cat = "CTRL";
+        // Set this once: peers aren't allowed to switch between modes
+        // Either set the flow control to "off" or continue processing the
+        // message and potentially send the next batch of items
+        if (stellarMsg.sendMoreMessage().numMessages == 0)
+        {
+            releaseAssert(!mRemotePeerWantsFlowControl);
```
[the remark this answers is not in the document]
marta-lokhova: no, we need to reflect three states here: enabled, disabled and don't know. The last one is needed to deal with accumulating flood traffic while peers connect and establish if they want flow control

src/overlay/Peer.cpp
marta-lokhova: good catch, this is fixed now with RAII

src/overlay/Peer.cpp
```diff
@@ -861,83 +865,101 @@ getOutboundFloodTimer(Application& app, StellarMessage const& msg)
...
-
     int dropped = 0;
-    auto it = mFlowControlledQueue.cbegin();
-    while (it != mFlowControlledQueue.cend())
+    if (lastEnqueuedMessageType == TRANSACTION)
     {
-        if (shouldTrim(*it))
+        auto& queue = mOutboundQueues[1];
+
+        while (queue.size() > Peer::OUTBOUND_TRANSACTION_DROP_LIMIT)
```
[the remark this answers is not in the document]
MonsieurNicolas: actually... I think that because you're calling this function every time a message is added to the queue, you will remove at most a message.

src/overlay/Peer.cpp
```diff
@@ -861,83 +865,101 @@ getOutboundFloodTimer(Application& app, StellarMessage const& msg)
...
+
+        for (auto it = queue.begin(); it != queue.end();)
         {
-            ++it;
+            if (it->mMessage.envelope().statement.slotIndex < minSlotToRemember)
+            {
+                it = queue.erase(it);
+                dropped++;
+            }
+            else if (mApp.getHerder().shouldReplaceSt(
```
MonsieurNicolas, line 901: as this code is invoked every time a new message is placed in the "back" position, all you need to do is something like:
```
            if (it->mMessage.envelope().statement.slotIndex < minSlotToRemember)
            {
                it = queue.erase(it);
                dropped++;
            }
            else if (!valueReplaced && it != queue.end()-1 && mApp.getHerder().shouldReplaceSt(
                         it->mMessage.envelope().statement,
                         queue.back().mMessage.envelope().statement))
            {
                valueReplaced = true;
                *it = std::move(queue.back());
                queue.pop_back(); // this does not invalidate "it"
                dropped++;
                ++it;
            }
            else
            {
                ++it;
            }
```

src/overlay/Peer.cpp
MonsieurNicolas, line 875: Should the max queue size for transactions should just be related to the maximum rate of transactions that can be included in a ledger `mApp.getLedgerManager().getLastMaxTxSetSizeOps()`? This avoids adding this hardcoded value.

src/overlay/Peer.cpp
```diff
@@ -861,83 +865,101 @@ getOutboundFloodTimer(Application& app, StellarMessage const& msg)
...
     {
-        sendAuthenticatedMessage(it->mMessage);
-        getOutboundFloodTimer(mApp, it->mMessage)
-            .Update(mApp.getClock().now() - it->mTimeEmplaced);
-
-        // Only count flood messages towards capacity
-        if (mApp.getOverlayManager().isFloodMessage(it->mMessage))
+        auto& queue = mOutboundQueues[i];
+        auto it = queue.begin();
+        while (it != queue.end() && mOutboundCapacity > 0)
```
MonsieurNicolas, line 952: the code would look better if you use empty/front/pop_front instead of iterator (makes it obvious that we're sending the first message in the queue)

src/scp/SCP.cpp
```diff
@@ -281,12 +281,20 @@ SCP::getLatestMessage(NodeID const& id)
 }
 
 bool
-SCP::isLatestStatement(SCPStatement const& st)
+SCP::shouldReplaceSt(SCPStatement const& st, SCPStatement const& hint)
```
MonsieurNicolas, line 284: a better name for this method and its arguments would be
`isNewerStatementSubprotocol` or something like that (it's basically `isNewerStatement`)

src/scp/SCP.cpp
```diff
@@ -281,12 +281,20 @@ SCP::getLatestMessage(NodeID const& id)
...
 {
+    bool sameSlot = st.slotIndex == hint.slotIndex;
+    bool sameNodeID = st.nodeID == hint.nodeID;
+
+    if (!sameSlot || !sameNodeID)
```
MonsieurNicolas, line 289: you probably don't want to use separate variables here: I don't trust compilers to lazy evaluate the second `==` in this case (that happens to be expensive)

src/scp/Slot.cpp
```diff
@@ -107,32 +107,31 @@ Slot::getLatestMessage(NodeID const& id) const
...
 {
-    bool isLatest = true;
+    bool replace = false;
 
-    auto node = st.nodeID;
-    if (st.pledges.type() == SCPStatementType::SCP_ST_NOMINATE)
+    bool stNomination = st.pledges.type() == SCPStatementType::SCP_ST_NOMINATE;
+    bool hintNomination =
+        hint.pledges.type() == SCPStatementType::SCP_ST_NOMINATE;
+
```
MonsieurNicolas, line 117: you can simplify the conditions below by just checking first that `stNomination == hintNomination`

src/overlay/Peer.cpp
```diff
@@ -1237,6 +1584,22 @@ Peer::recvAuth(StellarMessage const& msg)
...
+        if (mApp.getConfig().ENABLE_OVERLAY_FLOW_CONTROL)
+        {
+            sendSendMore(mApp.getConfig().PEER_FLOOD_READING_CAPACITY);
+        }
+        else
+        {
+            // 0 indicates peer doesn't want to use flow control
+            sendSendMore(0);
+        }
+    }
```
MonsieurNicolas, line 1600: I think you have a bug (missing test?):
you need to disable flow control in the else here as to avoid getting stuck in `DONT_KNOW` state

src/overlay/OverlayManagerImpl.cpp
```diff
@@ -871,6 +871,12 @@ OverlayManagerImpl::isPreferred(Peer* peer) const
     return false;
 }
 
+bool
```
MonsieurNicolas, line 874: adding comment here - I think we should add flow control information to the "peers" endpoint

src/overlay/Peer.cpp
MonsieurNicolas: (invariant: you have at most 1 message per validator per slot per consensus phase)

MonsieurNicolas (changes requested):
nice iteration. I found one problem that I missed earlier.

src/overlay/Peer.cpp
```diff
@@ -630,36 +770,237 @@ Peer::recvMessage(StellarMessage const& stellarMsg)
...
+        // Iterate over the message queue. If we found any messages for slots we
+        // don't keep in-memory anymore, delete those. Otherwise, compare
+        // messages for the same slot and validator against the latest SCP
+        // message and drop
+        auto minSlotToRemember = mApp.getHerder().getMinLedgerSeqToRemember();
+
+        auto latest = queue.back();
+        bool valueReplaced = false;
+
+        for (auto it = queue.begin(); it != queue.end();)
```
MonsieurNicolas, line 894: `deque` may not be the best choice if we end up doing a lot of random location erase (it only performs well when erasing from the head or tail). I think it depends on the type that you're using, if it ends up wrapping shared_ptrs (because of what I wrote elsewhere), then this should not be a problem.

MonsieurNicolas (changes requested):
things I missed yesterday

src/overlay/OverlayManagerImpl.cpp
marta-lokhova: i suggest we do this in a subsequent pr

MonsieurNicolas: ok let's open an issue for this then

src/overlay/Peer.h
```diff
@@ -142,6 +142,16 @@ class Peer : public std::enable_shared_from_this<Peer>,
     uint256 mSendNonce;
     uint256 mRecvNonce;
 
+    struct MsgCapacityTracker
+    {
+        std::weak_ptr<Peer> mWeakPeer;
+        StellarMessage mMsg;
```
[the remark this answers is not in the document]
MonsieurNicolas: Looks like I crossed the streams on this one. ⚡ 👻 
Now that I am back...
Yes you are totally right. This is the exact pattern that we need to follow for trackers (I even alluded to this in my first round of comments)

MonsieurNicolas (changes requested):

MonsieurNicolas:
r+ 6c27c408c9e4040e2fc834bf5484bd5c1e81ee8c

PR #3277, opened 2021-11-26
Use TRACY_NO_ISA_EXTENSIONS in Makefile.am

graydon:
Some of the auxiliary tracy tools (`tracy-csvexport` and `tracy-capture`) get built with -mcpu=native or -mcpu=sse4.2, which can produce instructions in compiled binaries (thus packages and docker images) that can only run those tools on machines as-new-or-newer than the build host. This PR attempts to use more conservative build flags for those tools.


MonsieurNicolas:
r+ 501178dc2ed3717348b16543d11f4cf902bd4207

adamlaska:
Скачайте Outlook для Android<https://aka.ms/AAb9ysg>
________________________________

PR #3280, opened 2021-11-29
Dampen arb flooding

graydon:
This adds a new statistical damping pass to the herder's transaction queueing/flooding machinery, to try to traffic-shape the flow of cyclical (arbitrage-attempt) path payments on a per-asset-pair basis. The network has been experiencing very high volumes of these txs (almost all of which fail); this gives validators a new tool for managing that volume. It has a controllable threshold and intensity (managed by a pair of config variables) and is set to a reasonable damping factor by default.


MonsieurNicolas:
r+ b36f935dd4f8b3f8bfda07cb8b00509e714efbd2

PR #3282, opened 2021-12-01
Captive core buffers ledgers on startup

### LINKED ISSUE #3237: Fast Captive-Core Restart
If captive-core is slow to rebuild, it can lose sync causing a 5 minute delay. One way to potentially avoid this would be to add a network-only-client that buffers ledgers from the previously synced captive-core state until it completes its rebuild. Then the network-only-client streams the relevant buffered ledgers back to captive-core, completing the catchup.

marta-lokhova:
Resolves #3237 

This change starts captive core and bucket application simultaneously, which allows core to buffer and save any new ledgers, which can be later replayed. While core is replaying state, certain paths that use the state are inhibited. This includes transactions processing, slot validation etc. 

I tested this by setting the config ARTIFICIALLY_DELAY_BUCKET_APPLICATION_FOR_TESTING to a high value (like 2 minutes), and verifying that captive core was able to join the network right away after applying buckets. 

One thing that worries me is that core is running and interacting with the network while its state is not ready, which can be potentially dangerous, so I'd love to get more eyes on this change, and see if there are additional ways to make this safer. I considered temporarily setting LCL to genesis while we're replaying buckets, but that doesn't seem that much better since we still need to handle all the places where LCL and the ledger state potentially mismatch.

### DIFF
--- a/src/catchup/ApplyBucketsWork.cpp
+++ b/src/catchup/ApplyBucketsWork.cpp
@@ -306,8 +306,9 @@ ApplyBucketsWork::isLevelComplete()
 std::string
 ApplyBucketsWork::getStatus() const
 {
+    auto size = mTotalSize == 0 ? 0 : (100 * mAppliedSize / mTotalSize);
     return fmt::format(
-        FMT_STRING("Applying buckets {:d}%. Currently on level {:d}"),
-        (100 * mAppliedSize / mTotalSize), mLevel);
+        FMT_STRING("Applying buckets {:d}%. Currently on level {:d}"), size,
+        mLevel);
 }
 }
--- a/src/catchup/ApplyBucketsWork.cpp
+++ b/src/catchup/ApplyBucketsWork.cpp
@@ -10,9 +10,9 @@
 #include "catchup/CatchupManager.h"
 #include "crypto/Hex.h"
 #include "crypto/SecretKey.h"
-#include "history/HistoryArchive.h"
 #include "historywork/Progress.h"
 #include "invariant/InvariantManager.h"
+#include "ledger/LedgerManager.h"
 #include "ledger/LedgerTxn.h"
 #include "main/Application.h"
 #include "transactions/TransactionUtils.h"
@@ -153,6 +153,7 @@ ApplyBucketsWork::onReset()
 
     mLevel = BucketList::kNumLevels - 1;
     mApplying = false;
+    mDelayChecked = false;
 
     mSnapBucket.reset();
     mCurrBucket.reset();
@@ -197,6 +198,20 @@ BasicWork::State
 ApplyBucketsWork::onRun()
 {
     ZoneScoped;
+
+    if (mApp.getLedgerManager().rebuildingInMemoryState() && !mDelayChecked)
+    {
+        mDelayChecked = true;
+        auto delay =
+            mApp.getConfig().ARTIFICIALLY_DELAY_BUCKET_APPLICATION_FOR_TESTING;
+        if (delay != std::chrono::seconds::zero())
+        {
+            CLOG_INFO(History, "Delay bucket application by {} seconds",
+                      delay.count());
+            setupWaitingCallback(delay);
+            return State::WORK_WAITING;
+        }
+    }
 
     // Check if we're at the beginning of the new level
     if (isLevelComplete())
--- a/src/catchup/CatchupConfiguration.cpp
+++ b/src/catchup/CatchupConfiguration.cpp
@@ -10,27 +10,58 @@
 namespace stellar
 {
 
+void
+CatchupConfiguration::checkInvariants() const
+{
+    if (mMode == CatchupConfiguration::Mode::LOCAL_BUCKETS_ONLY)
+    {
+        releaseAssert(mHAS && mHistoryEntry);
+        releaseAssert(toLedger() != CatchupConfiguration::CURRENT);
+        releaseAssert(count() == 0);
+    }
+    else
+    {
+        releaseAssert(!mHAS && !mHistoryEntry);
+    }
+}
+
 CatchupConfiguration::CatchupConfiguration(LedgerNumHashPair ledgerHashPair,
                                            uint32_t count, Mode mode)
     : mCount{count}, mLedgerHashPair{ledgerHashPair}, mMode{mode}
 {
+    checkInvariants();
+}
+
+CatchupConfiguration::CatchupConfiguration(HistoryArchiveState has,
+                                           LedgerHeaderHistoryEntry lhhe)
+    : mCount(0)
+    , mLedgerHashPair(LedgerNumHashPair(lhhe.header.ledgerSeq,
+                                        std::make_optional(lhhe.hash)))
+    , mMode(CatchupConfiguration::Mode::LOCAL_BUCKETS_ONLY)
+    , mHAS(std::make_optional(has))
+    , mHistoryEntry(std::make_optional(lhhe))
+{
+    checkInvariants();
 }
 
 CatchupConfiguration::CatchupConfiguration(uint32_t toLedger, uint32_t count,
                                            Mode mode)
     : mCount{count}, mLedgerHashPair{toLedger, std::nullopt}, mMode{mode}
 {
+    checkInvariants();
 }
 
 CatchupConfiguration
 CatchupConfiguration::resolve(uint32_t remoteCheckpoint) const
 {
+    checkInvariants();
+    auto cfg = *this;
     if (toLedger() == CatchupConfiguration::CURRENT)
     {
-        return CatchupConfiguration{remoteCheckpoint, count(), mMode};
+        cfg.mLedgerHashPair.first = remoteCheckpoint;
+        return cfg;
     }
-    return CatchupConfiguration{LedgerNumHashPair(toLedger(), hash()), count(),
-                                mode()};
+    return cfg;
 }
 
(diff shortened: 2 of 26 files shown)


MonsieurNicolas:
I started to review your change but then scrapped it as ultimately something bothered me.

The use of `RebuildInMemoryState` does not seem right:
* it introduces something similar to a global variable that needs to be handled in places that are not super obvious
* doesn't really help with correctness (previous point magnifies this)
* you still end up with the code written with "2 main loops" (captive-core right now is performing in `setupApp` a bunch of work synchronously before the "real" main loop)

I think that you will end up with a change smaller in scope (and easier to validate) by making `setupApp` just setup the app in such a way that it's already "catching up" and *not* special case much in the rest of the app when you want to reconstruct state:
* I think that you can initialize the app with lcl = ledger zero (~genesis ledger, as it will not be used for anything critical, it should be fine) - note that lcl would be at seq 0 *not* 1 (unlike what happens when starting a fresh node), so that no SCP message could possibly cause the node to advance.
* you need to tell herder that it's tracking whatever state we want to start from (like the last known state), without this, herder could track slots older than what it's reconstructing with catchup
* then you just need to change `CatchupWork` so that you can give it a "has"/"ledgerHeader" that corresponds to the last known state. It will apply buckets and update LCL properly as lcl.buckets = {} (so it will apply all buckets, etc)
   * we may want to do something here in case of failure, right now captive core just exits with an error (I am not sure it's actionable) so maybe just do that.

That way:
* we are a step close to get rid of the weird "2 main loops"
   * we should be able to do something similar with the blocking call to `catchup` in `setupApp` (if not in this PR, in a subsequent one)
   * it may still be worth spending a bit of time thinking about this to validate the approach
* when the "initial catchup" completes, it will just catchup to the network like normal
* there is nothing special here, in particular, you don't need the changes that you made in Herder and LedgerManager and going forward if we make changes we won't have to think too much about how captive-core differs

src/ledger/LedgerManagerImpl.cpp
```diff
@@ -408,6 +433,12 @@ LedgerManagerImpl::getLastClosedLedgerHeader() const
 HistoryArchiveState
 LedgerManagerImpl::getLastClosedLedgerHAS()
 {
+    if (rebuildingInMemoryState())
```
MonsieurNicolas, line 436: did you need to special case this here? We could not initialize the instance with `has = {}` in this case instead?

src/ledger/LedgerManagerImpl.h
```diff
@@ -57,6 +57,11 @@ class LedgerManagerImpl : public LedgerManager
     medida::Counter& mLedgerAge;
     medida::Timer& mMetaStreamWriteTime;
     VirtualClock::time_point mLastClose;
+    bool mRebuildInMemoryState{false};
+    // References to buckets being applied during state rebuild; this is needed
+    // to ensure we do not accidentally garbage-collect buckets during in-memory
+    // state rebuild
+    std::set<std::shared_ptr<Bucket>> mRetainBucketsDuringRebuild;
```
MonsieurNicolas, line 64: This seems a bit odd - shouldn't that be referenced by catchup as it's the "thing" that applies buckets: we just need to make sure that buckets are referenced as soon as we create the catchupwork that rebuilds state?

src/main/ApplicationImpl.cpp
```diff
@@ -681,14 +681,7 @@ ApplicationImpl::start()
...
     bool done = false;
-    mLedgerManager->loadLastKnownLedger([this,
-                                         &done](asio::error_code const& ec) {
-        if (ec)
-        {
-            throw std::runtime_error(
-                "Unable to restore last-known ledger state");
-        }
-
+    mLedgerManager->loadLastKnownLedger([this, &done]() {
```
MonsieurNicolas, line 684: this can't fail anymore?

MonsieurNicolas (changes requested):
A few suggestions. I think we'll want to adjust the `catchup` path as well (when we don't have local buckets) at some point (we can do that in a separate PR). That way we'll only have one way of "starting up captive core".

src/catchup/ApplyBucketsWork.cpp
```diff
@@ -306,8 +306,9 @@ ApplyBucketsWork::isLevelComplete()
 std::string
 ApplyBucketsWork::getStatus() const
 {
+    auto size = mTotalSize == 0 ? 0 : (100 * mAppliedSize / mTotalSize);
```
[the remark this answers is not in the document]
marta-lokhova: You are right that once `ApplyBucketsWork` is properly setup, we can't have 0 buckets. But it is possible that we call `getStatus` while Work is in between states. Specifically, when work has been created and passed to the WorkSequence, but it hasn't "started" (where we set state to RUNNING, call reset, etc). This issue was already present in master, but got revealed by this change

src/main/ApplicationImpl.cpp
marta-lokhova: yeah, this is a minor cleanup (needed regardless of this change). The only place we had an actual error code was in "repair buckets" work, but that was removed a long time ago in https://github.com/stellar/stellar-core/pull/2232/files#diff-ce04c2790c0d0c23aa7da5c9e2053e7a60f2e286fccb33a1b98dc704340c3f87L325 So we don't need `ec` and the lambda

src/catchup/CatchupConfiguration.h
```diff
@@ -46,13 +47,19 @@ class CatchupConfiguration
         // Do validity checks on all history archive file types for a given
         // range, regardless of whether files are used or not
         OFFLINE_COMPLETE,
-        ONLINE
+        ONLINE,
+        // Similar to online catchup, except in this mode there is no archive
+        // lookup and validation, rebuild local state present on disk, while
+        // buffering ledgers
+        LOCAL
```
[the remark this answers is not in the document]
marta-lokhova: renamed to `LOCAL_BUCKETS_ONLY`

src/catchup/CatchupConfiguration.cpp
```diff
@@ -16,6 +16,18 @@ CatchupConfiguration::CatchupConfiguration(LedgerNumHashPair ledgerHashPair,
...
+    : mCount(0)
+    , mLedgerHashPair(LedgerNumHashPair(lhhe.header.ledgerSeq,
+                                        std::make_optional(lhhe.hash)))
+    , mMode(CatchupConfiguration::Mode::LOCAL)
+    , mHAS(std::make_optional(has))
+    , mHistoryEntry(std::make_optional(lhhe))
+{
+    releaseAssert(mHAS->currentLedger == mHistoryEntry->header.ledgerSeq);
+}
+
```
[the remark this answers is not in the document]
marta-lokhova: done. All constructors now call `checkInvariants` function, which checks several things. This also revealed a bug in ` CatchupConfiguration::resolve`, which I did not update correctly to handle the local mode (it was not caught by tests because the result of `resolve` is only used to calculate catchup range, so it doesn't use the new fields)

src/ledger/LedgerManagerImpl.h
marta-lokhova: Updated, bucket references are now passed to catchup (but collected prior to application start, which invokes garbage collection)

MonsieurNicolas (changes requested):
Found a few minor things. I think we're very close at this point

src/catchup/ApplyBucketsWork.cpp
```diff
@@ -198,6 +199,20 @@ ApplyBucketsWork::onRun()
 {
     ZoneScoped;
 
+    if (mApp.getLedgerManager().rebuildingInMemoryState() && !mDelayChecked)
```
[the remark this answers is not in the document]
marta-lokhova: Yeah, it's safe to call, it's just... annoying :) When I'm testing things, I only want to delay the actual state rebuild (not normal catchup, or genesis rebuild)

marta-lokhova:
@MonsieurNicolas addressed your latest feedback, and added a commit to increase the retry limit for HAS downloads in catchup. I can rebase if things look good

src/catchup/CatchupWork.cpp
```diff
@@ -312,8 +312,12 @@ CatchupWork::getAndMaybeSetHistoryArchiveState()
                     ? CatchupConfiguration::CURRENT
                     : mApp.getHistoryManager().checkpointContainingLedger(
                           mCatchupConfiguration.toLedger());
+            // Set retries to something high in case the current checkpoint
+            // hasn't been published yet
+            size_t retries = mApp.getHistoryManager().getCheckpointFrequency() *
```
MonsieurNicolas, line 317: This seems way too much?
We potentially wait for 8 minutes per retry after the 9th retry, and here you're setting it to 340 retries... so this is basically "infinite".

Right now we use the default of "a few" (5 retries), and "a lot" (32) seems to be way too many already. Maybe the real fix here is to
* change `getRetryDelay` to pick a number between "m" and "m/2+1" for m > 2 (instead of 1...m). That way we have a low/high bound for retry timeout a little more sane.
* use a retry count of 9 for `GetHistoryArchiveStateWork` here (that will be a bit less than 512 seconds as low bound, high bound is 1024 seconds)

MonsieurNicolas (changes requested):

MonsieurNicolas:
r+ 50550ef2c0761c326ff1e07365bf28ee0d9db670

PR #3288, opened 2021-12-09
Shutdown fixes

### LINKED ISSUE #3163: Audit asynchronous events during application shutdown
Discovered in https://github.com/stellar/stellar-core/issues/3106. Application shutdown and IO service shutdown are not synchronous, so it is possible that events get executed in between. The problem is that at this point multiple application components (such as ProcessManager, BucketManager etc) are shutdown. We should audit the code, and ensure that either timers are properly cancelled, or the events triggered no-op if the application is shutting down.

marta-lokhova:
Resolves #3163 
 
The goal of this change is to ensure asynchronous callbacks behave appropriately during application shutdown (when certain subsystems are already in shutdown state, but the application continues to crank and execute events). There are two areas I looked into: timer-based callbacks, and scheduler queues. There doesn't seem to be a problem with timers, as those are properly cancelled in `shutdown` functions. There are a couple of places where events from the scheduler queues can be executed during shutdown, but most of them seem to be already mitigated with early exits if the app is shutting down (specifically, I'm referring to `processSCPQueueSomeMore` in Herder, and message-related functions in Peer). I added on check in `HistoryManagerImpl::publishQueuedHistory`.

In addition, I updated the scheduler shutdown code to not wait until completion, and updated callbacks in related works to no-op if the work scheduler is aborting.

### DIFF
--- a/src/history/HistoryManagerImpl.cpp
+++ b/src/history/HistoryManagerImpl.cpp
@@ -291,6 +291,11 @@ HistoryManagerImpl::takeSnapshotAndPublish(HistoryArchiveState const& has)
 size_t
 HistoryManagerImpl::publishQueuedHistory()
 {
+    if (mApp.isStopping())
+    {
+        return 0;
+    }
+
 #ifdef BUILD_TESTS
     if (!mPublicationEnabled)
     {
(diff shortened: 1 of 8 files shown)


MonsieurNicolas:
r+ 18f0d23ce2d644253f18bfc6e0a4fd32db12a0ed

PR #3290, opened 2021-12-10
Optionally use TransactionResultSets in replay, to accelerate failed txs.

graydon:
This is a sketch of a special mode (set to on-by-default in this PR) that accelerates replay of txsets with a lot of failed txs in them by reusing the result codes from the txresults file. If there's any discrepancy it'll diverge and halt, of course; I think the only thing this could really do _wrong_ is hide divergence in the database by supplying "correct" historical result vectors. But we aren't always able to catch divergence in the DB when it happens anyways; this might make it a little worse but it's not exactly new.

### DIFF
--- a/docs/stellar-core_example.cfg
+++ b/docs/stellar-core_example.cfg
@@ -292,6 +292,11 @@ CATCHUP_COMPLETE=false
 # new history
 CATCHUP_RECENT=0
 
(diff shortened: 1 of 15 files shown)


MonsieurNicolas:
closing stale / obsolete PR (this will not work anyways with Soroban)

PR #3306, opened 2022-01-07
Delete temporary files while catching up

graydon:
This limits temporary disk space used during catchup by deleting tx files as soon as we're done with them. It will still store MAX_CONCURRENT_SUBPROCESSES (default 16) worth of tx files at any time -- and since it verifies the entire header chain before starting it will download and keep that all up front -- but that's comparatively small.

### DIFF
--- a/src/catchup/DownloadApplyTxsWork.cpp
+++ b/src/catchup/DownloadApplyTxsWork.cpp
@@ -12,6 +12,7 @@
 #include "ledger/LedgerManager.h"
 #include "work/ConditionalWork.h"
 #include "work/WorkSequence.h"
+#include "work/WorkWithCallback.h"
 
 #include <Tracy.hpp>
 #include <fmt/format.h>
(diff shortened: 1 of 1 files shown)


graydon:
cc @paulbellamy

marta-lokhova (changes requested):
Thanks for fixing this! I think the file deletion logic belongs in `DownloadApplyTxsWork` though, since this is where we create a sequence of work "download + apply" for each checkpoint. So you can just add a new work, "delete", to the sequence (note that we can use `WorkWithCallback` for this)

graydon:
@marta-lokhova updated! I'll be out the next few days so if you want to make any further tweaks, go ahead!

paulbellamy:
With this, the reingestion that was using 25GB+ now levels off at ~5.8GB.

MonsieurNicolas:
r+ aa6f7c164ba92fcc0a2b8268d03154030c81f8e0

PR #3318, opened 2022-01-21
fix: close files after applying checkpoint

MonsieurNicolas:
#3306 introduced a bug on Windows:
attempting to run a command such as `catchup 2000/200` would fail due to "files in use".

Indeed files were kept open by the `ApplyCheckpointWork` class.


graydon:
r+ 908cde2

Issue #1427, opened 2017-12-01
Output all logging information to stderr instead of stdout, it will allow for piping output


vogel:
Please also check for all documentation that says logging goes to stdout (and all external tools that assume that).

MonsieurNicolas:
I imagine that this should be done when running subcommands that are supposed to be executed from the command line, like `offline-info` or `report-last-history-checkpoint`.
Also those should be changed to not use `LOG` for their result (so that it's possible to grab the result from shell).

MonsieurNicolas:
fixed

PR #3326, opened 2022-01-27
Stricter offline info

graydon:
This changes (very slightly) the way `offline-info` works:

 - Bypasses the logging stream, writes directly to stdout.
 - Avoids a single (I think stray / unintentional) pre-logging-config log-write.
 - Allows user to run `stellar-core --ll fatal offline-info` to get a machine-readable JSON blob.

This is something horizon wanted to be able to programmatically query a core database when running us captive.

cc @paulbellamy

### DIFF
--- a/src/main/CommandLine.cpp
+++ b/src/main/CommandLine.cpp
@@ -477,10 +477,9 @@ CommandLine::ConfigOption::getConfig(bool logToFile) const
     auto configFile =
         mConfigFile.empty() ? std::string{"stellar-core.cfg"} : mConfigFile;
 
-    LOG_INFO(DEFAULT_LOG, "Config from {}", configFile);
-
     // yes you really have to do this 3 times
     Logging::setLogLevel(mLogLevel, nullptr);
+    LOG_INFO(DEFAULT_LOG, "Config from {}", configFile);
     config.load(configFile);
 
     Logging::setFmt(KeyUtils::toShortString(config.NODE_SEED.getPublicKey()));
--- a/docs/quick-reference.md
+++ b/docs/quick-reference.md
@@ -225,9 +225,12 @@ consistent.
 #### Offline info
 
 The command `offline-info` fetches information similar to what `info` returns but
-without starting the node.
+without starting the node, writing it directly to standard output.
 
 This allows to quickly see information about the last loaded ledger, supported protocol version etc.
+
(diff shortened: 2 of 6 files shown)


src/main/CommandLine.cpp
```diff
@@ -477,10 +477,9 @@ CommandLine::ConfigOption::getConfig(bool logToFile) const
     auto configFile =
         mConfigFile.empty() ? std::string{"stellar-core.cfg"} : mConfigFile;
 
-    LOG_INFO(DEFAULT_LOG, "Config from {}", configFile);
-
     // yes you really have to do this 3 times
     Logging::setLogLevel(mLogLevel, nullptr);
+    LOG_INFO(DEFAULT_LOG, "Config from {}", configFile);
```
MonsieurNicolas, line 482: I think it would be better to do #1427 - since I opened that other issue, we moved to spdlog, and I see that we can just swap out `stdout_sink_mt` with `stderr_sink_mt` what do you think?

graydon: Plausible. I suspect it'll break someone, somewhere, who relies on how things currently are. And I _suspect_ horizon is merging both streams for log-consumption currently. But I agree it's "more of the right thing", ish.

(I think? I mean .. there is no stdlog stream. Logs are not always errors. But stderr is from the "strong, silent" era of software where you only wasted teletype output when something went wrong)

MonsieurNicolas: The concern with a change like the one in this PR is that it is brittle: if anybody later adds some logging in the wrong spot (potentially deep in the call stack), it will break clients that expect to find json.

I agree, doing it the other way can potentially break some clients but should be stable over time:
for clients that don't merge streams already, that break will be pretty obvious *and* we'll fix another problem with those clients as they're dropping on the floor whatever can end up in stderr like critical crash information when we abort (we use `std::fprintf(stderr, ` - some other libraries may do the same).

MonsieurNicolas: @graydon - it's a "rip off the band-aid" type of situation. Let's just do it

graydon: @MonsieurNicolas ok, band-aid ripped off.

MonsieurNicolas:
r+ 5c5285e5768e3022848c420f1ecbc61e9d20ff11

PR #3330, opened 2022-01-31
Log exception as INFO (instead ERROR) for older protocol version

### LINKED ISSUE #3073: reduce severity of known (and old) protocol issues
When we hit an error while replaying old ledgers, we still log and update metrics as if those are actual problems. This creates unnecessary stress :)

For protocol versions that are *not* live on the network anymore (as of this writing, protocols older than 17 for example), we should:
* not increment the `ledger.transaction.internal-error` metric as to not trigger alerts
* add information in the logs that this is for an older protocol version, maybe reduce to `INFO` as opposed to `ERROR` - alternatively we could just log the protocol version in the log line

jayz22:
# Description

Resolves #3073

# Checklist
- [x] Compiles
- [x] Ran all tests


src/transactions/TransactionFrame.cpp
```diff
@@ -740,6 +740,19 @@ TransactionFrame::applyOperations(SignatureChecker& signatureChecker,
                                   TransactionMeta& outerMeta)
 {
     ZoneScoped;
+    auto& internalErrorCounter = app.getMetrics().NewCounter(
+        {"ledger", "transaction", "internal-error"});
+
+    // We do not want to treat exception as internal-error for non-live ledger
```
sisuresh, line 746: It looks like the counter is incremented and an error is logged for both `LEDGER_PROTOCOL_VERSION` and `LEDGER_PROTOCOL_VERSION - 1` ledger versions. I would mention that it's still possible to get an internal-error for a non-live ledger due to this.

I wonder if there's a better solution here though. If you have the `LedgerHeader` for the ledger you're catching up to, you could use that version, but I'm not sure if that's easy/worth it to get and expose here.

src/transactions/TransactionFrame.cpp
```diff
@@ -740,6 +740,19 @@ TransactionFrame::applyOperations(SignatureChecker& signatureChecker,
...
+    // versions. The live ledger version is in the head of the ledger, and
+    // normally should be same as the app.getConfig().LEDGER_PROTOCOL_VERSION,
+    // which should also be same as maxAssertLedgerVersion here. However, right
+    // before a protocol upgrade, the LEDGER_PROTOCOL_VERSION (the software
+    // version) could lead the protocol version by 1, thus we need to offset it
+    // from the maxAssertLedgerVersion.
+    uint32_t maxAssertLedgerVersion =
```
sisuresh, line 753: We don't do this so it might not be a concern, but it's technically possible for us to bump `CURRENT_LEDGER_PROTOCOL_VERSION` by more than one. If we did this, the internal error would always be downgraded. It might be a good idea to also check that catchup is running (like we do [here](https://github.com/stellar/stellar-core/blob/a44e07d495ec91e55e120557886a6ee5f7363210/src/ledger/LedgerManagerImpl.cpp#L507)) before downgrading the internal error.

src/transactions/TransactionFrame.cpp
```diff
@@ -740,6 +740,19 @@ TransactionFrame::applyOperations(SignatureChecker& signatureChecker,
...
+        app.getConfig().LEDGER_PROTOCOL_VERSION - 1;
+    uint32_t ledgerVersion = app.getConfig().LEDGER_PROTOCOL_VERSION;
```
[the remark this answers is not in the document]
jayz22: I see what you're saying, but shouldn't we always [initialize an variable ](https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#es20-always-initialize-an-object)to something?
In this case, would you suggest `0` being a better value to initialize to instead of `LEDGER_PROTOCOL_VERSION`?

src/transactions/TransactionFrame.cpp
```diff
@@ -740,6 +740,19 @@ TransactionFrame::applyOperations(SignatureChecker& signatureChecker,
                                   TransactionMeta& outerMeta)
 {
     ZoneScoped;
+    auto& internalErrorCounter = app.getMetrics().NewCounter(
```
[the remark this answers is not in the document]
jayz22: Because previously it was being lazy-initialized. If no exception occurs, instead of outputting "internal-error" count = 0, it'll complain about "internal-error" not being a valid metric.

src/transactions/TransactionFrame.cpp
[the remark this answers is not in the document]
sisuresh: Yeah that's a good point. What about using https://en.cppreference.com/w/cpp/utility/optional? If `ledgerVersion` is accessed prior to it being set, an exception will be thrown, leading to a `txINTERNAL_ERROR`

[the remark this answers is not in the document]
jayz22: That should work, I've changed it to using `std::optional`

jayz22:
@MonsieurNicolas @jonjove Any thoughts on how we should determine the live protocol version?
As a simple implementation, I'm assuming `LEDGER_PROTOCOL_VERSION` and `LEDGER_PROTOCOL_VERSION-1` as the "active" (error-logging) version. However, as @sisuresh points out, this won't work if the code base jumps more than 1 major versions than the ledger header (internal-errors won't be captured). Also this means usually we are treating two versions as active as far as error logging is concerned.
Alternatively we'll need to fetch the network's live header version (>= catch up ledger header target) first and pass it down for logging. 
Want to get your guys' thoughts first before going that route.

jayz22:
@jonjove I think that makes sense. Should we default the new param's value to `LEDGER_PROTOCOL_VERSION-2` instead of 1?

jonjove:
When I posted that comment, I was thinking it would be a compile-time constant (updated by maintainers) rather than a configurable (updated by operators). But in retrospect, after looking at what you did, it's probably better to make it configurable with a default hardcoded by the maintainers.

jayz22:
After discussing with Jon, decided to make this a constant (currently = 18), rather than pinning it to the `LEDGER_PROTOCOL_VERSION`. This will need to be updated during a protocol upgrade so that we don't get old alerts. This small operational overhead is better than potentially missing a critical alert if the protocol version jumps more than what we expect.

src/transactions/TransactionFrame.cpp
```diff
@@ -740,6 +740,19 @@ TransactionFrame::applyOperations(SignatureChecker& signatureChecker,
...
+        {"ledger", "transaction", "internal-error"});
+
+    // We do not want to treat exception as internal-error for non-live ledger
+    // versions. The live ledger version is in the head of the ledger, and
+    // normally should be same as the app.getConfig().LEDGER_PROTOCOL_VERSION,
+    // which should also be same as maxAssertLedgerVersion here. However, right
+    // before a protocol upgrade, the LEDGER_PROTOCOL_VERSION (the software
+    // version) could lead the protocol version by 1, thus we need to offset it
```
sisuresh, line 751: This comment needs to be updated (it still references the -1 offset).

src/transactions/TransactionFrame.cpp
jonjove, line 746: Nitpick 1: this comment is a bit confusing. We do want to treat exceptions as internal errors, but we don't want to increment the metric internal-error. Maybe more precise language would help here.

Nitpick 2: we normally say "ledger header" instead of "head of the ledger"

src/transactions/TransactionFrame.cpp
```diff
@@ -826,22 +839,50 @@ TransactionFrame::applyOperations(SignatureChecker& signatureChecker,
...
+                      xdr_to_string(getContentsHash(), "contentsHash"));
+        }
     }
     // This is only reachable if an exception is thrown
     getResult().result.code(txINTERNAL_ERROR);
 
-    auto& internalErrorCounter = app.getMetrics().NewCounter(
-        {"ledger", "transaction", "internal-error"});
-    internalErrorCounter.inc();
+    // We only count internal error if the ledger number is and older version.
```
jonjove, line 881: Nitpick 1: typo "and" -> "an"

Nitpick 2: I think you mean ledger version here (ledger version is 18, ledger number is about 40M on production network)

Nitpick 3: I think you mean newer version

src/transactions/TransactionFrame.cpp
jonjove, line 753: Nitpick: I think you mean `minAssertLedgerVersion`? Also "Assert" is a bit strange in this context since we actually never `assert` on internal error, but perhaps "Alert" instead? If you rename here, please update the comment to match.

src/transactions/TransactionFrame.cpp
```diff
@@ -740,6 +740,19 @@ TransactionFrame::applyOperations(SignatureChecker& signatureChecker,
...
+    // We do not want to treat exception as internal-error for non-live ledger
+    // versions. The live ledger version is in the head of the ledger, and
+    // normally should be same as the app.getConfig().LEDGER_PROTOCOL_VERSION,
+    // which should also be same as maxAssertLedgerVersion here. However, right
+    // before a protocol upgrade, the LEDGER_PROTOCOL_VERSION (the software
+    // version) could lead the protocol version by 1, thus we need to offset it
+    // from the maxAssertLedgerVersion.
+    uint32_t maxAssertLedgerVersion =
+        app.getConfig().LEDGER_PROTOCOL_MIN_VERSION_INTERNAL_ERROR_REPORT;
+    std::optional<uint32_t> ledgerVersion;
```
jonjove, line 755: I'm not very confident about the use of `std::optional` here. Let's suppose that we don't reach line 765 where it's set (for example, imagine we throw on line 761). We jump down to line 842 in the first catch block, where `ledgerVersion.value()` throws `std::bad_optional_access` (see https://en.cppreference.com/w/cpp/utility/optional/value). This crashes the program.

So there are a few choices:
1. We could check in the `catch` blocks if `ledgerVersion` contains a value. If it doesn't, we just always alert (because it's better to alert when you shouldn't than it is to not alert when you should)
2. We could avoid the use of `std::optional` entirely. To do this, we revert the definition of `uint32_t ledgerVersion` to line 765--like it was before this change--and instead introduce a `bool shouldAlertOnInternalError = true` on this line (`true` for the same reason about always better to alert if you're not sure). Then on line 766 we can set the correct value for `shouldAlertOnInternalError`. This would probably eliminate the need for `maxAssertLedgerVersion` entirely, because it will only be used on that one line.

Second option seems significantly cleaner to me, but I'm open to other ideas.

sisuresh: The second option is a great idea.

jayz22: The reason I thought `std::optional` is fine here is because `TransactionFrame::applyOperations` is only called by `apply` under a try-catch block, so if it wasn't set before it will be caught there. But re-thinking that it is probably not a good assumption. 
I agree approach 2 is significantly better and cleaner. Will make the change.

jayz22: The problem with approach 2 is we'll still need the ledger version in the two `catch` blocks. For that reason, I'm going with option 1. Please see changes.

sisuresh: So you need the ledgerVersion in the catch block purely for logging in approach 2. Do we actually need this? I don't think we care about which old protocol version it is, just that it's less than `LEDGER_PROTOCOL_MIN_VERSION_INTERNAL_ERROR_REPORT`.

jayz22: Yes it is for logging only. It was mentioned in the issue so I interpreted it as a requirement.

sisuresh: The issue mentions logging the protocol version as a suggestion. I think the initial idea from Nico in that issue is sufficient ```"add information in the logs that this is for an older protocol version, maybe reduce to INFO as opposed to ERROR"```. What do you think?

jayz22: I agree, in that case, probably best to leave the version number out of the logging in favor of cleaner code. (Approach 2)

jayz22: See fix.

sisuresh:
LGTM. You should squash your commits and then @jonjove can take a final look.

jonjove:
r+ 1bb964f

PR #3334, opened 2022-02-04
Adding an option to halt on txINTERNAL_ERROR.

### LINKED ISSUE #2609: Add a way to configure a node to halt on "internal error"
Right now if a node runs into an error that corresponds to `txINTERNAL_ERROR` the node will log the problem and close the ledger as normal.

On some deployments, like testnet (and maybe some validators), it would be better to just crash so that the error is easy to spot (right now we only have logs and metrics).

dmkozh:
# Description

Resolves #2609 

With HALT_ON_INTERNAL_TRANSACTION_ERROR setting enabled the core will crash when an internal error is encountered during transaction processing.

Testing: as Catch2 doesn't seem to support death tests, I don't see a good way to cover this. I've checked locally that the old version of protocol cause the crash. I also checked that this testnet/pubnet configs work with the option enabled.

# Checklist
- [ ] Compiles
- [ ] Ran all tests

### DIFF
--- a/src/main/Config.cpp
+++ b/src/main/Config.cpp
@@ -217,6 +217,8 @@ Config::Config() : NODE_SEED(SecretKey::random())
 
     HISTOGRAM_WINDOW_SIZE = std::chrono::seconds(30);
 
+    HALT_ON_INTERNAL_TRANSACTION_ERROR = false;
+
 #ifdef BUILD_TESTS
     TEST_CASES_ENABLED = false;
 #endif
@@ -1286,6 +1288,10 @@ Config::processConfig(std::shared_ptr<cpptoml::table> t)
                         "HISTOGRAM_WINDOW_SIZE must divide 300 evenly");
                 }
                 HISTOGRAM_WINDOW_SIZE = std::chrono::seconds(s);
+            }
+            else if (item.first == "HALT_ON_INTERNAL_TRANSACTION_ERROR")
+            {
+                HALT_ON_INTERNAL_TRANSACTION_ERROR = readBool(item);
             }
             else
             {
--- a/src/main/Config.h
+++ b/src/main/Config.h
@@ -420,6 +420,14 @@ class Config : public std::enable_shared_from_this<Config>
     // the entry cache
     size_t PREFETCH_BATCH_SIZE;
 
+    // If set to true, the application will halt when an internal error is
+    // encountered during applying a transaction. Otherwise, the
+    // txINTERNAL_ERROR transaction is created but not applied.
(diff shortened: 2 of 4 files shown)


src/main/Config.cpp
```diff
@@ -217,6 +217,8 @@ Config::Config() : NODE_SEED(SecretKey::random())
 
     HISTOGRAM_WINDOW_SIZE = std::chrono::seconds(30);
 
+    HALT_ON_INTERNAL_TRANSACTION_ERROR = false;
```
MonsieurNicolas, line 220: need to update example config, this is how we test that parsing is working properly

src/main/Config.h
```diff
@@ -420,6 +420,14 @@ class Config : public std::enable_shared_from_this<Config>
     // the entry cache
     size_t PREFETCH_BATCH_SIZE;
 
+    // If set to true, the application will halt when an internal error is
+    // encountered during applying a transaction. Otherwise, the
```
MonsieurNicolas, line 424: we normally document external facing flags with the example config file

src/transactions/TransactionFrame.cpp
```diff
@@ -829,12 +829,22 @@ TransactionFrame::applyOperations(SignatureChecker& signatureChecker,
         CLOG_ERROR(Tx, "Exception while applying operations ({}, {}): {}",
                    xdr_to_string(getFullHash(), "fullHash"),
                    xdr_to_string(getContentsHash(), "contentsHash"), e.what());
+        if (app.getConfig().HALT_ON_INTERNAL_TRANSACTION_ERROR)
```
MonsieurNicolas, line 832: no need to duplicate this code: you can just do it later in that function

MonsieurNicolas (changes requested):

src/main/Config.cpp
dmkozh: Updated the config. I've been thinking about updating it after this change is in to prevent the mismatch between docs and binaries, but I suppose the scenario of someone just copying the example config at head is unlikely.

src/main/Config.h
dmkozh: Done. I still see the value of documenting this in code too, please let me know if you think this is unneeded.

src/transactions/TransactionFrame.cpp
dmkozh: Done.

jonjove (approved):

MonsieurNicolas:
r+ e8067a9ea481ae6dd0f394641014c458c4a5f0fa

PR #3336, opened 2022-02-08
Update the `quorum` endpoint logic

### LINKED ISSUE #2558: improve quorum endpoint to deal with a slot that didn't externalize yet
Looking at `BallotProtocol::getJsonQuorumInfo`

There are a few issues when the node didn't externalize yet:
  * it's going to report all nodes in its quorum set as "delayed"
  * "disagree" may not be accurate (this could be fine to leave it as is - as this would be temporary until the next timeout)
  * as the logic is implemented based on what the ballot protocol is doing, it's going to be very inaccurate during nomination

hidenori-shinohara:
# Description

Resolves #2558.

Update the `quorum` endpoint logic to improve the accuracy by checking the state of the node in both the ballot and nomination protocol in (possibly) multiple slots. For more details, see https://github.com/stellar/stellar-core/issues/2558#issuecomment-1028456974

# Checklist
- [x] Compiles
- [x] Ran all tests


src/scp/NominationProtocol.cpp
```diff
@@ -612,6 +617,64 @@ NominationProtocol::getJsonInfo()
...
+NominationProtocol::getState(NodeID const& n, bool checkingForNewerSlot)
+{
+    if (n == mSlot.getLocalNode()->getNodeID())
+    {
+        // always mark myself as IN_PROGRESS.
+        return SCP::QuorumInfoNodeState::IN_PROGRESS;
+    }
+    auto state = SCP::QuorumInfoNodeState::IN_PROGRESS;
+
+    bool enoughTimeHasPassed = mTimerExpCount >= 2 || checkingForNewerSlot;
```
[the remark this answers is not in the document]
hidenori-shinohara: Done!

src/scp/NominationProtocol.cpp
```diff
@@ -612,6 +617,64 @@ NominationProtocol::getJsonInfo()
     return ret;
 }
 
+SCP::QuorumInfoNodeState
+NominationProtocol::getState(NodeID const& n, bool checkingForNewerSlot)
```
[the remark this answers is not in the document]
hidenori-shinohara: Hmm I'm not sure I agree. I feel that we'd still be telling the same information to the nomination protocol (and the ballot protocol). Also, in the ballot protocol, we use this flag when checking for `DELAYED` in which case we don't check whether the timer has expired.

```
if (I externalized & they haven't externalized & checkingForNewerSlot)
{
    // mark the node as delayed as we've already moved onto a new slot
}
```
would be rather confusing if we used `!shouldCheckExpiredTimer` instead.

src/scp/SCP.h
```diff
@@ -70,8 +70,18 @@ class SCP
...
+        NO_INFO = 0,
+        IN_PROGRESS = 1,
+        MISSING = 2,
+        DELAYED = 3,
+        DISAGREE = 4
+    };
+
     // summary: only return object counts
-    // index = 0 for returning information for all slots
+    // index = 0 for returning information for the latest known slot.
```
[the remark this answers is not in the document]
hidenori-shinohara: We chatted offline, but for future reference:

This comment is incorrect and the code in master only returns information for the latest known slot, and we indeed call the function with `index = 0` (e.g., for the quorum endpoint). So I decided to fix the comment so it's consistent with what the code already does.

https://github.com/stellar/stellar-core/blob/76124d3eb0d3d21cdee754d2405c24a515b1e3ee/src/scp/SCP.cpp#L124-L147

docs/software/commands.md
```diff
@@ -194,6 +194,11 @@ format.
   If `compact` is set, only returns a summary version.
 
   If `fullkeys` is set, outputs unshortened public keys.
+  The quorum endpoint categorizes each node as following:
+  * `missing`: missing from the current slot and the previous slot.
```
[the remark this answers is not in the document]
hidenori-shinohara: Yup, that makes sense!

src/scp/BallotProtocol.cpp
```diff
@@ -2009,6 +2011,63 @@ BallotProtocol::getJsonInfo()
...
+    }
+    else if (mCurrentBallot)
+    {
+        auto b = mCurrentBallot->getBallot();
+        auto& st = it->second->getStatement();
+        auto t = st.pledges.type();
+        if (mPhase == SCP_PHASE_EXTERNALIZE &&
+            (t != SCPStatementType::SCP_ST_EXTERNALIZE ||
+             (t == SCPStatementType::SCP_ST_CONFIRM &&
+              st.pledges.confirm().ballot.counter == UINT32_MAX)) &&
```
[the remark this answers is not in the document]
hidenori-shinohara: Thanks for catching it!

src/scp/BallotProtocol.cpp
```diff
@@ -2009,6 +2011,63 @@ BallotProtocol::getJsonInfo()
...
+            !areBallotsCompatible(getWorkingBallot(st), b) &&
+            t != SCPStatementType::SCP_ST_PREPARE)
+        {
+            // n has accepted to commit a different value than mine!
+            // Even if this node has been marked something else,
+            // we will overwrite it since this is bad.
+            state = SCP::QuorumInfoNodeState::DISAGREE;
+        }
+    }
+    else
```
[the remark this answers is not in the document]
hidenori-shinohara: Makes sense! I removed the comment since I felt that it was stating the obvious.

src/scp/BallotProtocol.cpp
```diff
@@ -2009,6 +2011,64 @@ BallotProtocol::getJsonInfo()
...
+            state = SCP::QuorumInfoNodeState::MISSING;
+        }
+        else
+        {
+            // It's too soon to start calling this node
+            // MISSING.
+            state = SCP::QuorumInfoNodeState::NO_INFO;
+        }
+    }
+    else if (mCurrentBallot)
```
marta-lokhova, line 2038: hmm, should we only use local state that's been fully validated? Otherwise you might be potentially comparing slots with bogus votes, and I don't think that's correct

src/scp/NominationProtocol.cpp
```diff
@@ -612,6 +617,65 @@ NominationProtocol::getJsonInfo()
...
+{
+    if (n == mSlot.getLocalNode()->getNodeID())
+    {
+        // always mark myself as IN_PROGRESS.
+        return SCP::QuorumInfoNodeState::IN_PROGRESS;
+    }
+    auto state = SCP::QuorumInfoNodeState::IN_PROGRESS;
```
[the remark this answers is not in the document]
hidenori-shinohara: Fixed!

src/scp/NominationProtocol.h
```diff
@@ -122,5 +125,14 @@ class NominationProtocol
...
     SCPEnvelope const* getLatestMessage(NodeID const& id) const;
+
+  private:
+    // The number of times the timer has expired
+    // Used for the quorum endpoint.
+    int32 mTimerExpCount;
+
+    // The threshold of the number of times the timer has to expired
+    // This is used for the quorum endpoint.
+    int32 const NUM_TIMEOUTS_THRESHOLD = 2;
```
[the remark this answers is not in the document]
hidenori-shinohara: Fixed

src/scp/SCP.cpp
```diff
@@ -120,28 +120,93 @@ SCP::getJsonInfo(size_t limit, bool fullKeys)
     return ret;
 }
 
+SCP::QuorumInfoNodeState
+SCP::getState(NodeID const& node, uint64 slotIndex, int N = 2)
```
MonsieurNicolas, line 124: don't use this N=2 (I would remove this argument instead): this function is only called from one place and that syntax is bogus anyways and should be removed : I am not even sure why this compiles (default arguments should be defined in the signature ie the ".h" file).

MonsieurNicolas, line 124: `N` does not follow our coding standard for variable names

src/scp/SCP.cpp
```diff
@@ -120,28 +120,93 @@ SCP::getJsonInfo(size_t limit, bool fullKeys)
...
+{
+    for (int k = 0; k < N; k++)
+    {
+        if (slotIndex - k <= 0)
```
MonsieurNicolas, line 128: check should be written as `slotIndex <= k` (result is unsigned so what you did is invalid)

src/scp/SCP.cpp
```diff
@@ -120,28 +120,93 @@ SCP::getJsonInfo(size_t limit, bool fullKeys)
...
+    }
     if (index == 0)
     {
-        for (auto& item : mKnownSlots)
-        {
-            auto& slot = *item.second;
-            ret = slot.getJsonQuorumInfo(id, summary, fullKeys);
-            ret["ledger"] = static_cast<Json::UInt64>(slot.getSlotIndex());
-        }
+        index = mKnownSlots.rbegin()->first;
```
MonsieurNicolas, line 159: Interesting the old code was looping over all slots only to keep the last result... I am pretty sure that was not what we intended to do originally (was returning all slots) - and indeed I found that https://github.com/stellar/stellar-core/pull/2127 made that change.
I think this is fine. Funny that it took 3 years to notice.

src/scp/SCP.cpp
```diff
@@ -120,28 +120,93 @@ SCP::getJsonInfo(size_t limit, bool fullKeys)
...
+        int n_disagree = 0, n_missing = 0, n_delayed = 0, n_in_progress = 0;
+
+        // Categorize each node's state.
+        LocalNode::forAllNodes(qSet, [&](NodeID const& n) {
+            auto state = getState(n, index);
+            auto k = s->getSCPDriver().toStrKey(n, fullKeys);
+            switch (state)
+            {
+            case SCP::QuorumInfoNodeState::DISAGREE:
+                disagree.append(k);
```
MonsieurNicolas, line 180: need to only append to those things when `summary` is false

src/scp/SCP.cpp
```diff
@@ -120,28 +120,93 @@ SCP::getJsonInfo(size_t limit, bool fullKeys)
...
+    auto s = getSlot(index, false);
+    if (s)
     {
-        auto s = getSlot(index, false);
-        if (s)
+        ret = s->getJsonQuorumInfo(id, summary, fullKeys);
+        auto qSet = getLocalNode()->getQuorumSet();
+
+        Json::Value& disagree = ret["disagree"];
+        Json::Value& missing = ret["missing"];
```
MonsieurNicolas, line 168: need to update the documentation and examples in `admin.md`

src/scp/SCP.cpp
MonsieurNicolas: (I know you're overriding the value at the end, it's just weird and inefficient)

src/scp/SCP.cpp
```diff
@@ -120,28 +120,93 @@ SCP::getJsonInfo(size_t limit, bool fullKeys)
...
+        Json::Value& delayed = ret["delayed"];
+        Json::Value& in_progress = ret["in_progress"];
```
MonsieurNicolas, line 170: looking at other places where we use `agree`, we would have to update a bunch of downstream tooling/systems (like supercluster) to handle `in_progress` the same way. So I suggest to just keep using `agree` in json to just mean "in progress" (just need to update the documentation to explain that nuance).

MonsieurNicolas (changes requested):
looks pretty close to me

src/scp/BallotProtocol.cpp
```diff
@@ -515,6 +516,7 @@ BallotProtocol::stopBallotProtocolTimer()
 void
 BallotProtocol::ballotProtocolTimerExpired()
 {
+    mHasTimerExpired = true;
```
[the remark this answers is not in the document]
hidenori-shinohara: Updated this so they are the same as the nomination code. I also updated it to 2 timeouts for consistency.

src/scp/SCP.cpp
hidenori-shinohara: Removed. I turned this into a constant.

hidenori-shinohara: Removed.

src/scp/SCP.cpp
hidenori-shinohara: Fixed.

src/scp/SCP.cpp
hidenori-shinohara: Fixed.

src/scp/SCP.cpp
hidenori-shinohara: Updated `admin.md`.

src/scp/SCP.cpp
hidenori-shinohara: Fixed. I updated the documentation to clarify what this means.

src/scp/NominationProtocol.h
```diff
@@ -122,5 +125,14 @@ class NominationProtocol
...
     SCPEnvelope const* getLatestMessage(NodeID const& id) const;
+
+  private:
+    // The number of times the timer has expired
+    // Used for the quorum endpoint.
+    int32 mTimerExpCount;
+
+    // The number of times the timer has to expire before we consider the node
+    // missing. Used for reporting purposes only.
+    int32 const NUM_TIMEOUTS_THRESHOLD_FOR_REPORTING = 2;
```
MonsieurNicolas, line 136: why do you have the same constant defined in two places? Just move it to "Slot"

docs/software/admin.md
```diff
@@ -877,8 +877,8 @@ This output has two main sections: `qset` and `transitive`. The former describes
 
 Entries to watch for in the `qset` section -- describing the node and its quorum set -- are:
 
-  * `agree` : the number of nodes in the quorum set that agree with this instance.
-  * `delayed` : the nodes that are participating to consensus but seem to be behind.
+  * `agree` : the number of nodes in the quorum set that seem to be up and running as expected. The local node has no reason to believe that this node is `delayed`, `disagree` or `missing`. Note that `agree` has nothing to do with SCP terms such as "accept" or "confirming".
+  * `delayed` : the nodes that are participating to consensus in this round but seem to be behind.
```
MonsieurNicolas, line 881: forgot to update the definition of other fields like "missing" (it says "this consensus round")

MonsieurNicolas (changes requested):
you can make the minor updates requested and squash+rebase to prepare for merging

src/scp/NominationProtocol.h
hidenori-shinohara: Moved!

docs/software/admin.md
hidenori-shinohara: Updated it!

MonsieurNicolas:
r+ bf43d33a39958d45a4ee2462e12ef698873818fb

PR #3338, opened 2022-02-09
Rename isString32Valid to isStringValid

leighmcculloch:
# Description

Rename isString32Valid to isStringValid.

isString32Valid is named with 32 in the name which implies that it is
validating 32 byte strings, but it is used to validate strings of both 32 bytes
and 64 bytes in length. For example, it is used to validate 32 byte home
domains, and 64 byte data entry names. When perusing the code this initially
confused me and I thought the data entry names must also be 32 bytes. On closer
inspection I discovered that the isString32Valid does not check if the string
is valid for the string32 type as specified in the Stellar XDR, but rather that
the string is valid for specific string fields in the XDR, namely the home
domain and the data entry name.

It would be less misleading if this function didn't specify the string size in
the name when it is not relevant to the functions behavior or intended scope of
use.

# Checklist
- [x] Compiles
- [x] Ran all tests
- [ ] ~If change impacts performance, include supporting evidence per the [performance document](https://github.com/stellar/stellar-core/blob/master/performance-eval/performance-eval.md)~

Note: I ran `make format` with `clang-format version 13.0.1` and it modified hundreds of files, so I discarded those changes and have not run format. I'm unable to install clang-format v8.0.0 on macOS, so I'm not sure if that is the cause of the difference.

### DIFF
--- a/src/invariant/LedgerEntryIsValid.cpp
+++ b/src/invariant/LedgerEntryIsValid.cpp
@@ -151,7 +151,7 @@ LedgerEntryIsValid::checkIsValid(AccountEntry const& ae, uint32 version) const
         return "Account flags are invalid";
     }
 
-    if (!isString32Valid(ae.homeDomain))
+    if (!isStringValid(ae.homeDomain))
     {
         return "Account homeDomain is invalid";
     }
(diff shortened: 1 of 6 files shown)


MonsieurNicolas:
r+ 426ca7a22f3f733fb655deec8842ce522b3da620

PR #3342, opened 2022-02-15
Make validators from the same home domain preferred peers

MonsieurNicolas:
This PR implements a best practice for entities that operate multiple validators by upgrading the peer type for other validators in the same home domain from just "known" to "preferred" peers.

This allows an organization to better take advantage of their internal bandwidth while providing better guarantees to other nodes on the network as they will receive an organization's SCP messages in batch.

### DIFF
--- a/src/main/Config.cpp
+++ b/src/main/Config.cpp
@@ -528,6 +528,7 @@ Config::parseValidators(
         ValidatorEntry ve;
         std::string pubKey, hist;
         bool qualitySet = false;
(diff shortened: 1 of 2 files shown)


marta-lokhova:
r+ 70ecf7d3dd8a2b016c57107c9ae30cfabe952825

marta-lokhova:
r+ 57412e7126f18990be24955132a62ffeda383f52

PR #3343, opened 2022-02-15
Actually enforce SHA256 contract

MonsieurNicolas:
This change actually ensures that we don't make broken calls to libsodium when using the SHA256 class.

I did a code review of the call sites, and right now we don't make such broken calls so this is hardening "just in case".


graydon:
r+ 764efee

PR #3345, opened 2022-02-16
rust/C++ bridge

graydon:
This PR contains a bridge to support Rust code in stellar-core, as well as some example use of that bridge. It's based on the [cxx](cxx.rs) project, which is a relatively high-level bridging system that allows a subset of both C++ and Rust semantics to be shown to one another directly, without the user having to drop down to the C ABI connecting the two. It's worth reading the cxx website for an explanation.

The bridge is gated behind `--enable-next-protocol-version-unsafe-for-production`, on the assumption that whenever "next" occurs, it's going to have at least _some_ Rust code in it.

To get this to work, you need to install `rustc` and `cargo`, obviously, and you have to pass `--enable-next-protocol-version-unsafe-for-production` to configure. It will install `cxx` and the `cxxbridge` tool that it uses locally, in the stellar-core build dir. All this PR does when Rust is enabled is route some base64 encode/decode functions over the bridge to Rust implementations. It's just a basic smoke test.

Notably: this PR does not switch to building the whole program by cargo or anything. Cargo runs in the `src/rust` subdirectory only, and the usual autotools-driven (or msbuild-driven) build remains the primary, outer structure. When make sees a rust file has changed, it re-runs cargo, as well as `cxxbridge`, regenerating the bridge.


INSTALL.md
```diff
@@ -156,3 +156,22 @@ The GUI depends on the `capstone`, `freetype` and `glfw` libraries and their hea
 
     # On MacOS
     $ brew install capstone freetype2 glfw
+
+## Building with Rust
+
+Configuring with `--enable-rust` will build and embed components written in the [Rust](https://rust-lang.org) programming language. These components are currently optional and experimental.
```
MonsieurNicolas, line 162: Are there `CFLAGS` equivalent that need to be passed in/initialized so that symbols etc are created properly?

graydon: I'm not sure what you mean by symbols being created, but you shouldn't have to adjust CFLAGS.

None of this is wired up on windows yet at all. I can try that now if you like?

MonsieurNicolas: what I meant is that we have default optimization flags for C/C++ and I don't know how Rust configures those things, I imagine that there is some equivalent

MonsieurNicolas: re: Windows, we'll have to do that as some point yes but I don't think we need to block on it for this PR.

Now that I think of it (related):
we should remove the Github action stuff for now and we should make the appropriate changes when we actually take the Rust dependency (either for all builds or just for "protocol next" depending on where we'll start bringing in Rust), and when we do this we'll want to have Windows figured out for sure (to avoid having to manage many targets - right now Windows only allows to build "protocol next" for that reason).

jonjove (changes requested):
Looks good overall and worked easily for me. Very cool. I was able to trigger many of the toolchain version checks suggesting that they work correctly. Only big issue is in the actual base64 code.

src/rust/src/lib.rs
```diff
@@ -0,0 +1,31 @@
...
+fn to_base64(b: &CxxVector<u8>, mut s: Pin<&mut CxxString>)
+{
+    s.as_mut().clear();
+    s.push_str(base64::encode(b.as_slice()).as_str());
+}
+
+fn from_base64(s: &CxxString, mut b: Pin<&mut CxxVector<u8>>)
+{
+    assert!(b.len() == 0);
+    let ss = s.to_str().expect("non-str base64 string");
```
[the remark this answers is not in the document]
MonsieurNicolas: I guess related to this, what's the interop story when it comes to errors and exceptions? Is it just panic all around (with the appropriate crash handler being called to log & exit)?

Should the bridge function exposed to c++ be marked "noexcept" (and end up calling "std::terminate" on panic somehow) and do we want bridge functions to convert errors into exceptions (for the non panic case)?

What about the other way around? Are we going to call C++ from Rust? If so, we should also enforce that whatever is called is "noexcept" on the C++ side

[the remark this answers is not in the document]
graydon: @jonjove  I've pushed an update that makes it "do what the C++ does" which is just silently eat malformed bytes in the base64 input. woo. we might want to make this actually strict at some point.

@MonsieurNicolas the default behaviour is to [map Rust's `Result<T>` type to and from C++ exceptions](https://cxx.rs/binding/result.html) -- both directions, so it'll catch C++ exceptions and turn them _into_ Rust `Result<T>` values too -- with a little bit of customization available. They're going to continue to work on this part of the bridge but since we do use exceptions it's probably close to ok for our needs now.

I just, in this case, didn't return a Result because the existing C++ code had no error-handling paths. I assumed this was because we sanitized its input, but it's actually because it's error tolerant (perhaps unwisely, but ..)

INSTALL.md
graydon: Added a windows sketch which gets far enough along that it passes the round-trip tests between C++ and rust. It's a little ugly and I could use some help with it, I think it's probably not 100% right structurally, I think it rebuilds too often, and also even if working .. it has some limitations (eg. can't make a debug profile that uses it, as rust and/or cargo can only use the release CRT when linking statically; maybe the trick is to go dynamic in the debug config?)

I'll remove the github actions stuff tomorrow I guess. I am a little hesitant to do so since it'll bitrot and break if not checked, but I guess we're not _quite_ ready to turn it on yet.

src/rust/Cargo.toml
```diff
@@ -0,0 +1,13 @@
+[package]
+name = "stellar-core"
+version = "0.1.0"
```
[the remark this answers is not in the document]
graydon: Unless we publish it on crates.io, it's not going to make much difference what version we call it.

graydon:
@MonsieurNicolas I've disabled the `--enable-rust` action from the github actions scenario. I'm not actually sure if it affects the github cache rates all that much, and there's enough breakable plumbing that we'll want to turn this on for real eventually -- but for the time being it's turned off, so I think ready to merge.

src/rust/src/lib.rs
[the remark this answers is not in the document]
jonjove: Opened #3374

INSTALL.md
graydon: Thanks for @MonsieurNicolas finding [this github comment](https://github.com/rust-lang/rust/issues/39016?msclkid=6b0248f6c5ac11ecbd155fd40b4d4f30#issuecomment-853964918) the debug build now also works on windows.

Currently there's only one Rust source file (the crate root) so the CustomBuild step can just include it as its AdditionalInputs source; this will get unwieldy if we want visual studio / msbuild to re-run cargo on any edit to _any_ Rust file. There's an even-more-involved technique we can use for this [explained here](https://www.reedbeta.com/blog/custom-toolchain-with-msbuild/) but it's not really clear to me that we're going to be doing edit-and-rebuild of Rust code from inside visual studio -- it doesn't have a Rust editing component currently, only does very modest syntax highlighting. No intellisense or anything, you have to use a different IDE for that. I can investigate that custom toolchain approach if we want to go further in that direction, but I think for now it's probably good enough to land.

graydon:
I believe this is ready to land now. Contains a test to confirm the base64 identity and is on by default on vNext.

.github/workflows/build.yml
```diff
@@ -51,7 +51,7 @@ jobs:
           sudo apt-get -y install --no-install-recommends apt-utils dialog git iproute2 procps lsb-release
       - name: install tool chain
         run: |
-          sudo apt-get -y install libstdc++-8-dev clang-format-10 ccache
+          sudo apt-get -y install libstdc++-8-dev clang-format-10 ccache cargo rustc
```
MonsieurNicolas, line 54: we should add the same to `.devcontainer/Dockerfile` and `docker/Dockerfile.testing`

MonsieurNicolas:
r+ f78f04faf2eaeb4d612e2960c04ea88fab2e9ecf

PR #3352, opened 2022-02-23
Improve overlay idle timeout

marta-lokhova:


src/overlay/test/OverlayTests.cpp
```diff
@@ -289,6 +289,42 @@ TEST_CASE("drop peers that dont respect capacity", "[overlay][flow control]")
     testutil::shutdownWorkScheduler(*app1);
 }
 
+TEST_CASE("drop idle flow-contolled peers", "[overlay][flow control]")
```
MonsieurNicolas, line 292: typo "controlled"

src/overlay/Peer.cpp
```diff
@@ -239,6 +240,13 @@ Peer::recurrentTimerExpired(asio::error_code const& error)
             drop("idle timeout", Peer::DropDirection::WE_DROPPED_REMOTE,
                  Peer::DropMode::IGNORE_WRITE_QUEUE);
         }
+        else if (flowControlEnabled() == Peer::FlowControlState::ENABLED &&
+                 (now - mLastSendMoreReceived) >= timeout)
```
MonsieurNicolas, line 244: Is this correct? If the rate of messages is low and the capacity is high, we may not receive "send more" requests.
Maybe what we should be tracking is how long a connection is stuck "at 0 capacity" instead?

MonsieurNicolas (changes requested):

src/overlay/Peer.cpp
marta-lokhova: yeah I think you're right: while it seems unlikely, it's possible, especially on testnet. I updated the change to start counting when the outbound capacity is 0 on the sender side. This way we don't depend on message rate, and also handle potential bugs in capacity accounting on the receiver side (for example, if it releases _some_ capacity, but not enough to trigger sending "send more").

MonsieurNicolas:
r+ 2c512dcc4697984c77a868d290ced4800b7c3aa7

PR #3361, opened 2022-03-04
Make dropping sqlite databases fast

### LINKED ISSUE #3353: sqlite: change "new-db" to first delete the sqlite files
I noticed that trying to run `new-db` on a pubnet node running with sqlite takes a long time as it tries to delete tables one by one via the WAL.

As we clear out the entire database anyways, we should just delete the sqlite files (wal, etc) directly (or perform some sqlite specific pragma to do it).

Postgresql does not seem to have that problem as it looks like they also just directly delete files associated with the deleted tables.

MonsieurNicolas:
# Description

Resolves #3353

### DIFF
--- a/src/database/Database.cpp
+++ b/src/database/Database.cpp
@@ -10,6 +10,7 @@
 #include "main/Config.h"
 #include "overlay/StellarXDR.h"
 #include "util/Decoder.h"
+#include "util/Fs.h"
 #include "util/GlobalChecks.h"
 #include "util/Logging.h"
 #include "util/Timer.h"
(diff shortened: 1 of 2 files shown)


lib/soci/src/backends/sqlite3/session.cpp
```diff
@@ -59,7 +59,8 @@ sqlite3_session_backend::sqlite3_session_backend(
     int connection_flags = SQLITE_OPEN_READWRITE | SQLITE_OPEN_CREATE;
     std::string synchronous;
     std::string const & connectString = parameters.get_connect_string();
-    std::string dbname(connectString);
+    std::string& dbname = db_name_;
```
jonjove, line 62: I don't think the changes in this commit are necessary, you can use https://www.sqlite.org/pragma.html#pragma_database_list instead. I've tested it locally from the `sqlite3` command line interface, and it does return the path to the connected database.

I'm not _strongly_ opposed to making this change, but I do think it's better to avoid changing dependencies whenever possible to make it easier to upgrade in the future.

MonsieurNicolas: ah yes you are right - I missed it when I looked at the list of pragmas. Updated to use it instead

jonjove:
r+ 01e2cf0

PR #3380, opened 2022-03-19
WASM smart contracts experiment

graydon:
This is a **preliminary, draft experiment** with embedding a WASM VM in stellar-core, in order to execute WASM smart contracts, along with ledger entry types for storing contract code and data, a transaction type for invoking contracts, and a certain amount of plumbing to manage lifecycle and interconversion of values between XDR and "host" object formats.

This experiment takes a minimalist "uniform representation" approach where smart contracts never (or rarely) perform any serialization or deserialization in their code / on the VM, nor even hold large structured data values themselves. Rather they interact through uniform i64 values, some of which hold _references_ to host objects -- small-integer handle values. The goal is to keep the per-contract code size as low as possible.  Ideally, most contracts won't do any VM-side memory allocation at all (which is good, because in WASM if you want malloc you have to literally embed a copy of malloc in your code).

The WASM VM used here is [fizzy](https://github.com/wasmx/fizzy) which is a deliberately-simple (but relatively fast) interpreter. It may or may not be the right long-term choice. There are several others to choose from, this was just a tidy and easy-to-integrate codebase with (evidently) a focus on pedantic correctness.

The PR also includes the [immer](https://github.com/arximboldi/immer) library for immutable datatypes. This is to simplify contract code (and contract security) and enable transactional semantics on host objects: we present _only immutable objects_ through the host-object handle interface. For example, adding an entry to a map takes the old map handle, a key and value, and returns a new map handle with the new mapping in it. The immer library uses data structures from the functional language world (HAMTs and RRBs) to make this efficient on maps and vectors, by having structured values share substructures.

There is currently no interface to adding contracts, and only a single unit test that invokes one (which is compiled from some zig code -- turns out zig has a fairly good wasm backend that produces small binaries). There is also not yet an interface to the new ledger entries, or really much of any host functions.

I'm opening this now so that others can collaborate on design and experimentation. This is not a finished or even half-finished sketch, much less a CAP or final design. It is just an experiment that's now far enough along to run tests and use for prototyping in various other directions (SDKs, host objects, figuring out storage and inter-contract calling, etc. etc.)


graydon:
Closed in favor of https://github.com/stellar/stellar-core/pull/3428

PR #3381, opened 2022-03-21
Loadgen: better recovery for dropped transactions

marta-lokhova:
Noticed that our loadgen recovery mechanism for dropped transactions isn't working correctly. It broke after we introduced banned transactions. Specifically, before we would re-submit the same transaction that got dropped, but it would also be rejected because we ban dropped transactions. With this change, whenever we resubmit transactions to recover, we pick a different destination account, which allows the transaction to go through.


MonsieurNicolas:
r+ bb3a70829fcac55a416a90fbb8f0966ea0970d9a

PR #3391, opened 2022-04-04
Remove mutable result from TransactionFrame into test-only classes

jayz22:
# Description

Resolves "TransactionFrame" part of #2163
Refactor for `TransactionFrame` class-family (including `TransactionFrame`, `TransactionFrameBase`, `FeeBumpTransactionFrame`, `Fuzzer*TransactionFrame`, `TxSim*TransactionFrame`). 
Make the class immutable by removing the member `TransactionResult`, instead pass them externally. 
Separate test code from production code by creating `TestTransactionFrame` and `TestFeeBumpTransactionFrame` that contains the old (mutable) behavior. This is to save from changing boilerplate code in tests everywhere that interacts with `TransactionFrameBase` (and its derived classes). 

# Checklist
- [x] Compiles
- [x] Ran all tests

### DIFF
--- a/src/herder/Herder.h
+++ b/src/herder/Herder.h
@@ -117,7 +117,7 @@ class Herder
     virtual bool recvTxSet(Hash const& hash, TxSetFrame const& txset) = 0;
     // We are learning about a new transaction.
     virtual TransactionQueue::AddResult
-    recvTransaction(TransactionFrameBasePtr tx) = 0;
+    recvTransaction(TransactionFrameBasePtr tx, TransactionResult& txRes) = 0;
     virtual void peerDoesntHave(stellar::MessageType type,
                                 uint256 const& itemID, Peer::pointer peer) = 0;
     virtual TxSetFramePtr getTxSet(Hash const& hash) = 0;
--- a/src/herder/TransactionQueue.cpp
+++ b/src/herder/TransactionQueue.cpp
@@ -186,7 +186,7 @@ isDuplicateTx(TransactionFrameBasePtr oldTx, TransactionFrameBasePtr newTx)
 }
 
 TransactionQueue::AddResult
-TransactionQueue::canAdd(TransactionFrameBasePtr tx,
+TransactionQueue::canAdd(TransactionFrameBasePtr tx, TransactionResult& txRes,
                          AccountStates::iterator& stateIter,
                          TimestampedTransactions::iterator& oldTxIter)
 {
@@ -236,7 +236,7 @@ TransactionQueue::canAdd(TransactionFrameBasePtr tx,
             {
                 if (!findBySeq(tx, transactions, oldTxIter))
                 {
-                    tx->getResult().result.code(txBAD_SEQ);
+                    txRes.result.code(txBAD_SEQ);
                     return TransactionQueue::AddResult::ADD_STATUS_ERROR;
                 }
 
@@ -252,8 +252,8 @@ TransactionQueue::canAdd(TransactionFrameBasePtr tx,
                     int64_t minFee;
                     if (!canReplaceByFee(tx, oldTxIter->mTx, minFee))
                     {
-                        tx->getResult().result.code(txINSUFFICIENT_FEE);
-                        tx->getResult().feeCharged = minFee;
+                        txRes.result.code(txINSUFFICIENT_FEE);
+                        txRes.feeCharged = minFee;
                         return TransactionQueue::AddResult::ADD_STATUS_ERROR;
                     }
 
@@ -295,8 +295,8 @@ TransactionQueue::canAdd(TransactionFrameBasePtr tx,
         ban({tx});
         if (canAddRes.second != 0)
         {
-            tx->getResult().result.code(txINSUFFICIENT_FEE);
-            tx->getResult().feeCharged = canAddRes.second;
+            txRes.result.code(txINSUFFICIENT_FEE);
+            txRes.feeCharged = canAddRes.second;
             return TransactionQueue::AddResult::ADD_STATUS_ERROR;
         }
         return TransactionQueue::AddResult::ADD_STATUS_TRY_AGAIN_LATER;
@@ -320,7 +320,7 @@ TransactionQueue::canAdd(TransactionFrameBasePtr tx,
             mApp.getLedgerManager().getLastClosedLedgerNum() + 1;
     }
     if (!tx->checkValid(ltx, seqNum, 0,
-                        getUpperBoundCloseTimeOffset(mApp, closeTime)))
+                        getUpperBoundCloseTimeOffset(mApp, closeTime), txRes))
     {
         return TransactionQueue::AddResult::ADD_STATUS_ERROR;
     }
@@ -334,7 +334,7 @@ TransactionQueue::canAdd(TransactionFrameBasePtr tx,
                             : feeStateIter->second.mTotalFees;
     if (getAvailableBalance(ltx.loadHeader(), feeSource) - netFee < totalFees)
     {
-        tx->getResult().result.code(txINSUFFICIENT_BALANCE);
+        txRes.result.code(txINSUFFICIENT_BALANCE);
         return TransactionQueue::AddResult::ADD_STATUS_ERROR;
     }
 
@@ -481,12 +481,12 @@ TransactionQueue::findAllAssetPairsInvolvedInPaymentLoops(
 }
 
 TransactionQueue::AddResult
-TransactionQueue::tryAdd(TransactionFrameBasePtr tx)
+TransactionQueue::tryAdd(TransactionFrameBasePtr tx, TransactionResult& txRes)
 {
     ZoneScoped;
     AccountStates::iterator stateIter;
     TimestampedTransactions::iterator oldTxIter;
-    auto const res = canAdd(tx, stateIter, oldTxIter);
+    auto const res = canAdd(tx, txRes, stateIter, oldTxIter);
     if (res != TransactionQueue::AddResult::ADD_STATUS_PENDING)
     {
         return res;
--- a/src/ledger/LedgerManagerImpl.cpp
+++ b/src/ledger/LedgerManagerImpl.cpp
@@ -686,15 +686,17 @@ LedgerManagerImpl::closeLedger(LedgerCloseData const& ledgerData)
     // was sorted by hash; we reorder it so that transactions are
     // sorted such that sequence numbers are respected
     vector<TransactionFrameBasePtr> txs = ledgerData.getTxSet()->sortForApply();
+    vector<TransactionResult> txResults(txs.size());
 
     // first, prefetch source accounts for txset, then charge fees
     prefetchTxSourceIds(txs);
     auto curBaseFee = txSet->getBaseFee(header.current());
-    processFeesSeqNums(txs, ltx, curBaseFee, ledgerCloseMeta);
+    processFeesSeqNums(txs, txResults, ltx, curBaseFee, ledgerCloseMeta);
 
     TransactionResultSet txResultSet;
     txResultSet.results.reserve(txs.size());
-    applyTransactions(txs, ltx, txResultSet, ledgerCloseMeta, curBaseFee);
+    applyTransactions(txs, txResults, ltx, txResultSet, ledgerCloseMeta,
+                      curBaseFee);
 
     ltx.loadHeader().current().txSetResultHash = xdrSha256(txResultSet);
 
@@ -1047,7 +1049,8 @@ mergeOpInTx(std::vector<Operation> const& ops)
 
 void
 LedgerManagerImpl::processFeesSeqNums(
-    std::vector<TransactionFrameBasePtr>& txs, AbstractLedgerTxn& ltxOuter,
+    std::vector<TransactionFrameBasePtr>& txs,
+    std::vector<TransactionResult>& txResults, AbstractLedgerTxn& ltxOuter,
     int64_t baseFee, std::unique_ptr<LedgerCloseMeta> const& ledgerCloseMeta)
 {
     ZoneScoped;
@@ -1064,7 +1067,7 @@ LedgerManagerImpl::processFeesSeqNums(
         for (auto tx : txs)
         {
             LedgerTxn ltxTx(ltx);
-            tx->processFeeSeqNum(ltxTx, baseFee);
+            tx->processFeeSeqNum(ltxTx, baseFee, txResults.at(index));
 
             if (protocolVersionStartsFrom(
                     ltxTx.loadHeader().current().ledgerVersion,
@@ -1179,7 +1182,8 @@ LedgerManagerImpl::prefetchTransactionData(
 
 void
 LedgerManagerImpl::applyTransactions(
-    std::vector<TransactionFrameBasePtr>& txs, AbstractLedgerTxn& ltx,
+    std::vector<TransactionFrameBasePtr>& txs,
+    std::vector<TransactionResult>& txResults, AbstractLedgerTxn& ltx,
     TransactionResultSet& txResultSet,
     std::unique_ptr<LedgerCloseMeta> const& ledgerCloseMeta, int64 curBaseFee)
 {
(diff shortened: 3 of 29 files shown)


src/herder/Herder.h
```diff
@@ -116,8 +116,12 @@ class Herder
                                   SCPQuorumSet const& qset) = 0;
     virtual bool recvTxSet(Hash const& hash, TxSetFrame const& txset) = 0;
     // We are learning about a new transaction.
+    virtual TransactionQueue::AddResult
+    recvTransaction(TransactionFrameBasePtr tx, TransactionResult& txRes) = 0;
+#ifdef BUILD_TESTS
```
dmkozh, line 121: General thought: test-only code in the public interface is typically a bad smell, as it both adds maintenance burden to the non-test implementers, as well as gets around the purpose of the tests (which normally should cover the public interface used by the real, non-test class clients).

For example, here it seems like we just use this method about 10 times in the tests, would adding a dummy output `txRes` for these tests work (instead of re-using the old method)?

src/herder/TransactionQueue.cpp
```diff
@@ -473,6 +473,14 @@ TransactionQueue::tryAdd(TransactionFrameBasePtr tx)
     return res;
 }
 
+#ifdef BUILD_TESTS
+TransactionQueue::AddResult
```
dmkozh, line 477: This seems to be used in just a couple of test cases. Would it be possible to cover the new interface instead?

src/test/FuzzerImpl.cpp
```diff
@@ -984,7 +985,8 @@ applyFuzzOperations(LedgerTxn& ltx, PublicKey const& sourceAccount,
 {
     auto txFramePtr = createFuzzTransactionFrame(ltx, sourceAccount, begin, end,
                                                  app.getNetworkID());
-    txFramePtr->attemptApplication(app, ltx);
+    TransactionResult txRes;
```
dmkozh, line 988: Since the result is not used anywhere, would it make sense to just allocate `txRes` in `attemptApplication`?

src/transactions/TransactionFrameBase.h
```diff
@@ -56,8 +54,23 @@ class TransactionFrameBase
     insertKeysForFeeProcessing(UnorderedSet<LedgerKey>& keys) const = 0;
     virtual void insertKeysForTxApply(UnorderedSet<LedgerKey>& keys) const = 0;
 
-    virtual void processFeeSeqNum(AbstractLedgerTxn& ltx, int64_t baseFee) = 0;
+    virtual void processFeeSeqNum(AbstractLedgerTxn& ltx, int64_t baseFee,
+                                  TransactionResult& txResult) = 0;
 
     virtual StellarMessage toStellarMessage() const = 0;
+
+#ifdef BUILD_TESTS
```
dmkozh, line 62: I haven't examined each usage of this, but it seems like you create test tx frames using the special method (`makeTestTransactionFromWire`). Would it be possible then to not have these methods in the public interface and only declare/implement them in the respective test-only classes?

src/transactions/TransactionFrame.cpp
```diff
@@ -375,41 +376,48 @@ TransactionFrame::commonValidPreSeqNum(AbstractLedgerTxn& ltx, bool chargeFee,
         (protocolVersionStartsFrom(ledgerVersion, ProtocolVersion::V_13) &&
          mEnvelope.type() == ENVELOPE_TYPE_TX_V0))
     {
-        getResult().result.code(txNOT_SUPPORTED);
+        // getResult().result.code(txNOT_SUPPORTED);
```
dmkozh, line 379: Please remove the commented out code in this method.

src/transactions/TransactionFrame.cpp
```diff
@@ -735,25 +747,27 @@ TransactionFrame::insertKeysForTxApply(UnorderedSet<LedgerKey>& keys) const
 }
 
 void
-TransactionFrame::markResultFailed()
+TransactionFrame::markResultFailed(TransactionResult& txResult)
```
dmkozh, line 750: Does this still need to be a member method of `TransactionFrame`? Seems like could be just a free internal function.

src/transactions/TransactionFrame.cpp
```diff
@@ -300,28 +300,28 @@ TransactionFrame::makeOperation(Operation const& op, OperationResult& res,
 
 void
 TransactionFrame::resetResults(LedgerHeader const& header, int64_t baseFee,
```
dmkozh, line 302: Seems like this method still mutates `TransactionFrame` (it has `mOperations.clear()` call). Not sure if that's in the scope of this PR, just wanted to make sure this doesn't escape our attention.

src/transactions/test/TxEnvelopeTests.cpp
```diff
@@ -2031,6 +2031,9 @@ TEST_CASE_VERSIONS("txenvelope", "[tx][envelope]")
 
                     REQUIRE(tx1->getResultCode() == txSUCCESS);
                     REQUIRE(tx2->getResultCode() == txFAILED);
+                    OperationResult res = getFirstResult(*tx2);
```
dmkozh, line 2034: This seems like some leftover change, is this intentional?

jayz22: Yes, this was leftover from some debug printing statement, removed.

src/transactions/TransactionFrame.cpp
jayz22: Thanks for the catch, removed.

src/transactions/TransactionFrame.cpp
jayz22: Yes I also noticed that, I agree this is one place that needs attention.

src/transactions/TransactionFrame.cpp
jayz22: This is also being called in `FuzzTransactionFrame::attemptApplication`.

src/herder/TransactionQueue.cpp
jayz22: Yes, I agree that's better. Fixed.

jayz22: Actually I take that back. Although the `tryAdd` method is only being called in two places, one of them is in `TransactionQueueTest::add` which is then called by 100+ lines. In order to remove the test version of `tryAdd` (always passing `txRes` as arg), we need to change the `add` method and update all of its callers, which is the exact reason why we have the test-only method in the first place.

src/transactions/TransactionFrameBase.h
jayz22: That would be difficult, as many of the existing tests are calling those public methods via the base class pointer. To remove those test-only functions from the base class to the "Test" classes would require updating all (hundreds of) callers in the test code to pass in the additional parameters.

src/herder/Herder.h
jayz22: I agree with the general sentiment regarding the test-only code. Ideally, we shouldn't have them, since they potentially give rise to unit test logic diverging from actual (production) code, and it's harder to maintain (as you mentioned).
However, 1. we already have them in our current code base (this is not to give excuse for writing more of those) and 2. doing it in this PR saves a ton of boilerplate code change: 
The `TransactionFrameBase` class and its many implementations `TransactionFrame`, `FeeBumpTransactionFrame`, `Fuzzer*TransactionFrame` are pervasive in all the unit-test code, and many of them are currently accessing the `TransactionResult` via member variable, those callers would need to passed in external `TransactionResult` to the new interfaces. 
Although the `recvTransaction` method here is only called a couple dozen times, which makes good argument for not using test method here, I would lean towards having them since the main objective is to adapt tests in the bigger scope to the new interfaces without changing boilerplate code.

src/transactions/TransactionFrame.cpp
dmkozh: That's not a big issue, but I think in general it's better for classes to have less responsibilities when possible. Here it seems like the method no longer has anything to do with the `TransactionFrame` and could be either a public free function (given that it's reused), a static function or even removed altogether, given that it's one line and for some reason it's the only 'special' case for setting the result code (others seem to be just set in-place). In any case, that's up to your judgement.

jayz22: I do see your point.

jayz22: I am inclined to remove the method, since it is just one line of setting the result code, and we are already directly setting it for other kinds of code (other than `txFAILED`). The only thing I'm concerned is the comments in this function, which points out to some perf implication that caller might need to know. So I'm gonna keep this function for now and maybe think of a better place to put it later.

sisuresh (commented):
Nice work. This PR removes `mResult` from `TransactionFrame` without requiring many test changes, but it does mix in more `#ifdef BUILD_TESTS` code into non-test code, which isn't ideal. 

I think Nico had an idea to use a wrapper class in tests (instead of using inheritance) where we replace `TransactionFramePtr` with a class that holds the `TransactionFramePtr` and `txResult`. The wrapper could imitate the `TransactionFrame` interface to keep test changes to a minimum, and the `txResult` member can be passed in wherever you need This would allow you to remove the new `#ifdef BUILD_TESTS` blocks. I'm not sure yet if there are issues with this approach, but it's something we should discuss.

src/herder/Herder.h
jayz22: In that case, I agree with you both. I've made the change to remove the test-only `recvTransaction`.

src/herder/TransactionQueue.cpp
jayz22: Yes, you are absolutely right, not sure how I didn't see that. Fixed.

src/ledger/LedgerManagerImpl.cpp
```diff
@@ -1034,7 +1036,8 @@ LedgerManagerImpl::advanceLedgerPointers(LedgerHeader const& header,
 
 void
 LedgerManagerImpl::processFeesSeqNums(
-    std::vector<TransactionFrameBasePtr>& txs, AbstractLedgerTxn& ltxOuter,
+    std::vector<TransactionFrameBasePtr>& txs,
+    vector<TransactionResult>& txResults, AbstractLedgerTxn& ltxOuter,
```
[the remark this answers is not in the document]
jayz22: Yes, fixed.

src/ledger/LedgerManagerImpl.cpp
```diff
@@ -1045,10 +1048,10 @@ LedgerManagerImpl::processFeesSeqNums(
     {
         LedgerTxn ltx(ltxOuter);
         auto ledgerSeq = ltx.loadHeader().current().ledgerSeq;
-        for (auto tx : txs)
+        for (size_t i = 0; i < txs.size(); ++i)
         {
             LedgerTxn ltxTx(ltx);
-            tx->processFeeSeqNum(ltxTx, baseFee);
+            txs[i]->processFeeSeqNum(ltxTx, baseFee, txResults[i]);
```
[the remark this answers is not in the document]
jayz22: Thanks, this is a good catch. Also, I realize there already exists `index` that does what `i` does so I'm using that instead.

src/ledger/LedgerManagerImpl.cpp
```diff
@@ -1115,7 +1118,8 @@ LedgerManagerImpl::prefetchTransactionData(
 
 void
 LedgerManagerImpl::applyTransactions(
-    std::vector<TransactionFrameBasePtr>& txs, AbstractLedgerTxn& ltx,
+    std::vector<TransactionFrameBasePtr>& txs,
+    vector<TransactionResult>& txResults, AbstractLedgerTxn& ltx,
```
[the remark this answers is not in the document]
jayz22: Fixed.

src/ledger/LedgerManagerImpl.cpp
```diff
@@ -1143,20 +1147,22 @@ LedgerManagerImpl::applyTransactions(
...
     prefetchTransactionData(txs);
 
-    for (auto tx : txs)
+    for (size_t i = 0; i < numTxs; ++i)
     {
         ZoneNamedN(txZone, "applyTransaction", true);
         auto txTime = mTransactionApply.TimeScope();
         TransactionMeta tm(2);
+        auto& tx = txs[i];
+        auto& txRes = txResults[i];
```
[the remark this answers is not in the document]
jayz22: Fixed. 
Although I realize this code is not in a `try...catch` block like `processFeesSeqNums` is and if out-of-bound exception did occur we are not handling it. I wonder if we should catch and throw here.

src/herder/Herder.h
jayz22: This is fixed.

src/test/TestTransactionFrame.h
```diff
@@ -0,0 +1,137 @@
...
+// under the Apache License, Version 2.0. See the COPYING file at the root
+// of this distribution or at http://www.apache.org/licenses/LICENSE-2.0
+
+#include "transactions/FeeBumpTransactionFrame.h"
+#include "transactions/TransactionFrame.h"
+
+namespace stellar
+{
+
+class TestTransactionFrame : public TransactionFrame
```
MonsieurNicolas, line 13: This class can be reworked to avoid more copy/paste down the line: you can just add a member `std::shared_ptr<TransactionFrameBase> mTx` and derive from `TransactionFrameBase` -> you can implement a proxy pattern and update `mResult` as needed.

src/transactions/TransactionFrame.cpp
MonsieurNicolas: this is because the change in this PR is incomplete: if you look at `OperationFrame`, it has a `mResult` that gets updated in a bunch of places.
Like when "apply" is called -> it ends up calling `OperationFrame::doApply` that mutates the result with things like `innerResult().code(CLAIM_CLAIMABLE_BALANCE_DOES_NOT_EXIST);`
I think it's just that `doApply` is missing a parameter so that it uses the result passed to `TransactionFrame::apply` (same thing in other places like `doCheckValid`

MonsieurNicolas: After you remove `mResult from `OperationFrame` `mOperations` should be set once and be immutable

src/transactions/TransactionFrameBase.h
MonsieurNicolas: This is not entirely accurate.

See my other comment on `TransactionTestFrame`: I think that we should just use `TransactionTestFrame` in all test code instead of returning `TransactionFrameBase` or `TransactionFrame`. The change should be fairly transparent as we usually do things like
`auto b1 = root.create("b1", paymentAmount);` (that would return a `TransactionTestFrame` instead).

That way we can also move the other "mutators" from TransactionFrame, like `addSignature` that should only be used in test/simulation code.

MonsieurNicolas (changes requested):
Good start, but the changes are only partial. I put a few recommendations in the comments that should allow to get this one over the line

src/test/TestTransactionFrame.h
MonsieurNicolas: update: I am not even sure you should derive from `TransactionFrameBase`.
* test code needs the version of the methods that don't have a "result"
* the whole `syncResult` thing is misleading and does not catch all bugs if people are mixing using a local variable and using this class' member, so there is no real point in trying to keep those things in sync
* for the few places where you need a `TransactionFrameBase` , you can just get a `const` reference to `mTx`.

src/transactions/FeeBumpTransactionFrame.h
```diff
@@ -91,11 +84,50 @@ class FeeBumpTransactionFrame : public TransactionFrameBase
...
-    void processFeeSeqNum(AbstractLedgerTxn& ltx, int64_t baseFee) override;
-
+    void processFeeSeqNum(AbstractLedgerTxn& ltx, int64_t baseFee,
+                          TransactionResult& txResult) override;
     StellarMessage toStellarMessage() const override;
 
     static TransactionEnvelope
     convertInnerTxToV1(TransactionEnvelope const& envelope);
+
+#ifdef BUILD_TESTS
```
MonsieurNicolas, line 94: note: you don't need any of this we directly use `TestTransactionFrame` in test code

src/transactions/TransactionFrame.cpp
MonsieurNicolas: the reason this is incomplete and unsafe is that `OperationFrame::mResult` ends up _caching_ a reference to the Result that you use, this in turn can have catastrophic implications in production code if the the local variable goes out of scope before the TransactionFrame gets deleted.

src/transactions/TransactionFrameBase.h
```diff
@@ -27,11 +27,12 @@ class TransactionFrameBase
                             TransactionEnvelope const& env);
 
     virtual bool apply(Application& app, AbstractLedgerTxn& ltx,
-                       TransactionMeta& meta) = 0;
+                       TransactionMeta& meta, TransactionResult& txResult) = 0;
```
MonsieurNicolas, line 30: all methods exposed here should now be `const` (this would allow you to easily find remaining bugs related to mutating the object/caching)

MonsieurNicolas: If there are any non const methods left in derived classes, it probably means they belong to `TestTransactionFrame` (or there is something we missed outside of "result", in which case let me know!)

MonsieurNicolas (changes requested):
I added a few clarifying comments

PR #3392, opened 2022-04-05
Bump overlay version

marta-lokhova:
Bump overlay version in 19.0.0. Not bumping min version yet, currently it's 19 (=v18.2.0) and the next version corresponds to v18.4.0. We can bump the min version in 19.1.0


MonsieurNicolas:
r+ f2651e62f63f6a2ff1c786ee06d4dd05b8823b2c

PR #3393, opened 2022-04-06
Calculate and expose more individual peer info through `peers` endpoint

### LINKED ISSUE #3041: expose individual peer information tracked by overlay via the peers endpoint
Right now core expose aggregate overlay metrics (exported via the metrics endpoint) and computes per peer metrics (not exposed as metrics).

We should export those as part of the `peers` endpoint (maybe with a new flag if it's too verbose).

This would allow to get a sense of certain things like number of bytes transferred etc on a per peer basis when debugging.

From a quick look at the code, it looks like many metrics are not computed on a per peer basis (because this information is not useful in the "network survey" context), but should be exposed as part of this work.
For example, `getOverlayMetrics().mMessageDrop` doesn't have a per peer equivalent.

On that note, we may want to use some template helper of sorts to update both metric types at once instead of manually keeping metrics them in sync with code like
```
    getOverlayMetrics().mByteRead.Mark(byteCount);
    mPeerMetrics.mByteRead += byteCount;
```

hidenori-shinohara:
# Description

Resolves #3041

This PR calculates and exposes more metrics for each peer through the `peers` endpoint. The new output can be slightly verbose. If `compact=true`, the new output will be the same as the current output.

# Checklist
- [x] Compiles
- [x] Ran all tests

### DIFF
--- a/docs/software/commands.md
+++ b/docs/software/commands.md
@@ -179,9 +179,10 @@ format.
   Clear metrics for a specified domain. If no domain specified, clear all
   metrics (for testing purposes).
 
-* **peers?[&fullkeys=false]**
-  Returns the list of known peers in JSON format.
+* **peers?[&fullkeys=false&compact=true]**
+  Returns the list of known peers in JSON format with some metrics.
   If `fullkeys` is set, outputs unshortened public keys.
+  If `compact` is `false`, it will output extra metrics.
 
 * **quorum**
   `quorum?[node=NODE_ID][&compact=false][&fullkeys=false][&transitive=false]`<br>
--- a/src/overlay/Peer.cpp
+++ b/src/overlay/Peer.cpp
@@ -294,6 +294,23 @@ Peer::getFlowControlJsonInfo() const
     return res;
 }
 
+Json::Value
+Peer::getJsonInfo(bool compact) const
+{
+    Json::Value res;
+    res["address"] = mAddress.toString();
+    res["elapsed"] = (int)getLifeTime().count();
+    res["latency"] = (int)getPing().count();
+    res["ver"] = getRemoteVersion();
+    res["olver"] = (int)getRemoteOverlayVersion();
+    res["flow_control"] = getFlowControlJsonInfo();
+    if (!compact)
+    {
+    }
+
+    return res;
+}
+
 void
 Peer::sendAuth()
 {
--- a/src/overlay/Peer.cpp
+++ b/src/overlay/Peer.cpp
@@ -263,7 +263,7 @@ Peer::recurrentTimerExpired(asio::error_code const& error)
 }
 
 Json::Value
-Peer::getFlowControlJsonInfo() const
+Peer::getFlowControlJsonInfo(bool compact) const
 {
     Json::Value res;
     std::string stateStr;
@@ -291,6 +291,16 @@ Peer::getFlowControlJsonInfo() const
         res["peer_capacity"] = (Json::UInt64)mOutboundCapacity;
     }
 
+    if (!compact)
+    {
+        res["outbound_queue_delay_scp_p75"] = static_cast<Json::UInt64>(
+            mPeerMetrics.mOutboundQueueDelaySCP.GetSnapshot()
+                .get75thPercentile());
+        res["outbound_queue_delay_txs_p75"] = static_cast<Json::UInt64>(
+            mPeerMetrics.mOutboundQueueDelayTxs.GetSnapshot()
+                .get75thPercentile());
+    }
+
     return res;
 }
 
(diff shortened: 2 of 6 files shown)


src/overlay/Peer.cpp
```diff
@@ -1007,7 +1067,14 @@ Peer::maybeSendNextBatch()
             auto& timer = front.mMessage->type() == SCP_MESSAGE
                               ? om.mOutboundQueueDelaySCP
                               : om.mOutboundQueueDelayTxs;
-            timer.Update(mApp.getClock().now() - front.mTimeEmplaced);
+            auto& maxDelay = front.mMessage->type() == SCP_MESSAGE
+                                 ? mPeerMetrics.mMaxOutboundQueueDelaySCP
+                                 : mPeerMetrics.mMaxOutboundQueueDelayTxs;
+            auto const& diff = mApp.getClock().now() - front.mTimeEmplaced;
+            timer.Update(diff);
+            maxDelay = std::max(
```
marta-lokhova, line 1075: Same thing with the max here. Ideally, we can display some recent value to get a sense of how the peer is doing.

src/overlay/Peer.cpp
```diff
@@ -294,6 +294,65 @@ Peer::getFlowControlJsonInfo() const
...
+                mPeerMetrics.mMaxMessageDelayInWriteQueue.count());
+        res["max_message_delay_in_async_write_msec"] =
+            static_cast<Json::UInt64>(
+                mPeerMetrics.mMaxMessageDelayInAsyncWrite.count());
+        res["max_outbound_queue_delay_scp_msec"] = static_cast<Json::UInt64>(
+            mPeerMetrics.mMaxOutboundQueueDelaySCP.count());
+        res["max_outbound_queue_delay_txs_msec"] = static_cast<Json::UInt64>(
+            mPeerMetrics.mMaxOutboundQueueDelayTxs.count());
+
+        res["unique_flood_bytes_recv"] =
```
marta-lokhova, line 334: do we need all of these byte metrics? We already have a very coarse byte measure (total bytes read/written), so I wonder if only including unique and duplicate message counts is sufficient here.

marta-lokhova (changes requested):

docs/software/commands.md
```diff
@@ -179,9 +179,10 @@ format.
   Clear metrics for a specified domain. If no domain specified, clear all
   metrics (for testing purposes).
 
-* **peers?[&fullkeys=false]**
-  Returns the list of known peers in JSON format.
+* **peers?[&fullkeys=false&compact=false]**
```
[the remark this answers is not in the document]
hidenori-shinohara: Updated!

src/overlay/Peer.cpp
```diff
@@ -294,6 +294,65 @@ Peer::getFlowControlJsonInfo() const
...
+        res["message_drop"] =
+            static_cast<Json::UInt64>(mPeerMetrics.mMessageDrop);
+
+        res["max_message_delay_in_write_queue_msec"] =
+            static_cast<Json::UInt64>(
+                mPeerMetrics.mMaxMessageDelayInWriteQueue.count());
+        res["max_message_delay_in_async_write_msec"] =
+            static_cast<Json::UInt64>(
+                mPeerMetrics.mMaxMessageDelayInAsyncWrite.count());
+        res["max_outbound_queue_delay_scp_msec"] = static_cast<Json::UInt64>(
```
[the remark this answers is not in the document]
hidenori-shinohara: Updated!

src/overlay/Peer.cpp
hidenori-shinohara: Removed from reporting. (I kept the member variables since they are used for the survey.)

src/overlay/Peer.cpp
hidenori-shinohara: Updated to show P99s from the last 30-second window

hidenori-shinohara:
Here's a sample output:
```
         {                                                                                                            
            "address" : "54.173.54.200:11625",
            "async_read" : 149,    
            "async_write" : 511,              
            "byte_read" : 353760,            
            "byte_write" : 315188,            
            "duplicate_fetch_message_recv" : 0,
            "duplicate_flood_message_recv" : 547,
            "elapsed" : 18,                    
            "flow_control" : {                    
               "local_capacity" : {
                  "flood" : 200, 
                  "reading" : 201                 
               },                                 
               "outbound_queue_delay_scp_p99" : 0,
               "outbound_queue_delay_txs_p99" : 0,
               "peer_capacity" : 198,        
               "state" : "enabled"
            },                                     
            "id" : "sdf_1",                        
            "latency" : 22,        
            "message_delay_in_async_write_p99" : 0,
            "message_delay_in_write_queue_p99" : 92,
            "message_drop" : 0,
            "message_read" : 748,                  
            "message_write" : 705,                 
            "olver" : 20,                                                                                             
            "unique_fetch_message_recv" : 0,
            "unique_flood_message_recv" : 0,
            "ver" : "stellar-core 18.5.0 (d387c6a710322135ac076804490af22c4587b96d)"                                  
         }
```

src/overlay/Peer.h
```diff
@@ -98,10 +100,10 @@ class Peer : public std::enable_shared_from_this<Peer>,
         uint64_t mAsyncWrite;
         uint64_t mMessageDrop;
 
-        std::chrono::milliseconds mMaxMessageDelayInWriteQueue;
-        std::chrono::milliseconds mMaxMessageDelayInAsyncWrite;
-        std::chrono::milliseconds mMaxOutboundQueueDelaySCP;
-        std::chrono::milliseconds mMaxOutboundQueueDelayTxs;
+        medida::Timer& mMessageDelayInWriteQueueTimer;
```
marta-lokhova, line 103: We shouldn't duplicate these metrics. You can get the data you need in `getJsonInfo` by querying the app-wide metric registry. See `Application::getMetrics`

marta-lokhova (changes requested):

src/overlay/Peer.h
marta-lokhova: actually, scratch that, we don't track per-peer metrics in the global metric registry. So I think we just need to track these timers in the struct itself (we don't need a registry for this)

hidenori-shinohara: Updated it!

src/overlay/TCPPeer.cpp
```diff
@@ -311,14 +313,23 @@ TCPPeer::messageSender()
...
         mCompletedTime - mIssuedTime);
     metrics.mMessageDelayInWriteQueueTimer.Update(qdelay);
     metrics.mMessageDelayInAsyncWriteTimer.Update(wdelay);
+    auto updateMax = [](auto& a, auto const& b) {
+        if (a < b)
+        {
+            a = std::chrono::duration_cast<std::chrono::milliseconds>(b);
+        }
+    };
+    updateMax(peerMetrics.mMaxMessageDelayInWriteQueue, qdelay);
```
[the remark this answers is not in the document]
hidenori-shinohara: Updated to show P99s from the last 300-second window

src/overlay/Peer.cpp
hidenori-shinohara: Edit: 30 -> 300 seconds based on the offline discussion.

marta-lokhova (commented):
Looks good to me overall! One thing I realized is that reporting p99 for the timers maybe isn't as useful as, say, p75, so perhaps we should report that instead.

src/overlay/Peer.cpp
```diff
@@ -1007,7 +1064,12 @@ Peer::maybeSendNextBatch()
             auto& timer = front.mMessage->type() == SCP_MESSAGE
                               ? om.mOutboundQueueDelaySCP
                               : om.mOutboundQueueDelayTxs;
-            timer.Update(mApp.getClock().now() - front.mTimeEmplaced);
+            auto& peerTimer = front.mMessage->type() == SCP_MESSAGE
```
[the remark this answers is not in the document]
hidenori-shinohara: Fixed!

hidenori-shinohara:
I will squash the commits tomorrow

hidenori-shinohara:
I just finished squashing the commits.

MonsieurNicolas:
r+ 0064a59b19abc5c4b226c8363dbbdc6a6e4ccfa9

Issue #2363, opened 2019-12-07
TxSetFrame is messy and confusing


graydon:
The TxSetFrame class needs an overhaul. Among its problems:

  - It is stateful, allowing addition and deletion of txs from the set, which corrupts its sort order.
  - It has methods like trimInvalid or checkOrTrim that implicitly call the deletion method.
  - The deletion method is linear in the size of the txset, and is called from linear trimming loops over the txset, with quadratic results.
  - It has a calculated field (the content hash) that's cached, with a validity flag that's flipped when it changes.
  - It has a sortForHash() method that also does a potentially slightly-expensive calculation but _doesn't_ track its state with a validity flag, meaning it potentially re-sorts anytime anyone calls it, even if it's already sorted
  - It has a sortForApply() method which doesn't change the order of the members, even though the name suggests it does, and some tests appear to think it does (see herder/test/HerderTests.cpp)
  - It has a somewhat awkward abstract base class with both full (TxSetFrame) and simulated (SimulationTxSetFrame) subtypes. It's not clear to a reader what the difference is or why there are two.

As a general sketch -- hopefully without disrupting its users too much! -- I'd recommend the following adjustments:

  - [ ] Keep the cached calculated content hash but make it mutable and invalidated by a `const` helper method, revalidated in the `const` hash-getter (copy what TransactionFrame does here)
  - [ ] Make it a (documented) invariant that a `TxSetFrame`'s transactions are always in hash-sorted order:
    - [ ] When constructed from a `TxSet` off-the-wire, pass a boolean saying whether to sort in the ctor (because we're just mucking around in a test) or whether to require the input is pre-sorted and throw if it's not (eg. for values off the wire, which we should reject immediately at `Peer::recvTxSet` level).
    - [ ] Either store values in a sorted form that's always sorted (`std::set`) ...
    - [ ] Or store them in a sorted `std::vector`...
        - [ ] And re-sort eagerly on each change (possibly batching changes)...
        - [ ] Or flip a mutable `mSorted` bit and re-sort before _observation_, as with the hash
    - [ ] Either way, remove the `sortForHash` function from the public API. Callers shouldn't care.
  - [ ] Remove quadratic filtering loop: build a `std::vector` of txs-to-remove during the filter loop, then sort that vector by hash once (unfortuntately it's _processed_ in apply order) and merge-remove it from the remaining sorted type.
  - [ ] Remove the `AbstractTxSetForApply` base type
[...]

PR #3395, opened 2022-04-11
Refactor TxSetFrame

jayz22:
# Description

Resolves `TxSetFrame` part of #2163 as well as #2363
- Make `TxSetFrame` immutable
- Transactions in `TxSetFrame` are guaranteed to be in hash order, and hash computed on construction.
- Move out utility functions and methods that used to mutate `TxSetFrame` into `TxSetUtils.cpp`
- Remove abstract base class. Repurpose the "TxSim" class into and a child class of `TxSetFrame` that preserves the apply order. 

# Checklist
- [x] Compiles
- [x] Ran all tests

### DIFF
--- a/src/herder/HerderImpl.cpp
+++ b/src/herder/HerderImpl.cpp
@@ -12,6 +12,7 @@
 #include "herder/LedgerCloseData.h"
 #include "herder/QuorumIntersectionChecker.h"
 #include "herder/TxSetFrame.h"
+#include "herder/TxSetUtils.h"
 #include "ledger/LedgerManager.h"
 #include "ledger/LedgerTxn.h"
 #include "ledger/LedgerTxnEntry.h"
@@ -1098,13 +1099,13 @@ HerderImpl::triggerNextLedger(uint32_t ledgerSeqToTrigger,
     upperBoundCloseTimeOffset = nextCloseTime - lcl.header.scpValue.closeTime;
     lowerBoundCloseTimeOffset = upperBoundCloseTimeOffset;
 
-    auto invalidTxs = TxSetFrame::getInvalidTxList(
+    auto removed = TxSetUtils::getInvalidTxList(
         mApp, *proposedSet, lowerBoundCloseTimeOffset,
         upperBoundCloseTimeOffset, false);
 
-    mTransactionQueue.ban(invalidTxs);
-    proposedSet = TxSetFrame::removeTxs(proposedSet, invalidTxs);
-    proposedSet = TxSetFrame::surgePricingFilter(proposedSet, mApp);
+    mTransactionQueue.ban(removed);
+    proposedSet = TxSetUtils::removeTxs(proposedSet, removed);
+    proposedSet = TxSetUtils::surgePricingFilter(proposedSet, mApp);
 
     // we not only check that the value is valid for consensus (offset=0) but
     // also that we performed the proper cleanup above
@@ -1794,7 +1795,7 @@ HerderImpl::updateTransactionQueue(
     lhhe.hash = HashUtils::random();
     auto txSet = mTransactionQueue.toTxSet(lhhe);
 
-    auto invalidTxs = TxSetFrame::getInvalidTxList(
+    auto removed = TxSetUtils::getInvalidTxList(
         mApp, *txSet, 0,
         getUpperBoundCloseTimeOffset(mApp, lhhe.header.scpValue.closeTime),
         false);
--- a/src/herder/HerderImpl.cpp
+++ b/src/herder/HerderImpl.cpp
@@ -1099,12 +1099,13 @@ HerderImpl::triggerNextLedger(uint32_t ledgerSeqToTrigger,
     upperBoundCloseTimeOffset = nextCloseTime - lcl.header.scpValue.closeTime;
     lowerBoundCloseTimeOffset = upperBoundCloseTimeOffset;
 
-    auto removed = TxSetUtils::getInvalidTxList(
-        mApp, *proposedSet, lowerBoundCloseTimeOffset,
-        upperBoundCloseTimeOffset, false);
+    TxSetFrame::Transactions invalidTxs;
+    proposedSet =
+        TxSetUtils::trimInvalid(proposedSet, mApp, lowerBoundCloseTimeOffset,
+                                upperBoundCloseTimeOffset, invalidTxs);
 
-    mTransactionQueue.ban(removed);
-    proposedSet = TxSetUtils::removeTxs(proposedSet, removed);
+    mTransactionQueue.ban(invalidTxs);
+
     proposedSet = TxSetUtils::surgePricingFilter(proposedSet, mApp);
 
     // we not only check that the value is valid for consensus (offset=0) but
@@ -1795,8 +1796,8 @@ HerderImpl::updateTransactionQueue(
     lhhe.hash = HashUtils::random();
     auto txSet = mTransactionQueue.toTxSet(lhhe);
 
-    auto removed = TxSetUtils::getInvalidTxList(
-        mApp, *txSet, 0,
+    auto invalidTxs = TxSetUtils::getInvalidTxList(
+        *txSet, mApp, 0,
         getUpperBoundCloseTimeOffset(mApp, lhhe.header.scpValue.closeTime),
         false);
     mTransactionQueue.ban(invalidTxs);
--- a/src/herder/HerderImpl.cpp
+++ b/src/herder/HerderImpl.cpp
@@ -213,10 +213,10 @@ HerderImpl::newSlotExternalized(bool synchronous, StellarValue const& value)
     mLastExternalize = mApp.getClock().now();
 
     // perform cleanups
-    TxSetFramePtr externalizedSet = mPendingEnvelopes.getTxSet(value.txSetHash);
+    auto externalizedSet = mPendingEnvelopes.getTxSet(value.txSetHash);
     if (externalizedSet)
     {
-        updateTransactionQueue(externalizedSet->mTransactions);
+        updateTransactionQueue(externalizedSet->getTxsInHashOrder());
     }
 
     // Evict slots that are outside of our ledger validity bracket
@@ -265,7 +265,8 @@ HerderImpl::processExternalized(uint64 slotIndex, StellarValue const& value)
                      slotIndex, hexAbbrev(value.txSetHash));
     }
 
-    TxSetFramePtr externalizedSet = mPendingEnvelopes.getTxSet(value.txSetHash);
+    TxSetFrameConstPtr externalizedSet =
+        mPendingEnvelopes.getTxSet(value.txSetHash);
 
     // save the SCP messages in the database
     if (mApp.getConfig().MODE_STORES_HISTORY_MISC)
@@ -696,14 +697,14 @@ HerderImpl::recvSCPEnvelope(SCPEnvelope const& envelope,
     ZoneScoped;
     mPendingEnvelopes.addTxSet(txset.getContentsHash(),
                                envelope.statement.slotIndex,
-                               std::make_shared<TxSetFrame>(txset));
+                               std::make_shared<TxSetFrame const>(txset));
     mPendingEnvelopes.addSCPQuorumSet(xdrSha256(qset), qset);
     return recvSCPEnvelope(envelope);
 }
 
 void
-HerderImpl::externalizeValue(std::shared_ptr<TxSetFrame> txSet,
-                             uint32_t ledgerSeq, uint64_t closeTime,
+HerderImpl::externalizeValue(TxSetFrameConstPtr txSet, uint32_t ledgerSeq,
+                             uint64_t closeTime,
                              xdr::xvector<UpgradeType, 6> const& upgrades,
                              std::optional<SecretKey> skToSignValue)
 {
@@ -961,7 +962,7 @@ bool
 HerderImpl::recvTxSet(Hash const& hash, const TxSetFrame& t)
 {
     ZoneScoped;
-    auto txset = std::make_shared<TxSetFrame>(t);
+    auto txset = std::make_shared<TxSetFrame const>(t);
     return mPendingEnvelopes.recvTxSet(hash, txset);
 }
 
@@ -973,7 +974,7 @@ HerderImpl::peerDoesntHave(MessageType type, uint256 const& itemID,
     mPendingEnvelopes.peerDoesntHave(type, itemID, peer);
 }
 
-TxSetFramePtr
+TxSetFrameConstPtr
 HerderImpl::getTxSet(Hash const& hash)
 {
     return mPendingEnvelopes.getTxSet(hash);
@@ -1097,11 +1098,13 @@ HerderImpl::triggerNextLedger(uint32_t ledgerSeqToTrigger,
     upperBoundCloseTimeOffset = nextCloseTime - lcl.header.scpValue.closeTime;
     lowerBoundCloseTimeOffset = upperBoundCloseTimeOffset;
 
-    auto removed = proposedSet->trimInvalid(mApp, lowerBoundCloseTimeOffset,
-                                            upperBoundCloseTimeOffset);
-    mTransactionQueue.ban(removed);
+    auto invalidTxs = TxSetFrame::getInvalidTxList(
+        mApp, *proposedSet, lowerBoundCloseTimeOffset,
+        upperBoundCloseTimeOffset, false);
 
-    proposedSet->surgePricingFilter(mApp);
+    mTransactionQueue.ban(invalidTxs);
+    proposedSet = TxSetFrame::removeTxs(proposedSet, invalidTxs);
+    proposedSet = TxSetFrame::surgePricingFilter(proposedSet, mApp);
 
     // we not only check that the value is valid for consensus (offset=0) but
     // also that we performed the proper cleanup above
@@ -1597,7 +1600,7 @@ HerderImpl::persistSCPState(uint64 slot)
 
     // saves SCP messages and related data (transaction sets, quorum sets)
     xdr::xvector<SCPEnvelope> latestEnvs;
-    std::map<Hash, TxSetFramePtr> txSets;
+    std::map<Hash, TxSetFrameConstPtr> txSets;
     std::map<Hash, SCPQuorumSetPtr> quorumSets;
 
     for (auto const& e : getSCP().getLatestMessagesSend(slot))
@@ -1664,8 +1667,8 @@ HerderImpl::restoreSCPState()
 
             for (auto const& txset : latestTxSets)
             {
-                TxSetFramePtr cur =
-                    make_shared<TxSetFrame>(mApp.getNetworkID(), txset);
+                TxSetFrameConstPtr cur =
+                    make_shared<TxSetFrame const>(mApp.getNetworkID(), txset);
                 Hash h = cur->getContentsHash();
                 mPendingEnvelopes.addTxSet(h, 0, cur);
             }
@@ -1791,10 +1794,11 @@ HerderImpl::updateTransactionQueue(
     lhhe.hash = HashUtils::random();
     auto txSet = mTransactionQueue.toTxSet(lhhe);
 
-    auto removed = txSet->trimInvalid(
-        mApp, 0,
-        getUpperBoundCloseTimeOffset(mApp, lhhe.header.scpValue.closeTime));
-    mTransactionQueue.ban(removed);
+    auto invalidTxs = TxSetFrame::getInvalidTxList(
+        mApp, *txSet, 0,
+        getUpperBoundCloseTimeOffset(mApp, lhhe.header.scpValue.closeTime),
+        false);
+    mTransactionQueue.ban(invalidTxs);
 
     mTransactionQueue.rebroadcast();
 }
--- a/src/herder/HerderSCPDriver.cpp
+++ b/src/herder/HerderSCPDriver.cpp
@@ -106,7 +106,7 @@ class SCPHerderEnvelopeWrapper : public SCPEnvelopeWrapper
     HerderImpl& mHerder;
 
     SCPQuorumSetPtr mQSet;
-    std::vector<TxSetFramePtr> mTxSets;
+    std::vector<TxSetFrameConstPtr> mTxSets;
 
   public:
     explicit SCPHerderEnvelopeWrapper(SCPEnvelope const& e, HerderImpl& herder)
@@ -301,7 +301,7 @@ HerderSCPDriver::validateValueHelper(uint64_t slotIndex, StellarValue const& b,
     }
 
     Hash const& txSetHash = b.txSetHash;
-    TxSetFramePtr txSet = mPendingEnvelopes.getTxSet(txSetHash);
+    TxSetFrameConstPtr txSet = mPendingEnvelopes.getTxSet(txSetHash);
 
     SCPDriver::ValidationLevel res;
 
@@ -790,7 +790,7 @@ HerderSCPDriver::logQuorumInformation(uint64_t index)
 
 void
 HerderSCPDriver::nominate(uint64_t slotIndex, StellarValue const& value,
-                          TxSetFramePtr proposedSet,
+                          TxSetFrameConstPtr proposedSet,
                           StellarValue const& previousValue)
 {
     ZoneScoped;
@@ -801,7 +801,7 @@ HerderSCPDriver::nominate(uint64_t slotIndex, StellarValue const& value,
     CLOG_DEBUG(Herder,
                "HerderSCPDriver::triggerNextLedger txSet.size: {} "
                "previousLedgerHash: {} value: {} slot: {}",
-               proposedSet->mTransactions.size(),
+               proposedSet->sizeTx(),
                hexAbbrev(proposedSet->previousLedgerHash()),
                hexAbbrev(valueHash), slotIndex);
 
@@ -1084,7 +1084,7 @@ class SCPHerderValueWrapper : public ValueWrapper
 {
     HerderImpl& mHerder;
 
-    TxSetFramePtr mTxSet;
+    TxSetFrameConstPtr mTxSet;
 
   public:
     explicit SCPHerderValueWrapper(StellarValue const& sv, Value const& value,
--- a/src/herder/HerderSCPDriver.cpp
+++ b/src/herder/HerderSCPDriver.cpp
@@ -33,6 +33,8 @@
 namespace stellar
 {
 
+uint32_t const TXSETVALID_CACHE_SIZE = 1000;
+
 Hash
 HerderSCPDriver::getHashOf(std::vector<xdr::opaque_vec<>> const& vals) const
 {
@@ -79,6 +81,7 @@ HerderSCPDriver::HerderSCPDriver(Application& app, HerderImpl& herder,
     , mPrepareTimeout{mApp.getMetrics().NewHistogram(
           {"scp", "timeout", "prepare"})}
     , mLedgerSeqNominating(0)
+    , mTxSetValidCache(TXSETVALID_CACHE_SIZE)
 {
 }
 
@@ -314,7 +317,7 @@ HerderSCPDriver::validateValueHelper(uint64_t slotIndex, StellarValue const& b,
 
         res = SCPDriver::kInvalidValue;
     }
-    else if (!txSet->checkValid(mApp, closeTimeOffset, closeTimeOffset))
+    else if (!checkAndCacheTxSetValid(txSet, closeTimeOffset))
     {
         CLOG_DEBUG(Herder,
                    "HerderSCPDriver::validateValue i: {} invalid txSet {}",
@@ -1123,5 +1126,26 @@ HerderSCPDriver::wrapStellarValue(StellarValue const& sv)
     auto val = xdr::xdr_to_opaque(sv);
     auto res = std::make_shared<SCPHerderValueWrapper>(sv, val, mHerder);
     return res;
+}
+
+bool
+HerderSCPDriver::checkAndCacheTxSetValid(TxSetFrameConstPtr txSet,
+                                         uint64_t closeTimeOffset) const
+{
+    auto key = TxSetUtils::TxSetValidityKey{
+        mApp.getLedgerManager().getLastClosedLedgerHeader().hash,
+        txSet->getContentsHash(), closeTimeOffset, closeTimeOffset};
+
+    bool* pRes = mTxSetValidCache.maybeGet(key);
+    if (pRes == nullptr)
+    {
+        bool res = txSet->checkValid(mApp, closeTimeOffset, closeTimeOffset);
+        mTxSetValidCache.put(key, res);
+        return res;
+    }
+    else
+    {
+        return *pRes;
+    }
 }
 }
--- a/src/herder/HerderSCPDriver.h
+++ b/src/herder/HerderSCPDriver.h
@@ -85,7 +85,8 @@ class HerderSCPDriver : public SCPDriver
     // Submit a value to consider for slotIndex
     // previousValue is the value from slotIndex-1
     void nominate(uint64_t slotIndex, StellarValue const& value,
-                  TxSetFramePtr proposedSet, StellarValue const& previousValue);
+                  TxSetFrameConstPtr proposedSet,
+                  StellarValue const& previousValue);
 
     SCPQuorumSetPtr getQSet(Hash const& qSetHash) override;
 
--- a/src/herder/HerderSCPDriver.h
+++ b/src/herder/HerderSCPDriver.h
@@ -6,8 +6,10 @@
 
 #include "herder/Herder.h"
 #include "herder/TxSetFrame.h"
+#include "herder/TxSetUtils.h"
 #include "medida/timer.h"
 #include "scp/SCPDriver.h"
+#include "util/RandomEvictionCache.h"
 #include "xdr/Stellar-ledger.h"
 #include <optional>
 
@@ -191,6 +193,11 @@ class HerderSCPDriver : public SCPDriver
     // indexed by slotIndex, timerID
     std::map<uint64_t, std::map<int, std::unique_ptr<VirtualTimer>>> mSCPTimers;
 
+    // validity of txSet
+    mutable RandomEvictionCache<TxSetUtils::TxSetValidityKey, bool,
+                                TxSetUtils::TxSetValidityKeyHash>
+        mTxSetValidCache;
+
     SCPDriver::ValidationLevel validateValueHelper(uint64_t slotIndex,
                                                    StellarValue const& sv,
                                                    bool nomination) const;
@@ -207,5 +214,8 @@ class HerderSCPDriver : public SCPDriver
                          std::string const& logStr,
                          std::chrono::nanoseconds threshold,
                          uint64_t slotIndex);
+
+    bool checkAndCacheTxSetValid(TxSetFrameConstPtr TxSet,
+                                 uint64_t closeTimeOffset) const;
 };
 }
--- a/src/herder/LedgerCloseData.cpp
+++ b/src/herder/LedgerCloseData.cpp
@@ -14,9 +14,9 @@ using namespace std;
 namespace stellar
 {
 
-LedgerCloseData::LedgerCloseData(
-    uint32_t ledgerSeq, std::shared_ptr<AbstractTxSetFrameForApply> txSet,
-    StellarValue const& v, std::optional<Hash> const& expectedLedgerHash)
+LedgerCloseData::LedgerCloseData(uint32_t ledgerSeq, TxSetFrameConstPtr txSet,
+                                 StellarValue const& v,
+                                 std::optional<Hash> const& expectedLedgerHash)
     : mLedgerSeq(ledgerSeq)
     , mTxSet(txSet)
     , mValue(v)
--- a/src/herder/TxSetFrame.cpp
+++ b/src/herder/TxSetFrame.cpp
@@ -33,23 +33,26 @@ namespace stellar
 
 using namespace std;
 
+TxSetFrame::TxSetFrame(Hash const& previousLedgerHash,
+                       Transactions const& transactions)
+    : mPreviousLedgerHash(previousLedgerHash)
+    , mTxsInHashOrder(TxSetFrame::sortTxsInHashOrder(transactions))
+    , mHash(
+          TxSetFrame::computeContentsHash(previousLedgerHash, mTxsInHashOrder))
+{
+}
+
 TxSetFrame::TxSetFrame(Hash const& previousLedgerHash)
-    : mHash(std::nullopt)
-    , mValid(std::nullopt)
-    , mPreviousLedgerHash(previousLedgerHash)
+    : TxSetFrame(previousLedgerHash, Transactions{})
 {
 }
 
 TxSetFrame::TxSetFrame(Hash const& networkID, TransactionSet const& xdrSet)
-    : mHash(std::nullopt), mValid(std::nullopt)
+    : mPreviousLedgerHash(xdrSet.previousLedgerHash)
+    , mTxsInHashOrder(TxSetFrame::sortTxsInHashOrder(networkID, xdrSet))
+    , mHash(
+          TxSetFrame::computeContentsHash(mPreviousLedgerHash, mTxsInHashOrder))
 {
-    ZoneScoped;
-    for (auto const& env : xdrSet.txs)
-    {
-        auto tx = TransactionFrameBase::makeTransactionFromWire(networkID, env);
-        mTransactions.push_back(tx);
-    }
-    mPreviousLedgerHash = xdrSet.previousLedgerHash;
 }
 
 static bool
@@ -63,13 +66,28 @@ HashTxSorter(TransactionFrameBasePtr const& tx1,
 
 // order the txset correctly
 // must take into account multiple tx from same account
-void
-TxSetFrame::sortForHash()
+TxSetFrame::Transactions
+TxSetFrame::sortTxsInHashOrder(TxSetFrame::Transactions const& transactions)
 {
     ZoneScoped;
-    std::sort(mTransactions.begin(), mTransactions.end(), HashTxSorter);
-    mHash.reset();
-    mValid.reset();
+    TxSetFrame::Transactions sortedTxs(transactions);
+    std::sort(sortedTxs.begin(), sortedTxs.end(), HashTxSorter);
+    return sortedTxs;
+}
+
+TxSetFrame::Transactions
+sortTxsInHashOrder(Hash const& networkID, TransactionSet const& xdrSet)
+{
+    ZoneScoped;
+    TxSetFrame::Transactions txs;
+    txs.reserve(xdrSet.txs.size());
+    std::transform(xdrSet.txs.cbegin(), xdrSet.txs.cend(),
+                   std::back_inserter(txs),
+                   [&](TransactionEnvelope const& env) {
+                       return TransactionFrameBase::makeTransactionFromWire(
+                           networkID, env);
+                   });
+    return TxSetFrame::sortTxsInHashOrder(txs);
 }
 
 // We want to XOR the tx hash with the set hash.
@@ -91,13 +109,6 @@ struct ApplyTxSorter
     }
 };
 
-static bool
-SeqSorter(TransactionFrameBasePtr const& tx1,
-          TransactionFrameBasePtr const& tx2)
-{
-    return tx1->getSeqNum() < tx2->getSeqNum();
-}
-
 /*
     Build a list of transaction ready to be applied to the last closed ledger,
     based on the transaction set.
@@ -106,16 +117,16 @@ SeqSorter(TransactionFrameBasePtr const& tx1,
     * transactions for an account are sorted by sequence number (ascending)
     * the order between accounts is randomized
 */
-std::vector<TransactionFrameBasePtr>
-TxSetFrame::sortForApply()
+TxSetFrame::Transactions
+TxSetFrame::getTxsInApplyOrder() const
 {
     ZoneScoped;
-    auto txQueues = buildAccountTxQueues();
+    auto txQueues = TxSetFrame::buildAccountTxQueues(*this);
 
     // build txBatches
     // txBatches i-th element contains each i-th transaction for accounts with a
     // transaction in the transaction set
-    std::list<std::deque<TransactionFrameBasePtr>> txBatches;
+    std::list<AccountTransactionQueue> txBatches;
 
     while (!txQueues.empty())
     {
@@ -139,8 +150,8 @@ TxSetFrame::sortForApply()
         }
     }
 
-    vector<TransactionFrameBasePtr> retList;
-    retList.reserve(mTransactions.size());
+    Transactions txsInApplyOrder;
+    txsInApplyOrder.reserve(mTxsInHashOrder.size());
     for (auto& batch : txBatches)
     {
         // randomize each batch using the hash of the transaction set
@@ -149,11 +160,10 @@ TxSetFrame::sortForApply()
         std::sort(batch.begin(), batch.end(), s);
         for (auto const& tx : batch)
         {
-            retList.push_back(tx);
+            txsInApplyOrder.push_back(tx);
         }
     }
-
-    return retList;
+    return txsInApplyOrder;
 }
 
 struct SurgeCompare
@@ -192,17 +202,17 @@ struct SurgeCompare
 };
 
 UnorderedMap<AccountID, TxSetFrame::AccountTransactionQueue>
-TxSetFrame::buildAccountTxQueues()
+TxSetFrame::buildAccountTxQueues(TxSetFrame const& txSet)
 {
     ZoneScoped;
-    UnorderedMap<AccountID, AccountTransactionQueue> actTxQueueMap;
-    for (auto& tx : mTransactions)
+    UnorderedMap<AccountID, TxSetFrame::AccountTransactionQueue> actTxQueueMap;
+    for (auto const& tx : txSet.getTxsInHashOrder())
     {
         auto id = tx->getSourceID();
         auto it = actTxQueueMap.find(id);
         if (it == actTxQueueMap.end())
         {
-            auto d = std::make_pair(id, AccountTransactionQueue{});
+            auto d = std::make_pair(id, TxSetFrame::AccountTransactionQueue{});
             auto r = actTxQueueMap.insert(d);
             it = r.first;
         }
@@ -212,13 +222,17 @@ TxSetFrame::buildAccountTxQueues()
     for (auto& am : actTxQueueMap)
     {
         // sort each in sequence number order
-        std::sort(am.second.begin(), am.second.end(), SeqSorter);
+        std::sort(am.second.begin(), am.second.end(),
+                  [](TransactionFrameBasePtr const& tx1,
+                     TransactionFrameBasePtr const& tx2) {
+                      return tx1->getSeqNum() < tx2->getSeqNum();
+                  });
     }
     return actTxQueueMap;
 }
 
-void
-TxSetFrame::surgePricingFilter(Application& app)
+TxSetFrameConstPtr
+TxSetFrame::surgePricingFilter(TxSetFrameConstPtr txSet, Application& app)
 {
     ZoneScoped;
     LedgerTxn ltx(app.getLedgerTxnRoot());
@@ -229,16 +243,18 @@ TxSetFrame::surgePricingFilter(Application& app)
 
     size_t opsLeft = app.getLedgerManager().getLastMaxTxSetSizeOps();
 
-    auto curSizeOps = maxIsOps ? sizeOp() : (sizeTx() * MAX_OPS_PER_TX);
+    auto curSizeOps =
+        maxIsOps ? txSet->sizeOp() : (txSet->sizeTx() * MAX_OPS_PER_TX);
     if (curSizeOps > opsLeft)
     {
         CLOG_WARNING(Herder, "surge pricing in effect! {} > {}", curSizeOps,
                      opsLeft);
 
-        auto actTxQueueMap = buildAccountTxQueues();
+        auto actTxQueueMap = buildAccountTxQueues(*txSet);
 
-        std::priority_queue<AccountTransactionQueue*,
-                            std::vector<AccountTransactionQueue*>, SurgeCompare>
+        std::priority_queue<TxSetFrame::AccountTransactionQueue*,
+                            std::vector<TxSetFrame::AccountTransactionQueue*>,
+                            SurgeCompare>
             surgeQueue;
 
         for (auto& am : actTxQueueMap)
@@ -246,8 +262,8 @@ TxSetFrame::surgePricingFilter(Application& app)
             surgeQueue.push(&am.second);
         }
 
-        std::vector<TransactionFrameBasePtr> updatedSet;
-        updatedSet.reserve(mTransactions.size());
+        TxSetFrame::Transactions updatedSet;
+        updatedSet.reserve(txSet->sizeTx());
         while (opsLeft > 0 && !surgeQueue.empty())
         {
             auto cur = surgeQueue.top();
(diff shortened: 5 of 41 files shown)


src/herder/TxSetFrame.cpp
```diff
@@ -274,22 +297,22 @@ TxSetFrame::surgePricingFilter(Application& app)
...
+
+        transactions = std::move(updatedSet);
+        // sortForHash();
     }
 }
 
 bool
-TxSetFrame::checkOrTrim(Application& app,
-                        std::vector<TransactionFrameBasePtr>& trimmed,
+TxSetFrame::checkOrTrim(Application& app, TxSetFrame::Transactions& trimmed,
```
dmkozh, line 307: Seems like trimming logic is gone now, isn't it? Previously this mutated the tx set when `justCheck` is `false`.
I think logically there should be a bool method for check and a method that returns a `TxSetFrame` for trim (maybe with partially shared implementation).

src/herder/TxSetFrame.h
```diff
@@ -27,70 +27,94 @@ class AbstractTxSetFrameForApply
...
+// A wrapper for a set of transactions that maintains the hash order
+
 class TxSetFrame : public AbstractTxSetFrameForApply
 {
-    std::optional<Hash> mHash;
+    using AccountTransactionQueue = std::deque<TransactionFrameBasePtr>;
+    using Transactions = std::vector<TransactionFrameBasePtr>;
+
+    // std::optional<Hash> mHash;
+    Hash mHash;
```
dmkozh, line 50: Can we make all the members `const` to make sure they're immutable?

src/herder/TxSetFrame.h
```diff
@@ -17,8 +17,8 @@ namespace stellar
 class Application;
 
 class TxSetFrame;
-typedef std::shared_ptr<TxSetFrame> TxSetFramePtr;
-typedef std::shared_ptr<TxSetFrame const> TxSetFrameConstPtr;
+using TxSetFramePtr = std::shared_ptr<TxSetFrame>;
+using TxSetFrameConstPtr = std::shared_ptr<TxSetFrame const>;
 
 class AbstractTxSetFrameForApply
```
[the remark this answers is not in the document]
jayz22: Don't think it is still relevant any more (as per Nico's [comment](https://github.com/stellar/stellar-core/issues/2363#issuecomment-563348512) first point). 
We will just lazy compute the apply order, and it is not needed that often, only in closeLedger.

src/herder/LedgerCloseData.cpp
```diff
@@ -14,9 +14,10 @@ using namespace std;
 namespace stellar
 {
 
-LedgerCloseData::LedgerCloseData(
-    uint32_t ledgerSeq, std::shared_ptr<AbstractTxSetFrameForApply> txSet,
-    StellarValue const& v, std::optional<Hash> const& expectedLedgerHash)
+LedgerCloseData::LedgerCloseData(uint32_t ledgerSeq,
+                                 std::shared_ptr<TxSetFrame> txSet,
```
dmkozh, line 18: nit: We have `TxSetFramePtr`, which looks a bit more concise (and also consistent with `TxFramePtr`). Could you please update the relevant places to use that alias instead?

src/herder/TxSetUtils.cpp
```diff
@@ -0,0 +1,269 @@
...
+        if (cmp3 != 0)
+        {
+            return cmp3 < 0;
+        }
+        // use hash of transaction as a tie breaker
+        return lessThanXored(top1->getFullHash(), top2->getFullHash(), mSeed);
+    }
+};
+
+void
```
dmkozh, line 101: This should return a new `TxSetFrame`. Mutation breaks, at least, the tx hash and likely some other invariants too.

src/herder/TxSetFrame.cpp
```diff
@@ -106,16 +123,16 @@ SeqSorter(TransactionFrameBasePtr const& tx1,
     * transactions for an account are sorted by sequence number (ascending)
     * the order between accounts is randomized
 */
-std::vector<TransactionFrameBasePtr>
-TxSetFrame::sortForApply()
+void
+TxSetFrame::sortInApplyOrder()
```
dmkozh, line 127: I wonder if we really need to persist the apply order as a class member instead of just returning a properly sorted vector. What problem do we solve by this change?

src/herder/TxSetUtils.cpp
```diff
@@ -0,0 +1,281 @@
...
+            {
+                // drop this transaction -> we need to drop the others
+                cur->clear();
+            }
+        }
+
+        return std::make_shared<TxSetFrame>(txSetPtr->previousLedgerHash(),
+                                            updatedSet, false);
+    }
+    return std::make_shared<TxSetFrame>(*txSetPtr);
```
dmkozh, line 164: We don't need to allocate a new tx set in this case, do we? If we never mutate the tx set, we could just always use the const ptr, so that we don't need to do the cast (which, I suppose, was the reason to allocate a copy here).

src/herder/TxSetFrame.h
```diff
@@ -17,105 +16,99 @@ namespace stellar
 class Application;
 
 class TxSetFrame;
-typedef std::shared_ptr<TxSetFrame> TxSetFramePtr;
-typedef std::shared_ptr<TxSetFrame const> TxSetFrameConstPtr;
+using TxSetFramePtr = std::shared_ptr<TxSetFrame>;
```
dmkozh, line 19: Do we still need this?

src/herder/TxSetFrame.h
```diff
@@ -17,105 +16,99 @@ namespace stellar
...
+    Transactions const&
+    getTxsInHashOrder() const
+    {
+        return mTxsInHashOrder;
+    }
 
-    void removeTx(TransactionFrameBasePtr tx);
+#ifdef BUILD_TESTS
+    Transactions&
+    transactions()
```
dmkozh, line 65: How frequently do we use the mutable method in the tests? I recall only one or two test cases where we directly modify the internal tx vector. Instead, we could just modify the temporary vector and construct new tx sets.

src/herder/TxSetFrame.cpp
```diff
@@ -33,23 +34,48 @@ namespace stellar
...
+}
+
+TxSetFrame::TxSetFrame(Hash const& previousLedgerHash,
+                       Transactions const& transactions, bool inputIsApplyOrder)
+    : mPreviousLedgerHash(previousLedgerHash)
+    , mTxsInHashOrder(transactions)
+    , mTxsInApplyOrder(transactions)
+{
+    computeContentsHash();
+    if (!inputIsApplyOrder)
```
dmkozh, line 58: I wonder how frequently is this actually true. I suspect it should be just a single code path. Maybe we should at least make this explicitly default argument in declaration?

src/herder/TxSetUtils.h
```diff
@@ -0,0 +1,38 @@
...
+#include <deque>
+#include <herder/TxSetFrame.h>
+
+namespace stellar
+{
+namespace TxSetUtils
+{
+
+UnorderedMap<AccountID, TxSetFrame::AccountTransactionQueue>
+buildAccountTxQueues(TxSetFrame const& txSet);
```
dmkozh, line 18: This kind of demonstrates the issue with separating these functions from `TxSetFrame`: on one hand this is an implementation detail and hence shouldn't be exposed in the header (unlike other fucntions here), but since it needs to be shared between these two files, it has to be exposed. I don't think it's a big deal for now, just something to consider.

src/herder/TxSetUtils.h
```diff
@@ -0,0 +1,38 @@
...
+
+TxSetFramePtr surgePricingFilter(Application& app, TxSetFrameConstPtr txSetPtr);
```
dmkozh, line 20: nit: Can we make these functions a bit more consistent:
- Consistent `txSet` argument placement (e.g. the first argument)
- Consistent type of `txSet`. I'd use const reference if possible, but const ptr is also ok (due to some optimizations it enables)

src/herder/simulation/TxSimTxSetFrame.h
```diff
@@ -11,33 +11,14 @@ namespace stellar
 namespace txsimulation
 {
 
-class TxSimTxSetFrame : public AbstractTxSetFrameForApply
+class TxSimTxSetFrame
```
dmkozh, line 14: We don't this class anymore, do we? A free function is fine IMO.

src/herder/TxSetFrame.cpp
jayz22: It is not used frequently, just a couple of places (simulation and some tests). 
However, I don't think having default arguments is a good practice (e.g. see this [blog](https://quuxplusone.github.io/blog/2020/04/18/default-function-arguments-are-the-devil/)) and I don't see them anywhere in our code base. I'm open to it though.

src/herder/TxSetFrame.cpp
jayz22: I don't see a better way if we want to get rid of the abstract class approach. 
Here in `TxSetFrame`, the transactions are guaranteed to always maintain hash order. In the "simulation" case (previously `TxSimTxSetFrame`), the transactions also need to maintain its original order as the apply order. Unless we override the `TxSetFrame` class with a different "sortInApplyOrder" behavior, this is the best way I can think of.

dmkozh: Oh, right, I forgot about that. This is still a bit hacky, but should be ok. I guess an alternative would be to just make `sortForApply` virtual and have `TxSimTxSetFrame` inherited from `TxSetFrame` directly (no ABC), but I don't have a strong opinion on this.

src/herder/TxSetFrame.h
dmkozh: I don't see what exactly does this test cover now, given that `TxSetFrame` now maintains the inavriant of tx set being hash-sorted. So the scenario you described just cannot happen anymore.

src/herder/TxSetFrame.cpp
dmkozh: I don't see anything wrong with default arguments in normal usage scenarios (the blog goes into pretty convoluted edge cases). I guess in this particular case my issue is rather not about using/not using the default arguments, but more about the fact that we kind of leak the simulation-related code (as we only need this to be `true` for simulation, right?) into the generic class. This causes the weird `false` argument to be present in multiple places.
Besides the virtual function approach I suggested in the comment above, another approach could be to create the `TxSetFrame` from hash-sorted and apply-sorted txs and have helper functions that build it for regular and simulation cases (similar to what we do for `TransactionFrame` creation).

src/herder/test/HerderTests.cpp
```diff
@@ -274,138 +287,140 @@ testTxSet(uint32 protocolVersion)
...
             REQUIRE(txSet->checkValid(*app, 0, 0));
 
-            txSet->trimInvalid(*app, 0, 0);
+            trimInvalid(*app, txSet, 0, 0);
             REQUIRE(txSet->checkValid(*app, 0, 0));
-            REQUIRE(txSet->mTransactions.size() == (2 * nbTransactions));
+            REQUIRE(txSet->sizeTx() == (2 * nbTransactions));
         }
+        /* No longer possible for txs to be out of order, and transactions are
+        no longer mutable.
```
jayz22, line 337: I'll remove this block once I get a sign-off.

src/herder/test/TransactionQueueTests.cpp
```diff
@@ -1440,12 +1440,15 @@ TEST_CASE("transaction queue starting sequence boundary",
             auto lcl = app->getLedgerManager().getLastClosedLedgerHeader();
             lcl.header.ledgerSeq = ledgerSeq;
             auto txSet = tq.toTxSet(lcl);
-            REQUIRE(txSet->mTransactions.size() == size);
+            REQUIRE(txSet->sizeTx() == size);
+            /* This is no longer valid. The TxSet is always in Hash order, the
+            original queuing order is no longer preserved.
```
jayz22, line 1445: I'll remove this block once getting sign-off.

src/herder/TxSetFrame.cpp
jayz22: Fixed by separating the apply-order use case into a child class `TxSetInApplyOrder`, so not to mingle with the normal use case.

src/herder/TxSetFrame.h
jayz22: Removed the `TxSetFramePtr`

src/herder/TxSetFrame.cpp
jayz22: Fixed. Same comment as [below](https://github.com/stellar/stellar-core/pull/3395#discussion_r854371104).

src/herder/HerderImpl.cpp
```diff
@@ -1097,11 +1097,15 @@ HerderImpl::triggerNextLedger(uint32_t ledgerSeqToTrigger,
...
-    auto removed = proposedSet->trimInvalid(mApp, lowerBoundCloseTimeOffset,
-                                            upperBoundCloseTimeOffset);
+    auto removed = proposedSet->getTxTrimList(mApp, lowerBoundCloseTimeOffset,
+                                              upperBoundCloseTimeOffset);
     mTransactionQueue.ban(removed);
+    auto updated = proposedSet->getTransactions();
+    TxSetFrame::removeTxs(updated, removed);
 
-    proposedSet->surgePricingFilter(mApp);
+    proposedSet->surgePricingFilter(mApp, updated);
```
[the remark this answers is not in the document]
jayz22: Fixed.

src/herder/TxSetFrame.h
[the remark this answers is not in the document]
jayz22: Fixed.

src/herder/TxSetFrame.h
jayz22: I'm using `const` members whenever can. Some members, e.g Hash, Transactions, needs to be computed and assigned.

src/herder/TxSetFrame.cpp
jayz22: The `checkOfTrim` method has been moved and renamed to `getInvalidTxList` in `TxSetUtils` which is a lot clearer on what it supposed to do. The old "trimInvalid" function (which is only used 1-2 places in the old production code), is now replace by a call to `getInvalidTxList` followed by `removeTxs` passing in the invalid set. In the `HerderTest.cpp` there is now a free function `trimInvalid` that imitates the old function and only used in tests.

src/herder/TxSetFrame.cpp
```diff
@@ -149,357 +160,50 @@ TxSetFrame::sortForApply()
...
-    return mPreviousLedgerHash;
-}
-
-Hash const&
-TxSetFrame::previousLedgerHash() const
-{
-    return mPreviousLedgerHash;
+    return TxSetUtils::getInvalidTxList(app, *this, lowerBoundCloseTimeOffset,
+                                        upperBoundCloseTimeOffset, true)
+               .empty() == true;
```
dmkozh, line 206: nit: unnecessary `== true`

src/herder/TxSetFrame.cpp
```diff
@@ -558,23 +262,23 @@ TxSetFrame::getTotalFees(LedgerHeader const& lh) const
...
                            });
 }
 
 void
-TxSetFrame::toXDR(TransactionSet& txSet)
+TxSetFrame::toXDR(TransactionSet& txSet) const
 {
     ZoneScoped;
-    releaseAssert(std::is_sorted(mTransactions.begin(), mTransactions.end(),
+    releaseAssert(std::is_sorted(mTxsInHashOrder.begin(), mTxsInHashOrder.end(),
```
graydon, line 276: this is probably also redundant now, given that it's true by construction.

src/herder/TxSetFrame.cpp
```diff
@@ -33,23 +34,42 @@ namespace stellar
 
 using namespace std;
 
+void
+TxSetFrame::computeContentsHash()
```
graydon, line 38: I think if you make this function and sortInHashOrder static functions that return values rather than member functions, you can call them in the class initializer list for `mHash` and `mTxsInHashOrder`, and thereby make those fields `const`. If possible, that would be the strongest version of making this class immutable.

graydon (changes requested):
This is great! Thanks so much for doing this -- we've been wanting it done for years.

A handful of requested changes, happy to talk through any of them.

src/herder/TxSetFrame.cpp
jayz22: This is a great suggestion! 
Fixed.

src/herder/TxSetUtils.h
```diff
@@ -0,0 +1,39 @@
...
+                                      Application& app);
+
+// Returns transactions from a TxSet that are invalid. If returnIfFound is true,
+// return immediately if an invalid transaction is found (instead of finding all
+// of them), this is useful for checking if a TxSet is valid.
+TxSetFrame::Transactions getInvalidTxList(Application& app,
+                                          TxSetFrame const& txSet,
+                                          uint64_t lowerBoundCloseTimeOffset,
+                                          uint64_t upperBoundCloseTimeOffset,
+                                          bool returnIfFound);
```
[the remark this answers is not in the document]
jayz22: Fixed.

src/herder/TxSetUtils.cpp
```diff
@@ -0,0 +1,310 @@
...
+    return invalidTxs;
+}
+
+TxSetFrameConstPtr
+removeTxs(TxSetFrameConstPtr txSetPtr, TxSetFrame::Transactions const& filter)
+{
+    auto updated = txSetPtr->getTxsInHashOrder();
+    for (auto const& tx : filter)
+    {
+        auto it = std::find(updated.begin(), updated.end(), tx);
```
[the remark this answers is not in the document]
jayz22: That makes sense. Fixed it.

src/herder/TxSetFrame.cpp
jayz22: Fixed.

src/herder/TxSetFrame.h
jayz22: Fixed it based on Graydon's suggestion [below](https://github.com/stellar/stellar-core/pull/3395#discussion_r854563736). All members are `const` now.

src/herder/TxSetUtils.cpp
```diff
@@ -0,0 +1,310 @@
...
+#include <list>
+#include <numeric>
+
+namespace stellar
+{
+namespace TxSetUtils
+{
+
+UnorderedMap<AccountID, TxSetFrame::AccountTransactionQueue>
+buildAccountTxQueues(TxSetFrame const& txSet)
```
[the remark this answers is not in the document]
jayz22: You are right, this is mostly code moved over from the `TxSetFrame` class with the logic kept the same. Just some interface changes (e.g. change `checkOrTrim` to `getInvalidTxList`) and adaptation. On hindsight I probably should've make the interface changes in a separate commit from the code moving to make it easier to review/diff. I'll make sure to do that next time.

src/herder/TxSetFrame.cpp
```diff
@@ -33,23 +34,26 @@ namespace stellar
...
-    , mValid(std::nullopt)
-    , mPreviousLedgerHash(previousLedgerHash)
+    : TxSetFrame(previousLedgerHash, Transactions{})
 {
 }
 
 TxSetFrame::TxSetFrame(Hash const& networkID, TransactionSet const& xdrSet)
-    : mHash(std::nullopt), mValid(std::nullopt)
+    : mPreviousLedgerHash(xdrSet.previousLedgerHash)
+    , mTxsInHashOrder(TxSetUtils::sortTxsInHashOrder(networkID, xdrSet))
```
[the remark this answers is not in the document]
MonsieurNicolas: You probably need a dedicated function that allows to normalize things: we should really be explicit when we use this in production code.
It's fine if we construct another `TxSetFrame` (instead of "in place" normalization with the old code when calling `sortForHash`)

src/herder/TxSetUtils.cpp
```diff
@@ -0,0 +1,393 @@
...
+}
+
+TxSetFrame::Transactions
+getInvalidTxList(Application& app, TxSetFrame const& txSet,
+                 uint64_t lowerBoundCloseTimeOffset,
+                 uint64_t upperBoundCloseTimeOffset,
+                 bool returnEarlyOnFirstInvalidTx)
+{
+    ZoneScoped;
+    LedgerTxn ltx(app.getLedgerTxnRoot());
```
MonsieurNicolas, line 223: I'm surprised that we don't create a "read only" LedgerTxn here - @marta-lokhova did we miss this earlier?

MonsieurNicolas (changes requested):
I only did a cursory review of the test code as things are probably going to change again based on my other comments.

src/herder/TxSetFrame.h
```diff
@@ -17,105 +16,102 @@ namespace stellar
 class Application;
 
 class TxSetFrame;
-typedef std::shared_ptr<TxSetFrame> TxSetFramePtr;
-typedef std::shared_ptr<TxSetFrame const> TxSetFrameConstPtr;
+using TxSetFrameConstPtr = std::shared_ptr<TxSetFrame const>;
```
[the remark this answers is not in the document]
jayz22: I'm not sure what's the best practice on this, although moving it in the class adds a bit overhead of typing out the namespace, i.e. `TxSetFrame::TxSetFrameConstPtr` everywhere.

src/herder/TxSetFrame.h
```diff
@@ -17,105 +16,102 @@ namespace stellar
...
 
-    void
-    add(TransactionFrameBasePtr tx)
+class TxSetInApplyOrder : public TxSetFrame
+{
+  public:
+    TxSetInApplyOrder(Hash const& previousLedgerHash,
+                      Transactions const& transactions)
+        : TxSetFrame(previousLedgerHash, transactions)
+        , mTxsInApplyOrder(transactions)
```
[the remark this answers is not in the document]
jayz22: Per my understanding, this class is used purely for testing/simulation (`TxSimApplyTransactionsWork::onRun()`), and it wants to preserve the original order in which the transactions were generated during the apply (in other words, to "cheat out" the apply order). I don't think the input is intended to be pre-sorted in the apply order and we probably shouldn't computed it here.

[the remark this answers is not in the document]
jayz22: Didn't see @[MonsieurNicolas](https://github.com/MonsieurNicolas)'s comment, yeah I will move this out of the class and rename it.

src/herder/TxSetUtils.cpp
marta-lokhova: not following this comment, can you clarify? I haven't been reviewing this PR, so I need context.

MonsieurNicolas: I thought we were supposed to use `TransactionMode::READ_ONLY_WITHOUT_SQL_TXN` in places like this

jayz22:
I've broken down previous single commit into three separate commits and made a new push. Same code, no new logic. Hopefully this makes reviewing much easier. I will address the other review comments in subsequent commits.

src/herder/TxSetUtils.cpp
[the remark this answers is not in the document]
jayz22: Fixed. Apologize for the inconvenience.

src/herder/TxSetFrame.h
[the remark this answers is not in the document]
jayz22: Moved the class back to `TxSimSetFrame.h` and renamed it to `SimApplyOrderTxSetFrame`

src/herder/HerderImpl.cpp
```diff
@@ -213,10 +214,11 @@ HerderImpl::newSlotExternalized(bool synchronous, StellarValue const& value)
     mLastExternalize = mApp.getClock().now();
 
     // perform cleanups
-    TxSetFramePtr externalizedSet = mPendingEnvelopes.getTxSet(value.txSetHash);
+    TxSetFrameConstPtr externalizedSet =
```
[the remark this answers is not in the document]
jayz22: Fixed.

src/herder/TxSetUtils.cpp
```diff
@@ -0,0 +1,393 @@
+// Copyright 2022 Stellar Development Foundation and contributors. Licensed
```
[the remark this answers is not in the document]
jayz22: Fixed.

src/herder/HerderImpl.cpp
```diff
@@ -1097,11 +1100,13 @@ HerderImpl::triggerNextLedger(uint32_t ledgerSeqToTrigger,
     upperBoundCloseTimeOffset = nextCloseTime - lcl.header.scpValue.closeTime;
     lowerBoundCloseTimeOffset = upperBoundCloseTimeOffset;
 
-    auto removed = proposedSet->trimInvalid(mApp, lowerBoundCloseTimeOffset,
-                                            upperBoundCloseTimeOffset);
-    mTransactionQueue.ban(removed);
+    auto removed = TxSetUtils::getInvalidTxList(
```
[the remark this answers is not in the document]
jayz22: Fixed.

src/herder/HerderImpl.cpp
```diff
@@ -1791,9 +1796,10 @@ HerderImpl::updateTransactionQueue(
     lhhe.hash = HashUtils::random();
     auto txSet = mTransactionQueue.toTxSet(lhhe);
 
-    auto removed = txSet->trimInvalid(
-        mApp, 0,
-        getUpperBoundCloseTimeOffset(mApp, lhhe.header.scpValue.closeTime));
+    auto removed = TxSetUtils::getInvalidTxList(
```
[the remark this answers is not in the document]
jayz22: Fixed.

src/herder/TxSetFrame.cpp
```diff
@@ -33,23 +34,26 @@ namespace stellar
 
 using namespace std;
 
+TxSetFrame::TxSetFrame(Hash const& previousLedgerHash,
```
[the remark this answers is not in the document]
jayz22: The `networkID` is only used for `TransactionFrame` to compute hash? 
This constructor takes a `vector<TransactionFrame>` directly, which already contains the `networkID` in them.

src/herder/TxSetFrame.cpp
```diff
@@ -149,357 +160,50 @@ TxSetFrame::sortForApply()
...
     if (this->size(lcl.header) > lcl.header.maxTxSetSize)
     {
         CLOG_DEBUG(Herder, "Got bad txSet: too many txs {} > {}",
                    this->size(lcl.header), lcl.header.maxTxSetSize);
-        mValid = std::make_optional<std::pair<Hash, bool>>(lcl.hash, false);
         return false;
     }
 
-    if (!std::is_sorted(mTransactions.begin(), mTransactions.end(),
+    if (!std::is_sorted(mTxsInHashOrder.begin(), mTxsInHashOrder.end(),
```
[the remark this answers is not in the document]
jayz22: If I understood correctly, this check is necessary to make sure the `TransactionSet` we receive is in the correct hash order. 
Thus sounds like this check should happen during construction via `TxSetFrame::TxSetFrame(Hash const& networkID, TransactionSet const& xdrSet)`?  Should I remove this check here and move to the constructor, i.e. if we received a bad `xdrSet` we store the status in a flag?

src/herder/TxSetFrame.cpp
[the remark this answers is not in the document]
jayz22: It's not very clear to me what should be done here. 
My understanding here, and from the two "check sorted" discussions [above1](https://github.com/stellar/stellar-core/pull/3395#discussion_r857779243) and [above2](https://github.com/stellar/stellar-core/pull/3395#discussion_r857789597), it is possible we might receive a bad `xdrSet` (out of hash order), in which case, we should not normalize it and make it valid (and altering the set and introducing malleability attack)? I'm not getting the "dedicated function that allows to normalize things" part.

src/herder/TxSetUtils.cpp
marta-lokhova: We're currently using `TransactionMode::READ_ONLY_WITHOUT_SQL_TXN` on the transaction processing path (there's a significant perf impact because that path is called hundreds of times per second). We could call it on this path as well, though I'm not sure we'll see the benefits (we call this function for unique SCP messages it looks like)

src/herder/TxSetFrame.cpp
```diff
@@ -149,357 +160,50 @@ TxSetFrame::sortForApply()
...
-                       uint64_t upperBoundCloseTimeOffset)
+                       uint64_t upperBoundCloseTimeOffset) const
 {
     ZoneScoped;
     auto& lcl = app.getLedgerManager().getLastClosedLedgerHeader();
-    if (mValid && mValid->first == lcl.hash)
-    {
-        return mValid->second;
-    }
+
```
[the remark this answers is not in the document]
jayz22: I've added a RandomEvictionCache in the `HerderSCPDriver` to cache the validity of txSet, as you and [MonsieurNicolas](https://github.com/MonsieurNicolas) both suggested. Although I feel my hash key computation might be a bit ad-hoc and I would like to know if there is a better approach.

src/herder/TxSetFrame.cpp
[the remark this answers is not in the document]
MonsieurNicolas: Normalization is in the context of tests (maybe also in the context of TransactionQueue), as we construct a `TxSetFrame` from a vector of transactions, and we just want to construct a valid `TxSetFrame` in that case

src/herder/TxSetFrame.cpp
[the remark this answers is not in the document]
MonsieurNicolas: You mean cache the result from `std::is_sorted` (in the constructor) and use it in the right places? Yeah that would work.

src/herder/TxSetFrame.cpp
MonsieurNicolas: we can just keep the `releaseAssert` but check that the `mIsSorted` is true (you're just caching the result of `std::is_sorted`

src/herder/TxSetUtils.cpp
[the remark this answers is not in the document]
MonsieurNicolas: thank you!

src/herder/TxSetUtils.cpp
MonsieurNicolas: yeah perf wise would not have much impact, I was thinking about this as a strong signal that nothing changes and we're performing read only operations as in, the code ends up being easier for people to understand (and impossible to mess up)

src/herder/TxSetUtils.h
```diff
@@ -48,5 +50,32 @@ TxSetFrameConstPtr removeTxs(TxSetFrameConstPtr txSet,
...
+            tbs.extend_at(i) = (std::get<2>(key) >> (8 * i)) & 0xff;
+        }
+        for (size_t i = 0; i < 8; ++i)
+        {
+            tbs.extend_at(i + 8) = (std::get<3>(key) >> (8 * i)) & 0xff;
+        }
+        hasher.add(tbs);
+        return std::hash<Hash>()(hasher.finish());
+    }
+};
```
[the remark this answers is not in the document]
MonsieurNicolas: It's complicated because you're using crypto hashes.
For hash functions in `unsorted_map`, we use short hashes.

See https://github.com/stellar/stellar-core/blob/9dffed456d837e9e3e0fb937865278562e2671b2/src/ledger/LedgerHashUtils.h#L116 that you modified before.

src/herder/HerderSCPDriver.cpp
```diff
@@ -33,6 +33,8 @@
 namespace stellar
 {
 
+const uint32_t TXSETVALID_CACHE_SIZE = 10000;
```
MonsieurNicolas, line 36: uint32 const

MonsieurNicolas (changes requested):
better!

src/herder/test/HerderTests.cpp
jayz22: I've re-enabled this test to cover the case of an out-of-order `TransactionSet`.

src/herder/test/TransactionQueueTests.cpp
jayz22: This part should still be removed. TxSet constructed from the `TransactionQueue` is hash sorted, so the original indices won't work.

src/herder/TxSetFrame.cpp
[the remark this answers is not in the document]
jayz22: I've updated the `TxSetFrame` class to cache result from the `std::is_sorted`. So now, the TxSet is either sorted in hash order on construction, if it is constructed from a `vector<TransactionFrameBasePtr>`, or it is invalid (if constructed from an `xdrSet` that is in the wrong order).

src/herder/TxSetFrame.cpp
jayz22: Updated.

src/herder/TxSetFrame.cpp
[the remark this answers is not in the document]
jayz22: @dmkozh and I previously discussed about having a builder helper class to construct the `TxSetFrame` from a `generalizedTransactionSet`, maybe that's something that'll make it easier. I'll be happy to discuss if there are other ways to make this easier for CAP-42.

src/herder/TxSetFrame.cpp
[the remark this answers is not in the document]
jayz22: Turns out we didn't really need a dedicated normalization method. The one place where I did such a normalization was done [here](https://github.com/stellar/stellar-core/blob/928947a06a10a8bc44daea96166ebe70c39befd0/src/herder/test/HerderTests.cpp#L336-L353), which is done by transforming `xdrSet` to `vector<TransactionFrameBasePtr>` and call `TxSetFrame` constructor on that.

src/herder/TxSetUtils.h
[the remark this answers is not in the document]
jayz22: Updated it to use fast hash.

src/herder/HerderSCPDriver.cpp
jayz22: Fixed.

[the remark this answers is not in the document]
jayz22: Fixed.

src/herder/TxSetUtils.cpp
jayz22: I think makes sense. I've updated it.

src/herder/test/HerderTests.cpp
```diff
@@ -274,138 +287,145 @@ testTxSet(uint32 protocolVersion)
...
+            TransactionSet xdrSet;
+            txSet->toXDR(xdrSet);
+            // rearrange it out of order
+            std::swap(xdrSet.txs[0], xdrSet.txs[1]);
+            txSet =
+                std::make_shared<TxSetFrame const>(app->getNetworkID(), xdrSet);
             REQUIRE(!txSet->checkValid(*app, 0, 0));
 
-            txSet->trimInvalid(*app, 0, 0);
+            // reconstruct it from TxSetFrame::Transactions, which automatically
```
MonsieurNicolas, line 346: the goal here would be to test whatever code we're using in production when we perform `trimInvalid` --> another reason to have a `trimInvalid` equivalent

src/herder/TxSetFrame.h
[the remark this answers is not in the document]
dmkozh: I'm not suggesting to move this *into* the class, just move the declaration below the class and remove the forward declaration. In general I would avoid forward declarations when possible (i.e. when there is no circular dependency)

src/herder/TxSetFrame.h
```diff
@@ -7,115 +7,88 @@
...
+    TxSetFrame(TxSetFrame const& other) = default;
+
     virtual ~TxSetFrame(){};
 
     // returns the hash of this tx set
-    Hash const& getContentsHash() override;
+    Hash const&
+    getContentsHash() const
+    {
+        releaseAssert(mTxsInValidHashOrder);
```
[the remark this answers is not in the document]
dmkozh: I'm decoupling xdr validation from `TxSetFrame` construction in my CAP-42 PR (as it doesn't make sense to even try building a `TxSetFrame` around some malformed message). Maybe it should happen here instead, so that we have an assert to ensure that we always build correct frames.

src/herder/TxSetFrame.h
```diff
@@ -7,115 +7,88 @@
...
+        return mHash;
+    }
 
-    Hash& previousLedgerHash();
-    Hash const& previousLedgerHash() const override;
+    Hash const&
+    previousLedgerHash() const
```
dmkozh, line 48: nit; Is there some benefit of inlining the implementation into header? IMHO this both makes header harder to read (as it's longer) and also harder to modify (if I want to add something I'd need to move the implementation back to cpp file)

src/herder/TxSetFrame.cpp
```diff
@@ -33,43 +34,34 @@ namespace stellar
...
-    // need to use the hash of whole tx here since multiple txs could have
-    // the same Contents
-    return tx1->getFullHash() < tx2->getFullHash();
 }
 
-// order the txset correctly
-// must take into account multiple tx from same account
-void
-TxSetFrame::sortForHash()
+// Contruct a TxSetFrame from an xdrSet, which may directly come from the
```
dmkozh, line 54: typos: s/Contruct/Construct, s/comptue/compute

src/herder/TxSetFrame.cpp
```diff
@@ -33,43 +34,34 @@ namespace stellar
...
+// overlay and be invalid (in the incorrect order), in which case we mark
+// this set invalid and do not comptue its hash.
```
dmkozh, line 56: As I've mentioned in another comment, I would like to decouple XDR validation from construction; I don't think that validation in constructor is a good approach. I may take care of that in my PR though.

src/herder/TxSetFrame.cpp
jayz22: Thanks, fixed.

src/herder/TxSetFrame.h
jayz22: Moved out.

src/herder/test/HerderTests.cpp
```diff
@@ -1846,12 +1825,12 @@ testSCPDriver(uint32 protocolVersion, uint32_t maxTxSetSize, size_t expectedOps)
...
                                      : SCPDriver::kInvalidValue));
 
-                // Confirm that trimInvalid() as used by
+                // Confirm that getTxTrimList() as used by
                 // HerderImpl::triggerNextLedger() trims the transaction if and
                 // only if we expect it to be invalid.
                 auto closeTimeOffset = nextCloseTime - lclCloseTime;
                 auto removed =
-                    txSet->trimInvalid(*app, closeTimeOffset, closeTimeOffset);
+                    trimInvalid(*app, txSet, closeTimeOffset, closeTimeOffset);
```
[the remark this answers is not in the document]
jayz22: Updated.

src/herder/TxSetUtils.h
```diff
@@ -0,0 +1,82 @@
...
+
+namespace stellar
+{
+namespace TxSetUtils
+{
+
+bool HashTxSorter(TransactionFrameBasePtr const& tx1,
+                  TransactionFrameBasePtr const& tx2);
+
+bool TxsInValidHashOrder(TxSetFrame::Transactions const& txs);
```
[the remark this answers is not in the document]
jayz22: Updated.

src/herder/TxSetUtils.cpp
```diff
@@ -0,0 +1,399 @@
...
+                   txSet->getTxsInHashOrder().cend(),
+                   std::inserter(fullTxSetHashMap, fullTxSetHashMap.end()),
+                   [](TransactionFrameBasePtr const& tx) {
+                       return std::pair<Hash, TransactionFrameBasePtr>{
+                           tx->getFullHash(), tx};
+                   });
+
+    // candidate tx hashes to be removed
+    std::unordered_set<Hash> removeTxHashSet;
+    removeTxHashSet.reserve(txsToRemove.size());
```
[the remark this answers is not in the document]
jayz22: Yeah I must've been thinking some weird cases.. Cleaned it up.

src/herder/TxSetUtils.h
```diff
@@ -0,0 +1,82 @@
...
+// returnEarlyOnFirstInvalidTx is true, return immediately if an invalid
+// transaction is found (instead of finding all of them), this is useful for
+// checking if a TxSet is valid.
+TxSetFrame::Transactions getInvalidTxList(Application& app,
+                                          TxSetFrame const& txSet,
+                                          uint64_t lowerBoundCloseTimeOffset,
+                                          uint64_t upperBoundCloseTimeOffset,
+                                          bool returnEarlyOnFirstInvalidTx);
+
+TxSetFrameConstPtr removeTxs(TxSetFrameConstPtr txSet,
```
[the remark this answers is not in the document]
jayz22: This is a good point. 
I've added back the `trimInvalid` method, combining the `getInvalidTxList` and `removeTxs` (and removed `removeTxs` to test-only header). 
However, the main difference is the old `trimInvalid` first calls `sortForHash`, which shouldn't be necessary here because is called on `txSet` constructed by us (`mTransactionQueue`) so it is guaranteed to be in the correct order.

src/herder/TxSetUtils.h
```diff
@@ -0,0 +1,82 @@
...
+                             TxSetFrame::Transactions const& txsToRemove);
+
+TxSetFrameConstPtr addTxs(TxSetFrameConstPtr txSet,
```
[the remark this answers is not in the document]
jayz22: Moved.

src/herder/TxSetUtils.h
```diff
@@ -0,0 +1,82 @@
...
+#include <herder/TxSetFrame.h>
+#include <ledger/LedgerHashUtils.h>
+#include <tuple>
+
+namespace stellar
+{
+namespace TxSetUtils
+{
+
+bool HashTxSorter(TransactionFrameBasePtr const& tx1,
```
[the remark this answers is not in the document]
jayz22: Removed.

src/herder/TxSetUtils.h
```diff
@@ -0,0 +1,82 @@
+// Copyright 2022 Stellar Development Foundation and contributors. Licensed
+// under the Apache License, Version 2.0. See the COPYING file at the root
+// of this distribution or at http://www.apache.org/licenses/LICENSE-2.0
+
+#pragma once
+
+#include "util/UnorderedMap.h"
+#include "xdr/Stellar-types.h"
+#include <deque>
```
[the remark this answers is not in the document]
jayz22: Removed.

jayz22:
Addressed all comments except a few tests comments which are less trivial, will need more time to think about it.

src/herder/TxSetUtils.h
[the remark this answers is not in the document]
jayz22: Fixed.

src/herder/TxSetUtils.h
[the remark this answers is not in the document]
jayz22: Fixed.

src/herder/HerderSCPDriver.h
```diff
@@ -190,9 +193,14 @@ class HerderSCPDriver : public SCPDriver
...
 
+    // validity of txSet
+    RandomEvictionCache<TxSetUtils::TxSetValidityKey, bool,
+                        TxSetUtils::TxSetValidityKeyHash>
+        mTxSetValidCache;
+
     SCPDriver::ValidationLevel validateValueHelper(uint64_t slotIndex,
                                                    StellarValue const& sv,
-                                                   bool nomination) const;
+                                                   bool nomination);
```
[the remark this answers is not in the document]
jayz22: Fixed. That makes sense!

src/herder/TxSetUtils.h
```diff
@@ -0,0 +1,82 @@
...
+#include <herder/TxSetFrame.h>
```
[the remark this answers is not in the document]
jayz22: Fixed.

src/herder/test/HerderTests.cpp
```diff
@@ -274,138 +287,145 @@ testTxSet(uint32 protocolVersion)
...
     };
-
     for (size_t i = 0; i < nbAccounts; i++)
     {
         genTx(nbTransactions);
     }
-
-    txSet->previousLedgerHash() =
-        app->getLedgerManager().getLastClosedLedgerHeader().hash;
+    txSet = std::make_shared<TxSetFrame const>(
```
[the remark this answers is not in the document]
jayz22: This test section is a bit of messy. Cleaned it up.

src/herder/test/HerderTests.cpp
```diff
@@ -274,138 +287,145 @@ testTxSet(uint32 protocolVersion)
...
-        while (txSet->mTransactions.size() <=
-               cfg.TESTING_UPGRADE_MAX_TX_SET_SIZE)
+        while (txSet->sizeTx() <= cfg.TESTING_UPGRADE_MAX_TX_SET_SIZE)
         {
             genTx(1);
         }
-        txSet->previousLedgerHash() =
-            app->getLedgerManager().getLastClosedLedgerHeader().hash;
-        txSet->sortForHash();
+        txSet = std::make_shared<TxSetFrame const>(
```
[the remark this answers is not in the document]
jayz22: Cleaned up.

src/herder/test/HerderTests.cpp
jayz22: Please see my earlier [comment](https://github.com/stellar/stellar-core/pull/3395#discussion_r862284708) on the `trimInvalid`.  `trimInvalid` does not sort-in-hash any more (the txSet was supposed to be sorted on construction), thus the scenario described here doesn't match the actual production. The closest thing to it is here where we take the out-of-order vector and reconstruct the `TxSet` from it to get a valid one.

src/herder/test/TransactionQueueTests.cpp
jayz22: Fixed. Although I stared at this test for a long time and still not 100% clear what it's trying to do.

jayz22:
Addressed all comments, ready for the next round.

src/herder/TxSetFrame.cpp
```diff
@@ -33,43 +34,51 @@ namespace stellar
...
 }
 
+// Construct a TxSetFrame from an xdrSet, which may directly come from the
+// overlay and be invalid (in the incorrect order), in which case we mark
+// this set invalid and do not compute its hash.
 TxSetFrame::TxSetFrame(Hash const& networkID, TransactionSet const& xdrSet)
-    : mHash(std::nullopt), mValid(std::nullopt)
+    : mPreviousLedgerHash(xdrSet.previousLedgerHash)
+    , mTxs(TxSetUtils::extractTxsFromXdrSet(networkID, xdrSet))
+    , mTxsInValidHashOrder(TxSetUtils::isInValidHashOrder(mTxs))
```
MonsieurNicolas, line 60: nit: just realized that the name `mTxsInValidHashOrder` kinda reads too close to `mTxsInvalidHashOrder`, so we should make it `mTxsIsValidHashOrder` (same with `TxSetUtils::isInValidHashOrder` that should be `TxSetUtils::isValidHashOrder`)

src/herder/TxSetUtils.cpp
```diff
@@ -0,0 +1,389 @@
...
+        }
+    }
+
+    return invalidTxs;
+}
+
+// Target use case is to remove a subset of invalid transactions from a TxSet.
+// I.e. txSet.size() >= txsToRemove.size()
+TxSetFrameConstPtr
+removeTxs(TxSetFrameConstPtr txSet, TxSetFrame::Transactions const& txsToRemove)
```
MonsieurNicolas, line 343: mark as "static" or in the `{}` namespace as to not export

src/herder/TxSetUtils.cpp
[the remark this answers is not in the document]
MonsieurNicolas: I didn't see you were looping again at the end... so a better algo (which also will work better with GeneralizedTransactions groups) is:
* create an `UnorderedSet` for `txsToRemove` (note: we use that instead of `std::unordered_set` as much as possible)
* iterate over the input txset and construct `txs` by excluding transactions that are found in `txsToRemove`

less copying, less code.

src/herder/test/TestTxSetUtils.h
```diff
@@ -0,0 +1,24 @@
...
+// under the Apache License, Version 2.0. See the COPYING file at the root
+// of this distribution or at http://www.apache.org/licenses/LICENSE-2.0
+
+#pragma once
+
+#include "herder/TxSetUtils.h"
+
+namespace stellar
+{
+namespace TxSetUtils
```
MonsieurNicolas, line 11: do not use the prod namespace, just call it `TxSetTestUtils`

src/herder/TxSetUtils.cpp
MonsieurNicolas: actually, I see that you did something odd wrt test code. I am not sure if there is a way to do this with namespaces, but with classes this would done by having that defined as a private static method and have the test class defined as a "friend"

MonsieurNicolas: we want to make it basically impossible to call this function from production code outside of this file, and what you did does not guarantee this and is definitely not a usual pattern

src/herder/test/PendingEnvelopesTests.cpp
```diff
@@ -323,12 +325,27 @@ TEST_CASE("PendingEnvelopes recvSCPEnvelope", "[herder]")
...
         // Make sure to discard the envelope
         REQUIRE(pendingEnvelopes.recvSCPEnvelope(envNoSign) ==
                 Herder::ENVELOPE_STATUS_DISCARDED);
     }
+
+    SECTION("receiving malformed txset would crash but disabled check")
+    {
+        auto malformedTxSet =
+            TxSetUtils::makeIllSortedTxSet(app->getNetworkID(), txSet);
+        auto p = makeTxPair(malformedTxSet, 10, STELLAR_VALUE_SIGNED);
```
MonsieurNicolas, line 341: warning: `p` is shadowing

MonsieurNicolas (changes requested):
a few minor things left. Please address, rebase and squash so that we can merge

src/herder/TxSetUtils.cpp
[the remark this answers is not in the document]
jayz22: Fixed. Yeah I agree this is better.

src/herder/test/PendingEnvelopesTests.cpp
jayz22: Fixed.

src/herder/test/TestTxSetUtils.h
jayz22: Fixed. See above.

jayz22:
Converting this to draft while I'm sorting through the commit orders before squashing.

src/herder/TxSetUtils.cpp
```diff
@@ -0,0 +1,379 @@
+// Copyright 2022 Stellar Development Foundation and contributors. Licensed
```
MonsieurNicolas, line 1: I don't know what you did when you rebased... but you lost your local changes to the visual studio files

MonsieurNicolas:
```
diff --git a/Builds/VisualStudio/stellar-core.vcxproj b/Builds/VisualStudio/stellar-core.vcxproj
index 3c01bbf08..f3c8eaa9f 100644
--- a/Builds/VisualStudio/stellar-core.vcxproj
+++ b/Builds/VisualStudio/stellar-core.vcxproj
@@ -310,6 +310,7 @@ exit /b 0
     <ClCompile Include="..\..\src\herder\PendingEnvelopes.cpp" />
     <ClCompile Include="..\..\src\herder\simulation\TxSimTxSetFrame.cpp" />
     <ClCompile Include="..\..\src\herder\SurgePricingUtils.cpp" />
+    <ClCompile Include="..\..\src\herder\test\TestTxSetUtils.cpp" />
     <ClCompile Include="..\..\src\herder\TransactionQueue.cpp" />
     <ClCompile Include="..\..\src\herder\QuorumTracker.cpp" />
     <ClCompile Include="..\..\src\herder\test\HerderTests.cpp" />
@@ -319,6 +320,7 @@ exit /b 0
     <ClCompile Include="..\..\src\herder\test\UpgradesTests.cpp" />
     <ClCompile Include="..\..\src\herder\TxQueueLimiter.cpp" />
     <ClCompile Include="..\..\src\herder\TxSetFrame.cpp" />
+    <ClCompile Include="..\..\src\herder\TxSetUtils.cpp" />
     <ClCompile Include="..\..\src\herder\Upgrades.cpp" />
     <ClCompile Include="..\..\src\historywork\BatchDownloadWork.cpp" />
     <ClCompile Include="..\..\src\herder\QuorumIntersectionCheckerImpl.cpp" />
@@ -703,10 +705,12 @@ exit /b 0
     <ClInclude Include="..\..\src\herder\PendingEnvelopes.h" />
     <ClInclude Include="..\..\src\herder\simulation\TxSimTxSetFrame.h" />
     <ClInclude Include="..\..\src\herder\SurgePricingUtils.h" />
+    <ClInclude Include="..\..\src\herder\test\TestTxSetUtils.h" />
     <ClInclude Include="..\..\src\herder\TransactionQueue.h" />
     <ClInclude Include="..\..\src\herder\QuorumTracker.h" />
     <ClInclude Include="..\..\src\herder\TxQueueLimiter.h" />
     <ClInclude Include="..\..\src\herder\TxSetFrame.h" />
+    <ClInclude Include="..\..\src\herder\TxSetUtils.h" />
     <ClInclude Include="..\..\src\herder\Upgrades.h" />
     <ClInclude Include="..\..\src\historywork\BatchDownloadWork.h" />
     <ClInclude Include="..\..\src\herder\QuorumIntersectionChecker.h" />
diff --git a/Builds/VisualStudio/stellar-core.vcxproj.filters b/Builds/VisualStudio/stellar-core.vcxproj.filters
index 1e8c50c7f..c3b64393b 100644
--- a/Builds/VisualStudio/stellar-core.vcxproj.filters
+++ b/Builds/VisualStudio/stellar-core.vcxproj.filters
@@ -1233,6 +1233,12 @@
     <ClCompile Include="src\generated\rust\bridge.cpp">
       <Filter>rust\generated</Filter>
     </ClCompile>
+    <ClCompile Include="..\..\src\herder\test\TestTxSetUtils.cpp">
+      <Filter>herder\tests</Filter>
+    </ClCompile>
+    <ClCompile Include="..\..\src\herder\TxSetUtils.cpp">
+      <Filter>herder</Filter>
+    </ClCompile>
   </ItemGroup>
   <ItemGroup>
     <ClInclude Include="..\..\lib\util\cpptoml.h">
@@ -2144,6 +2150,12 @@
     <ClInclude Include="src\generated\rust\bridge.h">
       <Filter>rust\generated</Filter>
     </ClInclude>
+    <ClInclude Include="..\..\src\herder\test\TestTxSetUtils.h">
+      <Filter>herder\tests</Filter>
+    </ClInclude>
+    <ClInclude Include="..\..\src\herder\TxSetUtils.h">
+      <Filter>herder</Filter>
+    </ClInclude>
   </ItemGroup>
   <ItemGroup>
     <None Include="..\..\AUTHORS" />
```

MonsieurNicolas: I suspect it's that you had merge conflicts and just discarded your own changes. Always do a diff between your old branch and new rebased branch (that's how I found this right away).

jayz22:
Finished combing through commit history, ready for final review and merge.

MonsieurNicolas (changes requested):
Aside from git history cleanup and the bad merge issue, this looks pretty much done

src/herder/TxSetUtils.cpp
jayz22: I cleaned up the git history and added back the Visual Studio files. 
Re: the missing file, I remember seeing the conflict on the .vcxproj files, but my the diff tool on my vscode showed no difference, so I just accepted the incoming change. Super weird. Anyways, it's fixed now and will keep an eye out in the future.

MonsieurNicolas:
r+ 9bc2231da54f5fc8bc51714aa35cad65fd587aa8

PR #3396, opened 2022-04-11
Remove zipped buckets as soon as they are published

### LINKED ISSUE #2507: Publish sometimes leaves stale files

### LINKED ISSUE #3394: snapshots not properly cleaned up after publish
We noticed that even long after publish, files from the last snapshot are still on disk (in this case ~ 5 minutes after the previous checkpoint completed):

```
grep -i publish /var/log/stellar/stellar-core.log ; ls *.gz ; du -sh tmp/snapshot*                           22:45:17
2022-04-08T22:06:25.597 GCGB2 [History INFO] Publishing 1 queued checkpoints [40387135-40387135]: Ready to run: prepare-snapshot
2022-04-08T22:06:31.427 GCGB2 [History INFO] Publishing 1 queued checkpoints [40387135-40387135]: Waiting: prepare-snapshot
2022-04-08T22:06:41.195 GCGB2 [History INFO] Publishing 1 queued checkpoints [40387135-40387135]: update-archives-0268423f:uploading files
2022-04-08T22:12:52.791 GCGB2 [History INFO] Publishing 1 queued checkpoints [40387199-40387199]: Ready to run: prepare-snapshot
2022-04-08T22:12:58.241 GCGB2 [History INFO] Publishing 1 queued checkpoints [40387199-40387199]: Waiting: prepare-snapshot
2022-04-08T22:19:24.859 GCGB2 [History INFO] Publishing 1 queued checkpoints [40387263-40387263]: Ready to run: prepare-snapshot
2022-04-08T22:19:31.302 GCGB2 [History INFO] Publishing 1 queued checkpoints [40387263-40387263]: Waiting: prepare-snapshot
2022-04-08T22:19:39.666 GCGB2 [History INFO] Publishing 1 queued checkpoints [40387263-40387263]: update-archives-026842bf:uploading files
2022-04-08T22:25:58.776 GCGB2 [History INFO] Publishing 1 queued checkpoints [40387327-40387327]: Ready to run: prepare-snapshot
2022-04-08T22:26:03.378 GCGB2 [History INFO] Publishing 1 queued checkpoints [40387327-40387327]: Waiting: prepare-snapshot
2022-04-08T22:32:39.442 GCGB2 [History INFO] Publishing 1 queued checkpoints [40387391-40387391]: Ready to run: prepare-snapshot
2022-04-08T22:32:45.932 GCGB2 [History INFO] Publishing 1 queued checkpoints [40387391-40387391]: Waiting: prepare-snapshot
2022-04-08T22:39:08.139 GCGB2 [History INFO] Publishing 1 queued checkpoints [40387455-40387455]: Ready to run: prepare-snapshot
2022-04-08T22:39:15.691 GCGB2 [History INFO] Publishing 1 queued checkpoints [40387455-40387455]: Waiting: prepare-snapshot
bucket-0482eb898f740978272681922b881ee38bff22c7cb5c26da32126a77e8a63a93.xdr.gz  bucket-8c68d3fa6b41fbc962e7548fdaeb4444f5422dd0800cefb505b7c81372e8b9b3.xdr.gz
bucket-0afcbed52b98cd2dcd27fdc1885ff27eccf38770d82ac18f63a18d3d15d33e99.xdr.gz  bucket-9e5b01b50563bb3bec93ad1217abf6459652745bd8a0a2f5f7183312d139521b.xdr.gz
bucket-2204cc91211b53dc6ba56bf62f2278fc77a371cddb56daf92ac5af2e6356b9cf.xdr.gz  bucket-9fd894efb51fdbed5aedf9509e4f6600c829461117dd400c8017e42674a076d7.xdr.gz
bucket-34fcf23ac79dcc253e30f58a93f70deda8a04639755e994a3bd685ba6864ae9a.xdr.gz  bucket-a18409096181514428ae0bde65664a005e246e32142ee2b9e936bf33cdaadca0.xdr.gz
bucket-380c5adeb45bf53a37c54f678177a1a6c7a37b3b3ae00280e635a1aee821731f.xdr.gz  bucket-c70ee02104b2874273988906adce30a4b4e290d400893780f5282db94d224eee.xdr.gz
bucket-6ceecd9a7c8963e29559abca97e631aa9ed95abe459163ee63be3c2e6802280a.xdr.gz  bucket-cf523812e392e05fc2ce6b34a3abc289b6cc968d0c851bdd372d9c0e948a727b.xdr.gz
bucket-87a5a8f73ce20902b7df49d63d0813c9acfa98832f5ea1d7dcf053ecea1f95da.xdr.gz
29M     tmp/snapshot-85e62e392ad2b627
```

I would expect those files to be cleaned up right after publish finishes.

Not sure if this is related, but not all checkpoints emit things like `update-archives-026842bf:uploading files` even though snapshots are published (but this could just be a logging bug).

marta-lokhova:
Resolves #3394

### DIFF
--- a/src/historywork/PutSnapshotFilesWork.cpp
+++ b/src/historywork/PutSnapshotFilesWork.cpp
@@ -18,6 +18,16 @@
 namespace stellar
 {
 
+void
+PutSnapshotFilesWork::cleanup()
+{
+    // Delete `gz` files produced by this work
+    for (auto const& f : mFilesToUpload)
+    {
+        std::remove(f.second.localPath_gz().c_str());
+    }
+}
+
 PutSnapshotFilesWork::PutSnapshotFilesWork(
     Application& app, std::shared_ptr<StateSnapshot> snapshot)
     : Work(app,
@@ -73,10 +83,7 @@ PutSnapshotFilesWork::doWork()
         if (status == State::WORK_SUCCESS)
         {
             // Step 2: Gzip all unique files
-            for (auto const& f : getFilesToZip())
-            {
-                mGzipFilesWorks.emplace_back(addWork<GzipFileWork>(f, true));
-            }
+            createGzipWorks();
             return State::WORK_RUNNING;
         }
         else
@@ -99,13 +106,16 @@ PutSnapshotFilesWork::doWork()
 void
 PutSnapshotFilesWork::doReset()
 {
+    cleanup();
+
     mGetStateWorks.clear();
     mGzipFilesWorks.clear();
     mUploadSeqs.clear();
+    mFilesToUpload.clear();
 }
 
-UnorderedSet<std::string>
-PutSnapshotFilesWork::getFilesToZip()
+void
+PutSnapshotFilesWork::createGzipWorks()
 {
     // Sanity check: there are states for all archives
     if (mGetStateWorks.size() !=
@@ -114,17 +124,18 @@ PutSnapshotFilesWork::getFilesToZip()
         throw std::runtime_error("Corrupted GetHistoryArchiveStateWork");
     }
 
-    UnorderedSet<std::string> filesToZip{};
     for (auto const& getState : mGetStateWorks)
     {
         for (auto const& f :
              mSnapshot->differingHASFiles(getState->getHistoryArchiveState()))
         {
-            filesToZip.insert(f->localPath_nogz());
+            if (mFilesToUpload.emplace(f->localPath_nogz(), *f).second)
+            {
+                mGzipFilesWorks.emplace_back(
+                    addWork<GzipFileWork>(f->localPath_nogz(), true));
+            }
         }
     }
-
-    return filesToZip;
 }
 
 std::string
@@ -146,5 +157,11 @@ PutSnapshotFilesWork::getStatus() const
     }
 
     return BasicWork::getStatus();
+}
+
+void
+PutSnapshotFilesWork::onSuccess()
+{
+    cleanup();
 }
 }
--- a/src/historywork/PutSnapshotFilesWork.h
+++ b/src/historywork/PutSnapshotFilesWork.h
@@ -23,8 +23,10 @@ class PutSnapshotFilesWork : public Work
     std::list<std::shared_ptr<GetHistoryArchiveStateWork>> mGetStateWorks;
     std::list<std::shared_ptr<BasicWork>> mGzipFilesWorks;
     std::list<std::shared_ptr<BasicWork>> mUploadSeqs;
+    UnorderedMap<std::string, FileTransferInfo> mFilesToUpload;
 
-    UnorderedSet<std::string> getFilesToZip();
+    void createGzipWorks();
+    void cleanup();
 
   public:
     PutSnapshotFilesWork(Application& app,
@@ -36,5 +38,6 @@ class PutSnapshotFilesWork : public Work
   protected:
     State doWork() override;
     void doReset() override;
+    void onSuccess() override;
 };
 }


graydon:
I'm a little confused -- don't these files already get cleaned up (eventually) when some reference drops and destructor runs? I was expecting this change to just find an object that's being retained too long and reduce its lifecycle to a shorter duration.

marta-lokhova:
it doesn't seem to be the case for buckets, as they are not kept in temp folders (which are referenced by the publish work and cleaned up as soon as publish is done). Before this change, publish only reduced the refcount on the buckets published, but not cleanup the zipped files. What from what I see, `gz` files were cleaned up during GC: https://github.com/stellar/stellar-core/blob/9b54f2d9ea38caac0eabbdd8817af3e262ac6a89/src/bucket/BucketManagerImpl.cpp#L689 The problem is that if core continues referencing those buckets for some time, then the leftover `gz` files won't be cleaned up until there are no more references.

src/historywork/PutSnapshotFilesWork.cpp
```diff
@@ -102,6 +102,7 @@ PutSnapshotFilesWork::doReset()
     mGetStateWorks.clear();
     mGzipFilesWorks.clear();
     mUploadSeqs.clear();
+    bucketsToCleanup.clear();
```
MonsieurNicolas, line 105: should be called `mBucketsToCleanup`

src/historywork/PutSnapshotFilesWork.cpp
```diff
@@ -121,6 +122,12 @@ PutSnapshotFilesWork::getFilesToZip()
              mSnapshot->differingHASFiles(getState->getHistoryArchiveState()))
         {
             filesToZip.insert(f->localPath_nogz());
+            // Bucket files are tracked outside of the temp folder, so we
+            // manually collect buckets to clean up at the end
+            if (BucketManager::isBucketFile(f->baseName_gz()))
+            {
+                bucketsToCleanup.insert(f->localPath_gz());
```
MonsieurNicolas, line 129: it would be cleaner to return a `UnorderedMap<std::string, FileTransferInfo>` instead of `UnorderedSet<std::string>` and just store it in a member variable.

That way, your cleanup function can just delete all `localPath_gz()` and `localPath_gz_tmp()` associated with those. I don't think you need to only delete buckets: any file that got zipped can safely be deleted.

MonsieurNicolas (changes requested):

src/historywork/PutSnapshotFilesWork.cpp
marta-lokhova: other files are deleted automatically because they live in temp directories, so I don't think we need to do this.

marta-lokhova: we shouldn't need to delete `tmp` files here. Those are transient and owned by the unzip work. If they are left, that would indicate there's a bug in unzip work, that it's not cleaned up correctly (I verified that `GetAndUnzipRemoteFileWork` cleans up as expected).

src/historywork/PutSnapshotFilesWork.cpp
```diff
@@ -147,4 +154,13 @@ PutSnapshotFilesWork::getStatus() const
 
     return BasicWork::getStatus();
 }
+
+void
+PutSnapshotFilesWork::onSuccess()
```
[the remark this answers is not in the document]
marta-lokhova: On failure/abort, I expect work to automatically clean up after itself. `GzipFileWork` already deletes all zipped files in `onReset`, so I assumed if this Work fails, it resets children automatically and aborts. I think I was wrong about that though - checking the code now it looks we don't actually do it. I think we should (there might be other places where this is an issue).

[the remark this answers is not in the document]
MonsieurNicolas: I agree we should be looking into properly aborting in general, but that seems a bit more involved. We should open an issue about this.

src/historywork/PutSnapshotFilesWork.cpp
MonsieurNicolas: I think of the "zip files" (all of them) as resources that this work "inherits" on success (`GzipWork` on success stops tracking files) so it should be cleaning up after itself.

The fact that other files are downloaded in temp folders is a choice made somewhere elsee. The fact that they happen to be deleted by the caller is more of a "catch all" type of situation that happens to cleanup dangling files. The current work is the one that owns those files.

I would err on writing the simplest code possible here unless there are disadvantages.

Adding to this, I am not sure we gain much by trying to not delete files in the abort/failure case, which is why I was saying to just delete everything regardless of success/failure/abort.

We could *not* delete the tmp files, but that implies (I think) fixing the abort path. I would recommend deleting them for now, and fix abort in a subsequent PR.

MonsieurNicolas (changes requested):

marta-lokhova:
@MonsieurNicolas updated the change, and opened an issue to fix abort https://github.com/stellar/stellar-core/issues/3400

MonsieurNicolas (approved):
LGTM!

MonsieurNicolas:
r+ a2579ffab9a9ff17555799e2338409e551aee433

Issue #3402, opened 2022-04-19
Runtime Environment [jump cannon]


MonsieurNicolas:
- [x] make host object comparison deep
- [x] write some unit tests
- [x] wasmi - go no/go
   - [x] switch to stellar-specific wasmi fork
   - [x] merge wasmi integrated gas metering
   - [x] upgrade wasmi fork to the v1 API since they've deprecated the old one
- [x] add storage backend for contracts and contract data
- [x] prohibit FP ops or else normalize FP values and FP environment
- [ ] support contract "specs" / schemas / signatures / types
  - [ ] ~~add host object vocabulary to represent types as first class values~~
  - [ ] ~~add host functions for constructing types~~
  - [ ] ~~add host functions for checking value conformance to a type~~
  - [ ] check contract arguments match spec on invocation
  - [ ] check contract spec-compatibility on upgrade
- [x] event/logging subsystem #3487
- [x] record and export host metrics to core
- [x] tracy integration #3433
- [x] support host-crate multi-versioning in core: #3471
- [x] consolidate XDR definitions between C++ and Rust
  - [x] move XDR to submodule: #3470 
  - [x] check XDR consistency between Rust and C++: #3452
- [x] tighten "security" semantics
  - [x] never accept ScStatus values as input or store in objects
  - [x] enforce host-object read encapsulation: https://github.com/stellar/rs-stellar-contract-env/pull/132
  - [x] enforce non-reentry flag for cross-contract calls
- [x] update CAP-46 to reflect recent changes
  - [x] subtransactions and rollback
  - [x] vast expansion of ScStatusType and ScStatus
  - [x] renaming pos_i64 back to u63
  - [x] not-completely-empty section on metering 
  - [x] read encapsulation, non-reentry, scstatus prohibition
  - [x] replace "C++" with "rust" in various places
  - [x] update references to PRs and repos
- [x] #3488
- [x] security review
- [x] #3489 / https://github.com/stellar/rs-stellar-contract-env/issues/160
- [x] performance testing
  - [x] support & annotate tracy tracing in host crate
- [x] fuzzing

graydon:
Done as much as we're going to do (one section skipped).

PR #3413, opened 2022-04-21
WASM smart contract runime environment

graydon:
# Description

This is an initial, hopefully not-too-overwhelming piece of the WASM smart contract system. It is based on the prototype branch developed in https://github.com/stellar/stellar-core/pull/3380, but significantly cleaned up, extended and reorganized for review and merging. It is a step towards #3402, though there is more to do to complete that.

It includes two main pieces which are so tightly intertwined as to be pointless to subdivide:

  - A WASM interpreter ([fizzy](https://github.com/wasmx/fizzy)) with a "host" data structure to contain it (`HostContext`) with a convenience method for invoking "guest" WASM code within it.
  - A data model for values passed back and forth between host and guest, including handles to (immutable) objects that remain on the host, but that can be referred-to by handle values held in either host or guest.

It does not include any host functions yet -- that is for a future PR -- but it does include some packed values and host objects covering a small repertoire of useful types beyond the spartan few built in to WASM: symbols, bitsets, strings, blobs, vectors, maps, and arbitrary precision integers and rationals.

It also does not include any transactions to add smart contracts or invoke them, nor any ledger entries to store them on-chain or store their data records. Those are also for a future PR.

There is [a draft CAP explaining the design](https://github.com/stellar/stellar-protocol/pull/1176) in some detail.

Only the barest unit tests exist here to confirm initial functionality. More will follow.

### DIFF
--- a/src/transactions/contracts/HostVal.cpp
+++ b/src/transactions/contracts/HostVal.cpp
@@ -1,0 +1,676 @@
+#include "transactions/contracts/HostVal.h"
+#include "crypto/ShortHash.h"
+#include "src/ledger/LedgerHashUtils.h"
+#include "src/util/XDROperators.h"
+#include "transactions/contracts/HostContext.h"
+#include <Tracy.hpp>
+#include <algorithm>
+#include <boost/multiprecision/rational_adaptor.hpp>
+#include <variant>
+#include <vector>
+
+namespace stellar
+{
+
+std::string
+HostVal::asSymbol() const
+{
+    ZoneScoped;
+    size_t MAXCHARS = SCSymbol::max_size();
+    size_t MAXBITS = MAXCHARS * 6;
+    static char dict[64] =
+        "_0123456789ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz";
+    std::string out;
+    out.reserve(MAXCHARS);
+    auto tmp = getBody();
+    for (size_t off = 6; off <= MAXBITS; off += 6)
+    {
+        auto idx = (tmp >> (MAXBITS - off)) & 63;
+        if (idx != 0)
+        {
+            out += dict[idx - 1];
+        }
+    }
+    return out;
+}
+
+HostVal
+HostVal::fromSymbol(std::string const& s)
+{
+    ZoneScoped;
+    size_t MAXSZ = SCSymbol::max_size();
+    if (s.size() > MAXSZ)
+    {
+        throw std::runtime_error("bad symbol size");
+    }
+    uint64_t accum = 0;
+    for (auto const c : s)
+    {
+        accum <<= 6;
+        if (c == '_')
+        {
+            // 1 is underscore
+            accum |= 1;
+        }
+        else if ('0' <= c && c <= '9')
+        {
+            // 2-11 inclusive are digits
+            accum |= 2 + (c - '0');
+        }
+        else if ('A' <= c && c <= 'Z')
+        {
+            // 12-37 inclusive are uppercase
+            accum |= 12 + (c - 'A');
+        }
+        else if ('a' <= c && c <= 'z')
+        {
+            // 38-63 inclusive are lowercase
+            accum |= 38 + (c - 'a');
+        }
+        else
+        {
+            throw std::runtime_error("bad symbol char");
+        };
+    }
+    return fromBodyAndTag(accum, TAG_SYMBOL);
+}
+
+std::ostream&
+operator<<(std::ostream& out, HostVal const& v)
+{
+    ZoneScoped;
+    if (v.isU63())
+    {
+        out << "u63(" << v.asU63() << ")";
+    }
+    else if (v.isVoid())
+    {
+        out << "void";
+    }
+    else if (v.isBool())
+    {
+        out << "bool(" << v.asBool() << ')';
+    }
+    else if (v.isStatus())
+    {
+        auto pair = v.asStatusTypeAndCode();
+        out << "status(" << pair.first << "," << pair.second << ')';
+    }
+    else if (v.isU32())
+    {
+        out << "u32(" << v.asU32() << ')';
+    }
+    else if (v.isI32())
+    {
+        out << "i32(" << v.asI32() << ')';
+    }
+    else if (v.isSymbol())
+    {
+        out << "sym(" << v.asSymbol() << ')';
+    }
+    else if (v.isBitSet())
+    {
+        out << "bitset(";
+        auto tmp = v.asBitSet();
+        for (auto i = 0; i < 48; ++i)
+        {
+            out << (tmp & 1 ? '1' : '0');
+            tmp >>= 1;
+        }
+        out << ')';
+    }
+    else if (v.isObject())
+    {
+        out << "obj(" << v.asObject() << ')';
+    }
+    else
+    {
+        auto payload = v.payload();
+        out << "unknown(" << payload << ",0x" << std::hex << payload << ')';
+    }
+    return out;
+}
+
+struct HashHostObjectVisitor
+{
+    size_t
+    operator()(HostBox const& box) const
+    {
+        return std::hash<HostValInContext>()(box.get());
+    }
+
+    size_t
+    operator()(HostVec const& vec) const
+    {
+        size_t h{0};
+        for (auto const& v : vec)
+        {
+            hashMix(h, std::hash<HostValInContext>()(v));
+        }
+        return h;
+    }
+
+    size_t
+    operator()(HostMap const& map) const
+    {
+        size_t h{0};
+        for (auto const& pair : map)
+        {
+            hashMix(h, std::hash<HostValInContext>()(pair.first));
+            hashMix(h, std::hash<HostValInContext>()(pair.second));
+        }
+        return h;
+    }
+
+    size_t
+    operator()(uint64_t const& u) const
+    {
+        return shortHash::xdrComputeHash(u);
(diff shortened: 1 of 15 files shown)


src/xdr/Stellar-contract.x
```diff
@@ -0,0 +1,199 @@
...
+    SCO_BIGINT = 7,
+    SCO_BIGRAT = 8,
+
+    // Followed by a potentially much longer list of object types
+    // corresponding to any XDR ledger types users might want to manipulate
+    // directly. Separate type codes are required for composite types and
+    // each of their members, if it is reasonable for a user to want to hold
+    // a handle to the member separately from the composite type.
+
+    SCO_LEDGERKEY = 9,
```
leighmcculloch, line 127: Why do we need ledger key as a distinct handle type? I would expect there to be an object type for each identifier type we use to key into ledgers. i.e. AccountId, TrustLineLedgerKey, etc. At least on the contract side it seems not particularly useful to have a generic ledger key that could refer to anything.

graydon: The read-write sets of a contract invocation are sets of LedgerKeys.

leighmcculloch: So is LedgerKeys some subset of OBJs?

graydon: No, LedgerKeys keys that identify LedgerEntries in the ledger. Read-write sets have to do with concurrency control against the ledger, among multiple concurrent host contexts. There is no concurrent access to host objects.

leighmcculloch: How does a contract use the LedgerKeys object type? When would it encounter the type? What host functions would it use it with?

graydon: Well, at least my thinking was two contexts, both deriving from the assumption of deterministic concurrency control with ledger keys:

  - When a contract is invoked in a transaction that is bundled together with some set of LedgerKeys as its read-write sets, the contract will then need to _retrieve_ those keys somehow, in order to actually read or write them. It might use a host function to access the sets accompanying the transaction, or it might be passed elements of the sets as contract-function arguments, but either way it seems likely to me that it'd be receiving them as general LedgerKeys since that's how they'd be presented on the transaction's surface, not as specific subtypes of LedgerKey.
  -  Deterministic concurrency control sometimes requires a so-called "reconnaissance query" to run first in order to determine the read-write sets that will be used in a subsequent query with dynamic read-write sets. So in such a reconnaissance query, I was thinking it'd be returning a vector of vectors of LedgerKeys as its result value.
  
I suppose this will depend on whether classic interop winds up having access to the ledger via LedgerKeys at all, or whether it's restricted to CONTRACT_DATA entries subkeyed by SCVal -- in the latter case we'd just need to be indicating CONTRACT_DATA keys for the read-write sets.

leighmcculloch: Hmm okay got it. Is there an assumption that the contract would need to be given a list of read-write sets. Is that necessary? I, maybe naively, thought that the read-write sets would be given to core but not to the contract, and that a contract attempting to read or write outside of the set would be an error. The reason I thought that is I don't imagine writing code in a contract to handle the case where a read-write set missed including something I need. And I imagined that any ledger keys referenced by parameters to the contract call would be simple objects like AccountId.

I think it would be helpful to document some example contract calls which show how this will work.

graydon: Well, I mean, think it through: suppose your contract is going to read AccountID X. How is it going to _get_ AccountID X into the contract? X isn't going to appear out of thin air. It's going to be provided as a parameter to the contract, right? That is, a transaction that is an invocation of the contract will pass X. But we also want to indicate to core's concurrency-control mechanism (that is going to schedule the execution) that the contract's read-set includes LedgerKey(AccountID(X)). So suppose LedgerKey(AccountID(X)) is the Nth element of the read-set attached to the transaction. We have three-ish options:

  - We invoke the contract passing the Val N, and the contract turns around and calls a host function to fetch its Nth read-set element, in order to acquire LedgerKey(AccountID(X)), and then unwraps AccountID(X) from it.
  - We call the contract with an argument placeholder that says, effectively, "hey contract-invoking path, please grab read-set element N and put it in here as an argument", and then the contract is called with LedgerKey(AccountID(X)) again and can unwrap AccountID(X) from it again.
  - We cal the contract with a Val that holds AccountID(X) directly. Now we've provided 2 copies of AccountID in the transaction: one in the read-set and another redundant one in the contract argument, differing only by "being wrapped in LedgerKey(...)"
 
I figure providing two copies of the same key to a conract is not ideal (indeed if you look in [the project](https://github.com/orgs/stellar/projects/28/views/1) you'll see I have a work item "consolidate invoke tx environment and read-write sets" -- that's what this is about).

I was assuming we'd do the first or second option. But again, maybe all this is irrelevant since we might not be letting contracts access AccountIDs directly at all (they might be restricted to just be using CONTRACT_DATA). This is stuff that I expect will be determined by @jon and others working on the ledger interop plan; I don't feel very strongly about it, I'm just doing stuff here that follows the path I was on on the prototype branch.

Re adding example contracts to this PR: this PR is ahead of any of that. There are no host functions in here at all. It's all work-in-progress to arrive later.

leighmcculloch: If I understand correctly in all three options you describe the read-write set and the arguments are being provided to the contract, however what I was describing was that only the arguments get provided to the contract, and the contract is rather ignorant that a read-write set exists.

For example:

Tx would contain contract invocation with:
- Read-write set elements: `LK(A1)`, `LK(A2)`
- Arguments: `A1`, `A2`, `A3` (A3 is provided here as an argument, but the contract will not actually load its ledger key because it is not referencing a ledger entry)

Contract would receive:
- Arguments: `A1`, `A2`, `A3`

If the contract was provided with A4 without that account being included in the read-write set, and if the contract attempt to load the ledger entry referenced by A4, then the contract invocation would fail.

To highlight: the concern I raised was only with exposing the read-write to the contract. That seems like something that will complicate the contract interface. And I don't think contracts should try to handle cases where their read-write set was insufficient, or use their read-write set as some form of argument. I think that because then people will start using the read-write set list as sort of like another argument list.

To address the issue you raise of duplicate information in the transaction, I think we can address that without exposing ledger keys to contracts. We could say contract invocations provide a list of values, and then the read-write set and arguments are just indexes into that list of values. So the tx would contain a contract invocation like:
- Values: `A1`, `A2`, `A3`, ...
- Read-write set elements: `#1`, `#2`, ... (automatically converted into ledger keys within core)
- Arguments: `#1`, `#2`, `#3`

In this design there is no need to have "LedgerKey" as an object in the host to guest interface, because the guest doesn't concern itself with "ledger entries", it just concerns itself with objects that just happen to be backed by ledger entries.

leighmcculloch: Actually, even with both options 1 and 2 you mentioned I think we can still keep ledger keys out of the contracts themselves by translating a `LedgerKey(AccountID)` in the transaction interface into a `Object(OBJ_ACCOUNT)` in the host to contract interface. I've been assuming we will have `OBJ_ACCOUNT` in addition to `OBJ_ACCOUNTID`.

adamlaska:
@

src/transactions/contracts/HostVal.cpp
```diff
@@ -0,0 +1,676 @@
...
+        return std::lexicographical_compare(vec.begin(), vec.end(),
+                                            otherVec.begin(), otherVec.end());
+    }
+
+    bool
+    operator()(HostMap const& map) const
+    {
+        releaseAssert(std::holds_alternative<HostMap>(mOther));
+        auto const& otherMap = std::get<HostMap>(mOther);
+        return std::lexicographical_compare(map.begin(), map.end(),
```
jonjove, line 502: Per https://sinusoid.es/immer/containers.html#map, "Immutable unordered mapping of values from type K to type T." Given the "unordered", does this always work?

src/transactions/contracts/HostVal.cpp
```diff
@@ -0,0 +1,676 @@
...
+    {
+        return mVal.asU32() < other.mVal.asU32();
+    }
+    else if (mVal.isI32())
+    {
+        return mVal.asI32() < other.mVal.asI32();
+    }
+    else if (mVal.isVoid())
+    {
+        return false;
```
jonjove, line 620: Am I mistaken or does this permit comparing void to bool (they have the same tag)? That seems strange.

src/transactions/contracts/HostVal.cpp
```diff
@@ -0,0 +1,676 @@
...
+                return idx < otherIdx;
+            }
+            // Deep-compare if they're the same variant.
+            LtHostObjectVisitor lhov{*otherPtr};
+            return std::visit(lhov, *ptr);
+        }
+    }
+    else if (mVal.isSymbol())
+    {
+        return mVal.asSymbol() < other.mVal.asSymbol();
```
jonjove, line 660: Going back to my comment about `asSymbol` skipping 0s, this seems problematic. We check if two symbols are equal by bitwise comparison. But we use their string representation to compare them for inequality. So it's possible that `a != b` but also `!(a < b) && !(b < a)` which seems less than ideal.

jonjove (commented):
I don't want to block the PR over style/hardening. But I am slightly concerned that the `asSymbol` issue and the compare-void-to-bool issue might cause problems for future work, so maybe we should fix these now.

graydon:
Closed in favor of https://github.com/stellar/stellar-core/pull/3428

PR #3414, opened 2022-04-23
Fixes protocol upgrade bug in BucketManager test

### LINKED ISSUE #3349: Some bucket tests don't actually perform protocol upgrades
See https://github.com/stellar/stellar-core/pull/3329#discussion_r812222798 for more information.

SirTyson:
# Description

Resolves #3349

Fixes issue in some `BucketManager` tests where the protocol version would not upgrade properly.

# Checklist
- [x] Compiles
- [x] Ran all tests

### DIFF
--- a/src/bucket/test/BucketManagerTests.cpp
+++ b/src/bucket/test/BucketManagerTests.cpp
@@ -156,7 +156,8 @@ clearFutures(Application::pointer app, BucketList& bl)
 }
 
 static Hash
-closeLedger(Application& app, std::optional<SecretKey> skToSignValue)
+closeLedger(Application& app, std::optional<SecretKey> skToSignValue,
+            xdr::xvector<UpgradeType, 6> upgrades = emptyUpgradeSteps)
 {
     auto& lm = app.getLedgerManager();
     auto lcl = lm.getLastClosedLedgerHeader();
@@ -166,8 +167,8 @@ closeLedger(Application& app, std::optional<SecretKey> skToSignValue)
               hexAbbrev(app.getBucketManager().getBucketList().getHash()));
     auto txSet = std::make_shared<TxSetFrame const>(lcl.hash);
     app.getHerder().externalizeValue(txSet, ledgerNum,
-                                     lcl.header.scpValue.closeTime,
-                                     emptyUpgradeSteps, skToSignValue);
+                                     lcl.header.scpValue.closeTime, upgrades,
+                                     skToSignValue);
     return lm.getLastClosedLedgerHeader().hash;
 }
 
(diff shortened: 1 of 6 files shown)


src/ledger/test/LedgerTestUtils.cpp
```diff
@@ -535,5 +535,13 @@ generateLedgerHeadersForCheckpoint(
     makeValid(res, firstLedger, state);
     return res;
 }
+
+UpgradeType
```
SirTyson, line 539: I don't think this really belongs in LedgerTestUtils, but I couldn't see a more logical header to but it in.

src/bucket/test/BucketManagerTests.cpp
```diff
@@ -1242,7 +1243,33 @@ class StopAndRestartBucketMergesTest
             resolveAllMerges(app->getBucketManager().getBucketList());
             auto countersBeforeClose =
                 app->getBucketManager().readMergeCounters();
-            closeLedger(*app);
+
+            if (firstProtocol != secondProtocol &&
```
marta-lokhova, line 1247: I think we don't need this if block, as it doesn't actually impact upgrades past genesis ledger.

src/ledger/test/LedgerTestUtils.cpp
marta-lokhova: Any reason not to keep it where it used to be?

MonsieurNicolas:
@SirTyson any update on this one?

SirTyson:
> @SirTyson any update on this one?

I don't think any additional updates are required, I believe I addressed Marta's comments.

marta-lokhova:
@SirTyson I don't see any changes, maybe you forgot to push? (there's also a merge conflict that we should resolve)

src/ledger/test/LedgerTestUtils.cpp
SirTyson: Since several different test files use this function now, I thought it would make more sense to put it in a utils header instead of `TxTest.h`.

src/bucket/test/BucketManagerTests.cpp
marta-lokhova: oh I see, there's been some renaming, so LEDGER_PROTOCOL_VERSION is the max version supported by core. So yeah, we do need to update it. Why does it need to be here though? I think it'd be cleaner to update it in between application shutdown and start, since this is when the change takes effect.

marta-lokhova:
r+ aa519541374a5737c1aeb7aa46b05c990d681430

PR #3424, opened 2022-05-11
Clarify bucketlist duration arithmetic, again.

graydon:
The time intuitions section of the current BucketList.h comment isn't entirely wrong, but it's misleading as it does not differentiate "age of oldest object in a level" with "frequency at which a level receives an incoming spill". The latter is the important frequency when considering the time it takes to "rewrite the whole BL" for things like upgrades, so I have added a section here that clarifies that and gives time estimates for that (30-36 days for a full rewrite).


graydon:
cc @SirTyson and @MonsieurNicolas

MonsieurNicolas:
r+ d40d5a0bcbca46ed5a7d9eba0252937449e67dc5

PR #3427, opened 2022-05-13
Bump dependencies (asio 1.22.1, etc)

MonsieurNicolas:
Bump a few dependencies:

asio -- adds support for io_uring
others -- bug fixes, better compiler support

I did not touch sqlite (still quite behind) as there seems to be a long streak of issues related to bloom filter optimizations.

### DIFF
--- a/lib/catch.hpp
+++ b/lib/catch.hpp
@@ -1,13 +1,13 @@
 /*
- *  Catch v2.13.8
- *  Generated: 2022-01-03 21:20:09.589503
- *  ----------------------------------------------------------
- *  This file has been merged from multiple headers. Please don't edit it directly
- *  Copyright (c) 2022 Two Blue Cubes Ltd. All rights reserved.
- *
- *  Distributed under the Boost Software License, Version 1.0. (See accompanying
- *  file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
- */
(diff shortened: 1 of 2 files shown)


graydon:
r+ 1148d5a

graydon:
There's a comment in the file but basically this was failing because asio autoconfig was being half (but not completely) overridden by our explicit config-setting defines, and the reason we _have_ those config-setting defines is because asio autoconfig didn't seem to be working before. It turns out it wasn't working because it has to have a C++ stdlib header included in order to sniff definitions on libstdc++. So I removed all our config-setting defines and gave it stdbool as an include, which is enough to get the libstdc++ config defines it needs to autoconfig properly.

graydon:
@MonsieurNicolas I actually put an r+ on this a minute ago but removed it because you might want to review my "fix" here. I _think_ we're just getting to a better version of the same place as we were with the explicit defines (just eliminating forced inconsistencies we were causing) but it _might_ change some aspects of asio's behaviour. I don't think we're really in a position to eyeball those changes, but I can't think of an approach that's safer than just using our existing checks (sanitizers, regression testing etc.) against "asio doing its best". If you agree, r+ and land it.

MonsieurNicolas:
r+ c99d2d524e55a1d62210d19a7e610d57b2f1b873

PR #3428, opened 2022-05-14
Rust wasm runtime

graydon:
This is a much smaller and simpler smart contract PR for stellar-core, because most of the smart contract host environment is now in a separate rust crate stored in the [https://github.com/stellar/rs-stellar-contract-env](https://github.com/stellar/rs-stellar-contract-env) repository.

This PR just provides XDR types to the C++ side, connects core to the Rust crate, adds some plumbing for traversing the C++/Rust bridge by serializing and deserializing XDR types, and tests that it works.

PR #3432, opened 2022-05-20
Rust log to core

graydon:
This builds on the back of https://github.com/stellar/stellar-core/pull/3428 by connecting the Rust standard `log` crate API to the stellar-core logging subsystem, allowing us to fairly normally write `info!(...)` calls in Rust and have them come out of `spdlog`.

Along the way it adds a little bit of additional plumbing in the bridge module and renames C++ bridging-related files yet again (hopefully for the last time) to match our CamelCase naming convention on C++ files.


src/rust/src/log.rs
```diff
@@ -0,0 +1,83 @@
...
+            _ => LevelFilter::Info,
+        }
+    };
+    if HAVE_INITIALIZED
+        .compare_exchange(false, true, Ordering::SeqCst, Ordering::SeqCst)
+        .is_ok()
+    {
+        log::set_logger(&LOGGER)?;
+    }
+    log::set_max_level(maxFilter);
```
jonjove, line 54: I think this should also be getting called when `Logging::setLogLevel` is called, right?

jonjove:
r+ 51009b0

src/rust/src/log.rs
graydon: Yes, any time you change anything in the logging system it gets "deinitialized" and "reinitialized" in the C++ Logging.cpp / spdlog side. That's why there's the `HAVE_INITIALIZED` bool in the Rust side, to prevent multiple-initialization.

PR #3433, opened 2022-05-20
WIP: Rust Tracy

graydon:
This builds on the back of https://github.com/stellar/stellar-core/pull/3432 connecting the Rust Tracy API to the copy of Tracy we have in-tree. It involves a local Cargo override and some bindgen, which isn't delightful, but I got it as far as actually running. Will need to tidy up a little more before landing.


graydon:
closing in favor of https://github.com/stellar/stellar-core/pull/3830

Issue #3412, opened 2022-04-21
Make it easy to work on vNext protocol without breaking current


MonsieurNicolas:
Right now, event though we can configure core with `--enable-next-protocol-version-unsafe-for-production`, the xdr is the one for "vNext".

This creates problems: for "current", it means that some objects become valid and are not filtered by the xdr marshaling code, and instead have to be dealt with at the application layer. This is unnecessary risk when dealing with protocol versions being developed.
As we're about to make a lot of "vNext" changes as part of Jump Cannon, we need a way to really mitigate that risk.

Solution is to have different xdr files for "current" and "vNext".

Question is how to manage this.

A few options, I think the first one is the more practical:

### Separate xdr files into two folders

We'd have "current" (which could be the same folder that we use today) and "vNext" folders that contain the respective xdr files.

pros:
* easy to implement, no special tools needed
* easy to make arbitrary changes to the xdr, including deletion/addition of files
* clean git history

cons:
* backporting changes from "current" into "vNext" needs to be done manually (mitigated a bit by the fact that the git history is clean)

### use submodules for xdr files

`vNext` branch in an xdr submodule.

pros:
* We talked about this before, it allows to isolate xdr from core
* "vNext" is just a branch in git

cons:
* when working on "vNext", the "xdr" submodule is always going to be "dirty" because of the branch change, and will cause the "developer experience" to be bad.
    * rebasing on top of master requires stashing this submodule change (to switch from "vNext" to "current", which guarantees a "full rebuild"
    * special care has to be done to always exclude that change from commits
* backporting changes from "current" into "vNext" is done with merges into "vNext" needs to be done manually and the "vNext" branch will not be "based" off "current" (when we want to merge it).
* "configure" will make changes to the tree (has to call git) which is uncommon and potentially surprising for developers

### use "ifdefs" in xdr files

pros:
* "common" changes (like adding a case in a switch) are easy to do and read

cons:
* xdr files in tree would not really be be xdr files and need to be transformed into standard compliant files to be used in other projects
* some changes will require copy/pasting large portions of xdr (like the xdr change from CAP-0042), and adding a lot of conditionals
* hard to "cherry pick" changes from vNext into curent
* xdr tools have to change to support this or we have to run "CPP" on the xdr files as a pre-step. Right now, there is no support for preprocessors in xdr files

MonsieurNicolas:
done

PR #3434, opened 2022-05-20
Move XDR files to separate subdirs for vCurr and vNext

graydon:
This is a very simple implementation of the first strategy for bug #3412 -- allowing us to maintain two different protocol versions at the XDR level simultaneously. It does so by making two copies of the .x files in two separate subdirectories and then adjusting the include path to point to one or the other.

It is a _little_ different than expected in that it puts things in `src/protocol-{curr,next}/xdr` rather than `src/xdr/{curr,next}` as one might expect. We refer in many places to files by paths like `xdr/Stellar-Ledger.h` and I figured it'd be nice to maintain that rather than editing (or worse, conditionalizing) all those include sites.

Advantages:
  - Simple
  - Portable
  - Minimal code or build-system changes
  - Can diff current and next protocol easily
  - Always runs xdrc on both variants, will catch breakage early

Disadvantages
  - Can't eyeball differences without running diff
  - Need to port any change to curr over to next or risk losing it

Overall I am comfortable with trying this approach for a while to see how it sits. If agreeable I'll update the smart contract PR to match.


MonsieurNicolas:
r+ b9501ae3288a5879d457841168cb4b249691cb43

PR #3435, opened 2022-05-23
Implementation of CAP-42

### LINKED ISSUE #3382: Implement CAP-0042

dmkozh:
# Description

Resolves #3382

This is implementation of the protocol change only. It does not contain any behavior changes besides surge pricing change described in the CAP.

The digest of changes is as follows (more details in respective commits):

- Introduce `GeneralizedTransactionSet` XDR that supports multiple components with different base fees (or without one)
- Use `GeneralizedTransactionSet` starting from protocol 20 in all relevant systems (overlay, SCP state, history, tx meta etc.)
- Add functionality for transactions to have unique base fees
- Necessary refactoring: improved `TxSetFrame`  construction, cleaner SCP state storage, C++ frame over `LedgerCloseMeta` XDR and more minor changes.

# Checklist
- [ ] Compiles
- [ ] Ran all tests


jayz22 (commented):
Only had time to to look through the first few commits. Just a few minor comments.

src/transactions/TransactionFrame.h
```diff
@@ -183,22 +187,25 @@ class TransactionFrame : public TransactionFrameBase
 
     bool checkSignatureNoAccount(SignatureChecker& signatureChecker,
                                  AccountID const& accountID);
-
     bool checkExtraSigners(SignatureChecker& signatureChecker);
 
-    bool checkValid(AbstractLedgerTxn& ltxOuter, SequenceNumber current,
-                    bool chargeFee, uint64_t lowerBoundCloseTimeOffset,
-                    uint64_t upperBoundCloseTimeOffset);
+    bool checkValidWithOptionalFee(AbstractLedgerTxn& ltxOuter,
```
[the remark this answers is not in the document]
jayz22: Also, should this function be protected instead?

marta-lokhova (commented):
thanks for putting this together! I only did a partial review so far. Couple of minor suggestions, and one issue in overlay

src/transactions/TransactionSQL.cpp
```diff
@@ -8,14 +8,338 @@
...
+    XDROutputFileStream& txResultOut)
+{
+    if (encodedTxSets.empty() || ledgerSeq < encodedTxSets.front().first)
+    {
+        writeNonGeneralizedTxSetToStream(db, sess, ledgerSeq, txs, results,
+                                         txOut, txResultOut);
+    }
+    else
+    {
+        // Currently there can be gaps in this method calls when there are no
```
marta-lokhova, line 324: Is there a particular reason to keep iterating over the transactions table and handle gaps? Now that we store tx sets without gaps, would it be simpler/less error-prone to iterate over sets first and collect required transactions instead? Of course, we still need to deal with old-style tx sets, but we can extract the old-style logic into a separate function, and use the new logic on post-upgrade ledgers (asking because I spent quite a bit of time staring at this function, wondering if there's a way to make things easier to follow)

marta-lokhova (commented):
a couple more questions regarding history.

MonsieurNicolas (changes requested):
checkpoint (I spent most of my time looking at TxSetFrame.cpp/h) - overall changes look really good. Only found a few things so far

src/herder/TxSetFrame.cpp
```diff
@@ -176,21 +415,66 @@ TxSetFrame::checkValid(Application& app, uint64_t lowerBoundCloseTimeOffset,
...
+                continue;
+            }
+            if (*fee < lcl.header.baseFee)
+            {
+
+                CLOG_DEBUG(Herder, "Got bad txSet: {} has too low base fee {}",
+                           hexAbbrev(mPreviousLedgerHash), *fee);
+                return false;
+            }
+            // Here we validate the fee bid in relation to the respective bid
```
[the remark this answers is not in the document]
MonsieurNicolas: OK I see, the code and this comment are confusing. What is tested later is that the account can pay for the fee bid and that the fee bid is more than the "minimum ledger fee".

src/transactions/FeeBumpTransactionFrame.cpp
```diff
@@ -269,16 +274,27 @@ FeeBumpTransactionFrame::getFeeBid() const
 }
 
 int64_t
-FeeBumpTransactionFrame::getMinFee(LedgerHeader const& header) const
+FeeBumpTransactionFrame::getMinFee(LedgerHeader const& header,
+                                   std::optional<int64_t> baseFee) const
```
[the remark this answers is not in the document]
dmkozh: Actually I don't need to pass anything here at all (see `TransactionFrameBase` comment). The semantics here was indeed confusing.

I'm keeping the remaining optional `baseFee` fields for now. I believe the semantics in all the remaining places is consistent: the base fee is the discount to apply, so when it's missing no discount is applied. Using a custom struct everywhere is a bit more obvious on one hand, but also clunkier to work with on another.

Please let me know if you think we should go for a custom struct everywhere.

src/transactions/TransactionSQL.cpp
```diff
@@ -8,14 +8,338 @@
...
+
+    txOut.writeOne(hist);
+    txResultOut.writeOne(results);
+}
+
+// The following method handles the protocol version upgrade. Before
+// GENERALIZED_TX_SET_PROTOCOL_VERSION we don't store the tx set at
+// all, hence the results are empty. Starting from
+// GENERALIZED_TX_SET_PROTOCOL_VERSION we do store it. However, in
+// order to figure out if ledger `N` has generalized tx set we need
```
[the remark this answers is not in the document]
marta-lokhova: I see what you mean. That's not correct though: in published ledger headers, upgrade ledger contains the hash of an old-style tx set, so published tx sets must match what's in ledger headers, in order for core to properly verify history (otherwise the replay will fork). We do the check here: https://github.com/stellar/stellar-core/blob/ce981331000128a1c145ec364fbd83d3dd4be5ed/src/catchup/ApplyCheckpointWork.cpp#L219 So I think we should keep the tx set at the upgrade ledger as-is.

src/ledger/LedgerManagerImpl.cpp
```diff
@@ -689,12 +690,13 @@ LedgerManagerImpl::closeLedger(LedgerCloseData const& ledgerData)
...
     prefetchTxSourceIds(txs);
-    auto curBaseFee = txSet->getBaseFee(header.current());
-    processFeesSeqNums(txs, ltx, curBaseFee, ledgerCloseMeta);
+    processFeesSeqNums(txs, ltx, *txSet, ledgerCloseMeta);
 
     TransactionResultSet txResultSet;
     txResultSet.results.reserve(txs.size());
-    applyTransactions(txs, ltx, txResultSet, ledgerCloseMeta, curBaseFee);
+    applyTransactions(txs, ltx, txResultSet, ledgerCloseMeta);
+    storeTxSet(mApp.getDatabase(), ltx.loadHeader().current().ledgerSeq,
```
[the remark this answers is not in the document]
marta-lokhova: no worries, rewriting history can be really annoying. This is just something to keep in mind for the future to help improve the quality of our code reviews.

src/transactions/TransactionSQL.cpp
```diff
@@ -8,14 +8,338 @@
...
+                                         txOut, txResultOut);
+    }
+    else
+    {
+        // Currently there can be gaps in this method calls when there are no
+        // transactions in the set.
+        while (encodedTxSetIt != encodedTxSets.end() &&
+               ledgerSeq != encodedTxSetIt->first)
+        {
+            ++encodedTxSetIt;
```
MonsieurNicolas, line 329: maybe assert that we're skipping txsets that are empty?

MonsieurNicolas: actually found a bug: there are multiple ways to represent the empty tx set, so this could be a problem here (we would not have a way to verify that `H(txSet) == scpValue`).
I am adding a comment about this in TxSetFrame.cpp

src/herder/TxSetFrame.cpp
```diff
@@ -31,55 +31,135 @@
...
+                               return *c1.txsMaybeDiscountedFee().baseFee ==
+                                      *c2.txsMaybeDiscountedFee().baseFee;
+                           }) == phase.v0Components().end();
+    if (!componentBaseFeesUnique)
+    {
+        CLOG_DEBUG(Herder, "Got bad txSet: duplicate component base fees");
+        return false;
+    }
+
+    return true;
```
MonsieurNicolas, line 143: there is a missing check that `txs` cannot be empty (this is also missing from CAP-0042), otherwise we have an infinite number of ways to represent an empty set.

The only way should be to have `v0Components` empty.

src/transactions/TransactionSQL.cpp
MonsieurNicolas: so this points to missing tests.

MonsieurNicolas (changes requested):
I think I am done with my first round of review on this PR. It doesn't look too far off, pretty sure we can land this soon (just need a few more tests to catch issues found during this review)

src/overlay/Peer.cpp
```diff
@@ -1282,27 +1293,51 @@ void
...
+                           ? TX_SET
+                           : GENERALIZED_TX_SET;
     auto self = shared_from_this();
     if (auto txSet = mApp.getHerder().getTxSet(msg.txSetHash()))
     {
         StellarMessage newMsg;
-        newMsg.type(TX_SET);
-        txSet->toXDR(newMsg.txSet());
+        newMsg.type(messageType);
+        if (messageType == TX_SET)
```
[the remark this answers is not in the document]
marta-lokhova: I had a similar thought wrt supercluster: I'd expect version mix consensus to fail. Haven't investigated yet, but it's worth taking a look (we definitely still need as many unit tests as possible)

src/overlay/Peer.cpp
```diff
@@ -1286,23 +1297,66 @@ Peer::recvGetTxSet(StellarMessage const& msg)
...
+                                        .getLastClosedLedgerHeader()
+                                        .header.ledgerVersion,
+                                    GENERALIZED_TX_SET_PROTOCOL_VERSION)
+                ? TX_SET
+                : GENERALIZED_TX_SET;
+        if (messageType == GENERALIZED_TX_SET &&
+            mRemoteOverlayVersion <
+                Peer::FIRST_VERSION_SUPPORTING_GENERALIZED_TX_SET)
+        {
+            return;
```
marta-lokhova, line 1339: mmm, any reason not to respond with TX_SET here? If the peer is on older overlay version, it doesn't understand generalized tx sets yet, so it has to be TX_SET

src/herder/TxSetFrame.cpp
```diff
@@ -100,20 +180,179 @@ struct ApplyTxSorter
...
+TxSetFrameConstPtr
+TxSetFrame::makeFromHistoryTransactions(Hash const& previousLedgerHash,
+                                        Transactions const& txs)
+{
+    return std::shared_ptr<TxSetFrame>(
+        new TxSetFrame(false, previousLedgerHash, txs));
+}
+
+TxSetFrameConstPtr
+TxSetFrame::makeEmpty(LedgerHeaderHistoryEntry const& lclHeader)
```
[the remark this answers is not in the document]
marta-lokhova: Oh when I first reviewed HerderTests, there were some places that didn't use makeEmpty and similar functions, but maybe it got addressed in later commits? Anyways, looks good now. Resolving.

src/simulation/Simulation.cpp
```diff
@@ -487,6 +505,10 @@ Simulation::haveAllExternalized(uint32 num, uint32 maxSpread)
     uint32_t min = UINT32_MAX, max = 0;
     for (auto it = mNodes.begin(); it != mNodes.end(); ++it)
     {
+        if (mPausedNodes.find(it->first) != mPausedNodes.end())
+        {
+            continue;
+        }
```
marta-lokhova, line 511: looks like `Simulation::crankNode` doesn't handle paused nodes at the moment. But also, I'm not really sure we need the pause functionality at all. We can just restart the node, allowing the network to close a few ledgers in between.

src/herder/test/UpgradesTests.cpp
```diff
@@ -2484,3 +2486,135 @@ TEST_CASE("upgrade to generalized tx set", "[upgrades]")
...
+        upgrades.mUpgradeTime = lclCloseTime + std::chrono::seconds(15);
+        node->getHerder().setUpgrades(upgrades);
+    }
+
+    auto root = TestAccount::createRoot(*nodes[0]);
+
+    int txId = 0;
+    int64_t const minBalance =
+        nodes[0]->getLedgerManager().getLastMinBalance(0);
+    auto sendTxs = [&]() {
```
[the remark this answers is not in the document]
marta-lokhova: note that then you don't need to track each individual ledger like in `testLedger`. What you can do is stop one node at the beginning  (it should be on ledger 3), crank for a few ledgers past the upgrade, then restart the node and crank some more. We can check whether we're expecting a generalized ledger or not based on the ledger number when we look at closed ledgers. I think this will also make the code much simpler and easier to follow.

marta-lokhova (commented):
Thanks for the updates! Reviewed the Herder portion, still need to review/respond to history-related parts.

src/simulation/Simulation.cpp
marta-lokhova: And by "stop the node", I mean using `Simulation::removeNode` (it's used in a few other tests like this already)

src/overlay/Peer.cpp
```diff
@@ -1286,23 +1297,66 @@ Peer::recvGetTxSet(StellarMessage const& msg)
     if (auto txSet = mApp.getHerder().getTxSet(msg.txSetHash()))
     {
         StellarMessage newMsg;
-        newMsg.type(TX_SET);
-        txSet->toXDR(newMsg.txSet());
+        if (txSet->isGeneralizedTxSet())
+        {
+            if (mRemoteOverlayVersion <
```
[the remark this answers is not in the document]
marta-lokhova: Oh actually, I think the case you're handling is when the other peer doesn't know it's requesting something it doesn't understand. In this case yeah, I think it's ok to let it timeout. The other peer won't be able to sync with the network anyway after the upgrade.

src/overlay/test/OverlayTests.cpp
```diff
@@ -1734,4 +1736,66 @@ TEST_CASE("peer is purged from database after few failures",
...
+                return cfg;
+            });
+
+        simulation->startAllNodes();
+        auto nodeIDs = simulation->getNodeIDs();
+        auto node = simulation->getNode(nodeIDs[0]);
+
+        auto root = TestAccount::createRoot(*node);
+
+        int64_t const minBalance =
```
marta-lokhova, line 1771: similarly here, you can just use loadgen to generate transactions

src/transactions/TransactionSQL.cpp
marta-lokhova: alright, let's keep it as-is, I don't feel super strongly (also, now that the behavior has been clarified in the comments, it's easier to follow!)

src/herder/test/UpgradesTests.cpp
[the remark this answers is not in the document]
dmkozh: As per offline discussion, changed the test to just add a new node to the network and let it catchup, so that we don't need to worry about the restart code.

src/overlay/test/OverlayTests.cpp
dmkozh: As per additional offline discussion, removed this test and the logic for handling non-upgraded peers. Non-upgraded peers won't be able to parse new XDR anyway, so we don't need to specially handle them.

Builds/VisualStudio/stellar-core.vcxproj
```diff
@@ -1112,6 +1116,27 @@ exit /b 0
       <OutputItemType Condition="'$(Configuration)|$(Platform)'=='DebugNoPostgres|x64'">ClInclude</OutputItemType>
       <OutputItemType Condition="'$(Configuration)|$(Platform)'=='Release|x64'">ClInclude</OutputItemType>
     </CustomBuild>
+    <CustomBuild Include="..\..\src\xdr\Stellar-internal.x">
```
MonsieurNicolas, line 1119: this still references `src\xdr` instead of curr/next

src/herder/test/UpgradesTests.cpp
```diff
@@ -2458,3 +2456,148 @@ TEST_CASE_VERSIONS("upgrade flags", "[upgrades][liquiditypool]")
...
+                herder.getHerderSCPDriver().toStellarValue(
+                    env.statement.pledges.externalize().commit.value, sv);
+                return pe.getTxSet(sv.txSetHash);
+            }
+        }
+        return txSet;
+    };
+
+    // Make sure tx set format switches to generalized after upgrade.
+    for (int ledger = 4; ledger <= 11; ++ledger)
```
MonsieurNicolas, line 2592: needs to be `uint32_t` to avoid warning

src/overlay/test/OverlayTests.cpp
```diff
@@ -1734,4 +1736,66 @@ TEST_CASE("peer is purged from database after few failures",
...
+            node->getLedgerManager().getLastMinBalance(0);
+        REQUIRE(node->getHerder().recvTransaction(
+                    root.tx({txtest::createAccount(
+                        txtest::getAccount("acc").getPublicKey(), minBalance)}),
+                    false) == TransactionQueue::AddResult::ADD_STATUS_PENDING);
+        simulation->crankForAtLeast(Herder::EXP_LEDGER_TIMESPAN_SECONDS, false);
+
+        for (auto const& nodeID : simulation->getNodeIDs())
+        {
+            auto node = simulation->getNode(nodeID);
```
MonsieurNicolas, line 1781: warning: `declaration of 'node' hides previous local declaration`

Builds/VisualStudio/stellar-core.vcxproj
MonsieurNicolas: (this causes the 2 files that include `stellar-internal.h` to be recompiled all the time as msbuild cannot figure out dependencies)

src/overlay/Peer.cpp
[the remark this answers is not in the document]
MonsieurNicolas: change looks good now

src/protocol-curr/xdr/Stellar-ledger.x
```diff
@@ -176,6 +176,29 @@ case METAENTRY:
     BucketMetadata metaEntry;
 };
 
+enum TxSetComponentType
+{
+  // txs with effective fee <= bid derived from a base fee (if any).
+  // If base fee is not specified, no discount is applied.
+  TXSET_COMP_TXS_MAYBE_DISCOUNTED_FEE = 0
```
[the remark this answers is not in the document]
MonsieurNicolas: thank  you

src/transactions/FeeBumpTransactionFrame.cpp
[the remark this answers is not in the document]
MonsieurNicolas: yeah this is fine as this is now consistently used

src/herder/TxSetFrame.cpp
```diff
@@ -213,53 +497,107 @@ TxSetFrame::sizeOp() const
...
+                           [&](int64_t t, TransactionFrameBasePtr const& tx) {
+                               return t + tx->getFeeBid();
                            });
 }
 
 void
 TxSetFrame::toXDR(TransactionSet& txSet) const
 {
     ZoneScoped;
+    releaseAssert(!isGeneralizedTxSet());
```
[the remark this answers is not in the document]
MonsieurNicolas: yes this makes sense

src/herder/TxSetFrame.cpp
```diff
@@ -100,20 +180,179 @@ struct ApplyTxSorter
...
+    if (!invalidTxs)
+    {
+        invalidTxs = &unusedInvalidTxs;
+    }
+    auto validTxs =
+        TxSetUtils::trimInvalid(txs, app, lowerBoundCloseTimeOffset,
+                                upperBoundCloseTimeOffset, *invalidTxs);
+    auto const& lclHeader = app.getLedgerManager().getLastClosedLedgerHeader();
+    std::shared_ptr<TxSetFrame> txSet(new TxSetFrame(lclHeader, validTxs));
+    txSet->surgePricingFilter(app.getLedgerManager().getLastMaxTxSetSizeOps());
```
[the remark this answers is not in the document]
MonsieurNicolas: yeah next PR it is then

src/herder/TxSetFrame.cpp
```diff
@@ -176,21 +415,66 @@ TxSetFrame::checkValid(Application& app, uint64_t lowerBoundCloseTimeOffset,
...
+
     if (this->size(lcl.header) > lcl.header.maxTxSetSize)
     {
         CLOG_DEBUG(Herder, "Got bad txSet: too many txs {} > {}",
                    this->size(lcl.header), lcl.header.maxTxSetSize);
         return false;
     }
 
-    if (!mTxsIsValidHashOrder)
+    if (std::adjacent_find(mTxs.begin(), mTxs.end(),
```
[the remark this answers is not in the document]
MonsieurNicolas: ah yes this is to check that a tx is unique across components

src/main/Config.cpp
```diff
@@ -128,7 +128,7 @@ Config::Config() : NODE_SEED(SecretKey::random())
     MAXIMUM_LEDGER_CLOSETIME_DRIFT = 50;
 
     OVERLAY_PROTOCOL_MIN_VERSION = 19;
-    OVERLAY_PROTOCOL_VERSION = 21;
```
MonsieurNicolas, line 131: careful when rebasing here (should be 23)

src/overlay/Peer.h
```diff
@@ -56,6 +56,7 @@ class Peer : public std::enable_shared_from_this<Peer>,
 
   public:
     static constexpr uint32_t FIRST_VERSION_SUPPORTING_FLOW_CONTROL = 20;
+    static constexpr uint32_t FIRST_VERSION_SUPPORTING_GENERALIZED_TX_SET = 22;
```
MonsieurNicolas, line 59: careful when rebasing (should be 23)

src/main/PersistentState.cpp
```diff
@@ -78,7 +80,8 @@ PersistentState::getSCPStateAllSlots()
     std::vector<std::string> states;
     for (uint32 i = 0; i <= mApp.getConfig().MAX_SLOTS_TO_REMEMBER; i++)
     {
-        auto val = getFromDb(getStoreStateName(kLastSCPData, i));
+        auto val = states.emplace_back(
```
MonsieurNicolas, line 83: I think this got borked: it should just be `auto val = getFromDb(getStoreStateName(kLastSCPData, i));`

MonsieurNicolas (changes requested):
Looks pretty much ready!
I think you can address the remaining comments as you squash and rebase

src/main/PersistentState.cpp
dmkozh: Thanks for catching this, fixed.

src/overlay/test/OverlayTests.cpp
dmkozh: Fixed.

src/herder/test/UpgradesTests.cpp
dmkozh: Done.

Builds/VisualStudio/stellar-core.vcxproj
dmkozh: Thanks for the remainder. I'm aware of this, just wanted to postpone the update to rebase.

dmkozh: Rebased and updated this.

src/main/Config.cpp
dmkozh: Done, thanks for the reminder.

src/overlay/Peer.h
dmkozh: Done, thanks for the reminder.

src/overlay/Peer.cpp
marta-lokhova: the change still replies with generalized tx set as type. Maybe just remove line 1332 and let it timeout to be consistent with the behavior above

src/history/test/HistoryTests.cpp
```diff
@@ -958,6 +962,83 @@ TEST_CASE("Catchup post-shadow-removal works", "[history]")
     }
 }
 
+TEST_CASE("Catchup with generalized tx set upgrade",
```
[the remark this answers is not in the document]
marta-lokhova: yeah I think it's fine to remove those comments.

src/simulation/Simulation.cpp
marta-lokhova: Resolving this: the test was updated to spin up a new node.

src/overlay/Peer.cpp
dmkozh: It seems safer to reply with the old tx set for the sake of some weird catchup cases, doesn't it? I don't have a strong opinion on this, but unlike the case where we have tx set, giving a response here seems a bit more appropriate (as we don't say that we have something peer can't read).

src/history/test/HistoryTests.cpp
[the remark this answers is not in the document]
dmkozh: Sounds good, refactored the tests into one generalized test.

src/overlay/Peer.cpp
marta-lokhova: ok, that's fine, I think either behavior is appropriate. I see you updated this to be TX_SET, so I'm resolving this.

src/history/test/HistoryTests.cpp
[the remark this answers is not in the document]
marta-lokhova: thanks!

MonsieurNicolas:
r+ ed14258b8a1e8eb913d43031f890453007225cc3

PR #3436, opened 2022-05-23
fix bad dependencies in VS Project files

MonsieurNicolas:
There are still references to xdr files that got moved, causing projects to be rebuilt from scratch every time


MonsieurNicolas:
r+ 08be96dc5ebda211483b991892007643d165a59d

PR #3438, opened 2022-05-23
Improved protocol version support for VS project

dmkozh:
# Description

- Introduced build configurations for current and next protocol versions
- Added both current and next version XDR/generated headers to the project
- Use the appropriate XDRs for appropriate protocol versions
- Libraries only have one Debug/Release configuration so that they don't need to be rebuilt

# Checklist
- [ ] Compiles
- [ ] Ran all tests


Builds/VisualStudio/stellar-core.vcxproj
```diff
@@ -127,6 +166,65 @@
...
+    <PreBuildEvent>
+      <Command>
+      </Command>
+    </PreBuildEvent>
+    <PreBuildEvent>
+      <Message>
+      </Message>
+    </PreBuildEvent>
+    <CustomBuildStep>
+      <Inputs>src/generated/xdr/Stellar-overlay.h</Inputs>
```
MonsieurNicolas, line 207: I think this is incorrect (this is the kind of bug that was fixed in master recently)

Builds/VisualStudio/stellar-core.sln
```diff
@@ -1,6 +1,6 @@
 Microsoft Visual Studio Solution File, Format Version 12.00
-# Visual Studio Version 16
-VisualStudioVersion = 16.0.28803.202
+# Visual Studio Version 17
+VisualStudioVersion = 17.1.32421.90
```
MonsieurNicolas, line 3: why did this change? (so far I don't think changes caused this bump)

Builds/VisualStudio/stellar-core.vcxproj
MonsieurNicolas: should grep for `src/generated/xdr`

MonsieurNicolas: should we just make all those rules the same and use a variable? basically PR becomes:
replace `src/generated` by `src/$(Configuration)/generated`  which gets rid of the rust "generated" folder as well.
And this I think can be done across the board.

MonsieurNicolas (changes requested):

Builds/VisualStudio/stellar-core.vcxproj
dmkozh: Changed everything to use $(Configuration). The only downside is that VS index/navigation isn't updated properly (more details in commit message), but that issue must have existed for the previous approach too.

Builds/VisualStudio/stellar-core.sln
dmkozh: Must be some automatic VS stamp, reverted.

MonsieurNicolas:
Looks good (and tried it locally) but it needs to be rebased

MonsieurNicolas:
r+ 56c7292278257057d0d7e010368d56cf5ba4e241

PR #3443, opened 2022-06-04
re-enable ccache per protocol version

MonsieurNicolas:
I think this was broken when we split curr vs next protocol.

Looking at https://github.com/stellar/stellar-core/actions/runs/2436952384

"current" builds take 14 minutes
"next" builds take 1h+


graydon (approved):
Ah, nice catch.

graydon:
r+ de54b18

PR #3444, opened 2022-06-06
Experimental BucketList key-value store

SirTyson:
# Description

This PR adds a new experimental flag `EXPERIMENTAL_BUCKET_KV_STORE` that when set uses the `BucketList` for `LedgerEntry` loads. To reduce disc IO, an in memory index maps `LedgerKey`s to the corresponding `LedgerEntry`'s offset in the Bucket file. This allows us to do an in memory binary search for the key and a single disc IO operation for loading the entry once the key has been found.

Since there are too many `LedgerEntry`s to keep an in memory map for each key, we index ranges of keys that are within a given page for buckets above a certain size. `EXPERIMENTAL_BUCKET_KV_STORE_INDEX_PAGE_SIZE` determines the page size of the range based index, and `EXPERIMENTAL_BUCKET_KV_STORE_INDEX_CUTOFF` determines at which size the bucket switches from indexing individual keys to indexing key ranges.

After playing around with the config flags, I've found that a page size of 1024 bytes and a individual key index cutoff of 20 MB balances lookup speed and memory usage well, but I have not yet done a rigorous analysis of the trade offs. These values are currently the defaults for these config flags. 

Single entry loads and prefetch are both working properly. Loading trustlines by account and asset has not been tested well. For performance testing, For performance testing, I primarily ran catchup for ledgers from April 18-21, 2021. Currently, the BucketList prefetch lookup is 10% slower than postgres lookup. For individual loads, the BucketList seems to be significantly faster, but I need to double check the numbers because it seems a little too good to be true.

There's still some cleanup to be done. In particular, I want to unify a lot of the code between the different BucketIndex subclasses and clean up the prefetch function. That being said I think overall the code looks pretty good and is ready for some review and feedback.

# Checklist
- [x] Compiles
- [x] Ran all tests


graydon (changes requested):
Overall very nice work! I'm sorry for the bad news around synchronous index-building, but it looks extremely promising so far and I'm sure the remaining distance can be covered with a little reorg.

src/bucket/Bucket.cpp
```diff
@@ -29,10 +29,274 @@
...
+    while (in && in.readOne(be))
+    {
+        if (be.type() != METAENTRY)
+        {
+            LedgerKey key = Bucket::getBucketLedgerKey(be);
+            CLOG_TRACE(Bucket, "Indexed {} at {} in {}",
+                       xdr_to_string(key, "entry"), pos, filename.filename());
+
+            // If we need to create a new entry for this page
+            if (pos >= pageUpperBound)
```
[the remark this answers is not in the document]
SirTyson: Correct, nextPageBoundary advances to whatever the next page boundary is, not necessarily just one more page forward. It's not good if an entry exceeds pageSize, as both the upper and lower bound are the same key so you're doubling the memory over just the individual key index. Nothing breaks though, before I put in the minimum pageSize I did a few tests with pageSize = 1 and it was fine.

src/bucket/Bucket.cpp
```diff
@@ -42,19 +306,39 @@ Bucket::Bucket(std::string const& filename, Hash const& hash)
                    mFilename);
         mSize = fs::size(filename);
     }
+
+    if (pageSize)
+    {
+        // Use individual key indexing if size if below cutoff
+        initIndex(mSize < indexCutoff ? 0 : *pageSize);
```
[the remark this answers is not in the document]
SirTyson: I don't this this will be too bad, I was playing with separating the bucket and index more anyway. Probably the path of least resistance is building the index during merge and passing a unique_ptr of the constructed index to adoptFileAsBucket.

However, this does raise a question about persisting indexes on disk. Should we persist indexes on disk and support resuming a partial index like we resume partial merges? I haven't looked into it yet, but thinking through it I think it might be necessary if we want to do the indexing while we merge.

src/bucket/Bucket.cpp
```diff
@@ -29,10 +29,274 @@
...
+    }
+}
+
+BucketRangeIndex::BucketRangeIndex(std::filesystem::path const& filename,
+                                   size_t pageSize)
+    : mPageSize(pageSize)
+{
+    ZoneScoped;
+    releaseAssert(!filename.empty());
+    releaseAssert(pageSize > 256);
```
MonsieurNicolas, line 76: the assert on `pageSize` is probably not what you want:
* I think you intend to require to have at least 1 object per page but things will work anyways, so it's more of a tuning thing. I would say a warning is probably enough (in config.cpp) and the number should be picked so that we expect "reasonable performance" (so even 256 in that context may not be great).
* Jump Cannon introduces much larger key sizes anyways

@graydon  - as I write this I am not sure we should allow "any key size" in Jump Cannon (https://github.com/stellar/stellar-protocol/blob/master/core/cap-0053.md) , this has pretty big implications on the performance of any "in memory index". So it would be pretty bad if people craft ledger entries with 1KB keys type of thing. I think we should just force people to compute their own "short keys", we can keep `SCVal` as part of their internal key, but limit the overall key size to something reasonable like 128 bytes (contractID + 2 full keys + extra stuff seems plenty).

src/bucket/BucketList.cpp
```diff
@@ -371,9 +374,151 @@ BucketList::getHash() const
...
+        {
+            break;
+        }
+    }
+
+    return entries;
+}
+
+std::vector<LedgerEntry>
+BucketList::loadPoolShareTrustLinesByAccountAndAsset(AccountID const& accountID,
```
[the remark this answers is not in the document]
MonsieurNicolas: yeah this is buggy for a different reason.
See `LedgerTxnRoot::Impl::loadPoolShareTrustLinesByAccountAndAsset`:
it's a join between:
* pools that match on the asset
* trustlines for that account (this is where you need a range) that correspond to the pool IDs

This can be done by either looking up all trustlines that are pool shares first and then looking for pools (to match asset), or scan all pools first and then scan accounts (I suspect this is worst).

not sure why the assert in ledgerTxn didn't trip

src/main/ApplicationImpl.cpp
```diff
@@ -644,6 +645,42 @@ ApplicationImpl::validateAndLogConfig()
             "requires --in-memory");
     }
 
+    if (mConfig.EXPERIMENTAL_BUCKET_KV_STORE)
```
MonsieurNicolas, line 648: I would like to see a few tests with this enabled (both versions)

src/bucket/Bucket.cpp
```diff
@@ -42,19 +306,39 @@ Bucket::Bucket(std::string const& filename, Hash const& hash)
...
+Bucket::Bucket() : mIndex(std::make_unique<BucketIndex const>())
 {
 }
 
+XDRInputFileStream&
+Bucket::getStream()
+{
+    if (!mStream)
+    {
+        mStream = std::make_unique<XDRInputFileStream>();
```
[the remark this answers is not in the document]
MonsieurNicolas: actually from what I see, it would be a bug to call this with en empty filename, so throw right away seems right

MonsieurNicolas (changes requested):

src/bucket/BucketManagerImpl.cpp
```diff
@@ -474,7 +485,19 @@ BucketManagerImpl::getBucketByHash(uint256 const& hash)
...
+        if (mApp.getConfig().EXPERIMENTAL_BUCKET_KV_STORE)
+        {
+            pageSize = std::make_optional(
+                mApp.getConfig().EXPERIMENTAL_BUCKET_KV_STORE_INDEX_PAGE_SIZE);
+            pageCutoff =
+                mApp.getConfig().EXPERIMENTAL_BUCKET_KV_STORE_INDEX_CUTOFF;
+        }
+
+        auto p =
+            std::make_shared<Bucket>(canonicalName, hash, pageSize, pageCutoff);
```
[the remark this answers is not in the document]
SirTyson: Looking at this more, it might be a good idea to have two bucket types, `Bucket` and `IndexedBucket`. `IndexedBucket` can just extend the base type with indexing and search related functions. It looks like all the paths that require the "definitely fast bucket" never do any searches with it and can be the base `Bucket` type. This makes the semantics a little more clear too, so we can have a `getBucketByHash` and a `getIndexedBucketByHash`.

SirTyson:
Here are some updates regarding the last round of commits. Currently, all BucketList search related functions should be working properly. The last few updates have introduced unit tests, cleaned up lots of the code and bugs, as well as improved performance.

The performance enhancements included primarily the addition of a bloom filter to range based indexes and an overhaul of `XDRStream.h`. With the range based index, there were many instances where a key would fall within the range but not be in the indexed bucket. This happened especially often in small buckets, where the range of keys on a single page may be large. Whenever this happened, all entries in the range on disk needed to be searched before it could be determined that the target entry was not within that range. This was very expensive and wasted many disk IO ops. By using a bloom filter as an additional check, we can now eliminate most of these false positives and determine that a key is not in the current bucket without disk IO ops. The second improvement involves the `XDRStream`. I removed some unnecessary and expensive error checking. Additionally, for range based indexes, the `XDRStream` now reads the entire page into the buffer instead of individual entries.

There are two configurations I have been testing, a high-speed configuration suitable for most situations and a low memory configuration. These configurations are controlled by configuration settings. The high-speed configuration has a page size of 16kb and an individual index cutoff of 20 mb. Prefetch performs 2.3x faster than the current postrges implementation, and individual entry lookup is also significantly faster. This configuration requires 500 mb of additional memory, with spikes up to 1 gb, which is comparable to that of postgres. The memory spikes are due to new buckets being indexed on background threads. During merging, the freshly merged bucket is also indexed on the background thread. This means that in the worst case scenario, the current bucket list is entirely indexed, and the finished merges of every bucket are also indexed while pending `resolve()`, leading to a 2x memory spike. 

The low memory configuration has similar lookup performance to MySQL. The configuration has a 32kb page size with an individual index cutoff of 2 mb. It also has an additional flag set to enable lazy indexing. With lazy indexing, buckets are not indexed in advance on background threads, but only when needed. This reduces the memory usage and memory spike, but at a large runtime cost. For the larger buckets, indexing can take over 30 seconds. With lazy indexing enabled, this could cause the node to stall and de-sync with the network at any time. However, for certain operations like catchup, this flag can be enabled to reduce memory overhead. The low memory configuration has a consistent overhead of 300 mb with no spikes. Note: The lazy indexing change has not been pushed yet but will be updated soon.

SirTyson:
Added overview design doc @ `src/bucket/readme.md`

marta-lokhova (commented):
nice work! I left a few questions around making sure indexes are built in the right places, and tightening the code to help prevent bugs

src/bucket/BucketIndex.cpp
```diff
@@ -0,0 +1,244 @@
...
+    return key;
+}
+
+template <class IndexT>
+BucketIndexImpl<IndexT>::BucketIndexImpl(std::filesystem::path const& filename,
+                                         std::streamoff pageSize)
+    : mPageSize(pageSize)
+{
+    ZoneScoped;
+    if (filename.empty())
```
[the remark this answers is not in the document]
SirTyson: Yes, this doesn't represent an invalid index, but an empty index. This happens when you index the entire BucketList and there are some empty initialized buckets. My thinking is every Bucket in the Bucket list should be indexed, even if the bucket is empty. It seemed to be a little cleaner to do this for the empty indexed buckets so that they pass the assert in `getIndex()` instead of putting in special cases for the empty initialized buckets in the BucketList.

src/bucket/BucketList.cpp
```diff
@@ -371,9 +377,227 @@ BucketList::getHash() const
...
+    for (size_t i = 0; i < mLevels.size(); ++i)
+    {
+        auto const& lev = mLevels[i];
+        std::array<std::shared_ptr<Bucket>, 2> buckets = {lev.getCurr(),
+                                                          lev.getSnap()};
+        for (auto b : buckets)
+        {
+            // Once we see one empty bucket, the rest will also be empty, return
+            // early
+            if (b->getFilename().empty())
```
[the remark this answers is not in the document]
SirTyson: Correct

src/ledger/LedgerManagerImpl.cpp
```diff
@@ -345,6 +345,7 @@ LedgerManagerImpl::loadLastKnownLedger(function<void()> handler)
                     {
                         mApp.getBucketManager().assumeState(
                             has, header.current().ledgerVersion);
+                        mApp.getBucketManager().initAllIndexes();
```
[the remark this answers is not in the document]
SirTyson: Initially, we did have that function in `assumeState`. However, in several codepaths, `assumeState` gets called as part of a work task. This would cause bucket indexing to take a very long time on startup since the task would keep getting slept. By moving it to `loadLastKnownLedger`, we can index synchronously and it's a lot faster. I originally did not index in bucket application for the same reason. I'll have to investigate it a little more, but I don't seem to be failing anything without indexing during bucket application but I want to look into it.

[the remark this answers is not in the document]
marta-lokhova: I'm not sure I'm following how this would work without rebuilding indexes after bucket application: for example, in case of online catchup, we start core at LCL=1 then replay buckets to arrive at some ledger N, while buffering ledgers from the network. `loadLastKnownLedger` only  covers the case when we start core with all the buckets already in place, but this isn't always the case, right?

src/main/Config.h
```diff
@@ -258,6 +258,27 @@ class Config : public std::enable_shared_from_this<Config>
     // configuration) to delay emitting metadata by one ledger.
     bool EXPERIMENTAL_PRECAUTION_DELAY_META;
 
+    // A config parameter that when set uses the BucketList as the primary
+    // key-value store for LedgerEntry lookups
+    bool EXPERIMENTAL_BUCKET_KV_STORE;
```
marta-lokhova, line 263: discussed offline a bit but following up for posterity: I think we need to make sure that when we switch to the bucketlist as a store, we should clean up any old state to avoid bugs and inconsistencies. This means that we need to drop the DB on restart if the experimental flag is one (so not just new-db), and also ensure that in the catchup path we don't actually apply buckets (so I think ApplyBucketsWork only needs to assumeState at that point)

src/bucket/BucketIndex.h
```diff
@@ -0,0 +1,142 @@
...
+    // if file size is less than the cutoff, individual key index is used.
+    // Otherwise range index is used, with the range defined by pageSize.
+    static std::unique_ptr<BucketIndex const>
+    createIndex(Config const& cfg, std::filesystem::path const& filename);
+
+    // Returns next page boundary > pos
+    static inline std::streamoff
+    nextPageBoundary(std::streamoff pos, std::streamoff pageSize)
+    {
+        return (pos & ~(pageSize - 1)) + pageSize;
```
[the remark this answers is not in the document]
SirTyson: Yeah, originally it was due to low level disk IO optimizations. Those were all experimental though and mostly got cut out. Currently there's nothing that would break if the page size isn't a power of two (except for the nextPageBoundary calculation), but since we are reading the range index a page at a time it would be preferably if the pageSize was a multiple of the underlying disk sector size. I'm thinking I'll keep the assert and add comments.

src/bucket/BucketIndex.h
```diff
@@ -0,0 +1,142 @@
...
+
+class Config;
+
+struct BucketIndexRangeEntry
+{
+    LedgerKey lowerBound;
+    LedgerKey upperBound;
+
+    BucketIndexRangeEntry(LedgerKey low, LedgerKey high)
+        : lowerBound(low), upperBound(high)
```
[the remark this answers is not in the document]
SirTyson: I'll definitely add the bounds check, but it would be difficult to make the index keys const. C++17 containers don't support `const` template types so I think it would be too involved to make the keys const. The index itself is a protected member and all the functions are `const` so I think we'll be safe without `const` keys.

src/bucket/BucketIndex.h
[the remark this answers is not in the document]
marta-lokhova: ok. Also, we should avoid duplicating the code here - we already have `BucketList::mask` that applies the same bitwise logic

src/main/Config.h
marta-lokhova: or actually, we would still need to apply buckets to rebuild offers, but we shouldn't be rebuilding any other tables

src/bucket/BucketOutputIterator.cpp
```diff
@@ -136,7 +137,22 @@ BucketOutputIterator::getBucket(BucketManager& bucketManager,
...
-                                           mObjectsPut, mBytesPut, mergeKey);
+
+    auto hash = mHasher.finish();
+    std::unique_ptr<BucketIndex const> index{};
+
+    // If this bucket needs to be indexed and is not already indexed
+    if (shouldIndex)
+    {
+        if (auto b = bucketManager.getBucketIfExists(hash);
+            !b || !b->isIndexed())
```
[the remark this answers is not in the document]
SirTyson: There actually are a few edge cases where this happens. In particular, this happens when merging a bucket with an empty bucket. This happens whenever the bucketlist trickles down into a previously empty level. Since these are the lowest level buckets, it would be expensive to re-index, so we need to keep this check so we can attach the index to the "new" merge result.

src/bucket/BucketIndex.cpp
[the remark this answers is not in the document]
SirTyson: After working through this a little more, I think it does make more sense to throw. After a few changes all of the `BucketList` logic now only iterates through non-empty buckets, so indexing or getting the index of an empty bucket should be an error.

src/bucket/BucketList.cpp
```diff
@@ -597,7 +823,7 @@ BucketList::restartMerges(Application& app, uint32_t maxProtocolVersion,
         auto& next = level.getNext();
         if (next.hasHashes() && !next.isLive())
         {
-            next.makeLive(app, maxProtocolVersion, i);
+            next.makeLive(app, maxProtocolVersion, i, true);
```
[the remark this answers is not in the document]
SirTyson: I've taken out all of the `shouldIndex` params that were redundant with the experimental flag. There are still a few functions that need `shouldSynchronouslyIndex`. Essentially, the experimental flag determines if a bucket should eventually be indexed, while `shouldSynchronouslyIndex` indicates that a  bucket needs to be indexed right now and that the function should block until the given bucket is indexed.

src/main/Config.h
SirTyson: With these changes, there are no errors and everything works as expected when switching the experimental flag from 0 to 1. However, if you switch the flag from 1 to 0, the app crashes with the error message `Got an exception: sqlite3_statement_backend::prepare: no such table:`. I don't expect many validators to switch the experimental flag off after turning it on, but should there be a better error message? For example, should we add a better error suggesting the user run `new-db`? Right now it just says to report the bug with the internal DB exception.

src/main/Config.h
```diff
@@ -258,6 +258,27 @@ class Config : public std::enable_shared_from_this<Config>
...
+
+    // A config parameter that when set lazily initializes bucket indexes when
+    // needed. This lowers the bucket index memory usage, but significantly
+    // reduces lookup speed. Should never be set on a validator node, node will
+    // fall out of sync with network.
+    bool EXPERIMENTAL_BUCKET_KV_STORE_LAZY_INDEX;
```
marta-lokhova, line 269: I'm slightly on the fence about this config, as it seems to try to mitigate a problem elsewhere (lack of sufficient resources outside of core). Can we try to resolve the resource problem first? Also, do you have a rough idea of how much we're actually saving? From what I understand, the reason this works is because in a way we spread out index generation, instead of kicking off a bunch of indexes all at once, is that right?

src/catchup/IndexBucketsWork.cpp
```diff
@@ -0,0 +1,88 @@
...
+            }
+
+            mIndexedBuckets.insert(b->getHash());
+            auto indexCallback = [b](Application& app) {
+                b->setIndex(BucketIndex::createIndex(app.getConfig(),
+                                                     b->getFilename()));
+                return true;
+            };
+
+            addWork<WorkWithCallback>("index-" + hexAbbrev(b->getHash()),
```
marta-lokhova, line 47: This is still problematic because it's done from the main thread, so this will cause nodes to go out of sync

src/bucket/BucketManager.h
```diff
@@ -205,10 +210,13 @@ class BucketManager : NonMovableOrCopyable
     virtual std::vector<std::string>
     checkForMissingBucketsFiles(HistoryArchiveState const& has) = 0;
 
-    // Restart from a saved state: find and attach all buckets in `has`, set
-    // current BL. Pass `maxProtocolVersion` to any restarted merges.
-    virtual void assumeState(HistoryArchiveState const& has,
-                             uint32_t maxProtocolVersion) = 0;
+    // Assume state from `has` in BucketList: find and attach all buckets in
+    // `has`, set current BL. Note: Does not restart merging
+    virtual void assumeState(HistoryArchiveState const& has) = 0;
```
marta-lokhova, line 215: Hmm, not sure I understand why merge restart is now done outside of assumeState. I think separating functionality like this (set state, build index, start merges) is generally a wrong pattern: what we should be doing is centralizing functionality to restore bucket state in one place, as it's less error-prone and allows us to reason about what's supposed to happen at state restoration (previously, this was assumeState). What you can do is turn assumeState into a work that executes a sequence of events (we have WorkSequence for this): set state, rebuild indexes (this must be done in the background as per my other comment, using IndexBucketWork you already implemented), restart merges. Then in catchup you can add it to the sequence after bucket application. On startup, you'd just call executeWork for synchronous execution.

src/catchup/IndexBucketsWork.cpp
```diff
@@ -0,0 +1,88 @@
...
+    if (!mChildrenSpawned)
+    {
+        auto indexBucket = [&](std::shared_ptr<Bucket> b) {
+            // Don't index empty bucket or buckets that are already being
+            // indexed. Sometimes one level's snap bucket may be another level's
+            // future bucket. The indexing job may have started but not
+            // finished, so check mIndexedBuckets. Some buckets may have already
+            // been indexed when assuming LCL state, so also check b->isIndexed.
+            if (b->getFilename().empty() || b->isIndexed() ||
+                mIndexedBuckets.find(b->getHash()) != mIndexedBuckets.end())
```
marta-lokhova, line 35: is the last check for mIndexedBuckets needed? Isn't this already covered by b->isIndexed()?

src/bucket/BucketList.cpp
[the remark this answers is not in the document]
marta-lokhova: Yeah, older versions buckets are weird because they don't contain METAENTRY. So we should be conservative and always iterate over the whole bucketslist. 

Also - I think it's error-prone to check filenames like this. The caller isn't supposed to know an internal to Bucket detail that empty buckets have empty filenames. What we should do is have a public method in the Bucket class, which checks whether the bucket is empty. While we're there, we should check that both filename and hash are empty (otherwise throw, as it's an invalid bucket). We should clean up the code to use that function everywhere: from what I see, we also check for empty buckets in BucketList::restartMerges and in HistoryArchiveState::containsValidBuckets.

src/catchup/IndexBucketsWork.cpp
SirTyson: I think this code path is only for startup. When a validator is running, the merging is done on the background thread in `Bucket::merge` and synchronously for small buckets in `Bucket::fresh`. `IndexBucketsWork` is called during catchup and when loading the LCL initially, so it should be ok to block here?

marta-lokhova: We can't block during online catchup because we need to continue listening to the network to buffer ledgers

src/catchup/IndexBucketsWork.cpp
SirTyson: It is. There are a few weird edge cases where one level's snap bucket can be another level's future bucket. Before adding this check, we would kick off indexing jobs for all the buckets in the bucket list, including futures ready to resolve, and sometimes we would double index a bucket.

src/main/Config.h
SirTyson: It's about 2/3 to 1/2 the memory footprint. The largest issue with non-lazy indexing is the future buckets that are not part of the bucketlist yet. Once a future bucket is done merging, they index themselves on the background thread so they can be immediately searched when they are eventually added to the bucketlist. This means that, especially for the larger buckets at the lower levels, we have large indexes sitting around and not being used. The memory gains from this flag are primarily from not indexing future buckets.

SirTyson: I'm not sure how necessary this config is to parallel catchup, but it was fairly straight forward to implement and helpful in bug finding. We'd have to ping someone more familiar with supercluster to see how necessary this memory reduction really is (or isn't).

src/main/ApplicationImpl.cpp
```diff
@@ -157,61 +157,98 @@ maybeRebuildLedger(Application& app, bool applyBuckets)
...
+                    break;
+#endif
+                default:
+                    abort();
+                }
             }
         }
 
-        tx.commit();
+        if (toRebuild.empty())
```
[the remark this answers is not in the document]
SirTyson: I'm trying to preserve the original logic flow of the function here. Initially, this `toRebuild.empty()` check was at the beginning of the function and returned early. We don't want to do that now since we want to always drop the SQL tables if bucketlist loads are enabled, but I still want to exit early in that case so we don't hit the `applyBuckets` codepath.

src/catchup/IndexBucketsWork.cpp
SirTyson: Is this why the apply buckets work has the `advance` interface to break it down into fast chunks to not block and should we use a similar interface here?

src/main/ApplicationImpl.cpp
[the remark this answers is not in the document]
marta-lokhova: sure but there's no need to duplicate the commit call though - you can commit the tx regardless and exit early if there's nothing to rebuild (also, related to my other comment - if you setup toRebuild such that it takes the experimental flag into consideration , then there's no need to check the flag in the rest of the function, so you could use the same condition as before `toRebuild.empty()`)

src/catchup/IndexBucketsWork.cpp
marta-lokhova: yeah, that's why bucket application is done in chunks. Regarding what to do here, I don't think we need to mimic that pattern as it's super slow and requires a bunch of changes to how we index: simplest is to just spin up indexing in the background, and notify Work when it's ready. We have this pattern in a few places, see VerifyBucketWork for example (it sends an expensive verification task to the background, then the background task schedules a callback to the main thread to notify the work it's ready - it's not the cleanest pattern tbh, but it's sufficient for now)

src/bucket/FutureBucket.cpp
```diff
@@ -362,7 +364,7 @@ FutureBucket::startMerge(Application& app, uint32_t maxProtocolVersion,
                     bm, maxProtocolVersion, curr, snap, shadows,
                     BucketList::keepDeadEntries(level), countMergeEvents,
                     app.getClock().getIOContext(),
-                    !app.getConfig().DISABLE_XDR_FSYNC);
+                    !app.getConfig().DISABLE_XDR_FSYNC, mShouldIndex);
```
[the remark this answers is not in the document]
marta-lokhova: Still getting errors while trying to catchup locally with `catchup 27000000/0`:

```
2022-09-02T12:30:42.175 GAW4V [History INFO] ApplyBuckets : done, restarting merges
2022-09-02T12:30:42.176 GAW4V [Bucket INFO] BucketIndex::createIndex() indexing individual keys in bucket "buckets/tmp/bucket-83bae6622aa42060/tmp-bucket-c502a3ee99680d17.xdr"
b->isIndexed() at bucket/FutureBucket.cpp:418
```
I think this is related to the other thread about centralizing "restore state" functionality: right now catchup assumes state, restarts merges and then does the indexing, but that's out of order, hence the merges complain. Consolidating everything in one place with a clear order of events (set state -> build indexes -> restart merges) should help.

src/bucket/BucketManagerImpl.cpp
```diff
@@ -670,10 +695,10 @@ BucketManagerImpl::forgetUnreferencedBuckets()
         // have references to them, just us. It's ok to retain a few too
         // many buckets, a little longer than necessary.
         //
-        // This conservatism is important because we want to enforce that only
-        // one bucket ever exists in memory with a given filename, and that
-        // we're the first and last to know about it. Otherwise buckets might
-        // race on deleting the underlying file from one another.
+        // This conservatism is important because we want to enforce that
```
MonsieurNicolas, line 698: I thought you mentioned before that indexes for buckets that are only referenced by the publish queue were using a lot of memory - this seems to still be the case with this PR?

MonsieurNicolas (changes requested):
This looks great overall! Most of my comments are cleanup items. I'm going to try to add a couple commits to your PR to address Windows related issues

MonsieurNicolas:
I created a gist to apply after rebasing on top of master for Visual studio -- https://gist.github.com/MonsieurNicolas/a62c6e32f464d0dc237b1cfca765dc1c

src/bucket/Bucket.h
```diff
@@ -34,29 +35,66 @@ class Database;
...
+    // Lazily-constructed and retained for read path.
+    std::unique_ptr<XDRInputFileStream> mStream;
 
   public:
     // Create an empty bucket. The empty bucket has hash '000000...' and its
     // filename is the empty string.
     Bucket();
 
+    // Returns true if bucket is indexed, false otherwise
+    bool isIndexed() const;
```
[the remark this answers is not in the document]
SirTyson: Currently, this is only called in a few places where a class needs to decide whether or not to index a bucket when indexing the entire bucketlist synchronously. I don't know if there is a good way to do this without this method. In several startup paths, we need to index the entire bucket list twice. For example, if we are running a catchup but already have some buckets saved on disk, we need to index the bucketlist initially to assume LCL state, then again after the bucket apply step. If the catchup is over a small range, many of the bucket indexes from the LCL will still be valid, so we use `isIndexed` to avoid indexing redundantly. We could adjust the abstraction some and replace this with `indexIfnotIndexed` method, but I think we'll need some way of outside classes knowing whether or not a bucket is indexed for these scenarios.

src/bucket/BucketIndex.cpp
```diff
@@ -0,0 +1,241 @@
...
+    // Assert pageSize is a power of 2 for disk IO optimization
+    releaseAssert((pageSize & (pageSize - 1)) == 0);
+    releaseAssert(!filename.empty());
+
+    auto timer = LogSlowExecution("Indexing bucket");
+
+    size_t const estimatedLedgerEntrySize =
+        xdr::xdr_traits<BucketEntry>::serial_size(BucketEntry{});
+    auto fileSize = fs::size(filename);
+    auto estimatedNumElems = fileSize / estimatedLedgerEntrySize;
```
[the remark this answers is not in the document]
SirTyson: It may not matter depending on how good the estimation is, but strictly more is better. According to the bloom filter library the estimate only needs to be within an order of magnitude of the actual number. However playing around with [this calculator](https://hur.st/bloomfilter), it looks like an underestimation of 1% increases the false positive rate by 7%, while an underestimation of 5% increases false positives by 40%. Overestimation results in more memory usage, but that is much more preferable to an increase in false positive rate. I don't think we should log bloom filter setup every time, but maybe just a warning if the error is larger than some percentage.

src/bucket/Bucket.h
```diff
@@ -27,30 +28,58 @@ namespace stellar
  */
 
 class Application;
+class Config;
```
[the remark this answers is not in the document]
marta-lokhova: minor, but BucketList and Database are still there

src/catchup/CatchupWork.cpp
```diff
@@ -601,6 +619,13 @@ CatchupWork::runCatchupStep()
                 mBucketVerifyApplySeq = downloadApplyBuckets();
                 seq.push_back(mBucketVerifyApplySeq);
             }
+            // If there are no buckets to apply, still need to index BL before
+            // ledgers can be replayed
+            else if (mApp.getConfig().EXPERIMENTAL_BUCKETLIST_DB)
+            {
+                auto indexBuckets = std::make_shared<IndexBucketsWork>(mApp);
```
marta-lokhova, line 626: shouldn't this be taken care of by indexing on startup?

src/bucket/BucketList.cpp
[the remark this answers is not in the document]
marta-lokhova: thanks for cleaning this up!

src/bucket/BucketOutputIterator.cpp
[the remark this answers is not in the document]
marta-lokhova: makes sense! resolving

src/bucket/BucketIndex.h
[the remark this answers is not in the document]
marta-lokhova: ok. it looks like the bounds check is still missing though

src/bucket/BucketIndex.h
```diff
@@ -0,0 +1,142 @@
...
+
+// Index maps a range of BucketEntry's to the associated offset
+// within the bucket file. Index stored as two vectors, one stores
+// LedgerKey ranges sorted in the same scheme as LedgerEntryCmp, the other
+// stores offsets into the bucket file for a given key/ key range. pageSize
+// determines how large, in bytes, each range should be. pageSize == 0 indicates
+// that individual index.
+template <class IndexT> class BucketIndexImpl : public BucketIndex
+{
+    IndexT mKeys{};
```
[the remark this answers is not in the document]
marta-lokhova: this doesn't seem relevant anymore, as we now use pairs

src/catchup/CatchupWork.cpp
SirTyson: Yes, removed this block.

src/bucket/BucketOutputIterator.cpp
```diff
@@ -129,14 +132,33 @@ BucketOutputIterator::getBucket(BucketManager& bucketManager,
...
+    // If this bucket needs to be indexed and is not already indexed
+    if (shouldSynchronouslyIndex)
+    {
+        // either it's a new bucket or we just reconstructed a bucket
+        // we already have, in any case ensure we have an index
+        if (auto b = bucketManager.getBucketIfExists(hash);
+            !b || !b->isIndexed())
+        {
+            std::atomic_bool exit{false};
+            index = BucketIndex::createIndex(bucketManager.getConfig(),
```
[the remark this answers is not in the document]
SirTyson: I've changed the early exit interface to stop indexing early if the `BucketManager` is shutting down instead of relying on the passed in `atomic_bool`, so this should be resolved.

src/bucket/BucketIndex.cpp
```diff
@@ -0,0 +1,339 @@
...
+
+    // Conver to bytes
+    auto cutoff = cfg.EXPERIMENTAL_BUCKETLIST_DB_INDEX_CUTOFF * 1000000;
+    if (pageSize == 0 || fs::size(filename.string()) < cutoff)
+    {
+        CLOG_INFO(Bucket,
+                  "BucketIndex::createIndex() indexing individual keys in "
+                  "bucket {}",
+                  filename);
+        return std::unique_ptr<BucketIndexImpl<IndividualIndex> const>(
```
MonsieurNicolas, line 252: I think it may be incorrect to return a unique ptr like this if indexing was not completed for whatever reason (both variants)? In particular the index is partial when aborting on shutdown.
In other places, the assumption is that if not null then the index is valid

MonsieurNicolas (changes requested):
Looks very close to me!

src/bucket/BucketIndex.cpp
SirTyson: I agree that partial indexes shouldn't be returned. I initially had the index constructor throw if a shutdown was detected, but error messages were messy since the exception would not get caught. I'd like to keep the unique_ptr return interface, so maybe I'll put this throw back but add a try-catch in `createIndex`. If the `BucketIndex` constructor thows, `createIndex` can return `nullptr` so we still exit early but don't have an invalid index floating around.

src/bucket/BucketIndex.cpp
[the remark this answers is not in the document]
SirTyson: I'll add this message as a WARNING for now if we underestimate. Maybe is should be INFO instead? I will definitely revisit this issue before BucketListDB is on by default, but for this experimental PR I'll just add the log message.

src/util/XDRStream.h
```diff
@@ -118,10 +136,78 @@ class XDRInputFileStream
...
         }
+
         xdr::xdr_get g(mBuf.data(), mBuf.data() + sz);
         xdr::xdr_argpack_archive(g, out);
         return true;
     }
+
+    template <typename T>
+    bool
+    readPage(T& out, LedgerKey const& key, size_t pageSize)
```
graydon, line 147: A comment here would be good, explaining what `readPage` is doing and what its parameters mean.

src/util/XDRStream.h
```diff
@@ -118,10 +136,78 @@ class XDRInputFileStream
...
+            }
+        }
+
+        auto bytesToRead = mIn.gcount();
+        char* curr = mBuf.data();
+        for (std::streamsize read = 0; read < bytesToRead;)
+        {
+            ZoneNamedN(__unpack, "xdr_unpack_entry", true);
+
+            uint32_t sz = getXDRSize(curr);
```
graydon, line 173: This can do an out-of-bounds read if there aren't 4 bytes left in the buffer.

src/util/XDRStream.h
```diff
@@ -118,10 +136,78 @@ class XDRInputFileStream
...
+            read += sizeof(sz);
+            curr += sizeof(sz);
+
+            // If entry continues past end of page
+            if (read + sz > bytesToRead)
+            {
+                if (mBuf.size() < sz)
+                {
+                    mBuf.resize(sz);
```
graydon, line 182: This resize can invalidate `curr` which will make the next read a memory error.

src/util/XDRStream.h
```diff
@@ -118,10 +136,78 @@ class XDRInputFileStream
...
+                }
+
+                // Move undecoded bytes to front of buffer and read in the rest
+                // of the entry
+                auto bytesRemaining = bytesToRead - read;
+                memmove(mBuf.data(), curr, bytesRemaining);
```
graydon, line 188: I'm not really sure why the `memmove` is happening here, nor why it is enough to make space for the subsequent read to succeed (I think the previous resize likely _shrank_ the buffer?); at this point the offset and pointer arithmetic becomes too unclear for me to follow.

I think it would make more sense to expand the buffer to accommodate the remainder, rather than maybe-shrink or maybe-expand it to cover the current record. And fill only that remainder piece. I'll post a change that does this (and fixes the other issues).

src/ledger/test/LedgerCloseMetaStreamTests.cpp
```diff
@@ -421,7 +421,8 @@ TEST_CASE("METADATA_DEBUG_LEDGERS works", "[metadebug]")
     while (lm.getLastClosedLedgerNum() < (2 * cfg.METADATA_DEBUG_LEDGERS))
     {
         clock.crank(false);
-        if (app->getWorkScheduler().allChildrenDone())
+        if (app->getWorkScheduler().allChildrenDone() &&
+            lm.getLastClosedLedgerNum() >= cfg.METADATA_DEBUG_LEDGERS)
```
graydon, line 425: This is comparing an absolute ledger number to a count of ledgers -- is this intentional?

src/util/XDRStream.h
```diff
@@ -118,10 +136,78 @@ class XDRInputFileStream
...
+
+        if (!mIn.read(mBuf.data(), pageSize))
+        {
+            if (mIn.eof())
+            {
+                mIn.clear(std::ios_base::eofbit);
+            }
+            else
+            {
+                return false;
```
graydon, line 163: I also _think_ this `failbit` handling is wrong (as is the handling in `readOne` above, which of course I wrote) though I've only been banging my head against this part of `iostreams` for 25 years and it's possible I'm getting it wrong again. I _think_ if `mIn` is false after the read then either there was an IO fault (`badbit`) or the operation logically failed (`failbit`) and if EOF was the _cause_ of the logical failure then `eofbit` will also be set. So we want to treat anything _other_ than `eofbit` as an actual problem and escalate by throwing, not returning false. I think.

graydon: (I will fiddle with this locally to try to confirm)

src/bucket/BucketIndex.cpp
```diff
@@ -0,0 +1,339 @@
...
+    ZoneValue(static_cast<int64_t>(mKeysToOffset.size()));
+
+    auto internalStart = std::get<typename IndexT::const_iterator>(start);
+    auto keyIter =
+        std::lower_bound(internalStart, mKeysToOffset.end(), k,
+                         lower_bound_pred<typename IndexT::value_type>);
+
+    // If the key is not in the bloom filter or in the lower bounded index
+    // entry, return nullopt
+    if ((mFilter && !mFilter->contains(std::hash<stellar::LedgerKey>()(k))) ||
```
graydon, line 291: Bloom filters don't give false negatives, only false positives, right? Can you not check the filter before the binary search? (I tried it myself here and it does not seem to work, which suggests I don't understand something about how this works)

src/bucket/BucketIndex.cpp
```diff
@@ -0,0 +1,339 @@
...
+    size_t estimatedIndexEntries;
+
+    // Initialize bloom filter for range index
+    if constexpr (std::is_same<IndexT, RangeIndex>::value)
+    {
+        ZoneNamedN(bloomInit, "bloomInit", true);
+        bloom_parameters params;
+        params.projected_element_count = estimatedNumElems;
+        params.false_positive_probability = 0.001; // 1 in 1000
+        crypto_shorthash_keygen(params.random_seed.data());
```
graydon, line 110: Unfortunately this is setting the random seed using real entropy which makes the program nondeterministic in testing. Need to use `gRandomEngine`.

MonsieurNicolas: Or maybe use the same `gKey` that we use in shorthash.cpp ?

src/catchup/IndexBucketsWork.cpp
```diff
@@ -0,0 +1,134 @@
...
+        std::static_pointer_cast<IndexWork>(shared_from_this()));
+    app.postOnBackgroundThread(
+        [&app, &b = mBucket, weak]() {
+            auto self = weak.lock();
+            if (!self || self->isAborting())
+            {
+                return;
+            }
+
+            b->setIndex(BucketIndex::createIndex(app.getBucketManager(),
```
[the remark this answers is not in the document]
marta-lokhova: There's a check on line 51 that exits early if work is aborting, so it doesn't seem like we need to check that here

src/history/test/HistoryTests.cpp
```diff
@@ -585,88 +612,102 @@ TEST_CASE("History catchup", "[history][catchup][acceptance]")
...
     CatchupSimulation catchupSimulation{VirtualClock::REAL_TIME};
     auto checkpointLedger = catchupSimulation.getLastCheckpointLedger(3);
-    auto app = catchupSimulation.createCatchupApplication(
-        std::numeric_limits<uint32_t>::max(), Config::TESTDB_ON_DISK_SQLITE,
-        "app");
 
-    auto offlineNonCheckpointDestinationLedger =
-        checkpointLedger -
-        app->getHistoryManager().getCheckpointFrequency() / 2;
+    auto f = [&](bool useBucketListDB) {
```
marta-lokhova, line 616: Similarly here, we don't need to test the new flag for each catchup trigger condition (this test mostly focuses on how catchup is triggered, so independent of the new flag). These tests already take really long time to run, so we only want to test the cases where the flag actually matters.

marta-lokhova: This applies to other tests in this file, "History prefix catchup", "History catchup with different modes" etc. 

 I recommend writing just one new test, which trigger basic offline catchup with bucket application - this should be sufficient to test out the flow with BucketsDB. In this new test, just in case we should also have two versions of the app publishing original checkpoints, one with the flag and the other without. This way we'd also verify that publish flow is unaffected by the flag.

src/catchup/test/CatchupWorkTests.cpp
```diff
@@ -124,115 +124,141 @@ std::vector<std::pair<uint32_t, CatchupConfiguration>> gCatchupRangeCases{
...
+        f(cfg);
+    }
+}
 
-            CLOG_DEBUG(History,
-                       "computed CatchupRange: first={}, last={}, "
-                       "replayFirst={}, replayCount={}, replayLimit={}",
-                       range.first(), range.last(), range.getReplayFirst(),
-                       range.getReplayCount(), range.getReplayLimit());
+TEST_CASE("compute CatchupRange from CatchupConfiguration", "[catchup]")
```
marta-lokhova, line 144: Same thing here: these tests just verify that catchup configuration is correct, so it's unrelated to the flag.

src/bucket/BucketManagerImpl.cpp
```diff
@@ -395,6 +397,13 @@ BucketManagerImpl::adoptFileAsBucket(std::string const& filename,
         {
             auto timer = LogSlowExecution("Delete redundant bucket");
             std::remove(filename.c_str());
+
+            // race condition: two buckets + indexes were produced in parallel
```
marta-lokhova, line 401: mm, not quite following the comment here - what's the race condition? (I might be blanking, but probably good to elaborate in the comment anyway)

src/bucket/BucketOutputIterator.cpp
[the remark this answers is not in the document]
marta-lokhova: thanks!

src/history/test/HistoryTests.cpp
marta-lokhova: actually, publish isn't touched by the bucketlist DB changes right? So maybe that change isn't necessary

src/bucket/BucketManagerImpl.cpp
```diff
@@ -584,6 +612,20 @@ BucketManagerImpl::getReferencedBuckets() const
...
+
+    return referenced;
+}
+
+std::set<Hash>
+BucketManagerImpl::getAllReferencedBuckets() const
+{
+    ZoneScoped;
+    auto referenced = getBucketListReferencedBuckets();
+    if (!mApp.getConfig().MODE_ENABLES_BUCKETLIST)
```
graydon, line 624: I don't think I understand why this function was split in two / why this early return happens here.

src/bucket/BucketManagerImpl.cpp
```diff
@@ -656,24 +698,41 @@ BucketManagerImpl::forgetUnreferencedBuckets()
...
+        // other in-progress structures will add bucket to bucket list after
+        // deleting index
+        if (j->second->isIndexed() && j->second.use_count() == 1 &&
+            blReferenced.find(j->first) == blReferenced.end())
+        {
+            CLOG_TRACE(Bucket,
+                       "BucketManager::forgetUnreferencedBuckets deleting "
+                       "index for {}",
+                       j->second->getFilename());
+            j->second->freeIndex();
```
graydon, line 724: Is there really any need for a separate method `freeIndex` at all? I mean .. would it not be adequate to have the bucket keep its index alive as long as the bucket is alive? Do we have buckets that get an index, and then we delete the index, but then also somehow keep the bucket alive?

graydon (approved):
Handful of questions none of which are very serious -- this generally looks great and seems to work (been running for a couple days here and did a 100k ledger catchup no problem).

src/util/XDRStream.h
graydon: (have posted changes to this PR to resolve this)

graydon (changes requested):
Actually it'd be good at least to change the random seed issue. The others are more issues of curiosity / possible cleanups.

src/catchup/IndexBucketsWork.cpp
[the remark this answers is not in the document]
SirTyson: I think it still serves some purpose. Creating the index could take a significant amount of time (minutes), so it's very possible that the Work was not aborting when it started indexing but is now. In practice, it doesn't matter because we return null indexes if we detect the system is aborting, but still probably good defensive practice to not start indexing if you are aborting and to not set the index if you are aborting.

src/main/Config.h
marta-lokhova: Looks like this flag was removed, so I'm resolving this.

src/catchup/AssumeStateWork.cpp
```diff
@@ -0,0 +1,59 @@
...
+}
+
+BasicWork::State
+AssumeStateWork::doWork()
+{
+    if (!mWorkSpawned)
+    {
+        std::vector<std::shared_ptr<BasicWork>> seq;
+
+        mApp.getBucketManager().assumeState(mHas);
```
marta-lokhova, line 29: I'm worried that here we assign partially valid (unindexed) buckets to the bucketlist. If something further down goes wrong, then core is left in an invalid state. Safer would be to generate indexes to buckets first, then assign fully valid buckets to the bucketlist. The thing is that this cleanup only matters when the flag is on, so we could do this in a subsequent change

src/bucket/BucketList.cpp
```diff
@@ -241,12 +246,6 @@ BucketList::levelHalf(uint32_t level)
     return levelSize(level) >> 1;
 }
 
-uint32_t
```
marta-lokhova, line 244: looks like the move/rename isn't relevant anymore? would it be possible to revert it, as less changes are easier to review

src/catchup/IndexBucketsWork.cpp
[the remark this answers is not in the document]
marta-lokhova: oh I see, yes you both are right, I missed that there are two actions happening on this line

marta-lokhova (approved):
a few questions related to tests, otherwise this looks good! thank you, this is exciting!

src/crypto/ShortHash.cpp
```diff
@@ -24,6 +24,16 @@ initialize()
     std::lock_guard<std::mutex> guard(gKeyMutex);
     crypto_shorthash_keygen(gKey);
 }
+
+std::array<unsigned char, crypto_shorthash_KEYBYTES>
```
MonsieurNicolas, line 28: no need to make a copy here, just return a const ref

src/bucket/BucketIndex.cpp
```diff
@@ -107,7 +108,7 @@ BucketIndexImpl<IndexT>::BucketIndexImpl(BucketManager const& bm,
         bloom_parameters params;
         params.projected_element_count = estimatedNumElems;
         params.false_positive_probability = 0.001; // 1 in 1000
-        crypto_shorthash_keygen(params.random_seed.data());
+        crypto_shorthash_keygen(shortHash::getShortHashInitKey().data());
```
MonsieurNicolas, line 113: this is not what you want, this should be `params.random_seed = shortHash::getShortHashInitKey()`

src/bucket/BucketManagerImpl.cpp
```diff
@@ -115,6 +116,10 @@ BucketManagerImpl::BucketManagerImpl(Application& app)
     , mBucketSnapMerge(app.getMetrics().NewTimer({"bucket", "snap", "merge"}))
     , mSharedBucketsSize(
           app.getMetrics().NewCounter({"bucket", "memory", "shared"}))
+    , mBucketListDBQueryMeter(app.getMetrics().NewMeter(
```
MonsieurNicolas, line 119: need to add those to `metrics.md`

src/bucket/BucketManagerImpl.cpp
```diff
@@ -859,6 +864,71 @@ BucketManagerImpl::maybeSetIndex(std::shared_ptr<Bucket> b,
     }
 }
 
+medida::TimerContext
+BucketManagerImpl::getBulkLoadTimer(std::string const& label) const
+{
+    mBucketListDBQueryMeter.Mark();
+    return mApp.getMetrics()
+        .NewTimer({"bucketlistDB", "bulk", label})
```
MonsieurNicolas, line 872: From what I remember, medida is dad at this and this will be too slow given how often those metrics get used: this creates a new metric (expensive) and then throws it away every time, so you need to manage your own set/vector.

MonsieurNicolas (approved):
Just a few minor things to take care of and I don't need to review those updates,
I'll let @marta-lokhova / @graydon do the final sign off

src/util/XDRStream.h
SirTyson: I was using the `memove` to avoid having to resize the buffer and instead just throw out all the stuff we already read, move the unread portion to the front of the buffer, then read in the rest of the data that we need. I like yours a LOT better.

src/bucket/BucketIndex.cpp
SirTyson: I went back and forth on this one actually. Initially, I did check the bloom filter first and exit early before doing the `std::lower_bound` search. However, when you check the bloom filter first, the `scan` interface isn't quite right. The caller expects `scan` to return an iterator to `k` or to the first element greater than `k`. If we check the bloom filter before binary search, this interface is broken. We either just return the `start` iterator passed in as a parameter or some sort of null iterator. If we return a null iterator, the caller loses there place in the index, so follow up calls to `scan` will be less efficient. If we return `start`, everything works but we're not following the interface properly. For me, binary search first seems like the correct option. This way we can return the correct iterator that the caller expects, but still use the bloom filter to avoid expensive disk io. I think if we change the interface to return `start` if the key is not in the bloom filter, we are slightly more efficient. However this means the `scan` interface is different between `RangeIndex` and `IndividualIndex`, which is an implementation detail I'd rather not leak to the caller.

src/bucket/BucketManagerImpl.cpp
SirTyson: This has to do with freeing indexes, see comment on `freeIndex`.

src/bucket/BucketManagerImpl.cpp
SirTyson: This is to free memory for buckets that are no longer referenced by the `BucketList` (i.e., will never be searched again). These are buckets referenced by the LCL and the publish queue. There's no reason for these buckets to sit around in the queue with (potentially very large) indexes. This is also why we split `getReferencedBuckets` into `getBucketListReferencedBuckets` and `getAllReferencedBuckets`. Buckets that are referenced by the BucketList are not eligible to free their index, while the other referenced buckets are eligible.

src/bucket/BucketManagerImpl.cpp
SirTyson: `adoptFileAsBucket` is supposed to be thread safe and called by both the main and worker threads. Sometimes, the same bucket exists in different levels of the BucketList. On startup, these identical buckets can be produced in parallel since they are on different levels and both call `adoptFileAsBucket`. In practice, there shouldn't be a race for the indexes since we check for this case in `IndexBucketsWork` and only index identical buckets once, but we keep `maybeSetIndex` here for safety.

Even though this shouldn't happen on startup, there are valid reasons to call `adoptFileAsBucket` on a bucket that already has an index after startup, so we can't just assert that the bucket is not indexed. For example, if a snap bucket merges with an empty bucket during normal operation after startup, the index will exist twice: once for the bucket currently part of the BucketList, and another for the result of the merge between that bucket and the empty level.

While we may not strictly need this race condition check since we try very hard not to double index buckets, it's there for defensive purposes.

src/catchup/AssumeStateWork.cpp
SirTyson: This is a good point, but will take a fair bit of work since `Buckets` referenced by `HistoryArchiveState` are not yet managed by `BucketManager`. Commenting here for visibility, will address in follow-up PR.

marta-lokhova: yeah, it might be annoying to change (you'd need to pass around a list of bucket pointers not attached to the bucketlist instead I think). Best is to open an issue, so we can track this.

src/history/test/HistoryTests.cpp
SirTyson: Correct, I'll just add the one test without publish queue.

SirTyson:
This "should" be ready to merge and all the major concerns have been addressed. There are a few outstanding issues that need to be addressed in a follow-up update, but are outside the scope of this PR: #3604, #3605, and (maybe) #3606.

src/catchup/test/CatchupWorkTests.cpp
marta-lokhova: looks like this one still needs to be addressed

src/bucket/Bucket.cpp
```diff
@@ -29,10 +29,274 @@
...
+            else
+            {
+                mKeys.back().upperBound = key;
+            }
+        }
+
+        pos = in.pos();
+    }
+
+    CLOG_INFO(Bucket, "Indexed {} positions in {}", mKeys.size(),
```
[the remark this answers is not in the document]
graydon: This is still CLOG_INFO where it should be CLOG_DEBUG.

src/bucket/BucketIndex.cpp
graydon: Ok. I suspect it's a bit of a speed hit but I understand the rationale. It might be nice to put a comment in about this, because any future reader who knows "bloom filters are to save you work when it gets a negative hit" will consider hoisting it up.

src/bucket/BucketManagerImpl.cpp
graydon: ok

graydon:
@SirTyson I resolved _almost_ all my comments. The bloom filter late-check part should get a comment, I think we should bump logging from INFO down to DEBUG, but neither of those is a blocker. The fact that the CI run crashed on current is worrying. Any idea what that's about?

src/bucket/BucketIndex.h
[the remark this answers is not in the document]
marta-lokhova: ^ looks like this comment got lost

src/bucket/BucketList.cpp
```diff
@@ -82,7 +86,7 @@ BucketList::shouldMergeWithEmptyCurr(uint32_t ledger, uint32_t level)
     {
         // Round down the current ledger to when the merge was started, and
         // re-start the merge via prepare, mimicking the logic in `addBatch`
-        auto mergeStartLedger = mask(ledger, BucketList::levelHalf(level - 1));
+        auto mergeStartLedger = roundDown(ledger, BucketList::mask(level - 1));
```
marta-lokhova, line 89: something went wrong with the rename here - why is `levelHalf` named `mask` now? In the earlier comment, I was saying that simplest is to revert this change completely as the refactor isn't needed anymore

SirTyson: Sorry, I think I messed up my vim command. Interesting that it still passed tests. While I was fixing it though, I spotted that this rename actually is useful since `BucketIndex` uses this function as well when doing page aligning index entries. I'm reverting the erroneous refactor, but I think it's best to keep the rename for `roundDown` since it's used outside of the `BucketList` now.

src/bucket/BucketIndex.h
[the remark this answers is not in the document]
SirTyson: I believe I addressed the bounds check with `                  releaseAssert(rangeEntry.upperBound < key);` in the constructor for index `BucketIndexImpl`. This just checks to make sure the bounds are correct when constructing the range index.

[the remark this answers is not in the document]
marta-lokhova: oh I thought you were going to enforce this in the constructor/at assignment in `RangeEntry` itself (in case there are other call sites now or in the future). This is not critical for landing though, and can be cleaned up later, so we can consider this comment resolved

src/bucket/BucketManagerImpl.cpp
marta-lokhova: thanks for explaining, I seem to keep forgetting that we can produce identical valid buckets on different levels 😅

src/bucket/Bucket.cpp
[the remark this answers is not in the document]
SirTyson: I'm not sure what you're referencing here? I believe this message is `CLOG_DEBUG`, it's just been moved to `BucketIndex.cpp`?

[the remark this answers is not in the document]
SirTyson: Are you referring to the log messages in `BucketIndex::createIndex`?

src/crypto/ShortHash.cpp
SirTyson: We actually do need to copy here. `gKey` must be accessed only when `gKeyMutex` is locked, so we acquire the lock, make a copy, and return the copy. If we just return a direct reference to `gKey`, then we get race conditions and blow up. We also can't return a reference because we copy into a stack array and would return a stack reference. If your try to return a `const&` and return `std::move(arr)`, it works in `gcc` but not in `clang`.

graydon (approved):

graydon:
r+ e401177

dpblnt:
I'm very happy with this improvement. On my watcher node(with horizon) disk IO went from 100% to 3%, CPU 350% to 100%.

PR #3446, opened 2022-06-07
Require flow control

### LINKED ISSUE #3387: overlay: remove the option to disable flow control
Currently, peers can opt out of flow control. With protocol 19, we have an opportunity to force nodes to upgrade to the latest version of core, so we should take advantage of that and make flow control mandatory for all connections.

marta-lokhova:
Resolves #3387


graydon:
LGTM, assuming the config-param tweaks are the right numbers (not sure about them).

marta-lokhova:
the tweaks are only to the example config, to reflect our defaults for consistency

MonsieurNicolas:
r+ f741003fc6c2f689aa0aa2d17044b8ee26267abb

PR #3447, opened 2022-06-07
Added utility to dump ledger state from buckets to JSON

dmkozh:
# Description

Added utility to dump ledger state from buckets to JSON for debugging (optionally using `jq` for post-processing).

The full ledger in this format currently takes ~19GB and is not really feasible for meaningful processing. Hence initially I introduce the following ways to reduce the output (all of which can be combined):
- Filter expressions for conditionally selection the entries
- Limit to only recently updated entries
- Limit the output entry count

For the sake of filtering I introduced XDR field extraction utility, that can be used for other dynamic operations on XDR, e.g. group-by and/or reduction. However, for now I'm keeping this minimalistic until some concrete use-cases that aren't feasible with the current approach appear.

The current approach is a simple iteration over buckets from newer to older. This probably could be optimized (e.g. with parallel execution or XDR of intermediate results), but for now the performance seems acceptable: ~80s on my laptop for filtering the full ledger with a reasonably small output set.

# Checklist
- [ ] Compiles
- [ ] Ran all tests

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -104,3 +104,6 @@ min-testcases/
 *.vcxproj.user
 **/.vs/*
 
+/src/util/xdrquery/XDRQueryScanner.cpp
+/src/util/xdrquery/XDRQueryParser.h
+/src/util/xdrquery/XDRQueryParser.cpp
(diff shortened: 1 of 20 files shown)


MonsieurNicolas (approved):
this is great! Thank you

MonsieurNicolas:
r+ b7b86386646a60a8cf5328b3c8f5989d832a4560

PR #3451, opened 2022-06-17
Add an option to disable console logging

### LINKED ISSUE #3386: changed logging behavior to avoid duplicate logging
### Right now

We log to console and (if not set to "") to the file defined by `LOG_FILE_PATH`.
We also happen to configure by default `LOG_FILE_PATH` to log to a file in `cwd`.

Impact is that we end up with logs in both the desired log file and in the "daemon log" (like `/var/log/daemon.log`), which are more verbose (they contain an extra timestamp, hostname, and process information):
```
2022-03-26T00:16:41.585514+00:00 localhost stellar-core[1698452]: 2022-03-26T00:16:41.585 GCSKX [Herder INFO] Quorum information for 40192206 : {"agree":23,"cost":26827304,"delayed":0,"disagree":0,"fail_at":6,"hash":"b68a69","lag_ms":1416,"ledger":40192206,"missing":0,"phase":"EXTERNALIZE"}
```

### Instead

We should log to console only if `LOG_FILE_PATH` is empty or if (a new global option) `--console` is passed in.

marta-lokhova:
Resolves #3386 

Note that instead of disabling console logging by default, I added an option to disable it. I find console logging useful most of the time when developing/testing, as it's easier to quickly eyeball the logs instead of having to deal with files. But if anyone feels strongly about this, let me know. Otherwise, I think people who do not want logs in both places simply need to update their run command to include the new flag (I think it's also better this way in case some of the operators rely on default console logging)


MonsieurNicolas:
I think the behavior I proposed in the issue is the more sensible one:
it's very unusual for a daemon to log to stdout/stderr when a log file is specified.

This explains why sysadmins get confused with the existing behavior (including our own ops team).

I agree that for core developers logging to console is probably what people want even when logging to files, but that is niche enough as a use case that opt-in into this is probably the better option.

marta-lokhova:
okay, I updated the PR @MonsieurNicolas

MonsieurNicolas:
r+ caf8842f02548a1f9eeac6470c38994d7b5988d5

PR #3452, opened 2022-06-18
Cross-check XDR file hashes between Rust and C++.

graydon:
This adds the necessary machinery to core for cross-checking that the XDR files converted to C++ via xdrc are the same as those converted to rust using xdrgen (and coming in via dependency on rs-stellar-xdr). In other words it is a companion to PR https://github.com/stellar/xdrgen/pull/89 

It also includes a fix for some test breakage that occurred when https://github.com/stellar/rs-stellar-contract-env/pull/120 landed on the env crate.

cc @leighmcculloch @jonjove


leighmcculloch:
https://github.com/stellar/rs-stellar-xdr/pull/83 and https://github.com/stellar/rs-stellar-contract-env/pull/126 are merged.

MonsieurNicolas:
do we need this to be enforced in code or we could just use github actions for this and catch the problem even earlier?

MonsieurNicolas:
note: you have a merge commit in there

graydon:
> do we need this to be enforced in code or we could just use github actions for this and catch the problem even earlier?

Github is actually a bit _late_ -- the way this is now we will catch it on our own workstations when running `make check`.

MonsieurNicolas:
I was commenting on the core binary doing that check. As you're looking at moving the xdr around it looks to be a build problem not a runtime problem. The "hash of xdr" is a small subset of the kind of things that can go wrong if we mix incompatible SDK and core. It seems to me that we need to think about dependencies a bit more.

graydon:
Note: we had another instance of divergence between the rust XDR inputs and C++ XDR inputs today, based on mistaken commits to the wrong repo. I only noticed when something failed to build: https://github.com/stellar/stellar-xdr-next/pull/21

I appreciate that we need to think about dependencies quite a bit, but "having the same XDR on either side of an FFI" is a baseline part of any correct solution.

MonsieurNicolas:
ok let's merge this then. I was hoping that by now we'd be in a better spot wrt dependencies but whatever helps.

you need to rebase this though.

Also, I made a comment some time ago > "note: you have a merge commit in there".

graydon:
@MonsieurNicolas rebased, updated, cleaned out spurious merge commit, and actually synchronized the input rust and C++ XDRs (they had become out of sync on trunk again, and this change caught it). This should land ASAP.

MonsieurNicolas:
r+ d89c49983ff4420ebfe0a8352bb80732c3a4c669

PR #3454, opened 2022-06-21
Fix failing non-test build due to test in `tests/` subdir

graydon:
The `make-mks` script classifies `.cpp` files as "tests" or "not tests" depending on whether the file occurs in a path with `test/` in it. Unfortunately [the XDRQuery PR](https://github.com/stellar/stellar-core/pull/3447) introduced a test in a directory called `util/xdrquery/tests/` (trailing `s`) which means it gets classified as a non-test.

When building with `--disable-tests` (such as in packages, release builds, docker) this causes the build system to include the test in the build anyways, which causes a link error.

_This_ PR fixes that build breakage by just renaming that `tests` dir to `test`.

cc @leighmcculloch @MonsieurNicolas @dmkozh


dmkozh:
We should also update the VS project, but I could take care of that after this is pushed.

MonsieurNicolas:
I am r+ now to unblock building off master

MonsieurNicolas:
r+ e18ad453176ec3d77745aaaf25d620c151e9fffd

PR #3472, opened 2022-06-30
ifdef-BUILD_TESTS-guard a definition to match its declaration.

graydon:
This just fixes a build failure that slipped through -- only triggers when building with `--disable-tests` for a package / release.


MonsieurNicolas:
r+ 33fa715409476471b1533480ebe7d644eaad0aa8

PR #3491, opened 2022-07-19
Fix the ordering issue and add the respective check.

dmkozh:
# Description

Fix the `TxSetFrame` ordering issue and add the respective check.

Simply adding the check causes crash in tests (e.g. surge pricing), hence I'm not adding any new tests to specifically handle this issue.

# Checklist
- [ ] Compiles
- [ ] Ran all tests

### DIFF
--- a/src/herder/TxSetFrame.cpp
+++ b/src/herder/TxSetFrame.cpp
@@ -521,6 +521,13 @@ TxSetFrame::checkValid(Application& app, uint64_t lowerBoundCloseTimeOffset,
         return false;
     }
 
+    if (!std::is_sorted(mTxs.begin(), mTxs.end(), &TxSetUtils::hashTxSorter))
+    {
+        CLOG_DEBUG(Herder, "Got bad txSet: {} is not in hash order",
+                   hexAbbrev(mPreviousLedgerHash));
+        return false;
+    }
+
     if (std::adjacent_find(mTxs.begin(), mTxs.end(),
                            [](auto const& lhs, auto const& rhs) {
                                return lhs->getFullHash() == rhs->getFullHash();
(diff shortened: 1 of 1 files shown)


src/herder/TxSetFrame.cpp
```diff
@@ -521,6 +521,13 @@ TxSetFrame::checkValid(Application& app, uint64_t lowerBoundCloseTimeOffset,
         return false;
     }
 
+    if (!std::is_sorted(mTxs.begin(), mTxs.end(), &TxSetUtils::hashTxSorter))
```
marta-lokhova, line 524: I think we should use TxSetUtils::isValidHashOrder

marta-lokhova (approved):

src/herder/TxSetFrame.cpp
jonjove: As far as I can tell, this function no longer exists.

jonjove (approved):

src/herder/TxSetFrame.cpp
dmkozh: Yeah, I removed it because it has been used in a single place before and the comparator is now exposed... I guess we could bring it back, given that we need this twice again now.

src/herder/TxSetFrame.cpp
```diff
@@ -521,6 +521,13 @@ TxSetFrame::checkValid(Application& app, uint64_t lowerBoundCloseTimeOffset,
...
+    {
+        CLOG_DEBUG(Herder, "Got bad txSet: {} is not in hash order",
+                   hexAbbrev(mPreviousLedgerHash));
```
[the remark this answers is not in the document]
dmkozh: Agree, I was a bit confused about this yesterday too. This logging existed before my changes and I think more useful thing to log would be just the current ledger seq. Also I think this should be promoted to `CLOG_ERROR` as this code is only being invoked for sanity checking now and would crash the node if the tx set is incorrect.

src/herder/TxSetFrame.cpp
marta-lokhova: ah sorry, was looking at the wrong branch. All good then!

PR #3492, opened 2022-08-01
Extend load generator to produce high volume traffic

dmkozh:
# Description

Extend load generator to produce high volume traffic that easily triggers limiting and surge pricing logic. Also added fee diversification and did some refactoring to easier introduce new parameters.

# Checklist
- [ ] Compiles
- [ ] Ran all tests

### DIFF
--- a/docs/software/commands.md
+++ b/docs/software/commands.md
@@ -320,24 +320,33 @@ format.
   is started
 
 ### The following HTTP commands are exposed on test instances
-* **generateload**
-  `generateload[?mode=(create|pay|pretend)&accounts=N&offset=K&txs=M&txrate=R&batchsize=L&spikesize=S&spikeinterval=I]`<br>
-  Artificially generate load for testing; must be used with
-  `ARTIFICIALLY_GENERATE_LOAD_FOR_TESTING` set to true.
-  * `create` mode creates new accounts.
-    Additionally, allows batching up to 100 account creations per transaction via 'batchsize'.
+* **generateload** `generateload[?mode=
+    (create|pay|pretend)&accounts=N&offset=K&txs=M&txrate=R&batchsize=L&spikesize=S&spikeinterval=I&maxfeerate=F&skiplowfeetxs=(0|1)]`
+
+    Artificially generate load for testing; must be used with
+    `ARTIFICIALLY_GENERATE_LOAD_FOR_TESTING` set to true.
+  * `create` mode creates new accounts. Additionally, allows batching up to 100
+    account creations per transaction via `batchsize`.
   * `pay` mode generates `PaymentOp` transactions on accounts specified
     (where the number of accounts can be offset).
-  * `pretend` mode generates transactions on accounts specified
-    (where the number of accounts can be offset). Operations in `pretend` mode are
-    designed to have a realistic size to help users "pretend" that they have real traffic.
+  * `pretend` mode generates transactions on accounts specified(where the number
+    of accounts can be offset). Operations in `pretend` mode are designed to
+    have a realistic size to help users "pretend" that they have real traffic.
     You can add optional configs `LOADGEN_OP_COUNT_FOR_TESTING` and
     `LOADGEN_OP_COUNT_DISTRIBUTION_FOR_TESTING` in the config file to specify
-    the # of ops / tx and how often they appear. More specifically, the probability
-    that a transaction contains `COUNT[i]` ops is
-    `DISTRIBUTION[i] / (DISTRIBUTION[0] + DISTRIBUTION[1] + ...)`.
+    the # of ops / tx and how often they appear. More specifically, the
+    probability that a transaction contains `COUNT[i]` ops is `DISTRIBUTION
+    [i] / (DISTRIBUTION[0] + DISTRIBUTION[1] + ...)`.
 
-  For `pay` and `pretend`, when a nonzero I is given, a spike will occur every I seconds injecting S transactions on top of `txrate`.
+  Non-`create` load generation makes use of the additional parameters:
+  * when a nonzero `spikeinterval` is given, a spike will occur every
+    `spikeinterval` seconds injecting `spikesize` transactions on top of
+    `txrate`
+  * `maxfeerate` defines the maximum per-operation fee for generated
+    transactions (when not specified only minimum base fee is used)
+  * when `skiplowfeetxs` is set to `true` the transactions that are not accepted by
+    the node due to having too low fee to pass the rate limiting are silently
+    skipped. Otherwise (by default), such transactions would cause load generation to fail.
 
 * **manualclose**
   If MANUAL_CLOSE is set to true in the .cfg file, this will cause the current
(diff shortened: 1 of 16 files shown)


src/simulation/LoadGenerator.cpp
```diff
@@ -570,6 +603,15 @@ LoadGenerator::maybeHandleFailedTx(TestAccountPtr sourceAccount,
     if (status == TransactionQueue::AddResult::ADD_STATUS_ERROR &&
         code == txBAD_SEQ)
     {
+        auto* herder = static_cast<HerderImpl*>(&mApp.getHerder());
```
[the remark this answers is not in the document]
marta-lokhova: That's fine with me in the short term (we can still put the getter behind BUILD_TESTS). Long-term like you said we'd probably want to step back and rethink herder vs tx queue relationship, but that's outside of the scope of the PR.

src/simulation/LoadGenerator.cpp
```diff
@@ -49,6 +50,10 @@ const uint32_t LoadGenerator::TX_SUBMIT_MAX_TRIES = 10;
 // ledger.
 const uint32_t LoadGenerator::TIMEOUT_NUM_LEDGERS = 20;
 
+// After successfully submitting desired load, wait for this many ledgers
+// without checking for account consistency.
+const uint32_t LoadGenerator::COMPLETION_TIMEOUT_WITHOUT_CHECKS = 4;
```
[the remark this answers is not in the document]
marta-lokhova: Oh, I see. Should we just use one parameter across all loadgen modes for consistency? I think as soon loadgen is done submitting load, the transactions would either be picked up within the next 3 ledgers or dropped, so it _seems_ like there isn't much value in waiting for 20 ledgers (I can't recall how we landed on 20, it might have been an arbitrary choice. @MonsieurNicolas  might know better)

docs/software/commands.md
```diff
@@ -311,24 +311,33 @@ format.
...
+    probability that a transaction contains `COUNT[i]` ops is `DISTRIBUTION
+    [i] / (DISTRIBUTION[0] + DISTRIBUTION[1] + ...)`.
+
+  Non-`create` load generation makes use of the additional parameters:
+  * when a nonzero `spikeinterval` is given, a spike will occur every
+    `spikeinterval` seconds injecting `spikesize` transactions on top of
+    `txrate`
+  * `maxfeerate` defines the maximum per-operation fee for generated
+    transactions (when not specified only minimum base fee is used)
+  * when `skiplowfeetxs` is set to `1` the transactions that are not accepted by
```
marta-lokhova, line 338: nit: we use explicit "true" and "false" for bools in other http commands, so it'd be nice to do the same here for consistency.

src/simulation/LoadGenerator.cpp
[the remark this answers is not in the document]
dmkozh: Ok, I've exposed it from the interface (with `BUILD_TESTS`). I've actually realized it's test-only in impl too, so there is really no good reason to not expose it (at least in the test builds)

docs/software/commands.md
dmkozh: Done.

src/simulation/LoadGenerator.cpp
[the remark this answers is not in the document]
dmkozh: I guess for the old constant we wouldn't normally reach the limit (unless something is broken), so there was no point in fine-tuning it. For the new constant we always have to wait this much. I'm not opposed to using a single small constant, though maybe would do that separately to be safe.

src/simulation/LoadGenerator.cpp
```diff
@@ -700,12 +752,32 @@ LoadGenerator::TxMetrics::report()
...
+        auto baseFee = mApp.getLedgerManager().getLastTxFee();
+        if (baseFee > *maxGeneratedFeeRate)
+        {
+            throw std::runtime_error(
+                "maxfeerate is smaller than minimum base fee");
+        }
+        auto feeRateDistr =
+            uniform_int_distribution<uint32_t>(baseFee, *maxGeneratedFeeRate);
+        // Add a bit more fee to get non-integer fee rates, such that
+        // `ceil(fee / ops.size()) == feeRate`, but
```
marta-lokhova, line 772: should be `feeRate + 1`, right?

src/main/CommandHandler.cpp
```diff
@@ -213,6 +213,16 @@ parseOptionalParamOrDefault(std::map<std::string, std::string> const& map,
...
 }
 
+template <>
+bool
+parseOptionalParamOrDefault<bool>(std::map<std::string, std::string> const& map,
+                                  std::string const& key,
+                                  bool const& defaultValue)
+{
+    auto paramStr = parseOptionalParam<std::string>(map, key);
+    return paramStr && *paramStr == "true";
```
marta-lokhova, line 223: this condition doesn't respect the default value (if option is not provided, it should return the default)

src/simulation/LoadGenerator.cpp
```diff
@@ -700,12 +752,36 @@ LoadGenerator::TxMetrics::report()
...
 {
-    auto txf = from->tx(ops);
+    int fee = 0;
 
+    if (maxGeneratedFeeRate)
+    {
+        auto baseFee = mApp.getLedgerManager().getLastTxFee();
+        if (baseFee > *maxGeneratedFeeRate)
+        {
+            throw std::runtime_error(
```
marta-lokhova, line 766: hmm, this isn't an internal bug, just bad parameters from the user, so no need to crash here (need to gracefully fail instead). Also, is this the right place for the check? We could fast fail in CommandHandler right after parsing maxGeneratedFeeRate (and return an appropriate error message to the user in retStr)

src/main/CommandHandler.cpp
dmkozh: Nice catch, fixed.

src/simulation/LoadGenerator.cpp
```diff
@@ -4,6 +4,7 @@
 
 #include "simulation/LoadGenerator.h"
 #include "herder/Herder.h"
+#include "herder/HerderImpl.h"
```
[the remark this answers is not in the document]
dmkozh: Done.

src/simulation/LoadGenerator.cpp
dmkozh: True, moved the check to `CommandHandler`.

src/simulation/LoadGenerator.cpp
dmkozh: Actually it should be `floor` instead (when added value is 0 `ceil` would be equal to `feeRate`, so `+ 1` won't be correct).

marta-lokhova (approved):
LGTM! Just need to squash the commits

dmkozh:
> LGTM! Just need to squash the commits

Rebased & squashed.

MonsieurNicolas:
r+ 31566c8a7a660e08ad3ba93b8e3e52ce4a6b5035

PR #3500, opened 2022-08-05
First sketch of InvokeHostFunctionOp preflight endpoint.

graydon:
This implements a new `preflight?blob=<InvokeHostFunctionOp>` HTTP endpoint which then runs the requested op _synchronously_ against the current ledger in preflight / footprint-recording mode, and reports back

- the host-function call result
- the storage footprint that was recorded
- the cpu instruction and memory byte counts (since we don't currently have a way to translate these into "gas costs")

It also adds a new `getledgerentry?key=<LedgerKey>` HTTP endpoint, which returns a status, encoded ledger entry, and ledger number (for consistency validation on the caller), because there's some ongoing discussion about doing preflight in the CLI client / in an RPC host / closer to the operator and it's harmless to include at this point (it's short and not part of release builds or anything, easy to remove later).

I also very slightly cleaned up the `tx?blob=` endpoint because it kinda looks like nobody has touched it in 7 years and it's a little messy. And I slightly cleaned up the existing call path for `invoke_host_function`.

Addresses #3484 and https://github.com/stellar/soroban-cli/issues/63

### DIFF
--- a/src/main/CommandHandler.cpp
+++ b/src/main/CommandHandler.cpp
@@ -18,6 +18,11 @@
 #include "overlay/BanManager.h"
 #include "overlay/OverlayManager.h"
 #include "overlay/SurveyManager.h"
+#ifdef ENABLE_NEXT_PROTOCOL_VERSION_UNSAFE_FOR_PRODUCTION
+#include "ledger/InternalLedgerEntry.h"
+#include "ledger/LedgerTxnImpl.h"
+#include "transactions/InvokeHostFunctionOpFrame.h"
+#endif
 #include "transactions/TransactionBridge.h"
 #include "transactions/TransactionUtils.h"
 #include "util/Logging.h"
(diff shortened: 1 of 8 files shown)


src/main/CommandHandler.cpp
```diff
@@ -652,17 +663,83 @@ CommandHandler::ll(std::string const& params, std::string& retStr)
...
+        fromOpaqueBase64(op, blob);
+        root = InvokeHostFunctionOpFrame::preflight(mApp, op);
+    }
+    else
+    {
+        throw std::invalid_argument(
+            "Must specify an InvokeHostFunctionOp blob: preflight?blob=<op in "
+            "base64 XDR format>");
+    }
+    retStr = root.toStyledString();
```
MonsieurNicolas, line 689: `toStyledString` is not the right writer: this one produces a human readable string (with tabs etc), the way to do it is `Json::FastWriter().write(root)`

MonsieurNicolas: applies to all changes you made in this file, but not all endpoints (for example, the `quorum`, `info` endpoints are commonly used by operators and are not performance sensitive)

MonsieurNicolas (changes requested):

src/main/CommandHandler.cpp
graydon: Fixed.

MonsieurNicolas:
r+ 97cd4aaaec26281f7e5dcf96621ed4f380fc774a

PR #3501, opened 2022-08-09
Avoid persisting redundant tx sets

### LINKED ISSUE #3499: SCP state persistance is inefficient
Currently, whenever core persists new SCP state, it persists all tx sets, even though it likely already saved the same tx sets in the previous checkpoint. The problem is that it invokes expensive functionality like base64 encoding, which becomes really inefficient as the size of a tx set grows. To mitigate this, we should only save new tx sets.

marta-lokhova:
Resolves #3499

### DIFF
--- a/src/database/Database.cpp
+++ b/src/database/Database.cpp
@@ -64,9 +64,9 @@ bool Database::gDriversRegistered = false;
 // smallest schema version supported
 static unsigned long const MIN_SCHEMA_VERSION = 13;
 #ifdef ENABLE_NEXT_PROTOCOL_VERSION_UNSAFE_FOR_PRODUCTION
-static unsigned long const SCHEMA_VERSION = 19;
+static unsigned long const SCHEMA_VERSION = 20;
 #else
-static unsigned long const SCHEMA_VERSION = 18;
+static unsigned long const SCHEMA_VERSION = 19;
 #endif
 
 // These should always match our compiled version precisely, since we are
@@ -233,8 +233,11 @@ Database::applySchemaUpgrade(unsigned long vers)
         createTxSetHistoryTable(*this);
         mApp.getPersistentState().upgradeSCPDataFormat();
         break;
-#ifdef ENABLE_NEXT_PROTOCOL_VERSION_UNSAFE_FOR_PRODUCTION
     case 19:
+        mApp.getPersistentState().upgradeSCPDataV1Format();
+        break;
+#ifdef ENABLE_NEXT_PROTOCOL_VERSION_UNSAFE_FOR_PRODUCTION
+    case 20:
         mApp.getPersistentState().setRebuildForType(CONFIG_SETTING);
         mApp.getPersistentState().setRebuildForType(CONTRACT_DATA);
         break;
--- a/src/herder/HerderImpl.cpp
+++ b/src/herder/HerderImpl.cpp
@@ -85,6 +85,7 @@ HerderImpl::HerderImpl(Application& app)
     , mLastExternalize(app.getClock().now())
     , mTriggerTimer(app)
     , mOutOfSyncTimer(app)
+    , mTxSetGarbageCollectTimer(app)
     , mApp(app)
     , mLedgerManager(app.getLedgerManager())
     , mSCPMetrics(app)
@@ -248,6 +249,7 @@ HerderImpl::shutdown()
         mLastQuorumMapIntersectionState.mInterruptFlag = true;
     }
     mTransactionQueue.shutdown();
+    mTxSetGarbageCollectTimer.cancel();
 }
 
 void
@@ -1589,8 +1591,9 @@ HerderImpl::persistSCPState(uint64 slot)
     mLastSlotSaved = slot;
     // saves SCP messages and related data (transaction sets, quorum sets)
     PersistedSCPState scpState;
+    scpState.v(1);
 
-    auto& latestEnvs = scpState.v0().scpEnvelopes;
+    auto& latestEnvs = scpState.v1().scpEnvelopes;
     std::map<Hash, TxSetFrameConstPtr> txSets;
     std::map<Hash, SCPQuorumSetPtr> quorumSets;
 
@@ -1602,7 +1605,7 @@ HerderImpl::persistSCPState(uint64 slot)
         for (auto const& h : getTxSetHashes(e))
         {
             auto txSet = mPendingEnvelopes.getTxSet(h);
-            if (txSet)
+            if (txSet && !mApp.getPersistentState().hasTxSet(h))
             {
                 txSets.insert(std::make_pair(h, txSet));
             }
@@ -1615,7 +1618,7 @@ HerderImpl::persistSCPState(uint64 slot)
         }
     }
 
-    auto& latestQSets = scpState.v0().quorumSets;
+    auto& latestQSets = scpState.v1().quorumSets;
     for (auto it : quorumSets)
     {
         latestQSets.emplace_back(*it.second);
@@ -1623,25 +1626,29 @@ HerderImpl::persistSCPState(uint64 slot)
 
     stellar::Value latestSCPData;
 
-    auto& latestTxSets = scpState.v0().txSets;
+    std::unordered_map<Hash, std::string> txSetsToPersist;
     for (auto it : txSets)
     {
-        latestTxSets.emplace_back();
+        StoredTransactionSet tempTxSet;
         if (it.second->isGeneralizedTxSet())
         {
-            latestTxSets.back().v(1);
-            it.second->toXDR(latestTxSets.back().generalizedTxSet());
+            tempTxSet.v(1);
+            it.second->toXDR(tempTxSet.generalizedTxSet());
         }
         else
         {
-            it.second->toXDR(latestTxSets.back().txSet());
+            it.second->toXDR(tempTxSet.txSet());
         }
+        txSetsToPersist.emplace(
+            it.first, decoder::encode_b64(xdr::xdr_to_opaque(tempTxSet)));
     }
+
     latestSCPData = xdr::xdr_to_opaque(scpState);
 
     std::string encodedScpState = decoder::encode_b64(latestSCPData);
 
-    mApp.getPersistentState().setSCPStateForSlot(slot, encodedScpState);
+    mApp.getPersistentState().setSCPStateV1ForSlot(slot, encodedScpState,
+                                                   txSetsToPersist);
 }
 
 void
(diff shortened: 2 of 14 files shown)


marta-lokhova:
note: I realized some functionality is currently missing from the PR, so please don't review it yet

marta-lokhova:
This is up-to-date now. Note that protocol-next doesn't build at the moment because we need to update XDR upsteam. I'll do it after a round of review though, just in case XDR needs to change.

src/main/PersistentState.cpp
```diff
@@ -20,18 +21,52 @@ using namespace std;
 std::string PersistentState::mapping[kLastEntry] = {
     "lastclosedledger", "historyarchivestate", "lastscpdata",
     "databaseschema",   "networkpassphrase",   "ledgerupgrades",
-    "rebuildledger",    "lastscpdataxdr"};
+    "rebuildledger",    "lastscpdataxdr",      "txset"};
 
 std::string PersistentState::kSQLCreateStatement =
     "CREATE TABLE IF NOT EXISTS storestate ("
     "statename   CHARACTER(32) PRIMARY KEY,"
```
marta-lokhova, line 28: Interesting, it looks like sqlite (which is what we use for tests) doesn't actually enforce this limit, but it's now too small because we reference tx sets by hash. I got a failure on Postgres though, so am currently fixing this.

src/main/PersistentState.cpp
```diff
@@ -120,6 +207,64 @@ PersistentState::setRebuildForType(LedgerEntryType let)
...
+            newScpState.v(1);
+            newScpState.v1().scpEnvelopes = scpState.v0().scpEnvelopes;
+            newScpState.v1().quorumSets = scpState.v0().quorumSets;
+
+            std::unordered_map<Hash, std::string> txSets;
+            for (auto const& txSet : scpState.v0().txSets)
+            {
+                auto txSetPtr = TxSetFrame::getTxSetFramePtr(txSet, mApp);
+                txSets.emplace(
+                    std::pair(txSetPtr->getContentsHash(),
```
dmkozh, line 244: nit: there is no need to use the pair constructor when you're using `emplace`

src/main/ApplicationImpl.cpp
```diff
@@ -308,6 +308,9 @@ ApplicationImpl::initialize(bool createNewDB, bool forceRebuild)
         upgradeToCurrentSchemaAndMaybeRebuildLedger(true, forceRebuild);
     }
 
+    // Subtle: restore persistent state after a potential DB schema upgrade
+    mPersistentState->restoreState();
```
dmkozh, line 312: Just to double-check: this should be safe for restarts without upgrades, right?

src/herder/HerderImpl.cpp
```diff
@@ -1615,42 +1616,76 @@ HerderImpl::persistSCPState(uint64 slot)
...
-            it.second->toXDR(latestTxSets.back().generalizedTxSet());
+            tempTxSet.v(1);
+            it.second->toXDR(tempTxSet.generalizedTxSet());
         }
         else
         {
-            it.second->toXDR(latestTxSets.back().txSet());
+            it.second->toXDR(tempTxSet.txSet());
         }
+        txSetsToPersist.emplace(std::pair(
```
dmkozh, line 1640: nit: no need to use pair constructor with `emplace`

src/herder/test/HerderTests.cpp
```diff
@@ -2164,6 +2166,45 @@ TEST_CASE("SCP State", "[herder][acceptance]")
...
+            std::unordered_set<Hash> hashes;
+            for (auto i = app->getHerder().trackingConsensusLedgerIndex();
+                 i >= limit; --i)
+            {
+                if (i == LedgerManager::GENESIS_LEDGER_SEQ)
+                {
+                    continue;
+                }
+                auto msgs = herder.getSCP().getLatestMessagesSend(i);
+                if (expectedSCPState.has_value())
```
dmkozh, line 2189: nit: `if (expectedSCPState)` is sufficient, though I suppose you could use the explicit check for expressivity, so up to you

src/main/ApplicationImpl.cpp
marta-lokhova: Yeah, when a node upgrades its version to include this change, it'll do a schema upgrade, which will replace all of the old-style rows with new ones. If there is no upgrade, it means the schema is already up-to-date, which is what `restoreState` expects (as it queries PersistentStateV1)

src/herder/TxSetFrame.h
```diff
@@ -57,6 +58,9 @@ class TxSetFrame : public NonMovableOrCopyable
     makeFromWire(Hash const& networkID,
                  GeneralizedTransactionSet const& xdrTxSet);
 
+    static TxSetFrameConstPtr
+    getTxSetFramePtr(StoredTransactionSet const& storedSet, Application& app);
```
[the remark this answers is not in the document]
marta-lokhova: good idea, done.

src/main/PersistentState.cpp
marta-lokhova: True, just a habit of writing things down explicitly. Updated (in the other place too)

src/herder/test/HerderTests.cpp
marta-lokhova: kept it as-is if you don't mind: has_value is used pretty commonly in the codebase, and I do like the explicitness of it

src/main/PersistentState.cpp
```diff
@@ -4,32 +4,111 @@
...
+    }
+
+    startTxSetGCTimer();
+}
+
+void
+PersistentState::upgradeSizeLimit(Database& db)
+{
+    db.getSession()
+        << "ALTER TABLE storestate ALTER COLUMN statename TYPE CHARACTER(70);";
```
MonsieurNicolas, line 111: why 70? Is it just max string size `PersistentState::mapping` + 32 (to store hash) + 1 (for extra characters) ? It seems quite close to what we do (this max size does not really do anything, so may break again in the future)

MonsieurNicolas (changes requested):

src/database/Database.cpp
```diff
@@ -233,8 +233,16 @@ Database::applySchemaUpgrade(unsigned long vers)
         createTxSetHistoryTable(*this);
         mApp.getPersistentState().upgradeSCPDataFormat();
         break;
-#ifdef ENABLE_NEXT_PROTOCOL_VERSION_UNSAFE_FOR_PRODUCTION
     case 19:
+        // Sqlite does not enforce size limits
+        if (!isSqlite())
+        {
+            PersistentState::upgradeSizeLimit(mApp.getDatabase());
```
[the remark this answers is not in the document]
marta-lokhova: ALTER COLUMN doesn't exist in sqlite. Instead, you have to create a new table and copy its contents from the old table. It doesn't really make sense to do this, given that this upgrade is a no-op on sqlite.

src/main/PersistentState.cpp
```diff
@@ -4,32 +4,111 @@
...
     "); ";
 
-PersistentState::PersistentState(Application& app) : mApp(app)
+PersistentState::PersistentState(Application& app)
+    : mApp(app), mTxSetGarbageCollectTimer(app)
+{
+}
+
+void
+PersistentState::startTxSetGCTimer()
```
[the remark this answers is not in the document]
marta-lokhova: I'm not a fan of the suggested approaches, because they both make it harder to reason about what's happening:
* we don't know how many times SCP persistence is called per ledger, so we can't really reason about how often tx sets are cleaned up (same thing with the number of tx sets).
* It makes testing harder as now we have to into account cases where purging triggers at different times depending on consensus.
* I went with the timer, as it seems to be the simplest approach that avoids anything complicated where we have to map Herder's state to the database and try to estimate a good time to do the cleanup.

I suggest we keep the timer approach. I do like the idea of moving it to Herder, this way we don't need extra start/shutdown logic in PersistentState, and it'll be easy to refactor the purging function based on your other suggestion.

src/herder/HerderImpl.cpp
```diff
@@ -1615,42 +1616,76 @@ HerderImpl::persistSCPState(uint64 slot)
...
 HerderImpl::restoreSCPState()
 {
     ZoneScoped;
 
+    // First, load all known tx sets
+    auto latestTxSets = mApp.getPersistentState().getTxSetsForAllSlots();
+    for (auto const& txSet : latestTxSets)
+    {
+        std::vector<uint8_t> buffer;
+        decoder::decode_b64(txSet, buffer);
```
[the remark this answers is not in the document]
marta-lokhova: Fixed.

src/main/PersistentState.cpp
```diff
@@ -4,32 +4,111 @@
...
+    {
+        std::vector<uint8_t> buffer;
+        decoder::decode_b64(state, buffer);
+        PersistedSCPState scpState;
+        try
+        {
+            xdr::xdr_from_opaque(buffer, scpState);
+            for (auto const& e : scpState.v1().scpEnvelopes)
+            {
+                for (auto const& hash : getTxSetHashes(e))
```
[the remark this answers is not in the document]
marta-lokhova: Moved it to Herder.

src/database/Database.cpp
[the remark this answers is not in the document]
MonsieurNicolas: oh I see, so a sad sqlite limitation.
So maybe update the comment to something a little more specific: "sqlite does not enforce size constraints and does not support updating columns".

Also, you should just move this to `upgradeSCPDataV1Format`: there is no reason to split the schema upgrade into 2 parts.

src/main/PersistentState.cpp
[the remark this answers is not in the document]
MonsieurNicolas: actually you're right, I was thinking that we could get arbitrary number of values from other validators, so we should try to GC more often in some situations, but this is only tracking what the current validator is doing.

src/herder/HerderImpl.cpp
```diff
@@ -1615,42 +1616,76 @@ HerderImpl::persistSCPState(uint64 slot)
...
+    mApp.getPersistentState().setSCPStateV1ForSlot(slot, encodedScpState,
+                                                   txSetsToPersist);
 }
 
 void
 HerderImpl::restoreSCPState()
 {
     ZoneScoped;
 
+    // First, load all known tx sets
```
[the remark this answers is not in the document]
marta-lokhova: I had the same thought, but ultimately decided to keep them separate: the functions are sufficiently different, so refactoring didn't actually simplify much.

MonsieurNicolas:
r+ f13ddfe394f69687bd6a5ed7bc09539acaf9e7b1

lokera666 (commented):
Yeah

PR #3504, opened 2022-08-12
Do not force broadcast SCP messages

MonsieurNicolas:
As we're pulling for SCP messages when out of sync, this is not necessary anymore.

This change became possible after the network picked up #2786 - we just forgot to do this.

This should reduce the number of duplicate SCP messages broadcasted quite a bit.


marta-lokhova:
r+ 6f6832d4f0cde774d40adc7862c981191b7a2243

PR #3505, opened 2022-08-16
Fixes to verify-checkpoints command

marta-lokhova:
Looks like `verify-checkpoints` was broken since https://github.com/stellar/stellar-core/pull/3229/files#diff-1fe071ca6fa86c1b4830199b49f2cd39e811459529da0682484f4c0fbc8486bfL891. This change adds the buffering logic back, along with a new test (Note that the test focuses on the waiting condition only; the writing logic itself is already tested elsewhere)


MonsieurNicolas:
r+ e721111e22d5630d51fd6c460a6edd782c80978e

PR #3508, opened 2022-08-17
Unify fee-based transaction comparisions

dmkozh:
# Description

This introduces a common data structure (`SurgePricingPriorityQueue`) that orders the transactions by the fee rate and is aware of operation limits. This data structure is then used in the 3 places that do fee-based transaction comparisons: `TransactionQueue`, `TxQueueLimiter` and `TxSetFrame`.

The refactoring is functionally a no-op and serves as a preparation for introducing more complex transaction comparison logic.

# Checklist
- [ ] Compiles
- [ ] Ran all tests

### DIFF
--- a/src/herder/SurgePricingUtils.cpp
+++ b/src/herder/SurgePricingUtils.cpp
@@ -3,20 +3,30 @@
 // of this distribution or at http://www.apache.org/licenses/LICENSE-2.0
 
 #include "herder/SurgePricingUtils.h"
+#include "crypto/SecretKey.h"
+#include "util/Logging.h"
 #include "util/numeric128.h"
+#include "util/types.h"
+#include <numeric>
 
 namespace stellar
 {
-
-using namespace std;
+namespace
+{
 
 int
-feeRate3WayCompare(int64 lFeeBid, uint32 lNbOps, int64 rFeeBid, uint32 rNbOps)
+feeRate3WayCompare(TransactionFrameBase const& l, TransactionFrameBase const& r)
 {
-    // compare fee/numOps between l and r
-    // getNumOperations >= 1 because SurgeCompare can only be used on
-    // valid transactions
-    //
+    return stellar::feeRate3WayCompare(l.getFeeBid(), l.getNumOperations(),
+                                       r.getFeeBid(), r.getNumOperations());
+}
+
+} // namespace
+
+int
+feeRate3WayCompare(int64_t lFeeBid, uint32_t lNbOps, int64_t rFeeBid,
+                   uint32_t rNbOps)
+{
     // Let f1, f2 be the two fee bids, and let n1, n2 be the two
     // operation counts. We want to calculate the boolean comparison
     // "f1 / n1 < f2 / n2" but, since these are uint128s, we want to
@@ -27,7 +37,6 @@ feeRate3WayCompare(int64 lFeeBid, uint32 lNbOps, int64 rFeeBid, uint32 rNbOps)
     //               f1 / n1 < f2 / n2
     //  == f1 * n1 * n2 / n1 < f2 * n1 * n2 / n2
     //  == f1 *      n2      < f2 * n1
-
     auto v1 = bigMultiply(lFeeBid, rNbOps);
     auto v2 = bigMultiply(rFeeBid, lNbOps);
     if (v1 < v2)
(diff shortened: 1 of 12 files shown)


src/herder/SurgePricingUtils.cpp
```diff
@@ -41,10 +50,315 @@ feeRate3WayCompare(int64 lFeeBid, uint32 lNbOps, int64 rFeeBid, uint32 rNbOps)
...
+    releaseAssert(neededTotalOps > 0);
+    while (neededTotalOps > 0)
+    {
+        // The preconditions are ensured above that we should be able to fit
+        // the transaction by evicting some transactions (given high enough
+        // fee).
+        releaseAssert(iter != mTxStackSet.end());
+        auto const& evictTxStack = *iter;
+        auto const& evictTx = *evictTxStack->getTopTx();
+        auto evictOps = evictTxStack->getNumOperations();
```
MonsieurNicolas, line 311: I think this method (`canFitWithEviction`) should not be in this class, or should be written differently:
what that code does is use the top tx `evictTx` to decide on evicting the entire stack (and the code in , which does not make sense in the general case, and only in the context of `TxQueueLimiter` that right now uses `SingleTxStack`.
If we want to keep it in here, the challenge (I think) is that you can't really iterate over txs in stacks without popping.

src/herder/TxQueueLimiter.cpp
```diff
@@ -10,244 +10,177 @@
...
-// that sorts by base fee, and breaks ties using pointers
-struct QueueLimiterTxComparator
+namespace
+{
+class SingleTxStack : public TxStack
 {
-    size_t mSeed;
-    QueueLimiterTxComparator(size_t seed) : mSeed(seed)
+  public:
+    SingleTxStack(TransactionFrameBasePtr tx) : mTx(tx)
```
MonsieurNicolas, line 18: why not assert in the constructor instead of methods?

src/herder/TxQueueLimiter.cpp
```diff
@@ -10,244 +10,177 @@
...
-        if (size() == 0)
-        {
-            return false;
-        }
-        auto evictTx = getWorstTransaction();
-        mMinFeeNeeded = {evictTx->getFeeBid(), evictTx->getNumOperations()};
-        evict(evictTx);
+        auto tx = evictedStack->getTopTx();
+        mEvictedFeeBid = {tx->getFeeBid(), tx->getNumOperations()};
+        evict(tx);
```
MonsieurNicolas, line 155: I made this comment in the other PR:
`evict` may evict more transactions (because when kicking out a tx it also kicks out transactions with higher sequence numbers).
So here you're changing semantics (this probably points to a gap in `TransactionQueue` test code).

As `txsToEvict` contains transactions to kick out in ascending order, you should stop as soon as the goal is reached?

src/herder/TxQueueLimiter.cpp
```diff
@@ -10,244 +10,177 @@
...
+void
+TxQueueLimiter::resetEvictionState()
 {
-    return mMinFeeNeeded;
+    // We cannot normally initialize `mTxs` queue in the constructor
+    // because `maxQueueSizeOps()` may not be initialized. Hence we initialize
+    // lazily during the add/reset.
+    if (mTxs == nullptr)
+    {
+        resetTxs();
```
MonsieurNicolas, line 174: Why do you need this? The only path where this applies is when `resetEvictionState` is called from `TransactionQueue::shift()` when `TxQueueLimiter` is empty it seems.

src/herder/TxSetFrame.cpp
```diff
@@ -797,57 +779,35 @@ TxSetFrame::addTxsFromXdr(Hash const& networkID,
...
-        else
-        {
-            // drop this transaction -> we need to drop the others
-            cur->clear();
-        }
+        auto perOpFee = computePerOpFee(*tx, lclHeader.ledgerVersion);
+        lowestFee = std::min(lowestFee, perOpFee);
     }
-    mTxs = TxSetUtils::sortTxsInHashOrder(filteredTxs);
+    mTxs = TxSetUtils::sortTxsInHashOrder(includedTxs);
```
MonsieurNicolas, line 808: this conflicts with #3496 (so you'll have to rebase as we merge)

src/herder/TransactionQueue.cpp
```diff
@@ -1061,73 +1083,59 @@ TransactionQueue::broadcastSome()
...
+    };
+    SurgePricingPriorityQueue::visitTopTxs(trackersToBroadcast, opsToFlood,
+                                           mBroadcastSeed, visitor,
+                                           mBroadcastOpCarryover);
     ban(banningTxs);
     // carry over remainder, up to MAX_OPS_PER_TX ops
     // reason is that if we add 1 next round, we can flood a "worst case fee
     // bump" tx
-    mBroadcastOpCarryover = std::min<size_t>(opsToFlood, MAX_OPS_PER_TX + 1);
+    mBroadcastOpCarryover = std::min(mBroadcastOpCarryover, MAX_OPS_PER_TX + 1);
```
MonsieurNicolas, line 1138: changes in this file may conflict with the pull mode PR (I just checked and changes look to be in different places, so it will depend on how good auto-merge is)

src/herder/test/TransactionQueueTests.cpp
```diff
@@ -1312,74 +1319,87 @@ TEST_CASE("TransactionQueue limits", "[herder][transactionqueue]")
...
-            auto fr = limiter.getMinFeeNeeded();
-            return fr.second == 0 ? 0ll
-                                  : bigDivideOrThrow(fr.first, 1, fr.second,
-                                                     Rounding::ROUND_UP);
+
+        // Check that 1 operation transaction with `minFee` cannot be added to
+        // the limiter, but with `minFee + 1` can be added. Use for checking
+        // that fee threshold is applied even when there is enough space in
+        // the limiter, but some transactions were evicted before.
+        auto checkMinFeeToFitWithNoEvict = [&](int64_t minFee) {
```
MonsieurNicolas, line 1333: nice upgrade thank you!

MonsieurNicolas (changes requested):
Overall looks good - only thing I flagged is wrt `canFitWithEviction` that makes too many assumptions

src/herder/TxQueueLimiter.cpp
dmkozh: TBH I don't remember why exactly did I add this, doesn't seem like removing this breaks anything.

src/herder/TxSetFrame.cpp
dmkozh: Thanks for flagging this, I'll rebase after resolving this round of comments (for the sake of cleaner diffs)

src/herder/SurgePricingUtils.cpp
dmkozh: > what that code does is use the top tx evictTx to decide on evicting the entire stack 
Well, that's the invariant of this method and it happens to work for `TxQueueLimiter`. I can't say it's incorrect for seq num based stack implementations, as the same invariant would be maintained (whether it's useful or not is up to the caller; currently I guess we know we don't need to reuse it).

I cannot say I'm 100% happy with this class, however I think it's a decent option. The set of supported operations is indeed bespoke and isn't too abstract. This is due to the fact that we have quite specific use cases and there is only so much in common between them. In order to support the upcoming functionality I wanted to encapsulate as much logic as possible in these methods and let as little internals to be exposed as possible. The implementation with lanes is more involved in terms of using the internal structures. 

I guess we could make the queue to be 'low-level' data structure that exposes the internals and then a wrapper data structure to be used in `TxQueueLimiter` with `SingleTxStack` only. However, I'm not sure how much value does that provide, given that the complexity still lies in the iteration methods. I'd also still rather keep the logic in `SurgePricingUtils` so that there is a single source for all surge-pricing related stuff.

MonsieurNicolas: We talked offline about potentially getting rid of the "transaction chains" per account in the future, so I think it's fine to keep the code as is. As right now we have no use cases for this method outside of the `SingleTxStack`, and to avoid surprises if we were to use it in a different context, I think we should just add an assert that there is only one tx in the stack:
```
        auto const& evictTx = *evictTxStack->getTopTx();
        auto evictOps = evictTxStack->getNumOperations();
        releaseAssert(evictOps == evictTx.getNumOperations());
```

what do you think?

dmkozh: SG, added the assertion.

src/herder/TxSetFrame.cpp
dmkozh: Rebased/merged this.

src/herder/TransactionQueue.cpp
dmkozh: The queue has been merged without conflicts.

dmkozh:
> Left one suggestion in the only open comment left. We can rebase & squash it as it's pretty much ready to merge

I've rebased & squashed, just need to wait for the checks to pass.

MonsieurNicolas:
r+ 9d3ce58090296a5b07dee0391c4d6e529bb85a26

PR #3509, opened 2022-08-17
Allow passing trusted hash to catchup, warn if catchup is unverified

marta-lokhova:


MonsieurNicolas:
r+ 57e17a57d657bd91c379e441ffba495b07e3f848

Issue #3487, opened 2022-07-18
add event/log recording and reporting machinery for contracts


graydon:
- [x] Make event-log buffer in host
- [x] Make host functions and host infrastructure for logging diagnostic / error messages
- [x] Define general / non-diagnostic event type in XDR-next
- [x] Emit relevant host events into event-log buffer
- [x] Route events back to transaction-invoker / preflight-invoker
- [x] Emit events into the meta stream

graydon:
This is done enough for now. There will be ongoing work on the events-and-logging subsystem but there's one there now.

graydon:
Reopening with more clarity on partial progress.

graydon:
Progress on the host side as of https://github.com/stellar/rs-soroban-env/pull/342 -- just need to route it back to the invoker in the core endpoint / emit into the meta stream now.

smac711:
@graydon Should the iteration and priority of this issue be updated given it is in the past and is it still a critical issue to close prior to Meridian? 

1.  iteration = ? 
2. priority = ?

graydon:
stellar/stellar-core#3516 is now complete enough to review. There are a few rough edges in the design (CAP-0056 isn't entirely complete) but it should work.

PR #3516, opened 2022-08-24
Plumb events through to TxMeta / LedgerCloseMeta

graydon:
This plumbs ContractEvents through to the preflight endpoint, necessary for https://github.com/stellar/stellar-core/issues/3487

It also plumbs the events through to the LedgerCloseMeta v2 / TransactionMeta v3, as described in CAP-0056.

It also (hopefully correctly) accumulates and hashes the new TransactionResultSetV2 into the ledger header, while still (for now) continuing to publish the old TransactionResultSet values to the database and thus history archives. This will (currently) break offline catchup, which does a download-and-verify job on `results` files in the archives (which, being defined to contain TransactionResultSet, will still load ok but will not verify against the ledger hash), and will also break archivist and likely anything else that tires to check the identity between `results` and ledger headers. We should decide how to proceed here before finalizing the design.


src/ledger/LedgerManagerImpl.cpp
```diff
@@ -1214,29 +1219,24 @@ LedgerManagerImpl::applyTransactions(
...
 
         // Then potentially add that TRP and its associated TransactionMeta
         // into the associated slot of any LedgerCloseMeta we're collecting.
         if (ledgerCloseMeta)
         {
-            TransactionResultMeta& trm =
-                ledgerCloseMeta->txProcessing().at(index);
-            trm.txApplyProcessing = tm;
-            trm.result = results;
+            ledgerCloseMeta->setLastTxProcessingMetaAndResultPair(tm.getXDR(),
```
MonsieurNicolas, line 1238: this breaks meta:
you replaced an update at index `index` by an update of the last element

MonsieurNicolas (changes requested):
I am pushing a possible fix. We should add a test that does something like `closeLedger(*app, {tx1, tx2, tx3}, true);` and checks (maybe using the "debug meta" functionality) that we produce the meta that we expect?

MonsieurNicolas:
I pushed a couple commits to that branch:
* [fix: properly populate result in LedgerCloseMeta](https://github.com/stellar/stellar-core/pull/3516/commits/cfd69e77017a2e6463f37a9f49695fd7e881cfee) - properly sets the tx result and hashes when there are more than 1 tx in a ledger (note that the meta was broken for all versions)
* [fix: properly set result and corresponding hashes in tx meta](https://github.com/stellar/stellar-core/pull/3516/commits/c4dee576f0bf1fa000750f44756ad05929fd3189) - moves the responsibility of setting the result/hashes in meta outside of `TransactionFrame`, this makes `FeeBumpTransaction` work as expected for the new meta format.

We should add some tests for this (I suggested in a comment to maybe leverage the "debug stream" in some way: it should be fairly easy to add a few checks against the ledger close meta generated as part of the "meta stream test suite", a ledger containing a fee bump, a regular transaction and maybe a tx that produces events).

src/transactions/TransactionMetaFrame.cpp
```diff
@@ -0,0 +1,284 @@
...
+TransactionMetaFrame::TransactionMetaFrame(uint32_t protocolVersion)
+{
+    // The TransactionMeta v() switch can be in 4 positions 0, 1, 2, 3. We
+    // do not support 0 or 1 at all -- core does not produce it anymore and we
+    // have no obligation to consume it under any circumstance -- so this
+    // class just switches between cases 2 and 3.
+    mVersion = 2;
+#ifdef ENABLE_NEXT_PROTOCOL_VERSION_UNSAFE_FOR_PRODUCTION
+    if (protocolVersionStartsFrom(protocolVersion, SOROBAN_PROTOCOL_VERSION))
+    {
```
MonsieurNicolas, line 23: you forgot to update `normalizeMeta` in dumpxdr.cpp

MonsieurNicolas: mmm actually for hashes to work properly we have to normalize the meta in protocol (so here)

MonsieurNicolas: (and I'm not sure this can be done more efficiently that what was done in dumpxdr)

src/ledger/LedgerManagerImpl.cpp
graydon: `index` always points to the last element pushed onto the meta (it starts at 0 and is ++'ed by one on each iteration of the loop, as we process transactions). There's no difference.

MonsieurNicolas: when we charge fees earlier, we insert elements, so the last element is definitely not the same one

src/ledger/LedgerManagerImpl.cpp
```diff
@@ -1252,8 +1252,11 @@ LedgerManagerImpl::applyTransactions(
         if (mApp.getConfig().MODE_STORES_HISTORY_MISC)
         {
             auto ledgerSeq = ltx.loadHeader().current().ledgerSeq;
-            storeTransaction(mApp.getDatabase(), ledgerSeq, tx, tm,
-                             txResultSet);
+            // TODO: decide if we are going to keep storing the V1 txResultSet
```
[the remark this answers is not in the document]
graydon: Done: https://github.com/stellar/stellar-core/issues/3555

src/ledger/TransactionResultSetFrame.cpp
```diff
@@ -0,0 +1,86 @@
...
+#include "util/ProtocolVersion.h"
+
+namespace stellar
+{
+
+TransactionResultSetFrame::TransactionResultSetFrame(uint32_t protocolVersion)
+{
+    mVersion = 1;
+#ifdef ENABLE_NEXT_PROTOCOL_VERSION_UNSAFE_FOR_PRODUCTION
+    // TODO: for the time being, we are _not_ hashing or emitting
```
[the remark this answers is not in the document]
graydon: this is really .. the same issue -- https://github.com/stellar/stellar-core/issues/3555 -- so I've modified it to describe both and am linking to it in both

src/ledger/LedgerManagerImpl.cpp
graydon: Ah, I see, the fee-processing loop is a first pass over all the txs -- that extends the result vector -- and then the tx-applying loop does a second pass over all of those vector elements. Ok.

src/transactions/TransactionFrame.cpp
```diff
@@ -1085,9 +1097,11 @@ TransactionFrame::applyOperations(SignatureChecker& signatureChecker,
 
             ltxTx.commit();
             // commit -> propagate the meta to the outer scope
-            std::swap(outerMeta.v2().operations, newMeta.v2().operations);
-            std::swap(outerMeta.v2().txChangesAfter,
-                      newMeta.v2().txChangesAfter);
+            outerMeta.pushOperationMetas(operationMetas);
+            outerMeta.pushTxChangesAfter(changesAfter);
+#ifdef ENABLE_NEXT_PROTOCOL_VERSION_UNSAFE_FOR_PRODUCTION
+            outerMeta.pushContractEvents(mEvents);
```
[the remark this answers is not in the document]
graydon: fixed

src/transactions/TransactionFrame.cpp
```diff
@@ -1085,9 +1097,11 @@ TransactionFrame::applyOperations(SignatureChecker& signatureChecker,
 
             ltxTx.commit();
             // commit -> propagate the meta to the outer scope
-            std::swap(outerMeta.v2().operations, newMeta.v2().operations);
-            std::swap(outerMeta.v2().txChangesAfter,
-                      newMeta.v2().txChangesAfter);
+            outerMeta.pushOperationMetas(operationMetas);
```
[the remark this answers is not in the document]
graydon: fixed (at least some -- there's probably always going to be more possible)

src/ledger/LedgerManagerImpl.cpp
graydon: (I believe this conversation is resolved by your added commits below)

src/transactions/TransactionFrame.cpp
```diff
@@ -1187,18 +1201,22 @@ TransactionFrame::apply(Application& app, AbstractLedgerTxn& ltx,
...
         {
             // This should only throw if the logging during exception handling
             // for applyOperations throws. In that case, we may not have the
             // correct TransactionResult so we must crash.
-            return valid && applyOperations(signatureChecker, app, ltx, meta);
+            if (ok)
+            {
+                ok = applyOperations(signatureChecker, app, ltx, meta);
+            }
+            meta.setTxResult(mResult);
```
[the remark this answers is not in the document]
graydon: (I believe this conversation is resolved by your added commits below)

src/transactions/TransactionMetaFrame.cpp
graydon: fixed

src/transactions/FeeBumpTransactionFrame.cpp
```diff
@@ -83,15 +84,13 @@ updateResult(TransactionResult& outerRes, TransactionFrameBasePtr innerTx)
...
+                               TransactionMetaFrame& meta)
 {
     try
     {
         LedgerTxn ltxTx(ltx);
         removeOneTimeSignerKeyFromFeeSource(ltxTx);
-
-        auto& txChanges = meta.v2().txChangesBefore;
-        txChanges = ltxTx.getChanges();
+        meta.pushTxChangesBefore(ltxTx.getChanges());
```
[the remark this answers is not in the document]
graydon: I'm not sure I understand this comment. This is the part that sets the `txChangesBefore` part of the meta, which is only (as far as I can tell) used to record the pre-operation activities of a fee-bump transaction. And all I've done here is to change the way it gets recorded from direct field access to indirect access through the frame class.

Results don't figure into it. Maybe I'm misunderstanding your comment?

src/transactions/TransactionMetaFrame.cpp
```diff
@@ -0,0 +1,284 @@
+// Copyright 2022 Stellar Development Foundation and contributors. Licensed
```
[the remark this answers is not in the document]
graydon: fixed

graydon:
@MonsieurNicolas I've added a very basic test that does as suggested -- throws some txs at closeLedger and checks the meta is vaguely sensible -- but I'm out of time tonight; will extend it tomorrow, as well as wrap up whatever you mean above in the comment a bout fee bumps, and regenerate the meta test vectors again.

src/transactions/FeeBumpTransactionFrame.cpp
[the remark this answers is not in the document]
MonsieurNicolas: (Github does not allow me to leave comments outside of diffs) this was a comment that result in the meta was not updated in `updateResult` but this is irrelevant now that I moved the meta update related to setting the result out of the `TransactionFrame` class

src/transactions/TransactionMetaFrame.cpp
```diff
@@ -201,6 +202,7 @@ void
 TransactionMetaFrame::finalizeHashes()
 {
     releaseAssert(!mHashesFinalized);
+    normalizeMeta(mTransactionMeta);
```
MonsieurNicolas, line 205: we probably don't want to do this if we're not on v=3 (as this has some unknown perf impact)

src/util/MetaUtils.cpp
```diff
@@ -0,0 +1,101 @@
...
+    case 1:
+        sortChanges(m.v1().txChanges);
+        normalizeOps(m.v1().operations);
+        break;
+    case 2:
+        sortChanges(m.v2().txChangesBefore);
+        sortChanges(m.v2().txChangesAfter);
+        normalizeOps(m.v2().operations);
+        break;
+    }
```
MonsieurNicolas, line 98: shouldn't there be a `case 3` here?

src/ledger/test/LedgerCloseMetaStreamTests.cpp
```diff
@@ -427,3 +431,68 @@ TEST_CASE("METADATA_DEBUG_LEDGERS works", "[metadebug]")
...
+        auto bal = app->getLedgerManager().getLastMinBalance(2);
+
+        auto root = TestAccount::createRoot(*app);
+
+        // Ledger 2 creates the account, which we do directly so emit no meta
+        // for.
+        auto acc1 = root.create("acc1", bal);
+
+        // Ledger 3 uses closeLedger so emits interesting meta.
+        std::vector<TransactionFrameBasePtr> txs = {
```
MonsieurNicolas, line 459: I would add a couple other transactions in there (as the bug we had was that meta was broken with ledgers with more than 1 tx).
For good measure one of these could be a fee bump transaction

src/ledger/test/LedgerCloseMetaStreamTests.cpp
```diff
@@ -427,3 +431,68 @@ TEST_CASE("METADATA_DEBUG_LEDGERS works", "[metadebug]")
...
+    LedgerCloseMeta lcm;
+    uint32_t maxSeq = 0;
+    while (in.readOne(lcm))
+    {
+        if (lcm.v() == 2)
+        {
+            maxSeq = lcm.v2().ledgerHeader.header.ledgerSeq;
+            if (lcm.v2().ledgerHeader.header.ledgerSeq == 3)
+            {
+                REQUIRE(lcm.v2().txProcessing.size() == 1);
```
MonsieurNicolas, line 475: why not just convert to json and do a string comparison of the entire record for "ledger 3"?
That way we'll have in the test the actual expected meta (which will allow us to see when we make any meta changes - as this breaks Horizon).

Also note that vcurrent and vnext have a different format (which we'll see in the test)

src/transactions/FeeBumpTransactionFrame.cpp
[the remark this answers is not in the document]
MonsieurNicolas: (you can resolve this thread when you see my response)

src/transactions/TransactionFrame.cpp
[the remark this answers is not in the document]
MonsieurNicolas: yes thank you

MonsieurNicolas:
I resolved comments and added new ones based on the latest updates.
You also need to rebase from what I see.

src/util/MetaUtils.cpp
graydon: Oops! Yes of course, sorry. Also will add a default assert.

src/transactions/TransactionMetaFrame.cpp
graydon: Ok.

src/ledger/test/LedgerCloseMetaStreamTests.cpp
graydon: done

src/ledger/test/LedgerCloseMetaStreamTests.cpp
graydon: ok, doing so (it adds a lot of test vector files, I eyeballed them a bit but second opinions on whether they're correct welcome!)

graydon:
@MonsieurNicolas at last I believe this is rebased, passing all tests, and (I hope!) addressing all your comments. Should be ready to land.

MonsieurNicolas:
r+ d4b28eecef52584de73a39b6ead74a3fb7cd1cd2

PR #3527, opened 2022-09-14
Add install-rust.sh

graydon:
Add a script for installing rust using rustup, with a fixed expected SHA256 for the rustup binary.

### DIFF
--- a/install-rust.sh
+++ b/install-rust.sh
@@ -1,0 +1,48 @@
+#!/bin/sh
+#
+# This file should be run if you want to install the exact version of Rust that
+# stellar-core is building its packages and testing with.
+#
+# You are of course welcome to install your own version of Rust, but this file
+# is version-controlled and run in CI to make the dependency explicit (and to
+# use a version newer than the debian packages).
+
+# Fail on any error
+set -e
+
+# Print each step as we go
+set -x
+
+# Extracts the current release-channel (which in our case is a specific rust
+# version) that we want to install. This value will change relatively often, as
+# new Rust _compilers_ are released, but you should only have to change it in
+# rust-toolchain.toml. This file will pick it up automatically.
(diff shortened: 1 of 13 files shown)


install-rust.sh
```diff
@@ -0,0 +1,47 @@
...
+
+# We download rustup-init from a URL adjacent to the SHA256 file above, and
+# check that it matches the expected SHA256 wired-in to this file, and then run
+# it.
+#
+# Rustup with then install Rust binary components (compiler, stdlib, etc.) and
+# check their PGP signatures match the Rust project's signing key (the signing
+# key is embedded in rustup).
+rm -f rustup-init
+curl -o rustup-init https://static.rust-lang.org/rustup/dist/${HOST_TRIPLE}/rustup-init
```
leighmcculloch, line 44: Probably want to add a `-f` so that a non-2xx/3xx error causes a non-zero exit status.
```suggestion
curl -f -o rustup-init https://static.rust-lang.org/rustup/dist/${HOST_TRIPLE}/rustup-init
```

leighmcculloch (commented):
One suggestion, but looks greater otherwise.

install-rust.sh
```diff
@@ -0,0 +1,48 @@
+#!/bin/sh
```
MonsieurNicolas, line 1: a few more tweaks missing from what I see: both docker/Dockerfile.testing and .devcontainer/Dockerfile need to be adjusted (and probably need to use this script). Also README.md has wrong instructions (should point to use install-rust.sh instead)

graydon:
@MonsieurNicolas fixed issues described (also tweaked the Dockerfile.testing to enable protocol-next)

MonsieurNicolas:
r+ 489ca94fd5bfd24394b9172a3b6cf8ee8dc32acb

PR #3529, opened 2022-09-15
Update ops limit in `TxQueueLimiter` every time we try adding tx.

dmkozh:
# Description

This is how the limiter has worked before the refactoring.

I somehow convinced myself that we reset limiter on upgrades, but that's not true. So for simplicity I just update the limit to make sure we always have up-to-date ops limit.

# Checklist
- [ ] Compiles
- [ ] Ran all tests


MonsieurNicolas:
r+ 8c5f3a0455a8c5844150db9ba769bd3a84aaec0e

Issue #3521, opened 2022-08-30
Improve pulling for txs that have been applied (pull mode)


hidenori-shinohara:
https://github.com/stellar/stellar-core/pull/3479

There is no need to pull transactions that have already been applied. It wastes bandwidth.

Issue #3520, opened 2022-08-26
Remove old flow control reporting


marta-lokhova:
Now that flow control is fully enabled, we don't need reporting such as percentage of nodes with flow control enabled etc. We should audit the codebase and remove any obsolete code.

PR #3530, opened 2022-09-15
pull mode follow-up

marta-lokhova:
Resolves https://github.com/stellar/stellar-core/issues/3521 by banning applied transactions. Note that this broke one test which tries to re-add an applied transaction to the queue again (which now fails). I split the test into two sections that test two different sequences. 

Resolves https://github.com/stellar/stellar-core/issues/3520 by removing flow control percentage metrics plus minor changes to the peers endpoint. Note that we still report capacities as well as outbound queue stats, at these are useful to debugging.


MonsieurNicolas:
r+ 04646b1d481017b7c46e089439bf0d4032c6df9f

PR #3534, opened 2022-09-16
Enforce line endings to be "LF"

MonsieurNicolas:
CRLF is causing some issues on Windows, this PR fixes them by just ensuring that the workspace is always configured with LF.

I also included a couple commits to get rid of warnings and build errors (Windows again).


dmkozh (approved):

MonsieurNicolas:
r+ 56c832087695e5703652271d79e664e454bb56f5

PR #3540, opened 2022-09-19
Drop peers that do not respect pull mode settings

marta-lokhova:
Two changes:
* Drop peers that do not follow the protocol and send adverts/demands when pull mode is disabled
* Place adverts/demands onto the same scheduler queue as transactions, mark them as droppable


MonsieurNicolas:
r+ 97c00e1a65d065caafcd3d9f4502326e8ef03e64

PR #3542, opened 2022-09-21
Ignore adverts and demands when out of sync

marta-lokhova:


src/overlay/Peer.cpp
```diff
@@ -917,7 +919,7 @@ Peer::recvMessage(StellarMessage const& stellarMsg)
         cat = "MISC";
     }
 
-    if (!mApp.getLedgerManager().isSynced() && stellarMsg.type() == TRANSACTION)
+    if (!mApp.getLedgerManager().isSynced() && ignoreIfOutOfSync)
```
hidenori-shinohara, line 922: Looks like this will ignore transactions when we're out of sync (even when pull mode is disabled). Is that intended? It sounds like a good change but I was wondering since that'd change the behavior even when pull mode is disabled!

marta-lokhova: this doesn't change the behavior, see previous condition `!mApp.getLedgerManager().isSynced() && stellarMsg.type() == TRANSACTION`

hidenori-shinohara: Ah! Yes, that makes sense. I totally missed that.

hidenori-shinohara: This looks good to me!

MonsieurNicolas:
r+ dd338706136c838f615dec2fdc963a7323fb6405

PR #3544, opened 2022-09-22
Nomination: do not vote for new values after confirming a candidate

### LINKED ISSUE #3523: Nomination: do not introduce a new value when we have a candidate
We potentially introduce a new value (generated locally) for nomination, which is wasting resources (the txset needs to be propagated) which can make consensus slower.

We should never introduce a new value if `!mCandidates.empty()`: when we have candidates we already know that nomination will converge, and new values are extremely likely to be ignored by other nodes anyways (paper specifies that "As soon as `v` has a candidate value, however, it must cease voting to `nominate x` for any new values `x`.").

We should
 - [ ] add a metric that reports how many unique values were referenced by a consensus round.
 - [ ] stop the nomination timer as soon as we add something to `mCandidates`. There is no need for the timeout callback at this point: there is no point in picking a new leader or to introduce a new value (from self or from a leader). Values will continue to be promoted based on messages received.

marta-lokhova:
Resolves #3523 

- Cancel the nomination timer as soon as we have a candidate value. Turn `nominate` into a no-op when a node confirmed a candidate (just in case it ever gets called outside of the timer).
- add a metric to count how many values are referenced per consensus round. This is done at the slot purge time by counting referenced tx sets in the cache, so this way we know how many values we had to pull. Note that this metric includes value from self as well, even though sometimes the locally generated value might not actually be propagated to the network. If this is an issue, I can add an extra condition to count the local value only if the node actually nominated. Or we can move the cache population call (addTxSet) to when the node actually nominates its local value (but this might cause the node to pull the tx set it generated locally anyway)

### DIFF
--- a/src/scp/NominationProtocol.cpp
+++ b/src/scp/NominationProtocol.cpp
@@ -505,16 +505,19 @@ NominationProtocol::nominate(ValueWrapperPtr value, Value const& previousValue,
                              bool timedout)
 {
     ZoneScoped;
-    CLOG_DEBUG(SCP, "NominationProtocol::nominate ({}) {}", mRoundNumber,
-               mSlot.getSCP().getValueString(value->getValue()));
 
     // No need to continue nominating, as per the whitepaper:
     // "As soon as `v` has a candidate value, however, it must cease
     // voting to nominate `x` for any new values `x`"
     if (!mCandidates.empty())
     {
+        CLOG_DEBUG(SCP, "Skip nomination round {}, already have a candidate",
+                   mRoundNumber);
         return false;
     }
+
+    CLOG_DEBUG(SCP, "NominationProtocol::nominate ({}) {}", mRoundNumber,
+               mSlot.getSCP().getValueString(value->getValue()));
 
     bool updated = false;
 
--- a/src/scp/NominationProtocol.cpp
+++ b/src/scp/NominationProtocol.cpp
@@ -447,6 +447,12 @@ NominationProtocol::processEnvelope(SCPEnvelopeWrapperPtr envelope)
             {
                 mCandidates.emplace(a);
                 newCandidates = true;
+                // Stop the timer, as there's no need to continue nominating,
+                // per the whitepaper:
+                // "As soon as `v` has a candidate value, however, it must cease
+                // voting to nominate `x` for any new values `x`"
+                mSlot.getSCPDriver().stopTimer(mSlot.getSlotIndex(),
+                                               Slot::NOMINATION_TIMER);
             }
         }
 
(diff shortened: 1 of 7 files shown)


hidenori-shinohara (commented):
The change looks good to me! I had one small question about logging.

nano-o:
This makes sense to me at a high level. As I understand it, this applies to voting only, and we should keep accepting and confirming values even if we have a candidate already.

hidenori-shinohara:
LGTM!

> This makes sense to me at a high level. As I understand it, this applies to voting only, and we should keep accepting and confirming values even if we have a candidate already.

Yeah, that's how I understood it too.

src/scp/NominationProtocol.cpp
```diff
@@ -502,6 +502,14 @@ NominationProtocol::nominate(ValueWrapperPtr value, Value const& previousValue,
     CLOG_DEBUG(SCP, "NominationProtocol::nominate ({}) {}", mRoundNumber,
                mSlot.getSCP().getValueString(value->getValue()));
 
+    // No need to continue nominating, as per the whitepaper:
+    // "As soon as `v` has a candidate value, however, it must cease
```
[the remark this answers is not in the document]
marta-lokhova: updated!

src/herder/PendingEnvelopes.cpp
```diff
@@ -674,10 +677,20 @@ PendingEnvelopes::eraseBelow(uint64 slotIndex)
 
     // 0 is special mark for data that we do not know the slot index
     // it is used for state loaded from database
-    mTxSetCache.erase_if([&](TxSetFramCacheItem const& i) {
+    auto erased = mTxSetCache.erase_if([&](TxSetFramCacheItem const& i) {
```
MonsieurNicolas, line 680: I don't think this is a clean data source:
* txset cache is a cache -> there is no guarantee that you'll find all entries
* as we're counting when slots are evicted, there is a higher chance that eviction took place
* more noise may be added if multiple slots are purged at once

There is a function `logQuorumInformation` that seems to be a better place for this kind of tracking: it's only called when externalizing the latest slot, and with a ~fixed window to allow for messages to propagate.
I _think_ that you could combine something like `processCurrentState` and `getTxSetHashes` to build a set of unique values referenced by that slot.

marta-lokhova: Yeah, I think the initial idea was to get an estimate on how many values the node is processing per ledger, which included values we pulled but didn't record (if an envelope got rejected, for example). I agree that looking at the cache can be noisy, though in this case eviction is extremely unlikely (the cache is sufficiently big, as I'm assuming we want to be able to answer queries about recent ledgers). That being said, I think a metric for values propagated network-wide makes more sense here. I updated the code based on your suggestion.

MonsieurNicolas:
r+ 74a672e5c86f40572b902da2cf3a548185223ba5

PR #3545, opened 2022-09-22
nomination: reuse the value from votes

### LINKED ISSUE #3525: Nomination: consider not introducing a new value if we already voted for something
Right now when ["nominate" kicks in](https://github.com/stellar/stellar-core/blob/96951d3de4b12c7ad63577c878178f16a602fe3d/src/scp/NominationProtocol.cpp#L498) and the local node is a round leader, it will always add its locally generated value to its set of votes.

Instead we should not [introduce the new locally generated value](https://github.com/stellar/stellar-core/blob/96951d3de4b12c7ad63577c878178f16a602fe3d/src/scp/NominationProtocol.cpp#L529) at all if `!mVotes.empty()` (and letting things play out with whatever is in `mVotes`, which are valid values).
That way, the local node will only introduce a new locally generated value when it is a round leader and didn't hear from previous leaders before that.

By not introducing a new value on timeout, this increases the chance of fast convergence of the nomination protocol while reducing resource utilization.

note: we could also consider moving the introduction of the locally generated value after the code that introduces values from leaders. This would only matter when the local node is picked as leader at the same time than some other nodes. I do not think this presents an actual security risk (and the other leaders already voted, so they're faster than the local node anyways).

marta-lokhova:
resolves #3525

### DIFF
--- a/src/scp/NominationProtocol.cpp
+++ b/src/scp/NominationProtocol.cpp
@@ -525,18 +525,6 @@ NominationProtocol::nominate(ValueWrapperPtr value, Value const& previousValue,
     std::chrono::milliseconds timeout =
         mSlot.getSCPDriver().computeTimeout(mRoundNumber);
 
-    // if we're leader, add our value
-    if (mRoundLeaders.find(mSlot.getLocalNode()->getNodeID()) !=
-        mRoundLeaders.end())
-    {
-        auto ins = mVotes.insert(value);
-        if (ins.second)
-        {
-            updated = true;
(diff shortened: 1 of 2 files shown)


hidenori-shinohara:
I think the change does exactly what the issue describes! 

I'm not sure about the implication of this change in the pubnet after reading this change and the issue. Is the idea that it's better to increase the chance of fast convergence at the expense of possibly making it harder for local values to be included?

nano-o:
Makes sense to me.

MonsieurNicolas:
r+ 513c9065ee1e6a2683c6eaa1c127d2a2fb40678b

PR #3554, opened 2022-09-27
Avoid redundant transaction advertising

### LINKED ISSUE #3515: Pull mode advertises a transaction to a node even if it has advertised it to the node
Came up in a discussion on https://github.com/stellar/stellar-core/pull/3479

In pull mode, a node sends a transaction after it sends its peer the hash and the peer told that it doesn't that the txn. 

Consider the following scenario:

- Node 1 and Node 2 *advertise* tx 1 to Node 3.
- Node 3 demands and receives the txn from Node 1.
- Now Node 3 broadcasts tx 1 and advertises it to Node 2.

Since Node 2 has already advertised it to Node 3, Node 3 can assume that Node 2 already has the txn. This is not a bug, but this may improve the performance of pull mode as it reduces duplicate traffic.

Note: We don't advertise a transaction if the peer has already sent the same transaction. This issue is for when the peer has *advertised* it to us.

Relevant issue: https://github.com/stellar/stellar-core/issues/3384

marta-lokhova:
Resolves #3515 

This change:
- Populates floodmap with tx hashes received in adverts
- Forces the caller to pass a hash used a key in order to support adverts

I'm not a huge fan of using different hashes like this, as it's quite error-prone, so I'm open to suggestions. The alternative is to have transactions-specific functions and data structures in Floodgate, but then we'll essentially have to duplicate all the functional calls to accommodate transactions and everything else separately.

### DIFF
--- a/src/overlay/Floodgate.h
+++ b/src/overlay/Floodgate.h
@@ -74,8 +74,5 @@ class Floodgate
     void forgetRecord(Hash const& msgID);
 
     void shutdown();
-
-    void updateRecord(StellarMessage const& oldMsg,
-                      StellarMessage const& newMsg);
 };
 }
--- a/src/overlay/Floodgate.h
+++ b/src/overlay/Floodgate.h
@@ -14,7 +14,8 @@
  * either send M to P once (and only once), or receive M _from_ P (thereby
  * inhibit sending M to P at all).
  *
- * The broadcast message types are TRANSACTION and SCP_MESSAGE.
+ * The broadcast message types are TRANSACTION, SCP_MESSAGE, SURVEY_RESPONSE and
+ * SURVEY_REQUEST.
  *
  * All messages are marked with the ledger sequence number to which they
  * relate, and all flood-management information for a given ledger number
@@ -54,14 +55,11 @@ class Floodgate
     // forget data strictly older than `maxLedger`
     void clearBelow(uint32_t maxLedger);
     // returns true if this is a new record
-    // fills msgID with msg's hash
-    bool addRecord(StellarMessage const& msg, Peer::pointer fromPeer,
-                   Hash& msgID);
+    bool addRecord(Peer::pointer fromPeer, Hash const& msgID);
 
     // returns true if msg was sent to at least one peer
     // The hash required for transactions
-    bool broadcast(StellarMessage const& msg, bool force,
-                   std::optional<Hash> const& hash = std::nullopt);
+    bool broadcast(StellarMessage const& msg, bool force, Hash const& index);
 
     // returns the list of peers that sent us the item with hash `msgID`
     // NB: `msgID` is the hash of a `StellarMessage`
--- a/src/overlay/Floodgate.h
+++ b/src/overlay/Floodgate.h
@@ -68,6 +68,7 @@ class Floodgate
     // removes the record corresponding to `msgID`
     // `msgID` corresponds to a `StellarMessage`
     void forgetRecord(Hash const& msgID);
+    void forgetRecordForPeer(Hash const& msgID, Peer::pointer peer);
 
     void shutdown();
 };
--- a/src/overlay/Floodgate.h
+++ b/src/overlay/Floodgate.h
@@ -37,11 +37,9 @@ class Floodgate
         typedef std::shared_ptr<FloodRecord> pointer;
 
         uint32_t mLedgerSeq;
-        StellarMessage mMessage;
         std::set<std::string> mPeersTold;
 
-        FloodRecord(StellarMessage const& msg, uint32_t ledger,
-                    Peer::pointer peer);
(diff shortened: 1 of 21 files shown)


src/overlay/test/OverlayManagerTests.cpp
```diff
@@ -285,13 +285,6 @@ class OverlayManagerTests
         crank(10);
         std::vector<int> expectedFinal{2, 2, 1, 2, 2};
         REQUIRE(sentCounts(pm) == expectedFinal);
-
```
MonsieurNicolas, line 288: Thanks! I did some archeology and this was added for CAP-0015 transition and we forgot to remove this as part of the post rollout cleanup

src/overlay/Floodgate.h
```diff
@@ -37,11 +37,9 @@ class Floodgate
         typedef std::shared_ptr<FloodRecord> pointer;
 
         uint32_t mLedgerSeq;
-        StellarMessage mMessage;
```
MonsieurNicolas, line 40: nice! This should save quite a bit of memory

src/overlay/OverlayManagerImpl.cpp
```diff
@@ -959,11 +966,10 @@ OverlayManagerImpl::shufflePeerList(std::vector<Peer::pointer>& peerList)
 }
 
 bool
-OverlayManagerImpl::recvFloodedMsgID(StellarMessage const& msg,
-                                     Peer::pointer peer, Hash& msgID)
+OverlayManagerImpl::recvFloodedMsg(Peer::pointer peer, Hash const& msgID)
```
MonsieurNicolas, line 969: This change makes it quite error prone: we had bugs in the past because of hash scheme mismatch. While we may not be able to avoid this at the floodmap level, we can avoid exposing this complexity outside of `OverlayManager`.
If we want to track transactions and msgs separately, we should just add  overloads at the `OverlayManager` level that take both (and do the right thing when passing a transaction wrapped in a `StellarMessage`)

src/overlay/Peer.cpp
```diff
@@ -2009,6 +2007,12 @@ Peer::recvSurveyResponseMessage(StellarMessage const& msg)
 void
 Peer::recvFloodAdvert(StellarMessage const& msg)
 {
+    auto self = shared_from_this();
+    for (auto const& hash : msg.floodAdvert().txHashes)
+    {
+        mApp.getOverlayManager().recvFloodedMsg(self, hash);
```
MonsieurNicolas, line 2013: This creates quite of an amplification potential with the current implementation as we bypass flow control here by a factor of 1000. We probably need to either not use the flood map to track adverts (we kinda already do that), or make the floodmap more resilient to DoS (note that `OverlayManager` already has access to both flood map and pull mode related data)

src/transactions/TransactionFrameBase.cpp
```diff
@@ -24,4 +24,10 @@ TransactionFrameBase::makeTransactionFromWire(Hash const& networkID,
         abort();
     }
 }
+
+Hash
+TransactionFrameBase::txHashForFlooding(TransactionFrameBasePtr const& tx)
```
MonsieurNicolas, line 29: this does not seem to be the right place for this helper: `TransactionFrameBase` should not decide what overlay does

MonsieurNicolas (changes requested):

src/overlay/OverlayManagerImpl.cpp
marta-lokhova: Yeah I wasn't super happy with this approach either. The problem is that with adverts, we only have transaction hashes, not full messages, so we have to provide an interface that only takes a hash. So I figured the simplest thing would be to force callers to compute the "keys" for whatever messages they want to remember

src/overlay/OverlayManagerImpl.cpp
```diff
@@ -47,6 +48,18 @@ constexpr std::chrono::seconds PEER_IP_RESOLVE_RETRY_DELAY(10);
 // longer than 2 seconds between re-issuing demands.
 constexpr std::chrono::seconds MAX_DELAY_DEMAND{2};
 
+Hash
+OverlayManager::defaultFloodingHash(StellarMessage const& msg)
```
hidenori-shinohara, line 52: Maybe we can throw if `msg` is a transaction?

hidenori-shinohara (commented):
I understand how this may be error prone, but also I don't have a good idea to change the design. Initially, I thought that maybe we can create a wrapper for a hash so we won't accidentally mix sha256 and blake2, but I realized that we do want to be able to use both of them as a key to the same map, so that won't make sense.

marta-lokhova:
We decided to go ahead with https://github.com/stellar/stellar-core/pull/3577, as it's much simpler to reason about and less error-prone. Closing this PR.

PR #3560, opened 2022-10-03
Include ledger sequence in preflight response

tamirms:
# Description

This commit adds a ledger sequence to the preflight response which is useful because execution of a smart contract will depend on the state of the ledger when the operation is applied.

# Checklist
- [x] Compiles
- [x] Ran all tests

### DIFF
--- a/src/transactions/InvokeHostFunctionOpFrame.cpp
+++ b/src/transactions/InvokeHostFunctionOpFrame.cpp
@@ -220,9 +220,11 @@ InvokeHostFunctionOpFrame::preflight(Application& app,
 {
     auto cb = std::make_unique<PreflightCallbacks>(app);
     Json::Value root;
+    LedgerTxn ltx(app.getLedgerTxnRoot());
+    root["ledger"] = ltx.loadHeader().current().ledgerSeq;
+
     try
     {
-        LedgerTxn ltx(app.getLedgerTxnRoot());
         PreflightHostFunctionOutput out = rust_bridge::preflight_host_function(
             toVec(op.function), toVec(op.parameters), toVec(sourceAccount),
             getLedgerInfo(ltx, app.getConfig()), std::move(cb));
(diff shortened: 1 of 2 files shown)


paulbellamy (approved):
lgtm, but idk

src/transactions/InvokeHostFunctionOpFrame.cpp
```diff
@@ -220,9 +220,11 @@ InvokeHostFunctionOpFrame::preflight(Application& app,
 {
     auto cb = std::make_unique<PreflightCallbacks>(app);
     Json::Value root;
+    LedgerTxn ltx(app.getLedgerTxnRoot());
+    root["ledger"] = ltx.loadHeader().current().ledgerSeq;
```
MonsieurNicolas, line 224: Any reason for moving `ltx` out of the try/catch block? The constructor call throw and this would crash the node instead of returning an error

tamirms: I assumed that if the invoke function operation failed (e.g. the smart contract reverted) then that would throw an exception and we would enter the catch clause. In that casem it would still be nice to know the ledger sequence of the invoke function attempt. 

I assumed that moving `ltx` out of the try/catch block should be safe because we already do that in the get ledger entry endpoint:

https://github.com/stellar/stellar-core/blob/e7f06eaa81448891c8503cb45b25adb280c55d84/src/main/CommandHandler.cpp#L722-L723

Note how the same code is executed without any try / catch block in the handler.

MonsieurNicolas:
r+ e7f06eaa81448891c8503cb45b25adb280c55d84

src/transactions/InvokeHostFunctionOpFrame.cpp
MonsieurNicolas: Actually commandhandler has a try/catch anyways at the top level (just returns a 500) so we're good

PR #3561, opened 2022-10-04
Add some soroban metrics.

graydon:
This adds an initial rough set of new metrics related to soroban. They're all grouped under `host-fn` and track basic parameters of invocations and preflights: ledger entries and their sizes read and written, events emitted, CPU and memory budget consumption and overall execution time.

I did not document them in metrics.md because they're entirely ifdef'ed out at this point, and likely to change before we finalize them.

### DIFF
--- a/src/transactions/InvokeHostFunctionOpFrame.cpp
+++ b/src/transactions/InvokeHostFunctionOpFrame.cpp
@@ -6,6 +6,8 @@
 // This needs to be included first
 #include "util/GlobalChecks.h"
 #include <json/json.h>
+#include <medida/metrics_registry.h>
+#include <xdrpp/types.h>
 #ifdef ENABLE_NEXT_PROTOCOL_VERSION_UNSAFE_FOR_PRODUCTION
 #include "rust/RustVecXdrMarshal.h"
 #endif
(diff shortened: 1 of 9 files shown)


src/transactions/InvokeHostFunctionOpFrame.cpp
```diff
@@ -82,9 +84,159 @@ InvokeHostFunctionOpFrame::doApply(AbstractLedgerTxn& ltx)
...
+        }
+        else
+        {
+            mReadDataByte += n;
+        }
+    }
+
+    ~HostFunctionMetrics()
+    {
+        mMetrics.NewMeter({"host-fn", mFnType, "read-entry"}, "entry")
```
MonsieurNicolas, line 182: If I remember correctly - doing `NewMeter` often like this is very slow and should be avoided (because medida). We probably do not care right now, but will probably show up when we profile later on

src/transactions/InvokeHostFunctionOpFrame.cpp
```diff
@@ -93,35 +245,48 @@ InvokeHostFunctionOpFrame::doApply(AbstractLedgerTxn& ltx, Config const& cfg)
...
             getLedgerInfo(ltx, cfg), ledgerEntryCxxBufs);
+        metrics.mSuccess = true;
     }
     catch (std::exception&)
     {
         innerResult().code(INVOKE_HOST_FUNCTION_TRAPPED);
         return false;
     }
 
+    metrics.mCpuInsn = out.cpu_insns;
```
MonsieurNicolas, line 287: we are not doing this accounting if the invoke traps which may hide some patterns. Any way we can still get a read on this in all cases? (I think the limitation is coming from relying on the result for this)

MonsieurNicolas (changes requested):
looks good - just one question wrt failure cases

src/transactions/InvokeHostFunctionOpFrame.cpp
graydon: Yeah, I made this local counter-accumulating object in part to push it back to "one call per metric" rather than "one per entry" but caching the metrics object references higher up be .. a bit more plumbing. I guess I could make a different (non `medida::MetricsRegistry`) object like `OpMetrics` that cache such references, and is passed in and out of each op method, similar to the way `OverlayMetrics` caches them. But for now there's only one op per tx, so this would have to be living up above the tx level -- spanning txs -- to make a difference, and it seemed likely to me not to be worth the effort (these ops are already very very costly, it's probably lost in the noise).

src/transactions/InvokeHostFunctionOpFrame.cpp
graydon: Yeah, we get no output in case of an exception, so there's nothing to read this out of. The information channel in Rust errors translated to C++ exceptions is quite narrow, I don't think it'd be feasible to convey that way.

MonsieurNicolas:
We can merge this as is then, but it needs to be rebased

src/transactions/InvokeHostFunctionOpFrame.cpp
MonsieurNicolas: ok. I guess we'll still have the overall timing so it's not that bad. Still annoying that we can't know how much actual instructions we executed for an entire ledger

graydon:
(Assuming you're afk and r+'ing the rebase myself, given above comment and desire to be included in any of tomorrow's builds.)

graydon:
r+ 5705607

PR #3563, opened 2022-10-04
Fix nested-ltx bug introduced in #3557

graydon:
PR #3557 introduced a root ltx that is kept alive during the preflight call, which causes preflight callbacks to fail (they establish a secondary root ltx). This change hoists that outer ltx to an earlier block that completes and destroys the ltx before the preflight.


MonsieurNicolas:
r+ 1ff3fb97ab111b0ae5c15a4c5c9c0dc75c67f6dc

PR #3564, opened 2022-10-04
Propagate cxx::Exception strings from preflight callbacks

graydon:
This PR does exception-explanation string propagation in an awkward context: when core's preflight endpoint calls the soroban host, and the soroban host calls core _back_ to synchronously serve preflight ledger accesses, and then one of those ledger accesses fails and generates a C++ exception, which we receive back in Rust as a `cxx::Exception` (i.e. a bridge value carrying a string converted from the C++ exception).

Before this PR, we just converted that `cxx::Exception` to a low-content `HostError` generated from a fairly uninformative status code. With this PR, we plumb a reference to the soroban host into the callbacks, so that we can record the `what()` string from the `cxx::Exception` into the soroban host debug-event buffer, and then generate a `HostError` that captures that buffer, ultimately giving the user (and ultimately the core dev diagnosting the problem) a bit more context about the error.

Doing this introduces an awkward temporary cyclical-ownership situation as well as an awkward potential borrow failure, but the former can be handled by explicitly breaking the edge after the call (as is done in the code here) and the latter can only fail with behavior no worse than the existing behavior.

All of this may be temporary, depending on the fate of (synchronous, dangerous) preflight-in-core. We might keep it around forever, hard to tell. But this change improves things for the time being.

### DIFF
--- a/src/rust/src/contract.rs
+++ b/src/rust/src/contract.rs
@@ -15,7 +15,7 @@ use std::{cell::RefCell, fmt::Display, io::Cursor, panic, pin::Pin, rc::Rc};
 
 use soroban_env_host::{
     budget::Budget,
-    events::{Events, HostEvent},
+    events::{DebugError, DebugEvent, Events, HostEvent},
     storage::{self, AccessType, SnapshotSource, Storage},
     xdr,
     xdr::{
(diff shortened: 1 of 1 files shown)


MonsieurNicolas:
r+ 33da13520ee234e6d9f72850f1442985c6d3434a

PR #3567, opened 2022-10-05
Bump env and add complex-wasm testcase.

graydon:
This bumps the env by a few days (to 17ada13), absorbing recent changes over the past week, including those that disable ed25519 contract creation, and a recent new wasm testcase I added today that exercises a bit more of the recently-enabled host embedding interface (preflight, events, etc) from the guest side.

It should probably only land if @paulbellamy and @leighmcculloch want it currently -- I don't want to get in anyone's way / break e2e tests just about to start working, but .. if it's desirable to shift to a slightly newer env, this does so.

(It also removes the ed25519 tests @sisuresh added a while back -- they can be fished out of git and re-enabled when that redesign is completed)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -527,7 +527,7 @@ dependencies = [
 [[package]]
 name = "soroban-env-common"
 version = "0.0.5"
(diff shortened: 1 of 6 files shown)


MonsieurNicolas:
r+ 15baacfc90e8567a5218d0c219021db36a183a2e

PR #3569, opened 2022-10-06
Vote for accepted values from round leaders

### LINKED ISSUE #3524: Nomination: consider reducing which values to use from round leaders
Right now we pick the "highest value" from the set of values that a leader votes for as per https://github.com/stellar/stellar-core/blob/96951d3de4b12c7ad63577c878178f16a602fe3d/src/scp/NominationProtocol.cpp#L335

But if a round leader has already accepted at least one value (meaning a quorum already saw those values) we should limit our value selection to one of those values (assuming at least one of them is considered "valid" by the local node).

This would reduce the chance of multiple values having to compete during nomination, and as a consequence speed up consensus.

marta-lokhova:
Resolves #3524

### DIFF
--- a/src/scp/NominationProtocol.cpp
+++ b/src/scp/NominationProtocol.cpp
@@ -339,8 +339,9 @@ NominationProtocol::getNewValueFromNomination(SCPNomination const& nom)
     // sorted using hashValue.
     ValueWrapperPtr newVote;
     uint64 newHash = 0;
+    bool foundValidValue = false;
 
-    applyAll(nom, [&](Value const& value) {
+    auto pickValue = [&](Value const& value) {
         ValueWrapperPtr valueToNominate;
         auto vl = validateValue(value);
         if (vl == SCPDriver::kFullyValidatedValue)
@@ -353,6 +354,7 @@ NominationProtocol::getNewValueFromNomination(SCPNomination const& nom)
         }
         if (valueToNominate)
         {
+            foundValidValue = true;
             if (mVotes.find(valueToNominate) == mVotes.end())
             {
                 uint64 curHash = hashValue(valueToNominate->getValue());
@@ -363,7 +365,22 @@ NominationProtocol::getNewValueFromNomination(SCPNomination const& nom)
                 }
             }
         }
-    });
+    };
+
+    for (auto const& val : nom.accepted)
+    {
+        pickValue(val);
+    }
+
+    // Move on to votes if we have not found a valid accepted value
+    if (!foundValidValue)
+    {
+        for (auto const& val : nom.votes)
+        {
+            pickValue(val);
+        }
+    }
+
     return newVote;
 }
 
--- a/src/scp/test/SCPTests.cpp
+++ b/src/scp/test/SCPTests.cpp
@@ -2821,6 +2821,57 @@ TEST_CASE("nomination tests core5", "[scp][nominationprotocol]")
                 verifyNominate(scp.mEnvs[1], v0SecretKey, qSetHash0, 0, votesXK,
                                {});
             }
+            SECTION("select accepted value from leader")
+            {
+                REQUIRE(xValue < yValue);
+                REQUIRE(yValue < zValue);
+                REQUIRE(scp.mEnvs.size() == 1);
+
+                // Update round leader to v1
+                scp.mPriorityLookup = [&](NodeID const& n) {
+                    return (n == v1NodeID) ? 1000 : 1;
+                };
+
+                SCPEnvelope nom1 = makeNominate(v1SecretKey, qSetHash, 0,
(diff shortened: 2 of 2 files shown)


src/scp/NominationProtocol.cpp
```diff
@@ -363,7 +363,23 @@ NominationProtocol::getNewValueFromNomination(SCPNomination const& nom)
...
+    };
+
+    for (auto const& val : nom.accepted)
+    {
+        pickValue(val);
+    }
+
+    // Only if we have not selected an accepted value, pick value a round leader
+    // voted for
+    if (!newVote)
```
MonsieurNicolas, line 375: this is not exactly right: you don't want to introduce values from `nom.votes` if there was any `valueToNominate`. Consider the situation where `mVotes` already contains `nom.accepted`

marta-lokhova: Oh hmm, I thought we want to introduce a value we haven't voted for either way, not no-op in this case. I can update that though, but want to make sure it's safe (so we should wait to get input from Giuliano on the issue as well)

src/scp/test/SCPTests.cpp
```diff
@@ -2821,6 +2821,48 @@ TEST_CASE("nomination tests core5", "[scp][nominationprotocol]")
...
                                {});
             }
+            SECTION("select accepted value from leader")
+            {
+                REQUIRE(xValue < yValue);
+                REQUIRE(yValue < zValue);
+                REQUIRE(scp.mEnvs.size() == 1);
+
+                // Update round leader to v1
+                scp.mPriorityLookup = [&](NodeID const& n) {
```
MonsieurNicolas, line 2831: You probably want this section to be under `SECTION("v1 is top node")` so that you don't end up with `leaders {= v0, v1}`

src/scp/test/SCPTests.cpp
```diff
@@ -2821,6 +2821,48 @@ TEST_CASE("nomination tests core5", "[scp][nominationprotocol]")
...
+
+                    // Vote for accepted value (z)
+                    scp.receiveEnvelope(nom1);
+                    REQUIRE(scp.mEnvs.size() == 2);
+                }
+
+                std::vector<Value> votesXY;
+                votesXY.emplace_back(xValue);
+                votesXY.emplace_back(yValue);
+                std::sort(votesXY.begin(), votesXY.end());
```
MonsieurNicolas, line 2861: `x < y` so you don't need to sort

src/scp/test/SCPTests.cpp
MonsieurNicolas: actually we probably want to check both cases (v0 leader first and v1 leader from the beginning)?

src/scp/test/SCPTests.cpp
```diff
@@ -2821,6 +2821,48 @@ TEST_CASE("nomination tests core5", "[scp][nominationprotocol]")
...
+
+                verifyNominate(scp.mEnvs[1], v0SecretKey, qSetHash0, 0, votesXY,
```
MonsieurNicolas, line 2863: you're missing receiving
`SCPEnvelope nom2 = makeNominate(v1SecretKey, qSetHash, 0,
                                                {yValue, zValue, zzValue}, {yValue});`
and perform checks that if we receive this before and after timeout we still only vote for the same thing (this would have caught the bug that you had in your first implementation)

MonsieurNicolas (changes requested):

src/scp/test/SCPTests.cpp
marta-lokhova: not sure I follow why both are needed. In the current check, we know exactly who the leaders are and what they voted for. And since the values are all different, we just need to make sure the updated function is called correctly and the right values are selected.

src/scp/test/SCPTests.cpp
marta-lokhova: I can add this, though I don't think both before/after timeout sections are needed. We already tested above that the updated function is called appropriately in different scenarios wrt timeout. Here we just need to test that the actual function does what we want (i.e. does not select votes)

src/scp/test/SCPTests.cpp
MonsieurNicolas: yes you are correct

src/scp/test/SCPTests.cpp
MonsieurNicolas: agree - we just need to test any code path that hits `getNewValueFromNomination`

marta-lokhova:
@MonsieurNicolas all comments have been addressed, so I believe this is ready to land

MonsieurNicolas:
r+ 0c990f9f6d16fded9cc41d83c3d7ef85af069c04

PR #3570, opened 2022-10-07
Plumb errors status and event for failed contract host fn invocations.

dmkozh:
# Description

This is a proof-of-concept PR, maybe it could be improved somehow.

I'm also not sure about plumbing the events; we can get rid of them if the proposed approach isn't feasible for some reason.

The XDR changes are in https://github.com/stellar/stellar-xdr-next/pull/48, so this won't build initially.

# Checklist
- [ ] Compiles
- [ ] Ran all tests

### DIFF
--- a/src/transactions/InvokeHostFunctionOpFrame.cpp
+++ b/src/transactions/InvokeHostFunctionOpFrame.cpp
@@ -283,6 +283,24 @@ InvokeHostFunctionOpFrame::doApply(AbstractLedgerTxn& ltx, Config const& cfg,
         innerResult().code(INVOKE_HOST_FUNCTION_TRAPPED);
         return false;
     }
+    
+    // Append events to the enclosing TransactionFrame, where
+    // they'll be picked up and transferred to the TxMeta.
+    for (auto const& buf : out.contract_events)
+    {
+        metrics.mEmitEvent++;
+        metrics.mEmitEventByte += buf.data.size();
+        ContractEvent evt;
+        xdr::xdr_from_opaque(buf.data, evt);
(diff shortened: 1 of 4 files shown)


src/transactions/InvokeHostFunctionOpFrame.cpp
```diff
@@ -283,6 +283,24 @@ InvokeHostFunctionOpFrame::doApply(AbstractLedgerTxn& ltx, Config const& cfg,
         innerResult().code(INVOKE_HOST_FUNCTION_TRAPPED);
         return false;
     }
+    
+    // Append events to the enclosing TransactionFrame, where
+    // they'll be picked up and transferred to the TxMeta.
+    for (auto const& buf : out.contract_events)
```
graydon, line 289: I don't think we want to report events on error -- the transaction is going to be rolled back, which implies that whatever the events are describing didn't actually happen.

graydon (changes requested):
I'm surprised this is necessary but I guess it is! I would just leave the events where they are, I don't think they should be emitted on error -- it's a failed transaction after all.

graydon:
I pushed a commit here that makes the change above, as well as the SCStatus version matching the XDR change, and also a couple other minor cleanups in the way the code was written. Otherwise LGTM now!

src/transactions/InvokeHostFunctionOpFrame.cpp
dmkozh: Sounds good, I confused those with the debug events which could be meaningful for errors. Thanks for fixing this.

Issue #3572, opened 2022-10-07
Turn on pull mode by default


marta-lokhova:
- Flip the switch on the pull mode config
- Update overlay versions

marta-lokhova:
note that we should bump the overlay version, and _maybe_ bump minimum overlay version to support ~3 month backward-compatibility

PR #3575, opened 2022-10-07
Turn on pull mode by default

hidenori-shinohara:
# Description

Resolves https://github.com/stellar/stellar-core/issues/3572

This turns on pull mode by default. Note that node operators can still turn it off through the config file. Furthermore, this bumps the overlay protocol version.

# Checklist
- [x] Compiles
- [x] Ran all tests


marta-lokhova (changes requested):
thanks. we also need to make sure that minimum overlay version is up-to-date

marta-lokhova:
note: we should merge this change last, before closing the tree, so we can do perf run before/after with all the other changes already in

src/main/Config.cpp
```diff
@@ -127,7 +127,7 @@ Config::Config() : NODE_SEED(SecretKey::random())
 
     MAXIMUM_LEDGER_CLOSETIME_DRIFT = 50;
 
-    OVERLAY_PROTOCOL_MIN_VERSION = 21;
+    OVERLAY_PROTOCOL_MIN_VERSION = 22;
```
MonsieurNicolas, line 130: we can actually bump to 23 that corresponds to 19.3.0 released in July (this is in line with our release policy of 90 days)

MonsieurNicolas (changes requested):

src/main/Config.cpp
hidenori-shinohara: Updated!

MonsieurNicolas:
r+ 83da984d3dc05eb67ed50b0045de0045c697dcac

PR #3577, opened 2022-10-08
Pull mode advert cache

marta-lokhova:
Replaces https://github.com/stellar/stellar-core/pull/3554
Resolves https://github.com/stellar/stellar-core/issues/3515

### DIFF
--- a/src/overlay/test/OverlayTests.cpp
+++ b/src/overlay/test/OverlayTests.cpp
@@ -1886,6 +1886,12 @@ auto numTxHashesAdvertised = [](std::shared_ptr<Application> app) {
         .count();
 };
 
+auto numFulfilled = [](std::shared_ptr<Application> app) {
+    return app->getMetrics()
+        .NewMeter({"overlay", "flood", "fulfilled"}, "message")
+        .count();
+};
+
 TEST_CASE("overlay pull mode", "[overlay][pullmode]")
 {
     VirtualClock clock;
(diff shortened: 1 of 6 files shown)


src/overlay/test/OverlayTests.cpp
```diff
@@ -1928,6 +1934,92 @@ TEST_CASE("overlay pull mode", "[overlay][pullmode]")
...
+                    TransactionQueue::AddResult::ADD_STATUS_PENDING);
+        };
+
+        SECTION("pull mode enabled on all")
+        {
+            twoNodesRecvTx();
+
+            // Give enough time for Node2 to issue a demand and receive tx0
+            testutil::crankFor(clock,
+                               3 * apps[2]->getConfig().FLOOD_DEMAND_PERIOD_MS +
```
hidenori-shinohara, line 1958: What do you think about using an arbitrarily large number like 10 seconds? 

If the default values for `FLOOD_ADVERT_PERIOD_MS` or `FLOOD_DEMAND_PERIOD_MS` change, this test may break. (e.g., If `ADVERT_PERIOD = 200 ms` and `DEMAND_PERIOD = 50 ms`, then the node may not advertise.) 

I had the same thought about line 2000.

marta-lokhova: Updated with a fixed number.

marta-lokhova:
There was a bug in the test (hardcoded value for the peer fulfilling the demand, but we demand from peers randomly, so it could be any peer). It is now fixed and all checks are passing.

src/overlay/test/OverlayTests.cpp
```diff
@@ -1928,6 +1934,92 @@ TEST_CASE("overlay pull mode", "[overlay][pullmode]")
...
+            // Node0 and Node1 know about tx0 and will advertise it to Node2
+            // Node3 will flood the transaction directly
+            twoNodesRecvTx();
+            REQUIRE(apps[3]->getHerder().recvTransaction(tx, true) ==
+                    TransactionQueue::AddResult::ADD_STATUS_PENDING);
+
+            // Give enough time for Node2 to receive tx0 from Node3 and adverts
+            // from Node0 and Node1
+            testutil::crankFor(clock,
+                               3 * apps[2]->getConfig().FLOOD_DEMAND_PERIOD_MS +
```
hidenori-shinohara: What do you think about using a fixed number here as well?

marta-lokhova: This is fixed already in the last commit.

hidenori-shinohara: Oops, my bad!

hidenori-shinohara:
This looks good to me!

marta-lokhova:
@MonsieurNicolas I believe your concerns are addressed now. Specifically, we now purge advert caches every time we do cleanups during ledger close similar to flood map. For a moment, I considered reviving https://github.com/stellar/stellar-core/pull/3554 to avoid duplicating functionality. However, realized that I still like this change much better:
- having per-peer cache allows much better DoS resistance, as the impact is bounded without affecting other peers
- No need to force users to pass differently calculated hashes (super error-prone)
- We still get a little bit of duplication, but it seems not too bad (the change to add purging is very small, and the purging behavior is duplicated across subsystems already anyway)

MonsieurNicolas:
Indeed @marta-lokhova this is a good change (and a lot simpler than the other PR). Thank you

MonsieurNicolas:
r+ 3ac0a579489f9b6f604a61274c360f3736343ad7

PR #3579, opened 2022-10-08
Build rust code with heavier optimization settings.

graydon:
This just changes the release build parameters for the rust code so that, in particular, the wasmi VM inlines all of its opcode processing blocks. This wins a factor of 2-3 in performance, in microbenchmarks; it makes builds take longer but seemes worth doing.


leighmcculloch (approved):
🎉 

Out of curiosity how much does reducing the codeunits slow down the build?

Mario24433 (approved):

MonsieurNicolas:
r+ 81783a49ad9374cdeeaeed8976870f802fc48321

PR #3580, opened 2022-10-14
adjust demand period in flood test to avoid dropping txs

MonsieurNicolas:
default value was causing nodes to pull a lot slower than adverts, causing advert queues to drop transactions (as expected)


graydon:
r+ 0e9b474

PR #3584, opened 2022-10-25
Added BUILD_TEST_COMMON flag and refactored fuzzer

### LINKED ISSUE #2835: It should be possible to build a "fuzzing enabled" build without all tests
Right now to build with `--enable-afl` requires building with tests (ie, can't build with `--disable-tests`).

This slows down builds a lot: tests represent a lot of additional code, and fuzzing instrumentation itself is very slow. On my machine a full build compiled like this takes almost an hour (without tests, I can build in about 5).

It also slows down the runtime: as the binary is a lot bigger and as the fuzzer loads the binary in a loop, fuzzing is noticeably slower.

As a consequence, we should move the fuzzing code into its own folder outside of the "test" folders.

SirTyson:
# Description

Resolves #2835. This PR refactors fuzzer related code to its own folder and divides test code into a `test` folder and `test-common` folder. By doing this, we can now build with both `--enable-afl` and `--disable-tests` flags set. Building with AFL instrumentation is slow, so by disabling unit tests, compilation is much faster and the resulting binary is 45% smaller. The smaller binary also speeds up fuzzing by around 2x since the fuzzer must load and execute this binary in a loop (see attached screenshot).

The new `test-common` folder contains code used by both unit tests and the simulation or fuzzer, and the `test` folder contains unit test code. As a general rule, if test code requires the Catch library, it's in `test`, otherwise it is in `test-common`. Corresponding to the `test` and `test-common` folders, there are also `BUILD_TESTS` and `BUILD_TESTS_COMMON` build flags. Note that if `BUILD_TESTS` is set, `BUILD_TESTS_COMMON` will also be set. 

# Checklist
- [x] Compiles
- [x] Ran all tests

Screenshot from 2022-10-25 14-53-33


SirTyson:
Bit rotted :(

PR #3585, opened 2022-10-25
Pin `cxxbridge` tool at v 1.0.68 to match `Cargo.lock`.

dmkozh:
# Description

Pin `cxxbridge` tool at v 1.0.68 to match `cxx` crates in `Cargo.lock`.

Ideally we would do that automatically, but that addresses the issue of version mismatch for the time being.

# Checklist
- [ ] Compiles
- [ ] Ran all tests

### DIFF
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -76,7 +76,7 @@ stellar_core_LDADD += $(LIBRUST_STELLAR_CORE) -ldl
 
 $(RUST_CXXBRIDGE):
 	mkdir -p $(RUST_BIN_DIR)
-	$(CARGO) install --root $(RUST_BUILD_DIR) cxxbridge-cmd
+	$(CARGO) install --root $(RUST_BUILD_DIR) cxxbridge-cmd --version 1.0.68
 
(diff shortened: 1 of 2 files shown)


src/Makefile.am
```diff
@@ -76,7 +76,7 @@ stellar_core_LDADD += $(LIBRUST_STELLAR_CORE) -ldl
 
 $(RUST_CXXBRIDGE):
 	mkdir -p $(RUST_BIN_DIR)
-	$(CARGO) install --root $(RUST_BUILD_DIR) cxxbridge-cmd
+	$(CARGO) install --root $(RUST_BUILD_DIR) cxxbridge-cmd --version 1.0.68
```
leighmcculloch, line 79: Fyi if you want to extract this from the `Cargo.lock` file so you don't have to manually keep it in sync, you can use this command:

```
cargo metadata --format-version 1 | jq -r '.packages[] | select(.name == "cxx") | .version'
```

I don't know how anything like that could be applied to the VisualStudio project file, but this would work for the Makefile at least.

MonsieurNicolas:
r+ 3c16f193f763e0d9c1dd724722467dba27b4f987

PR #3586, opened 2022-10-28
Remove spurious whitespace diffs between protocol-curr and next

graydon:
This is just a whitespace / no-op change to protocol-next to reduce spurious diffs against protocol-curr.


MonsieurNicolas:
We're now using the xdr repo as source of truth for everything

PR #3590, opened 2022-11-02
Require pull mode

### LINKED ISSUE #3583: Make tx pull mode required
Right now peers can opt-out of pull mode. This issue tracks enforcing pull mode.

hidenori-shinohara:
# Description

Resolves #3583 

- Bumps the overlay version and the min overlay version to the version that turns on pull mode by default.
- Removes the option to turn off pull mode and variables/logic/metrics that become obsolete when all connections use pull mode (e.g., pull mode percentage)
- Refuses a connection if the peer doesn't intend to turn on pull mode.

# Checklist
- [x] Compiles
- [x] Ran all tests


hidenori-shinohara:
This will be for v19.7.0 as pull mode is off by default for [v19.4.0](https://github.com/stellar/stellar-core/releases/tag/v19.4.0).

marta-lokhova:
@hidenori-shinohara looks like this PR needs to be refreshed

hidenori-shinohara:
I updated & rebased the PR. When I originally put this PR together almost 2 months ago, I added a metric to track the number of pulled txns with no demand history. I decided to delete it as I don't remember the context, I'm not sure that's useful and also that doesn't really belong in this PR to require pull mode.

src/overlay/Peer.cpp
```diff
@@ -1905,13 +1880,14 @@ Peer::recvAuth(StellarMessage const& msg)
     // the other peer about the local node's reading capacity.
     sendSendMore(mApp.getConfig().PEER_FLOOD_READING_CAPACITY);
 
-    if (mRemoteOverlayVersion >= Peer::FIRST_VERSION_SUPPORTING_PULL_MODE &&
-        mApp.getConfig().OVERLAY_PROTOCOL_VERSION >=
-            Peer::FIRST_VERSION_SUPPORTING_PULL_MODE &&
-        msg.auth().flags == AUTH_MSG_FLAG_PULL_MODE_REQUESTED &&
-        mApp.getConfig().ENABLE_PULL_MODE)
+    if (mRemoteOverlayVersion < Peer::FIRST_VERSION_REQUIRING_PULL_MODE &&
+        msg.auth().flags != AUTH_MSG_FLAG_PULL_MODE_REQUESTED)
```
marta-lokhova, line 1884: this check doesn't seem right: version 27 still sets msg.auth().flags == AUTH_MSG_FLAG_PULL_MODE_REQUESTED

src/overlay/Peer.cpp
```diff
@@ -1998,6 +1974,7 @@ Peer::recvFloodAdvert(StellarMessage const& msg)
         rememberHash(hash, seq);
     }
     mTxAdvertQueue.queueAndMaybeTrim(msg.floodAdvert().txHashes);
+    mApp.getOverlayManager().startDemandTimer();
```
marta-lokhova, line 1977: did this change because of the timer tests?

src/overlay/Peer.cpp
hidenori-shinohara: I want Core to eventually stop using `AUTH_MSG_FLAG_PULL_MODE_REQUESTED` https://github.com/stellar/stellar-core/issues/3636 once every node's overlay version is >= 27.

I decided to go with this logic since:
- Case 1: peer's overlay version < 27
    - The peer may have turned off pull mode. We'll know that by checking the flag.
- Case 2: peer's overlay version >=.
    -  The peer definitely wants to use pull mode. The peer may have used `AUTH_MSG_FLAG_PULL_MODE_REQUESTED`, but we don't have to check it.

src/overlay/Peer.cpp
hidenori-shinohara: The current pull-mode implementation starts the demand timer as soon as possible even if we haven't received an advert. This broke some timer tests as they're not aware of this new timer. The current workaround is to manually disable pull mode for those tests. 

Since this PR forces pull mode, the workaround is no longer possible. Instead of updating timer tests to accommodate pull mode, i decided to change the demand timer logic so it'll start demanding only after receiving the first advert.

src/overlay/Peer.cpp
hidenori-shinohara: I just realized that we know the peer's overlay version before we get to this line. (I mistakenly thought that that wasn't the case before 🤔 )

I updated the code so that we set `AUTH_MSG_FLAG_PULL_MODE_REQUESTED` only if the peer's overlay version is `< 27 = FIRST_VERSION_REQUIRING_PULL_MODE`.

src/overlay/test/OverlayTests.cpp
```diff
@@ -1970,49 +1969,6 @@ TEST_CASE("overlay pull mode", "[overlay][pullmode]")
                         .count() == 1);
             REQUIRE(overlaytestutils::getAdvertisedHashCount(apps[2]) == 0);
         }
-        SECTION("pull mode disabled on Node3")
```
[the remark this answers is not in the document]
hidenori-shinohara: I added a test case for that.

src/overlay/Peer.cpp
marta-lokhova: I don't think pull mode logic should be dictated by tests, as it's based on whatever policy we choose (in this case always start the timer), and the tests need to accommodate that. The problem you ran into isn't unique to pull mode, as there are other timers in the application. What you can do is cancel the demand timer, like we do with Herder timers here:
https://github.com/stellar/stellar-core/blob/master/src/util/test/TimerTests.cpp#L102

marta-lokhova (changes requested):

src/overlay/Peer.cpp
marta-lokhova: okay, that makes sense

src/overlay/test/LoopbackPeer.cpp
```diff
@@ -378,6 +401,18 @@ LoopbackPeer::getDropProbability() const
     return mDropProb.p();
 }
 
+bool
+LoopbackPeer::getDisablePullMode() const
```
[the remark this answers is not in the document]
hidenori-shinohara: Removed.

src/overlay/test/LoopbackPeer.cpp
```diff
@@ -60,6 +61,28 @@ LoopbackPeer::sendMessage(xdr::msg_ptr&& msg)
...
+    if (mDisablePullMode)
+    {
+        // This surgically changes the auth flag to disable pull mode.
+        // This is hacky but unfortunately necessary.
+        // The main code path forces pull mode on all connections,
+        // so no longer has the ability to express the intention to
+        // disable pull mode. This if clause along with `mDisablePullMode`
+        // can be safely removed once the min overlay version becomes
+        // FIRST_VERSION_REQUIRING_PULL_MODE.
+        AuthenticatedMessage am;
```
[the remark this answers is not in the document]
hidenori-shinohara: I tried to pursue this approach, but I started to feel that it probably wasn't a good idea in the first place.  I realized that I can turn on/off the pull mode flag only in tests by using `#ifdef BUILD_TESTS`, and I think that makes the tests much simpler. Let me know what you think.

hidenori-shinohara:
@marta-lokhova I updated the PR. The changes were big and the commit history was getting out of hand, I decided to rebase and force push. I hope the commits are easy to review despite the force push! 🤞

src/overlay/Peer.h
```diff
@@ -337,6 +336,15 @@ class Peer : public std::enable_shared_from_this<Peer>,
 
     bool mShuttingDown{false};
 
+#ifdef BUILD_TESTS
+    // For testing purposes, sometimes we want to force
+    // to set or not set the pull mode flag in the auth
+    // message. Set this optional bool value to force it.
+    // This will be obsolete once the min overlay version
+    // becomes >= FIRST_VERSION_REQUIRING_PULL_MODE.
+    std::optional<bool> mMaybeOverwritePullModeFlagForTesting;
```
marta-lokhova, line 345: Setting this to true has no effect since pull mode is already required by default, so this should just be `bool mOverrideDisablePullModeForTesting` which defaults to false, and can be set to true.

src/overlay/OverlayManagerImpl.cpp
```diff
@@ -1300,15 +1278,13 @@ OverlayManagerImpl::demand()
...
-    std::vector<Peer::pointer> pullModePeers;
-    for (auto const& peer : authenticatedPeers)
-    {
-        if (peer.second->isPullModeEnabled())
-        {
-            pullModePeers.push_back(peer.second);
-        }
-    }
-    stellar::shuffle(pullModePeers.begin(), pullModePeers.end(), gRandomEngine);
+    std::vector<Peer::pointer> peers;
```
marta-lokhova, line 1281: We should use `OverlayManagerImpl::getRandomAuthenticatedPeers` here instead

src/overlay/Peer.cpp
```diff
@@ -371,10 +363,30 @@ Peer::sendAuth()
...
+        // If the peer's overlay version indicates that
+        // they might attempt a non-pull-mode connection,
+        // then we should explicitly communicate the intention to
+        // use pull mode. HELLO includes the overlay version and
+        // is exchanged before AUTH, so we should have the remote
+        // overlay version by now.
         msg.auth().flags = AUTH_MSG_FLAG_PULL_MODE_REQUESTED;
     }
+#ifdef BUILD_TESTS
+    if (mMaybeOverwritePullModeFlagForTesting)
```
marta-lokhova, line 378: No need for an optional here, we can use a bool 
```
#ifdef BUILD_TESTS
if (mOverrideDisablePullModeForTesting)
{
    msg.auth().flags = 0;
}
#endif
```

marta-lokhova (changes requested):
Thanks for the updates. A few minor tweaks needed, otherwise LGTM!

src/overlay/OverlayManagerImpl.cpp
hidenori-shinohara: I didn't know about this! Thanks!

src/overlay/Peer.h
hidenori-shinohara: If it's a bool, i don't think we can test the case when the peer is using an old overlay version and wants to turn on pull mode. 

- A: Old overlay version that wants to turn on pull mode
- B: The latest overlay version that always turns on pull mode

Then, in this test, A will see that B uses the latest overlay version that turns on pull mode, and will not turn on the flag. Therefore, B will receive an `AUTH` message without the pull mode flag from an old overlay version, so B will refuse to connect.

marta-lokhova: that's because the "old version" really is running the new code from this PR, so it creates a discrepancy. In reality, old version will set the flag either way. This can be addressed by changing the condition in sendAuth to `if (mRemoteOverlayVersion < Peer::FIRST_VERSION_REQUIRING_PULL_MODE) || mOverlayVersion < Peer::FIRST_VERSION_REQUIRING_PULL_MODE`

marta-lokhova: I don't think it's reproducing old version behavior, as this condition is basically a no-op during normal operation. It makes things a bit simpler since we'll just request pull mode always by default, then in the macro you can disable it if needed.

hidenori-shinohara: Done

src/overlay/Peer.cpp
marta-lokhova: Thanks!

src/overlay/Peer.h
marta-lokhova: Thanks!

marta-lokhova (approved):
Thank you, this looks good to me

marta-lokhova:
@hidenori-shinohara looks like the commits needs to be squashed and there's a merge conflict

hidenori-shinohara:
@marta-lokhova I cleaned up the commit history & resolved the merge conflicts!

MonsieurNicolas:
I fixed a small confusing change as to avoid blocking

MonsieurNicolas:
r+ 1fe08bbe41f660ee5aacc2fb157a38d10444f9dc

PR #3594, opened 2022-11-08
Flow control capacity accounting

### LINKED ISSUE #3589: flow control: ensure capacity accounting is correct during auth
Currently, there's a race condition in flow control logic during auth. We process some messages sent during auth asynchronously (e.g. `PEERS`), which means that we can start processing the message while flow control is off (no capacity decrement occurs), but finish processing it when flow control is on (so we increment the capacity even though we didn’t decrement it).

hidenori-shinohara:
# Description

Resolves #3589 

As outlined in #3589, the race condition occurs when we start processing incoming messages before we turn on flow control and end processing them after we turn it on. This PR assumes that flow control is on always as all Overlay versions that meet the min overlay requirement enables pull mode on by default. 

- The first commit adds asserts to prevent this said behavior. 
- The second commit removes code for not turning on flow control, effectively fixing the bug.
- The third commit removes the constant specifying the minimum flow control version as it's already lower than the min overlay version. Furthermore, we remove a test that uses that as that test fails due to not meeting the min overlay version.

# Checklist
- [x] Compiles
- [x] Ran all tests

### DIFF
--- a/src/overlay/Peer.cpp
+++ b/src/overlay/Peer.cpp
@@ -65,7 +65,6 @@ Peer::Peer(Application& app, PeerRole role)
           std::make_optional<VirtualClock::time_point>(app.getClock().now()))
     , mEnqueueTimeOfLastWrite(app.getClock().now())
     , mPeerMetrics(app.getClock().now())
-    , mFlowControlState(Peer::FlowControlState::DONT_KNOW)
     , mCapacity{app.getConfig().PEER_FLOOD_READING_CAPACITY,
                 app.getConfig().PEER_READING_CAPACITY}
     , mTxAdvertQueue(app)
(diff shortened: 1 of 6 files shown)


src/overlay/test/OverlayTests.cpp
```diff
@@ -1558,13 +1558,16 @@ TEST_CASE("overlay flow control", "[overlay][flowcontrol]")
 
     auto configs = std::vector<Config>{};
 
+    uint32_t const floodReadingCapacity = 1;
```
marta-lokhova, line 1575: This commit is not needed because of the asserts in the second commit (so the test should fail regardless)

src/overlay/Peer.cpp
```diff
@@ -1820,6 +1826,11 @@ Peer::recvHello(Hello const& elo)
         return;
     }
 
+    if (mRemoteOverlayVersion >= Peer::FIRST_VERSION_REQUIRING_FLOW_CONTROL)
```
marta-lokhova, line 1829: Not following the reasoning for this if condition: remote version must be at least min version. min version right now is 23, so the condition should always be true.

src/overlay/Peer.cpp
```diff
@@ -852,7 +852,13 @@ Peer::recvMessage(StellarMessage const& stellarMsg)
             return;
         }
 
-        mFlowControlState = Peer::FlowControlState::ENABLED;
+        if (mFlowControlState != Peer::FlowControlState::ENABLED)
```
marta-lokhova, line 855: not following why this if condition is here

marta-lokhova (changes requested):

src/overlay/test/OverlayTests.cpp
hidenori-shinohara: Removed.

src/overlay/Peer.cpp
hidenori-shinohara: I removed some logic around the version as the minimum version must enable flow control by default. I removed this code during the cleanup.

src/overlay/Peer.cpp
hidenori-shinohara: Removed.

src/overlay/test/OverlayTests.cpp
```diff
@@ -1677,17 +1647,6 @@ TEST_CASE("overlay flow control", "[overlay][flowcontrol]")
                 3 * Herder::EXP_LEDGER_TIMESPAN_SECONDS, false),
             std::runtime_error);
     }
-    SECTION("one peer doesn't support flow control")
```
marta-lokhova, line 1680: Why is this removed?

src/overlay/test/OverlayTests.cpp
```diff
@@ -189,27 +181,20 @@ TEST_CASE("loopback peer flow control activation", "[overlay][flowcontrol]")
...
         }
     }
     SECTION("one disables")
     {
         // Peer tries to disable flow control during auth
         // Set capacity to 0 so that the peer sends SEND_MORE with numMessages=0
         cfg2.PEER_FLOOD_READING_CAPACITY = 0;
-        runTest({cfg1, cfg2}, Peer::FlowControlState::DISABLED);
-    }
-    SECTION("one does not support")
```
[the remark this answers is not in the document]
marta-lokhova: This test needs to fail with "wrong protocol version"

src/overlay/Peer.cpp
```diff
@@ -991,8 +952,6 @@ Peer::recvSendMore(StellarMessage const& msg)
 {
     ZoneScoped;
     // Flow control must be "on"
-    auto fc = flowControlEnabled();
```
marta-lokhova, line 994: comment for the removed lines is still there

marta-lokhova (changes requested):

src/overlay/test/OverlayTests.cpp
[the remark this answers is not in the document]
hidenori-shinohara: I removed it since (If I'm not mistaken!) the original intent of this test is no longer relevant. This test was about a peer which doesn't support flow control but passes the min overlay version check. But such a case no longer exists as the min overlay version supports flow control. And we have a test called `"reject peers with incompatible overlay versions"` about the case when the peer's version doesn't meet the min overlay version requirement and this test indeed fails with `wrong protocol version`

marta-lokhova (approved):
This looks good now. Need to squash commits though.

hidenori-shinohara:
Thanks for reviewing! Just squashed commits!

marta-lokhova:
we should merge this to unblock other changes (like #3607). @MonsieurNicolas I approved this change already, could you take a look?

MonsieurNicolas:
r+ eb7ea67e2b9c0fda2f67772bd5982291efa7e416

PR #3595, opened 2022-11-09
Bump vs 2022 project files and postgres (Windows)

MonsieurNicolas:
While setting up a new machine I discovered that some dependencies were very stale for no good reason:
* We now use VS2022 -> we should only depend on toolchains that are maintained with VS2022 to make it easier to setup. When libsodium releases a new stable version we can point to the updated project files (but I didn't want to rev libsodium just for that)
* PG 9.5: latest stable is 15

### DIFF
--- a/Builds/VisualStudio/stellar-core.vcxproj
+++ b/Builds/VisualStudio/stellar-core.vcxproj
@@ -124,8 +124,8 @@
       <WarningLevel>Level4</WarningLevel>
       <Optimization>Disabled</Optimization>
       <SDLCheck>true</SDLCheck>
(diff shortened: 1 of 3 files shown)


Builds/VisualStudio/stellar-core.vcxproj
```diff
@@ -124,8 +124,8 @@
...
-      <AdditionalIncludeDirectories>src;../../src;../../lib;../../lib/tracy;../../lib/spdlog/include;../../lib/libmedida/src;../../lib/soci/src/core;../../lib/autocheck/include;../../lib/cereal/include;../../lib/asio/asio/include;../../lib/xdrpp;../../lib/libsodium/src/libsodium/include;../../lib/fmt/include;../..;src/$(Configuration)/generated;../../lib/sqlite;c:\Program Files\PostgreSQL\9.5\include;%(AdditionalIncludeDirectories)</AdditionalIncludeDirectories>
+      <PreprocessorDefinitions>USE_POSTGRES;ENABLE_NEXT_PROTOCOL_VERSION_UNSAFE_FOR_PRODUCTION=1;USE_SPDLOG;FMT_HEADER_ONLY=1;BUILD_TESTS;WIN32_LEAN_AND_MEAN;NOMINMAX;ASIO_STANDALONE;_WINSOCK_DEPRECATED_NO_WARNINGS;SODIUM_STATIC;ASIO_SEPARATE_COMPILATION;ASIO_ERROR_CATEGORY_NOEXCEPT=noexcept;TRACY_ENABLE;TRACY_ON_DEMAND;TRACY_NO_BROADCAST;TRACY_ONLY_LOCALHOST;USE_TRACY;_CRT_SECURE_NO_WARNINGS;_WIN32_WINNT=0x0601;WIN32;_MBCS;_CRT_NONSTDC_NO_DEPRECATE;YY_NO_UNISTD_H;%(PreprocessorDefinitions)</PreprocessorDefinitions>
```
dmkozh, line 127: Seems like a flex def,  I'm curious where did it come from?

dmkozh (approved):

Builds/VisualStudio/stellar-core.vcxproj
MonsieurNicolas: interestingly, we were picking up that def from the old postgres SDK (and they stopped doing that)

dmkozh: Oh, I see. I thought that's some breakage caused by xdr query... Good to know

MonsieurNicolas:
r+ bd25f37f55a6005472db0c0b1e0ad383447270a3

PR #3598, opened 2022-11-15
Improvements to overlay metrics and logging

marta-lokhova:


src/overlay/Peer.cpp
```diff
@@ -137,6 +141,39 @@ Peer::MsgCapacityTracker::getPeer()
     return mWeakPeer;
 }
 
+bool
+Peer::isPeerSlow()
```
hidenori-shinohara, line 145: Nice!!! 👍

src/herder/TransactionQueue.cpp
```diff
@@ -658,6 +661,8 @@ TransactionQueue::removeApplied(Transactions const& appliedTxs)
     {
         auto& bannedFront = mBannedTransactions.front();
         bannedFront.emplace(h);
+        CLOG_DEBUG(Overlay, "Ban transaction {}", hexAbbrev(h));
```
hidenori-shinohara, line 664: nit: what do you think about specifying that it's applied, like "Ban applied transaction {}"?

hidenori-shinohara (commented):
I had a few comments!

src/herder/TransactionQueue.cpp
marta-lokhova: Updated!

src/overlay/OverlayManagerImpl.cpp
```diff
@@ -1171,14 +1172,48 @@ OverlayManagerImpl::getMaxDemandSize() const
...
+                Overlay, "Pulled transaction {} in {} milliseconds from peer",
+                hexAbbrev(hash),
+                std::chrono::duration_cast<std::chrono::milliseconds>(val)
+                    .count(),
+                peer->toString());
+        }
+        else
+        {
+            CLOG_WARNING(Overlay, "Received tx {} from peer we did not ask",
+                         hexAbbrev(hash));
```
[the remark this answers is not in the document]
marta-lokhova: Yeah I ended up removing that log message as it's not that useful anyway

src/overlay/OverlayManagerImpl.cpp
```diff
@@ -1171,14 +1172,48 @@ OverlayManagerImpl::getMaxDemandSize() const
...
+        // Record pull time from individual peer
+        auto peerIt = it->second.peers.find(peer->getPeerID());
+        if (peerIt != it->second.peers.end())
+        {
+            auto val = now - peerIt->second;
+            peer->updateLatencyBasedMetrics(
+                mOverlayMetrics.mPeerTxPullLatency,
+                mOverlayMetrics.mPeerTxPullLatencySlow, val);
+            CLOG_DEBUG(
+                Overlay, "Pulled transaction {} in {} milliseconds from peer",
```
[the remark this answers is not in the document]
marta-lokhova: good catch, fixed

src/overlay/OverlayMetrics.cpp
```diff
@@ -144,8 +166,12 @@ OverlayMetrics::OverlayMetrics(Application& app)
           {"overlay", "flood", "unfulfilled-banned"}, "message"))
     , mUnknownMessageUnfulfilledMeter(app.getMetrics().NewMeter(
           {"overlay", "flood", "unfulfilled-unknown"}, "message"))
-    , mTxPullLatency(
-          app.getMetrics().NewTimer({"overlay", "flood", "tx-pull-latency"}))
+    , mTotalTxPullLatency(
+          app.getMetrics().NewTimer({"overlay", "flood", "total-pull-latency"}))
```
MonsieurNicolas, line 170: I don't see a reason to rename the metric (this breaks the dashboard)

src/overlay/OverlayMetrics.cpp
```diff
@@ -83,6 +83,10 @@ OverlayMetrics::OverlayMetrics(Application& app)
           app.getMetrics().NewTimer({"overlay", "delay", "write-queue"}))
     , mMessageDelayInAsyncWriteTimer(
           app.getMetrics().NewTimer({"overlay", "delay", "async-write"}))
+    , mMessageDelayInWriteQueueTimerSlow(
+          app.getMetrics().NewTimer({"overlay", "delay", "write-queue-slow"}))
+    , mMessageDelayInAsyncWriteTimerSlow(
```
MonsieurNicolas, line 88: I think that we'll evolve over time how we want to categorize peers. I think that "quality" is probably the dimension we're going after (deriving for now from "latency") so the suffix should be `low` (for low quality).

Should there be, `foo` (for all), `foo-high` (for high) and `foo-low` (for low) ? I am a bit worried that if anything is wrong with quality we end up with unusable metrics as we can't reconstruct "all".

src/overlay/Peer.cpp
```diff
@@ -137,6 +141,39 @@ Peer::MsgCapacityTracker::getPeer()
...
+{
+    auto now = mApp.getClock().now();
```
MonsieurNicolas, line 147: I don't think you want to categorize peers like this:
you probably want to categorize peers all at the same time as to "snapshot" their quality (so maybe from the "tick" callback?).

src/overlay/Peer.cpp
```diff
@@ -137,6 +141,39 @@ Peer::MsgCapacityTracker::getPeer()
...
+    if ((now - mLastLatencyCheck) > LATENCY_CHECK_FREQUENCY)
+    {
+        mLatencyReasonable =
+            mPeerMetrics.mConnectionLatency.GetSnapshot().get95thPercentile() <=
```
MonsieurNicolas, line 151: peer latency used like this is going to be pretty bad.
I pulled this from the last few hours:
image

Here are a few things to take away from this:
* latency can get very high - like minutes. This may cause to classify as "good" most peers (I think those peers end up getting dropped by the straggler logic)
* by the time we adjust latency for a peer, if it was "fast" before, it will have contributed a lot of bad data to the "fast" partition
* we should probably take into account `mPingSentTime` so that we can "fast fail" peers that don't respond.
* because you're using medida's window, you will often ignore high latency peers (in any 30 seconds window, you will not observe slow peers every time)
* (maybe related to the previous point) we probably want the classification to be relatively stable over time, ideally a peer that gets stuck for 20 seconds every 5 minutes but is fast the rest of the time should just be "bad" quality all the time (if other peers do not do that)
* maybe "ping latency" is too broken on its own. There are a bunch of other ways that peers can "go bad" (see `recurrentTimerExpired`), so it might be more reliable to aggregate most of those things into a norm of sorts, so that every "tick" we can "credit" badness to peers.

MonsieurNicolas (changes requested):

src/herder/TransactionQueue.cpp
```diff
@@ -658,6 +661,8 @@ TransactionQueue::removeApplied(Transactions const& appliedTxs)
     {
         auto& bannedFront = mBannedTransactions.front();
         bannedFront.emplace(h);
+        CLOG_DEBUG(Overlay, "Ban applied transaction {}", hexAbbrev(h));
```
hidenori-shinohara, line 664: nit: Overlay -> Herder?

src/herder/TransactionQueue.cpp
```diff
@@ -690,6 +695,7 @@ TransactionQueue::ban(Transactions const& banTxs)
     {
         auto& transactions = transactionsByAccount[tx->getSourceID()];
         transactions.emplace_back(tx);
+        CLOG_DEBUG(Overlay, "Ban transaction {}", hexAbbrev(tx->getFullHash()));
```
hidenori-shinohara, line 698: nit: overlay -> herder?

src/simulation/LoadGenerator.cpp
```diff
@@ -368,8 +368,8 @@ LoadGenerator::submitPaymentOrPretendTx(GeneratedLoadConfig const& cfg,
         if (cfg.skipLowFeeTxs &&
             (status ==
                  TransactionQueue::AddResult::ADD_STATUS_TRY_AGAIN_LATER ||
-             status == TransactionQueue::AddResult::ADD_STATUS_ERROR &&
-                 code == txINSUFFICIENT_FEE))
+             (status == TransactionQueue::AddResult::ADD_STATUS_ERROR &&
```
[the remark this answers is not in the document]
marta-lokhova: I get this
```
simulation/LoadGenerator.cpp:371:70: warning: '&&' within '||' [-Wlogical-op-parentheses]
             status == TransactionQueue::AddResult::ADD_STATUS_ERROR &&
             ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^~
simulation/LoadGenerator.cpp:371:70: note: place parentheses around the '&&' expression to silence this warning
             status == TransactionQueue::AddResult::ADD_STATUS_ERROR &&
```

marta-lokhova:
@MonsieurNicolas the updated PR contains a few more metrics. I added demand timeouts back, as I think it can be helpful to see how often we're retrying, as if the rate goes up, it indicates that something is wrong. In addition, @hidenori-shinohara had an interesting idea to partition pulled transactions based on its relevancy to the tx queue - I think it can help evaluate the impact of duplicate pulling on the node.

graydon (approved):
LGTM!

graydon:
r+ 71c87df

PR #3601, opened 2022-11-19
Make install script easier to use on macOS

leighmcculloch:
### What
Move the dependencies that need brew installing onto a single line in a single command.

### Why
The commands are listed one after the other and are tedious to run one by one. The tools can be installed all together in a single command which is more convenient.


MonsieurNicolas:
thank you!

MonsieurNicolas:
r+ 3ce5e32d4be901dc10b1fca3d5821807373891ce

PR #3602, opened 2022-11-21
bump overlay version

marta-lokhova:
Keeping min version as-is, per 3-month backward compatibility policy


MonsieurNicolas:
r+ 1e0c67ab100022eb5dd0c00a023a17fd74ab72bc

PR #3607, opened 2022-11-22
Flow control in bytes

marta-lokhova:
Updates to switch flow control capacity accounting to bytes instead of message counts. This allows us to support larger transactions (i.e. Soroban) better.

### DIFF
--- a/src/main/Config.cpp
+++ b/src/main/Config.cpp
@@ -130,7 +130,7 @@ Config::Config() : NODE_SEED(SecretKey::random())
     MAXIMUM_LEDGER_CLOSETIME_DRIFT = 50;
 
     OVERLAY_PROTOCOL_MIN_VERSION = 24;
-    OVERLAY_PROTOCOL_VERSION = 27;
+    OVERLAY_PROTOCOL_VERSION = 28;
 
     VERSION_STR = STELLAR_CORE_VERSION;
 
@@ -214,6 +214,9 @@ Config::Config() : NODE_SEED(SecretKey::random())
     PEER_READING_CAPACITY = 200;
     PEER_FLOOD_READING_CAPACITY = 200;
     FLOW_CONTROL_SEND_MORE_BATCH_SIZE = 40;
+
+    PEER_FLOOD_READING_CAPACITY_BYTES = 100000;
+    FLOW_CONTROL_SEND_MORE_BATCH_SIZE_BYTES = 20000;
 
     // WORKER_THREADS: setting this too low risks a form of priority inversion
     // where a long-running background task occupies all worker threads and
@@ -936,11 +939,24 @@ Config::processConfig(std::shared_ptr<cpptoml::table> t)
 
             if (item.first == "PEER_READING_CAPACITY")
             {
-                PEER_READING_CAPACITY = readInt<uint32_t>(item, 2);
+                PEER_READING_CAPACITY = readInt<uint32_t>(item, 1);
             }
             else if (item.first == "PEER_FLOOD_READING_CAPACITY")
             {
                 PEER_FLOOD_READING_CAPACITY = readInt<uint32_t>(item, 1);
+            }
+            else if (item.first == "FLOW_CONTROL_SEND_MORE_BATCH_SIZE")
+            {
+                FLOW_CONTROL_SEND_MORE_BATCH_SIZE = readInt<uint32_t>(item, 1);
+            }
+            else if (item.first == "PEER_FLOOD_READING_CAPACITY_BYTES")
+            {
+                PEER_FLOOD_READING_CAPACITY_BYTES = readInt<uint32_t>(item, 1);
+            }
+            else if (item.first == "FLOW_CONTROL_SEND_MORE_BATCH_SIZE_BYTES")
+            {
+                FLOW_CONTROL_SEND_MORE_BATCH_SIZE_BYTES =
+                    readInt<uint32_t>(item, 1);
             }
             else if (item.first == "PEER_PORT")
             {
@@ -1366,10 +1382,6 @@ Config::processConfig(std::shared_ptr<cpptoml::table> t)
                 ARTIFICIALLY_SLEEP_MAIN_THREAD_FOR_TESTING =
                     std::chrono::microseconds(readInt<uint32_t>(item));
             }
-            else if (item.first == "FLOW_CONTROL_SEND_MORE_BATCH_SIZE")
-            {
-                FLOW_CONTROL_SEND_MORE_BATCH_SIZE = readInt<uint32_t>(item, 1);
-            }
             else if (item.first == "MAX_DEX_TX_OPERATIONS_IN_TX_SET")
             {
                 auto value = readInt<uint32_t>(item);
@@ -1403,6 +1415,16 @@ Config::processConfig(std::shared_ptr<cpptoml::table> t)
             std::string msg =
                 "Invalid configuration: FLOW_CONTROL_SEND_MORE_BATCH_SIZE "
                 "can't be greater than PEER_FLOOD_READING_CAPACITY";
+            throw std::runtime_error(msg);
+        }
+
+        if (FLOW_CONTROL_SEND_MORE_BATCH_SIZE_BYTES >
+            PEER_FLOOD_READING_CAPACITY_BYTES)
+        {
+            std::string msg =
+                "Invalid configuration: "
+                "FLOW_CONTROL_SEND_MORE_BATCH_SIZE_BYTES "
+                "can't be greater than PEER_FLOOD_READING_CAPACITY_BYTES";
             throw std::runtime_error(msg);
         }
 
--- a/src/main/Config.cpp
+++ b/src/main/Config.cpp
@@ -1428,6 +1428,18 @@ Config::processConfig(std::shared_ptr<cpptoml::table> t)
             throw std::runtime_error(msg);
         }
 
+        if (!(PEER_FLOOD_READING_CAPACITY_BYTES -
+                  FLOW_CONTROL_SEND_MORE_BATCH_SIZE_BYTES >=
+              FlowControlCapacity::MAX_FLOOD_MESSAGE_SIZE_BYTES))
+        {
+            std::string msg =
+                "Invalid configuration: the difference between "
+                "PEER_FLOOD_READING_CAPACITY_BYTES and "
+                "FLOW_CONTROL_SEND_MORE_BATCH_SIZE_BYTES must be at least "
+                "(MAX_FLOOD_MESSAGE_SIZE_BYTES)";
+            throw std::runtime_error(msg);
+        }
+
         verifyLoadGenOpCountForTestingConfigs();
 
         gIsProductionNetwork = NETWORK_PASSPHRASE ==
--- a/src/main/Config.cpp
+++ b/src/main/Config.cpp
@@ -52,6 +52,8 @@ static const std::unordered_set<std::string> TESTING_ONLY_OPTIONS = {
     "OP_APPLY_SLEEP_TIME_WEIGHT_FOR_TESTING",
     "LOADGEN_OP_COUNT_FOR_TESTING",
     "LOADGEN_OP_COUNT_DISTRIBUTION_FOR_TESTING",
+    "LOADGEN_OP_SIZE_FOR_TESTING",
+    "LOADGEN_OP_SIZE_DISTRIBUTION_FOR_TESTING",
     "CATCHUP_WAIT_MERGES_TX_APPLY_FOR_TESTING",
     "ARTIFICIALLY_DELAY_BUCKET_APPLICATION_FOR_TESTING",
     "ARTIFICIALLY_SLEEP_MAIN_THREAD_FOR_TESTING",
@@ -119,6 +121,8 @@ Config::Config() : NODE_SEED(SecretKey::random())
     OP_APPLY_SLEEP_TIME_WEIGHT_FOR_TESTING = std::vector<uint32>();
     LOADGEN_OP_COUNT_FOR_TESTING = {};
     LOADGEN_OP_COUNT_DISTRIBUTION_FOR_TESTING = {};
+    LOADGEN_OP_SIZE_FOR_TESTING = {};
+    LOADGEN_OP_SIZE_DISTRIBUTION_FOR_TESTING = {};
     CATCHUP_WAIT_MERGES_TX_APPLY_FOR_TESTING = false;
     ARTIFICIALLY_SLEEP_MAIN_THREAD_FOR_TESTING =
         std::chrono::microseconds::zero();
@@ -824,6 +828,20 @@ Config::verifyLoadGenOpCountForTestingConfigs()
                                     "must have the exact same size.");
     }
 
+    if (LOADGEN_OP_SIZE_FOR_TESTING.size() !=
+        LOADGEN_OP_SIZE_DISTRIBUTION_FOR_TESTING.size())
+    {
+        throw std::invalid_argument("LOADGEN_OP_SIZE_FOR_TESTING and "
+                                    "LOADGEN_OP_SIZE_DISTRIBUTION_FOR_TESTING "
+                                    "must be defined together and "
+                                    "must have the exact same size.");
+    }
+
+    if (LOADGEN_OP_SIZE_FOR_TESTING.empty())
+    {
+        return;
+    }
+
     if (LOADGEN_OP_COUNT_FOR_TESTING.empty())
     {
         return;
@@ -1355,6 +1373,15 @@ Config::processConfig(std::shared_ptr<cpptoml::table> t)
             else if (item.first == "LOADGEN_OP_COUNT_DISTRIBUTION_FOR_TESTING")
             {
                 LOADGEN_OP_COUNT_DISTRIBUTION_FOR_TESTING =
+                    readIntArray<uint32>(item);
+            }
+            else if (item.first == "LOADGEN_OP_SIZE_FOR_TESTING")
+            {
+                LOADGEN_OP_SIZE_FOR_TESTING = readIntArray<uint32>(item);
+            }
+            else if (item.first == "LOADGEN_OP_SIZE_DISTRIBUTION_FOR_TESTING")
+            {
+                LOADGEN_OP_SIZE_DISTRIBUTION_FOR_TESTING =
                     readIntArray<uint32>(item);
             }
             else if (item.first == "CATCHUP_WAIT_MERGES_TX_APPLY_FOR_TESTING")
--- a/src/main/Config.cpp
+++ b/src/main/Config.cpp
@@ -221,6 +221,7 @@ Config::Config() : NODE_SEED(SecretKey::random())
 
     PEER_FLOOD_READING_CAPACITY_BYTES = 100000;
     FLOW_CONTROL_SEND_MORE_BATCH_SIZE_BYTES = 20000;
+    OUTBOUND_TX_QUEUE_BYTE_LIMIT = 1024 * 1024 * 3;
 
     // WORKER_THREADS: setting this too low risks a form of priority inversion
     // where a long-running background task occupies all worker threads and
@@ -975,6 +976,10 @@ Config::processConfig(std::shared_ptr<cpptoml::table> t)
             {
                 FLOW_CONTROL_SEND_MORE_BATCH_SIZE_BYTES =
                     readInt<uint32_t>(item, 1);
+            }
+            else if (item.first == "OUTBOUND_TX_QUEUE_BYTE_LIMIT")
+            {
+                OUTBOUND_TX_QUEUE_BYTE_LIMIT = readInt<uint32_t>(item, 1);
             }
             else if (item.first == "PEER_PORT")
             {
--- a/src/main/Config.h
+++ b/src/main/Config.h
@@ -248,6 +248,14 @@ class Config : public std::enable_shared_from_this<Config>
     // processes `FLOW_CONTROL_SEND_MORE_BATCH_SIZE` messages
     uint32_t FLOW_CONTROL_SEND_MORE_BATCH_SIZE;
 
+    // A config parameter that controls how many bytes worth of flood messages
+    // (tx or SCP) from a particular peer core can process simultaneously
+    uint32_t PEER_FLOOD_READING_CAPACITY_BYTES;
+
+    // When flow control is enabled, peer asks for more data every time it
+    // processes `FLOW_CONTROL_SEND_MORE_BATCH_SIZE_BYTES` bytes
+    uint32_t FLOW_CONTROL_SEND_MORE_BATCH_SIZE_BYTES;
+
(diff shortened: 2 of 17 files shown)


src/protocol-curr/xdr/Stellar-overlay.x
```diff
@@ -22,9 +22,37 @@ struct Error
     string msg<100>;
 };
 
-struct SendMore
+enum SendMoreType
```
marta-lokhova, line 25: Heads up, this change isn't right - we need to still support old-style XDR for backward compatibility

src/overlay/Peer.cpp
```diff
@@ -957,15 +976,35 @@ Peer::sendSendMore(uint32_t numMessages)
...
+
 void
 Peer::recvSendMore(StellarMessage const& msg)
 {
     ZoneScoped;
-    releaseAssert(msg.sendMoreMessage().numMessages > 0);
+    releaseAssert(getNumMessages(msg) > 0);
+    if (mFlowControlBytes)
+    {
+        releaseAssert(msg.type() == SEND_MORE_EXTENDED);
```
MonsieurNicolas, line 1007: this is wrong: if a peer sends us the wrong message type/bad data, we should not crash but drop it instead

src/overlay/Peer.cpp
```diff
@@ -1163,50 +1224,65 @@ Peer::maybeSendNextBatch()
...
+    if (shouldAbort() || !mFlowControlMessages)
     {
         return;
     }
 
-    mCapacity.mTotalCapacity++;
-    if (mApp.getOverlayManager().isFloodMessage(msg))
+    bool isFloodMessage = mApp.getOverlayManager().isFloodMessage(msg);
+
+    mFlowControlMessages->releaseLocalCapacity(msg);
```
MonsieurNicolas, line 1269: maybe `release` should return the amount of capacity released? That way you don't need to have this tight coupling between whatever criteria is used internally (like if it's a flood message) and the call site.

src/overlay/Peer.cpp
MonsieurNicolas: also the opposite is true (missing else): we should not allow a peer to send us `SEND_MORE_EXTENDED` if we didn't enable flow control in bytes

src/overlay/Peer.cpp
```diff
@@ -50,6 +50,14 @@ using namespace soci;
 
 static constexpr VirtualClock::time_point PING_NOT_SENT =
     VirtualClock::time_point::min();
+static constexpr size_t OUTBOUND_TX_QUEUE_BYTE_LIMIT =
+    1024 * 1024 * 10; // 10 MB
```
MonsieurNicolas, line 54: I am having a hard time understanding how we'd set this limit:
* is "10 MB" too high or too low?
* other limit is expressed relative to ledger size, which is easy to understand
* there is a lower bound that comes from business requirements
* the limit itself is trying to control for latency? something else?

Right now that limit *seems* too high, but I don't fully understand what we're trying to do yet and if a simple limit like this is enough.

MonsieurNicolas (changes requested):

src/overlay/FlowControlCapacity.cpp
```diff
@@ -0,0 +1,141 @@
...
+
+bool
+FlowControlCapacity::lockLocalCapacity(StellarMessage const& msg)
+{
+    releaseAssert(getCapacityLimits().mFloodCapacity >=
+                  mCapacity.mFloodCapacity);
+    releaseAssert(getCapacityLimits().mTotalCapacity >=
+                  mCapacity.mTotalCapacity);
+    releaseAssert(mCapacity.mTotalCapacity > 0);
+    mCapacity.mTotalCapacity -= getMsgResourceCount(msg);
```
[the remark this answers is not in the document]
marta-lokhova: Addressing both this and the other comment around `tryLock`: I don't believe the proposed approach is correct. I intentionally implemented it this way, because right now messages take up arbitrary capacity that the receiver does not know ahead of time.

Consider a simple scenario: the receiver sets its batch size to 100 bytes, and asks the sender for 100 bytes worth of traffic. The sender has the next SCP message to send of size 300 bytes. Unless we allow capacity to temporarily go negative, the system deadlocks. To unblock it, we allow going past capacity for up to (message_size - 1).

Note that we wouldn't need to do this if we implemented flow control based on uniformly sized batches (no need to go negative if we know max batch size ahead of time), but I'm not doing it in this iteration because (a) benefits of this approach are unclear and (b) seems like it'll add more complexity to the code since we'll need to deconstruct/reconstruct messages, reconcile when to request more data based on complete messages vs batches received, etc

src/overlay/Peer.cpp
marta-lokhova: The limit here is related to memory DoS:
* Not sure how we can reason about ledger size here, because we don't know transaction size distribution per ledger. 1000 ops of 300 bytes is very different from 1000 ops of 128kb.
* I'm assuming transactions of arbitrary sizes can end up in this queue. Because the variance is too big, instead I decided to put a limit that is high enough in a sense that it'll still likely allow most/all transactions, but in some degenerate cases we don't blow up our memory (large txs add up fast with the current limit of 1000 - 128kb*1000=128MB per just one peer).
* I'm not sure it's too high - we're still enforcing 1000 txs per queue limit, this limit is just to prevent the queue from growing too big.

src/overlay/Peer.cpp
```diff
@@ -822,22 +826,38 @@ Peer::recvMessage(StellarMessage const& stellarMsg)
...
             drop("unexpected SEND_MORE message",
                  Peer::DropDirection::WE_DROPPED_REMOTE,
                  Peer::DropMode::IGNORE_WRITE_QUEUE);
             return;
         }
 
-        if (stellarMsg.sendMoreMessage().numMessages >
-            UINT64_MAX - mOutboundCapacity)
+        auto overflow =
+            mFlowControlMessages->getOutboundCapacity() > 0 &&
```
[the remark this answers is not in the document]
marta-lokhova: We don't, message outbound capacity is guaranteed to always be positive (I initially made this change when I made outbound capacity int64 instead of uint64, but forgot to remove it). Removed it.

src/overlay/Peer.cpp
marta-lokhova: Good catch, both cases are handled in `Peer::recvMessage(StellarMessage const& stellarMsg)` now where we perform other validity checks, so this assert can stay. Added test coverage as well.

src/overlay/Peer.cpp
marta-lokhova: Good idea, done.

src/overlay/FlowControlCapacity.cpp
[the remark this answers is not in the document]
MonsieurNicolas: I am not sure I understand the deadlock comment, that sounds more like a limitation of the design that I imagine we can overcome? In your example, we could just wait to receive more "send more" messages until there is more than 300?
Accounting in bytes or in terms of messages (where we already send multiple messages until we use up capacity) should behave the same way: if a peer sends us too much data, that means it's not respecting flow control, right?

Reason I flagged this in the first place is that if you have capacity left at 100 bytes we allow to receive a 100KB message (or whatever the max message size is) but we would block after receiving a small (200 bytes) transaction which means we'd allow ~2x the configured quota for large messages.

src/overlay/Peer.cpp
MonsieurNicolas: My concern here is that this limit is fine for protecting against out of memory DoS, but it may still trigger a lot of badness: I think what we discussed in a different context was maybe moving to latency based queue policies that would avoid creating "spiral of death" situations. 10MB seems like a lot of data to move especially if there are a lot of connections competing against each other (100 peers type of situation -> 1GB of stuff queued up to be sent)

src/overlay/FlowControlCapacity.cpp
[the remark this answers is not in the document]
marta-lokhova: > I am not sure I understand the deadlock comment, that sounds more like a limitation of the design that I imagine we can overcome? In your example, we could just wait to receive more "send more" messages until there is more than 300?

Send more is triggered when capacity is released. This means that the node needs to receive a 100 bytes worth of data before requesting more data. If it doesn't do it and requests more data arbitrarily, then that defeats the purpose of flow control. I'm not sure I understand what you're proposing, but I think it's the same thing as the current implementation: at the end, node receives 300 bytes when it only has 100 bytes of capacity to spare. This limitation comes from the fact that we transmit full messages, and we don't know their sizes ahead of time.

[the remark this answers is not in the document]
marta-lokhova: I looked into this further after our chat, and even sketched the change, bit still found several issues:
* We can't apply "capacity is always positive" reasoning to reading capacity as we don't know what non-flood traffic a peer is sending us and it's not controlled with SEND_MORE messages (we only define max message size for flood traffic, like transactions).
* I still find it hard to reason about the proposal and the relationship between capacity and batch size - specifically, I can't tell if this logic will actually unblock nodes in all possible situations, with all possible combinations of messages in the queue. I find it easier to reason about the "capacity goes negative" approach as we we allow at most one more message to be sent to unblock the node. 
* With this change, we're re-purposing the batch flag as now there are strict requirements for it vs capacity. I don't really like that we're taking a flag that was only meant to be there to improve perf and make it part of correctness requirements. I think this will also cause confusion for the operators if they decide to change their configs. I think it'll be easier to communicate that whatever capacity they pick implies up to (capacity + max_message_size - 1) max capacity.

Let me know what you think, but I'd like to re-visit the currently implemented approach, as it seems easier to reason about

src/protocol-curr/xdr/Stellar-overlay.x
```diff
@@ -27,6 +27,12 @@ struct SendMore
     uint32 numMessages;
 };
 
+struct SendMoreExtended
+{
+    uint32 numMessages;
+    uint32 numBytes;
```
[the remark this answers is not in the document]
marta-lokhova: This is on purpose, as we rely on peer versions in overlay and drop support for old versions eventually, so with a new message type, it's easy to deprecate old messages.

src/overlay/Peer.cpp
marta-lokhova: I'd like to move away from reasoning that ~100 peers is the norm. We have work in flight that actively tries to reduce this number by a lot. Even without it, a more realistic number based on pubnet data is ~50-60. That being said, I agree that we may still create too much overhead. 

What I was thinking we could do is use the configs for "max classic ops" vs "max Soroban ops" per ledger to calculate this limit. The reason I didn't do it is because we don't have those configs yet, and we don't even know what the values would be. Hence the static limit for now. I think we can change it in the next iteration once we have a better idea about what the distribution is going to be like. In the meantime, I'll reduce the limit to something reasonable, like 3MB.

MonsieurNicolas:
r+ 4b726190cd9a538fd7c1f2fcaee8b68076deff3c

src/overlay/FlowControlCapacity.cpp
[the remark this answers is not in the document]
MonsieurNicolas: > We can't apply "capacity is always positive" reasoning to reading capacity as we don't know what non-flood traffic a peer is sending us and it's not controlled with SEND_MORE messages (we only define max message size for flood traffic, like transactions).

Oh I see, I missed this - what is the point of the global in bytes reading capacity?
I think the reason we want to track in bytes is to avoid network congestion, and this does not contribute to this.
This one is more for DoS protection then right? If so, we should just have enough capacity for the largest message possible with some buffer on top - so something in the order of `MAX_MESSAGE_SIZE*5` (right now `MAX_MESSAGE_SIZE` is 16MB btw).

> I still find it hard to reason about the proposal and the relationship between capacity and batch size - specifically, I can't tell if this logic will actually unblock nodes in all possible situations, with all possible combinations of messages in the queue. I find it easier to reason about the "capacity goes negative" approach as we we allow at most one more message to be sent to unblock the node.

We went over this offline already.

I actually find it harder to reason when we go negative because we have implicit limits.
In the current version for example, you have `PEER_READING_CAPACITY_BYTES` of about 100KB but you can read 16MB of data.

Capacity & limits should be explicit so that we can properly reason about resource utilization, right now based on settings it looks like we would only about 100KB of memory per connection but this is way off.

Same thing applies to what is "on the wire": when modeling based on bandwidth numbers, you really want to configure based on the "real" numbers.


> With this change, we're re-purposing the batch flag as now there are strict requirements for it vs capacity. I don't really like that we're taking a flag that was only meant to be there to improve perf and make it part of correctness requirements. I think this will also cause confusion for the operators if they decide to change their configs. I think it'll be easier to communicate that whatever capacity they pick implies up to (capacity + max_message_size - 1) max capacity.

Those settings were always "advanced" in a way, because what we really want to control with them is max latency. I don't think making additional checks on startup to ensure that people don't do bad things is that special.

I agree that we need to make configuration as simple as possible. We went back and forth on how to configure batch size in the original flow control PR (if you remember we had things expressed as a fraction of the capacity at some point).

This is related to my previous comment on what is easier to reason about for people, in the context of controlling latency, I think that expressing that in terms of true bytes makes sense (because if people know the bandwidth they have, they can do the math on how to allocate it).
We could look into better ways to express those settings (but I don't think anybody is really going to try to configure those things).

src/main/Config.cpp
```diff
@@ -934,11 +938,28 @@ Config::processConfig(std::shared_ptr<cpptoml::table> t)
...
             {
-                PEER_READING_CAPACITY = readInt<uint32_t>(item, 2);
+                PEER_READING_CAPACITY = readInt<int32_t>(item, 1);
             }
             else if (item.first == "PEER_FLOOD_READING_CAPACITY")
             {
-                PEER_FLOOD_READING_CAPACITY = readInt<uint32_t>(item, 1);
+                PEER_FLOOD_READING_CAPACITY = readInt<int32_t>(item, 1);
+            }
+            else if (item.first == "PEER_READING_CAPACITY_BYTES")
```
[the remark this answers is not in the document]
MonsieurNicolas: but see the main convo, we may want to pick different parameters exposed in config (and derive those) if this is not easy to configure

src/main/Config.cpp
```diff
@@ -213,6 +213,10 @@ Config::Config() : NODE_SEED(SecretKey::random())
     PEER_FLOOD_READING_CAPACITY = 200;
     FLOW_CONTROL_SEND_MORE_BATCH_SIZE = 40;
```
mbsdf, line 214: are we planning to remove these constants in the future?

src/main/Config.cpp
```diff
@@ -1368,10 +1389,6 @@ Config::processConfig(std::shared_ptr<cpptoml::table> t)
                 ARTIFICIALLY_SLEEP_MAIN_THREAD_FOR_TESTING =
                     std::chrono::microseconds(readInt<uint32_t>(item));
             }
-            else if (item.first == "FLOW_CONTROL_SEND_MORE_BATCH_SIZE")
```
mbsdf, line 1371: +951

src/main/Config.cpp
```diff
@@ -1408,6 +1425,16 @@ Config::processConfig(std::shared_ptr<cpptoml::table> t)
             throw std::runtime_error(msg);
         }
 
+        if (FLOW_CONTROL_SEND_MORE_BATCH_SIZE_BYTES >
+            PEER_FLOOD_READING_CAPACITY_BYTES)
+        {
+            std::string msg =
```
mbsdf, line 1431: +1

src/main/Config.cpp
```diff
@@ -934,11 +938,28 @@ Config::processConfig(std::shared_ptr<cpptoml::table> t)
 
             if (item.first == "PEER_READING_CAPACITY")
             {
-                PEER_READING_CAPACITY = readInt<uint32_t>(item, 2);
+                PEER_READING_CAPACITY = readInt<int32_t>(item, 1);
```
mbsdf, line 941: what is the reason to change it from uint to int?

src/main/Config.h
```diff
@@ -234,15 +234,29 @@ class Config : public std::enable_shared_from_this<Config>
     // core can process simultaneously. If core is at capacity, it temporarily
     // stops reading from a peer until it completes processing of at least one
     // in-flight message
-    uint32_t PEER_READING_CAPACITY;
+    int32_t PEER_READING_CAPACITY;
 
     // A config parameter that controls how many flood messages (tx or SCP) from
     // a particular peer core can process simultaneously
-    uint32_t PEER_FLOOD_READING_CAPACITY;
+    int32_t PEER_FLOOD_READING_CAPACITY;
```
mbsdf, line 241: Can these  have negative values? Or do we want to keep it same as all others, int instead of uint?

src/overlay/FlowControlCapacity.cpp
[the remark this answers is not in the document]
mbsdf: We are implementing traffic shaping but in a different way. I think this as a way of some sort of bandwidth limit on top of message count instead. Message size can be arbitrary and we do not divide messages into multiple iterations.  For example, if capacity size is 100 and we have already received two messages with 40 and 50, and the next message will be 30. Receiver does not know the size beforehand but still need to ask more because there is 10 available capacity; and hence we accept 30 and make it -20 and now stop asking more messages.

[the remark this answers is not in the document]
mbsdf: I believe this PR serves its purpose for now.  For future as an improvement,we can introduce a burst size in addition to capacity size, and that may help us make it better represented.

src/main/Config.h
```diff
@@ -231,6 +231,10 @@ class Config : public std::enable_shared_from_this<Config>
     std::vector<unsigned short> LOADGEN_OP_COUNT_FOR_TESTING;
     std::vector<uint32> LOADGEN_OP_COUNT_DISTRIBUTION_FOR_TESTING;
 
+    // Additional number of bytes to add to every payment operation in loadgen
+    std::vector<uint32> LOADGEN_OP_SIZE_FOR_TESTING;
+    std::vector<uint32> LOADGEN_OP_SIZE_DISTRIBUTION_FOR_TESTING;
+
```
mbsdf, line 237: +1

src/main/Config.h
```diff
@@ -248,6 +252,17 @@ class Config : public std::enable_shared_from_this<Config>
...
 
+    // A config parameter that controls how many bytes worth of flood messages
+    // (tx or SCP) from a particular peer core can process simultaneously
+    uint32_t PEER_FLOOD_READING_CAPACITY_BYTES;
+
+    // When flow control is enabled, peer asks for more data every time it
+    // processes `FLOW_CONTROL_SEND_MORE_BATCH_SIZE_BYTES` bytes
+    uint32_t FLOW_CONTROL_SEND_MORE_BATCH_SIZE_BYTES;
+
+    // For testing only, remove later
```
mbsdf, line 263: +1

src/overlay/FlowControlCapacity.cpp
```diff
@@ -0,0 +1,187 @@
...
+        // No capacity to process flood message
+        if (mCapacity.mFloodCapacity < msgResources)
+        {
+            return false;
+        }
+
+        mCapacity.mFloodCapacity -= msgResources;
+        if (mCapacity.mFloodCapacity == 0)
+        {
+            CLOG_DEBUG(Overlay, "No flood capacity for peer {}",
```
mbsdf, line 71: msg is accepted (returns true) but prints "No flood capacity", you mean no flood capacity left anymore in debug?

src/overlay/FlowControlCapacity.cpp
```diff
@@ -0,0 +1,187 @@
...
+    if (mCapacity.mTotalCapacity)
+    {
+        *mCapacity.mTotalCapacity += resourcesFreed;
+    }
+
+    if (mApp.getOverlayManager().isFloodMessage(msg))
+    {
+        if (mCapacity.mFloodCapacity == 0)
+        {
+            CLOG_DEBUG(Overlay, "Got flood capacity for peer {} ({})",
```
mbsdf, line 94: +1 adding to both total capacity and flood capacity, and debug log when will make flood capacity >=0

src/overlay/FlowControlCapacity.cpp
```diff
@@ -0,0 +1,187 @@
...
+    return 1;
+}
+
+void
+FlowControlCapacityMessages::releaseOutboundCapacity(StellarMessage const& msg)
+{
+    ZoneScoped;
+    releaseAssert(msg.type() == SEND_MORE || msg.type() == SEND_MORE_EXTENDED);
+    auto numMessages = Peer::getNumMessages(msg);
+    if (!hasOutboundCapacity(msg) && numMessages != 0)
```
mbsdf, line 133: do we use hasOutboundCapacity() somewhere else?

src/overlay/FlowControlCapacity.cpp
```diff
@@ -0,0 +1,187 @@
...
+{
+    ZoneScoped;
+    releaseAssert(msg.type() == SEND_MORE_EXTENDED);
+    if (!hasOutboundCapacity(msg) &&
+        (msg.sendMoreExtendedMessage().numBytes != 0))
+    {
+        CLOG_DEBUG(Overlay, "Got outbound capacity for peer {}",
+                   mApp.getConfig().toShortString(mNodeID));
+    }
+    mOutboundCapacity += msg.sendMoreExtendedMessage().numBytes;
```
mbsdf, line 178: Can you add a few lines of comments (or a brief description in PR summary) regarding: sendMoreExtendedMessage + releaseOutbound() both in FlowControlCapacityBytes and FlowControlCapacity

src/overlay/Peer.cpp
```diff
@@ -1160,50 +1276,63 @@ Peer::maybeSendNextBatch()
...
-
-        if (mFloodMsgsProcessed ==
-            mApp.getConfig().FLOW_CONTROL_SEND_MORE_BATCH_SIZE)
+        else
         {
-            sendSendMore(mApp.getConfig().FLOW_CONTROL_SEND_MORE_BATCH_SIZE);
-            mFloodMsgsProcessed = 0;
+            sendSendMore(mFloodDataProcessed);
         }
+        mFloodDataProcessed = 0;
```
marta-lokhova, line 1323: rename to mFloodDataProcessedMessages

src/overlay/FlowControlCapacity.cpp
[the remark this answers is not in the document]
marta-lokhova: I made some changes that should hopefully help unblock this thread:
* Capacity is now always positive, but there's a requirement for capacity vs maximum flood message size (what we discussed offline)
* Byte accounting does not apply to reading capacity (for now). This is for simplicity, as currently maximum message size is really big, so I'd like to think about the best way to protect against DoS with large messages. We can still move this change forward, and add reading capacity incrementally (it's a purely internal change, so can be added separately without a version bump)

mbsdf (commented):
LGTM so far

marta-lokhova:
Closing this PR: the change has been split into #3737 and #3686

PR #3612, opened 2022-11-29
Better bloom filter metrics

### LINKED ISSUE #3604: Investigate impact of bloom filter element underestimation
Currently, whenever we build bloom filters in BucketListDB, we very roughly estimate how many entries we expect to insert into the filter based on the `Bucket` file size. This estimation is not very good and sometimes we can be off by over 20%. 

The bloom filter can be very sensitive to bad estimations. Playing around with a bloom filter parameter calculator [here](https://hur.st/bloomfilter/), underestimating entry count by 1% increses the false positive rate by 7%, and a 5% underestimation increases false positives by 40%. While these numbers are all theoretical, we need to evaluate the actual impact of estimation and find a solution if the estimation is causing significant false positives in practice.

SirTyson:
# Description

Resolves #3604 

This PR provides better metrics and log messages for BucketListDB bloom filters. Previously, it appeared that the bloom filter false positive rate was too high due to element count underestimation. However, this was due to faulty metrics. In practice, the false positive rate is significantly better than our target and underestimation is not an issue. 

To calculate false positive rates previously, we used the number of ledger entries queried. However, one ledger entry query may lead to several bloom filter queries since we have to check multiple levels of the BucketList for an entry. This PR adds an explicit bloom filter query metric to provide more accurate false positive rates and also logs bloom filter setup parameters. 

# Checklist
- [x] Compiles
- [x] Ran all tests

### DIFF
--- a/docs/metrics.md
+++ b/docs/metrics.md
@@ -31,6 +31,7 @@ bucket.batch.objectsadded                | meter     | number of objects added p
 bucket.memory.shared                     | counter   | number of buckets referenced (excluding publish queue)
 bucket.merge-time.level-<X>              | timer     | time to merge two buckets on level <X>
 bucket.snap.merge                        | timer     | time to merge two buckets
+bucketlistDB.bloom.lookups               | meter     | number of bloom filter lookups
 bucketlistDB.bloom.misses                | meter     | number of bloom filter false positives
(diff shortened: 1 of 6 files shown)


MonsieurNicolas:
r+ ecae078d5e66ce91546ce6ef35a23120534e717f

PR #3615, opened 2022-11-30
Enforce trusted catchup

### LINKED ISSUE #3503: Audit and make commands "secure by default" - add "force" options when bypassing safety
Example command is "offline catchup" that is insecure when invoked without the `trusted-checkpoint-hashes` option:
instead we should require catchup to have that option or a force option with a name that tells people that they're about to do something unsafe `--force-unsafe-trust-archives` in this case for example.

marta-lokhova:
Resolves #3503


MonsieurNicolas:
waiting to hear from the Horizon team before we can merge this

marta-lokhova:
pushed a minor change to not require trusted hash when catching up to `current` and log a warning that this is unsafe.

MonsieurNicolas:
I noticed warnings in supercluster, so we'll have to make sure that all systems are updated before we can merge this

mbsdf (approved):

mbsdf:
relevant issue for reference: https://github.com/stellar/stellar-core/pull/3615

marta-lokhova:
Closing this one as stale - re-open when downstream is ready.

PR #3619, opened 2022-12-05
Disable BucketListDB when using in-memory mode

SirTyson:
# Description

Whenever In-memory mode is enabled, the BucketListDB flag is not overriden. This means that the BucketList might still index buckets even when in-memory mode is enabled even though the buckets will never be searched. This PR disables BucketListDB when in-memory mode is set.

# Checklist
- [x] Compiles
- [x] Ran all tests

### DIFF
--- a/src/main/CommandLine.cpp
+++ b/src/main/CommandLine.cpp
@@ -1281,6 +1281,7 @@ run(CommandLineArgs const& args)
                     cfg.MODE_STORES_HISTORY_MISC = false;
                     cfg.MODE_USES_IN_MEMORY_LEDGER = false;
                     cfg.MODE_ENABLES_BUCKETLIST = false;
+                    cfg.EXPERIMENTAL_BUCKETLIST_DB = false;
                     cfg.PREFETCH_BATCH_SIZE = 0;
                 }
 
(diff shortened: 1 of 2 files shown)


src/main/CommandLine.cpp
```diff
@@ -1281,6 +1281,7 @@ run(CommandLineArgs const& args)
                     cfg.MODE_STORES_HISTORY_MISC = false;
                     cfg.MODE_USES_IN_MEMORY_LEDGER = false;
                     cfg.MODE_ENABLES_BUCKETLIST = false;
+                    cfg.EXPERIMENTAL_BUCKETLIST_DB = false;
```
marta-lokhova, line 1284: Should we just have a single free function that determines if this flag should is true? The condition would be `EXPERIMENTAL_BUCKETLIST_DB && !MODE_USES_IN_MEMORY_LEDGER && MODE_ENABLES_BUCKETLIST`. Hard-coding it manually can lead to weird configurations (like the flag is on, but core doesn't enable bucketlist).

marta-lokhova: We should probably update it here as well with a warning to the user: 
https://github.com/stellar/stellar-core/blob/master/src/main/Config.cpp#L980

PR #3621, opened 2022-12-06
Index buckets before adding them to BucketList

### LINKED ISSUE #3605: Index buckets before assuming state in BucketListDB
In `assumeStateWork`, we assume state based on `HistoryArchiveState` before the buckets are indexed, then index the buckets once they have been added to the `BucketList`. This means that for a time, the entire `BucketList` is populated, but none of the buckets are indexed or searchable. We should index `HistoryArchiveState` buckets before calling `assumeState` to ensure that the `BucketList` is never in an invalid, un-indexed state.

SirTyson:
# Description

Resolves #3605. 

This PR enforces better checks around bucket indexing and guarantees that all buckets are indexed before being added to the BucketList.

# Checklist
- [x] Compiles
- [x] Ran all tests

### DIFF
--- a/src/catchup/IndexBucketsWork.h
+++ b/src/catchup/IndexBucketsWork.h
@@ -32,11 +32,14 @@ class IndexBucketsWork : public Work
         bool onAbort() override;
     };
 
+    std::vector<std::shared_ptr<Bucket>> const& mBuckets;
+
     bool mWorkSpawned{false};
     void spawnWork();
 
   public:
-    IndexBucketsWork(Application& app);
+    IndexBucketsWork(Application& app,
+                     std::vector<std::shared_ptr<Bucket>> const& buckets);
 
   protected:
     State doWork() override;
--- a/src/bucket/BucketManager.h
+++ b/src/bucket/BucketManager.h
@@ -248,11 +248,8 @@ class BucketManager : NonMovableOrCopyable
 
     // Assume state from `has` in BucketList: find and attach all buckets in
     // `has`, set current BL. Note: Does not restart merging
-    virtual void assumeState(HistoryArchiveState const& has) = 0;
-
-    // Restart BucketList Merges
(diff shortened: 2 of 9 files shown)


MonsieurNicolas:
@marta-lokhova can you take a look at this one please?

src/catchup/IndexBucketsWork.h
```diff
@@ -32,11 +33,13 @@ class IndexBucketsWork : public Work
...
     };
 
+    HistoryArchiveState const& mHas;
+
     bool mWorkSpawned{false};
     void spawnWork();
 
   public:
-    IndexBucketsWork(Application& app);
+    IndexBucketsWork(Application& app, HistoryArchiveState const& has);
```
marta-lokhova, line 42: should we pass a vector of bucket pointers here instead? Because indexing can take a while and our policy for bucket garbage collection is kind of arbitrary, I think it's safer to keep strong references to buckets you need

SirTyson: Sounds reasonable to me. I'll do a set of pointers instead of a vector though. We already used a set of buckets to avoid duplicate indexing in `IndexBucketsWork`. Having a set as an input parameter means we can get rid of that check.

SirTyson: Wait never mind, it's a set of shared pointers, not Buckets, so we still need the check. A vector is better.

marta-lokhova: sounds good, you probably want to have the vector as a member variable of AssumeStateWork though as it can get reset/restarted asynchronously, losing buckets references with the current change.

marta-lokhova (changes requested):
Couple of nits/safety precautions, otherwise almost ready to merge! Please squash the commits as well

marta-lokhova:
r+ d4bd127d9fc9f99332b0551d867bf0e831f4bc40

PR #3626, opened 2022-12-14
Added BucketListDB init tests and invariant checks

### LINKED ISSUE #3623: Enable buckets DB by default on captive-core
We probably need to add a new config element if we do this to allow people to "opt-out" of this if needed, but the default should be for captive core to not duplicate state when not running "in memory".

(we will eventually deprecate the in-memory option as buckets are needed anyways)

Note:

I don't know if the "in-memory" behavior needs to be handled at all, or if Horizon now uses catchup+run (and if it does, the same edge cases need to be handled there --> it's unclear to me if Horizon just runs `new-db`, which causes the node to forget everything about the network, not just ledger state, in particular overlay information).

It's unclear right now if our test suite covers typical scenarios that can happen with captive-core when in memory mode:
* start from fresh (new-db)
* restart with a starting ledger (`--start-at-ledger`)
    * exactly equal to lcl
    * in the future
    * in the past

SirTyson:
# Description

Resolves #3623.

This PR adds several captive-core and BucketListDB related unit tests. In particular, it tests the BucketListDB startup path and several edge cases for in-memory mode in captive-core. There are also several bug fixes, including fixing BucketList related invariants that were broken when BucketListDB was enabled and some captive-core performance bugs introduced by BucketListDB.

# Checklist
- [x] Compiles
- [x] Ran all tests

### DIFF
--- a/src/bucket/Bucket.cpp
+++ b/src/bucket/Bucket.cpp
@@ -377,7 +377,7 @@ Bucket::fresh(BucketManager& bucketManager, uint32_t protocolVersion,
     }
 
     return out.getBucket(bucketManager,
-                         bucketManager.getConfig().EXPERIMENTAL_BUCKETLIST_DB);
+                         bucketManager.getConfig().isUsingBucketListDB());
 }
 
 static void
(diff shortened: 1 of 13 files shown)


MonsieurNicolas:
r+ 9c1f1b45ffda19a5c9e33c01938264918579e33c

PR #3633, opened 2022-12-19
Improvements to survey script

marta-lokhova:
List of changes, in order of importance:
* Submit maximum 10 survey requests per ledger. This change should help the network drop less survey requests, as the input is throttled. In addition, this change should mitigate premature end of survey due to expiration (example is the script submits 100 requests and on the core side survey expires before getting through all the requests)
* Add Tier1 analysis for graphs augmented with StellarBeat info
* Add reporting of survey progress and nodes we're still waiting for hear from
* Use set for accumulating new survey requests. Surveyed nodes are de-duped in core as well, but there's no need to add additional overhead of duplicate requests on the script side
* Run autopep8 for general formatting consistency

### DIFF
--- a/scripts/OverlaySurvey.py
+++ b/scripts/OverlaySurvey.py
@@ -56,6 +56,7 @@ def update_results(graph, parent_info, parent_key, results, is_inbound):
                        numTotalInboundPeers=parent_info[
                            "numTotalInboundPeers"
                        ])
+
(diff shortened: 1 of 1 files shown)


MonsieurNicolas:
r+ f2e9fb6b669031cdea642aa6d8c6cc0019394d01

PR #3638, opened 2022-12-21
Add a new load generation mode that produces DEX transactions.

dmkozh:
# Description

This allows to test setups that involve the mix of non-DEX and DEX traffic and its interaction with the DEX-limiting config setting.

Also refactored the loadgen code a bit and fixed the existing test.

# Checklist
- [ ] Compiles
- [ ] Ran all tests

### DIFF
--- a/src/main/CommandHandler.cpp
+++ b/src/main/CommandHandler.cpp
@@ -989,7 +989,9 @@ CommandHandler::generateLoad(std::string const& params, std::string& retStr)
             parseOptionalParam<uint32_t>(map, "maxfeerate");
         cfg.skipLowFeeTxs =
             parseOptionalParamOrDefault<bool>(map, "skiplowfeetxs", false);
-
+        // Only for MIXED_TX mode; fraction of DEX transactions.
+        cfg.dexTxPercent =
+            parseOptionalParamOrDefault<uint32_t>(map, "dextxpercent", 0);
         if (cfg.batchSize > 100)
         {
             cfg.batchSize = 100;
--- a/src/simulation/LoadGenerator.cpp
+++ b/src/simulation/LoadGenerator.cpp
@@ -81,6 +81,10 @@ LoadGenerator::getMode(std::string const& mode)
     {
         return LoadGenMode::PRETEND;
     }
+    else if (mode == "mixed_txs")
+    {
+        return LoadGenMode::MIXED_TXS;
+    }
     else
     {
         // unknown mode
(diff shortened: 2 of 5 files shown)


src/simulation/LoadGenerator.cpp
```diff
@@ -258,21 +262,69 @@ LoadGenerator::generateLoad(GeneratedLoadConfig cfg)
...
+                generateTx;
+
+            switch (cfg.mode)
+            {
+            case LoadGenMode::CREATE:
+                releaseAssert(false);
+                break;
+            case LoadGenMode::PAY:
+                generateTx = [&]() {
+                    return paymentTransaction(cfg.nAccounts, cfg.offset,
```
MonsieurNicolas, line 286: I think asan complains that you're capturing `sourceAccountId` and `opCount` but those variables go out of scope

dmkozh: Nice, your change has already saved me a couple of hours debugging time!

src/simulation/LoadGenerator.cpp
```diff
@@ -382,23 +425,17 @@ LoadGenerator::submitPaymentOrPretendTx(GeneratedLoadConfig const& cfg,
             status != TransactionQueue::AddResult::ADD_STATUS_ERROR)
         {
             mFailed = true;
-            return 0;
+            return false;
```
MonsieurNicolas, line 428: you missed the `return cfg.nTxs;` above that should be `false` I think given the refactor (C++ is nice like that)

src/simulation/LoadGenerator.cpp
```diff
@@ -258,21 +262,69 @@ LoadGenerator::generateLoad(GeneratedLoadConfig cfg)
...
+                    }
+                };
+            }
             break;
-        }
+            }
 
+            if (submitTx(cfg, generateTx))
+            {
+                --cfg.nTxs;
```
MonsieurNicolas, line 325: note that with this refactor, unlike the previous version, loadgen attempts to keep submitting transactions even after running into an error. I think this is fine @marta-lokhova can confirm

MonsieurNicolas (changes requested):

src/main/CommandHandler.cpp
```diff
@@ -989,7 +989,9 @@ CommandHandler::generateLoad(std::string const& params, std::string& retStr)
             parseOptionalParam<uint32_t>(map, "maxfeerate");
         cfg.skipLowFeeTxs =
             parseOptionalParamOrDefault<bool>(map, "skiplowfeetxs", false);
-
+        // Only for MIXED_TX mode; fraction of DEX transactions.
```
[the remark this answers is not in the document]
dmkozh: Done.

src/simulation/LoadGenerator.cpp
```diff
@@ -81,6 +81,10 @@ LoadGenerator::getMode(std::string const& mode)
     {
         return LoadGenMode::PRETEND;
     }
+    else if (mode == "mixed_txs")
```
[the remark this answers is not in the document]
dmkozh: Done.

src/simulation/LoadGenerator.cpp
dmkozh: Nice catch, fixed.

MonsieurNicolas (approved):
LGTM.
Please rebase & squash so that we can merge

src/simulation/LoadGenerator.cpp
marta-lokhova: Sorry I didn't see this earlier. I don't think we should keep submitting transactions after failures. We have some missions in CI that wait until loadgen succeeds or fails so the mission can "reset" core and retry. The expectation is that after completion no new tx load is generated by loadgen.

dmkozh: Actually I didn't intend to change the behavior here; it was just a bug. Added a break on failure.

dmkozh:
> LGTM. Please rebase & squash so that we can merge

Done.

MonsieurNicolas:
r+ adaed55a088a7ab70083cc5483074f66d35700bb

PR #3642, opened 2023-01-06
Extend survey functionality to include connection targets

marta-lokhova:
Update network survey functionality to include connectivity targets (inbound/outbound). Note that the new-style survey requests will only be flooded to version-compatible nodes, so we'll see better results overtime as more nodes upgrade their core version.

Since this PR changes XDR, an appropriate change is needed in the XDR next repo (CI won't build currently)


docs/software/commands.md
```diff
@@ -298,10 +298,13 @@ format.
         ledger<br>
 
 * **surveytopology**
-  `surveytopology?duration=DURATION&node=NODE_ID`<br>
+  `surveytopology?duration=DURATION&node=NODE_ID&surveyversion=0`<br>
```
MonsieurNicolas, line 301: I don't think it should work like this:
relaying nodes don't have access to the encrypted survey result (ie: `SurveyResponseBody` is not visible on the wire), it's probably better to just change the surveying code to respond with the new extended result, and leave everything else as is.

That way:
* existing nodes will be able to relay requests and (encrypted) result even if they're running an older version of the survey code
* we can deprecate right away responding with "old style" survey responses
* old nodes can't survey the network (because they may receive responses from newer nodes), but that seems fine

src/overlay/Floodgate.cpp
```diff
@@ -124,6 +124,11 @@ Floodgate::broadcast(StellarMessage const& msg, bool force,
     for (auto peer : peers)
     {
         releaseAssert(peer.second->isAuthenticated());
+        if (!peer.second->isMessageSupported(msg))
```
MonsieurNicolas, line 127: this change is not needed

src/main/Config.cpp
```diff
@@ -128,7 +128,7 @@ Config::Config() : NODE_SEED(SecretKey::random())
     MAXIMUM_LEDGER_CLOSETIME_DRIFT = 50;
 
     OVERLAY_PROTOCOL_MIN_VERSION = 23;
-    OVERLAY_PROTOCOL_VERSION = 26;
+    OVERLAY_PROTOCOL_VERSION = 27;
```
MonsieurNicolas, line 131: with my proposed change, you don't even need to update this version

MonsieurNicolas: (I guess you *could* keep this to make it easier for testing: a node with `OVERLAY_PROTOCOL_VERSION<=26` would return old style surveys)

src/overlay/SurveyManager.cpp
```diff
@@ -284,28 +298,44 @@ SurveyManager::processTopologyRequest(SurveyRequestMessage const& request) const
...
+            static_cast<uint32_t>(randomInboundPeers.size());
+        topologyBody.totalOutboundPeerCount =
+            static_cast<uint32_t>(randomOutboundPeers.size());
+    };
 
-    topologyBody.totalInboundPeerCount =
-        static_cast<uint32_t>(randomInboundPeers.size());
-    topologyBody.totalOutboundPeerCount =
-        static_cast<uint32_t>(randomOutboundPeers.size());
+    bool extendedSurvey = request.commandType == SURVEY_TOPOLOGY_EXTENDED;
```
MonsieurNicolas, line 324: updated way will be `bool extendedSurvey = body.type()` than couple with `commandType`

MonsieurNicolas (changes requested):

src/main/Config.cpp
marta-lokhova: This changes behavior of overlay, so the version here needs to be bumped regardless (it will be 27 in 19.7.0. either way due to other changes too)

docs/software/commands.md
marta-lokhova: Hmm. My thinking was that as per our usual overlay processes, we provide backward compatibility for at least some time, which should allow old nodes to use survey functionality. We've never upgraded survey before, so perhaps we're lacking a process and backward compatibility expectations here. Given that this isn't threatening to consensus and can be mitigated with a version upgrade, maybe it's ok to stop supporting survey for older nodes (I don't know who is using though and if it will cause problems).

Also - the fact that old nodes can still properly process newer requests/responses is true for this particular version but it's not guaranteed in the future. As the survey format changes, we'll still have to do what this PR does.

MonsieurNicolas: > Hmm. My thinking was that as per our usual overlay processes, we provide backward compatibility for at least some time, which should allow old nodes to use survey functionality. We've never upgraded survey before, so perhaps we're lacking a process and backward compatibility expectations here. Given that this isn't threatening to consensus and can be mitigated with a version upgrade, maybe it's ok to stop supporting survey for older nodes (I don't know who is using though and if it will cause problems).

I think it's acceptable to require surveying nodes to run the latest version.

> Also - the fact that old nodes can still properly process newer requests/responses is true for this particular version but it's not guaranteed in the future. As the survey format changes, we'll still have to do what this PR does.

I don't think this is true: the survey request and relaying messages should not change format. The only thing that should change is the format of the survey response (in response to a generic `SURVEY_REQUEST` message) and is encoded as an encrypted opaque blob.

src/main/Config.cpp
MonsieurNicolas: Technically speaking it's not because the part that is really part of the overlay protocol does not include the survey response format (old nodes would forward and respond to survey requests just fine)

docs/software/commands.md
marta-lokhova: >I don't think this is true: the survey request and relaying messages should not change format. The only thing that should change is the format of the survey response (in response to a generic SURVEY_REQUEST message) and is encoded as an encrypted opaque blob.

I don't think we can guarantee this. We don't know what the survey needs will look like in the future. The request itself contains SurveyMessageCommandType which assumes the possibility of running survey types. All I'm trying to say is that it's not a problem right now, but we might still need to deal with this in the future

src/overlay/SurveyManager.cpp
```diff
@@ -568,6 +586,8 @@ SurveyManager::commandTypeName(SurveyMessageCommandType type)
     {
     case SURVEY_TOPOLOGY:
         return "SURVEY_TOPOLOGY";
+    case SURVEY_TOPOLOGY_EXTENDED:
```
MonsieurNicolas, line 589: We should probably cleanup the XDR: there are more than 1 enum.

One contains the list of "commands", and the only supported command is `SURVEY_TOPOLOGY`.

Then there are different `SurveyResponseBody` versions (this PR adds 1 new one), so we should just create a new enum for that `SurveyResponseType` with values `SURVEY_TOPOLOGY_RESPONSE_V0=0` and `SURVEY_TOPOLOGY_RESPONSE_V1=1`

As for this function we can just call `xdr::xdr_traits<SurveyMessageCommandType>::enum_name(type)` instead in `getMsgSummary`

src/main/Config.cpp
marta-lokhova: Not sure I follow - if I'm talking to a node of a particular overlay version, I'm expecting a concrete protocol to be followed and a specific message format that I know about. Having the two nodes with "the same version", yet not being able to process each other's messages doesn't seem right.

src/overlay/SurveyManager.cpp
marta-lokhova: Ok, it's updated. I will submit a PR in the vNext repo when this PR is approved.

src/main/Config.cpp
MonsieurNicolas: Survey is not implemented at this layer: you can receive a new format survey response via an old peer just fine.

MonsieurNicolas (approved):
Changes look good to me. You just need to rev the xdr for vNext and we'll merge.

Thank you

MonsieurNicolas:
r+ 4b726190cd9a538fd7c1f2fcaee8b68076deff3c

PR #3648, opened 2023-01-20
Support extended survey in the survey script

marta-lokhova:
In addition, add a comment with sample response, which is useful to test the script (instead of re-generating sample response in core every time for testing)


MonsieurNicolas:
r+ a39629a34b5822caf0f4a0ec44926e262ce5224a

charlestonXdao (commented):
good

PR #3650, opened 2023-01-26
Out of sync peer recovery

marta-lokhova:


src/overlay/OverlayManagerImpl.cpp
```diff
@@ -586,13 +629,24 @@ OverlayManagerImpl::tick()
...
 
         releaseAssert(pendingUsedByPreferred <= availablePendingSlots);
         availablePendingSlots -= pendingUsedByPreferred;
     }
 
+    // Only trigger reconnecting if:
+    //   * no outbound slots are available
+    //   * we didn't establish any new preferred peers connections (those
+    //      will evict regular peers anyway)
+    if (availableAuthenticatedSlots == 0 && pendingUsedByPreferred == 0)
```
MonsieurNicolas, line 643: Should the second criteria just be that we can initiate connections? I don't see why trying to connect to a preferred peer should stop the node from triggering this logic

src/overlay/OverlayManagerImpl.cpp
```diff
@@ -517,6 +519,46 @@ OverlayManagerImpl::connectTo(std::vector<PeerBareAddress> const& peers,
...
+                }
+            }
+            else
+            {
+                // Still waiting for the timeout
+                return;
+            }
+        }
+    }
+    mLastOutOfSyncReconnect = std::make_optional<VirtualClock::time_point>(now);
```
MonsieurNicolas, line 559: Should this be moved to when we actually drop a peer?

MonsieurNicolas (changes requested):
Just a couple questions for you

src/overlay/OverlayManagerImpl.cpp
marta-lokhova: I think it's not the only place where we should set the time: when we go first go out of sync, we want to timeout before this logic kicks in (in case there's a transient issue unrelated to disconnected graph).

src/overlay/OverlayManagerImpl.cpp
marta-lokhova: I kept it like this because preferred peers have their own eviction policy when all connections are taken, so I didn't override it just in case (though it seems a bit arbitrary)

MonsieurNicolas: what I mean is that if we have 2 pending slots available and 1 preferred peer to connect to, there is no reason to not trigger this logic.

Trying to connect to a preferred peer does not mean that we'll be able to connect to it, so in this case we'd just be wasting cycles. So I'd rather not couple those behaviors.

src/overlay/OverlayManagerImpl.cpp
MonsieurNicolas: I am not sure I understand:
I thought the intent of `mLastOutOfSyncReconnect` was to throttle how often we drop peers.

Why would we want to wait the very first time? When that code triggers we already have connected to a lot of peers (because we don't have capacity) so I don't think this would be a transient issue.

If we want to avoid doing certain things right when we lose sync, we need to reset `mLastOutOfSyncReconnect` when we lose sync. That said the node was already waiting `CONSENSUS_STUCK_TIMEOUT_SECONDS` (35 seconds) before going out of sync, so I don't think the extra delay is needed (ie: when we lose sync, it's possible that we're only connected to bad peers).

MonsieurNicolas (changes requested):

src/overlay/OverlayManagerImpl.cpp
marta-lokhova: I don't think the extra delay is controversial, it's a conservative change to prevent unnecessary peer churn because "not tracking" does not always equal "bad connection graph".

MonsieurNicolas: Which comment are you responding to? I was asking which problem you're trying to address as it looks like you're trying to do something related to when a node transitions to out of sync but it's potentially incomplete

marta-lokhova: I'm talking about waiting for timeout before dropping peer for the first time after going out of sync. `mLastOutOfSyncReconnect` is reset either way in that function, whether we're tracking or not tracking. I'm not sure what you're saying above about the logic being incomplete.

marta-lokhova:
@MonsieurNicolas I updated this PR based on our offline discussion.

MonsieurNicolas:
r+ 174cddc403e452db84d954ba2f37f066db11a20c

PR #3669, opened 2023-02-27
Tests: allow overlay to work automatically in loopback mode

marta-lokhova:


src/simulation/Simulation.cpp
```diff
@@ -116,14 +117,26 @@ Simulation::addNode(SecretKey nodeKey, SCPQuorumSet qSet, Config const* cfg2,
     Application::pointer app;
     if (newDB)
     {
-        app = Application::create(*clock, *cfg, newDB);
+        if (mMode == OVER_LOOPBACK)
```
MonsieurNicolas, line 120: There might be something wrong.
with this change, "peer discovery" should work in loopback mode, right?

As a consequence, I would expect a need to disable this new behavior when testing specific properties:
I see that tests such as "herder externalizes values" rely on specific network topologies as to halt the network.

I know flooding tests also want to test what happens over various  topologies; that said, those tests are also testing using TCP -- so maybe there is something else I am missing (or that's a bug in the flood tests that we don't disable peer discovery)?

marta-lokhova: By default, RUN_STANDALONE in tests is true, which means we won't start the overlay manager. To enable peer discovery, that flag needs to be disabled manually.

MonsieurNicolas:
r+ 72909efb9a2599159eca9846be65f6314e70c3be

Issue #3679, opened 2023-03-20
Containerized dev environment installation complaining about install-rush.sh


anupsdf:
Following the steps in [Containerized dev environment](https://github.com/stellar/stellar-core/blob/master/INSTALL.md#containerized-dev-environment), I am getting this error when trying to open container in VSCode.
```

 => ERROR [dev_container_auto_added_stage_label  9/11] RUN ./install-rust  0.2s
------
 > [dev_container_auto_added_stage_label  9/11] RUN ./install-rust.sh:
#0 0.193 /bin/sh: 1: ./install-rust.sh: not found
------
ERROR: failed to solve: executor failed running [/bin/sh -c ./install-rust.sh]: exit code: 127
```

graydon:
@anupsdf [this](https://github.com/stellar/stellar-core/pull/3680) ought to get you unblocked, but .. I also don't think you're going to have a good time with devcontainers in general. core should build outside of them and inside it seems to be just excruciatingly slow (probably docker's amazingly bad union-filesystem performance). I would recommend building outside.

PR #3680, opened 2023-03-21
Use less-rigorously-checked rustup in devcontainer due to bug in vscode

graydon:
This is a "fix" for #3679 that I am not especially thrilled with but I spent half a day trying to fix it better and concluded that vscode's devcontainers don't really support anything better (build.context doesn't, for example, do what it's documented to do, and there appears to be no way to COPY or ADD things from outside the .devcontainer directory). Feel free to fix better sometime if this version distresses you, but .. I think the qualifiers in the comment here are accurate (it's really not much of a risk, this isn't building canonical packages or even CI) and .. I also think devcontainer use of stellar-core is probably not something that's ever going to work well. It's exceedingly slow and fragile.

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -67,7 +67,12 @@ RUN apt-get -y install cpp-10 gcc-10 g++-10
 # clang
 RUN apt-get -y install clang-12 llvm-12
 # rust
-RUN ./install-rust.sh
+# there is a bug in vscode's handling of devcontainer contexts that means
+# we can't access files outside this directory even with build.context,
+# but since we're building inside a container and only doing so for local
+# development, the security risks of running the normal rustup installer
+# are considerably less; for now we will tolerate them.
+RUN curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y
 ENV PATH "/root/.cargo/bin:$PATH"
 
 # clang by default


.devcontainer/Dockerfile
```diff
@@ -67,7 +67,12 @@ RUN apt-get -y install cpp-10 gcc-10 g++-10
 # clang
 RUN apt-get -y install clang-12 llvm-12
 # rust
-RUN ./install-rust.sh
+# there is a bug in vscode's handling of devcontainer contexts that means
```
MonsieurNicolas, line 70: a better workaround might be to do:
```
diff --git a/.devcontainer/Dockerfile b/.devcontainer/Dockerfile
index 9a325083c..ed531f707 100644
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -67,7 +67,6 @@ RUN apt-get -y install cpp-10 gcc-10 g++-10
 # clang
 RUN apt-get -y install clang-12 llvm-12
 # rust
-RUN ./install-rust.sh
 ENV PATH "/root/.cargo/bin:$PATH"

 # clang by default
diff --git a/.devcontainer/devcontainer.json b/.devcontainer/devcontainer.json
index fa6c17e38..253a6b775 100644
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -8,6 +8,7 @@
                        "APT_MIRROR": "${localEnv:APT_MIRROR}"
                }
        },
+       "postStartCommand": "./install-rust.sh",
        // The optional 'runArgs' property can be used to specify additional runtime arguments.
        "runArgs": [
                // Uncomment the line if you will use a ptrace-based debugger like C++, Go, and Rust.
```

It's possible to use `postCreateCommand ` instead of `postStartCommand` but in the context of a dev environment, we probably just want to run that script all the time anyways

graydon: Won't that reinstall rust in the container _every time_ you start the container? I thought the point was to cache an image that serves as a build environment.

MonsieurNicolas: the scripts no-ops if there is nothing to do

graydon: Ok but is the container _stateful_? I thought it would reinitialize itself every time it starts. I honestly don't know what sort of mess of union filesystems or temporary filesystems is involved, if that works go for it!

MonsieurNicolas: yes it works - dev containers need to be explicitly reset to lose state

MonsieurNicolas: I ended up opening https://github.com/stellar/stellar-core/pull/3683 as I was debugging something and there was some issues on Windows in general

graydon:
closing in favor of #3683

PR #3687, opened 2023-04-03
overlay tech debt: pull mode

### LINKED ISSUE #3675: overlay tech debt: pull mode
Since we've been adding new features to overlay, we've accumulated quite a bit of tech debt that would be useful to address before we add new features. I think this is worth investing resources into as it's becoming harder to unit test and validate new features in overlay because we don't have proper abstractions and separation of responsibilities. Auditing the code, I think we have an opportunity to cleanup the Peer class, and focus on pull mode specifically:
* Right now there is no single module responsible for pull mode: OverlayManager does demanding, while Peer deals with advert processing. Neither of these classes is actually supposed to be responsible for transactions processing (Peer represents a "connection" and handles transport, authentication, etc) while OverlayManager is responsible for managing who we're connected to. This makes it hard to make and unit test any new changes, as it requires knowing everything about Peer and OverlayManager. Pull mode is essentially a queue of transaction information and a timer to process those transactions on a schedule, so it should be fairly easy to extract this functionality into its own module.
* We need to provide a lean API for other parts of the system to use. For example, right now functions dealing with internal policies such as purging are exposed publicly. 
* Ideally, with this refactor we should be able to unit test a narrow set of pull mode scenarios, rather than spinning up a complete flow every time just to test a specific policy.

hidenori-shinohara:
# Description

Resolves #3675 

Moves the pull-mode logic into a separate class. This PR lacks a purging policy for when we lose a connection with a peer. This PR changes the advert behavior slightly as it has only one advert timer for all connections, which may not be desirable. 

# Checklist
- [ ] Compiles
- [ ] Ran all tests


marta-lokhova:
thanks! one step closer to a cleaner codebase :)

src/overlay/Peer.h
```diff
@@ -315,22 +315,11 @@ class Peer : public std::enable_shared_from_this<Peer>,
...
-    // of flood messages, then periodically flush the advert and await a
-    // _demand_ message with a list of flood messages to send. Adverts are
-    // typically smaller than full messages and batching them means we also
-    // amortize the authentication framing.
-    TxAdvertVector mTxHashesToAdvertise;
-    void flushAdvert();
-    VirtualTimer mAdvertTimer;
-    void startAdvertTimer();
     // transaction hash -> ledger number
     RandomEvictionCache<Hash, uint32_t> mAdvertHistory;
```
marta-lokhova, line 324: TODO: this should probably live in separate pull mode modules

marta-lokhova:
I would like to push a few tweaks to this change, so will update this PR. Hopefully it's alright to push directly to the branch.

src/overlay/Peer.h
```diff
@@ -315,25 +311,11 @@ class Peer : public std::enable_shared_from_this<Peer>,
...
     size_t mAdvertQueueTxHashCount{0};
     size_t mDemandQueueTxHashCount{0};
 
-    // As of MIN_OVERLAY_VERSION_FOR_FLOOD_ADVERT, peers accumulate an _advert_
-    // of flood messages, then periodically flush the advert and await a
-    // _demand_ message with a list of flood messages to send. Adverts are
-    // typically smaller than full messages and batching them means we also
-    // amortize the authentication framing.
-    TxAdvertVector mTxHashesToAdvertise;
-    void flushAdvert();
```
marta-lokhova, line 331: TODO: move advert/demand metrics to flood manager

src/overlay/TxFloodManager.cpp
```diff
@@ -0,0 +1,431 @@
...
+    {
+        // We move `demand` here and also pass `retry` as a reference
+        // which gets appended. Don't touch `demand` or `retry` after here.
+        peer->sendTxDemand(std::move(demandMap[peer].first));
+        if (demandMap[peer].second.empty())
+        {
+            continue;
+        }
+        auto it = mQueuedIncomingAdverts.find(peer);
+        if (it == mQueuedIncomingAdverts.end())
```
marta-lokhova, line 342: TODO: peer must already be present in incoming adverts, otherwise can't populate retries to begin with.

src/overlay/TxFloodManager.h
```diff
@@ -0,0 +1,89 @@
...
+    void recordTxPullLatency(Hash const& hash, std::shared_ptr<Peer> peer);
+
+    void shutdown();
+
+    void clearBelow(uint32_t seq);
+
+  private:
+    Application& mApp;
+
+    bool mShuttingDown;
```
marta-lokhova, line 40: TODO: uninitialized

src/main/Application.h
```diff
@@ -215,6 +216,7 @@ class Application
     virtual HerderPersistence& getHerderPersistence() = 0;
     virtual InvariantManager& getInvariantManager() = 0;
     virtual OverlayManager& getOverlayManager() = 0;
+    virtual TxFloodManager& getTxFloodManager() = 0;
```
marta-lokhova, line 219: TODO: OverlayManager owns and manages TxFloodManager

src/overlay/TxFloodManager.cpp
```diff
@@ -0,0 +1,275 @@
...
+    {
+        // We move `demand` here and also pass `retry` as a reference
+        // which gets appended. Don't touch `demand` or `retry` after here.
+        peer->sendTxDemand(std::move(demandMap[peer].first));
+        peer->retryAdvert(demandMap[peer].second);
+    }
+
+    // mPendingDemands and mDemandHistoryMap must always contain exactly the
+    // same tx hashes.
+    releaseAssert(mPendingDemands.size() == mDemandHistoryMap.size());
```
bboston7, line 229: This assert does not check what the comment says it checks. Maybe that's OK, but the comment should be updated to reflect that the check is intentionally different.

src/overlay/OverlayManager.h
```diff
@@ -90,6 +90,8 @@ class OverlayManager
         return recvFloodedMsgID(msg, peer, msgID);
     }
 
+    virtual void recvTransaction(Hash const& txHash, Peer::pointer peer) = 0;
```
bboston7, line 93: Please document this function

src/overlay/Peer.cpp
```diff
@@ -2205,4 +2127,60 @@ Peer::fulfillDemand(FloodDemand const& dmd)
...
+{
+    if (!hasAdvert())
+    {
+        throw std::runtime_error("No advert to pop");
+    }
+
+    if (!mTxPullMode)
+    {
+        throw std::runtime_error("Pull mode is not set");
+    }
```
bboston7, line 2181: I think the order of these conditionals needs to be reversed. `hasAdvert` returns false when `!mTxPullMode`, so the exception will always contain `No advert to pop` regardless of whether `popAdvert` fails because pull mode is disabled or it is enabled but empty.

src/overlay/Peer.h
```diff
@@ -364,6 +346,10 @@ class Peer : public std::enable_shared_from_this<Peer>,
     void sendGetScpState(uint32 ledgerSeq);
     void sendErrorAndDrop(ErrorCode error, std::string const& message,
                           DropMode dropMode);
+    void sendAdvert(Hash const& txHash);
+    void retryAdvert(std::list<Hash>& hashes);
+    bool hasAdvert();
+    std::pair<Hash, std::optional<VirtualClock::time_point>> popAdvert();
```
bboston7, line 352: Docs on these public functions please

src/overlay/Peer.h
```diff
@@ -459,13 +445,8 @@ class Peer : public std::enable_shared_from_this<Peer>,
...
     void fulfillDemand(FloodDemand const& dmd);
-    void queueTxHashToAdvertise(Hash const& hash);
-    void queueTxHashAndMaybeTrim(Hash const& hash);
-    TxAdvertQueue&
-    getTxAdvertQueue()
-    {
-        return mTxAdvertQueue;
-    };
+    bool peerKnowsHash(Hash const& hash);
+    void clearBelow(uint32_t ledgerSeq);
```
bboston7, line 449: Please document these public functions

bboston7 (changes requested):
Looks good! I requested mostly small changes + some additional documentation. "You" in my comments refers to whoever is taking this PR across the finish line.

src/overlay/Peer.cpp
marta-lokhova: Good catch, fixed

src/overlay/OverlayManager.h
marta-lokhova: Done (I also moved transaction processing to OverlayManager, as I don't think that functionality really belongs to Peer. At a high level, I think it makes more sense for Peer to just pass inbound messages to higher-level abstractions that will perform the business logic)

src/overlay/Peer.h
marta-lokhova: Done. Note on `peerKnowsHash`: I actually removed this method from the public API, so now the decision to broadcast an advert is isolated inside the Peer class.

src/overlay/Peer.h
marta-lokhova: Done.

src/overlay/TxPullMode.cpp
```diff
@@ -0,0 +1,195 @@
...
+    {
+        startAdvertTimer();
+    }
+
+    mOutgoingTxHashes.emplace_back(txHash);
+
+    // Flush adverts at the earliest of the following two conditions:
+    // 1. The number of hashes reaches the threshold.
+    // 2. The oldest tx hash hash been in the queue for FLOOD_TX_PERIOD_MS.
+    if (mOutgoingTxHashes.size() == getMaxAdvertSize())
```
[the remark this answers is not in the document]
marta-lokhova: It's a bit misleading since there are two conditions for flushing adverts, one is implemented in this function, and the other is managed elsewhere by the advert time. I edited to comment to clarify this.

src/overlay/TxPullMode.h
```diff
@@ -0,0 +1,67 @@
...
+    RandomEvictionCache<Hash, uint32_t> mAdvertHistory;
+    TxAdvertVector mOutgoingTxHashes;
+    VirtualTimer mAdvertTimer;
+    std::weak_ptr<Peer> mWeakPeer;
+
+    void rememberHash(Hash const& hash, uint32_t ledgerSeq);
+    void flushAdvert();
+    void startAdvertTimer();
+
+  public:
```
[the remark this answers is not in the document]
marta-lokhova: Done!

src/overlay/TxFloodManager.h
```diff
@@ -0,0 +1,70 @@
...
+/**
+ * @class TxFloodManager
+ *
+ * Manages queuing incoming and outgoing tx hashes including demands
+ *
+ */
+
+class TxFloodManager : private NonMovableOrCopyable
+{
+  public:
```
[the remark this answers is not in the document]
marta-lokhova: Of course, updated.

src/overlay/TxFloodManager.h
```diff
@@ -0,0 +1,70 @@
...
+    explicit TxFloodManager(Application& app);
+    static std::unique_ptr<TxFloodManager> create(Application& app);
```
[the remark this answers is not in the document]
marta-lokhova: Agreed, I think this was a remnant from a previous iteration, as it wasn't actually used. I removed it.

src/overlay/TxFloodManager.cpp
```diff
@@ -0,0 +1,275 @@
...
+        static_cast<double>(mApp.getHerder().getMaxQueueSizeOps()) *
+        opRatePerLedger;
+    releaseAssertOrThrow(queueSizeInOpsDbl >= 0.0);
+    int64_t queueSizeInOps = static_cast<int64_t>(queueSizeInOpsDbl);
+
+    size_t res = static_cast<size_t>(
+        bigDivideOrThrow(queueSizeInOps, cfg.FLOOD_DEMAND_PERIOD_MS.count(),
+                         ledgerCloseTime, Rounding::ROUND_UP));
+    res = std::max<size_t>(1, res);
+    res = std::min<size_t>(TX_DEMAND_VECTOR_MAX_SIZE, res);
```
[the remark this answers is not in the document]
marta-lokhova: Indeed, I updated the code.

src/overlay/TxFloodManager.cpp
marta-lokhova: I _think_ this was just meant to be a quick sanity check to avoid extra iterations, I'll make a note of it in the comment

marta-lokhova:
Closing in favor of https://github.com/stellar/stellar-core/pull/4294

PR #3699, opened 2023-04-06
overlay: add new tests to help evaluate connectivity strategy

marta-lokhova:
A subset of tests used to help evaluate changes to connectivity strategy (the rest are still WIP as we hash out the new strategy). These help verify that the network topology remains connected with various connection targets. In addition, new tests were added to evaluate network churn: specifically, gradually adding new nodes, randomly removing and re-adding nodes on the network, and churning nodes for longer periods of time to evaluate potential long-term degradation in topology. 

Note that it should not be expected for these to always pass, given the random nature of our connection management. However, it is useful to run these tests in a loop to evaluate how new changes impact connectivity.

### DIFF
--- a/src/overlay/test/LoopbackPeer.cpp
+++ b/src/overlay/test/LoopbackPeer.cpp
@@ -153,16 +153,17 @@ LoopbackPeer::drop(std::string const& reason, DropDirection direction, DropMode)
 
     if (mState != GOT_AUTH)
     {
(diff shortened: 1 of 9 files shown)


mbsdf:
r+ 2dcd65520ffa3250966f539bebed2de332fe0b4e

graydon:
r+ 2dcd655

Issue #3700, opened 2023-04-07
MAXIMUM_LEDGER_CLOSETIME_DRIFT should be consistent with SCP state sharing policy


marta-lokhova:
As per https://github.com/stellar/stellar-core/pull/3655#discussion_r1160976699

PR #3709, opened 2023-04-19
Calculate close time drift based on max slots to remember

marta-lokhova:
Resolves https://github.com/stellar/stellar-core/issues/3700


SirTyson:
Do you think we should also update `stellar-core_example.cfg` now that 50 seconds is no longer the actual default and add a comment about this?

SirTyson (approved):
LGTM

MonsieurNicolas:
r+ 645f52f26c0c66360f3adeb202cf51eabb549da9

Issue #3670, opened 2023-03-01
Limit core to ever store 1 transaction per source account


dmkozh:
## Background

Currently the Core implementation contains a pretty complex machinery built around maintaining multiple transactions per account, as long as their sequence numbers make a sequence starting from the ledger seqNum for that acount. This allows submitting multiple transactions to the Core in a short succession that most likely will be executed eventually (as long as fees are high enough or the ledger load is low). This also makes the Core implementation a fair bit more complex and results in a pretty tricky to interpret behavior (for example, it's not possible to 'kick out' your own transaction from the queue in case of traffic surge, which leads to a special sort of transaction rejections).

## Problem

The current approach may not guarantee that the fee can be charged from the account at the transaction application time, as in order to do that, the Core would need to simulate all the effects of the transaction to make sure that the account balance is enough. So it is possible, for example, to send a payment of the whole account balance in one transaction and another transaction with arbitrary large fee. If both transactions are to be included in the ledger, then the second transaction will fail due to insufficient fee. However, it has to be flooded in order to be included into ledger in the first place. 

This hasn't been much of on issue before, because all the transactions have a similar size. But with the upcoming Soroban launch, the current approach is unlikely to remain feasible. Soroban transactions may reach hundreds of kilboytes in size. Soroban works around such large transactions by introducing a 'bandwidth' fee dependent on the transaction size, However, due to the problem described above, it's possible to make the network flood very large transactions without actually paying any fees.

Also purely from the implementation standpoint, the Soroban transactions will likely need a separate transaction queue, so maintaining the already complicated logic would become even harder.

## Solution

The most straightforward solution to the problem is to only allow storing and flooding one transaction per source account at any given time. Basically, every transaction that for which sequence number doesn't match the account sequence number stored in the ledger will be immediately rejected by the Core node.

### Transition plan

This change may be staged in steps:

- Introduce a core configuration parameter that allows to store only a single transaction per account, disabled by default
- Enable the parameter on some validators to make sure there is no significant impact on the network behavior
- Enable the parameter by default
- Cleanup the Core code from the old logic

Only the first step has to happen before Soroban launch; other steps can be combined with it, although it is preferable to perform as many steps as possible before Soroban launch (as they don't require protocol upgrades)

## Consequences

The described behavior change may affect the client systems (mainly Horizon), as we are moving all the batching capabilities out of Core. 

[...]

PR #3726, opened 2023-05-01
Introduce source account limit in the tx queue

marta-lokhova:
First step outlined in https://github.com/stellar/stellar-core/issues/3670: introduce a flag that allows core to limit tx queue to 1 tx per source account. The flag is added on top of the current tx queue implementation. To avoid maintaining two implementations of tx queue, when the network switches to this limit, the code can be cleaned up to remove account chains. Note that tx set validation is not affected with this change. That is, nodes can still validate and vote on values introduced by nodes without this limit. If the limit becomes mandatory, we can enforce it in tx set validity checks in protocol 20.

Also, note that our test suite will need to be cleaned up when the limit becomes mandatory, as it currently heavily depends on multi-tx account chains logic.

### DIFF
--- a/src/herder/test/HerderTests.cpp
+++ b/src/herder/test/HerderTests.cpp
@@ -3023,6 +3023,134 @@ TEST_CASE("SCP checkpoint", "[catchup][herder]")
     }
 }
 
+// This test confirms that tx set processing and consensus are independent of
+// the tx queue source account limit (for now)
+TEST_CASE("tx queue source account limit", "[herder][transactionqueue]")
+{
+    std::shared_ptr<Simulation> simulation;
+    std::shared_ptr<Application> app;
+    std::shared_ptr<Application> limitApp;
+
+    auto setup = [&](bool mix) {
+        auto networkID = sha256(getTestConfig().NETWORK_PASSPHRASE);
+        simulation = std::make_shared<Simulation>(
+            Simulation::OVER_LOOPBACK, networkID, [mix](int i) {
+                auto cfg = getTestConfig(i, Config::TESTDB_ON_DISK_SQLITE);
+                if (!mix || i % 2 == 1)
+                {
+                    cfg.LIMIT_TX_QUEUE_SOURCE_ACCOUNT = true;
+                }
+                return cfg;
+            });
+
+        auto validatorAKey = SecretKey::fromSeed(sha256("validator-A"));
+        auto validatorBKey = SecretKey::fromSeed(sha256("validator-B"));
+        auto validatorCKey = SecretKey::fromSeed(sha256("validator-C"));
+
+        SCPQuorumSet qset;
+        // Everyone needs to vote to proceed
+        qset.threshold = 3;
+        qset.validators.push_back(validatorAKey.getPublicKey());
+        qset.validators.push_back(validatorBKey.getPublicKey());
+        qset.validators.push_back(validatorCKey.getPublicKey());
+
+        simulation->addNode(validatorAKey, qset);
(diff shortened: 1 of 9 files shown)


src/herder/test/HerderTests.cpp
```diff
@@ -3023,6 +3023,134 @@ TEST_CASE("SCP checkpoint", "[catchup][herder]")
...
+    std::shared_ptr<Simulation> simulation;
+    std::shared_ptr<Application> app;
+    std::shared_ptr<Application> limitApp;
+
+    auto setup = [&](bool mix) {
+        auto networkID = sha256(getTestConfig().NETWORK_PASSPHRASE);
+        simulation = std::make_shared<Simulation>(
+            Simulation::OVER_LOOPBACK, networkID, [mix](int i) {
+                auto cfg = getTestConfig(i, Config::TESTDB_ON_DISK_SQLITE);
+                if (!mix || i % 2 == 1)
```
dmkozh, line 3039: I think for additional coverage it might be worth covering mixed scenarios where the nodes with new setting is both majority and minority. Or do we have a consensus threshold of 1 here, so that wouldn't matter?

src/herder/test/HerderTests.cpp
```diff
@@ -3023,6 +3023,134 @@ TEST_CASE("SCP checkpoint", "[catchup][herder]")
...
+            },
+            3 * Herder::EXP_LEDGER_TIMESPAN_SECONDS, false);
+
+        for (auto const& node : simulation->getNodes())
+        {
+            // Applied txs were removed and banned
+            REQUIRE(node->getHerder().getTx(tx1->getFullHash()) == nullptr);
+            REQUIRE(node->getHerder().getTx(tx2->getFullHash()) == nullptr);
+            REQUIRE(node->getHerder().isBannedTx(tx1->getFullHash()));
+            // Second tx is not banned because it's never been flooded and
```
dmkozh, line 3143: I think it would make sense to also submit and get this tx applied to make sure we don't have any staleness issues for the 'prohibited' accounts set.

src/herder/test/TransactionQueueTests.cpp
```diff
@@ -1783,11 +1946,13 @@ TEST_CASE("transaction queue starting sequence boundary",
     }
 }
 
-TEST_CASE("transaction queue with fee-bump", "[herder][transactionqueue]")
+void
+testTxQueueFeeBump(bool limit)
```
dmkozh, line 1950: nit: `limitSourceAccounts`? It's unclear what exactly is limited here.

dmkozh (commented):
Thanks for doing this, looks good to me overall. Just left a few minor nitpicks concerning the tests.

src/herder/test/HerderTests.cpp
marta-lokhova: I believe this shouldn't matter because of the threshold, which requires all nodes to vote in order to confirm a tx set (which means all nodes must consider the value "valid")

src/herder/test/HerderTests.cpp
marta-lokhova: Good idea, added.

src/herder/test/TransactionQueueTests.cpp
marta-lokhova: agreed, updated.

dmkozh: Please also update `testReplaceByFee` as well

dmkozh (approved):

dmkozh:
r+ d23e175

PR #3737, opened 2023-05-16
overlay: add byte limit to flow control

marta-lokhova:
Functionally the change is equivalent to https://github.com/stellar/stellar-core/pull/3607, but opening a new PR since with the latest refactor the code is sufficiently different. 

One more TODO I'm currently working on is proper handling of Soroban txMaxSize upgrades, but we can start reviewing this change already (the upgrade handling is behind the vNext flag anyways)


marta-lokhova:
(note: still need to update vNext XDR, so the build is currently failing)

src/overlay/Peer.h
```diff
@@ -22,6 +22,10 @@ namespace stellar
 
 typedef std::shared_ptr<SCPQuorumSet> SCPQuorumSetPtr;
 
+static auto const MAX_MESSAGE_SIZE = 0x1000000;
+// max tx size is 64KB
+static const uint64_t MAX_CLASSIC_TX_SIZE_BYTES = 64000;
```
marta-lokhova, line 27: heads up: this const and capacity configs are placeholders for now. I'll update those once I finish simulation tuning

src/main/Config.cpp
```diff
@@ -1427,6 +1476,36 @@ Config::processConfig(std::shared_ptr<cpptoml::table> t)
             throw std::runtime_error(msg);
         }
 
+        if (FLOW_CONTROL_SEND_MORE_BATCH_SIZE_BYTES >
+            PEER_FLOOD_READING_CAPACITY_BYTES)
+        {
+            std::string msg =
+                "Invalid configuration: "
+                "FLOW_CONTROL_SEND_MORE_BATCH_SIZE_BYTES "
```
mbsdf, line 1484: +1

src/main/Config.cpp
```diff
@@ -1427,6 +1476,36 @@ Config::processConfig(std::shared_ptr<cpptoml::table> t)
...
+
+        // This condition is needed to ensure flow control never reaches a
+        // deadlock state. Specifically, it prevents a situation where the
+        // sender is waiting for available capacity to send new messages, but is
+        // unable to do so because the receiver has not yet processed enough
+        // bytes to hit the FLOW_CONTROL_SEND_MORE_BATCH_SIZE_BYTES threshold
+        // and request additional capacity. IOW, by satisfying this condition,
+        // we guarantee that the sender won't be perpetually stalled, waiting
+        // for the receiver's request for more data.
+        if (!(PEER_FLOOD_READING_CAPACITY_BYTES -
```
mbsdf, line 1497: why minus send_more_batch_size?

src/main/Config.cpp
```diff
@@ -1427,6 +1476,36 @@ Config::processConfig(std::shared_ptr<cpptoml::table> t)
...
+                  FLOW_CONTROL_SEND_MORE_BATCH_SIZE_BYTES >=
+              MAX_CLASSIC_TX_SIZE_BYTES))
```
mbsdf, line 1499: what is max classic tx size? do we read it from somewhere?

src/main/Config.cpp
```diff
@@ -1427,6 +1476,36 @@ Config::processConfig(std::shared_ptr<cpptoml::table> t)
...
+        {
+            std::string msg =
+                "Invalid configuration: the difference between "
+                "PEER_FLOOD_READING_CAPACITY_BYTES and "
```
mbsdf, line 1503: can send_more_batch_size be less than max_tx_size?

src/overlay/FlowControl.cpp
```diff
@@ -96,7 +150,7 @@ FlowControl::maybeSendNextBatch()
         {
             auto& front = queue.front();
             auto const& msg = *(front.mMessage);
-            if (!mFlowControlCapacity->hasOutboundCapacity(msg))
+            if (!hasOutboundCapacity(msg))
```
mbsdf, line 153: +1

src/overlay/FlowControl.cpp
```diff
@@ -188,48 +256,91 @@ FlowControl::endMessageProcessing(StellarMessage const& msg,
...
     releaseAssert(mFloodDataProcessed <=
                   mApp.getConfig().FLOW_CONTROL_SEND_MORE_BATCH_SIZE);
     bool shouldSendMore = mFloodDataProcessed ==
                           mApp.getConfig().FLOW_CONTROL_SEND_MORE_BATCH_SIZE;
+    if (mFlowControlBytesCapacity)
+    {
+        shouldSendMore =
+            shouldSendMore ||
+            mFloodDataProcessedBytes >=
+                mApp.getConfig().FLOW_CONTROL_SEND_MORE_BATCH_SIZE_BYTES;
```
mbsdf, line 274: what it means if less than batch_size?

src/main/Config.h
```diff
@@ -248,6 +252,22 @@ class Config : public std::enable_shared_from_this<Config>
     // processes `FLOW_CONTROL_SEND_MORE_BATCH_SIZE` messages
     uint32_t FLOW_CONTROL_SEND_MORE_BATCH_SIZE;
 
+    // A config parameter that controls how many bytes worth of flood messages
+    // (tx or SCP) from a particular peer core can process simultaneously
+    uint32_t PEER_FLOOD_READING_CAPACITY_BYTES;
```
mbsdf, line 257: I think it will be nice to add a description in this PR to explain peer flood capacity, batch size for both sender and  receiver (maybe a simple graph will be better?)

src/overlay/FlowControl.cpp
```diff
@@ -188,48 +256,91 @@ FlowControl::endMessageProcessing(StellarMessage const& msg,
...
 }
 
 bool
 FlowControl::isSendMoreValid(StellarMessage const& msg,
                              std::string& errorMsg) const
 {
-    releaseAssert(msg.type() == SEND_MORE);
-    if (getNumMessages(msg) == 0)
+    bool sendMoreExtendedType =
+        mFlowControlBytesCapacity && msg.type() == SEND_MORE_EXTENDED;
```
mbsdf, line 314: send more and send more extended, how they are different?

src/main/Config.h
marta-lokhova: Added!

src/overlay/FlowControlCapacity.cpp
```diff
@@ -29,24 +29,74 @@ FlowControlMessageCapacity::getMsgResourceCount(StellarMessage const& msg) const
...
+
+FlowControlByteCapacity::FlowControlByteCapacity(Application& app,
+                                                 NodeID const& nodeID)
+    : FlowControlCapacity(app, nodeID)
+{
+    mCapacity = getCapacityLimits();
+}
+
+FlowControlCapacity::ReadingCapacity
+FlowControlByteCapacity::getCapacityLimits() const
```
SirTyson, line 67: Just curious, why do we have a message count reading capacity but not a byte based reading capacity?

src/overlay/test/OverlayTests.cpp
```diff
@@ -97,45 +97,226 @@ TEST_CASE("loopback peer send auth before hello", "[overlay][connections]")
...
+            auto& sendMoreRecvMeter = conn.getInitiator()
+                                          ->getApp()
+                                          .getOverlayManager()
+                                          .getOverlayMetrics()
+                                          .mRecvSendMoreTimer;
+            auto currentSendCount = sendMoreMeter.count();
+            auto currentRecvCount = sendMoreRecvMeter.count();
+            testutil::crankSome(clock);
+            REQUIRE(sendMoreMeter.count() ==
+                    (currentSendCount + !!shouldRequestMore));
```
SirTyson, line 165: Nit: Double negation

src/protocol-curr/xdr/Stellar-overlay.x
```diff
@@ -47,16 +53,7 @@ struct Hello
...
-// During the roll-out phrase, pull mode will be optional.
-// Therefore, we need a way to communicate with other nodes
-// that we want/don't want pull mode.
-// However, the goal is for everyone to enable it by default,
-// so we don't want to introduce a new member variable.
-// For now, we'll use the `flags` field (originally named
-// `unused`) in `Auth`.
-// 100 is just a number that is not 0.
-const AUTH_MSG_FLAG_PULL_MODE_REQUESTED = 100;
+const AUTH_MSG_FLAG_FLOW_CONTROL_BYTES_REQUESTED = 200;
```
SirTyson, line 56: This value seems cryptic, I'd add some comment saying something like "must be a value > 0 && value != 100" with a brief explanation as to why.

SirTyson (approved):
A few questions but LGTM

src/protocol-curr/xdr/Stellar-overlay.x
```diff
@@ -27,6 +27,12 @@ struct SendMore
     uint32 numMessages;
 };
 
+struct SendMoreExtended
+{
+    uint32 numMessages;
+    uint32 numBytes;
```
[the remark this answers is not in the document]
marta-lokhova: for overlay, we normally introduce new messages instead of extending old ones, since older overlay versions become obsolete and this approach makes it easy to deprecate old behavior (note that this is different from protocol, where we need to support old behavior forever)

src/overlay/FlowControlCapacity.cpp
marta-lokhova: good question. the reading capacity is a purely internal concept, needed to ensure fairness across all of our connections (what we used to do was read from a peer while there was any data available, potentially starving other peers). the reason I decided to keep only the message counting for now was because the largest message we allow in overlay is 16MB anyway, so it seemed like it wasn't super useful to try to enforce the existing byte limits, since they would be way off when we receive larger messages like tx sets. That being said, we might do something better than this in the future, but for now I'm leaving it as-is.

src/overlay/FlowControl.cpp
```diff
@@ -188,48 +256,91 @@ FlowControl::endMessageProcessing(StellarMessage const& msg,
...
+            shouldSendMore ||
+            mFloodDataProcessedBytes >=
+                mApp.getConfig().FLOW_CONTROL_SEND_MORE_BATCH_SIZE_BYTES;
+    }
     auto peerPtr = peer.lock();
 
     if (shouldSendMore && peerPtr)
     {
-        sendSendMore(mFloodDataProcessed, peerPtr);
+        if (mFlowControlBytesCapacity)
```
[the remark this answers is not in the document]
marta-lokhova: Yeah, this pointer is set if a connection decides to use fc in bytes

src/main/Config.cpp
marta-lokhova: yes, batch size is just a frequency of data requests.

src/main/Config.cpp
marta-lokhova: it's a const. In the future, the actual max tx size value will be derived from this const _and_ Soroban network config.

src/main/Config.cpp
```diff
@@ -1434,6 +1456,61 @@ Config::processConfig(std::shared_ptr<cpptoml::table> t)
...
+        if (FLOW_CONTROL_SEND_MORE_BATCH_SIZE_BYTES >
+            PEER_FLOOD_READING_CAPACITY_BYTES)
+        {
+            std::string msg =
+                "Invalid configuration: "
+                "FLOW_CONTROL_SEND_MORE_BATCH_SIZE_BYTES "
+                "can't be greater than PEER_FLOOD_READING_CAPACITY_BYTES";
+            throw std::runtime_error(msg);
+        }
+
```
mbsdf, line 1468: +1

mbsdf (approved):
Look good to me. Only have a few comments.  

PS: already reviewed #3607 and reviewing this PR was much easier after overlay refactoring #3686

src/overlay/FlowControl.cpp
marta-lokhova: send_more is for current fc (messages only), while `extended` includes byte counts

mbsdf:
r+ da8e7c7

marta-lokhova:
r+ f01256999c4b8d1e03256050fef696ea7e4837fd

Issue #3693, opened 2023-04-04
Refundable Soroban fees


dmkozh:
https://github.com/stellar/stellar-protocol/blob/master/core/cap-0046-07.md defines two types of fees that are withdrawn at apply time and with the amounts based on the actual resource usage:

- Historical storage fee 
- Metadata fee

Besides implementing the post-apply fee withdrawal process (or, alternatively, fee refund process), we need to figure out what to do with transactions that didn't fail during the normal executiong, but didn't declare enough metadata usage.

PR #3738, opened 2023-05-17
Implement refunds for the Soroban metadata fees.

dmkozh:
# Description

Resolves https://github.com/stellar/stellar-core/issues/3693

Implement refunds for the Soroban metadata fees.

The refunds are recorded as `changesAfter` tx meta entries.

# Checklist
- [ ] Compiles
- [ ] Ran all tests

### DIFF
--- a/src/ledger/LedgerManagerImpl.cpp
+++ b/src/ledger/LedgerManagerImpl.cpp
@@ -1304,6 +1304,7 @@ LedgerManagerImpl::applyTransactions(
                    tx->getSeqNum(),
                    mApp.getConfig().toShortString(tx->getSourceID()));
         tx->apply(mApp, ltx, tm);
+        tx->processPostApply(mApp, ltx, tm);
         TransactionResultPair results;
         results.transactionHash = tx->getContentsHash();
         results.result = tx->getResult();
--- a/src/transactions/FeeBumpTransactionFrame.cpp
+++ b/src/transactions/FeeBumpTransactionFrame.cpp
@@ -155,6 +155,14 @@ FeeBumpTransactionFrame::apply(Application& app, AbstractLedgerTxn& ltx,
     }
 }
 
+void
+FeeBumpTransactionFrame::processPostApply(Application& app,
+                                          AbstractLedgerTxn& ltx,
+                                          TransactionMetaFrame& meta)
+{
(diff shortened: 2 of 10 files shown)


src/ledger/LedgerManagerImpl.cpp
```diff
@@ -1304,6 +1304,7 @@ LedgerManagerImpl::applyTransactions(
                    tx->getSeqNum(),
                    mApp.getConfig().toShortString(tx->getSourceID()));
         tx->apply(mApp, ltx, tm);
+        tx->processPostApply(mApp, ltx, tm);
```
MonsieurNicolas, line 1307: I think  it's better to perform this `PostApply` there so that changes get merged as needed: https://github.com/stellar/stellar-core/blob/d9ecb7e727eb24a046bafc7878ef22466c268e71/src/transactions/TransactionFrame.cpp#L1349

if you do it like this, you end up with the same ledger entry (the account paying for the fee) being modified twice in the same `LedgerEntryChanges` if the code at the other location does something (signers are in the same ledger entry), which wastes resources and may potentially confuse downstream systems.

MonsieurNicolas (changes requested):
This looks good to me. I have a suggestion in there to avoid introducing a new subtle behavior within the meta

src/ledger/LedgerManagerImpl.cpp
dmkozh: Currently the only other case for changes after is pre-V10, so there is no conflict with Soroban. Going forward I would actually prefer to have post-apply changes to be called post-apply and not as a part of it. E.g. in this case if we put the refund logic inside apply, then we might omit refund in case of internal tx error (which is of course unlikely, but semantically it still doesn't seem right to not do the refund in such case). 

FWIW I've also tried to put this into apply, but besides the awkward failure cases, it is also inconvenient for tests - we charge fees and apply txs separately and hence if the test just applies the tx without charging the fee, it might produce some XLM, which is confusing.

MonsieurNicolas:
r+ 8e06280eb117c542f8ad6965e72d987fae3dc1a9

src/ledger/LedgerManagerImpl.cpp
MonsieurNicolas: yes makes sense thank you

PR #3743, opened 2023-05-22
Correct initial network configs for contract cost

jayz22:
# Description

Correct the network config parameters for  `VmInstantiation` contract cost to match the host [defaults](https://github.com/stellar/rs-soroban-env/blob/main/soroban-env-host/src/budget.rs#L565). 
Previously it was mis-configured due to copy-paste error which could result in contracts running over the limits. 

# Checklist
- [ ] Compiles
- [ ] Ran all tests

### DIFF
--- a/src/ledger/NetworkConfig.cpp
+++ b/src/ledger/NetworkConfig.cpp
@@ -212,7 +212,7 @@ initialCpuCostParamsEntry()
             break;
         case VmInstantiation:
             params[val] =
-                ContractCostParamEntry{805445, 307, ExtensionPoint{0}};
+                ContractCostParamEntry{1'000'000, 0, ExtensionPoint{0}};
             break;
         case InvokeVmFunction:
             params[val] = ContractCostParamEntry{6212, 0, ExtensionPoint{0}};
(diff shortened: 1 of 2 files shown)


MonsieurNicolas:
>  Previously it was mis-configured due to copy-paste error which could result in contracts running over the limits.

this does not seem right: the settings defined in core should always be used by the host when enforcing limits.

Is there something else going on here?

jayz22:
> > Previously it was mis-configured due to copy-paste error which could result in contracts running over the limits.
> 
> this does not seem right: the settings defined in core should always be used by the host when enforcing limits.
> 
> Is there something else going on here?

So these values were copied over from https://github.com/stellar/rs-soroban-env/blob/main/soroban-env-host/src/budget.rs#LL566C27-L566C27, where were calibrated and tested since the last release.

The values in core now is the source of truth and will decide host metering. But with this error (there is a linear term in both cpu and memory w.r.t the contract size in the core version), this will lead to significant over-counting of contract invocation cost and likely cause some contracts to fail. 

Note these initial values can be overridden by network config upgrade but this PR is to fix the default initial values.

dmkozh:
r+ 96f0c84

PR #3744, opened 2023-05-22
Reject txs that are too big before placing them into outbound queues

marta-lokhova:


mbsdf:
cc: @marta-lokhova 

what are the chances of rejecting a valid classic transaction? The current MAX_CLASSIC_TX_SIZE_BYTES is 100KB for classic transactions (flow control in bytes enabled). One may have a large claimable balance in transaction and can go over this limit (100 operations per transaction). However they will be able to resubmit the same by splitting into multiple transactions.  Historical data shows we are within this limits.

SirTyson (approved):
LGTM

marta-lokhova:
r+ dcfae95476b988f00113cfd02e909c6f82a7d237

PR #1570, opened 2018-02-27
Delay publish by one ledger

vogel:
Fixes #1268

Issue #448, opened 2015-04-15
Block credits with XLM currency code


thejollyrogers:
In order to prevent potential confusion, and to maintain the sanctity of XLM, stellar-core should reject transactions that attempt to send credit with the currency code XLM.

MonsieurNicolas:
on that front, do we want to make currency codes case insensitive and low
ASCII only? Right now I don't think we enforce anything.
This includes things like use of "space" or other special characters in
there.

irisli:
+1 on having a clear defined currency code case (as to not repeat https://github.com/stellar/stellard/issues/90 )

I think this is a must have before the live network is launched so that we don't have to do backwards support for ledger cruft.

Drawing from inspiration: http://counterparty.io/docs/counterparty_features/#creating-assets
Counterparty limits it to: `A unique string of 4 to 12 uppercase Latin characters (inclusive) not beginning with ‘A’. `

I think whatever we do, we just need a very clear definition

MonsieurNicolas:
We have a different currency type for custom length asset strings.
For ISO like, I'd say we can go with something simple that will minimize chances of collisions:
mandatory 3 characters (A-Z only), with an optional 1 character prefix (A-Za-z).
so that people can do things like mUSD as we don't have exponents for those type of currencies.

thejollyrogers:
"do things like mUSD"
That doesn't make sense. Issue a credit that's a fraction of another
credit, but specified in the name of the currency?
That is really weird and doesn't work in the system. For example this would
mean that I could set a trustline to issuer A for
mBTC but someone with a trustline to issuer A for BTC wouldn't be able to
hold my mBTC credits. There would need to be an
exchange for each fractional type.

MonsieurNicolas:
You would most likely not have people trade many of these. I used mUSD but that was misleading - it would be for some currency ABC where the exchange rate is so high that we drive balances out of bound (ie, can't represent 1,000,000 BTC in this other currency).

Alternative is to use one as the bytes as a signed exponent (base 10) to apply to the stored balance if this makes it clearer.

Note that this is a client side thing: core doesn't care about multipliers and where the fixed point precision is (math is performed on opaque 64 bit "amounts"), but we still want to have some standardization on how amounts are used in clients to avoid complexity.

jedmccaleb:
I can see making it case insensitive although I don't really see the issue with allowing whatever. You are trusting the gateway in the first place so them "tricking" you into holding some weird currency code is the least of your worries.

[...]

PR #3750, opened 2023-05-23
Add offline-close command for establishing test network state in quickstart.

graydon:
This adds a new command `offline-close` which just cranks stellar-core forward from wherever it currently is by N ledgers, closing them with no content and waiting for publishes to complete. It's intended for setting up network state when booting the quickstart image, which really wants to start "after the first checkpoint" rather than at genesis. This is therefore an alternative to https://github.com/stellar/quickstart/pull/448 (and a cluster of related problems)

The only tricky part in here is that `ResolveSnapshotWork` had a 1-ledger delay in it (made by Rafał back in https://github.com/stellar/stellar-core/pull/1570) and it was applied unconditionally, which would make this wedge when trying to advance past a ledger-close and waiting for the publish associated with that close to complete. Rather than do acrobatics about going-one-past-and-waiting, I decided to just make this 1-ledger delay conditional on not-being-`STANDALONE`, since it's a meaningless check in that case anyways.

### DIFF
--- a/docs/software/commands.md
+++ b/docs/software/commands.md
@@ -93,6 +93,10 @@ Command options can only by placed after command.
   HISTORY-LABEL. HISTORY-LABEL should be one of the history archives you have
   specified in the stellar-core.cfg. This will write a
   `.well-known/stellar-history.json` file in the archive root.
+* **offline-close**: Forces stellar-core to close a specified number of empty
+  ledgers, strictly offline and starting from its current state, generating and
+  publishing history as it goes. Should only be used for special scenarios like
+  setting up test networks with artificial history.
 * **offline-info**: Returns an output similar to `--c info` for an offline
   instance, but written directly to standard output (ignoring log levels).
 * **print-xdr <FILE-NAME>**:  Pretty-print a binary file containing an XDR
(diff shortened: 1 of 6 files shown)


leighmcculloch (approved):
👏🏻

leighmcculloch:
This may not cause us to undo https://github.com/stellar/quickstart/pull/448, because its reasoning was two fold, it also made default the accelerated mode I expect folks will benefit from in dev and testing.

However, it would open up the possibility for us to have quickstart run an accelerated mode that did a checkpoint less frequently while still keeping ledger closes fast. So maybe we could have quickstart run accelerated with ledgers every one second, as it does now, but with checkpoints at every 64th, or 320th, ledger.

@graydon Do you think that's possible?

Of course, it would also allow us to undo #448 which removed the ability to control acceleration, and add that control back with the default being on but the possibility for a user to disable it. In which case, this new feature of stellar-core would allow us to skip the 5 min wait time when acceleration is disabled.

marta-lokhova (approved):

marta-lokhova:
r+ e41ec89fad4c7832e2ab1cce38c663150a35ff84

PR #3751, opened 2023-05-23
Make sed usage more compat with git refs

leighmcculloch:
# Description

Make `sed` usage more compat with git refs where we use sed for setting the version, which is a git ref.

Git refs can contain `/`, and so if the version of stellar-core is a build of stellar-core from a branch that has `/` in the name, the sed command will value because the sed command is using `/` as the separator.

Using an alternative separator that is guaranteed to never be found in git refs eliminates this conflict.

Related discussion is at https://stellarfoundation.slack.com/archives/C0367MGFL/p1684881613131079.

# Checklist
- [ ] Compiles
- [ ] Ran all tests

### DIFF
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -125,7 +125,7 @@ main/StellarCoreVersion.cpp: always
 	@vers=$$(cd "$(srcdir)" \
 		&& git describe --always --dirty --tags 2>/dev/null \
 		|| echo "$(PACKAGE) $(VERSION)"); \
-		sed -e "s/%%VERSION%%/$$vers/" \
+		sed -e "s@%%VERSION%%@$$vers@" \
 			< "$(srcdir)/main/StellarCoreVersion.cpp.in" > $@~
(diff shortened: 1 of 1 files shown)


leighmcculloch:
cc @sreuland This is what caused `sed` errors when you, and I, attempted to build stellar-core using a `pulls/N/head` reference.

leighmcculloch:
Tested with this command to confirm it worked:

```bash
docker build \
  -f docker/Dockerfile.testing \
  --build-arg BUILDKIT_CONTEXT_KEEP_GIT_DIR=true \
  https://github.com/stellar/stellar-core.git\#pull/3751/head
```

In the build output I saw this message, confirming the replace works:

```
 => => # echo '#include "main/StellarCoreVersion.h"                                                         
 => => # const std::string STELLAR_CORE_VERSION = "pull/3751/head";' > main/StellarCoreVersion.cpp         
```

marta-lokhova (approved):

PR #3753, opened 2023-05-24
Avoid BUILD_TESTS code in offline-close, fix other nits

graydon:
This just fixes the recent `offline-close` command to not use any of the `BUILD_TESTS` machinery. While I was here I also changed it to emit log messages (avoiding the herder means it otherwise runs silent), crank nonblocking (so it runs faster), allow early shutdown with ctrl-C and emit a warning about being out of consensus with any peers when it's done.


marta-lokhova:
r+ 8987c66647014fed406f18ea9bb245cc62c6c6aa

marta-lokhova (approved):

PR #3754, opened 2023-05-26
Ensure we can always find txs to evict in tx limiter

marta-lokhova:
Currently, it is possible to hit [this assert](https://github.com/stellar/stellar-core/blob/7fb6d5e8858fed6ea365f8717b0266635f578477/src/herder/SurgePricingUtils.cpp#L431) during tx limiting if we have a more expensive transaction, but not enough ops to evict from the limited lane. This is because we stop iterating over the generic lane too early, preventing the expensive transaction from fitting in the generic limit. Solution is to allow evicting cheaper transactions from the generic lane, therefore preserving the invariant that we can always find txs to evict (assuming pre-conditions pass). Adding an extra test to confirm the failure, and the fix.

### DIFF
--- a/src/herder/SurgePricingUtils.cpp
+++ b/src/herder/SurgePricingUtils.cpp
@@ -398,11 +398,11 @@ SurgePricingPriorityQueue::canFitWithEviction(
             // there is not enough ops to evict from limited lane to fit the
             // current transaction
             if (lane == GENERIC_LANE || lane == evictLane ||
-                neededLaneOps <= 0 || neededTotalOps > neededLaneOps)
+                neededTotalOps > neededLaneOps)
             {
                 canEvict = true;
             }
-            else if (neededTotalOps <= neededLaneOps)
+            else
             {
                 // When we need to evict more ops from the tx's lane than from
                 // the generic lane, then we can only evict from tx's lane (any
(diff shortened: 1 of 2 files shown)


src/herder/SurgePricingUtils.cpp
```diff
@@ -389,14 +389,16 @@ SurgePricingPriorityQueue::canFitWithEviction(
...
             // Check if it makes sense to evict the current top tx stack.
             //
             // Simple cases: generic lane txs can evict anything; same lane txs
             // can evict each other; any transaction can be evicted if per-lane
-            // limit hasn't been reached.
-            if (lane == GENERIC_LANE || lane == evictLane || neededLaneOps <= 0)
+            // limit hasn't been reached; any transaction can be evicted if
+            // there is not enough ops to evict from limited lane to fit the
+            // current transaction
+            if (lane == GENERIC_LANE || lane == evictLane ||
```
dmkozh, line 400: I think this can be further simplified to
```
if (lane == GENERIC_LANE || lane == evictLane || neededTotalOps > neededLaneOps)
{
    canEvict = true;
}
else
{
    evictedDueToLaneLimit = true;
}
```
`neededLaneOps <= 0` check is redundant (0 is less than any non-0 value of total ops)
`neededTotalOps <= neededLaneOps` has to be true if we get into else branch.

marta-lokhova: Indeed, fixed

marta-lokhova:
@dmkozh r+ if this looks good? :)

dmkozh:
r+ 4334daa

Issue #3689, opened 2023-04-04
Soroban-specific transaction prioritization


dmkozh:
Core prioritizes and throttles transactions in a few places. Soroban transactions are subject to a different comparison approach compared to all the other ('classic') transactions, as per https://github.com/stellar/stellar-protocol/blob/master/core/cap-0046-07.md.

The Soroban transactions should be comparted using the inclusion fee (instead of the transaction fee per operation) and limited not by the transaction count, but by the sum of resources that are limited per ledger: total instruction count, total ledger IO and sum of transaction sizes (the latter is probably only applicable for flooding/limiting). 

`SurgePricingPriorityQueue` should either be modified or forked to handle the Soroban limits and comparisons. The same goes for its users: tx queue limiter, tx queue itself, tx set nomination. Since Soroban transactions can't be directly compared to the classic transactions, they need to be handled separately for each of these users and in the corresponding use cases (e.g. in case of transaction queue Soroban transactions should be flooded separately from the classic transactions or during tx set building Soroban transactions should be compared with each other and probably belong to a separate phase).

This issue is for tracking these data structure and algorithm changes in Core, without considering any specific limits/constants etc. The scope of work/changes is similar to https://github.com/stellar/stellar-core/pull/3508/files#diff-ce22e2ef03fa436ff55f680514d24c84a5f755e7e39242315956dc87d368c0a1 or https://github.com/stellar/stellar-core/pull/3481/files#diff-656b2f63bc75a5971ad0f893e88fe8df2bce5ee3df5f1fbdce4526dce62c2a88

PR #3758, opened 2023-05-31
Flooding, tx limiting and tx set construction strategy for Soroban transactions

marta-lokhova:
This PR implements the flooding/tx limiting portion of https://github.com/stellar/stellar-core/issues/3689 by updating SurgePricingQueue to support limit tuples (previously we had a one-dimensional limit, aka op count). Herder now maintains two separate tx queues so limiting and flooding of Soroban and classic txs are done independently based on fees in each group. 

This PR is very much a draft, as I'm still hashing out bugs and doing cleanups/strengthening of the code, so some parts are messy/error-prone rn. I'm continuing to make improvements, and starting to think about how we'll deal with tx set construction, but wanted to at least get this change out to make sure this is going in the right direction.


marta-lokhova:
need to audit usage of vNext flag, and ensure current protocol isn't impacted by accident

src/herder/test/HerderTests.cpp
```diff
@@ -3960,6 +3962,176 @@ TEST_CASE("do not flood invalid transactions", "[herder]")
     REQUIRE(txSet->checkValid(*app, 0, 0));
 }
 
+#ifdef ENABLE_NEXT_PROTOCOL_VERSION_UNSAFE_FOR_PRODUCTION
```
marta-lokhova, line 3965: Currently adding another test here that mixes Soroban and classic txs in the same ledger (to ensure two queues are working as expected)

PR #3766, opened 2023-06-07
Flooding, tx limiting and tx set construction strategy for Soroban

marta-lokhova:
Implementation of https://github.com/stellar/stellar-core/issues/3689


src/herder/TxSetFrame.cpp
```diff
@@ -88,72 +91,80 @@ class InvalidTxSetFrame : public TxSetFrame
...
 {
     if (txSet.v() != 1)
     {
         CLOG_DEBUG(Herder, "Got bad txSet: unsupported version {}", txSet.v());
         return false;
     }
     auto const& txSetV1 = txSet.v1TxSet();
-    if (txSetV1.phases.size() != 1)
+    // There was no protocol with 1 phase, so checking for 2 phases only
+    if (txSetV1.phases.size() != 2)
```
dmkozh, line 104: nit: Use a constant you defined in the header?

src/herder/TxSetFrame.cpp
```diff
@@ -283,7 +349,7 @@ TxSetFrame::makeFromHistoryTransactions(Hash const& previousLedgerHash,
     // This may cause leaks in case of exceptions, so keep the constructors
     // simple and exception-safe.
     return std::shared_ptr<TxSetFrame>(
-        new TxSetFrame(false, previousLedgerHash, txs));
+        new TxSetFrame(false, previousLedgerHash, TxSetFrame::TxPhases{txs}));
```
dmkozh, line 352: This looks suspicious - wouldn't history contain 2 phases after protocol upgrade?

src/herder/TxSetFrame.cpp
```diff
@@ -832,10 +998,12 @@ TxSetFrame::isGeneralizedTxSet() const
 bool
 TxSetFrame::addTxsFromXdr(Application& app,
                           xdr::xvector<TransactionEnvelope> const& txs,
-                          bool useBaseFee, std::optional<int64_t> baseFee)
+                          bool useBaseFee, std::optional<int64_t> baseFee,
+                          Phase phase)
```
dmkozh, line 1002: I think this is a good place to verify phase consistency (soroban phase should have only soroban txs and vice versa). Maybe I've missed the check elsewhere though...

src/herder/TxSetFrame.cpp
```diff
@@ -398,56 +481,66 @@ TxSetFrame::previousLedgerHash() const
     return mPreviousLedgerHash;
 }
 
-TxSetFrame::Transactions const&
+TxSetFrame::Transactions
```
dmkozh, line 484: Just a consideration: if we use this frequently we might consider using a visitor instead in order to avoid copying the whole tx set.

src/herder/test/HerderTests.cpp
```diff
@@ -1519,12 +1526,146 @@ TEST_CASE("surge pricing", "[herder][txset]")
...
+                REQUIRE(phase.empty());
+            }
+            REQUIRE(txSet->sizeTx() == 3);
+            for (auto const& tx : txSet->getTxs())
+            {
+                // smallSorobanLowFee was picked over sorobanTx to fill the gap
+                REQUIRE(tx != sorobanTx);
+            }
+        }
+        SECTION("soroban tx limits")
```
dmkozh, line 1662: Just a reminder this still is not done - ok to do as a followup (probably do a smoke test where we create 100-1000 random tx sets that preserve the limits?).

src/herder/TxQueueLimiter.cpp
```diff
@@ -153,12 +124,34 @@ TxQueueLimiter::canAddTx(TransactionFrameBasePtr const& newTx,
...
+    return canAddTx(newTx, oldTx, txsToEvict, ltx);
+}
+
+std::pair<bool, int64>
+TxQueueLimiter::canAddTx(TransactionFrameBasePtr const& newTx,
+                         TransactionFrameBasePtr const& oldTx,
+                         std::vector<std::pair<TxStackPtr, bool>>& txsToEvict,
+                         AbstractLedgerTxn& ltxOuter)
+{
+    // Sanity check: oldTx is either not provided, or we're dealing with classic
```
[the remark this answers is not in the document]
marta-lokhova: you're right, this was a remnant from the first version, where I wasn't sure what to do with fee bumps. Removed.

src/herder/HerderImpl.cpp
```diff
@@ -450,7 +460,40 @@ TransactionQueue::AddResult
...
+
+    // Allow txs of the same kind to reach the tx queue in case it can be
+    // replaced by fee
+    bool hasSoroban =
+        mSorobanTransactionQueue.sourceAccountPending(tx->getSourceID()) &&
+        !tx->isSoroban();
+    bool hasClassic =
+        mTransactionQueue.sourceAccountPending(tx->getSourceID()) &&
+        tx->isSoroban();
+    bool reject = mApp.getConfig().LIMIT_TX_QUEUE_SOURCE_ACCOUNT &&
```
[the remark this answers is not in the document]
marta-lokhova: Yes, given that having two queues breaks any assumptions about source account tx chains, we should disallow core to run protocol 20 without the flag. We should enforce it prior to protocol 20 upgrade so that we don't need to drop any chains at upgrade time (we're planning to do this in the upcoming proto 19 core release). For now, to force it on futurenet, I'm going to add a check on startup behind the vNext flag.

src/herder/HerderImpl.cpp
```diff
@@ -1089,6 +1139,13 @@ HerderImpl::getMinLedgerSeqToAskPeers() const
 SequenceNumber
 HerderImpl::getMaxSeqInPendingTxs(AccountID const& acc)
 {
+#ifdef ENABLE_NEXT_PROTOCOL_VERSION_UNSAFE_FOR_PRODUCTION
```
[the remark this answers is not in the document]
marta-lokhova: I believe it should be fine to keep this code as-is, assuming misconfiguration handling I mentioned above.

src/util/TxResource.cpp
```diff
@@ -0,0 +1,186 @@
...
+    return false;
+}
+
+Resource
+multiplyByDouble(Resource const& res, double m)
+{
+    auto newRes = res;
+    for (auto& resource : newRes.mResources)
+    {
+        auto tempResultDbl = resource * m;
```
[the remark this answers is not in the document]
marta-lokhova: I extracted this logic from our current tx queue broadcasting mechanism: 
https://github.com/stellar/stellar-core/blob/287fe4c5d7f7ebce1a327cd26f25dd5858cd37a0/src/herder/TransactionQueue.cpp#L1002

I can see two sides of the argument here: on one hand, setting FLOOD_OP_RATE_PER_LEDGER too low may lead to nothing being flooded at all. On the other hand, if we round up, then we may not respect the configured rate, and "overbroadcast". I think it's fine to keep it as-is: if the config is so low that we don't broadcast anything, most likely it's an operator error (in which case the recovery is easy - just need to update the config)

src/simulation/LoadGenerator.cpp
```diff
@@ -719,9 +767,16 @@ LoadGenerator::maybeHandleFailedTx(TestAccountPtr sourceAccount,
...
         code == txBAD_SEQ)
     {
-        auto const& txQueue = mApp.getHerder().getTransactionQueue();
         auto txQueueSeqNum =
-            txQueue.getInQueueSeqNum(sourceAccount->getPublicKey());
+#ifdef ENABLE_NEXT_PROTOCOL_VERSION_UNSAFE_FOR_PRODUCTION
+            tx->isSoroban()
+                ? mApp.getHerder()
+                      .getSorobanTransactionQueue()
+                      .getInQueueSeqNum(sourceAccount->getPublicKey())
```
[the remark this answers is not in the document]
marta-lokhova: We only expose the queues for testing (the getters are behind BUILD_TESTS), but a helper makes loadgen code less ugly, so I added it anyway

src/herder/TxSetFrame.cpp
marta-lokhova: This logic is only used for old-style transaction sets (here when creating TxSetFrame, we pass false for isGeneralized). With the protocol upgrade, we're going to switch to GeneralizedTxSets, and for historical purposes we actually don't reconstruct the TxSetFrame object+convert it to XDR, but instead use the encoded tx set stored in the DB, so it should deal with phases correctly. Please let me know if this makes sense.

src/herder/TxSetFrame.cpp
marta-lokhova: I ended up changing this function to return const ref to txs in a particular phase - its probably safer to force callers to deal with individual phases instead of pooling all txs together.

src/ledger/test/LedgerCloseMetaStreamTests.cpp
```diff
@@ -434,113 +434,131 @@ TEST_CASE("METADATA_DEBUG_LEDGERS works", "[metadebug]")
...
-                REQUIRE(inJson);
-                std::string expect(std::istreambuf_iterator<char>{inJson}, {});
-                REQUIRE(expect == have);
-            }
-        }
-        maxSeq = ledgerSeq;
-    }
-    REQUIRE(maxSeq == targetSeq);
-}
+// TEST_CASE_VERSIONS("meta stream contains reasonable meta",
```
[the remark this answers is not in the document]
marta-lokhova: Ah yes, so this broke because I added a new phase to generalized tx sets in protocol 20, so I needed to regenerate the meta. Fixed.

src/herder/TxSetFrame.cpp
marta-lokhova: Fixed.

src/util/TxResource.h
```diff
@@ -0,0 +1,103 @@
...
+#include <vector>
+
+namespace stellar
+{
+
+constexpr size_t NUM_CLASSIC_TX_RESOURCES(1);
+constexpr size_t NUM_SOROBAN_TX_RESOURCES(6);
+
+// Small helper class to allow arithmetic operations on tuples
+class Resource
```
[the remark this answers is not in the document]
marta-lokhova: Agreed. This exact change was in my TODO list actually, but I didn't get to it because it was a bit annoying to propagate it to all the users of Resource (tx queue, limiter, etc). I do think it's worth doing to avoid any footguns. I'll submit a patch as a follow-up after we merge this change.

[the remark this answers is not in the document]
dmkozh: Sounds good, we can do that as a followup after the preview release (if at all) - when I left this comment I didn't know the scope of usage of this class. Propagating it everywhere would likely be really awkward; maybe the runtime checks are good enough of a tradeoff here.

src/herder/test/HerderTests.cpp
marta-lokhova: thanks, I'll fix the test in a subsequent PR (maybe we'll have soroban tx limit incorporated in the network config by then).

src/herder/HerderImpl.cpp
```diff
@@ -450,7 +459,40 @@ TransactionQueue::AddResult
 HerderImpl::recvTransaction(TransactionFrameBasePtr tx, bool submittedFromSelf)
 {
     ZoneScoped;
+#ifdef ENABLE_NEXT_PROTOCOL_VERSION_UNSAFE_FOR_PRODUCTION
+    TransactionQueue::AddResult result;
+
+    // Allow txs of the same kind to reach the tx queue in case it can be
```
dmkozh, line 465: I'm not sure I understand this logic, do we actually replace txs with the same seq num? Even if we do, I'm not sure if it's correct to only allow one certain kind of transactions...

src/herder/TxSetFrame.cpp
dmkozh: Ah, right, I just forgot what this is used for (this doesn't have components either).

src/herder/TxSetFrame.cpp
```diff
@@ -659,12 +637,22 @@ TxSetFrame::checkValid(Application& app, uint64_t lowerBoundCloseTimeOffset,
...
     {
-        CLOG_DEBUG(Herder, "Got bad txSet: too many txs {} > {}",
-                   this->size(lcl.header), lcl.header.maxTxSetSize);
+        CLOG_DEBUG(Herder, "Got bad txSet: too many classic txs {} > {}",
+                   this->size(lcl.header, Phase::CLASSIC),
+                   lcl.header.maxTxSetSize);
         return false;
     }
+#ifdef ENABLE_NEXT_PROTOCOL_VERSION_UNSAFE_FOR_PRODUCTION
+    if (this->size(lcl.header, Phase::SOROBAN) > SOROBAN_TX_LIMIT_PER_LEDGER)
```
dmkozh, line 648: classic->Soroban in the debug log

Also, 2 seems too low (though probably doesn't really matter since we're going to update this soon.

src/herder/TxSetFrame.cpp
```diff
@@ -681,27 +669,56 @@ TxSetFrame::checkValid(Application& app, uint64_t lowerBoundCloseTimeOffset,
...
+TxSetFrame::size(LedgerHeader const& lh, std::optional<Phase> phase) const
 {
-    return protocolVersionStartsFrom(lh.ledgerVersion, ProtocolVersion::V_11)
-               ? sizeOp()
-               : sizeTx();
+    size_t sz = 0;
+#ifdef ENABLE_NEXT_PROTOCOL_VERSION_UNSAFE_FOR_PRODUCTION
+    if (!phase && phase == Phase::SOROBAN)
+    {
+        sz += sizeTx(Phase::SOROBAN);
```
dmkozh, line 678: I think it's better to use `sizeOp` here for the sake of consistency (I realize they're equivalent, that's just for clarity)

src/test/TestAccount.cpp
```diff
@@ -149,6 +149,7 @@ TestAccount::create(SecretKey const& secretKey, uint64_t initialBalance)
     }
     catch (...)
     {
+        throw;
```
dmkozh, line 152: Is this intentional?

src/test/TxTests.cpp
```diff
@@ -822,6 +822,35 @@ createSimpleDexTx(Application& app, TestAccount& account, uint32 nbOps,
                                      ops, fee);
 }
 
+#ifdef ENABLE_NEXT_PROTOCOL_VERSION_UNSAFE_FOR_PRODUCTION
+TransactionFramePtr
+createSimpleDeployContractTx(Application& app, TestAccount& account,
```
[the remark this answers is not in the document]
dmkozh: Seems like this hasn't been updated

src/herder/TxSetFrame.cpp
```diff
@@ -851,15 +1076,19 @@ TxSetFrame::addTxsFromXdr(Application& app,
 #ifdef ENABLE_NEXT_PROTOCOL_VERSION_UNSAFE_FOR_PRODUCTION
         tx->maybeComputeSorobanResourceFee(ledgerVersion, sorobanConfig,
                                            appConfig);
+        // Phase should be consistent with the tx we're trying to add
+        if (tx->isSoroban() && phase != Phase::SOROBAN)
```
dmkozh, line 1080: What if the opposite condition is true? Should this be just `if (tx->isSoroban() != Phase::SOROBAN)`?

src/herder/TxSetFrame.cpp
marta-lokhova: yeah, the number doesn't matter, this variable will go away as soon as merge my other PR anyway, I just wanted to have a placeholder, so we don't forget to have this check.

src/herder/HerderImpl.cpp
marta-lokhova: discussed offline, and agreed on the correctness of this logic. Herder cross checks the two queues, but still allows txs of the same kind to enter the respective queue in case it's a fee bump. If the tx can't actually replace an older tx, then the queue will return an appropriate error.

src/herder/TxSetFrame.cpp
marta-lokhova: good catch. i'd prefer to be explicit here though, so changed to `((tx->isSoroban() && phase != Phase::SOROBAN) || (!tx->isSoroban() &&  phase != Phase::CLASSIC))`

src/herder/TxSetFrame.cpp
marta-lokhova: Done.

marta-lokhova:
Alright, I believe all feedback has been addressed now (this time for real!). Just waiting for the CI is go through.

dmkozh:
r+ 2978cc9

PR #3771, opened 2023-06-12
Wire up soroban PRNG seed from txset hash, tx and op numbers.

graydon:
This does a very basic (and not at all tested) wiring-up of the soroban PRNG service based on the txset hash, tx and op numbers as we run a tx. I figured it's small enough to sneak in to the preview 10 release and, if there's a problem with it, it won't at this point be any worse than the current situation (which, being unseeded, will [just fault if someone tries to use it](https://github.com/stellar/rs-soroban-env/blob/main/soroban-env-host/src/host/frame.rs#L270-L275)).


jayz22 (approved):

dmkozh:
r+ 20cc4e3

Issue #3768, opened 2023-06-09
SCP sharing code sets up a global timer incorrectly


marta-lokhova:
Pointed out by @MonsieurNicolas 

https://github.com/stellar/stellar-core/blob/287fe4c5d7f7ebce1a327cd26f25dd5858cd37a0/src/herder/HerderImpl.cpp#L808

This code is triggered by an individual peer, yet Herder sets up one global timer to track the delay. Instead, each peer should have its own timer to avoid races.

marta-lokhova:
Resolved via https://github.com/stellar/stellar-core/pull/3789

PR #3789, opened 2023-06-22
fix global early catch up timer

JuI3s:
#3768 Description

Fix issue global early catchup timer issue in Herder implementation. 

# Checklist
- [ ] Compiles
- [ ] Ran all tests

### DIFF
--- a/src/herder/HerderImpl.cpp
+++ b/src/herder/HerderImpl.cpp
@@ -91,13 +91,13 @@ HerderImpl::HerderImpl(Application& app)
     , mTriggerTimer(app)
     , mOutOfSyncTimer(app)
     , mTxSetGarbageCollectTimer(app)
-    , mEarlyCatchupTimer(app)
     , mApp(app)
     , mLedgerManager(app.getLedgerManager())
     , mSCPMetrics(app)
     , mState(Herder::HERDER_BOOTING_STATE)
 {
     auto ln = getSCP().getLocalNode();
+
     mPendingEnvelopes.addSCPQuorumSet(ln->getQuorumSetHash(),
                                       ln->getQuorumSet());
 }
@@ -246,7 +246,6 @@ HerderImpl::shutdown()
     mTrackingTimer.cancel();
     mOutOfSyncTimer.cancel();
     mTriggerTimer.cancel();
-    mEarlyCatchupTimer.cancel();
     if (mLastQuorumMapIntersectionState.mRecalculating)
     {
         // We want to interrupt any calculation-in-progress at shutdown to
@@ -861,9 +860,8 @@ HerderImpl::sendSCPStateToPeer(uint32 ledgerSeq, Peer::pointer peer)
     // ledger to achieve this
     if (delayCheckpoint)
     {
-        mEarlyCatchupTimer.expires_from_now(
-            Herder::SEND_LATEST_CHECKPOINT_DELAY);
-        mEarlyCatchupTimer.async_wait(
+        peer->startExecutionDelayedTimer(
+            Herder::SEND_LATEST_CHECKPOINT_DELAY,
             [checkpoint, this, sendSlot]() {
                 getSCP().processCurrentState(
                     checkpoint,
--- a/src/herder/test/HerderTests.cpp
+++ b/src/herder/test/HerderTests.cpp
@@ -14,6 +14,8 @@
 #include "test/TestUtils.h"
 #include "test/test.h"
 
+#include "history/test/HistoryTestsUtils.h"
+
 #include "catchup/CatchupManagerImpl.h"
 #include "crypto/SHA.h"
 #include "database/Database.h"
@@ -48,6 +50,7 @@
 using namespace stellar;
 using namespace stellar::txbridge;
 using namespace stellar::txtest;
+using namespace historytestutils;
 
 TEST_CASE_VERSIONS("standalone", "[herder][acceptance]")
 {
@@ -3220,8 +3223,11 @@ TEST_CASE("SCP checkpoint", "[catchup][herder]")
     auto simulation =
         std::make_shared<Simulation>(Simulation::OVER_LOOPBACK, networkID);
 
+    auto histCfg = std::make_shared<TmpDirHistoryConfigurator>();
+
     SIMULATION_CREATE_NODE(0);
     SIMULATION_CREATE_NODE(1);
+    SIMULATION_CREATE_NODE(2);
 
     SCPQuorumSet qSet;
     qSet.threshold = 1;
@@ -3229,9 +3235,17 @@ TEST_CASE("SCP checkpoint", "[catchup][herder]")
 
     Config cfg1 = getTestConfig(1);
     Config cfg2 = getTestConfig(2);
+    Config cfg3 = getTestConfig(3);
+
     cfg2.FORCE_SCP = false;
     cfg2.MODE_DOES_CATCHUP = true;
+    cfg3.FORCE_SCP = false;
+    cfg3.MODE_DOES_CATCHUP = true;
     cfg1.MODE_DOES_CATCHUP = false;
+
+    cfg1 = histCfg->configure(cfg1, true);
+    cfg3 = histCfg->configure(cfg3, false);
+    cfg2 = histCfg->configure(cfg2, false);
 
     auto mainNode = simulation->addNode(v0SecretKey, qSet, &cfg1);
     simulation->startAllNodes();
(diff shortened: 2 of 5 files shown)


src/herder/HerderImpl.cpp
```diff
@@ -861,16 +860,20 @@ HerderImpl::sendSCPStateToPeer(uint32 ledgerSeq, Peer::pointer peer)
...
-            Herder::SEND_LATEST_CHECKPOINT_DELAY);
-        mEarlyCatchupTimer.async_wait(
-            [checkpoint, this, sendSlot]() {
-                getSCP().processCurrentState(
-                    checkpoint,
-                    [&](SCPEnvelope const& e) { return sendSlot(e, true); },
-                    false);
-            },
-            &VirtualTimer::onFailureNoop);
+        auto weakPeer = std::weak_ptr<Peer>(peer);
```
SirTyson, line 863: It looks like we have redundant peer pointers. If you lock the weak pointer here, you should modify `sendSlot` to take `peerPtr` as an argument instead of creating and locking another weak_ptr again for the same object.

SirTyson (changes requested):

src/overlay/Peer.h
```diff
@@ -193,6 +197,8 @@ class Peer : public std::enable_shared_from_this<Peer>,
     VirtualClock::time_point mCreationTime;
 
     VirtualTimer mRecurringTimer;
+    VirtualTimer mEarlyCatchupTimer;
```
marta-lokhova, line 200: I know this was here before, but could we rename this variable? This timer triggers delayed execution of a function, and in our case the function then _maybe_ triggers another process, i.e. "early catchup". I think we can simply name the variable after what it does, `executeDelayed` or `scheduleExecution` or something like that.

marta-lokhova (changes requested):
Thank you for fixing this! Left a couple of suggestions

src/overlay/Peer.h
JuI3s: Named to it mExecutionDelayedTimer.

src/herder/HerderImpl.cpp
JuI3s: Just replaced with a using `peer`.

mbsdf (approved):

mbsdf:
@JuI3s  It will be good to have a unit test, to make sure each peer is having its own timer.  Before that we may want to have a test for multiple peers first.

@marta-lokhova @SirTyson please comment above ^^^

marta-lokhova:
@JuI3s We can probably add another section to the "SCP checkpoint" test (introduced in [early catchup PR](https://github.com/stellar/stellar-core/pull/3655/files#diff-677b80ed54027166ac32b9237769e289f0037dcf5793f76328345ce4c99453e6)). What we could do is have two out of sync nodes start at the same time, and request a checkpoint ledger. With your change, we should be able to verify that both out of sync nodes can catchup properly (in master right now, there's a race condition where one peer might override the timer of the other peer)

JuI3s:
> @JuI3s We can probably add another section to the "SCP checkpoint" test (introduced in [early catchup PR](https://github.com/stellar/stellar-core/pull/3655/files#diff-677b80ed54027166ac32b9237769e289f0037dcf5793f76328345ce4c99453e6)). What we could do is have two out of sync nodes start at the same time, and request a checkpoint ledger. With your change, we should be able to verify that both out of sync nodes can catchup properly (in master right now, there's a race condition where one peer might override the timer of the other peer)

I just added some a test for the two nodes case. I'm still trying to figure out how the test harness works. In this case when I create a new test section, do I need to create two new nodes or can I reuse the one in the previous test section for single node checkpoint? Is there any cleanup I need to do to for each test section?

marta-lokhova:
> I just added some a test for the two nodes case. I'm still trying to figure out how the test harness works. In this case when I create a new test section, do I need to create two new nodes or can I reuse the one in the previous test section for single node checkpoint? Is there any cleanup I need to do to for each test section?

Yeah, you'd need to create two new nodes, as sections are independent of each other (there's a bit of documentation on this here: https://github.com/catchorg/Catch2/blob/devel/docs/tutorial.md#test-cases-and-sections)

src/herder/test/HerderTests.cpp
```diff
@@ -3304,6 +3308,52 @@ TEST_CASE("SCP checkpoint", "[catchup][herder]")
...
+                2 * Herder::SEND_LATEST_CHECKPOINT_DELAY, false);
+        };
+
+        // History archves have not been configured, so this should throw once
+        // catchup starts
+        REQUIRE_THROWS_WITH(f(), "No GET-enabled history archive in config");
+
+        auto const& bufferedLedgers1 = cm1.getBufferedLedgers();
+        REQUIRE(!bufferedLedgers1.empty());
+        REQUIRE(bufferedLedgers1.begin()->first == firstCheckpoint);
```
marta-lokhova, line 3348: hmm, looks like this condition is failing
```

2	herder/test/HerderTests.cpp:3348: FAILED:
2	
2	  REQUIRE( bufferedLedgers1.begin()->first == firstCheckpoint )
2	
2	with expansion:
2	
2	  85 == 64
2	
```
This suggests that the out of sync node either didn't receive the checkpoint ledger or dropped it by the time we reached this check. Could you verify that both nodes actually receive the checkpoint ledger (IOW, that the main node calls this lambda for both out of sync nodes: https://github.com/stellar/stellar-core/blob/2109a168a895349f87b502ae3d182380b378fa47/src/herder/HerderImpl.cpp#L867)

JuI3s: > hmm, looks like this condition is failing
> 
> ```
> 
> 2	herder/test/HerderTests.cpp:3348: FAILED:
> 2	
> 2	  REQUIRE( bufferedLedgers1.begin()->first == firstCheckpoint )
> 2	
> 2	with expansion:
> 2	
> 2	  85 == 64
> 2	
> ```
> 
> This suggests that the out of sync node either didn't receive the checkpoint ledger or dropped it by the time we reached this check. Could you verify that both nodes actually receive the checkpoint ledger (IOW, that the main node calls this lambda for both out of sync nodes:
> 
> https://github.com/stellar/stellar-core/blob/2109a168a895349f87b502ae3d182380b378fa47/src/herder/HerderImpl.cpp#L867
> 
> )

`        auto f = [&]() {
            simulation->crankUntil(
                [&]() {
                    return cm1.isCatchupInitialized() &&
                           cm1.getCatchupWorkState() ==
                               BasicWork::State::WORK_RUNNING &&
                           cm2.isCatchupInitialized() &&
                           cm2.getCatchupWorkState() ==
                               BasicWork::State::WORK_RUNNING;
                },
                2 * Herder::SEND_LATEST_CHECKPOINT_DELAY, false);
        };
`

Is the crankUntil condition correct? I assume it checks that both nodes received catch up messages.

JuI3s: > 

I verified that the main node calls the lambda for both out of sync nodes.

SirTyson: It looks like the `SCP checkpoint` tests are still failing. I think this check isn't working `cm1.isCatchupInitialized() && cm1.getCatchupWorkState() == BasicWork::State::WORK_RUNNING` now that you've added the history archive config. It looks like the catchup state is stuck in `State::WORK_WAITING` and causing the `crankUntil` loop to timeout. I'm not 100% sure so maybe @marta-lokhova can follow up, but I think `getCatchupWorkState()` goes into `State::WORK_WAITING`while waiting for the history archive download. This didn't happen before because we didn't have a history archive to download from previously. If you change the loop to this it should avoid the timeout:
```
        simulation->crankUntil(
            [&]() {
                return cm.isCatchupInitialized();
            },
            2 * Herder::SEND_LATEST_CHECKPOINT_DELAY, false);
```

marta-lokhova: yeah, @SirTyson is right, removing `getCatchupWorkState` check should fix it (we still ensure that catchup was started with `isCatchupInitialized` check)

mbsdf: Thank you @SirTyson  and @marta-lokhova  Test will pass once ```getCatchupWorkState() ==BasicWork::State::WORK_RUNNING``` is deleted. Are we ready to merge this PR after that?

JuI3s: I just made a new commit removing the getCatchupWorkState check. Interestingly on my local machine that condition doesn't cause the test to fail?

SirTyson (approved):
LGTM

marta-lokhova (approved):
Looks good! nit: could you please squash the commits and we'll merge this (we typically try to keep the commit history cleaned up so it's easy to navigate/review/revert changes)

JuI3s:
> Looks good! nit: could you please squash the commits and we'll merge this (we typically try to keep the commit history cleaned up so it's easy to navigate/review/revert changes)

Just fixed it!

marta-lokhova:
r+ a2396521284b58db84168e2d48ef5d3c9ed2552c

Issue #3746, opened 2023-05-22
overlay: support Soroban max tx size upgrades


marta-lokhova:
now that https://github.com/stellar/stellar-core/pull/3737 is merged, we should add support for soroban network config upgrades at the overlay layer, and tweak throttling parameters appropriately.

mbsdf:
We should also ensure that MAX_CLASSIC_TX_SIZE_BYTES is set correctly (i.e. #3744) What are the chances of rejecting a valid classic transaction?

PR #3806, opened 2023-06-30
Support soroban upgrades in overlay

marta-lokhova:
Resolves https://github.com/stellar/stellar-core/issues/3746 + bump overlay version

The change potentially triggers upgrade handling every time the network externalizes a new ledger, automatically adjusting capacities to accommodate the new size. Note that when the capacity is increased, all peer connections need to be notified as well, which is done by sending an additional SEND_MORE message.

This upgrade handling applies only in scenarios where the maximum transaction size increases. No capacity adjustment is needed if the max size is decreased since the overlay can already handle those transactions. However, it's worth noting that values from the config file might become invalidated after the upgrade, as they need to maintain a specific invariance with the max tx size. To address this, I updated the code to automatically configure the values if they weren't specified by the operator. If an operator did provide config values, they will be validated during startup. Since these config options are considered more advanced, it's unlikely that many operators actually use them and instead rely on the default settings.


src/herder/HerderImpl.cpp
```diff
@@ -1934,9 +1950,69 @@ HerderImpl::restoreUpgrades()
...
+        LedgerTxn ltx(mApp.getLedgerTxnRoot(),
+                      /* shouldUpdateLastModified */ true,
+                      TransactionMode::READ_ONLY_WITHOUT_SQL_TXN);
+        auto conf = mApp.getLedgerManager().getSorobanNetworkConfig(ltx);
+
+        if (conf.txMaxSizeBytes() > mMaxTxSize)
+        {
+            diff = conf.txMaxSizeBytes() - mMaxTxSize;
+        }
+        mMaxTxSize =
```
mbsdf, line 1968: do we need to update FlowControl.cpp L388

src/herder/HerderImpl.cpp
```diff
@@ -1934,9 +1950,69 @@ HerderImpl::restoreUpgrades()
...
+#ifdef ENABLE_NEXT_PROTOCOL_VERSION_UNSAFE_FOR_PRODUCTION
+    {
+        LedgerTxn ltx(mApp.getLedgerTxnRoot(),
+                      /* shouldUpdateLastModified */ true,
+                      TransactionMode::READ_ONLY_WITHOUT_SQL_TXN);
+        auto conf = mApp.getLedgerManager().getSorobanNetworkConfig(ltx);
+        mMaxTxSize = std::max<uint64_t>(sz, conf.txMaxSizeBytes());
+    }
+#else
+    mMaxTxSize = sz;
```
mbsdf, line 1993: is this by default MAX_CLASSIC_TX_SIZE_BYTES = 100 * 1024

src/herder/HerderImpl.h
```diff
@@ -127,6 +134,13 @@ class HerderImpl : public Herder
 
     void processSCPQueue();
 
+    uint64_t getMaxClassicTxSize() override;
+    uint64_t
+    getMaxTxSize() const override
+    {
+        return mMaxTxSize;
```
mbsdf, line 141: ok

src/herder/HerderImpl.cpp
```diff
@@ -1934,9 +1950,69 @@ HerderImpl::restoreUpgrades()
...
+    // Core will calculate default values automatically
+    bool calculateDefaults = cfg.PEER_FLOOD_READING_CAPACITY_BYTES == 0 &&
+                             cfg.FLOW_CONTROL_SEND_MORE_BATCH_SIZE_BYTES == 0;
+
+    if (!calculateDefaults &&
+        !(cfg.PEER_FLOOD_READING_CAPACITY_BYTES -
+              cfg.FLOW_CONTROL_SEND_MORE_BATCH_SIZE_BYTES >=
+          mMaxTxSize))
+    {
+        std::string msg = fmt::format(
```
mbsdf, line 2006: just to make sure; if we increase tx size in ledger, is it possible that it can create a config error (condition above) when core is restarted?

src/main/Config.cpp
```diff
@@ -246,8 +247,10 @@ Config::Config() : NODE_SEED(SecretKey::random())
     PEER_FLOOD_READING_CAPACITY = 200;
     FLOW_CONTROL_SEND_MORE_BATCH_SIZE = 40;
 
-    PEER_FLOOD_READING_CAPACITY_BYTES = 300000;
-    FLOW_CONTROL_SEND_MORE_BATCH_SIZE_BYTES = 100000;
+    // If set to 0, calculate automatically (this will be done after application
+    // startup as we need to load soroban configs)
+    PEER_FLOOD_READING_CAPACITY_BYTES = 0;
```
mbsdf, line 252: regarding comment in HerderImp.cpp L2000 condition;  only if calculating defaults. LGTM

src/main/Config.cpp
```diff
@@ -1514,50 +1517,6 @@ Config::processConfig(std::shared_ptr<cpptoml::table> t)
             LIMIT_TX_QUEUE_SOURCE_ACCOUNT = true;
         }
 #endif
-        // PEER_FLOOD_READING_CAPACITY_BYTES (C): This is the initial credit
```
mbsdf, line 1517: Can we update these comments to reflect new changes if any? I think these comments were useful.

src/main/Config.cpp
```diff
@@ -1514,50 +1517,6 @@ Config::processConfig(std::shared_ptr<cpptoml::table> t)
...
-
-        // Receiver processes received bytes and once
-        // FLOW_CONTROL_SEND_MORE_BATCH_SIZE_BYTES (A) or more is processed, an
-        // acknowledgement is sent, which replenishes the sender's credit
-        // Sender (C-M1+A) <-- A bytes ---------- Receiver
-        //             \--- (C-M1+A)-M2 --->/
-
-        // Note:  M1, M2... are message sizes such that M <=
-        // MAX_CLASSIC_TX_SIZE_BYTES
-        if (!(PEER_FLOOD_READING_CAPACITY_BYTES -
```
mbsdf, line 1550: this condition does not apply anymore, correct?

src/main/Config.cpp
mbsdf: nvm

mbsdf (commented):
couple of comments and questions

src/main/Config.cpp
marta-lokhova: It still applies, but this check got moved to `HerderImpl::start` because we now need to load Soroban config to verify this condition (and we can't load the config until we start the application and initialize the database)

src/herder/HerderImpl.cpp
marta-lokhova: If operator relies on default (automatically calculated) values, no action is needed on restart. If the operator has custom values configured, at upgrade time, they'll be automatically adjusted to keep core running. However, on restart core will crash due to invalid parameters, so the operator will need to manually update them.

src/herder/HerderImpl.cpp
```diff
@@ -1934,9 +1950,69 @@ HerderImpl::restoreUpgrades()
...
+            diff = conf.txMaxSizeBytes() - mMaxTxSize;
+        }
+        mMaxTxSize =
+            std::max<uint64_t>(getMaxClassicTxSize(), conf.txMaxSizeBytes());
+    }
+
+    // Maybe update capacity to reflect the upgrade
+    for (auto& peer : mApp.getOverlayManager().getAuthenticatedPeers())
+    {
+        peer.second->handleMaxTxSizeIncrease(diff);
```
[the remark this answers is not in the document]
marta-lokhova: yes, the node sends an additional SEND_MORE message to indicate an increase in capacity.

src/herder/HerderImpl.cpp
marta-lokhova: Are you referring to this? https://github.com/stellar/stellar-core/pull/3806/files#diff-d37d5bbb43097772a9cbca1b1215a7599e1fdd7be78ffb575fdc309dc9cf93d6R420 This PR already updates all places where getMaxTxSize should be used.

src/herder/HerderImpl.h
marta-lokhova: is there an action item?

src/overlay/FlowControlCapacity.cpp
```diff
@@ -97,6 +100,15 @@ FlowControlByteCapacity::canRead() const
     return true;
 }
 
+#ifdef ENABLE_NEXT_PROTOCOL_VERSION_UNSAFE_FOR_PRODUCTION
+void
+FlowControlByteCapacity::handleTxSizeIncrease(uint64_t increase)
+{
+    mCapacity.mFloodCapacity += increase;
```
SirTyson, line 107: Do we need to check for overflow here since this increase value is coming from another node and could potentially be malicious?

src/overlay/OverlayManagerImpl.h
```diff
@@ -227,5 +228,6 @@ class OverlayManagerImpl : public OverlayManager
     void extractPeersFromMap(std::map<NodeID, Peer::pointer> const& peerMap,
                              std::vector<Peer::pointer>& result);
     void shufflePeerList(std::vector<Peer::pointer>& peerList);
+    std::pair<uint32_t, uint32_t> getFlowControlBytesConfig() const override;
```
SirTyson, line 231: Nit: This std::pair interface is a bit hard to follow. Maybe break this into a struct or typedef? Or at least add a comment on the return type.

src/herder/HerderImpl.cpp
```diff
@@ -1934,9 +1950,69 @@ HerderImpl::restoreUpgrades()
...
+    auto sz = mApp.getHerder().getMaxClassicTxSize();
+#ifdef ENABLE_NEXT_PROTOCOL_VERSION_UNSAFE_FOR_PRODUCTION
+    {
+        LedgerTxn ltx(mApp.getLedgerTxnRoot(),
+                      /* shouldUpdateLastModified */ true,
+                      TransactionMode::READ_ONLY_WITHOUT_SQL_TXN);
+        auto conf = mApp.getLedgerManager().getSorobanNetworkConfig(ltx);
+        mMaxTxSize = std::max<uint64_t>(sz, conf.txMaxSizeBytes());
+    }
+#else
```
SirTyson, line 1992: Nit: IDK if this is valid or not but I'm not a huge fan of the `#ifdef #else` value assignment. If you change 1983 to `mMaxTxSize = mApp.getHerder().getMaxClassicTxSize();` you can get rid of the `#else`.

SirTyson (commented):
Overall looks good, just a few minor cleanups and one potential security issue.

src/overlay/FlowControlCapacity.cpp
marta-lokhova: so here the increase is coming from the local node's upgrade logic, which I assume won't overflow (if the network upgrades to those types of values, then probably something is really wrong anyway). We do send another SEND_MORE message with `increase` in it, but we already handle potential overflow here (the mitigation is that the problematic peer is dropped) https://github.com/stellar/stellar-core/blob/0eab126b4343cf7d93af2cc9ca97c5bf7728ee4a/src/overlay/FlowControl.cpp#L340

src/herder/HerderImpl.cpp
```diff
@@ -1934,9 +1950,69 @@ HerderImpl::restoreUpgrades()
     }
 }
 
+#ifdef ENABLE_NEXT_PROTOCOL_VERSION_UNSAFE_FOR_PRODUCTION
+void
+HerderImpl::maybeHandleUpgrade()
+{
```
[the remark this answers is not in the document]
marta-lokhova: mmm, `getSorobanNetworkConfig` doesn't fail if protocolVersion < 20, but instead creates an empty config, right? 
https://github.com/stellar/stellar-core/blob/0eab126b4343cf7d93af2cc9ca97c5bf7728ee4a/src/ledger/LedgerManagerImpl.cpp#L1148

So I believe this function should be a no-op if we're running protocol 19 (as confirmed by "upgrade to generalized tx set in network" test which runs the network at proto 19 for some time, then upgrades to 20). That being said, it does seem error-prone (e.g. what if this logic handles default values incorrectly), so I agree we can add a protocol check just in case.

src/overlay/FlowControlCapacity.cpp
```diff
@@ -97,6 +100,15 @@ FlowControlByteCapacity::canRead() const
...
+    mCapacityLimits.mFloodCapacity += increase;
```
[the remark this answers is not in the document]
marta-lokhova: discussed offline, but re-iterating here for posterity: one is node's current capacity, and the other is node's capacity limit.

src/overlay/FlowControl.cpp
```diff
@@ -94,7 +94,7 @@ FlowControl::start(std::weak_ptr<Peer> peer,
         mFlowControlBytesCapacity =
             std::make_shared<FlowControlByteCapacity>(mApp, mNodeID);
         sendSendMore(mApp.getConfig().PEER_FLOOD_READING_CAPACITY,
-                     mApp.getConfig().PEER_FLOOD_READING_CAPACITY_BYTES,
+                     mApp.getOverlayManager().getFlowControlBytesConfig().first,
```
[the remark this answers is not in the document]
marta-lokhova: yeah, exactly

src/overlay/OverlayManagerImpl.h
marta-lokhova: good idea, using pair of the same type is also kind of error-prone. switched to a struct

src/overlay/FlowControlCapacity.cpp
SirTyson: Looks good. Sorry, didn't realize this overflow check was already there!

src/herder/HerderImpl.cpp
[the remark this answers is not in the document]
SirTyson: Yeah it's probably fine as is, I just get a little spooked calling anything with Soroban in the name pre protocol 20.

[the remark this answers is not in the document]
marta-lokhova: yeah, I also opened https://github.com/stellar/stellar-core/issues/3828 to prohibit retrieving soroban config prior to protocol 20

marta-lokhova:
@SirTyson @mbsdf I believe the comments should be addressed now. Let me know if the change looks good.

mbsdf (approved):
LGTM

src/herder/HerderImpl.cpp
```diff
@@ -1959,12 +1959,20 @@ HerderImpl::maybeHandleUpgrade()
         LedgerTxn ltx(mApp.getLedgerTxnRoot(),
                       /* shouldUpdateLastModified */ true,
                       TransactionMode::READ_ONLY_WITHOUT_SQL_TXN);
-        auto conf = mApp.getLedgerManager().getSorobanNetworkConfig(ltx);
+        if (ltx.loadHeader().current().ledgerVersion <
```
SirTyson, line 1962: Nit: Why not `protocolVersionIsBefore`?

SirTyson (approved):
LGTM

MonsieurNicolas:
r+ 09c4774016a81a79b6d67022c7ae01709d21d29d

PR #3807, opened 2023-07-05
Bucketlist size meta

### LINKED ISSUE #3783: Emit totalByteSizeOfBucketList size in Meta
We need to calculate the 30 day average of the BucketList size and emit it in meta so that downstream systems can correctly calculate fees. We should store 30 days worth of daily snapshots in the BucketList `CONFIG_ENTRY` and emit meta via `LedgerCloseMetaV2`.

SirTyson:
# Description

Resolves #3783

This PR emits `totalByteSizeOfBucketList` in ledger close meta so that downstream systems can properly calculate Soroban storage related fees. This value is the 30 day average of the size of the BucketList calculated via daily snapshots. These snapshots are stored on disk in a sliding window via `CONFIG_ENTRY`.

# Checklist
- [x] Compiles
- [x] Ran all tests

### DIFF
--- a/src/herder/test/UpgradesTests.cpp
+++ b/src/herder/test/UpgradesTests.cpp
@@ -3,6 +3,9 @@
 // of this distribution or at http://www.apache.org/licenses/LICENSE-2.0
 
 #include "bucket/BucketInputIterator.h"
+#include "bucket/BucketList.h"
+#include "bucket/BucketManager.h"
+#include "bucket/BucketManagerImpl.h"
 #include "bucket/test/BucketTestUtils.h"
 #include "herder/Herder.h"
 #include "herder/HerderImpl.h"
@@ -281,6 +284,15 @@ getMaxContractSizeKey()
     maxContractSizeKey.configSetting().configSettingID =
         CONFIG_SETTING_CONTRACT_MAX_SIZE_BYTES;
     return maxContractSizeKey;
+}
+
+LedgerKey
+getBucketListSizeWindowKey()
+{
+    LedgerKey windowKey(CONFIG_SETTING);
+    windowKey.configSetting().configSettingID =
+        CONFIG_SETTING_BUCKETLIST_SIZE_WINDOW;
+    return windowKey;
 }
 
 #endif
@@ -2015,6 +2027,7 @@ TEST_CASE("configuration initialized in version upgrade", "[upgrades]")
         REQUIRE(!ltx.load(getMaxContractSizeKey()));
     }
 
+    auto blSize = app->getBucketManager().getBucketList().getSize();
     executeUpgrade(*app, makeProtocolVersionUpgrade(
                              static_cast<uint32_t>(SOROBAN_PROTOCOL_VERSION)));
 
@@ -2025,6 +2038,32 @@ TEST_CASE("configuration initialized in version upgrade", "[upgrades]")
             CONFIG_SETTING_CONTRACT_MAX_SIZE_BYTES);
     REQUIRE(maxContractSizeEntry.contractMaxSizeBytes() ==
             InitialSorobanNetworkConfig::MAX_CONTRACT_SIZE);
+
+    // Check that BucketList size window initialized with current BL size
+    auto& networkConfig = app->getLedgerManager().getSorobanNetworkConfig(ltx);
+    REQUIRE(networkConfig.getAverageBucketListSize() == blSize);
+
+    // Check in memory window
+    auto const& inMemoryWindow =
+        networkConfig.getBucketListSizeWindowForTesting();
+    REQUIRE(inMemoryWindow.size() ==
+            InitialSorobanNetworkConfig::BUCKET_LIST_SIZE_WINDOW_SAMPLE_SIZE);
+    for (auto const& e : inMemoryWindow)
+    {
+        REQUIRE(e == blSize);
+    }
+
+    // Check LedgerEntry with window
+    auto onDiskWindow = ltx.load(getBucketListSizeWindowKey())
+                            .current()
+                            .data.configSetting()
+                            .bucketListSizeWindow();
+    REQUIRE(onDiskWindow.size() ==
+            InitialSorobanNetworkConfig::BUCKET_LIST_SIZE_WINDOW_SAMPLE_SIZE);
+    for (auto const& e : onDiskWindow)
+    {
+        REQUIRE(e == blSize);
+    }
 }
 #endif
 
--- a/src/ledger/NetworkConfig.cpp
+++ b/src/ledger/NetworkConfig.cpp
@@ -3,6 +3,9 @@
 // of this distribution or at http://www.apache.org/licenses/LICENSE-2.0
 
 #include "ledger/NetworkConfig.h"
+#include "bucket/BucketList.h"
+#include "bucket/BucketManager.h"
+#include "main/Application.h"
 #include "util/ProtocolVersion.h"
 
 namespace stellar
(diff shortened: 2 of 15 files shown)


src/bucket/BucketManagerImpl.cpp
```diff
@@ -960,6 +1136,26 @@ BucketManagerImpl::getBloomLookupMeter() const
     return mBucketListDBBloomLookups;
 }
 
+uint64_t
+BucketManagerImpl::getAverageBucketListSize(AbstractLedgerTxn& ltx) const
```
dmkozh, line 1140: This probably should belong to the `NetworkConfig` in order to have all the lazy loading logic in one spot (and then we could hopefully get rid of lazy loading altogether). 
 think most of the code here should actually be there, besides the bits that actually compute the bucket list size. Is there any particular reason for having this in the bucket list manager?

src/bucket/test/BucketManagerTests.cpp
```diff
@@ -327,6 +327,99 @@ TEST_CASE("bucketmanager missing buckets fail", "[bucket][bucketmanager]")
...
+        // Check on-disk sliding window
+        LedgerKey key(CONFIG_SETTING);
+        key.configSetting().configSettingID =
+            ConfigSettingID::CONFIG_SETTING_BUCKETLIST_SIZE_WINDOW;
+        auto txle = ltx.loadWithoutRecord(key, /*loadExpiredEntry=*/false);
+        releaseAssert(txle);
+        auto const& leVector =
+            txle.current().data.configSetting().bucketListSizeWindow();
+
+        REQUIRE(correctWindow.size() == leVector.size());
```
dmkozh, line 382: Can we just use `REQUIRE(correctWindow == leVector)` ? (you may convert deque to a vec for this to work)

src/herder/test/UpgradesTests.cpp
```diff
@@ -2025,6 +2040,30 @@ TEST_CASE("configuration initialized in version upgrade", "[upgrades]")
             CONFIG_SETTING_CONTRACT_MAX_SIZE_BYTES);
     REQUIRE(maxContractSizeEntry.contractMaxSizeBytes() ==
             InitialSorobanNetworkConfig::MAX_CONTRACT_SIZE);
+
+    // Check that BucketList size window initialized with current BL size
+    REQUIRE(bm.getAverageBucketListSize(ltx) == blSize);
```
dmkozh, line 2045: We'll also need some tests to cover upgrades involving increasing/decreasing the window size. It's ok to do that as a followup though.

src/bucket/BucketManagerImpl.cpp
```diff
@@ -825,7 +825,183 @@ BucketManagerImpl::addBatch(Application& app, uint32_t currLedger,
...
+    auto txle = ltx.load(key);
+    releaseAssert(txle);
+    auto& leVector = txle.current().data.configSetting().bucketListSizeWindow();
+    leVector.clear();
+    for (auto e : mBucketListSizeSnapshots)
+    {
+        leVector.push_back(e);
+    }
+
+    ltx.commit();
```
dmkozh, line 879: We don't emit any meta here, do we? I think that's an issue.

SirTyson: I don't think so? This is writing the sliding window to disk, which is a `CONFIG_ENTRY`. This is a Validator, BucketList specific entry that's only used to maintain the BucketList size average. The only BucketList size related meta we need to emit is a single value for the average. I think it's fine we don't emit anything hear because even though the sliding window is technically a `LedgerEntry`, it shouldn't be useful for any downstream systems and isn't a result of a transaction, so I don't think there would be a good place to put it in the meta and I don't think it would be useful.

src/bucket/BucketManagerImpl.cpp
```diff
@@ -825,7 +825,183 @@ BucketManagerImpl::addBatch(Application& app, uint32_t currLedger,
...
+            mBucketListSizeSnapshots.pop_front();
+        }
+    }
+    // If newSize > currSize, initialize new slots with current BucketList size
+    else
+    {
+        auto blSize = mBucketList->getSize();
+        while (mBucketListSizeSnapshots.size() != newSize)
+        {
+            mBucketListSizeSnapshots.push_back(blSize);
```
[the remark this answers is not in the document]
SirTyson: I'm using a `std::deque` to implement an iterable FIFO queue. The iteration order doesn't matter for us, and using `push_front/pop_back` vs. `push_back/pop_front` is pretty much identical on `std::deque`. I picked `push_back/pop_front` because [stack overflow](https://stackoverflow.com/questions/20219736/which-is-better-with-a-dequeue-to-use-push-back-pop-front-or-push-front-p) says it's used in the STL, but I can change it, it's pretty arbitrary.

src/bucket/BucketManagerImpl.cpp
dmkozh: Sounds good, I somehow haven't noticed that we emit the average (and not the current) size.

src/bucket/BucketManagerImpl.cpp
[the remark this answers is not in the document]
dmkozh: It's not arbitrary, is it? We have older sizes at front and newer sizes at back. Here you're adding the new size to the back of the queue, which effectively shifts the history back and breaks it. What I'm suggesting is backfilling the old slots with the old size (so that they will be popped as soon as possible).

src/ledger/NetworkConfig.cpp
```diff
@@ -868,6 +957,92 @@ SorobanNetworkConfig::ledgerMaxTxCount() const
...
+    }
+
+    if (newSize < currSize)
+    {
+        while (mBucketListSizeSnapshots.size() != newSize)
+        {
+            mBucketListSizeSnapshots.pop_front();
+        }
+    }
+    // If newSize > currSize, initialize new slots with current BucketList size
```
dmkozh, line 999: Please update the comment.

src/ledger/NetworkConfig.cpp
```diff
@@ -868,6 +957,92 @@ SorobanNetworkConfig::ledgerMaxTxCount() const
...
+    else
+    {
+        auto oldestSize = mBucketListSizeSnapshots.front();
+        while (mBucketListSizeSnapshots.size() != newSize)
+        {
+            mBucketListSizeSnapshots.push_back(oldestSize);
```
dmkozh, line 1005: Shouldn't this be `push_front`? (again, these should be the 'old' slots and you'd want to pop them ASAP, right?)

src/ledger/NetworkConfig.h
```diff
@@ -175,9 +179,32 @@ class SorobanNetworkConfig
...
     uint32_t ledgerMaxTxCount() const;
 
+    // Number of samples in slidign window
+    uint32_t getBucketListSizeSnapshotPeriod() const;
+
+    // If newSize is different than the current BucketList size sliding window,
+    // update the window. If newSize < currSize, pop entries off window. If
+    // newSize > currSize, add as many copies of the current BucketList size to
+    // window until it has newSize entries.
+    void maybeUpdateBucketListWindowSize(AbstractLedgerTxn& ltx,
```
dmkozh, line 189: I think this should be a static/free function and not fields should be mutable. Instead, this would just update the ledger entry and then we will automatically update the fields as a part of `SorobanNetworkConfig::loadFromLedger`  (it happens after any update).

I think it's not necessary to do the change in this PR though - probably just open an issue to do cleanup?

dmkozh (commented):
Looks good overall, just a couple minor things and maybe a followup

src/bucket/BucketManagerImpl.cpp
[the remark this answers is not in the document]
SirTyson: Sorry, I misread your comment initially, you're right

dmkozh:
r+ 9219a44

Issue #3769, opened 2023-06-09
Make it easier to capture SCP messages


MonsieurNicolas:
Right now SCP messages go to 2 places:
* meta, as soon as we close a ledger
* history archives (via SCPHistory, as soon as we close a ledger)

This issue tracks improving the quality of the SCP messages uploaded to archives as it's the easiest of the two (we could emit SCP updates from older ledgers in the meta, but it seems a bit more involved).

The issue can be summarized as:
SCP messages that are emitted are the ones used by that node to externalize.
To externalize, the node needs to receive either `CONFIRM` or `EXTERNALIZE` messages from nodes in its quorum set.

Some systems (like bridges) rely on SCP messages, so the more `EXTERNALIZE` messages we can gather, the better it is (they're smaller and easier to process from contracts).

The proposal tracked here is to add a mechanism to core to improve the chances that we have `EXTERNALIZE` messages from as many validators in the transitive quorum as possible published to archive.

Given how the system works right now the following may work:
- [x] audit herder/SCP filtering logic to ensure that we continue to store `EXTERNALIZE` messages in memory after a ledger is closed. I think this is already the case (right now we accumulate those so that we can serve them to other peers).
- [x] add a new setting that controls a delay to publish to archive.
   * the delay could be expressed in terms of number of ledgers (anything less than `MAX_SLOTS_TO_REMEMBER`) or in time (maybe harder to configure/implement?) - in practice we probably need only a few more seconds to receive all messages (validators tend to track the same ledger)
- [ ] continue to save SCP messages in the database for older slots. Right now this only occurs [right when we externalize](https://github.com/stellar/stellar-core/blob/287fe4c5d7f7ebce1a327cd26f25dd5858cd37a0/src/herder/HerderImpl.cpp#L280)
       * after the delay, update the `scphistory` table to reflect the internal SCP state. we could compute the differences in terms of SCP messages (or just rewrite the whole thing? -- I don't know how much data we're really talking about here) for the effected slot
           *  note that we must write both when we externalize and later as we want to have in the database at least the messages that were used when externalizing in case of a crash.
       * the delay is also used to delay the actual snapshot (that produces files from database state) & publish step
   * this setting should probably not be set on validators, and only for watchers (we may fail to start on validators) as the delay can potentially put the network at risk of an outage

marta-lokhova:
The delay implementation (second bullet) was introduced in https://github.com/stellar/stellar-core/pull/3832

PR #3832, opened 2023-07-17
Capture scp messages

JuI3s:
# Description

Draft working on #3769 

# Checklist
- [ ] Compiles
- [ ] Ran all tests


mbsdf:
```
local:2/4/100%/51.0s parallel: This job failed:
runpart 1 'application setup' 'XDRSHA256 is identical to byte SHA256' 'bucketmanager do not leak empty-merge futures' 'History publish to multiple archives' 'parse peer record'
FAIL: test/selftest-pg
```

src/herder/HerderImpl.cpp
```diff
@@ -314,12 +318,41 @@ HerderImpl::processExternalized(uint64 slotIndex, StellarValue const& value)
         }
     }
 
-    // tell the LedgerManager that this value got externalized
-    // LedgerManager will perform the proper action based on its internal
-    // state: apply, trigger catchup, etc
-    LedgerCloseData ledgerData(static_cast<uint32_t>(slotIndex),
-                               externalizedSet, value);
-    mLedgerManager.valueExternalized(ledgerData);
+    auto publishToArchive = [this, value, slotIndex]() {
```
MonsieurNicolas, line 321: This change is too broad and probably not doing what you want:
* this delays externalizing ledgers in addition to delaying publish
* by not applying ledgers synchronously with `processExternalized`, I *think* you're potentially introducing new classes of race conditions where the node will potentially externalize the next slot before having applied the previous one. This in turn will cause all sorts of problems (like skipping applying ledgers as you only have one timer to apply ledgers)

src/herder/HerderImpl.cpp
```diff
@@ -330,6 +363,7 @@ void
 HerderImpl::valueExternalized(uint64 slotIndex, StellarValue const& value,
                               bool isLatestSlot)
 {
+
```
MonsieurNicolas, line 366: white space?

MonsieurNicolas (changes requested):
general approach probably has too many implications (which would require a lot of new tests to be written).
Also: rebase branches instead of merging as this creates hard-to-follow git history

src/herder/HerderImpl.cpp
JuI3s: I see. My understanding now is that we should only delay publishing not externalising ledgers?

For that purpose, does it suffice to delay executing [these steps](https://github.com/stellar/stellar-core/blob/ee13db4b76b8d68cf534b4c4afb74a100ac9c25a/src/ledger/LedgerManagerImpl.cpp#L923C1-L929C1)?

src/ledger/LedgerManagerImpl.cpp
```diff
@@ -920,12 +926,30 @@ LedgerManagerImpl::closeLedger(LedgerCloseData const& ledgerData)
...
+        hm.logAndUpdatePublishStatus();
 
-    // step 4
-    mApp.getBucketManager().forgetUnreferencedBuckets();
+        // step 4
+        mApp.getBucketManager().forgetUnreferencedBuckets();
+    }
+    else
+    {
+        mDelayPublishToArchiveTimer.expires_from_now(
```
marta-lokhova, line 940: I would suggest moving the delay logic to the actual publishing task. This codepath is quite sensitive (the node is atomically updating state of the ledger), so I recommend keeping the sequence of events as-is.

 Here we're snapshotting the current state of the ledger, then scheduling the publish work asynchronously https://github.com/stellar/stellar-core/blob/c2599d62274931c11ea7f36c5ee9255a202f5739/src/history/HistoryManagerImpl.cpp#L287. So I would suggest tweaking PublishWork to begin "work" with a small delay. Note that we're using a sequence of "works", so what you could do is add a WorkWithCallback at the front of the sequence, which succeeds if enough time has passed. Let me know if you have any questions about the Work subsystem, I think using it is the cleanest approach.

src/ledger/LedgerManagerImpl.h
```diff
@@ -155,7 +160,8 @@ class LedgerManagerImpl : public LedgerManager
                  std::shared_ptr<HistoryArchive> archive,
                  std::set<std::shared_ptr<Bucket>> bucketsToRetain) override;
 
-    void closeLedger(LedgerCloseData const& ledgerData) override;
```
marta-lokhova, line 158: looks like protocol-next/xdr changes ended up in this PR by accident :)

src/herder/Herder.h
```diff
@@ -50,6 +50,9 @@ class Herder
     // state
     static std::chrono::seconds const SEND_LATEST_CHECKPOINT_DELAY;
 
+    // Timeout before publishing externalized values to archive
```
marta-lokhova, line 53: I think it would be useful for nodes to opt out of this, so you could add a new config to the Config class instead of the const.

marta-lokhova (changes requested):

src/main/Config.cpp
```diff
@@ -142,6 +142,7 @@ Config::Config() : NODE_SEED(SecretKey::random())
     EXPERIMENTAL_BUCKETLIST_DB_INDEX_PAGE_SIZE_EXPONENT = 14; // 2^14 == 16 kb
     EXPERIMENTAL_BUCKETLIST_DB_INDEX_CUTOFF = 20;             // 20 mb
     EXPERIMENTAL_BUCKETLIST_DB_PERSIST_INDEX = true;
+    PUBLISH_TO_ARCHIVE_DELAY = std::chrono::seconds{2};
```
marta-lokhova, line 145: actually, should we set this to 0 by default, and operators that need this functionality can adjust the config manually?

src/history/HistoryManagerImpl.cpp
```diff
@@ -278,14 +279,32 @@ HistoryManagerImpl::takeSnapshotAndPublish(HistoryArchiveState const& has)
...
+    ConditionFn delayTimeout = [start](Application& app) {
+        auto delayDuration = app.getConfig().PUBLISH_TO_ARCHIVE_DELAY;
+        if (delayDuration == std::chrono::seconds::zero())
+        {
+            return true;
+        }
+
+        auto time = app.getClock().now();
+        auto dur = time - start;
+        return std::chrono::duration_cast<std::chrono::seconds>(dur) >
```
marta-lokhova, line 293: `>=`? This way you don't need to special case `std::chrono::seconds::zero`

marta-lokhova (approved):
Couple of minor suggestions, otherwise LGTM

marta-lokhova (approved):

marta-lokhova:
r+ 419329cf782a5e5c3d6f967f61ef7165dfa3a323

PR #3833, opened 2023-07-18
Bump spdlog and fmt library to latest

MonsieurNicolas:
This fixes "internal compiler" issues on current version of Visual C++.

This also fixes a couple places where we were logging the wrong thing (binary blobs instead of hashes).

I am not sure why customer formatters are not always used (in some cases I had to call them explicitely) -- maybe somebody will figure this out.


MonsieurNicolas:
I also bumped asio that has a fix for spurious connection failures on Linux (not handling `EINTR` properly).

dmkozh:
r+ b44ded8

PR #3852, opened 2023-07-22
Add support for building rust code with sanitizers

graydon:
This attempts to plumb the sanitizer flags through to cargo and rustc. I'm not 100% sure it's working right yet (it does not catch the error I was hoping to see it catch, but maybe tests don't exercise that).


graydon:
This is obsolete -- support for this landed back in e8bd4853e4c75c80c93335f36e0c3b71c16f596c

PR #3856, opened 2023-07-26
Nomination prototype

JuI3s:
Draft working off Hidenori's branch on the nomination protocol prototype. Replace transaction sets with their hashes.

# Description

Resolves #X

# Checklist
- [ ] Compiles
- [ ] Ran all tests


src/overlay/Peer.cpp
```diff
@@ -1141,6 +1175,21 @@ Peer::recvTxSet(StellarMessage const& msg)
     ZoneScoped;
     auto frame = TxSetFrame::makeFromWire(mApp, msg.txSet());
     mApp.getHerder().recvTxSet(frame->getContentsHash(), frame);
+    CLOG_INFO(Overlay,
+              "Peer::recvTxSet received {}, trying to fulfill pending requests",
+              hexAbbrev(frame->getContentsHash()));
+    auto pendingTxSetRequests =
+        mApp.getOverlayManager().getPendingTxSetRequests();
+
+    for (auto& weakPeer : pendingTxSetRequests[frame->getContentsHash()])
```
SirTyson, line 1184: I'm not 100% percent sure, but I think this unchecked map access may open up a DOS angle. Probably want to check with @marta-lokhova, but it looks like this assumes the incoming message asks for a pending TxSet hash that we currently have in `pendingTxSetRequests`. I don't think this assumption is valid given byzantine nodes or dropped messages from a well intentioned node. You should check that the hash exists in the map and not do an unchecked access.

SirTyson (changes requested):
I think you're off to a good start! I'm not super knowledgeable on specific SCP issues so this is more of a high-level overview for now, but I'll come back through and validate the actual logic a little more once the unit tests are passing. It might seem like a bunch of comments, but most of the stuff is more code style and git related issues which is always a bit tricky to get right the first couple times :).

src/overlay/test/LoopbackPeer.h
```diff
@@ -110,6 +119,12 @@ class LoopbackPeer : public Peer
     double getReorderProbability() const;
     void setReorderProbability(double d);
 
+    void
```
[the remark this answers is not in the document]
JuI3s: I'm not sure if this should be be only used for testing purposes (Hide made this change)? Is there any benefit in the real environment we might want to suspend flooding?

src/overlay/test/LoopbackPeer.h
```diff
@@ -35,6 +35,15 @@ class LoopbackPeer : public Peer
...
     bool mDamageAuth{false};
+
+    // Test tx set flooding by suspending tx flooding.
+    // If we allow tx flooding, it is possible that
+    // nodes don't need to flood tx sets because
+    // nodes may have the exact same txns
+    // in their TransactionQueue.
+    // This is more likely to be true
+    // esp in test environments with not many real-world obstacles.
+    bool mSuspendTxFlooding{false};
```
marta-lokhova, line 46: that's an interesting idea! I agree tx flooding can prevent us from flooding what we want.

src/overlay/test/OverlayTests.cpp
```diff
@@ -2956,4 +2956,68 @@ TEST_CASE("overlay pull mode with many peers",
...
+    simulation->startAllNodes();
+
+    REQUIRE(node1->getLedgerManager().isSynced());
+    REQUIRE(node2->getLedgerManager().isSynced());
+    REQUIRE(node3->getLedgerManager().isSynced());
+
+    auto& loadGen = node1->getLoadGenerator();
+    auto const numAccounts = 1;
+    loadGen.generateLoad(
+        GeneratedLoadConfig::createAccountsLoad(/* nAccounts */ numAccounts,
```
marta-lokhova, line 3001: I think it would be nice to strengthen this test by having all nodes generate load: right now, if node2 or node3 are selected as nomination leaders that introduce a tx set, they'll just introduce an empty tx sets since tx flooding is disabled. You can use loadgen's offset parameter to make sure transactions generated by multiple nodes don't clash, see `"soroban txs accepted by the network"` test for an example.

src/herder/test/HerderTests.cpp
```diff
@@ -5339,3 +5340,199 @@ TEST_CASE("exclude transactions by operation type", "[herder]")
...
+        StellarMessage getTxSet;
+        getTxSet.type(GET_TX_SET);
+        getTxSet.txSetHash() = setID;
+
+        return std::make_shared<StellarMessage>(getTxSet);
+    };
+
+    auto createTxSetMessage = [](TxSetFrameConstPtr txSet) {
+        StellarMessage msg;
+        msg.type(TX_SET);
```
marta-lokhova, line 5395: Since we recently enabled a new protocol (v20) in master which this change was rebased on top of, this should be generalized tx set.

src/herder/TxSetFrame.h
```diff
@@ -47,14 +47,14 @@ class TxSetFrame : public NonMovableOrCopyable
...
-    makeFromTransactions(TxPhases const& txPhases, Application& app,
-                         uint64_t lowerBoundCloseTimeOffset,
-                         uint64_t upperBoundCloseTimeOffset);
-    static TxSetFrameConstPtr
     makeFromTransactions(TxPhases const& txPhases, Application& app,
                          uint64_t lowerBoundCloseTimeOffset,
                          uint64_t upperBoundCloseTimeOffset,
-                         TxPhases& invalidTxsPerPhase);
+                         bool forceIsNotGeneralized = false);
+    static TxSetFrameConstPtr makeFromTransactions(
```
marta-lokhova, line 54: I think the issue is in the test, which is creating tx sets rather than generalized tx sets, so feel free to revert this change.

src/overlay/Peer.cpp
```diff
@@ -1105,10 +1191,52 @@ Peer::recvGetTxSet(StellarMessage const& msg)
...
+        }
+
+        if (wait)
+        {
+            CLOG_INFO(Overlay,
+                      "Peer::recvGetTxSet {} did not find the tx set for {}, "
+                      "triggering mTxSetRequestTimer.",
+                      toString(), hexAbbrev(msg.txSetHash()));
+
+            mTxSetRequestTimer.expires_from_now(
```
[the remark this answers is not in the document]
marta-lokhova: Could we add some tests that verify the expected behavior?

marta-lokhova (changes requested):
Exciting first pass at this fairly complicated piece of code. Thanks for putting this together!

I left some questions around the general tx set fetching mechanism, as well as some suggestions around safety/code cleanup.

src/herder/TxSetFrame.h
JuI3s: Sorry I thought we're meant to create tx sets rather than generalized tx sets, as the node is meant to receive the tx set? My recollection is that generalized tx set basically just contains the hash?

src/scp/SCPDriver.h
```diff
@@ -177,6 +178,10 @@ class SCPDriver
     // `computeTimeout` computes a timeout given a round number
     // it should be sufficiently large such that nodes in a
     // quorum can exchange 4 messages
+
+    // TODO: does 'can exchange 4 messages' consider sending getTxSet messages?
```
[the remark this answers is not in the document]
JuI3s: Sorry I'll just delete it.. I was confused at some point what the 4 messages mean. I think it refers to the nomination and ballot messages?

src/overlay/Peer.cpp
```diff
@@ -1105,10 +1191,52 @@ Peer::recvGetTxSet(StellarMessage const& msg)
...
         self->sendMessage(newMsgPtr);
     }
     else
     {
+        auto slotIndex =
+            mApp.getHerder().getLastSeenSlotIndexForTxSet(msg.txSetHash());
+
+        if (slotIndex == 0)
+        {
+            CLOG_INFO(Overlay,
```
marta-lokhova, line 1207: I think we should just send DONT_HAVE immediately here, since we're not tracking this tx set.

JuI3s: I think the issue is what if malicious nodes send too many malformed requests? If we reply to every message we might flood the network?

src/overlay/Peer.cpp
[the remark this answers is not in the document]
JuI3s: I think if receive the tx set while we wait, the condition auto txSet = mApp.getHerder().getTxSet(msg.txSetHash() should evaluate to true, so the next time we call recvGetTxSet request, we should have the tx set to send back to the peer?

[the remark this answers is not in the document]
JuI3s: The "delay sending DONT_HAVE" test case has a section "tx set received before timeout." to make sure we're sending the tx set back to peer.

src/overlay/Peer.cpp
```diff
@@ -1078,9 +1080,87 @@ Peer::recvDontHave(StellarMessage const& msg)
...
+//      │    │   ┃                 ┃                   │
+//      │    │   ┃                 ┃                   │
+//      │    │   ┗━━━━━━━━━━━━━━━━━┛                  Y│
+//      │    │            │                            │
+//      Peer times    Peer recvs                       │
+//      │  out.       GetTxnSet                        │
+//      │    │         request.                        │
+//      │    │            │                            │
+//    Peer   │   Peer has Λ                   First    Λ
+//  receives │   txn set?▕ ▏───────────N──────time?──▶▕ ▏
```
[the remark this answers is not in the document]
JuI3s: 1. By the second tx set fetching attempt, I just meant that if we do not have the tx set, we try again after a delay. Yeah maybe the diagram is not being clear and I should fix this.
2. I'll think about it.. If we don't send tx set to peers we've previously sent DONT_HAVE for, is there any. chance that the network might get stuck?
4. Sorry that part of the diagram is stale - I'll update it. Currently it's cleared with the periodic garbage collection.

src/herder/TxSetFrame.h
marta-lokhova: generalized transaction sets are also transactions sets, but with a slightly different structure to accommodate protocol 20. See definition here: https://github.com/stellar/stellar-xdr/blob/curr/Stellar-ledger.x#L234.

src/overlay/Peer.cpp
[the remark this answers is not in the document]
marta-lokhova: >I think if receive the tx set while we wait, the condition auto txSet = mApp.getHerder().getTxSet(msg.txSetHash() should evaluate to true, so the next time we call recvGetTxSet request, we should have the tx set to send back to the peer?

While this is technically true, we are not guaranteed to receive another fetching request from the same peer. Even when we do, this is after we ask all other peers (so tens of seconds later potentially). The current mechanism will cause significant performance/network liveness issues. Therefore, we need to forward the tx set to the peer that asked for it as soon as we receive the tx set.

src/overlay/Peer.cpp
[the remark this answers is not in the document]
marta-lokhova: hmm, still not following why the mechanism differentiates between first time fetching and any subsequent attempt. The state transitions should be the same regardless: if we receive a request, place it into the "pending request" set, and start a timer for DONT_HAVE. If we receive the tx set, forward it to the peer that asked and cancel the DONT_HAVE timer.

src/overlay/Peer.cpp
```diff
@@ -1078,9 +1080,81 @@ Peer::recvDontHave(StellarMessage const& msg)
...
+//      │          ┃                 ┃    │     ┌────────────────┐
+//      │          ┗━━━━━━━━━━━━━━━━━┛    │     │Add remote peer │
+//      │                   │             │     │   to pending   │
+//      │                   │             │     │    getTxSet    │
+//      │                   │             │     │requests for tx │
+//      │                   │             │     └────────────────┘
+//      │          ┌────────┴───────┐     │              ▲
+// Local node      │Local node recvs│     │              │
+//  receives       │   GetTxnSet    │◀────┘              │
+//  new txn        │    request.    │                   Y│
```
marta-lokhova, line 1123: What does "Retry after timeout" mean here? It would be the remote node retrying, right?

src/overlay/test/LoopbackPeer.h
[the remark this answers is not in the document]
marta-lokhova: LoopbackPeer is only used for testing, so no need for BUILD_TESTS here.

marta-lokhova:
Closing this as stale, we'll probably open a new PR picking up this work

PR #3896, opened 2023-08-25
rust bridge

SirTyson:
This is a draft with changes to the rust bridge interface. We pass LedgerEntries via two `Vec<CxxBuf>`. For each entry in `ledger_entries`, there is a corresponding `ExpriationEntry` in `expiration_entries`. If a given LedgerEntry does not have an associated ExpirationEntry, the `CxxBuf` at the given index in `expiration_entries` will be empty.

### DIFF
--- a/src/rust/src/contract.rs
+++ b/src/rust/src/contract.rs
@@ -7,7 +7,7 @@ use crate::{
     rust_bridge::{
         CxxBuf, CxxFeeConfiguration, CxxLedgerEntryRentChange, CxxLedgerInfo,
         CxxRentFeeConfiguration, CxxTransactionResources, CxxWriteFeeConfiguration, FeePair,
-        InvokeHostFunctionOutput, ReadOnlyBump, RustBuf, XDRFileHash,
+        InvokeHostFunctionOutput, RustBuf, XDRFileHash,
     },
 };
 use log::debug;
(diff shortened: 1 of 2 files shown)


src/rust/src/contract.rs
```diff
@@ -209,30 +209,17 @@ fn encode_diagnostic_events(events: &Vec<DiagnosticEvent>) -> Vec<RustBuf> {
...
-                if expiration_change.new_expiration_ledger > expiration_change.old_expiration_ledger
-                {
-                    read_only_bumps.push(ReadOnlyBump {
-                        ledger_key: change.encoded_key.into(),
-                        min_expiration: expiration_change.new_expiration_ledger,
-                    });
-                }
-            }
-        } else {
+        assert!(!change.read_only);
```
dmkozh, line 216: This is not correct - there may be an expiration entry change for RO entry (there won't be a separate `LedgerEntryChange` for expiration entries).

src/rust/src/contract.rs
```diff
@@ -209,30 +209,17 @@ fn encode_diagnostic_events(events: &Vec<DiagnosticEvent>) -> Vec<RustBuf> {
...
             if let Some(encoded_new_value) = change.encoded_new_value {
                 modified_entries.push(encoded_new_value.into());
             }
-        }
     }
 
-    (modified_entries, read_only_bumps)
+    modified_entries
```
dmkozh, line 222: I think this function should be converting `LedgerEntryChange` to (maybe) a LE change and (maybe) an EE change. But that really depends on the host-side implementation at this point.

src/rust/src/contract.rs
SirTyson: I think I'm misunderstanding this function then. I thought the `ExpirationEntry` itself would have `LedgerEntryChanges` populated, meaning that every lifetime bump, even if the associated ContractCode/Data entry is in the RO set, would never be a read_only `LedgerEntryChange`?

dmkozh: No, I think we should keep expiration changes correlated with their 'main' entries. Otherwise it will be hard to e.g. compute the rent fees. Basically at the `LedgerEntryChange` level we don't really want to expose how we store the expiration changes.

PR #3904, opened 2023-08-31
Add medida metric for cpu_insn/time_nsecs ratio

jayz22:
# Description

Compute the cpu_insn/time_nsecs ratio and export that as a metric.

# Why
The cpu/time ratio is important for understanding the accuracy of our metering model.
Currently we compute the ratio during query time, which depends on a number of external factors like the scrape rate. And as a result we are only displaying the cumulative trend which is going upwards over time (this is expected because the metering model overestimates).
(https://grafana.stellar-ops.com/d/DwjDwtPVz/soroban?orgId=1&from=now-90d&to=now&forceLogin&viewPanel=20)
To get the more accurate and useful signals, we need the ratio computed and exported as its own metric.

### DIFF
--- a/src/transactions/InvokeHostFunctionOpFrame.cpp
+++ b/src/transactions/InvokeHostFunctionOpFrame.cpp
@@ -173,6 +173,7 @@ struct HostFunctionMetrics
     uint32 mCpuInsn{0};
     uint32 mMemByte{0};
     uint32 mInvokeTimeNsecs{0};
+    uint32 mInsnNsecRatio{0};
 
     // max single entity size metrics
     uint32 mMaxReadWriteKeyByte{0};
@@ -262,6 +263,8 @@ struct HostFunctionMetrics
         mMetrics
             .NewMeter({"soroban", "host-fn-op", "invoke-time-nsecs"}, "time")
             .Mark(mInvokeTimeNsecs);
+        mMetrics.NewMeter({"soroban", "host-fn-op", "insn-nsec-ratio"}, "ratio")
+            .Mark(mInsnNsecRatio);
 
         mMetrics.NewMeter({"soroban", "host-fn-op", "max-rw-key-byte"}, "byte")
             .Mark(mMaxReadWriteKeyByte);
@@ -491,6 +494,9 @@ InvokeHostFunctionOpFrame::doApply(Application& app, AbstractLedgerTxn& ltx,
     metrics.mCpuInsn = static_cast<uint32>(out.cpu_insns);
     metrics.mMemByte = static_cast<uint32>(out.mem_bytes);
     metrics.mInvokeTimeNsecs = static_cast<uint32>(out.time_nsecs);
+    metrics.mInsnNsecRatio =
+        metrics.mCpuInsn / std::max(metrics.mInvokeTimeNsecs, (uint32)1);
+
     if (!out.success)
     {
         if (resources.instructions < out.cpu_insns ||


MonsieurNicolas:
Can you explain which problem this is trying to solve? I do not understand the comment on having a dependency on scrape period (also note that if there are multiple transactions per ledger, you have a similar problem).

We have both `{"soroban", "host-fn-op", "cpu-insn"}` that tracks the cumulative cost and `{"soroban", "host-fn-op", "invoke-time-nsecs"}` that tracks the  cumulative time spent executing, so in grafana we should be able to plot the ratio of the deltas over arbitrary periods?

The problem with having the ratio emitted per transaction is that you can't really aggregate that metric. I guess this could be a metric used to find outliers (so only way to aggregate is to use min/max)?

graydon:
> Can you explain which problem this is trying to solve? I do not understand the comment on having a dependency on scrape period (also note that if there are multiple transactions per ledger, you have a similar problem).

Ok so there's a timer (histogram) for the real time and a counter for the instructions. The timer produces a prometheus summary (which has a count-of-timed-events and a sum-of-all-time value) as well as some percentile gauges (which we're ignoring for now). The summary's sum-of-all-time is the only usable part.

> We have both `{"soroban", "host-fn-op", "cpu-insn"}` that tracks the cumulative cost and `{"soroban", "host-fn-op", "invoke-time-nsecs"}` that tracks the  cumulative time spent executing, so in grafana we should be able to plot the ratio of the deltas over arbitrary periods?

Right! So this _is_ what we're plotting in grafana right now -- sum-of-all-model-instructions divided by sum-of-all-real-time -- and if you look at it, it's fairly stable for long periods, and then has a little jump-up every few hours or days.

This is what Jay is talking about (and what I asked about, and what motivates this PR): when someone runs a bunch of txs that each hit a path we're overestimating a bit too aggressively, the model instruction count climbs a lot, but the real time doesn't climb much. And that's fine for like .. a single group of txs. Interesting! But it's also problematic because when those txs are gone, the memory of them is still in the sum-of-all-model-instructions and sum-of-all-real-time. The sums have diverged. So you get a gradually increasing ratio that's now something like 17. Which is misleading. Earlier in the month it was 8 for a while. The 10 for a while. Then 13 for a while. We're _mostly_ not overestimating -- the ratio is mostly stable and around 1:1 ish -- but there are periodic periods of people running txs where we overestimate them, and those accumulate into the sums.

This PR is about trying to make a metric that doesn't do that. Whether it's exactly the right way to do so, I'm less certain about. I want to look at a couple things first.

graydon:
I've thought about this a bit more and done some more precise digging and I think probably this PR is not what we want to do. I apologize, I did actually suggest this PR and I was actually about to rubber stamp it on that basis! But given @MonsieurNicolas's probing question and a little more thought, I think it's maybe not the right thing to do after all.

A few observations:

1. We can fix the drifting-upwards thing in the ratio by using a ratio-of-rates-of-sums in grafana. That is, wrap the two cumulative-sum metrics in `rate(the-cumulative-sum-metric[$__rate_interval])`, or I guess `increase(...[...])` to avoid some surprising infinities / divides-by-zero, and calculate the ratio between those. It'll be much noisier as the rate interval is small, not smoothed over all time, but you can smooth it over days if you like. And it's at least stable-ish. So that's step 1, which I've done, and it works ok: [see here](https://grafana.stellar-ops.com/d/DwjDwtPVz/soroban?orgId=1&from=now-90d&to=now&viewPanel=20). I think that's adequate for estimation purposes / not being confused about the metric anymore. It shows a noisy signal, but with a mean around 9 (instructions per nanosecond) in the current build, which is .. less wildly off base, closer to expectations.
2. I noticed that this PR calculates the ratio of model instructions against `soroban.host-fn-op.invoke-time-nsecs` which is in fact _not_ a medida timer (specifically not the timer `soroban-host-fn-op.exec` used in that graph above!) but rather a [medida meter Jay added in late july](https://github.com/stellar/stellar-core/commit/888db2b2335144fc661357e6a2b5b156d4e95957). The reason grafana is not currently using that new finer-precision metric from July 25 -- that is scoped to the same zone that the budget is scoped to -- is that the code is not yet deployed! Because futurenet is running a build from July 21 still. So I believe what I did in step 1 above will actually be a fair bit nicer if we can move it from the timer to the meter when we get a new futurenet (or testnet) build.

jayz22:
> Can you explain which problem this is trying to solve? I do not understand the comment on having a dependency on scrape period

@MonsieurNicolas I think the scrape rate dependency is eliminated with the ratio-of-rates-of-sums approach @graydon suggested. Both sides of the ratio depends on the rate and thus the rate dependency cancels out. I wasn't able to figure this out earlier. 

With that said, I'm still unsure why the [graph](https://grafana.stellar-ops.com/d/DwjDwtPVz/soroban?orgId=1&from=now-90d&to=now&viewPanel=20) looks noisier and ratio became significantly higher since July. 

The `soroban.host-fn-op.invoke-time-nsecs` meter is supposed to reduce the time (denominator) part so the ratio will end up even higher when it goes live. But I guess it will help with the noise, since it is a better proxy to instruction count. 

I also thought about making the ratio metric (in this PR) a histogram type so that we can see the outliers better, that is maybe an important signal to catch?

MonsieurNicolas:
@jayz22 having as an additional goal to find outliers seems fine to me, but you have to be careful on which metric type to use as medida tends to not do what we need.

I am not sure what we would want to plot/track in a reporting system:
* what would you need for "outlier detection"? I *think* that aggregating using "min" (or "max") of ratios would work, it would tell us worst behavior in any given time period (from a ledger to weeks/months at a time)
* what would you need for "trend analysis"? Metrics that depend on how long a node has been up tend to be useless: we should be able to compare "side-by-side" two nodes applying the same transactions (but potentially using slightly different implementations that only differ in terms of execution time). In this case, maybe the ratio of increase like @graydon mentioned above is enough (assuming this is meaningful over longer periods of time).

jayz22:
Closing this for now.

MonsieurNicolas:
why did you close this @jayz22 ?

jayz22:
@MonsieurNicolas This will be added in https://github.com/stellar/stellar-core/pull/4175/commits/c486379bf4d36e3dfe2c6e10df6f76f93655c07a. (as a histogram, for outlier detection)

PR #3919, opened 2023-09-08
Fixed off by one in eviction scan

### LINKED ISSUE #3913: incorrect iterator reset during eviction scan
During scans for eviction, we reset the iterator if the underlying bucket have changed. However, here it looks like we reset the iterator too early, before the new bucket actually spills: https://github.com/stellar/stellar-core/blob/23f3cfb821d79f3958b615e0404d6f6ff446cdaa/src/bucket/BucketList.cpp#L866

This is because ledgerSeq passed to ledgerShouldSpill isn't technically fully closed yet, so the BucketList isn't updated with spills. This means that we reset the iterator, and scan the same bucket again, then scan the new bucket at a wrong offset (therefore, not evicting some entries). 

One way to fix and simplify this logic is to stop relying on ledger sequence spill calculation completely. Instead, we could have a map of active bucket hashes that we're iterating over, and update it as soon as a new ledger is closed, and reset the iterator then.

SirTyson:
# Description

Resolves #3913 
Fixes eviction scan off by one error in spill ledger calculation.

# Checklist
- [x] Compiles
- [x] Ran all tests

### DIFF
--- a/src/bucket/BucketList.cpp
+++ b/src/bucket/BucketList.cpp
@@ -858,12 +858,18 @@ BucketList::scanForEviction(Application& app, AbstractLedgerTxn& ltx,
             evictionIter.bucketListLevel = firstScanLevel;
         }
 
-        // Check if bucket has changed since iterator was set
+        // Whenever a Bucket changes (spills or receives an incoming spill), the
+        // iterator offset in that bucket is invalidated. After scanning, we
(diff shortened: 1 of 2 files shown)


marta-lokhova (approved):

marta-lokhova:
r+ 7aefc6e58a8561bb146c187246d0537fa39c2cbd

Issue #3765, opened 2023-06-06
Make tx size limits consistent across the codebase


marta-lokhova:
This issue tracks a slight inconsistency between the tx queue and overlay when it comes to validating transactions against max allowed size. Specifically, tx queue takes into account envelope size only (https://github.com/stellar/stellar-core/blob/287fe4c5d7f7ebce1a327cd26f25dd5858cd37a0/src/transactions/TransactionFrame.cpp#L603), while overlay computes the size of the whole message (which includes signatures) https://github.com/stellar/stellar-core/blob/287fe4c5d7f7ebce1a327cd26f25dd5858cd37a0/src/overlay/FlowControlCapacity.cpp#L76

We should ensure we validate size consistently to prevent any unnecessary drops at the overlay layer.

dmkozh:
We'll use full tx size for the fees, so there won't be discrepancy. I'll keep this open to update when the change is merged.

marta-lokhova:
@dmkozh Looks like there are still discrepancies in two places:
* tx validation https://github.com/stellar/stellar-core/blob/03fa3a7a4f76ae1131ced5d9f0342c0af13374ba/src/transactions/TransactionFrame.cpp#L738
* tx resource count https://github.com/stellar/stellar-core/blob/03fa3a7a4f76ae1131ced5d9f0342c0af13374ba/src/transactions/TransactionFrame.cpp#L233

I thought the idea was to use full tx size across codebase; the limit itself is bandwidth-related, and we never flood just the tx portion of the tx envelope, so it makes sense to count the whole transaction. If we don't, this is also problematic for overlay as it relies on `txMaxSizeBytes` for calculations that expect full tx size.

PR #3937, opened 2023-09-14
Make tx size calculation consistent

marta-lokhova:
Follow-up to https://github.com/stellar/stellar-core/pull/3786 to resolve https://github.com/stellar/stellar-core/issues/3765. Also, I updated the calculation on the overlay side to not include XDR type overhead (4 bytes).

### DIFF
--- a/src/overlay/FlowControlCapacity.cpp
+++ b/src/overlay/FlowControlCapacity.cpp
@@ -75,8 +75,7 @@ FlowControlByteCapacity::getCapacityLimits() const
 uint64_t
 FlowControlByteCapacity::getMsgResourceCount(StellarMessage const& msg) const
 {
-
-    return static_cast<uint64_t>(xdr::xdr_argpack_size(msg));
+    return msgBodySize(msg);
 }
 
 void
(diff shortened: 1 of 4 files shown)


dmkozh:
Thanks for the fix, please update the test.

src/overlay/FlowControlCapacity.cpp
```diff
@@ -75,8 +75,55 @@ FlowControlByteCapacity::getCapacityLimits() const
 uint64_t
 FlowControlByteCapacity::getMsgResourceCount(StellarMessage const& msg) const
 {
-
-    return static_cast<uint64_t>(xdr::xdr_argpack_size(msg));
+    switch (msg.type())
```
dmkozh, line 78: How is this different from the generated implementation?

marta-lokhova: we need to exclude size of `type` (additional 4 bytes) for the calculation to be consistent with the rest of the codebase. I updated this function to be less verbose though.

dmkozh:
r+ 0843e92

PR #3961, opened 2023-10-06
Remove unused LedgerCloseMeta struct

### LINKED ISSUE #3923: Cleanup LedgerCloseMetaFrameV1
LedgerCloseMetaFrameV1 is never used since it's been superseded by LedgerCloseMetaFrameV2. Using v1 in protocol 20 is invalid, yet the code currently allows this, which makes it harder to spot bugs related to wrong versioning. We should clean this up and not allow using v1.

SirTyson:
# Description

Resolves #3923 

`LedgerCloseMetaFrameV1` is never used since it's been superseded by `LedgerCloseMetaFrameV2`. Using v1 in protocol 20 is invalid, yet the code currently allows this, which makes it harder to spot bugs related to wrong versioning. This removes LedgerCloseMetaFrameV1 and renames what was previously LedgerCloseMetaFrameV2 to V1.

Opening as a draft until [XDR changes](https://github.com/stellar/stellar-xdr/pull/153) are merged.

# Checklist
- [x] Compiles
- [x] Ran all tests


dmkozh:
> I also removed `src/testdata/ledger-close-meta-v2-protocol-21.json`, which I think is an artifact left over from a preview release fix that we don't need anymore, but I'm not 100% sure.

I'm not sure we don't need it - now that p20 is the 'safe' version, p21 is 'vnext unsafe' version. It makes sense to have it to account for p21 changes, in the same way as we had p20 file for a while. OTOH I don't have a strong opinion about keeping this right now, as we have no p21 changes yet. I'm slightly inclined to keeping though, because it's easy to forget about it when it actually becomes relevant.

SirTyson:
> > I also removed `src/testdata/ledger-close-meta-v2-protocol-21.json`, which I think is an artifact left over from a preview release fix that we don't need anymore, but I'm not 100% sure.
> 
> I'm not sure we don't need it - now that p20 is the 'safe' version, p21 is 'vnext unsafe' version. It makes sense to have it to account for p21 changes, in the same way as we had p20 file for a while. OTOH I don't have a strong opinion about keeping this right now, as we have no p21 changes yet. I'm slightly inclined to keeping though, because it's easy to forget about it when it actually becomes relevant.

Hmm, so I tried to regenerate p21 meta with LedgerCloseMetaV1, and I don't think it's actually possible with the current code base. It appears that the code that generates protocol 21 meta is no longer in main. I could just be doing it incorrectly, but when I run the following command I only get meta dumps up to p20:
```
env GENERATE_TEST_LEDGER_CLOSE_META=1 ./stellar-core test "meta stream contains reasonable meta" --ll FATAL -b -a --all-versions
```

SirTyson:
Closing, included in #3966.

PR #3963, opened 2023-10-11
Code cleanup, fixes to tx queue

marta-lokhova:
Couple changes from code review:
* Deprecate FIRST_VERSION_SUPPORTING_GENERALIZED_TX_SET, as minimum overlay version is past it at this point
* Some renaming for clarity
* Most importantly, some fixes to tx queue: ensure we properly reject unsupported Soroban transactions when in protocol v19 (this can occur on lagging nodes that haven't upgraded yet). To further strengthen the code, initialize Soroban queue post-v20-upgrade. This way we ensure we don't call any functions on Soroban tx queue before the upgrade. This allows us to remove some `protocolVersion` checks (specifically, in flooding).


src/herder/HerderImpl.h
```diff
@@ -230,9 +230,10 @@ class HerderImpl : public Herder
     void writeDebugTxSet(LedgerCloseData const& lcd);
 
     ClassicTransactionQueue mTransactionQueue;
-    SorobanTransactionQueue mSorobanTransactionQueue;
+    std::shared_ptr<SorobanTransactionQueue> mSorobanTransactionQueue;
```
dmkozh, line 233: Why not `unique_ptr` or `optional`? We don't intend to share the ownership for tx queues, do we?

src/herder/HerderImpl.cpp
```diff
@@ -1280,7 +1288,7 @@ HerderImpl::triggerNextLedger(uint32_t ledgerSeqToTrigger,
                                   SOROBAN_PROTOCOL_VERSION))
     {
         txPhases.emplace_back(
-            mSorobanTransactionQueue.getTransactions(lcl.header));
+            mSorobanTransactionQueue->getTransactions(lcl.header));
```
dmkozh, line 1291: Do we need to check that queue is initialized (just in case)? Same below

src/herder/test/HerderTests.cpp
```diff
@@ -3534,6 +3534,91 @@ TEST_CASE("soroban txs each parameter surge priced")
...
+                ->getLedgerManager()
+                .getLastClosedLedgerHeader()
+                .header.ledgerVersion ==
+            static_cast<uint32_t>(SOROBAN_PROTOCOL_VERSION));
+
+    // Now generate Soroban txs
+    auto sorobanConfig =
+        GeneratedLoadConfig::txLoad(LoadGenMode::SOROBAN, 50,
+                                    /* nTxs */ 15, 1, /* offset */ 0);
+    sorobanConfig.skipLowFeeTxs = true;
```
dmkozh, line 3605: Do we need this because we don't have even loadgen spacing for soroban yet? If so, we need to at least check that *some* txs succeeded, otherwise it's possible to pass the test while not accepting a single soroban tx.

src/overlay/Peer.cpp
```diff
@@ -1056,14 +1056,6 @@ Peer::recvGetTxSet(StellarMessage const& msg)
         StellarMessage newMsg;
         if (txSet->isGeneralizedTxSet())
         {
-            if (mRemoteOverlayVersion <
```
dmkozh, line 1059: Why is this no longer necessary?

marta-lokhova: Core's current minimum overlay version is 27, which means it won't even be able to authenticate with peers that are below that. So this check is redundant.

dmkozh: Nice!

src/herder/HerderImpl.cpp
marta-lokhova: If we got to this point (nominating a tx set in v20), Soroban tx queue must exist. If it doesn't, then core is in invalid state. I added extra asserts in both places to ensure the queue is initialized.

src/herder/test/HerderTests.cpp
marta-lokhova: It's because when we upgrade to protocol 20, our Soroban limits are too conservative (even 1 tx/s rate is too high). Technically, we could do another upgrade and bump ledger-wide limits, but for the purpose of this test I just wanted to make sure the queue is initialized and nodes are able to include Soroban txs. Added a check that _some_ txs went in based on your suggestions, so I think that should be sufficient.

dmkozh:
r+ 1655657

src/herder/HerderImpl.h
marta-lokhova: I think this was a remnant of some testing infra that got removed later. Updated to unique_ptr.

marta-lokhova:
CI failed due to `DEFAULT_TEST_REFUNDABLE_FEE` -> `DEFAULT_TEST_RESOURCE_FEE` rename in master, fixing.

dmkozh:
r+ 69a69b0

Issue #3973, opened 2023-10-16
Add functionality for testing transaction subsystem


dmkozh:
Transaction simulation framework has been incompatible for some time and notably won't be compatible with Soroban. Thus we're removing it for now in order to reduce the maintenance burden.

Going forward, we'd probably want to introduce some similar framework that would account for protocol 20 changes in design, especially the Soroban support.

PR #3974, opened 2023-10-17
Remove transaction simulation machinery from Core.

dmkozh:
# Description

Remove transaction simulation machinery from Core.

This feature is currently not well maintained and will be even more broken with Soroban launch.

We might consider bringing this back when we have more time and a better idea about Soroban support requirements: https://github.com/stellar/stellar-core/issues/3973

# Checklist
- [ ] Compiles
- [ ] Ran all tests


src/herder/TxSetFrame.cpp
```diff
@@ -308,9 +330,20 @@ TxSetFrame::makeFromTransactions(TxPhases const& txPhases, Application& app,
...
+            validatedPhases.emplace_back(
+                TxSetUtils::trimInvalid(txs, app, lowerBoundCloseTimeOffset,
+                                        upperBoundCloseTimeOffset, invalid));
+#ifdef BUILD_TESTS
+        }
+        else
+        {
+            validatedPhases.emplace_back(txs);
+        }
+#endif
```
marta-lokhova, line 346: nit: a slightly more readable way would be
```
#ifdef BUILD_TESTS
        if (skipValidation)
        {
             // Add unvalidated
        }
       else
#endif
        {
             // Add validated
        }
```

src/herder/TxSetFrame.cpp
```diff
@@ -336,6 +369,14 @@ TxSetFrame::makeFromTransactions(TxPhases const& txPhases, Application& app,
         txSet->toXDR(xdrTxSet);
         outputTxSet = TxSetFrame::makeFromWire(app, xdrTxSet);
     }
+#ifdef BUILD_TESTS
+    if (skipValidation)
+    {
+        txSet->mHash = outputTxSet->mHash;
```
marta-lokhova, line 375: why not `return outputTxSet;`?

marta-lokhova (changes requested):

src/herder/TxSetFrame.cpp
dmkozh: Updated. I still wrapped the scope in macros just in case (I don't quite like having some arbitrary scopes in production build)

src/herder/TxSetFrame.h
```diff
@@ -162,17 +177,22 @@ class TxSetFrame : public NonMovableOrCopyable
     static TxSetFrameConstPtr
     makeFromTransactions(Transactions txs, Application& app,
                          uint64_t lowerBoundCloseTimeOffset,
-                         uint64_t upperBoundCloseTimeOffset);
+                         uint64_t upperBoundCloseTimeOffset,
+                         Transactions const* orderOverride = nullptr);
```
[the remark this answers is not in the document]
dmkozh: Nice catch, we don't. I suppose I was just thinking that we accept `Phases` as in non-test methods.

src/herder/TxSetFrame.cpp
dmkozh: This is needed to not use the new tx frame objects, otherwise the apply order override wouldn't match the pointers stored in tx set (and we do need them to match in order for fee lookups to work).

src/test/TxTests.cpp
```diff
@@ -497,7 +497,7 @@ closeLedgerOn(Application& app, uint32 ledgerSeq, TimePoint closeTime,
     TxSetFrameConstPtr txSet;
     if (strictOrder)
     {
-        txSet = TxSetFrame::makeFromTransactions(txs, app, 0, 0, &txs);
+        txSet = TxSetFrame::makeFromTransactions(txs, app, 0, 0, true);
```
marta-lokhova, line 500: this should be `strictOrder`, right?

src/herder/TxSetFrame.cpp
marta-lokhova: Oh I see, makes sense, thanks

marta-lokhova (approved):
LGTM, one more small issue in tests. A bit sad we're removing this entirely since it used to be quite useful, but I suppose we can always revive this code if needed.

src/test/TxTests.cpp
dmkozh: We're already in `strictOrder` branch here, so it shouldn't really matter.

marta-lokhova:
r+ fa97a688caae09b9609c493d12327e7a5f780d53

PR #3975, opened 2023-10-17
initialize app's prngs with sane defaults

### LINKED ISSUE #3917: Customize random number generator seed on startup
For test and simulation release, we can have a more deterministic approach. This could be a config parameter. For production version, on node startup, we can start the random number generation, seeding for example with something nondeterministic, for example with time. This is to customize random number generator seed on startup, and to follow up randomness of peer selection.

MonsieurNicolas:
# Description

Resolves #3917

This PR changes the way we initialize PRNGs in the default case.

The main difference is going to be that when running certain scenarios like `catchup`, we properly randomize which history archive is getting hit allowing to properly "round robin" between archives.
Scenarios like `test` should not be effected as they already "reseed" during execution and commands that require determinism should have options for this (`catchup` for example has an argument to force using a specific archive if needed).

### DIFF
--- a/src/util/Math.cpp
+++ b/src/util/Math.cpp
@@ -159,25 +159,44 @@ k_means(std::vector<double> const& points, uint32_t k)
     return centroids;
 }
 
-#ifdef BUILD_TESTS
 static unsigned int lastGlobalSeed{0};
-void
-reinitializeAllGlobalStateWithSeed(unsigned int seed)
+static void
+reinitializeAllGlobalStateWithSeedInternal(unsigned int seed)
 {
     lastGlobalSeed = seed;
     PubKeyUtils::clearVerifySigCache();
     srand(seed);
     gRandomEngine.seed(seed);
-    shortHash::seed(seed);
-    Catch::rng().seed(seed);
-    autocheck::rng().seed(seed);
     randHash::initialize();
 }
+
+void
+initializeAllGlobalState()
+{
+    releaseAssert(lastGlobalSeed == 0);
+    auto const seed = static_cast<unsigned int>(
+        std::chrono::system_clock::now().time_since_epoch().count());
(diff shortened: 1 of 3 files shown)


src/util/Math.cpp
```diff
@@ -159,25 +159,41 @@ k_means(std::vector<double> const& points, uint32_t k)
...
+static
+#endif
+    void
+    reinitializeAllGlobalStateWithSeed(unsigned int seed)
 {
     lastGlobalSeed = seed;
     PubKeyUtils::clearVerifySigCache();
     srand(seed);
     gRandomEngine.seed(seed);
+#ifdef BUILD_TESTS
```
marta-lokhova, line 173: up to you, but given that test vs prod build randomness are sufficiently different, it seems cleaner/more readable to just have two different functions (one behind BUILD_TESTS flag) that initialize randomness.

marta-lokhova (approved):
Looks good, just a few nits

src/util/Math.h
```diff
@@ -53,6 +53,10 @@ rand_element(std::vector<T>& v)
     return v.at(rand_uniform<size_t>(0, v.size() - 1));
 }
 
+// initializes all global state that depend on prngs
+// using sane default (pseudo) random values
+void initializeAllGlobalStateWithoutSeed();
```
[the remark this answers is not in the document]
MonsieurNicolas: yeah I agree, I kept the "re" for the test helper as it's used later

src/util/Math.cpp
MonsieurNicolas: good point. fixed

marta-lokhova:
LGTM, thanks!

marta-lokhova:
r+ 8684848515c3dbaa95c0a4bef72a47e7f618c9c4

PR #3981, opened 2023-10-23
Improve catchup retry strategy

### LINKED ISSUE #3965: Bad retry strategy during catchup
It looks like the code in our catchup logic does not handle failures properly:
* There are unrecoverable failures (running wrong core version), where retrying will never work
   * note that right now the way we return `VERIFY_STATUS_ERR_BAD_LEDGER_VERSION` is *probably* [done too early](https://github.com/stellar/stellar-core/blob/2e2d1ef2e1e4c2a89c61285baaad124af67d06ea/src/catchup/VerifyLedgerChainWork.cpp#L178). If we were going to make this "fatal", we first need to establish trust of the ledger header based off latest SCP.
* Retries on other errors that potentially are transient failures (corrupt history archive like getting an exception `xdr::xdr_bad_message_size`), should implement some sort of exponential back-off
   * note that in the second example below (retries on state corruption), we need to first verify that curr links properly to latest SCP before deciding if it's a fatal error or just a transient error.

When running into errors that don't allow the node to make progress, catchup retries right away and downloads more and more checkpoints with each retry as it tries to verify the ledger chain between latest ledger and last closed ledger on the node. This causes a lot of unnecessary downloads to history archives.

## Example: bad stelllar-core version

```
2023-09-20T15:00:04.391 GCT2W [Ledger INFO] Got consensus: [seq=1615422, prev=fa5ab2, txs=5, ops=5, sv: [ SIGNED@sdf2 txH: 16f9cb, ct: 1695222004, upgrades: [ ] ]]
2023-09-20T15:00:04.392 GCT2W [Tx INFO] applying ledger 1615422 (txs:5, ops:5, base_fee:100)
2023-09-20T15:00:04.396 GCT2W [Ledger INFO] Closed ledger: [seq=1615422, hash=80e056]

2023-09-20T15:42:06.378 GCT2W [History ERROR] Catchup material failed verification - unsupported ledger version, propagating failure

2023-09-20T23:59:37.070 GBWBL [History ERROR] Catchup material failed verification - unsupported ledger version, propagating failure
2023-09-20T23:59:37.070 GBWBL [History ERROR] Catchup material failed verification - unsupported ledger version, propagating failure
2023-09-20T23:59:37.070 GBWBL [History ERROR] Upgrade this stellar-core installation to newest version
2023-09-20T23:59:37.070 GBWBL [History INFO] Verifying ledgers [1615423,1620736)
2023-09-20T23:59:37.070 GBWBL [History INFO] Catching up to ledger 1620735: Failed: verify-ledger-chain
2023-09-20T23:59:37.070 GBWBL [History INFO] Catching up to ledger 1620735: Succeeded: batch-download-ledger-0018a63f-0018bb3f : 252/252 children completed
2023-09-20T23:59:38.594 GBWBL [Herder INFO] Quorum information for 1620784 : {"agree":3,"cost":19192,"delayed":0,"disagree":0,"fail_at":2,"hash":"59d361","lag_ms":13,"ledger":1620784,"missi
ng":0,"phase":"EXTERNALIZE"}

2023-09-20T23:59:38.595 GBWBL [Ledger INFO] Got consensus: [seq=1620786, prev=000000, txs=0, ops=0, sv: [ SIGNED@sdf2 txH: aa6308, ct: 1695254378, upgrades: [ ] ]]
2023-09-20T23:59:38.595 GBWBL [Ledger INFO] Close of ledger 1620786 buffered. mSyncingLedgers has 51 ledgers
2023-09-20T23:59:41.070 GBWBL [History INFO] Catching up to ledger 1620735: downloading ledger files 17/84 (20%)
...
2023-09-21T00:06:18.222 GBWBL [History INFO] Catching up to ledger 1620799: downloading ledger files 85/85 (100%)
2023-09-21T00:06:18.246 GBWBL [History INFO] Catching up to ledger 1620799: Succeeded: batch-download-ledger-0018a63f-0018bb7f : 170/170 children completed
2023-09-21T00:06:18.246 GBWBL [History INFO] Verifying ledgers [1615423,1620800)
2023-09-21T00:06:18.246 GBWBL [History ERROR] Catchup material failed verification - unsupported ledger version, propagating failure
2023-09-21T00:06:18.246 GBWBL [History ERROR] Upgrade this stellar-core installation to newest version
2023-09-21T00:06:18.246 GBWBL [History INFO] Verifying ledgers [1615423,1620800)
2023-09-21T00:06:18.246 GBWBL [History INFO] Catching up to ledger 1620799: Failed: verify-ledger-chain
2023-09-21T00:06:18.246 GBWBL [History INFO] Catching up to ledger 1620799: Succeeded: batch-download-ledger-0018a63f-0018bb7f : 170/170 children completed
2023-09-21T00:06:20.247 GBWBL [History INFO] Catching up to ledger 1620799: downloading ledger files 17/85 (20%)
...
2023-10-10T10:27:52.064 GBWBL [History INFO] Catching up to ledger 1940223: downloading ledger files 3673/5076 (72%)
```

## Example: corrupted local state

In this example, state was corrupted by bug #3958 

```
2023-10-10T23:59:30.394 GA4V2 [History INFO] Catching up to ledger 1949503: downloading ledger files 118/118 (100%)
2023-10-10T23:59:30.417 GA4V2 [History INFO] Catching up to ledger 1949503: Succeeded: batch-download-ledger-001da1ff-001dbf7f : 708/708 children completed
2023-10-10T23:59:30.417 GA4V2 [History INFO] Verifying ledgers [1941982,1949504)
2023-10-10T23:59:30.418 GA4V2 [History INFO] Verifying ledger [seq=1949503, hash=19421f] against SCP hash
2023-10-10T23:59:30.418 GA4V2 [History INFO] Catching up to ledger 1949503: verifying checkpoint 1/118 (0%)
2023-10-10T23:59:30.440 GA4V2 [History ERROR] Bad ledger-header history entry: claimed ledger [seq=1941981, hash=6d207f] does not agree with LCL [seq=1941981, hash=a5777e]
2023-10-10T23:59:30.441 GA4V2 [History ERROR] Catchup material failed verification - hash mismatch, propagating failure
2023-10-10T23:59:30.441 GA4V2 [History ERROR] One or more of history archives may be corrupted. Update HISTORY configuration entry to only contain valid ones
2023-10-10T23:59:30.441 GA4V2 [History INFO] Verifying ledgers [1941982,1949504)
2023-10-10T23:59:30.441 GA4V2 [History INFO] Catching up to ledger 1949503: Failed: verify-ledger-chain
2023-10-10T23:59:30.441 GA4V2 [History INFO] Catching up to ledger 1949503: Failed: download-verify-ledgers-seq
2023-10-10T23:59:30.441 GA4V2 [History WARNING] Catchup failed
2023-10-10T23:59:34.886 GA4V2 [Herder INFO] Quorum information for 1949572 : {"agree":3,"cost":14772,"delayed":0,"disagree":0,"fail_at":2,"hash":"59d361","lag_ms":16,"ledger":1949572,"missi
ng":0,"phase":"EXTERNALIZE"}

2023-10-10T23:59:34.887 GA4V2 [Ledger INFO] Got consensus: [seq=1949574, prev=3747c0, txs=0, ops=0, sv: [ SIGNED@sdf2 txH: 80e32c, ct: 1696982374, upgrades: [ ] ]]
2023-10-10T23:59:34.887 GA4V2 [Ledger INFO] Close of ledger 1949574 buffered. mSyncingLedgers has 7 ledgers
2023-10-10T23:59:34.887 GA4V2 [History INFO] Starting catchup after ensuring checkpoint ledger 1949574 was closed on network
2023-10-10T23:59:34.890 GA4V2 [History INFO] Starting catchup with configuration:
  lastClosedLedger: 1941981
  toLedger: 1949567
  count: 18000
2023-10-10T23:59:34.890 GA4V2 [History INFO] Catching up to ledger 1949567: Downloading state file history/00/1d/bf/history-001dbf7f.json for ledger 1949567
2023-10-10T23:59:34.890 GA4V2 [History INFO] Downloading history archive state: history/00/1d/bf/history-001dbf7f.json
2023-10-10T23:59:35.081 GA4V2 [History INFO] Catching up to ledger 1949567: downloading ledger files 1/119 (0%)
```

marta-lokhova:
Resolves #3965

### DIFF
--- a/src/catchup/CatchupManagerImpl.cpp
+++ b/src/catchup/CatchupManagerImpl.cpp
@@ -107,8 +107,14 @@ void
 CatchupManagerImpl::processLedger(LedgerCloseData const& ledgerData)
 {
     ZoneScoped;
-    if (mCatchupWork && mCatchupWork->isDone())
+    if (catchupWorkIsDone())
     {
+        if (mCatchupWork->getState() == BasicWork::State::WORK_FAILURE &&
+            mCatchupWork->fatalFailure())
+        {
+            CLOG_FATAL(History, "Catchup failed and cannot recover");
+            mCatchupFatalFailure = true;
+        }
         mCatchupWork.reset();
         logAndUpdateCatchupStatus(true);
     }
@@ -181,12 +187,21 @@ CatchupManagerImpl::processLedger(LedgerCloseData const& ledgerData)
         hm.isFirstLedgerInCheckpoint(firstLedgerInBuffer) &&
         firstLedgerInBuffer < lastLedgerInBuffer)
     {
-        message =
-            fmt::format(FMT_STRING("Starting catchup after ensuring checkpoint "
-                                   "ledger {:d} was closed on network"),
-                        lastLedgerInBuffer);
-
-        startOnlineCatchup();
+        // No point in processing ledgers as catchup won't ever be able to
+        // succeed
+        if (!mCatchupFatalFailure)
+        {
+            message = fmt::format(
+                FMT_STRING("Starting catchup after ensuring checkpoint "
+                           "ledger {:d} was closed on network"),
+                lastLedgerInBuffer);
+            startOnlineCatchup();
+        }
+        else
+        {
+            CLOG_FATAL(History, "Skipping catchup: incompatible core version "
+                                "or invalid local state");
+        }
     }
     else
     {
--- a/src/catchup/VerifyLedgerChainWork.cpp
+++ b/src/catchup/VerifyLedgerChainWork.cpp
@@ -84,10 +84,31 @@ verifyLastLedgerInCheckpoint(LedgerHeaderHistoryEntry const& ledger,
     return HistoryManager::VERIFY_STATUS_OK;
 }
 
+template <typename T>
+void
+trySetFuture(std::promise<T>& promise, T value)
+{
+    try
+    {
+        promise.set_value(value);
+    }
+    catch (std::future_error const& err)
+    {
+        // If promise value has already been set, ignore exception from
+        // attempting to set it twice. This will happen if we're reset
+        // and re-run.
+        if (err.code() != std::future_errc::promise_already_satisfied)
+        {
+            throw;
+        }
+    }
+}
+
 VerifyLedgerChainWork::VerifyLedgerChainWork(
     Application& app, TmpDir const& downloadDir, LedgerRange const& range,
     LedgerNumHashPair const& lastClosedLedger,
     std::shared_future<LedgerNumHashPair> trustedMaxLedger,
+    std::promise<bool>&& fatalFailure,
     std::shared_ptr<std::ofstream> outputStream)
     : BasicWork(app, "verify-ledger-chain", BasicWork::RETRY_NEVER)
     , mDownloadDir(downloadDir)
@@ -97,6 +118,7 @@ VerifyLedgerChainWork::VerifyLedgerChainWork(
                           : mApp.getHistoryManager().checkpointContainingLedger(
                                 mRange.last()))
     , mLastClosed(lastClosedLedger)
+    , mFatalFailurePromise(std::move(fatalFailure))
     , mTrustedMaxLedger(trustedMaxLedger)
     , mVerifiedMinLedgerPrevFuture(mVerifiedMinLedgerPrev.get_future().share())
     , mOutputStream(outputStream)
@@ -130,6 +152,8 @@ VerifyLedgerChainWork::onReset()
                           ? 0
                           : mApp.getHistoryManager().checkpointContainingLedger(
                                 mRange.last());
+    mLocalFailure.reset();
+    mHasTrustedHash = false;
 }
 
 HistoryManager::LedgerVerificationStatus
@@ -178,10 +202,15 @@ VerifyLedgerChainWork::verifyHistoryOfSingleCheckpoint()
         if (curr.header.ledgerVersion >
             mApp.getConfig().LEDGER_PROTOCOL_VERSION)
         {
-            return HistoryManager::VERIFY_STATUS_ERR_BAD_LEDGER_VERSION;
+            // Note that local state does not agree with the archives; depending
+            // on the presence of trusted hash
+            mLocalFailure =
+                HistoryManager::VERIFY_STATUS_ERR_BAD_LEDGER_VERSION;
         }
 
         // Verify ledger with local state by comparing to LCL
+        // When checking against LCL, see it the local node is in the bad state,
+        // or if the archive is in a bad state (in which case, retry)
         if (curr.header.ledgerSeq == mLastClosed.first)
         {
             if (sha256(xdr::xdr_to_opaque(curr.header)) != *mLastClosed.second)
@@ -192,7 +221,7 @@ VerifyLedgerChainWork::verifyHistoryOfSingleCheckpoint()
                            LedgerManager::ledgerAbbrev(curr),
                            LedgerManager::ledgerAbbrev(mLastClosed.first,
                                                        *mLastClosed.second));
-                return HistoryManager::VERIFY_STATUS_ERR_BAD_HASH;
+                mLocalFailure = HistoryManager::VERIFY_STATUS_ERR_BAD_HASH;
             }
         }
         // Verify LCL that is just before the first ledger in range
@@ -207,7 +236,7 @@ VerifyLedgerChainWork::verifyHistoryOfSingleCheckpoint()
                            LedgerManager::ledgerAbbrev(curr),
                            LedgerManager::ledgerAbbrev(mLastClosed.first,
                                                        *mLastClosed.second));
-                return lclResult;
+                mLocalFailure = lclResult;
             }
         }
 
@@ -315,6 +344,8 @@ VerifyLedgerChainWork::verifyHistoryOfSingleCheckpoint()
         releaseAssert(futureIsReady(mTrustedMaxLedger));
 
         incoming = mTrustedMaxLedger.get();
+        mHasTrustedHash = incoming.second.has_value();
+
         releaseAssert(incoming.first == curr.header.ledgerSeq);
         CLOG_INFO(History, "{} ledger {} against SCP hash",
                   (incoming.second ? "Verifying" : "Skipping verification for"),
@@ -355,20 +386,7 @@ VerifyLedgerChainWork::verifyHistoryOfSingleCheckpoint()
         outgoing.second =
             std::make_optional<Hash>(first.header.previousLedgerHash);
 
-        try
-        {
-            mVerifiedMinLedgerPrev.set_value(outgoing);
-        }
-        catch (std::future_error const& err)
-        {
-            // If outgoing link has already been set, ignore exception from
-            // attempting to set it twice. This will happen if we're reset
-            // and re-run.
-            if (err.code() != std::future_errc::promise_already_satisfied)
-            {
-                throw;
-            }
-        }
+        trySetFuture<LedgerNumHashPair>(mVerifiedMinLedgerPrev, outgoing);
 
         // Also write the max ledger in this min-valued checkpoint, as it
         // will be read by catchup as the ledger number for bucket-apply.
@@ -383,6 +401,8 @@ VerifyLedgerChainWork::verifyHistoryOfSingleCheckpoint()
 void
 VerifyLedgerChainWork::onSuccess()
 {
+    trySetFuture<bool>(mFatalFailurePromise, mLocalFailure && mHasTrustedHash);
+
     if (mOutputStream)
     {
         for (auto const& pair : mVerifiedLedgers)
@@ -391,6 +411,12 @@ VerifyLedgerChainWork::onSuccess()
                              << binToHex(*pair.second) << "\"],";
         }
     }
+}
+
+void
+VerifyLedgerChainWork::onFailureRaise()
+{
+    trySetFuture<bool>(mFatalFailurePromise, mLocalFailure && mHasTrustedHash);
 }
 
 BasicWork::State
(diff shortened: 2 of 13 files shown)


src/catchup/VerifyLedgerChainWork.cpp
```diff
@@ -178,10 +202,15 @@ VerifyLedgerChainWork::verifyHistoryOfSingleCheckpoint()
         if (curr.header.ledgerVersion >
```
SirTyson, line 202: Should we also set `mLocalFailure` when we catch an xdr exception? Or does it not matter because we won't be able to see if we can set `mHasTrustedHash` to true anyway?

src/catchup/VerifyLedgerChainWork.cpp
```diff
@@ -178,10 +202,15 @@ VerifyLedgerChainWork::verifyHistoryOfSingleCheckpoint()
...
+            // Note that local state does not agree with the archives; depending
+            // on the presence of trusted hash
+            mLocalFailure =
+                HistoryManager::VERIFY_STATUS_ERR_BAD_LEDGER_VERSION;
         }
 
         // Verify ledger with local state by comparing to LCL
+        // When checking against LCL, see it the local node is in the bad state,
+        // or if the archive is in a bad state (in which case, retry)
         if (curr.header.ledgerSeq == mLastClosed.first)
```
SirTyson, line 214: Should be `else if` now that we no longer longer return early above.

src/catchup/VerifyLedgerChainWork.cpp
```diff
@@ -84,10 +84,31 @@ verifyLastLedgerInCheckpoint(LedgerHeaderHistoryEntry const& ledger,
     return HistoryManager::VERIFY_STATUS_OK;
 }
 
+template <typename T>
+void
+trySetFuture(std::promise<T>& promise, T value)
```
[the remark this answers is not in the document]
marta-lokhova: > Instead of using shared_future with promises, I think it might be better to just use std::shared_ptr<std::atomic<bool>>. You never actually access the shared_future in an async way, all accesses are concurrent, so you're not benefiting from the async interface. std::atomic<bool> can be set multiple times and we wouldn't have to catch any exceptions. You could just have a single assert in the constructor that the initial shared_ptr passed to the constructor in non-null and then you should be good to go without any further error checking.

So I actually had a similar debate :) I agree that we aren't doing anything async, and perhaps using futures is a bit of an overkill, but I picked it for two reasons (lmk if you disagree):
* The Work subsystem doesn't really have a good way to share state between works. https://github.com/stellar/stellar-core/pull/2598 introduced the promise/future pattern to partially deal with this problem. So I wanted to stay consistent with the codebase, since in this change we're trying to do the same thing.
* I think futures provide more clarity to the interface: this way we know that it is a responsibility of verify-ledger-chain to properly set the state at the end. Similarly, catchup-work can only check and maybe retrieve the state, but it can't make any modifications to it. 

> Currently, it's actually possible to miss a fatal error if we have bad local state and if the first history archive we download from is corrupt. On our first download, mHasTrustedHash will be false, so we set the shared_future to false. On the next download assuming the next history archive we try is correct, mHasTrustedHash and mLocalFailure will both be true, but we will be unable to set the fatal flag and would erroneously retry downloading history again a 3rd time. This would only occur if a node tries to download history 3+ times, the first history archive is corrupt, and if local state is also corrupt, so definitely a corner case, but atomic<bool> would help us there.

This is a good point! I _think_ the current implementation should still work though: we don't do retries on verify-ledger-chain, so if we fail the first time with the corrupt archive, catchup would fail and get reset (resetting the future as well) before a new catchup is triggered.

src/catchup/VerifyLedgerChainWork.cpp
marta-lokhova: Yeah, my thinking was that we won't be able to verify that the hash is trusted, and given that fatal failure causes us to _never_ retry catchup (which is serious if done erroneously), we should allow catchup to retry in this case.

src/catchup/VerifyLedgerChainWork.cpp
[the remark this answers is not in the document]
SirTyson: Yeah, I see where you're coming from. I don't think it's a huge deal so I'm good with leaving it as is, but using an async interface synchronously that we have to set in a try-catch loop...makes me feel a certain type of way lol.

src/catchup/VerifyLedgerChainWork.cpp
marta-lokhova: mm, should it? I guess it depends on what error we want to report first. i was thinking local corrupt state is probably more important to notify the user about, no?

src/catchup/VerifyLedgerChainWork.cpp
graydon: Hmm, doesn't this mean validators that are sufficiently old that that they can't parse the current network header will retry forever? (I had the same concern here).

(I don't mind leaving this as a followup if so, but I think it's at least a reasonably-plausible case we currently don't handle..)

src/catchup/VerifyLedgerChainWork.h
```diff
@@ -27,6 +27,13 @@ class VerifyLedgerChainWork : public BasicWork
     uint32_t mCurrCheckpoint;
     LedgerNumHashPair const mLastClosed;
 
+    std::optional<HistoryManager::LedgerVerificationStatus> mLocalFailure;
+    bool mHasTrustedHash{false};
```
graydon, line 31: I misunderstood these fields on first-through-Nth-reading so I suspect some comments would help and possibly a rename.
```suggestion
    // Record any instance where the chain we're verifying disagrees with the local node state.
    // This _might_ mean we can't possibly catch up (eg. we're on a too-old core or something)
    // or it _might_ mean we're reading from a corrupt archive. The next flag differentiates cases.
    std::optional<HistoryManager::LedgerVerificationStatus> mChainDisagreesWithLocalState;
    
    // Record whether we were instantiated with an initial head-of-chain hash that we trust to
    // be the network consensus. If this is true, then any local-node disagreement is our problem,
    // not a possible corrupt archive, and we should treat it as an unrecoverable error (i.e. set the
    // promise below, to stop further catchup-retries).
    bool mHasTrustedHash{false};
```

src/history/test/HistoryTests.cpp
```diff
@@ -224,91 +225,136 @@ TEST_CASE("Ledger chain verification", "[ledgerheaderverification]")
...
+            tmpDir, ledgerRange, lclPair, ledgerRangeEndFuture,
+            std::move(fataFailurePromise));
         REQUIRE(expectedState == w->getState());
+        REQUIRE(fatalFailureFuture.valid());
+        return fatalFailureFuture.get();
     };
 
     LedgerHeaderHistoryEntry lcl, last;
-    LOG_DEBUG(DEFAULT_LOG, "fully valid");
+    SECTION("fully valid")
```
graydon, line 253: Just checking -- the change to `SECTION`s makes the test re-run once for each `SECTION` -- is this intentional? eg. to reset the fatal failure promise?

src/catchup/CatchupManagerImpl.cpp
```diff
@@ -181,12 +187,21 @@ CatchupManagerImpl::processLedger(LedgerCloseData const& ledgerData)
...
+        {
+            message = fmt::format(
+                FMT_STRING("Starting catchup after ensuring checkpoint "
+                           "ledger {:d} was closed on network"),
+                lastLedgerInBuffer);
+            startOnlineCatchup();
+        }
+        else
+        {
+            CLOG_FATAL(History, "Skipping catchup: incompatible core version "
```
graydon, line 202: This is interesting -- core will just loop forever at this point right? Should we notice at some higher level and shut down the node with an error? I mean the user might still have core restart endlessly but maybe they're more likely to notice that than just endless LOG_FATALs?

graydon (changes requested):
Ok, I got confused about the meaning of "local failure" at first (comments suggesting a little clarifying addition -- I thought it meant like "any local failure in this method") but otherwise nice work! As far as my brain can simulate, it ought to do what it's aiming to do :)

src/history/test/HistoryTests.cpp
marta-lokhova: This is intentional: this test was broken before. Every "section" generated ledger chain files, which I think all shared the same directory/files. Out XDR output stream opens files in append mode, so all sections below "fully valid" were actually failing with hash mismatch (since we pass a new trusted hash, but verify-ledger-chain would read old data from "fully valid" section). Of course, we haven't noticed this for years, because the test only checks that the resulting work is in WORK_FAILURE state :/

src/catchup/CatchupManagerImpl.cpp
marta-lokhova: That's fair, though if we shut down the node and restart it, we'll go through the catchup flow again, since we'll lose any in-memory state. So maybe we just keep it as-is, wdyt?

src/catchup/VerifyLedgerChainWork.h
marta-lokhova: thanks, these are great! added.

src/catchup/VerifyLedgerChainWork.cpp
graydon: ok

src/history/test/HistoryTests.cpp
graydon: oh wow. I'm .. very sad to hear this indeed. thanks for fixing.

graydon: oh wow. I'm .. very sad to hear this indeed. thanks for fixing.

src/catchup/CatchupManagerImpl.cpp
graydon: yeah, fair point!

src/catchup/VerifyLedgerChainWork.cpp
graydon: yeah I think this is ok -- the second check is possibly-more important to report, and if either happens we're equally sunk.

graydon:
r+ 2d0b8f6

PR #3991, opened 2023-10-26
Added BucketListSizeSnapshots upgrade test

### LINKED ISSUE #3798: More comprehensive state expiration unit tests

SirTyson:
# Description

Resolves #3798

Adds upgrade tests for `BucketListSizeSnapshots` upgrades.

# Checklist
- [x] Compiles
- [x] Ran all tests

### DIFF
--- a/src/herder/test/UpgradesTests.cpp
+++ b/src/herder/test/UpgradesTests.cpp
@@ -245,6 +245,25 @@ makeMaxContractSizeBytesTestUpgrade(
     return makeConfigUpgradeSet(ltx, configUpgradeSet, expiredEntry, type);
 }
 
+ConfigUpgradeSetFrameConstPtr
+makeBucketListSizeWindowSampleSizeTestUpgrade(Application& app,
+                                              AbstractLedgerTxn& ltx,
+                                              uint32_t newWindowSize)
+{
+    // Modify window size
+    auto sas = app.getLedgerManager()
+                   .getSorobanNetworkConfig(ltx)
+                   .stateArchivalSettings();
+    sas.bucketListSizeWindowSampleSize = newWindowSize;
+
+    // Make entry for the upgrade
+    ConfigUpgradeSet configUpgradeSet;
+    auto& configEntry = configUpgradeSet.updatedEntry.emplace_back();
+    configEntry.configSettingID(CONFIG_SETTING_STATE_ARCHIVAL);
+    configEntry.stateArchivalSettings() = sas;
+    return makeConfigUpgradeSet(ltx, configUpgradeSet);
+}
+
 LedgerKey
 getMaxContractSizeKey()
 {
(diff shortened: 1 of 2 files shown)


src/herder/test/UpgradesTests.cpp
```diff
@@ -843,6 +862,86 @@ TEST_CASE("config upgrades applied to ledger", "[soroban][upgrades]")
                 CONFIG_SETTING_CONTRACT_MAX_SIZE_BYTES);
         REQUIRE(sorobanConfig.maxContractSizeBytes() == 32768);
     }
+
+    SECTION("modify BucketListSizeWindowSampleSize")
+    {
+        auto upgradeSize = [&](uint32_t size) {
```
marta-lokhova, line 868: nit: should this be a separate function like in other sections so we can re-use it later?

src/herder/test/UpgradesTests.cpp
```diff
@@ -843,6 +862,86 @@ TEST_CASE("config upgrades applied to ledger", "[soroban][upgrades]")
...
+                ltx2.commit();
+            }
+
+            REQUIRE(configUpgradeSet);
+            executeUpgrade(*app, makeConfigUpgrade(*configUpgradeSet));
+        };
+
+        SECTION("decrease size")
+        {
+            upgradeSize(20);
```
marta-lokhova, line 897: nit: extract 20 into a const to reuse on line 905 and 914 for clarity (same in the other section)

src/herder/test/UpgradesTests.cpp
```diff
@@ -843,6 +862,86 @@ TEST_CASE("config upgrades applied to ledger", "[soroban][upgrades]")
...
+                {
+                    ++expectedValue;
+                }
+
+                REQUIRE(cfg.mBucketListSizeSnapshots[i] == expectedValue);
+                sum += expectedValue;
+            }
+
+            // Verify average has been properly updated as well
+            REQUIRE(cfg.getAverageBucketListSize() == (sum / 40));
```
marta-lokhova, line 941: is there value in testing upgrades to 0 (should throw instead of UB), and to the same size as before (should no-op)?

marta-lokhova (changes requested):
Couple of small suggestions, otherwise looks fine

src/herder/test/UpgradesTests.cpp
SirTyson: I don't think so. I doubt we'll need to do this upgrade in other tests as this is pretty obscure and this test should be sufficient. This also isn't just an upgrade, but we also modify the BucketList size values before the actual upgrade. I think this should just be limited to this particular test, as I don't think other upgrade tests will need or expect these dummy values.

src/herder/test/UpgradesTests.cpp
SirTyson: Good idea

src/ledger/NetworkConfig.h
```diff
@@ -355,6 +352,9 @@ class SorobanNetworkConfig
   public:
 #endif
 
+    void writeBucketListSizeWindow(AbstractLedgerTxn& ltxRoot) const;
```
marta-lokhova, line 355: ah, didn't see this last time I reviewed, but I think these functions should be kept private. if you want to use them in tests, safer would be to make public methods with the same name plus "forTesting" behind BUILD_TESTS.

marta-lokhova (changes requested):

src/ledger/NetworkConfig.h
SirTyson: I believe this is essentially what this change is doing. When BUILD_TESTS is not set, the functions are still private. All this does is make the functions public when BUILD_TESTS is set, which I personally prefer over redefining the functions with a `forTesting` appended to them, especially given that NetworkConfig already does this for member access for testing.

marta-lokhova: ah yes, good point, the diff view did not show
```
#ifdef BUILD_TESTS
  public:
#endif
```

marta-lokhova:
r+ 6a5ed0d75180511ef82923d1b9b4ab87a8ed0948

PR #3995, opened 2023-10-27
Add utility for building random wasms

graydon:
This adds a helper to the rust side to generate random wasms. They should all be roughly the size of requested input plus 94 (or 96) bytes for the header.


dmkozh (approved):
Thanks! I'll plumb this to the core tests.

dmkozh:
r+ 532abcb

PR #4015, opened 2023-11-08
Add Soroban transactions to catchup tests

marta-lokhova:
Add basic Soroban transactions to random ledger generation logic to test publish/catchup. We can use @SirTyson's loadgen changes once we have them to generate more interesting Soroban txs in these tests as well

### DIFF
--- a/src/history/test/HistoryTestsUtils.cpp
+++ b/src/history/test/HistoryTestsUtils.cpp
@@ -427,6 +427,8 @@ CatchupSimulation::generateRandomLedger(uint32_t version)
 
     std::vector<TransactionFrameBasePtr> txs;
     std::vector<TransactionFrameBasePtr> sorobanTxs;
+    bool check = false;
+
     if (ledgerSeq < 5)
     {
         txs.push_back(
@@ -482,12 +484,17 @@ CatchupSimulation::generateRandomLedger(uint32_t version)
                 SOROBAN_PROTOCOL_VERSION))
         {
             SorobanResources res;
-            res.instructions = InitialSorobanNetworkConfig::TX_MAX_INSTRUCTIONS;
+            res.instructions =
+                mApp.getLedgerManager().maxSorobanTransactionResources().getVal(
+                    Resource::Type::INSTRUCTIONS) /
+                10;
+            res.writeBytes = 100'000;
             uint32_t inclusion = 100;
             sorobanTxs.push_back(createUploadWasmTx(
                 mApp, stroopy, inclusion, DEFAULT_TEST_RESOURCE_FEE, res));
             sorobanTxs.push_back(createUploadWasmTx(
                 mApp, eve, inclusion * 5, DEFAULT_TEST_RESOURCE_FEE, res));
+            check = true;
         }
     }
 
@@ -517,7 +524,19 @@ CatchupSimulation::generateRandomLedger(uint32_t version)
                                           upgrades, mApp.getConfig().NODE_SEED);
 
     mLedgerCloseDatas.emplace_back(ledgerSeq, txSet, sv);
+
+    auto& txsSucceeded =
+        mApp.getMetrics().NewCounter({"ledger", "apply", "success"});
+    auto lastSucceeded = txsSucceeded.count();
+
     lm.closeLedger(mLedgerCloseDatas.back());
+
+    if (check)
+    {
+        // Make sure all classic transactions and at least some Soroban
+        // transactions succeeded
+        REQUIRE(txsSucceeded.count() > lastSucceeded + phases[0].size());
+    }
 
     auto const& lclh = lm.getLastClosedLedgerHeader();
     mLedgerSeqs.push_back(lclh.header.ledgerSeq);
--- a/src/history/test/HistoryTestsUtils.cpp
+++ b/src/history/test/HistoryTestsUtils.cpp
@@ -64,6 +64,7 @@ TmpDirHistoryConfigurator::configure(Config& cfg, bool writable) const
     }
 
     cfg.HISTORY[d] = HistoryArchiveConfiguration{d, getCmd, putCmd, mkdirCmd};
+    cfg.TESTING_SOROBAN_HIGH_LIMIT_OVERRIDE = true;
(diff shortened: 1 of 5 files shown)


SirTyson (approved):
LGTM

src/history/test/HistoryTestsUtils.cpp
```diff
@@ -468,16 +475,36 @@ CatchupSimulation::generateRandomLedger(uint32_t version)
...
             txs.push_back(carol.tx({payment(bob, small)}));
         }
+
+        // Add soroban transactions
+        if (protocolVersionStartsFrom(
+                lm.getLastClosedLedgerHeader().header.ledgerVersion,
+                SOROBAN_PROTOCOL_VERSION))
+        {
+            SorobanResources res;
+            res.instructions = InitialSorobanNetworkConfig::TX_MAX_INSTRUCTIONS;
```
dmkozh, line 485: Is the intent here to get some of the txs applied? (I'd expect so). If so, you should use the current tx max (because initial is pretty low). You also need to set write bytes high enough.

src/history/test/HistoryTestsUtils.cpp
```diff
@@ -64,6 +64,7 @@ TmpDirHistoryConfigurator::configure(Config& cfg, bool writable) const
     }
 
     cfg.HISTORY[d] = HistoryArchiveConfiguration{d, getCmd, putCmd, mkdirCmd};
+    cfg.TESTING_SOROBAN_HIGH_LIMIT_OVERRIDE = true;
```
dmkozh, line 67: Is it possible to use `overrideSorobanNetworkConfigForTest` instead? I realize you'd need to inject it in a different place because we don't have app access here.

marta-lokhova: Hmm, it looks like `overrideSorobanNetworkConfigForTest` is kind of broken. It commits directly to the DB, but not to the bucketlist, so the two state snapshots end up diverging and subsequently cause catching up node to diverge. So our options are either to amend the genesis ledger creation (what TESTING_SOROBAN_HIGH_LIMIT_OVERRIDE is doing), or submit a proper upgrade and let it get committed through consensus.

dmkozh: Ah, you're right, this is not supposed to work with more complex setups where nodes need to catchup. It's not necessarily 'broken', but it's just a utility that only works on standalone apps or simulated networks where every node has the same override. This is good enough for now then.

src/history/test/HistoryTestsUtils.cpp
marta-lokhova: updated. also added a check that at least some soroban txs succeeded

src/history/test/HistoryTestsUtils.cpp
marta-lokhova: We don't need to act on it now, but I do think we need a more universal solution for network config overrides for testing that ensures changes to ledger state are correctly reflected everywhere, _and_ the whole simulated network applies the override (without that, nodes also diverge). Otherwise, it's easy right now to get stuck debugging divergences due to test misconfigurations, and folks outside of core had similar problems doing testing (like with eviction iterator config overrides, for example)

dmkozh: The only way to do that properly is to run the proper upgrade procedure, which is a bit cumbersome for most of the core tests. We do introduce the tools to simplify that, but I don't think we need to try to apply them everywhere. I do want to get rid of `TESTING_SOROBAN_HIGH_LIMIT_OVERRIDE` flag though, as it's the most error-prone thing (`overrideSorobanNetworkConfigForTest` is an internal function, so the impact of potential confusion is very limited).

marta-lokhova: hmm, how is it less error prone? the config flag at least updates both the DB and the bucketlist. it doesn't need to be user-facing, we have configs that are internal, for testing only

marta-lokhova: again I don't want to spend too much time on it right now, but it does seem like the current infra potentially creates a debugging time sink for people

marta-lokhova:
sorry looks like I accidentally pushed a bogus commit. this is fixed now

dmkozh:
r+ 4f07e06

src/history/test/HistoryTestsUtils.cpp
dmkozh: The flag is something that we expose in non-test builds, so that's a potential footgun for prod builds. It also very easily leads to divergence when not every node has this flag (and even if every node does have it, I'm still not sure if the timing of applying the override would always be correct).
The test function is there only for tests and we can set clear expectations on what it does/doesn't do. That doesn't mean that it can't be improved, but it does make it much safer.
FWIW initially it only modified the values in-memory, so there wasn't even a way to use it for multi-node setups (which meant that you have to do proper upgrade there, which is more cumbersome, but also less error-prone). I think the way to go further is to maybe create a helper in simulation that runs the upgrade, so that tests with multiple nodes could use it.
Anyway, it's definitely a tangent topic within the context of this PR, just something we need to clean up eventually.

marta-lokhova: i mean, it doesn't have to be included in prod builds, it's easy to make it test-only. setting the flag on all nodes is the same problem as calling the function on all nodes. its actually safer right now to use the flag because it modified the genesis ledger, whereas calling overrideSorobanNetworkConfigForTest in a wrong place for each node can create a divergence (if nodes are on different ledgers). It is also useful to have the config because we can use it in supercluster. agreed that it's tangential, maybe I'll just open an issue and we can decide what to do later

dmkozh:
r+ 0842288

marta-lokhova:
Not sure what's going with the CI: looks like `Catchup with protocol upgrade` is failing and I'm seeing a bunch of 
```
Process 61237 stopped and restarted: thread 2 received signal: SIGCHLD
```
This tests passes locally on my machine

dmkozh:
r+ 83a30e9

PR #4025, opened 2023-11-12
Soroban loadgen v2

### LINKED ISSUE #3706: loadgen: support a few Soroban transactions
We need to perform various load tests analysis for Soroban traffic.

If we want to reuse our existing load testing infra, we need to update `loadgen` to support a few minimalists contracts:
* simple payment, SAC and generic
* AMM swap equivalent

We should also use those as part of our smoke tests

SirTyson:
# Description

Resolves #3706

Adds `SOROBAN_INVOKE` mode to LoadGen tests, where generated Soroban invoke TXs  applied. Generated TXs are valid and applied by validators in order to test the Soroban TX subsystem. The generated TXs are of arbitrary size and require an arbitrary amount of disk IO and CPU work at apply time. The `GeneratedLoadConfig` `nDataEntries` and `kiloBytesPerDataEntry` determine how much disk IO each TX will require. Currently, for each TX loadgen picks a random TX size and amount of CPU work such that the TX respects NetworkConfig limits, but additional parameters can be added to control this if required.

For each source account, loadgen sets up a unique contract instance and a set of `ContractData` entries that  the TX will load on each invocation. Each account has its own instance and keys to produce the most work possible, but share the same WASM entry. A future update will add support for unique WASM entries. Due to Soroban TX op limits, it takes several dependent TXs across multiple ledgers to set up these contract instances (see `SorobanPhase` for setup).

The previous lodagen Soroban test produces random WASM upload operations, most of which are invalid and never applied. This is still useful for herder and overlay testing and has been renamed to `SOROBAN_WASM`.

# Checklist
- [x] Compiles
- [x] Ran all tests

### DIFF
--- a/src/main/CommandHandler.cpp
+++ b/src/main/CommandHandler.cpp
@@ -1249,11 +1249,22 @@ CommandHandler::generateLoad(std::string const& params, std::string& retStr)
 
         uint32_t numItems = isCreate ? cfg.nAccounts : cfg.nTxs;
         std::string itemType = isCreate ? "accounts" : "txs";
-
-        retStr +=
+        Json::Value res;
+        res["status"] =
             fmt::format(FMT_STRING(" Generating load: {:d} {:s}, {:d} tx/s"),
                         numItems, itemType, cfg.txRate);
         mApp.generateLoad(cfg);
+
+        if (cfg.mode == LoadGenMode::SOROBAN_CREATE_UPGRADE)
+        {
+            auto configUpgradeKey =
+                mApp.getLoadGenerator().getConfigUpgradeSetKey(cfg);
+            auto configUpgradeKeyStr = stellar::decoder::encode_b64(
+                xdr::xdr_to_opaque(configUpgradeKey));
+            res["config_upgrade_set_key"] = configUpgradeKeyStr;
+        }
+
+        retStr = res.toStyledString();
     }
     else
     {
--- a/src/main/CommandHandler.cpp
+++ b/src/main/CommandHandler.cpp
@@ -1173,6 +1173,23 @@ CommandHandler::generateLoad(std::string const& params, std::string& retStr)
         // Only for MIXED_TX mode; fraction of DEX transactions.
         cfg.dexTxPercent =
             parseOptionalParamOrDefault<uint32_t>(map, "dextxpercent", 0);
+        // Only for SOROBAN_INVOKE mode, resource consumption bounds
+        cfg.nDataEntriesLow =
+            parseOptionalParamOrDefault<uint32_t>(map, "dataentrieslow", 0);
+        cfg.nDataEntriesHigh =
+            parseOptionalParamOrDefault<uint32_t>(map, "dataentrieshigh", 0);
+        cfg.kiloBytesPerDataEntryLow =
+            parseOptionalParamOrDefault<uint32_t>(map, "kilobyteslow", 0);
+        cfg.kiloBytesPerDataEntryHigh =
+            parseOptionalParamOrDefault<uint32_t>(map, "kilobyteshigh", 0);
+        cfg.txSizeBytesLow =
+            parseOptionalParamOrDefault<uint32_t>(map, "txsizelow", 0);
+        cfg.txSizeBytesHigh =
+            parseOptionalParamOrDefault<uint32_t>(map, "txsizehigh", 0);
+        cfg.instructionsLow =
+            parseOptionalParamOrDefault<uint64_t>(map, "cpulow", 0);
+        cfg.instructionsHigh =
+            parseOptionalParamOrDefault<uint64_t>(map, "cpuhigh", 0);
 
         if (cfg.maxGeneratedFeeRate)
         {
--- a/src/main/CommandHandler.cpp
+++ b/src/main/CommandHandler.cpp
@@ -1173,6 +1173,7 @@ CommandHandler::generateLoad(std::string const& params, std::string& retStr)
         // Only for MIXED_TX mode; fraction of DEX transactions.
         cfg.dexTxPercent =
             parseOptionalParamOrDefault<uint32_t>(map, "dextxpercent", 0);
+
         // Only for SOROBAN_INVOKE mode, resource consumption bounds
         cfg.nDataEntriesLow =
             parseOptionalParamOrDefault<uint32_t>(map, "dataentrieslow", 0);
@@ -1190,6 +1191,50 @@ CommandHandler::generateLoad(std::string const& params, std::string& retStr)
             parseOptionalParamOrDefault<uint64_t>(map, "cpulow", 0);
         cfg.instructionsHigh =
             parseOptionalParamOrDefault<uint64_t>(map, "cpuhigh", 0);
+
+        // SOROBAN_CREATE_UPGRADE parameters
+        cfg.maxContractSizeBytes =
+            parseOptionalParamOrDefault<uint32_t>(map, "mxcntrctsz", 0);
+        cfg.maxContractDataKeySizeBytes =
+            parseOptionalParamOrDefault<uint32_t>(map, "mxcntrctkeysz", 0);
+        cfg.maxContractDataEntrySizeBytes =
+            parseOptionalParamOrDefault<uint32_t>(map, "mxcntrctdatasz", 0);
+        cfg.ledgerMaxInstructions =
+            parseOptionalParamOrDefault<uint64_t>(map, "ldgrmxinstrc", 0);
+        cfg.txMaxInstructions =
+            parseOptionalParamOrDefault<uint64_t>(map, "txmxinstrc", 0);
+        cfg.txMemoryLimit =
+            parseOptionalParamOrDefault<uint64_t>(map, "txmemlim", 0);
+        cfg.ledgerMaxReadLedgerEntries =
+            parseOptionalParamOrDefault<uint32_t>(map, "ldgrmxrdntry", 0);
+        cfg.ledgerMaxReadBytes =
+            parseOptionalParamOrDefault<uint32_t>(map, "ldgrmxrdbyt", 0);
+        cfg.ledgerMaxWriteLedgerEntries =
+            parseOptionalParamOrDefault<uint32_t>(map, "ldgrmxwrntry", 0);
+        cfg.ledgerMaxWriteBytes =
+            parseOptionalParamOrDefault<uint32_t>(map, "ldgrmxwrbyt", 0);
+        cfg.ledgerMaxTxCount =
+            parseOptionalParamOrDefault<uint32_t>(map, "ldgrmxtxcnt", 0);
+        cfg.txMaxReadLedgerEntries =
+            parseOptionalParamOrDefault<uint32_t>(map, "txmxrdntry", 0);
+        cfg.txMaxReadBytes =
+            parseOptionalParamOrDefault<uint32_t>(map, "txmxrdbyt", 0);
+        cfg.txMaxWriteLedgerEntries =
+            parseOptionalParamOrDefault<uint32_t>(map, "txmxwrntry", 0);
+        cfg.txMaxWriteBytes =
+            parseOptionalParamOrDefault<uint32_t>(map, "txmxwrbyt", 0);
+        cfg.txMaxContractEventsSizeBytes =
+            parseOptionalParamOrDefault<uint32_t>(map, "txmxevntsz", 0);
+        cfg.ledgerMaxTransactionsSizeBytes =
+            parseOptionalParamOrDefault<uint32_t>(map, "ldgrmxtxsz", 0);
+        cfg.txMaxSizeBytes = parseOptionalParamOrDefault<uint32_t>(
+            map, "txmxsz", cfg.ledgerMaxTransactionsSizeBytes);
+        cfg.bucketListSizeWindowSampleSize =
+            parseOptionalParamOrDefault<uint32_t>(map, "wndowsz", 0);
+        cfg.evictionScanSize = parseOptionalParamOrDefault<uint64_t>(
+            map, "evctsz", cfg.bucketListSizeWindowSampleSize);
+        cfg.startingEvictionScanLevel =
+            parseOptionalParamOrDefault<uint32_t>(map, "evctlvl", 0);
 
         if (cfg.maxGeneratedFeeRate)
         {
--- a/src/main/CommandHandler.cpp
+++ b/src/main/CommandHandler.cpp
@@ -1175,6 +1175,9 @@ CommandHandler::generateLoad(std::string const& params, std::string& retStr)
             parseOptionalParamOrDefault<uint32_t>(map, "dextxpercent", 0);
 
         // Only for SOROBAN_INVOKE mode, resource consumption bounds
+        cfg.nInstances =
+            parseOptionalParamOrDefault<uint32_t>(map, "instances", 0);
+        cfg.nWasms = parseOptionalParamOrDefault<uint32_t>(map, "wasms", 0);
         cfg.nDataEntriesLow =
             parseOptionalParamOrDefault<uint32_t>(map, "dataentrieslow", 0);
         cfg.nDataEntriesHigh =
--- a/src/simulation/LoadGenerator.cpp
+++ b/src/simulation/LoadGenerator.cpp
@@ -224,20 +224,12 @@ LoadGenerator::checkPendingTxs(GeneratedLoadConfig const& cfg)
             {
                 releaseAssert(cfg.sorobanPhase == SorobanPhase::UPLOAD);
                 releaseAssert(lkVec.size() == 1);
-                // TODO: Uncomment this when we have unique WASM
-                // releaseAssert(mIncompleteContractInstances.find(lk) ==
-                //               mIncompleteContractInstances.end());
-                // ContractInstance instance;
-                // instance.contractKeys.emplace_back(lk);
-                // mIncompleteContractInstances.emplace(id, instance);
                 releaseAssert(mCodeKey == std::nullopt);
                 mCodeKey = firstLk;
             }
             else
             {
                 releaseAssert(firstLk.type() == CONTRACT_DATA);
-
-                // TODO: Consolidate logic when we have unique WASM
                 if (firstLk.contractData().key.type() ==
                     SCV_LEDGER_KEY_CONTRACT_INSTANCE)
                 {
@@ -254,8 +246,7 @@ LoadGenerator::checkPendingTxs(GeneratedLoadConfig const& cfg)
                 else
                 {
                     releaseAssert(cfg.sorobanPhase == SorobanPhase::STORAGE);
-                    releaseAssert(firstLk.contractData().key.type() ==
-                                  SCV_SYMBOL);
+                    releaseAssert(firstLk.contractData().key.type() == SCV_U32);
                     auto& instance = mIncompleteContractInstances.at(id);
                     // Instance should already be populated with WASM and
                     // instance keys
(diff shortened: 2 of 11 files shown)


src/simulation/LoadGenerator.cpp
```diff
@@ -244,20 +372,56 @@ LoadGenerator::scheduleLoadGeneration(GeneratedLoadConfig cfg)
...
     {
-        errorMsg = "Soroban mode requires protocol version 20 or higher";
+        errorMsg = "Soroban modes require protocol version 20 or higher";
+    }
+
+    if (cfg.mode == LoadGenMode::SOROBAN_INVOKE)
+    {
+        LedgerTxn ltx(mApp.getLedgerTxnRoot());
+        auto& sorobanCfg = mApp.getLedgerManager().getSorobanNetworkConfig(ltx);
+        if (cfg.nDataEntries == 0)
```
dmkozh, line 389: These checks don't make much sense even with the current way to setup loadgen - why shouldn't I be able to skip the storage phase completely (and with suggested refactoring it would make even less sense).

dmkozh (commented):
I think we need more loadgen sophisticated configuration than what we have here. I think I wrote up about that in the supercluster/loadgen issues. Here are some ideas:

- `do_work` and `write` functions in the loadgen contract should be combined into one (sorry, I kind of forgot about that during the review)
- Loadgen configuration should include parameters for both CPU (`do_work`) and IO loads
- Loadgen parameters should define ranges, not the fixed values. Every tx we invoke should consume CPU in `min_cpu_work..max_cpu_work` range (where `cpu_work` is defined as e.g. a tuple of `(guest_cycles, host_cycles)`. It should *also* consume IO in the range `(min_entries, min_size)..(max_entries, max_size)`. We probably can start with uniform distributions here, so I guess we might also want to add normal distribution going forward.
- The same approach should apply goes to the tx size padding - we need a range for min/max tx padding size

I think this should be more or less enough to generate pretty complex missions. We could be even more sophisticated (e.g. add spikes in one certain resource dimension), but that is probably not critical initially.

Please feel free to postpone that to a followup PR. Though it seems like you've already got most of the bulding blocks, so leaving the decision on this up to your judgement.

Also of course please let me know if you have ideas for simpler configuration that would allow to build various mission types. The key requirement really is to be able to generate traffic that exercises all the resource dimensions at once.

src/testdata/ledger-close-meta-v1-protocol-20-soroban.json
```diff
@@ -6,27 +6,27 @@
                 "v": 0
             },
             "ledgerHeader": {
-                "hash": "86ccb8fdd5c17f3f10b78f9ea23223cc40f27d73aeac03a7cabeeaf3ef623c53",
+                "hash": "52d0e06821b3e9af8032d0058971828ad327402720b6d85d1a3423a1abad49e3",
```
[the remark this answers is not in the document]
SirTyson: As discussed in slack, this is expected and is the result of a change in Wasm test binary size.

src/simulation/test/LoadGeneratorTests.cpp
```diff
@@ -90,6 +90,166 @@ TEST_CASE("generate load with unique accounts", "[loadgen]")
...
+    simulation->crankUntil(
+        [&]() { return simulation->haveAllExternalized(3, 1); },
+        2 * Herder::EXP_LEDGER_TIMESPAN_SECONDS, false);
+
+    auto nodes = simulation->getNodes();
+    for (auto node : nodes)
+    {
+        overrideSorobanNetworkConfigForTest(*node);
+        modifySorobanNetworkConfig(*node, [&](SorobanNetworkConfig& cfg) {
+            // Entries should never expire
```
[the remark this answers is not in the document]
SirTyson: Yeah I think I can lower these numbers, I just picked a number higher than I thought we would ever simulate to make sure my hard coded fees were sufficient. I think we still need to up the minimum lifetime. We bump the contract data in the contract, but not the instance or Wasm, so those could still expire. I need to redo the loadgen contract anyway, so I think I'm going to remove the rent bump and just have a high minimum lifetime for loadgen so we don't have to worry about restoration and rent bumps. I can also remove the memory override. I think we still need the ledger and write related settings though just to make sure that no otherwise valid TXs get dropped during the setup phase.

src/simulation/LoadGenerator.h
```diff
@@ -160,10 +199,54 @@ class LoadGenerator
 
     uint32_t mWaitTillCompleteForLedgers{0};
 
-    void reset();
+    // The following maps store state information for the SOROBAN_INVOKE mode.
+    // For all other loadgen modes, there are no dependent TXs after creating
+    // source accounts. However, for SOROBAN_INVOKE, we must deploy WASM blobs,
+    // create contract instances, and create storage entries before invoking the
```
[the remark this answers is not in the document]
SirTyson: If we just have one Wasm entry, we only load the Wasm once, even though every TX is charged read resources for the entry. For the worst case scenario, every TX would be loading a different Wasm blob such that we have no cache hits. My thinking is that each Wasm blob has the same source code, but we add padding so that the size and hash varies, creating more IO work. This might be overkill, we could also just load more contract data and probably achieve something similar...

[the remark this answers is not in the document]
dmkozh: Sorry, I probably was unclear. The confusion point to me is why storage entry setup is mentioned alongside Wasm/contract instances. It should be possible to invoke contract without setting up any additional entries, right?

src/simulation/LoadGenerator.h
```diff
@@ -32,14 +32,23 @@ enum class LoadGenMode
     PRETEND,
     // Mix of payments and DEX-related transactions.
     MIXED_TXS,
-    SOROBAN
+    // Deploy random WASM blobs, for overlay/herder testing
```
marta-lokhova, line 35: need to also update commands.md since this impacts loadgen HTTP endpoint

src/simulation/LoadGenerator.cpp
```diff
@@ -668,6 +787,29 @@ LoadGenerator::getNextAvailableAccount()
     return sourceAccountId;
 }
 
+// Returns account ID for the next available account that has not already set up
+// a contract instance
+uint64_t
+LoadGenerator::getNextAvailableAccountForSorobanSetup()
```
marta-lokhova, line 793: what is different about contract instances that we can't use `getNextAvailableAccount` here? I might not have enough soroban background, but why do we have 1-1 mapping of accounts vs contract instances?

src/simulation/LoadGenerator.cpp
```diff
@@ -244,13 +291,46 @@ LoadGenerator::scheduleLoadGeneration(GeneratedLoadConfig cfg)
...
                                 SOROBAN_PROTOCOL_VERSION))
     {
-        errorMsg = "Soroban mode requires protocol version 20 or higher";
+        errorMsg = "Soroban modes require protocol version 20 or higher";
+    }
+
+    if (cfg.mode == LoadGenMode::SOROBAN_INVOKE)
+    {
+        auto& sorobanCfg = mApp.getLedgerManager().getSorobanNetworkConfig();
+        if (mContractInstances.size() < cfg.nAccounts)
```
marta-lokhova, line 308: I'm not sure this check makes sense: mContractInstances is populated during loadgen, whereas nAccounts is const config, so mContractInstances should depend on nAccounts to begin with

src/simulation/LoadGenerator.cpp
```diff
@@ -244,13 +291,46 @@ LoadGenerator::scheduleLoadGeneration(GeneratedLoadConfig cfg)
...
+        {
+            errorMsg = "must run SOROBAN_INVOKE_SETUP with at least nAccounts";
+        }
+        else if (cfg.nDataEntriesHigh > sorobanCfg.mTxMaxWriteLedgerEntries)
+        {
+            errorMsg = "nDataEntriesHigh larger than max write ledger entries";
+        }
+        // Wasm + instance + data entry reads
+        else if (cfg.nDataEntriesHigh + 2 > sorobanCfg.mTxMaxReadLedgerEntries)
```
marta-lokhova, line 317: what are the magic numbers in these checks (2, 2'000)?

src/simulation/LoadGenerator.cpp
SirTyson: The issue is that `mContractInstances` is populated during `SOROBAN_INVOKE_SETUP`, but used during `SOROBAN_INVOKE`. It's possible that these get run with different configurations. This is an error check for situations where `nAccounts == 10` during setup loadgen but `nAccounts == 20` during the invoke loadgen.

src/simulation/LoadGenerator.cpp
SirTyson: The 2 is the Wasm + instance entry that we need to read, and the 2'000 is a rough estimate of wasm size. The comments are supposed to explain these magic numbers, but I can make them a little more clear if it's still not apparent what they are.

src/simulation/LoadGenerator.cpp
SirTyson: A 1-1 mapping has a few advantages. The way we charge resource fees, every TX pays for every resource it uses. This means that if 2 TXs access the same entry, they both get charged for the read even though core only does one load. By having each account have it's own instance, you guarantee that each TX will have no shared state such that we have no cached data. Additionally, each contract instance has it's own key space, so having unique contract instance greatly simplifies IO loads since we can use the same SCVal keys for every TX but they still refer to separate `LedgerKeys` such that they require separate loads. I could just have a pool of available contract instances instead of having a one-to-one mapping, but the setup overhead is not very significant (it just takes one TX per instance), and with a one-to-one mapping we can reuse the existing `getNextAvailableAccount` functionality without having to reimplement something like `getNextAvailableInstance`.

SirTyson: During setup we want each account to submit a single deploy instance TX. If we use `getNextAvailableAccount`, we will get an available account that may or may not have already deployed an instance. `getNextAvailableAccountForSorobanSetup` is just an extension of `getNextAvailableAccount` that makes sure the returned account does not yet have a contract instance deployed.

src/simulation/test/LoadGeneratorTests.cpp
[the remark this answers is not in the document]
dmkozh: Yeah, it's fine to override whatever we need, I just wouldn't change more settings than necessary.

src/simulation/LoadGenerator.h
dmkozh, line 35: nit: Wasm

src/simulation/LoadGenerator.cpp
```diff
@@ -866,11 +1025,224 @@ LoadGenerator::manageOfferTransaction(
...
+{
+    auto account = findAccount(accountId, ledgerNum);
+    auto instanceIter = mContractInstances.find(accountId);
+    releaseAssert(instanceIter != mContractInstances.end());
+    auto const& instance = instanceIter->second;
+
+    auto const& networkCfg = mApp.getLedgerManager().getSorobanNetworkConfig();
+
+    // Approximate instruction measurements from loadgen contract. While the
+    // guest and host cycle counts are accurate, it very difficult to estimate
```
dmkozh, line 1141: I wouldn't spend much time on figuring this out, but these costs shouldn't be that difficult to figure out exactly, especially for the host calls. We do need to be vary of Wasm optimization for the guest cycles; but I'd hope we'll be able to freeze the loadgen contract soon. In any case, I think if cost per cycle has the correct upper bound, then we should never actually run out of resources at execution time. I suppose the tricky part is the storage cost, which depends on the inputs. 
So for now I would probably just rephrase this comment to explain where the cost discrepancies come from (IIUC currently these are just storage workload costs).

src/simulation/LoadGenerator.cpp
```diff
@@ -866,11 +1025,224 @@ LoadGenerator::manageOfferTransaction(
...
+    for (uint32_t i = 0; i < numEntries; ++i)
+    {
+        auto lk = contractDataKey(instance.contractID, makeU32(i),
+                                  ContractDataDurability::PERSISTENT);
+        resources.footprint.readWrite.emplace_back(lk);
+    }
+
+    // Use 2'000 as an estimate for Wasm size
+    uint32_t maxKiloBytesPerEntry =
+        (networkCfg.mTxMaxReadBytes - 2'000) / numEntries / 1024;
```
dmkozh, line 1184: nit: It should actually be easy to just get the actual Wasm size (which is just the length of the Wasm bytes vector).

src/simulation/LoadGenerator.cpp
```diff
@@ -866,11 +1025,224 @@ LoadGenerator::manageOfferTransaction(
...
+        networkCfg.mTxMaxInstructions - baseInstructionCount;
+
+    uint64_t guestCyclesMax = maxInstructions / instructionsPerGuestCycle;
+    guestCyclesMax = std::min(cfg.guestCyclesHigh, guestCyclesMax);
+    uint64_t guestCyclesMin = std::min(cfg.guestCyclesLow, guestCyclesMax);
+    uint64_t guestCycles =
+        rand_uniform<uint64_t>(guestCyclesMin, guestCyclesMax);
+
+    maxInstructions -= guestCycles * instructionsPerGuestCycle;
+    uint64_t hostCyclesMax = maxInstructions / instructionsPerHostCycle;
```
dmkozh, line 1161: It's a bit hard to argue about the distribution properties here (I can't say much about the instructions distribution while looking at this). I believe a simpler option would be to generate the 'target' number of instructions first, then split randomly between guest and host (i.e. generate guest cycles and dedicate the remainder to the host cycles). This way we'll be able to say that instructions are (roughly) uniformly distributed.

src/simulation/LoadGenerator.cpp
```diff
@@ -866,11 +1025,224 @@ LoadGenerator::manageOfferTransaction(
...
+    Operation op;
+    op.body.type(INVOKE_HOST_FUNCTION);
+    auto& ihf = op.body.invokeHostFunctionOp().hostFunction;
+    ihf.type(HOST_FUNCTION_TYPE_INVOKE_CONTRACT);
+    ihf.invokeContract().contractAddress = instance.contractID;
+    ihf.invokeContract().functionName = "do_work";
+    ihf.invokeContract().args = {guestCyclesU64, hostCyclesU64, numEntriesU32,
+                                 kiloBytesPerEntryU32};
+
+    // We don't have a good way of knowing how many bytes we will have to read,
```
dmkozh, line 1206: That's a bit surprising, is there no way to come up with some upper bound on the read size?

SirTyson: Well the upper bound is easy, it's the TX readBytes limit since that's what we base all our random values on. The issue is the current read values depend on the last invocation's write values, which are randomly selected. I thought about having some mapping of instance -> last invocation writeBytes, but the issue is there's no good way to maintain that mapping. We could update the writeBytes value when we submit the TX, but the TX may fail, leaving us with incorrect state. We'd have to check that the TX actually got applied to know the correct bytes value, and this would require that loadgen do disk lookups. Loadgen is already expensive as it is, so I think leaving these values as network max is fine.

dmkozh: Ok, this probably is good enough initially. We could be more precise though if we set `kiloBytesPerDataEntryHigh * nDataEntriesHigh + <Wasm+instance size>` as the limit (this way it's still possible to have load that is not bounded by the read bytes when we skip the IO).
Going forward, we probably could come up with a way to improve this, but we definitely should get something working first.

src/simulation/LoadGenerator.cpp
```diff
@@ -562,6 +655,86 @@ LoadGenerator::generateLoad(GeneratedLoadConfig cfg)
     scheduleLoadGeneration(cfg);
 }
 
+uint32_t
+LoadGenerator::submitSorobanPrepareInvokeTX(uint32_t nAccounts,
```
marta-lokhova, line 659: I'm not sure this works as we don't seem to be handling tx failure modes here: any transaction successfully submitted to the tx queue via loadgen isn't guaranteed to actually make it to block application. Even at block application, it may still fail. From what I see, if this happens loadgen will just get stuck. In other modes, we implement a timeout to get it unstuck (e.g. `waitTillComplete`). It seems the submission pattern that would handle this properly is something like: submit upload wasms and wait until they all show up in DB (timeout and fail if they don't), submit deploy instances, and wait until they are reflected in the DB (with the same timeout strategy), and only then return success in this setup mode

SirTyson:
Given the time constraints for getting a supercluster soroban network config upgrade mission working, the config upgrade loadgen modes have been added to this PR.

src/main/CommandHandler.cpp
```diff
@@ -1174,6 +1174,68 @@ CommandHandler::generateLoad(std::string const& params, std::string& retStr)
...
+        cfg.txMaxInstructions =
+            parseOptionalParamOrDefault<uint64_t>(map, "txmxinstrc", 0);
+        cfg.txMemoryLimit =
+            parseOptionalParamOrDefault<uint64_t>(map, "txmemlim", 0);
+        cfg.ledgerMaxReadLedgerEntries =
+            parseOptionalParamOrDefault<uint32_t>(map, "ldgrmxrdntry", 0);
+        cfg.ledgerMaxReadBytes =
+            parseOptionalParamOrDefault<uint32_t>(map, "ldgrmxrdbyt", 0);
+        cfg.ledgerMaxWriteLedgerEntries =
+            parseOptionalParamOrDefault<uint32_t>(map, "ldgrmxwrntry", 0);
```
dmkozh, line 1213: nit:  This is really hard to read... Is it really necessary to get rid of most of the vowels here? Why not just use underscore and proper long names?

src/simulation/LoadGenerator.cpp
dmkozh: Can we update this as suggested above?

src/simulation/LoadGenerator.h
```diff
@@ -67,6 +119,60 @@ struct GeneratedLoadConfig
     bool skipLowFeeTxs = false;
     // Percentage (from 0 to 100) of DEX transactions
     uint32_t dexTxPercent = 0;
+
+    // The following ranges are only used for SOROBAN_INVOKE txs:
```
dmkozh, line 123: I think the list of the parameters and possible combinations is growing unwieldy. For the sake of time we probably can keep this as is for now, but as a followup it would be nice to introduce a few structs that group the related parameters together (invoke parameters, soroban upgrades, soroban setup parameters, maybe something else).

dmkozh (commented):
Looks fine overall, just a couple minor things.

src/main/CommandHandler.cpp
SirTyson: Unfortunately this is necessary. For supercluster runs, we set all loadgen parameters via a single HTTP string. With the addition of these config parameters, we're running up against the HTTP string limits. If these names are longer, we'd have to break network upgrades into several small upgrades, which slows down runs and increases the chance of failure, since loadgen in supercluster is a little flaky to begin with.

dmkozh: Well, maybe it's time to switch to POST request :) That's fine for now then.

src/simulation/LoadGenerator.cpp
SirTyson: Sorry, missed this comment.

dmkozh (approved):

src/simulation/LoadGenerator.cpp
```diff
@@ -999,37 +1677,54 @@ LoadGenerator::checkAccountSynced(Application& app, bool isCreate)
...
-    else
+    // If we have an inconsistency, reset the timer and wait for another ledger
+    else if (!inconsistencies.empty() || !sorobanInconsistencies.empty())
     {
         if (++mWaitTillCompleteForLedgers >= TIMEOUT_NUM_LEDGERS)
         {
             CLOG_INFO(LoadGen, "Load generation failed.");
             mLoadgenFail.Mark();
-            reset();
+            reset(/*resetSoroban=*/!sorobanInconsistencies.empty());
```
marta-lokhova, line 1706: nit: it seems that resetting soroban state doesn't actually belong to `reset`, which resets all state between runs. probably better to have a dedicated function for soroban state reset, this way you can avoid duplication on line 322 as well.

src/simulation/LoadGenerator.h
marta-lokhova: Agreed, loadgen has increased in complexity quite a bit. We should probably open an issue so we don't forget to clean this up. It also wouldn't hurt to make sure we don't allow invalid combinations of parameters (e.g. it doesn't make sense to accept txs/accounts/etc in soroban upgrade mode)

marta-lokhova (approved):

src/main/CommandHandler.cpp
```diff
@@ -1187,11 +1249,22 @@ CommandHandler::generateLoad(std::string const& params, std::string& retStr)
 
         uint32_t numItems = isCreate ? cfg.nAccounts : cfg.nTxs;
         std::string itemType = isCreate ? "accounts" : "txs";
-
-        retStr +=
+        Json::Value res;
+        res["status"] =
```
marta-lokhova, line 1253: please update reporting here and in the logs to give info about the new modes

dmkozh:
r+ 752c6b1

marta-lokhova:
r+ 1217ebf2f5fc63e48ab448a1b3d649ca9139ba01

PR #4039, opened 2023-11-17
code review updates part 2

MonsieurNicolas:
See individual commit descriptions for details

### DIFF
--- a/src/main/PersistentState.cpp
+++ b/src/main/PersistentState.cpp
@@ -37,7 +37,7 @@ PersistentState::PersistentState(Application& app) : mApp(app)
 void
 PersistentState::deleteTxSets(std::unordered_set<Hash> hashesToDelete)
 {
-    // No need for soci::transaction here; try to delete as much as we can
+    soci::transaction tx(mApp.getDatabase().getSession());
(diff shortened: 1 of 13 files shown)


src/main/PersistentState.cpp
```diff
@@ -37,7 +37,7 @@ PersistentState::PersistentState(Application& app) : mApp(app)
 void
 PersistentState::deleteTxSets(std::unordered_set<Hash> hashesToDelete)
 {
-    // No need for soci::transaction here; try to delete as much as we can
+    soci::transaction tx(mApp.getDatabase().getSession());
```
marta-lokhova, line 40: curious, did you observe perf degradation in a profiler?

MonsieurNicolas: I didn't look for this particular one as this tends to be quite unpredictable. without this you have one sql transaction per txset. even on sqlite this means one fsync for each. So this new version is at least as fast as the original one and we don't have to maintain a comment.

marta-lokhova (approved):

marta-lokhova:
r+ c16d5b0ce57f7a902078204c3c01800c6913e856

PR #4047, opened 2023-11-20
annotate core with more tracy zones for profiling

marta-lokhova:
Looks like we haven't done a pass at tracy zones for some time. This change should help with perf analysis

### DIFF
--- a/src/bucket/Bucket.cpp
+++ b/src/bucket/Bucket.cpp
@@ -190,6 +190,8 @@ void
 Bucket::loadKeys(std::set<LedgerKey, LedgerEntryIdCmp>& keys,
                  std::vector<LedgerEntry>& result)
 {
+    ZoneScoped;
+
     auto currKeyIt = keys.begin();
     auto const& index = getIndex();
     auto indexIter = index.begin();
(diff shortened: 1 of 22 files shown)


src/bucket/Bucket.cpp
```diff
@@ -859,6 +863,7 @@ Bucket::scanForEviction(AbstractLedgerTxn& ltx, EvictionIterator& iter,
     {
         if (be.type() == INITENTRY || be.type() == LIVEENTRY)
         {
+            ZoneNamedN(maybeEvict, "maybe evict entry", true);
```
SirTyson, line 866: This zone isn't very meaningful here, since most entries we scan will not be temporary and will require almost no work. I think if we rename this to `should evict` and move this inside the lambda it would be more meaningful.

SirTyson (approved):
One small suggestion, thanks for doing this! A very helpful tedious change :)

src/bucket/Bucket.cpp
marta-lokhova: sounds good. I just moved it inside the `if (isTemporaryEntry(le.data))` block

dmkozh:
r+ 348e4f8

marta-lokhova:
@latobarita: retry

PR #4057, opened 2023-11-22
No-op refactor of TxSetFrame.

### LINKED ISSUE #4017: Cleanup: rename `TxSetFrame` to `TxSetXDRFrame`
This is a followup refactor for https://github.com/stellar/stellar-core/pull/3977: we can now rename `TxSetFrame` for clarity to something like ``TxSetXDRFrame` (as it only wraps around the XDR without interpretation). Also we might move some factory functions out of `TxSetFrame` (it's a bit weird that `makeFromTransactions` also returns `ApplicableTxSetFrame` while being a static function of `TxSetFrame`).

dmkozh:
# Description

Resolves #4017 

- Rename TxSetFrame to TxSetXDRFrame
- Move the factory functions that create 2 tx set frames (XDR and applicable) out of the TxSetFrame class as they don't make much semantic sense as member functions
- Same goes for the type aliases and `TxPhase` enum - moved all to be standalone types

# Checklist
- [ ] Compiles
- [ ] Ran all tests

### DIFF
--- a/src/herder/TxSetFrame.h
+++ b/src/herder/TxSetFrame.h
@@ -20,103 +20,117 @@
 namespace stellar
 {
 class Application;
-class TxSetFrame;
+class TxSetXDRFrame;
 class ApplicableTxSetFrame;
-using TxSetFrameConstPtr = std::shared_ptr<TxSetFrame const>;
+using TxSetXDRFrameConstPtr = std::shared_ptr<TxSetXDRFrame const>;
 using ApplicableTxSetFrameConstPtr =
     std::unique_ptr<ApplicableTxSetFrame const>;
+
+enum class TxSetPhase
+{
+    CLASSIC,
+    SOROBAN,
+    PHASE_COUNT
+};
+
+using TxSetTransactions = std::vector<TransactionFrameBasePtr>;
+using TxSetPhaseTransactions = std::vector<TxSetTransactions>;
+
+std::string getTxSetPhaseName(TxSetPhase phase);
+
+// Creates a valid ApplicableTxSetFrame and corresponding TxSetXDRFrame
+// from the provided transactions.
+//
+// Not all the transactions will be included in the result: invalid
+// transactions are trimmed and optionally returned via `invalidTxs` and if
+// there are too many remaining transactions surge pricing is applied.
+// The result is guaranteed to pass `checkValid` check with the same
+// arguments as in this method, so additional validation is not needed.
+//
+// **Note**: the output `ApplicableTxSetFrame` will *not* contain the input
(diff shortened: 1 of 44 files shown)


src/herder/TxSetFrame.h
```diff
@@ -20,103 +20,117 @@
...
+// Not all the transactions will be included in the result: invalid
+// transactions are trimmed and optionally returned via `invalidTxs` and if
+// there are too many remaining transactions surge pricing is applied.
+// The result is guaranteed to pass `checkValid` check with the same
+// arguments as in this method, so additional validation is not needed.
+//
+// **Note**: the output `ApplicableTxSetFrame` will *not* contain the input
+// transaction pointers.
+std::pair<TxSetXDRFrameConstPtr, ApplicableTxSetFrameConstPtr>
+makeTxSetFromTransactions(
```
marta-lokhova, line 53: should these live in TxSetUtils? it seems more natural to have all the tx set helpers grouped together for cohesion and discoverability

dmkozh: I treat factory functions (which this kind of is) the same way as constructors, so it's more natural for me to have it together with the class definitions. It also would be a bit annoying to make these friends of the tx set frame classes (I'm not very happy with having friends at all, but moving them across files feels even more hacky).

marta-lokhova: well, ok. if you want to treat them as constructors, then it makes sense to stay consistent with the codebase, and have a static factory method that returns xdr tx set inside tx set frame class (like what makeFromWire factories are doing; otherwise we now have several creation patterns). This factory is only really used in nomination, and we throw away applicable set anyway. I guess we use it lot in tests, so it would be annoying to re-resolve created tx sets. So I suppose this is fine then

dmkozh: The reason I moved these away is that we create two frame types, so having it in either one is awkward. I really wouldn't be opposed to moving this elsewhere, but as I said, I just don't quite like having to friend a function from a separate header.

marta-lokhova: yeah, what I was saying is that we don't really use the applicable tx set returned outside of tests, so it doesn't have to create two frame types. but I don't feel strongly about the current setup, so I'm fine merging this as-is.

marta-lokhova:
r+ 3634ba8c1b895b3b383ecdfcef495511954ef219

src/herder/TxSetFrame.h
dmkozh: Thanks, I think we can polish this further when we have some time and desire.

PR #4060, opened 2023-11-28
Added test and enforcement for max ledger entry size writes

### LINKED ISSUE #4002: [Test] Cover ledger entry size reductions
While it is unlikely, we might need to reduce the maximum key or ledger entry size via a network upgrade. We need to cover that in the tests. Currently the expected behavior is that we allow reading, but not writing such entries. We need to cover:

- [ ] Reading the existing ledger key that is larger than the key size limit
- [ ] Reading the existing ledger entry larger than the entry size limit

SirTyson:
# Description

Resolves #4002

Adds tests for enforcing `maxContractDataEntrySizeBytes` and `maxContractDataKeySizeBytes`. Previously, we allowed reads that were beyond entry and key limits, but not writes. After discussing with @dmkozh, we decided to not allow both read and writes that exceeded entry or key limits. This change adds test cases as well as enforcement for reads that exceed the maximum contract data entry limit.

# Checklist
- [x] Compiles
- [x] Ran all tests


src/transactions/InvokeHostFunctionOpFrame.cpp
```diff
@@ -440,6 +441,20 @@ InvokeHostFunctionOpFrame::doApply(Application& app, AbstractLedgerTxn& ltx,
 
             metrics.noteReadEntry(isCodeKey(lk), keySize, entrySize);
 
+            if (isSorobanEntry(lk) &&
```
dmkozh, line 444: You should reuse `validateContractLedgerEntry` (you're missing the contract code case here). I also think we need to move it to a separate utility and use in all the operations.

src/transactions/TransactionUtils.cpp
```diff
@@ -1827,6 +1827,35 @@ getMinInclusionFee(TransactionFrameBase const& tx, LedgerHeader const& header,
...
+validateContractLedgerEntry(LedgerKey const& lk, size_t entrySize,
+                            SorobanNetworkConfig const& config,
+                            TransactionFrame& parentTx)
+{
+    // check contract code size limit
+    if (lk.type() == CONTRACT_CODE && config.maxContractSizeBytes() < entrySize)
+    {
+        parentTx.pushSimpleDiagnosticError(
+            SCE_BUDGET, SCEC_EXCEEDED_LIMIT,
+            "WASM size exceeds network config maximum contract size",
```
dmkozh, line 1840: Wasm

src/transactions/RestoreFootprintOpFrame.cpp
```diff
@@ -108,6 +108,12 @@ RestoreFootprintOpFrame::doApply(Application& app, AbstractLedgerTxn& ltx,
         // To maintain consistency with InvokeHostFunction, TTLEntry
         // writes come out of refundable fee, so only add entrySize
         metrics.mLedgerWriteByte += entrySize;
+        if (!validateContractLedgerEntry(lk, entrySize, sorobanConfig,
```
dmkozh, line 111: I guess if we're doing this check everywhere, then we should also add it to extend ttl operation.

dmkozh:
r+ d5a9465

PR #4061, opened 2023-11-28
Pin cargo deps more

graydon:
This does the remaining part of stellar/rs-soroban-env#1222:

- pins versions with '=' constraints in core
- adds CI and workspace plumbing for `cargo-deny` and `cargo-acl`
- fixes issues they found
- uses `--locked` everywhere we issue `cargo` commands (in `Makefile.am`)

Note that there is a small lockfile change here but it doesn't impact the baked-in env dependency trees, it only reduces some divergence between those trees and dependencies of `cargo-lock`, which is one of core's dependencies _outside_ the env trees (we assume it has no effect on the part we care about: bit-precise stability of the env code)

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -28,7 +28,7 @@ jobs:
     steps:
       - uses: actions/checkout@v3
       - uses: cackle-rs/cackle-action@387fe459d8c037a0e7fbcd3d868f1aaba389d057
-      - run: cargo acl -n run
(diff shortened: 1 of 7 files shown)


dmkozh:
r+ 59a29aa

graydon:
@latobarita: retry

graydon:
r+ 8d2934c

graydon:
r+ a8f3c79

PR #4064, opened 2023-11-29
Centralize txn size calculation 

### LINKED ISSUE #4023: centralize tx size computation across tx subsytem and overlay
Multiple subsystems in core rely on transaction size computation. For these systems to interact correctly, they must compute tx size in the same way. We've already encountered bugs related to the fact that core has several places where the same computation is done, e.g. in overlay:
https://github.com/stellar/stellar-core/blob/0448bf0e314d1afb68f11924802534fe03989d69/src/overlay/FlowControlCapacity.cpp#L223
and in tx the subsystem:
https://github.com/stellar/stellar-core/blob/0448bf0e314d1afb68f11924802534fe03989d69/src/transactions/TransactionFrame.cpp#L227

We should centralize this logic in one util function to prevent footguns.

ThomasBrady:
…from txn and overlay systems

# Description

Resolves #4023 

# Checklist
- [ ] Compiles
- [ ] Ran all tests

### DIFF
--- a/src/transactions/TransactionFrame.cpp
+++ b/src/transactions/TransactionFrame.cpp
@@ -236,7 +236,7 @@ TransactionFrame::getNumOperations() const
 Resource
 TransactionFrame::getResources(bool useByteLimitInClassic) const
 {
-    int64_t txSize = xdr::xdr_size(mEnvelope);
+    auto txSize = static_cast<int64_t>(this->getSize());
     if (isSoroban())
     {
         auto r = sorobanResources();
@@ -751,7 +751,7 @@ TransactionFrame::validateSorobanResources(SorobanNetworkConfig const& config,
             return false;
         }
     }
-    auto txSize = xdr::xdr_size(mEnvelope);
+    auto txSize = this->getSize();
     if (txSize > config.txMaxSizeBytes())
     {
         pushSimpleDiagnosticError(
@@ -1899,5 +1899,12 @@ xdr::xvector<DiagnosticEvent> const&
 TransactionFrame::getDiagnosticEvents() const
 {
     return mSorobanExtension->mDiagnosticEvents;
+}
+
+uint32_t
+TransactionFrame::getSize() const
+{
+    ZoneScoped;
+    return static_cast<uint32_t>(xdr::xdr_size(mEnvelope));
 }
 } // namespace stellar
--- a/src/transactions/TransactionFrame.h
+++ b/src/transactions/TransactionFrame.h
@@ -47,6 +47,9 @@ using TransactionFramePtr = std::shared_ptr<TransactionFrame>;
 
 class TransactionFrame : public TransactionFrameBase
 {
+  private:
+    uint32_t getSize() const;
+
   protected:
     TransactionEnvelope mEnvelope;
     TransactionResult mResult;


src/transactions/TransactionFrame.cpp
```diff
@@ -225,9 +225,10 @@ TransactionFrame::getNumOperations() const
 Resource
 TransactionFrame::getResources(bool useByteLimitInClassic) const
 {
-    int64_t txSize = xdr::xdr_size(mEnvelope);
     if (isSoroban())
     {
+        auto txSize = static_cast<int64_t>(
```
SirTyson, line 230: Why did this move inside the if? Seems cleaner to get the size just once like before.

src/transactions/TransactionFrame.cpp
```diff
@@ -244,6 +245,8 @@ TransactionFrame::getResources(bool useByteLimitInClassic) const
     }
     else if (useByteLimitInClassic)
     {
+        auto txSize = static_cast<int64_t>(
```
SirTyson, line 248: It looks like you also need to use the new function on line 745.

src/transactions/TransactionUtils.cpp
```diff
@@ -1951,4 +1951,11 @@ makeClaimAtom(uint32_t ledgerVersion, AccountID const& accountID,
     }
     return atom;
 }
+
+uint64_t
+getTransactionEnvelopeSize(TransactionEnvelope const& envelope)
+{
+    return static_cast<uint64_t>(xdr::xdr_size(envelope));
```
SirTyson, line 1958: Why cast and return this as a `uint64_t`? It looks like you use this function as the original `int64_t` more often.

SirTyson (changes requested):
Thanks for this cleanup! I think there are a few more places in the code where we should be using the new `getTransactionEnvelopeSize` instead of the old calculation. I see some places in `src/herder/test/HerderTests.cpp`, `src/transactions/test/InvokeHostFunctionTests.cpp`, and `src/transactions/test/SorobanTxTestUtils.cpp`. Any place where we have `xdr::xdr_size(txEnvelope)` should probably use the new function for consistency.

src/overlay/FlowControlCapacity.cpp
```diff
@@ -220,8 +221,15 @@ FlowControlCapacity::msgBodySize(StellarMessage const& msg,
     if (remoteVersion >= Peer::FIRST_VERSION_UPDATED_FLOW_CONTROL_ACCOUNTING &&
         localVersion >= Peer::FIRST_VERSION_UPDATED_FLOW_CONTROL_ACCOUNTING)
     {
-        return static_cast<uint64_t>(xdr::xdr_size(msg) -
-                                     xdr::xdr_size(msg.type()));
+        if (msg.type() == TRANSACTION)
+        {
+            return stellar::getTransactionEnvelopeSize(msg.transaction());
```
dmkozh, line 226: The goal of the corresponding issue is to have consistency in how we measure tx envelope sizes at different layers. However, computations are still inconsistent (and remain inconsistent): for the purpose of the Soroban resource accounting we use the size of the 'regular' transaction envelope. At the overlay layer we might still deal with fee bump envelopes, for which we still use the inner envelope size for the resource accounting.
IIUC to remove the inconsistency, the helper should correctly handle the fee bump envelopes and use the inner envelope size only. I'm not 100% confident though whether this is the desired behavior at the overlay layer. That also would need the overlay version bump.
I think we need some input from @marta-lokhova as well.

src/transactions/TransactionFrame.cpp
ThomasBrady: We don't use the size in the final `else` case (non-soroban, non-useByteLimitInClassic), so I moved the computation into the cases where we are using it. But I agree its cleaner to only have the computation in one place.

src/transactions/TransactionUtils.cpp
ThomasBrady: `xdr_size` returns a `size_t` (unsigned), so I cast this to `uint64_t`. IMO generally sizes should be unsigned, and it seems thats how things are in the codebase (e.g. `FlowControlCapacity::msgBodySize` returns `uint64_t`), however the `Resource` class uses `int64_t` to store resource values (but afaik the inputs it stores nearly always unsigned and cast when populating it). I'd rather keep it unsigned here rather than cast unsigned to signed here and back to unsigned elsewhere.

src/overlay/FlowControlCapacity.cpp
marta-lokhova: thanks @dmkozh, you're right; since fee bumps don't account for the fee-bump envelope itself for resource calculation, overlay needs to support it separately. Special-casing fee bumps here is an option, but it doesn't seem super safe. As we discussed offline, it's probably safer to just add a big enough buffer to overlay limits to ensure that any unaccounted message parts still fit within overlay limits. In this case, we can just go back to using size of the whole message in overlay.

src/transactions/FeeBumpTransactionFrame.cpp
```diff
@@ -551,4 +551,10 @@ FeeBumpTransactionFrame::toStellarMessage() const
     msg.transaction() = mEnvelope;
     return msg;
 }
+
+uint64_t
+FeeBumpTransactionFrame::getSize() const
+{
+    return static_cast<uint64_t>(xdr::xdr_size(mEnvelope));
```
ThomasBrady, line 558: I'm unsure if this is the correct approach for `FeeBumpTransactionFrame`

src/transactions/TransactionFrameBase.h
```diff
@@ -94,5 +94,6 @@ class TransactionFrameBase
     getDiagnosticEvents() const = 0;
     virtual int64 declaredSorobanResourceFee() const = 0;
     virtual bool XDRProvidesValidFee() const = 0;
+    virtual uint64_t getSize() const = 0;
```
dmkozh, line 97: I don't think this needs to be an interface method, you're only using it for the regular tx frame (so it can be just a private method there). Note, that if it's going to be used by the external clients, the function should be renamed to something like `getEnvelopeSizeForSoroban` or something like that  and return the size of the inner envelope (not the whole envelope). But again, I'm not sure if that's necessary at all.

ThomasBrady: I agree this doesn't need to be exposed like this, previously I had added 
`uint64_t getTransactionEnvelopeSize(TransactionEnvelope const& envelope)` to TransactionUtils, but after the flow control accounting simplification, I moved the functionality `TransactionFrameBase::getSize`. Either this needs to be a private function as you said or I need to revert the last couple commits and have 
`TransactionUtils::getTransactionEnvelopeSize`. Depends on what level we want to "centralize" the size calculation, thoughts?

dmkozh: Well, I'd say this should be dictated by the users of this. Maybe there is no longer a significant benefit from this. If the only remaining user is tx frame itself (it seems like it is), then this should be just a private helper. I quickly looked through the code and only found the usage in tests (besides tx frame) - we could make the helper public for tests, but that would require a downcast from tx frame base to tx frame, so I don't think it's worth the hassle.

src/transactions/FeeBumpTransactionFrame.cpp
marta-lokhova: this is unused in FeeBumpTransactionFrame, right? I agree with Dima's suggestion that we should just extract size calculation as a private method of TransactionFrame.

src/transactions/TransactionFrameBase.h
ThomasBrady: Cool, I made `getSize` a private function of `TransactionFrame` and removed it from `TransactionFrameBase`.

src/transactions/TransactionFrame.cpp
marta-lokhova: looks like txSize is still calculated twice

ThomasBrady: Reverted.

marta-lokhova:
r+ b44f884ab4e5d5ea046093da986b2abc4343ddee

PR #4077, opened 2023-12-02
Add medida metrics for cpu_insn, time excluding VM

jayz22:
# Description

Add medida metrics for cpu_insn, time excluding VM instantiation. So that we monitor their ratio.
Since cpu cost is dominated by VM, the current cpu/time ratio is masked by the VM calibration (which we have overestimated a bit heavily), which doesn't tell how effective our metering is in general . This new metrics will hopefully solve that.

# Checklist
- [ ] Compiles
- [ ] Ran all tests


graydon:
r+ 85654cf

PR #4080, opened 2023-12-04
Loadgen cleanup

### LINKED ISSUE #4063: Clean up Soroban Loadgen
- [x] Cleanup `GeneratedLoadConfig`
- [x] Better status logs for loadgen
- [x] Metrics for Soroban loads
- [x] Soroban state reset function
- [ ] Discrete distribution for soroban loadgen parameters instead of random

SirTyson:
# Description

Mostly resolves #4063.

Additionally, `SOROBAN_INVOKE` mode now takes the total amount of IO kilobytes as a parameter instead of kilobytes per data entry, making it possible to write tests that respect network config bounds while still having variance in the number of data entries written.

# Checklist
- [x] Compiles
- [x] Ran all tests

### DIFF
--- a/src/simulation/LoadGenerator.cpp
+++ b/src/simulation/LoadGenerator.cpp
@@ -366,8 +366,6 @@ LoadGenerator::scheduleLoadGeneration(GeneratedLoadConfig cfg)
 
     if (cfg.mode == LoadGenMode::SOROBAN_INVOKE)
     {
-        auto& sorobanNetworkCfg =
-            mApp.getLedgerManager().getSorobanNetworkConfig();
         auto const& sorobanLoadCfg = cfg.getSorobanConfig();
         auto const& invokeCfg = cfg.getSorobanInvokeConfig();
         releaseAssertOrThrow(sorobanLoadCfg.nInstances != 0);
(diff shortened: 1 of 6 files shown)


src/simulation/LoadGenerator.cpp
```diff
@@ -1208,33 +1303,31 @@ LoadGenerator::invokeSorobanLoadTransaction(uint32_t ledgerNum,
...
+        kiloBytesPerEntry =
+            (totalWriteBytes - mContactOverheadBytes) / numEntries / 1024;
 
-    maxKiloBytesPerEntry =
-        std::min(maxKiloBytesPerEntry, cfg.kiloBytesPerDataEntryHigh);
-    uint32_t minKiloBytesPerEntry =
-        std::min(maxKiloBytesPerEntry, cfg.kiloBytesPerDataEntryLow);
-    auto kiloBytesPerEntry =
-        rand_uniform<uint32_t>(minKiloBytesPerEntry, maxKiloBytesPerEntry);
+        // If numEntries > 0, we can't write a 0 byte entry
```
dmkozh, line 1325: Hmm, we actually can, can't we? It won't be 0 bytes, probably about 8 bytes to write an empty Bytes ScVal.

src/simulation/LoadGenerator.cpp
```diff
@@ -1250,30 +1343,26 @@ LoadGenerator::invokeSorobanLoadTransaction(uint32_t ledgerNum,
...
     // baseInstructionCount is a very rough estimate and may be a significant
     // underestimation based on the IO load used, so use max instructions
     resources.instructions = networkCfg.mTxMaxInstructions;
-    resources.readBytes = readBytesUpperBound;
-    resources.writeBytes = writeBytes;
+
+    // We don't have a good way of knowing how many bytes we will need to read
+    // since the previous invocation writes a random number of bytes, so use
+    // upper bound
+    resources.readBytes = invokeCfg.ioKiloBytesHigh * 1024;
```
dmkozh, line 1353: Why did you change this back to the max value?

src/simulation/LoadGenerator.cpp
SirTyson: Sorry, I didn't know this was valid behavior. Currently the loadgen contract panics if `size_kilo_bytes == 0 && num_write_entries != 0`. We can fix this later, but personally I don't think an env change and wasm recompile is worth fixing this right now.

src/simulation/LoadGenerator.cpp
SirTyson: This is the same value that we had before. Previously this was 
```
readBytes = (cfg.kiloBytesPerDataEntryHigh * 1024 + leOverhead) * numEntries + mContactOverheadBytes;
```
Since the config parameter now dictates total TX io size instead of the size of individual data entries, we can simplify it. Note that this is the LoadGen config max, not the network settings max. Unfortunately we have a lot of different config types floating around so it's a bit confusing.

src/simulation/LoadGenerator.cpp
dmkozh: Ah, ok. Empty bytes sequence is a valid value for a byte sequence, but I agree there is no need to update Wasm due to that. Also, it doesn't provide any significant value for coverage.

src/simulation/LoadGenerator.cpp
dmkozh: Ah, right, I've got confused by this a bit.

dmkozh:
r+ 113f5d9

marta-lokhova:
r+ 8b825c06c95240c58992ccbf5ff2ed6d02b30df2

PR #4083, opened 2023-12-05
Overlay: Do not initiate connections that will be immediately dropped

bboston7:
# Description

This change ensures that the overlay does not initiate outbound connections when it is shutting down (and therefore `OverlayManagerImpl::addOutboundConnection` will immediately drop the connection).

This change also adds a line to the logs indicating how to increase the number of outbound pending slots when
`OverlayManagerImpl::connectToImpl` fails due to insufficient slots.

# Checklist
- [x] Compiles
- [x] Ran all tests

### DIFF
--- a/src/overlay/OverlayManagerImpl.cpp
+++ b/src/overlay/OverlayManagerImpl.cpp
@@ -54,6 +54,30 @@ constexpr uint32_t INITIAL_FLOW_CONTROL_SEND_MORE_BATCH_SIZE_BYTES{100000};
 // longer than 2 seconds between re-issuing demands.
 constexpr std::chrono::seconds MAX_DELAY_DEMAND{2};
 
+bool
+OverlayManagerImpl::canAcceptOutboundPeer(PeerBareAddress const& address) const
+{
+    if (availableOutboundPendingSlots() <= 0)
+    {
+        CLOG_DEBUG(Overlay,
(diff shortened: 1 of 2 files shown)


src/overlay/OverlayManagerImpl.cpp
```diff
@@ -371,6 +371,16 @@ OverlayManagerImpl::connectToImpl(PeerBareAddress const& address,
                        "Peer rejected - all outbound pending connections "
                        "taken: {}",
                        address.toString());
+            CLOG_DEBUG(Overlay, "If you wish to allow for more pending "
```
marta-lokhova, line 374: I think it would be nice to extract the logging here and in `addOutboundConnection` into a separate function to avoid duplication

marta-lokhova (changes requested):
Thanks for the PR! I had a small cleanup suggestion

src/overlay/OverlayManagerImpl.cpp
bboston7: Good idea, I added a function `logPeerRejectionNoOutboundSlots` that deduplicates this.

bboston7: Ah, got it. I just made that change!

marta-lokhova:
r+ 2c55512d1dcc8ee3fc29ef2bd78842f7efb0fe82

PR #4086, opened 2023-12-07
Add BucketList size meter

### LINKED ISSUE #4085: Add metric for BucketList size
We currently have no metrics tracking `BucketList` size. This would be very useful, especially given that Soroban fees are dependent on BucketList size.

SirTyson:
# Description

Resolves #4085

Adds a meter to track BucketList size. 

# Checklist
- [x] Compiles
- [x] Ran all tests

### DIFF
--- a/docs/metrics.md
+++ b/docs/metrics.md
@@ -31,6 +31,7 @@ bucket.batch.objectsadded                 | meter     | number of objects added 
 bucket.memory.shared                      | counter   | number of buckets referenced (excluding publish queue)
 bucket.merge-time.level-<X>               | timer     | time to merge two buckets on level <X>
 bucket.snap.merge                         | timer     | time to merge two buckets
+bucketlist.size.bytes                     | counter   | total size of the BucketList in bytes
 bucketlistDB.bloom.lookups                | meter     | number of bloom filter lookups
 bucketlistDB.bloom.misses                 | meter     | number of bloom filter false positives
 bucketlistDB.query.loads                  | meter     | number of BucketListDB load queries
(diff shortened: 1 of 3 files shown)


docs/metrics.md
```diff
@@ -31,6 +31,7 @@ bucket.batch.objectsadded                 | meter     | number of objects added
 bucket.memory.shared                      | counter   | number of buckets referenced (excluding publish queue)
 bucket.merge-time.level-<X>               | timer     | time to merge two buckets on level <X>
 bucket.snap.merge                         | timer     | time to merge two buckets
+bucketlist.size.bytes                     | meter     | size of the BucketList in bytes, excluding future buckets
```
marta-lokhova, line 34: I would avoid mentioning future buckets to prevent confusion, since it's more of an implementation detail (future buckets aren't actually part of the bucketlist, just an optimization we do ahead of time)

src/bucket/BucketManagerImpl.cpp
```diff
@@ -828,6 +830,7 @@ BucketManagerImpl::addBatch(Application& app, uint32_t currLedger,
                                   deadEntries.size());
     mBucketList->addBatch(app, currLedger, currLedgerProtocol, initEntries,
                           liveEntries, deadEntries);
+    mBucketListSizeMeter.Mark(mBucketList->getSize());
```
marta-lokhova, line 833: I find this metric a bit confusing. Meters are good for "rate of something happening per desired unit of time". Here we update the meter with full BL size every ledger, but the meter's default unit of time is seconds, so the resulting rate you'll see will be bytes per second (BL size divided by 5 roughly) aggregated over several ledgers. So I'm not sure how to reason about it. 

You probably want to use a counter that you set once a ledger to get precise BL size, then our metrics exporter can properly scrape it and plot the size changes overtime

marta-lokhova (changes requested):

marta-lokhova (approved):

marta-lokhova:
r+ 2a09c8182dfe3806ee68548f8e36337a1c86ec38

PR #4090, opened 2023-12-13
3437 improve handling of slow archives

### LINKED ISSUE #3437: Improve handling of slow archives
Right now, if a slow archive is being queried, there is very little information on what is going on and archive commands don't have a good way to handle failures and retries.

For the first issue, it would be nice to add some logging/status when certain operations take too long (or at a minimum log when archive operations start/stop).

For the second issue, a potential solution would be to add new variables that can be used in archive patterns (see below) that track things like:
* the number of retries so far (regardless of source), this can be used to calculate better timeouts etc
* "good defaults" (derived from the number of retries, and maybe some other configs) that can be used with commands like `curl`
   * connection timeout, can be used with curl's `--connect-timeout`
   * maximum time, can be used with curl's `--max-time`
   * minimum download rate, can be used with curl's `--speed-limit`

As a result, an `HISTORY` (and `get`) entry could look something like (actual variable names TBD):
```
HISTORY="curl --max-time {max_time_seconds} --speed-limit {min_download_bps} -sf http://history.stellar.org/prd/core-testnet/core_testnet_001/{source_name} -o {output_filename}"
```

ThomasBrady:
# Description

Resolves #3437 

* Exposes config to control number of retries when fetching archives
* Adds metrics for retries and throughput of files downloaded
* Logs what commands are run as part of GetRemoteFileWork

# Checklist
- [ ] Compiles
- [ ] Ran all tests

Screenshot 2024-01-02 at 6 45 24 PM

### DIFF
--- a/docs/metrics.md
+++ b/docs/metrics.md
@@ -51,6 +51,8 @@ history.check.success                     | meter     | history archive status c
 history.publish.failure                   | meter     | published failed
 history.publish.success                   | meter     | published completed successfully
 history.publish.time                      | timer     | time to successfully publish history
+history.get.throughput                    | meter     | bytes per second of history archive retrieval
+history.get.failure                       | meter     | history archive downloads failed
 ledger.age.closed                         | bucket    | time between ledgers
 ledger.age.current-seconds                | counter   | gap between last close ledger time and current time
 ledger.apply.success                      | counter   | count of successfully applied transactions
--- a/src/historywork/GetRemoteFileWork.cpp
+++ b/src/historywork/GetRemoteFileWork.cpp
@@ -8,6 +8,7 @@
 #include "history/HistoryArchiveManager.h"
 #include "history/HistoryManager.h"
 #include "main/Application.h"
+#include "util/Fs.h"
 #include "util/GlobalChecks.h"
 #include "util/Logging.h"
 
@@ -22,6 +23,10 @@ GetRemoteFileWork::GetRemoteFileWork(Application& app,
     , mRemote(remote)
     , mLocal(local)
     , mArchive(archive)
+    , mFailuresPerSecond(
+          app.getMetrics().NewMeter({"history", "get", "failure"}, "failure"))
+    , mBytesPerSecond(
+          app.getMetrics().NewMeter({"history", "get", "throughput"}, "bytes"))
 {
 }
 
@@ -52,6 +57,7 @@ void
 GetRemoteFileWork::onSuccess()
 {
     releaseAssert(mCurrentArchive);
+    mBytesPerSecond.Mark(fs::size(mLocal));
     RunCommandWork::onSuccess();
 }
 
@@ -59,6 +65,7 @@ void
 GetRemoteFileWork::onFailureRaise()
 {
     releaseAssert(mCurrentArchive);
+    mFailuresPerSecond.Mark(1);
     CLOG_ERROR(History,
                "Could not download file: archive {} maybe missing file {}",
                mCurrentArchive->getName(), mRemote);
--- a/src/historywork/GetRemoteFileWork.h
+++ b/src/historywork/GetRemoteFileWork.h
@@ -5,6 +5,7 @@
 #pragma once
 
 #include "historywork/RunCommandWork.h"
+#include "medida/medida.h"
 
 namespace stellar
 {
@@ -18,6 +19,8 @@ class GetRemoteFileWork : public RunCommandWork
     std::shared_ptr<HistoryArchive> const mArchive;
     std::shared_ptr<HistoryArchive> mCurrentArchive;
     CommandInfo getCommand() override;
+    medida::Meter& mFailuresPerSecond;
+    medida::Meter& mBytesPerSecond;
 
   public:
     // Passing `nullptr` for the archive argument will cause the work to
--- a/src/ledger/readme.md
+++ b/src/ledger/readme.md
@@ -23,7 +23,7 @@ behaves just like pointers in typical data structures but with added
 security guarantees.
 
 See the protocol file for the object definitions.
-[`src/xdr/Stellar-ledger.x`](../xdr/Stellar-ledger.x)
+[`src/protocol-curr/xdr/Stellar-ledger.x`](../protocol-curr/xdr/Stellar-ledger.x)
 
 One can think of the historical chain as a linked list of LedgerHeaders:
 


src/main/Config.h
```diff
@@ -634,5 +634,8 @@ class Config : public std::enable_shared_from_this<Config>
     void processOpApplySleepTimeForTestingConfigs();
 
     std::chrono::seconds HISTOGRAM_WINDOW_SIZE;
+
+    // How many times to retry fetching a history archive.
+    uint16_t HISTORY_ARCHIVE_RETRY_LIMIT;
```
SirTyson, line 639: This should be a `uint32_t`. I'd recommend compiling with `./configure --enable-extrachecks` during development to get warnings for issues like type mismatch. This would have given you a cast error in `src/main/Config.cpp:1442` and may have helped you spot this bug.

src/test/test.cpp
```diff
@@ -308,6 +308,7 @@ getTestConfig(int instanceNumber, Config::TestDbMode mode)
         thisConfig.PEER_READING_CAPACITY = 20;
         thisConfig.PEER_FLOOD_READING_CAPACITY = 20;
         thisConfig.FLOW_CONTROL_SEND_MORE_BATCH_SIZE = 10;
+        thisConfig.HISTORY_ARCHIVE_RETRY_LIMIT = 10;
```
SirTyson, line 311: This is redundant, the default value is already set to 10 in `Config.cpp`

docs/stellar-core_example.cfg
```diff
@@ -29,7 +29,7 @@ HISTOGRAM_WINDOW_SIZE=30
 # This will get written to a lot and will grow as the size of the ledger grows.
 BUCKET_DIR_PATH="buckets"
 
-
```
SirTyson, line 32: You should add a comment here explaining this config. Also, I think it would be helpful to move this down over by the History section on line 686.

docs/metrics.md
```diff
@@ -50,6 +50,8 @@ history.check.success                     | meter     | history archive status c
 history.publish.failure                   | meter     | published failed
 history.publish.success                   | meter     | published completed successfully
 history.publish.time                      | timer     | time to successfully publish history
+history.get.throughput                    | meter     | bytes per second of history archive retrieval
+history.get.retry                         | counter     | count of retries to get history
```
SirTyson, line 54: Nit: spacing is off for `history.get.retry`. I'd also make this a little more clear, something like "count of history archive download retries"

src/catchup/CatchupWork.cpp
```diff
@@ -342,10 +342,12 @@ CatchupWork::getAndMaybeSetHistoryArchiveState()
                     ? CatchupConfiguration::CURRENT
                     : mApp.getHistoryManager().checkpointContainingLedger(
                           mCatchupConfiguration.toLedger());
+
```
SirTyson, line 345: Nit: Extra space. Everyone has a bit of a different newline style, but we like to keep the diffs clean.

src/catchup/CatchupWork.cpp
```diff
@@ -342,10 +342,12 @@ CatchupWork::getAndMaybeSetHistoryArchiveState()
...
             // Set retries to 10 to ensure we retry enough in case current
```
SirTyson, line 346: Need to update comment since this is now a config defined value.

src/history/test/HistoryTests.cpp
```diff
@@ -137,6 +137,7 @@ TEST_CASE("History bucket verification", "[history][catchup]")
     VirtualClock clock;
     auto cg = std::make_shared<TmpDirHistoryConfigurator>();
     cg->configure(cfg, true);
+    releaseAssert(cfg.HISTORY_ARCHIVE_RETRY_LIMIT == 10);
```
SirTyson, line 140: Since this is a default value and we're not testing the value setting, we don't need to assert here. Also, for tests we prefer using `REQUIRE` over `releaseAssert`. It's the same thing, but `REQUIRE` gives you better traces.

I usually like writing a ton of asserts like this during development, but remove them once before I open a PR. It's good to keep `releaseAsserts` to check for things like null pointers, or for situations like if a given function should only be called in a certain scenario. However for sanity checking simple values that won't change, it's a little cleaner to leave these out once everything's working.

src/historywork/GetRemoteFileWork.cpp
```diff
@@ -37,13 +41,23 @@ GetRemoteFileWork::getCommand()
     releaseAssert(mCurrentArchive);
     releaseAssert(mCurrentArchive->hasGetCmd());
     auto cmdLine = mCurrentArchive->getFileCmd(mRemote, mLocal);
-
+    // getCommand is called from RunCommandWork::onRun immediately before
+    // running the command.  Record the time here so that we can calculate
+    // throughput in onSuccess.
+    mDownloadStart = std::chrono::steady_clock::now();
+    if (getRetryCount() > 0)
```
SirTyson, line 48: I don't think this currently works, and the retry metric will always be 0. It looks like in `src/catchup/CatchupWork.cpp`, you set `GetHistoryArchiveStateWork` to retry `HISTORY_ARCHIVE_RETRY_LIMIT` times. `GetHistoryArchiveStateWork` then spawns `GetRemoteFileWork`. The issue is, `GetHistoryArchiveStateWork` spawns `GetRemoteFileWork` with `RETRY_NEVER` at `src/historywork/GetHistoryArchiveStateWork.cpp:87`. What will happen is that `GetHistoryArchiveStateWork` will keep retrying, but on each retry spin up a fresh `GetRemoteFileWork` work. This means that on each iteration, `GetRemoteFileWork.getRetryCount()` will always be 0.

I could be wrong though, since `GetRemoteFileWork` is a child of `GetHistoryArchiveStateWork` it may inherit it's retry counter or something. One way to test this behavior is to run a watcher node, but set invalid URLs for the history archives. You can then run the node and use the `--metric` flag to print your new metrics to the console. This will help you verify that the metric counter is working properly, since it should have to retry 10 times since the URL is invalid.

src/ledger/readme.md
```diff
@@ -23,7 +23,7 @@ behaves just like pointers in typical data structures but with added
 security guarantees.
 
 See the protocol file for the object definitions.
-[`src/xdr/Stellar-ledger.x`](../xdr/Stellar-ledger.x)
+[`src/protocol-curr/xdr/Stellar-ledger.x`](../protocol-curr/xdr/Stellar-ledger.x)
```
SirTyson, line 26: Good catch, thanks!

src/historywork/GetRemoteFileWork.cpp
```diff
@@ -37,13 +41,23 @@ GetRemoteFileWork::getCommand()
...
+    }
+    CLOG_INFO(History, "Downloading file '{}' from archive '{}', COMMAND='{}'",
+              mRemote, mCurrentArchive->getName(), cmdLine);
     return CommandInfo{cmdLine, std::string()};
 }
 
 void
 GetRemoteFileWork::onReset()
 {
+    mDownloadStart = std::chrono::steady_clock::now();
```
SirTyson, line 60: I don't think this is the correct place to set the start timer. I think it would be cleaner to wrap `mDownloadStart` in an optional and clear it on reset. Then in the path that begins the download, you can check if `mDownloadStart` is empty and start the timer then. This should be cleaner also if you move this up to `GetHistoryArchiveStateWork`

SirTyson (changes requested):
Off to a good start! I think it may be useful to step up a level and do the metrics and timings in `GetHistoryArchiveStateWork`. What you could do is still set the number of retries for `GetHistoryArchiveStateWork`. In the `else if (state == State::WORK_FAILURE && archive)` branch after you see a failure, you can then update the metrics with the time and count. This would also avoid noisy metrics `GetAndUnzipRemoteFileWork`, as I think we're only interested in `GetHistoryArchiveStateWork` (I'm not sure on this point though and need to double check).

docs/stellar-core_example.cfg
ThomasBrady: Good shout. But maybe it should live in the catchup config section? (~331)

docs/stellar-core_example.cfg
```diff
@@ -29,7 +29,7 @@ HISTOGRAM_WINDOW_SIZE=30
...
+HISTORY_ARCHIVE_RETRY_LIMIT = 10
```
marta-lokhova, line 32: Any particular reason to pick 10 here? This changes the retry strategy we currently have implemented. For downloading files from the archives, we use `GetAndUnzipRemoteFileWork` which defaults to BasicWork::RETRY_A_LOT (32), so I think we should keep that as default.

marta-lokhova: Thinking about this a bit more, what do we want to achieve with this config in general? Glancing at the original issue, the discussion was around adding more visibility into download patterns, so I'm trying to understand how operators would use this new config.

src/catchup/CatchupWork.cpp
```diff
@@ -342,10 +342,12 @@ CatchupWork::getAndMaybeSetHistoryArchiveState()
...
             // checkpoint isn't published yet
             mGetHistoryArchiveStateWork = addWork<GetHistoryArchiveStateWork>(
-                toCheckpoint, mArchive, true, 10);
+                toCheckpoint, mArchive, true,
+                mApp.getConfig().HISTORY_ARCHIVE_RETRY_LIMIT);
```
marta-lokhova, line 350: Catchup has a few other steps that download files (`DownloadApplyTxsWork`, `DownloadBucketsWork`, etc), but looks like those do not use the new config?

src/historywork/GetRemoteFileWork.cpp
marta-lokhova: agreed with @SirTyson. You could probably create a new Work wrapper class `DownloadWorkWithMetrics` or something like that, which `GetHistoryArchiveStateWork` and `GetAndUnzipRemoteFileWork` would inherit from. This way we'll automatically have metrics for all archive download works. In that new work, `onFailureRetry` is a good place to update the retry metric, this way you don't need to track retries manually.

src/historywork/GetRemoteFileWork.cpp
```diff
@@ -37,13 +41,23 @@ GetRemoteFileWork::getCommand()
...
-
+    // getCommand is called from RunCommandWork::onRun immediately before
+    // running the command.  Record the time here so that we can calculate
+    // throughput in onSuccess.
+    mDownloadStart = std::chrono::steady_clock::now();
+    if (getRetryCount() > 0)
+    {
+        mRetryRemoteFile.inc();
+    }
+    CLOG_INFO(History, "Downloading file '{}' from archive '{}', COMMAND='{}'",
```
marta-lokhova, line 52: this can get pretty noisy if we're running a large catchup work. We try to keep the info log level tidy, and as compact as possible (verbose logs can cause log partitions to run out of space before they get rotated). We already have some DEBUG logs in `GetAndUnzipRemoteFileWork`, so perhaps that should be sufficient?

src/historywork/GetRemoteFileWork.cpp
```diff
@@ -52,6 +66,17 @@ void
 GetRemoteFileWork::onSuccess()
 {
     releaseAssert(mCurrentArchive);
+
+    // Get size of local file and calculate throughput as size_bytes / (now() -
+    // mDownloadStart)
+    auto size = fs::size(mLocal);
+    auto duration = std::chrono::steady_clock::now().time_since_epoch() -
```
marta-lokhova, line 73: nit: you can use `mApp.getClock().now() - mDownloadStart` here

src/historywork/GetRemoteFileWork.h
```diff
@@ -18,6 +19,9 @@ class GetRemoteFileWork : public RunCommandWork
     std::shared_ptr<HistoryArchive> const mArchive;
     std::shared_ptr<HistoryArchive> mCurrentArchive;
     CommandInfo getCommand() override;
+    medida::Counter& mRetryRemoteFile;
+    medida::Meter& mBytesPerSecondRemoteFile;
+    std::chrono::time_point<std::chrono::steady_clock> mDownloadStart;
```
marta-lokhova, line 24: nit: use `VirtualClock::time_point` typedef here

src/historywork/GetRemoteFileWork.cpp
```diff
@@ -52,6 +66,17 @@ void
...
+                    mDownloadStart.time_since_epoch();
+    auto durationSeconds =
+        std::chrono::duration_cast<std::chrono::seconds>(duration);
+    auto throughput = size / durationSeconds.count();
```
marta-lokhova, line 77: this crashes for fast downloads (divide by 0)

src/historywork/GetRemoteFileWork.cpp
```diff
@@ -52,6 +66,17 @@ void
...
+    mBytesPerSecondRemoteFile.Mark(throughput);
```
marta-lokhova, line 78: meter already computes rate per second, do I don't think you need to compute rate here manually. So this should be mBytesPerSecondRemoteFile.Mark(size)

marta-lokhova (changes requested):

SirTyson:
I’m not sure if the semantics of `HISTORY_ARCHIVE_RETRY_LIMIT` are quite right. Currently, if we try to download from an archive and it fails, we try another archive in a round robin fashion. I don’t know if the `HISTORY_ARCHIVE_RETRY_LIMIT` would be better as a config that determines how many times you try another archive before failing vs. trying to download from the same archive multiple times. Looking at the original issue, I don’t know if this config setting addresses the archive pattern variables that Nico originally mentioned. Sorry about that, I think I lost track of the original issue when going back and forth on the design.

I think what makes the most sense is to keep the metrics in this PR (which we definitely want), but move the config issues to a follow up issue where we can discuss a little more what we want. I agree with Marta’s `DownloadWorkWithMetrics` suggestion and we can keep the metrics moving. Again, sorry about getting confused on the issue.

src/historywork/GetRemoteFileWork.h
```diff
@@ -18,6 +19,9 @@ class GetRemoteFileWork : public RunCommandWork
     std::shared_ptr<HistoryArchive> const mArchive;
     std::shared_ptr<HistoryArchive> mCurrentArchive;
     CommandInfo getCommand() override;
+    medida::Counter& mRetryRemoteFile;
```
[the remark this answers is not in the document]
marta-lokhova: Just a small nit on resolving PR comment threads: we typically allow reviewers to resolve their original comments (or request further changes in the same thread after reviewing new commits)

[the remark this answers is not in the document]
marta-lokhova: (this makes it easy to keep track of work left to get PRs across the finish line)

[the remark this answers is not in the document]
ThomasBrady: Noted! 😄

src/historywork/DownloadWorkWithMetrics.h
```diff
@@ -0,0 +1,28 @@
+// Copyright 2015 Stellar Development Foundation and contributors. Licensed
```
SirTyson, line 1: Nit: update copyright

src/historywork/GetRemoteFileWork.h
```diff
@@ -4,7 +4,7 @@
 
 #pragma once
 
-#include "historywork/RunCommandWork.h"
```
SirTyson, line 7: This include should not have changed.

src/historywork/GetAndUnzipRemoteFileWork.h
```diff
@@ -13,7 +14,7 @@ namespace stellar
 class HistoryArchive;
 class GetRemoteFileWork;
 
-class GetAndUnzipRemoteFileWork : public Work
```
SirTyson, line 16: I think `GetHistoryArchiveStateWork` should also inherit from `DownloadWorkWithMetrics`.

SirTyson (changes requested):
Looking pretty close! A few minor nits, and I think another class needs to inherit from `DownloadWorkWithMetrics`.

src/historywork/DownloadWorkWithMetrics.cpp
```diff
@@ -0,0 +1,37 @@
+// Copyright 2015 Stellar Development Foundation and contributors. Licensed
```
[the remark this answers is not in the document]
ThomasBrady: Happy new year 🎉

src/historywork/GetHistoryArchiveStateWork.h
```diff
@@ -5,15 +5,15 @@
...
+#include "historywork/DownloadWorkWithMetrics.h"
 
 namespace stellar
 {
 
 class HistoryArchive;
 class GetRemoteFileWork;
 
-class GetHistoryArchiveStateWork : public Work
+class GetHistoryArchiveStateWork : public DownloadWorkWithMetrics
```
SirTyson, line 16: Sorry, I didn't catch this initially, but it looks like there's an inconsistency here with the `mBytesPerSecond` metric. The issue is the meter assumes that the download finishes whenever the work that inherits `DownloadWorkWithMetrics` finishes. For `GetHistoryArchiveStateWork`, this is fine because the entire class is effectively a wrapper around `GetRemoteFileWork` and does no additional work. However, `GetAndUnzipRemoteFileWork` first downloads the file, then unzips it, then finishes. This means that even if the speed of the download is identical, files downloaded via `GetAndUnzipRemoteFileWork` will always appear to be slower because you count unzip time.

I think the easiest way to fix this is probably just to do your metering in `GetRemoteFileWork`. There's no non-history related work that uses this, so we can probably just do this metering directly in the class. A wrapper class like `DownloadWithMetrics` would be useful if `GetRemoteFileWork` was used for use cases outside of history downloads, but I don't see any. You can mark the `bytesPerSecond` value in the `onSuccess` function still. Since we never retry the work directly, you can mark the retry/failure meter in `onFailureRaise`.

SirTyson (changes requested):

src/historywork/GetHistoryArchiveStateWork.h
ThomasBrady: Ah right, good catch on the unzip time being included in the denominator. Regarding moving the metrics to `GetRemoteFileWork`, it doesn't make sense to call the metric `history.get.retries` (as its not exactly a retry when incremented from a failing `GetRemoteFileWork`), perhaps `history.get.failures`?

SirTyson (approved):
Looks good to me! A nice clean change after all the back and forth.

marta-lokhova:
r+ c81c74084f6924670093f35f8b358c40eb9935ac

PR #4091, opened 2023-12-13
Cache tx set resolution

marta-lokhova:


dmkozh:
r+ 8b336b4

dmkozh:
r+ 3e1849a

PR #4093, opened 2023-12-14
Bucket cleanup

### LINKED ISSUE #4031: Harden Bucket File IO
Currently, `Bucket` uses a single IO stream for both BucketDB lookups and eviction scans. This is dangerous, as we must perform BucketDB lookups during the eviction scan process. We should allocate a separate streams for eviction scans and investigate if there is a better IO interface other than streams for the random access required by BucketDB.

SirTyson:
# Description

Resolves #4031

This PR includes cleanups and hardening of `Bucket` related code.

- `loadPoolShareTrustlinesByAccountAndAsset`: Extended a test to make sure this function does not return outdated entries. This was working properly, but did not have an explicit test. I've also added an explicit `seenKeys` set instead of relying on the fact that `unordered_map::emplace` will not update a key if it already exists for more defensive programming. Finally, there was an edge case where if a `Bucket` only contained offers and did not contain a `METAENTRY`, the scan would return early. This should never happen in practice, but it has been fixed.
- XDRStream audit: I took a deep dive into `ifstream` and ran several experiments such as disabling stream buffering. I've concluded that `ifstream` with buffering is still the optimal file IO interface for BucketsDB. I've also checked that we're handling EOF and bad bits properly, and could not find any dependencies on out-of-spec behavior. The only code change was a minor scoping issue on a tracy metric.
- Added separate stream for BucketDB scans and eviction scans.
- Warn more aggressively if eviction scans get stuck. Previously, there was only a single warning per stuck bucket, and there we certain edge cases where the warning would not be triggered, such as if a large bucket was almost entirely TEMPORARY entries. Now, we warn on every ledger and check much more aggressively.

# Checklist
- [x] Compiles
- [x] Ran all tests

### DIFF
--- a/src/bucket/Bucket.cpp
+++ b/src/bucket/Bucket.cpp
@@ -255,7 +255,7 @@ Bucket::loadPoolShareTrustLinessByAccount(
     // Get upper and lower bound for poolshare trustline range associated
     // with this account
     auto searchRange = getIndex().getPoolshareTrustlineRange(accountID);
-    if (searchRange.first == 0 && searchRange.second == 0)
+    if (!searchRange)
     {
         // No poolshare trustlines, exit
         return;
@@ -263,8 +263,8 @@ Bucket::loadPoolShareTrustLinessByAccount(
 
     BucketEntry be;
     auto& stream = getIndexStream();
-    stream.seek(searchRange.first);
-    while (stream && stream.pos() < searchRange.second && stream.readOne(be))
+    stream.seek(searchRange->first);
+    while (stream && stream.pos() < searchRange->second && stream.readOne(be))
     {
         LedgerEntry entry;
         switch (be.type())
@@ -292,7 +292,7 @@ Bucket::loadPoolShareTrustLinessByAccount(
         }
 
         // If this is a pool share trustline that matches the accountID and
-        // is not shadowed, add it to results
+        // is the newest version of the key, add it to results
         if (trustlineCheck(entry.data) &&
             seenTrustlines.find(LedgerEntryKey(entry)) == seenTrustlines.end())
         {
--- a/src/bucket/Bucket.cpp
+++ b/src/bucket/Bucket.cpp
@@ -221,7 +221,7 @@ Bucket::loadKeys(std::set<LedgerKey, LedgerEntryIdCmp>& keys,
 
 void
 Bucket::loadPoolShareTrustLinessByAccount(
-    AccountID const& accountID, UnorderedSet<LedgerKey>& deadTrustlines,
+    AccountID const& accountID, UnorderedSet<LedgerKey>& seenTrustlines,
     UnorderedMap<LedgerKey, LedgerEntry>& liquidityPoolKeyToTrustline,
     LedgerKeySet& liquidityPoolKeys)
 {
@@ -238,7 +238,7 @@ Bucket::loadPoolShareTrustLinessByAccount(
     // Get upper and lower bound for poolshare trustline range associated
     // with this account
     auto searchRange = getIndex().getPoolshareTrustlineRange(accountID);
-    if (searchRange.first == 0)
+    if (searchRange.first == 0 && searchRange.second == 0)
     {
         // No poolshare trustlines, exit
         return;
@@ -265,7 +265,7 @@ Bucket::loadPoolShareTrustLinessByAccount(
             // later
             if (trustlineCheck(key))
             {
-                deadTrustlines.emplace(key);
+                seenTrustlines.emplace(key);
             }
             continue;
         }
@@ -277,8 +277,9 @@ Bucket::loadPoolShareTrustLinessByAccount(
         // If this is a pool share trustline that matches the accountID and
         // is not shadowed, add it to results
         if (trustlineCheck(entry.data) &&
-            deadTrustlines.find(LedgerEntryKey(entry)) == deadTrustlines.end())
+            seenTrustlines.find(LedgerEntryKey(entry)) == seenTrustlines.end())
         {
+            seenTrustlines.emplace(LedgerEntryKey(entry));
             auto const& poolshareID =
                 entry.data.trustLine().asset.liquidityPoolID();
 
--- a/src/bucket/Bucket.cpp
+++ b/src/bucket/Bucket.cpp
@@ -77,16 +77,33 @@ Bucket::Bucket()
 {
 }
 
-XDRInputFileStream&
-Bucket::getStream()
+std::unique_ptr<XDRInputFileStream>
+Bucket::openStream()
 {
-    if (!mStream)
+    releaseAssertOrThrow(!mFilename.empty());
+    auto streamPtr = std::make_unique<XDRInputFileStream>();
+    streamPtr->open(mFilename.string());
+    return std::move(streamPtr);
+}
+
+XDRInputFileStream&
+Bucket::getIndexStream()
+{
+    if (!mIndexStream)
     {
-        mStream = std::make_unique<XDRInputFileStream>();
-        releaseAssertOrThrow(!mFilename.empty());
-        mStream->open(mFilename.string());
+        mIndexStream = openStream();
     }
-    return *mStream;
+    return *mIndexStream;
+}
+
+XDRInputFileStream&
+Bucket::getEvictionStream()
+{
+    if (!mEvictionStream)
+    {
+        mEvictionStream = openStream();
+    }
+    return *mEvictionStream;
 }
 
 Hash const&
@@ -139,7 +156,7 @@ void
 Bucket::freeIndex()
 {
     mIndex.reset(nullptr);
-    mStream.reset(nullptr);
+    mIndexStream.reset(nullptr);
 }
 
 std::optional<BucketEntry>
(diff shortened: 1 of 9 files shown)


src/bucket/Bucket.cpp
```diff
@@ -238,14 +250,14 @@ Bucket::loadPoolShareTrustLinessByAccount(
     // Get upper and lower bound for poolshare trustline range associated
     // with this account
     auto searchRange = getIndex().getPoolshareTrustlineRange(accountID);
-    if (searchRange.first == 0)
+    if (searchRange.first == 0 && searchRange.second == 0)
```
dmkozh, line 253: Isn't this actually a protocol change?

src/bucket/Bucket.cpp
```diff
@@ -78,15 +78,27 @@ Bucket::Bucket()
...
+}
+
+XDRInputFileStream&
+Bucket::getEvictionStream()
+{
+    if (!mEvictionStream)
+    {
+        mEvictionStream = std::make_unique<XDRInputFileStream>();
+        releaseAssertOrThrow(!mFilename.empty());
+        mEvictionStream->open(mFilename.string());
```
dmkozh, line 99: IIUC the intention is to open two streams on the same file, right? A helper would be useful here then to avoid duplication and improve clarity.

src/bucket/Bucket.h
```diff
@@ -109,10 +115,10 @@ class Bucket : public std::enable_shared_from_this<Bucket>,
     // stored with their corresponding liquidity pool key in
     // liquidityPoolKeyToTrustline. All liquidity pool keys corresponding to
     // loaded trustlines are also reduntantly stored in liquidityPoolKeys.
-    // If a trustline key is in deadTrustlines, it is not loaded. Whenever a
-    // dead trustline is found, its key is added to deadTrustlines.
+    // If a trustline key is in seenTrustlines, it is not loaded. Whenever a
+    // dead trustline is found, its key is added to seenTrustlines.
```
dmkozh, line 119: dead->seen? I don't quite understand the logic of the change here though.

src/bucket/BucketList.cpp
```diff
@@ -892,6 +891,22 @@ BucketList::scanForEviction(Application& app, AbstractLedgerTxn& ltx,
         auto initialLevel = evictionIter.bucketListLevel;
         auto initialIsCurr = evictionIter.isCurrBucket;
         auto b = getBucketFromIter(evictionIter);
+
+        // Check to see if we can finish scanning the bucket before it receives
+        // an update
+        auto period = bucketUpdatePeriod(evictionIter.bucketListLevel,
```
dmkozh, line 897: Is this is a protocol change?

src/bucket/Bucket.cpp
SirTyson: No, if there are no poolshare trustline keys in a bucket, `getPoolshareTrustlineRange` returns the pair {0, 0}. We only checked the first 0, so this hardens the check slightly. This pair is the file offsets within the bucket to scan for the trustline entries. In practice, it should be impossible for any `LedgerEntry` to start at an offset of 0, because the first entry should always be a `MetaEntry`. This is why checking just the first member of the pair is sufficient. The first possible offset for a `LedgerEntry` should always by 16. This just hardens the check a little, since `getPoolshareTrustlineRange` is expected to return {0, 0} if no keys exist.

src/bucket/BucketList.cpp
SirTyson: No, this only effects metrics and logging.

src/bucket/Bucket.h
SirTyson: For a given key, there is only one valid version of the entry. However, because the bucket list is a log-structured merge tree, multiple different versions of a key can exist simultaneously on different levels. For example, if a key has been recently deleted, there may be a DEADENTRY of the key marking that it has been deleted on level 0, but a LIVEENTRY of the same key on some lower level N. For this particular query, we have to scan all levels of the BucketList. Previously, this set was called `deadTrustlines` because it kept track of all the DEADENTRY keys we had already seen. That way, if we scan a DEADENTRY on level 0, we know not to later load that entry if we find a LIVEENTRY version on a lower level.

However, if an entry has been updated multiple times, it's possible that there are multiple LIVEENTRIES on multiple levels of the bucket list with different values of the same key. Suppose we have some key k that exists as a LIVEENTRY on both level 0 and 2. This function would first load the newest version on level 0, and emplace it into `liquidityPoolKeyToTrustline`. Later on, we then load the outdated version on level 2. This key is not in `deadTrustlines`, since there has been no DEADENTRY. So, we would again call `emplace` into `liquidityPoolKeyToTrustline` for the outdated version of the entry. In practice this is fine, since the key already exists in the map so emplace would just be a no-op.

This change cleans up the code to make this behavior more explicit. The first time we see a key, we add it to `seenTrustlines`. Whenever we call emplace, we check to make sure we haven't seen the key yet. If we have already seen the key, we know the version we just scanned is outdated, so we don't call emplace and continue.

src/bucket/Bucket.cpp
```diff
@@ -77,16 +77,33 @@ Bucket::Bucket()
 {
 }
 
+void
+Bucket::openStream(std::unique_ptr<XDRInputFileStream>& streamPtr)
+{
+    releaseAssertOrThrow(!streamPtr);
```
dmkozh, line 83: nit: just return a unique_ptr  and you won't need to worry about verifying the reference. It's generally a bit weird style.

src/bucket/Bucket.h
dmkozh: Thanks for the explanation!

src/bucket/BucketList.cpp
```diff
@@ -891,6 +891,22 @@ BucketList::scanForEviction(Application& app, AbstractLedgerTxn& ltx,
...
+                                         evictionIter.isCurrBucket);
+
+        // Ledgers remaining until the current Bucket changes
+        auto ledgersRemainingToScanBucket = period - (ledgerSeq % period);
+        auto bytesRemaining = b->getSize() - evictionIter.bucketFileOffset;
```
marta-lokhova, line 902: Is the bytes remaining calculation here compared to using total bucket size previously done to get more accuracy in case we hit LE limit first during eviction? Also, any particular reason for the code move?

marta-lokhova (changes requested):

src/bucket/BucketList.cpp
SirTyson: Correct, the previous check assumed you read the maximum amount of bytes possible each ledger, which may not be true given the LE limit. The code moved so that we check this every ledger. Previously, we only did this check once when switching to a new bucket. Now, we check every ledger. This will warn more aggressively so it's harder to miss and also allows us to catch the case where we get stuck due to the LE limit.

SirTyson: Correct, the previous check assumed you read the maximum amount of bytes possible each ledger, which may not be true given the LE limit. The code moved so that we check this every ledger. Previously, we only did this check once when switching to a new bucket. Now, we check every ledger. This will warn more aggressively so it's harder to miss and also allows us to catch the case where we get stuck due to the LE limit.

src/util/XDRStream.h
```diff
@@ -200,14 +200,16 @@ class XDRInputFileStream
...
 
-            ZoneNamedN(__unpack, "xdr_unpack_entry", true);
-            releaseAssert(xdrStart <= xdrEnd);
-            releaseAssert(xdrEnd <= mBuf.size());
-            xdr::xdr_get g(mBuf.data() + xdrStart, mBuf.data() + xdrEnd);
-            xdr::xdr_argpack_archive(g, out);
-            if (getBucketLedgerKey(out) == key)
             {
-                return true;
+                ZoneNamedN(__unpack, "xdr_unpack_entry", true);
```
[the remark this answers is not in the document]
marta-lokhova: mm, it doesn't seem to be the case for [ZoneNamedN](https://github.com/wolfpld/tracy/blob/3c4b06dff5472e8f28be361195c32db4731e2f46/manual/tracy.tex#L1256) though (it requires supplying a name for the created stack variable)

src/bucket/Bucket.cpp
marta-lokhova: I don't think we can make assumptions regarding offset validity: earlier versions of buckets do not support meta entries, so offset 0 is possible. It would be nice to make the API less error-prone (and fix the offset 0 issue) by explicitly returning a nullopt if there are no trustline keys in the bucket

src/bucket/Bucket.cpp
```diff
@@ -855,7 +873,7 @@ Bucket::scanForEviction(AbstractLedgerTxn& ltx, EvictionIterator& iter,
         return true;
     }
 
-    auto& stream = getStream();
+    auto& stream = getEvictionStream();
     stream.seek(iter.bucketFileOffset);
```
marta-lokhova, line 877: isn't this `seek` call redundant given that eviction stream should now guarantee position validity? (we should probably also assert that bucketFileOffset is consistent with stream pos)

SirTyson: No, if a node is catching up or joining the network, the stream will need to be seeked to the correct position based on the config parameter. I'd rather error on the safer side and re-seek every time instead of special casing an init path.

marta-lokhova: Oh, so the correct position is not initialized until we call eviction scan for the first time. Then the comment in Bucket.h is a bit confusing, as it implies we get a valid stream without the need to re-seek. So maybe clarify that in the comment?

marta-lokhova (approved):

marta-lokhova:
r+ 4f0789d26fc339304b47cbad2f6c764e21499ce7

PR #4097, opened 2023-12-15
Add standalone `check-quorum-intersection` command

### LINKED ISSUE #4062: Consider exposing quorum intersection as a standalone command
`QuorumIntersectionChecker` implements an efficient quorum intersection checker used as to report based on the current internal state of the SCP state machine.

We should allow to invoke the same functionality based on some arbitrary input:
* it should be a subcommand invoked from the command line (not over http) so that it can be used in scripts
* input should probably be json (file or stdin) as it's what we use in other places, and it's easy to transform json into json outside of core
* we may want to make it compatible with the output of the http command `quorum?transitive=true` (allowing to use short/friendly names)

bboston7:
# Description

Resolves #4062

This change adds a command `check-quorum-intersection` that takes a JSON file produced by the `quorum` endpoint and checks whether it enjoys a quorum intersection.

Most of the change centers around reading JSON in, but this change also required refactoring `QuorumIntersectionChecker` to use a map from `NodeID` to `SCPQuorumSetPtr` instead of a map from `NodeID` to `NodeInfo`, as the intersection checking only relies on the `mQuorumSet` field of a `QuorumMap` and the JSON input does not contain everything in `NodeInfo`.

# Checklist
- [x] Compiles
- [x] Ran all tests


src/main/test/ApplicationUtilsTests.cpp
```diff
@@ -658,4 +658,42 @@ TEST_CASE("application major version numbers", "[applicationutils]")
     CHECK(getStellarCoreMajorReleaseVersion("v19.1.2-10") == std::nullopt);
     CHECK(getStellarCoreMajorReleaseVersion("v19.9.0-30-g726eabdea-dirty") ==
           std::nullopt);
+}
+
+TEST_CASE("standalone quorum intersection check", "[applicationutils]")
+{
+    Config cfg = getTestConfig();
+    const std::string JSON_ROOT = "testdata/check-quorum-intersection-json/";
```
nano-o, line 666: This fails to open the file if stellar-core is not run from src/

src/main/CommandLine.cpp
```diff
@@ -1360,6 +1360,48 @@ getSettingsUpgradeTransactions(CommandLineArgs const& args)
         });
 }
 
+int
+runCheckQuorumIntersection(CommandLineArgs const& args)
+{
+    CommandLine::ConfigOption configOption;
+    std::string jsonPath;
+    return runWithHelp(
+        args, {configurationParser(configOption), fileNameParser(jsonPath)},
```
nano-o, line 1369: It seems that a configuration file should not be necessary to check the provided json file? If that's the case then it might be better not to require `--conf`.

bboston7: You're right that the algorithm itself doesn't need the config, but `QuorumIntersectionCheckerImpl` requires one and uses it to generate human readable names (if available). I'll try to make the config optional instead.

bboston7: I just pushed up [a commit](https://github.com/stellar/stellar-core/pull/4097/commits/32a690e220403b367dfca7b62355c908a6ea7a7f) that makes `--conf` optional.

nano-o: Thanks! It looks like it's trying load a default config file though:
```
./src/stellar-core check-quorum-intersection src/testdata/test-network.json 
2023-12-19T15:54:58.703 [default INFO] Config from stellar-core.cfg
2023-12-19T15:54:58.703 [SCP FATAL] check-quorum-intersection encountered an error: No config file stellar-core.cfg found
```

src/main/test/ApplicationUtilsTests.cpp
bboston7: Fixed in 3f6f7e65dc8477b872e6db8e7c7c6bf8113ca467 by inlining the JSON into the tests.

src/scp/SCPDriver.cpp
```diff
@@ -67,6 +67,12 @@ SCPDriver::toStrKey(NodeID const& pk, bool fullKey) const
     return fullKey ? KeyUtils::toStrKey(pk) : toShortString(pk);
 }
 
+NodeID
+SCPDriver::fromStrKey(std::string const& pkStr) const
+{
+    return KeyUtils::fromStrKey<NodeID>(pkStr);
```
marta-lokhova, line 73: should we also update https://github.com/stellar/stellar-core/blob/89b1dcd20e61d1c0970ffa589da6913220554e76/src/main/CommandHandler.cpp#L1145C5-L1145C5 to use this new helper?

src/main/CommandLine.cpp
bboston7: Whoops, I forgot to actually mark it optional in this case. *now* it should work with 2408801b6410d31d02f261dd82e65907e8a5555e

src/main/test/ApplicationUtilsTests.cpp
marta-lokhova: we run tests from `src` normally anyways, as other tests rely on `testdata` folder, so I think this is fine. Having some test json files is helpful since they can also be used to manually test the command from command line, so I'd recommend we keep them.

src/main/CommandLine.cpp
marta-lokhova: Alternatively, it seems cleaner to just make the Config argument in QuorumIntersectionChecker optional, then we don't need to construct the default config at all (I think it's also a bit error-prone to pass default config to a module that is expecting the real user config).

src/main/test/ApplicationUtilsTests.cpp
bboston7: That makes sense, it's nice to have the JSON files. I reverted the change that inlined the tests.

src/main/CommandLine.cpp
bboston7: Good idea. I changed the `Config` in `QuorumIntersectionChecker` to a `std::optional<Config>` changed the front end to pass `std::nullopt` when `--conf` isn't passed (see 31dcd9a31c448fdc749895aacf74b8649077e523 for details)

src/scp/SCPDriver.cpp
bboston7: Good idea. `CommandHandler` doesn't have access to an `SCPDriver`, so I added a function to `Herder` that exposes the `fromStrKey` function in its `HerderSCPDriver` and called that from `CommandHandler::surveyTopology` (see 2dec3aefcb8f09940b43add26887e2b69586f400)

marta-lokhova: Thanks! I think the new helper should live in KeyUtils then, since it's not related to SCPDriver (it should also be static/free function)

src/main/CommandLine.cpp
```diff
@@ -1360,6 +1360,65 @@ getSettingsUpgradeTransactions(CommandLineArgs const& args)
...
+                if (configOption.mConfigFile.empty())
+                {
+                    // Need to set up logging in this case because there is no
+                    // `getConfig` call (which would otherwise set up logging)
+                    Logging::setLoggingToConsole(true);
+                    Logging::setLogLevel(configOption.mLogLevel, nullptr);
+                }
+                else
+                {
+                    cfg.emplace(configOption.getConfig(false));
```
marta-lokhova, line 1389: why not log to file?

src/scp/LocalNode.cpp
```diff
@@ -407,6 +407,52 @@ LocalNode::toJson(SCPQuorumSet const& qSet,
...
+SCPQuorumSet
+LocalNode::fromJson(Json::Value const& qSetJson) const
+{
+    return fromJson(
+        qSetJson, [&](std::string const& s) { return mDriver.fromStrKey(s); });
+}
+
+SCPQuorumSet
+LocalNode::fromJson(Json::Value const& qSetJson,
+                    std::function<NodeID(std::string const&)> const toNodeID)
```
marta-lokhova, line 419: Do we need toNodeID here? Seems like in all use cases we'd use the same fromStrKey function?

marta-lokhova (approved):
Looks great overall, thank you! Just a few minor suggestions. Also, please squash the commits so we can merge.

src/scp/SCPDriver.cpp
bboston7: Done in 2b2ecb16c119848fd26b86ec60dd20601fe61e04

src/scp/LocalNode.cpp
bboston7: I originally added this for symmetry with the `toJson` function that takes a similar function, but I agree that if we're not using this version it probably shouldn't exist. I removed it in 2b2ecb16c119848fd26b86ec60dd20601fe61e04.

src/main/CommandLine.cpp
bboston7: This is a holdover from when config files were mandatory and I thought console logging was more practical for this use case (otherwise there's very little output, and the log is tiny anyway). Now that the config file is optional I agree the command should just interpret the config file as-is. I changed this to log to a file and added a `--console` flag for those who want console logging. (see 7ebfcf096e2a971e448fc6a4b80b69fd259b1440)

bboston7:
This is rebased, squashed, and ready to merge!

marta-lokhova:
r+ a93dfe65e0a73ccde01d447d9c3bdad36bafb026

PR #4101, opened 2023-12-19
Report drifting close time on local node

### LINKED ISSUE #1815: Rethink how we check whether node's time is synchronized

marta-lokhova:
Resolves #1815

### DIFF
--- a/src/herder/HerderImpl.cpp
+++ b/src/herder/HerderImpl.cpp
@@ -52,6 +52,11 @@ using namespace std;
 namespace stellar
 {
 
+// Roughly ~10 minutes of consensus
+constexpr uint32 const CLOSE_TIME_DRIFT_LEDGER_WINDOW_SIZE = 120;
+// 10 seconds of drift threshold
+constexpr uint32 const CLOSE_TIME_DRIFT_SECONDS_THRESHOLD = 10;
+
 constexpr uint32 const TRANSACTION_QUEUE_TIMEOUT_LEDGERS = 4;
 constexpr uint32 const TRANSACTION_QUEUE_BAN_LEDGERS = 10;
 constexpr uint32 const TRANSACTION_QUEUE_SIZE_MULTIPLIER = 2;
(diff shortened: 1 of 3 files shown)


src/herder/HerderImpl.cpp
```diff
@@ -1323,6 +1328,29 @@ HerderImpl::triggerNextLedger(uint32_t ledgerSeqToTrigger,
     if (nextCloseTime <= lcl.header.scpValue.closeTime)
```
SirTyson, line 1328: Is there any particular reason we only check if a clock is drifting behind and not drifting ahead?

SirTyson (dismissed):
LGTM, just a few questions.

marta-lokhova:
@SirTyson PTAL. I re-worked the approach a little bit: now validators compare their own close time used for nomination with close time externalized by the network. If the drift is over a certain threshold, we fire a warning. This approach is based on the assumption that most validators on the network have healthy clocks. It also catches both drifting behind and drifting ahead, so no more special-casing is needed.

src/herder/HerderImpl.cpp
```diff
@@ -399,13 +404,47 @@ HerderImpl::writeDebugTxSet(LedgerCloseData const& lcd)
...
+{
+    auto it = ctMap.find(slotIndex);
+    if (it != ctMap.end())
+    {
+        it->second.second = value.closeTime;
+    }
+
+    if (ctMap.size() >= CLOSE_TIME_DRIFT_LEDGER_WINDOW)
+    {
+        medida::Histogram h(medida::SamplingInterface::SampleType::kSliding);
```
SirTyson, line 420: A comment here for this histogram would be helpful, it's pretty unclear what's happening unless you look up the definition of ctMap because of the `ct.second.second` lines.

src/herder/HerderImpl.cpp
```diff
@@ -1321,6 +1360,28 @@ HerderImpl::triggerNextLedger(uint32_t ledgerSeqToTrigger,
...
+            mDriftCTSlidingWindow[ledgerSeqToTrigger] =
+                std::make_pair(nextCloseTime, std::nullopt);
+            while (mDriftCTSlidingWindow.size() > CLOSE_TIME_DRIFT_LEDGER_WINDOW)
+            {
+                mDriftCTSlidingWindow.erase(mDriftCTSlidingWindow.begin());
+            }
+        }
+        else
+        {
+            CLOG_WARNING(Herder,
```
SirTyson, line 1379: Should we assert or throw here instead of warn? Does this indicate a logic bug, or is this possible if we have a longer than expected ledger such that the trigger timer fires multiple times on one ledger?

src/herder/HerderImpl.cpp
```diff
@@ -52,6 +52,11 @@ using namespace std;
 namespace stellar
 {
 
+// Roughly ~10 minutes of consensus
+constexpr uint32 const CLOSE_TIME_DRIFT_LEDGER_WINDOW = 120;
```
SirTyson, line 56: Nit: `CLOSE_TIME_DRIFT_LEDGER_WINDOW_SIZE` might be a better name here, at first I thought this was a time based constant and was trying to figure out how 120 mapped to 10 minutes.

SirTyson (approved):
LGTM, just a few nit pics.

src/herder/HerderImpl.cpp
marta-lokhova: we should handle this as gracefully as possible so the node can continue participating in consensus in case of buggy trigger (then the worst case is delays in nomination, which is still much better than a halt)

src/herder/HerderImpl.cpp
marta-lokhova: I re-wrote the loop with the use of structured bindings, I think that should make it more clear than a comment. Lmk what you think though.

src/herder/HerderImpl.cpp
marta-lokhova: Done.

src/herder/HerderImpl.cpp
SirTyson: Great, thanks!

SirTyson (approved):

PR #4117, opened 2024-01-04
Bump overlay31

MonsieurNicolas:
Bumps overlay version to 31 ahead of protocol 20 upgrade


src/herder/HerderSCPDriver.cpp
```diff
@@ -1223,25 +1223,15 @@ HerderSCPDriver::checkAndCacheTxSetValid(TxSetXDRFrame const& txSet,
...
-        ApplicableTxSetFrameConstPtr applicableTxSet;
-        if (txSet.previousLedgerHash() ==
-            mApp.getLedgerManager().getLastClosedLedgerHeader().hash)
-        {
-            applicableTxSet = txSet.prepareForApply(mApp);
-        }
-
-        bool res = true;
+        bool res;
+        auto applicableTxSet = txSet.prepareForApply(mApp);
```
marta-lokhova, line 1227: prepareForApply asserts that previous ledger hash equals LCL, so I don't think we can make this change without reworking `prepareForApply` as well.

marta-lokhova:
r+ 2ca423033ed721fcba5c436c0415ca803482d093

PR #4119, opened 2024-01-04
Cleanup tx queue interface and address warnings/false positives from static analysis

marta-lokhova:


src/herder/HerderImpl.cpp
```diff
@@ -2132,14 +2120,6 @@ HerderImpl::start()
     }
 
     restoreUpgrades();
-    // make sure that the transaction queue is setup against
-    // the lcl that we have right now
-    mTransactionQueue.maybeVersionUpgraded();
```
SirTyson, line 2137: Why was this removed? I assume the current ledger version is no longer required after all the refactoring we've been doing?

src/overlay/test/FloodTests.cpp
```diff
@@ -167,12 +167,15 @@ TEST_CASE("Flooding", "[flood][overlay][acceptance]")
         auto ackedTransactions = [&](std::shared_ptr<Application> app) {
             // checks if an app received all transactions or not
             size_t okCount = 0;
+            auto& herder = static_cast<HerderImpl&>(app->getHerder());
+
             for (auto const& s : sources)
             {
-                okCount +=
```
SirTyson, line 172: I'm not a huge fan of this change, as it looks like we've made `getMaxSeqInPendingTxs` more involved. I see you have to use a `HerderImpl` cast to hide it under the `BUILD_TESTS` flag. Personally, given  that `getMaxSeqInPendingTxs` is a const function (or rather a function that could be made const) I don't think it's super important to hide behind BUILD_TESTS. That being said, you're more familiar with the semantics of this function so I'll defer to you.

SirTyson (approved):
LGTM, a couple questions.

src/herder/HerderImpl.cpp
marta-lokhova: This is dead code at this point, as it's dealing with v13 protocol upgrade (this will never happen on a production network)

src/overlay/test/FloodTests.cpp
marta-lokhova: Mmm, not sure what you mean here -`getMaxSeqInPendingTxs` was removed from Herder's API completely as it was used only in this one place, so I don't see much value in exposing it publicly.

SirTyson: I'm just not a huge fan of casting to an `impl` class, but it's once place in a test so I don't think it matters.

marta-lokhova: yeah it's not great, but it's really common across our tests. Ideally, we could have a wrapper or a subclass that properly exposes test-only functions to keep the actual herder API tidy, but that's more involved...

dmkozh:
r+ 5626aba

PR #4120, opened 2024-01-05
Drop old peers post v20 upgrade

marta-lokhova:


src/overlay/Peer.h
```diff
@@ -212,7 +213,7 @@ class Peer : public std::enable_shared_from_this<Peer>,
 
     std::string mRemoteVersion;
     uint32_t mRemoteOverlayMinVersion;
-    uint32_t mRemoteOverlayVersion;
+    std::optional<uint32_t> mRemoteOverlayVersion;
```
marta-lokhova, line 216: Made this optional instead of 0, so that we don't drop pending peers when we don't know their version yet (this can happen if we've sent HELLO, but haven't received a response)

src/overlay/OverlayManagerImpl.cpp
```diff
@@ -157,6 +157,58 @@ OverlayManagerImpl::PeersList::removePeer(Peer* peer)
...
+}
+
+void
+OverlayManagerImpl::PeersList::dropIf(
+    std::function<bool(Peer::pointer, uint32_t)> predicate, uint32_t version,
+    std::string const& reason)
+{
+    maybeDrop<decltype(OverlayManagerImpl::PeersList::mPending)>(
+        mPending, predicate, version, reason);
+    maybeDrop<decltype(OverlayManagerImpl::PeersList::mAuthenticated)>(
```
MonsieurNicolas, line 208: you don't need to do all this templating stuff: you can just call `getPendingPeers` and `getRandomAuthenticatedPeers` (that also gives you a list), if you don't want to randomize for the later, you can refactor it.

MonsieurNicolas (changes requested):
looks good -- I suggested a simpler version for the way we loop over peers

src/overlay/OverlayManagerImpl.cpp
marta-lokhova: Updated.

MonsieurNicolas:
r+ de714e688562b8e9c2b08a4f4399f6185e36200d

PR #4121, opened 2024-01-05
Continue to capture SCP messages for previous ledger in database

### LINKED ISSUE #3769: Make it easier to capture SCP messages

bboston7:
# Description

Resolves #3769

Upon closing a ledger, nodes record SCP messages received during the ledger period in the `scphistory` database table. However, any messages about that ledger heard *after* close were ignored. This change captures more of those messages by modifying the `processExternalized` function (which is responsible for recording SCP messages on ledger close) to also record any new messages heard about the previous ledger.

## Performance impact

As this change adds additional database interactions to the main thread, it does have a performance impact. I've mitigated this by [wrapping the new and old `saveSCPHistory` calls in a single transaction](https://github.com/stellar/stellar-core/blob/23aa763b643375322271496dea7f5be633b7a67f/src/herder/HerderImpl.cpp#L316). I measured the performance impacts using Tracy and have summarized the results for SQLite and Postgres below.

On SQLite, the use of a single transaction for both `saveSCPHistory` calls effectively negates the performance impact of the additional call. I suspect this is because it saves an expensive write to disk, which dominates the call.

On Postgres the performance improvement from using a single transaction is minimal, and does not entirely make up for the additional saveSCPHistory call. On average, saving the SCP history now takes 10% longer than before (averaging 0.82 miliseconds more on my machine). For reference, 0.82 milliseconds makes up 0.5% of the average `processExternalized` call when using the Postgres backend.

A 0.5% impact on one of two backends seemed like a good place to stop tweaking and measuring performance impacts, but if that's too large an impact I have a couple more ideas I can try:
1. Using a single multi-row insert into `scphistory` may be faster than the multiple single row inserts `saveSCPHistory` currently performs.
2. At first glance, it seems like saving SCP history could be moved off the main thread and onto a worker thread.

# Checklist
- [x] Compiles
- [x] Ran all tests

### DIFF
--- a/src/herder/HerderImpl.cpp
+++ b/src/herder/HerderImpl.cpp
@@ -307,6 +307,18 @@ HerderImpl::processExternalized(uint64 slotIndex, StellarValue const& value,
     // save the SCP messages in the database
     if (mApp.getConfig().MODE_STORES_HISTORY_MISC)
     {
+        ZoneNamedN(updateSCPHistoryZone, "update SCP history", true);
+        if (slotIndex != 0)
+        {
+            // Save any new SCP messages received about the previous ledger.
+            // NOTE: This call uses an empty `QuorumTracker::QuorumMap` because
+            // there is no new quorum map for the previous ledger.
+            mApp.getHerderPersistence().saveSCPHistory(
+                static_cast<uint32>(slotIndex - 1),
+                getSCP().getExternalizingState(slotIndex - 1),
+                QuorumTracker::QuorumMap());
+        }
+        // Store SCP messages received about the current ledger being closed.
         mApp.getHerderPersistence().saveSCPHistory(
             static_cast<uint32>(slotIndex),
             getSCP().getExternalizingState(slotIndex),
--- a/src/herder/HerderPersistenceImpl.cpp
+++ b/src/herder/HerderPersistenceImpl.cpp
@@ -62,6 +62,11 @@ HerderPersistenceImpl::saveSCPHistory(uint32_t seq,
             st.execute(true);
         }
     }
+
+    // Prepare multi-row insert into scphistory
+    std::vector<std::string> nodeIDs;
+    std::vector<uint32_t> seqs;
+    std::vector<std::string> envelopes;
     for (auto const& e : envs)
     {
         auto const& qHash =
@@ -76,21 +81,28 @@ HerderPersistenceImpl::saveSCPHistory(uint32_t seq,
         std::string envelopeEncoded;
         envelopeEncoded = decoder::encode_b64(envelopeBytes);
 
+        nodeIDs.push_back(nodeIDStrKey);
+        seqs.push_back(seq);
+        envelopes.push_back(envelopeEncoded);
+    }
+
+    if (!envs.empty())
+    {
+        // Perform multi-row insert into scphistory
         auto prepEnv =
             db.getPreparedStatement("INSERT INTO scphistory "
                                     "(nodeid, ledgerseq, envelope) VALUES "
                                     "(:n, :l, :e)");
-
         auto& st = prepEnv.statement();
-        st.exchange(soci::use(nodeIDStrKey));
-        st.exchange(soci::use(seq));
-        st.exchange(soci::use(envelopeEncoded));
+        st.exchange(soci::use(nodeIDs, "n"));
+        st.exchange(soci::use(seqs, "l"));
+        st.exchange(soci::use(envelopes, "e"));
         st.define_and_bind();
         {
             ZoneNamedN(insertSCPHistoryZone, "insert scphistory", true);
             st.execute(true);
         }
-        if (st.get_affected_rows() != 1)
+        if (st.get_affected_rows() != envs.size())
         {
             throw std::runtime_error("Could not update data in SQL");
         }
--- a/src/herder/test/HerderTests.cpp
+++ b/src/herder/test/HerderTests.cpp
@@ -5803,3 +5803,155 @@ TEST_CASE("exclude transactions by operation type", "[herder]")
                 TransactionQueue::AddResult::ADD_STATUS_PENDING);
     }
 }
+
+// Test that Herder updates the scphistory table with additional messages from
+// ledger `n-1` when closing ledger `n`
+TEST_CASE("SCP message capture from previous ledger", "[herder]")
+{
+    constexpr uint32_t version =
+        static_cast<uint32_t>(SOROBAN_PROTOCOL_VERSION);
+
+    // Initialize simulation
+    auto networkID = sha256(getTestConfig().NETWORK_PASSPHRASE);
+    auto simulation = std::make_shared<Simulation>(
+        Simulation::OVER_LOOPBACK, networkID, [](int i) {
+            auto cfg = getTestConfig(i, Config::TESTDB_ON_DISK_SQLITE);
+            cfg.TESTING_UPGRADE_LEDGER_PROTOCOL_VERSION = version;
+            return cfg;
+        });
+
+    // Create three validators: A, B, and C
+    auto validatorAKey = SecretKey::fromSeed(sha256("validator-A"));
+    auto validatorBKey = SecretKey::fromSeed(sha256("validator-B"));
+    auto validatorCKey = SecretKey::fromSeed(sha256("validator-C"));
+
+    // Put all validators in a quorum set of threshold 2
+    SCPQuorumSet qset;
+    qset.threshold = 2;
+    qset.validators.push_back(validatorAKey.getPublicKey());
+    qset.validators.push_back(validatorBKey.getPublicKey());
+    qset.validators.push_back(validatorCKey.getPublicKey());
+
+    // Connect validators A and B, but leave C disconnected
+    auto A = simulation->addNode(validatorAKey, qset);
+    auto B = simulation->addNode(validatorBKey, qset);
+    auto C = simulation->addNode(validatorCKey, qset);
+    simulation->addPendingConnection(validatorAKey.getPublicKey(),
+                                     validatorBKey.getPublicKey());
+    simulation->startAllNodes();
+
+    // Crank A and B until they're on ledger 2
+    simulation->crankUntil(
+        [&]() {
+            return A->getLedgerManager().getLastClosedLedgerNum() == 2 &&
+                   B->getLedgerManager().getLastClosedLedgerNum() == 2;
+        },
+        4 * Herder::EXP_LEDGER_TIMESPAN_SECONDS, false);
+
+    // Check that a node's scphistory table for a given ledger has the correct
+    // number of entries of each type in `expectedTypes`
+    auto checkSCPHistoryEntries =
+        [&](Application::pointer node, uint32_t ledgerNum,
+            UnorderedMap<SCPStatementType, size_t> const& expectedTypes) {
+            // Prepare query
+            auto& db = node->getDatabase();
+            auto prep = db.getPreparedStatement(
+                "SELECT envelope FROM scphistory WHERE ledgerseq = :l");
+            auto& st = prep.statement();
+            st.exchange(soci::use(ledgerNum));
+            std::string envStr;
+            st.exchange(soci::into(envStr));
+            st.define_and_bind();
+            st.execute(false);
+
+            // Count the number of entries of each type
+            UnorderedMap<SCPStatementType, size_t> actualTypes;
+            while (st.fetch())
+            {
+                Value v;
+                decoder::decode_b64(envStr, v);
+                SCPEnvelope env;
+                xdr::xdr_from_opaque(v, env);
+                ++actualTypes[env.statement.pledges.type()];
+            }
+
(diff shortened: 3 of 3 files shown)


src/herder/HerderImpl.h
```diff
@@ -288,6 +289,10 @@ class HerderImpl : public Herder
     Application& mApp;
     LedgerManager& mLedgerManager;
 
+    // Set of nodes whose SCP messages from the previous ledger have already
+    // been stored in the scphistory database table.
+    UnorderedSet<NodeID> mPrevExternalizedEnvs;
```
marta-lokhova, line 294: This isn't _quite_ right: current implementation only records new statements from nodes we haven't heard from at all in the last consensus round. The scenario we also want to handle is when we receive _newer_ SCP messages from nodes in our quorum set. This can occur if, for example, we have CONFIRM messages from our quorum slice (which is enough to externalize that ledger), but then receive EXTERNALIZE messages from the same nodes after some time. We'd like to store those newer EXTERNALIZE messages as well, to make things easier for downstream systems (e.g. bridges). So I think we'd either want to store the diff in SCP messages, or rewrite the whole externalizing state, depending on the performance impact.

marta-lokhova (changes requested):
Thanks for putting this together! I had a question regarding another scenario we'd want to handle

src/herder/HerderImpl.h
bboston7: Ah, this was an oversight on my part, I forgot about that scenario. I'll fix this and do another round of performance testing.

bboston7: I pushed up a fix for this in 07defe8ceb992f810e47f5047ae60fd7dfb58dfe.  Weirdly enough the performance of rewriting the entire externalizing state is faster than tracking diffs. I think this is because rewriting can be done with a single `DELETE` followed by a single bulk `INSERT` (I changed `saveSCPHistory` to perform a single multi-row `INSERT` into `scphistory`, which provided a small performance improvement), while tracking diffs requires multiple `MODIFY`s and `INSERT`s.

Regardless, the performance impact of all of this isn't too bad. Compared to `master` on pubnet, the new process of saving SCP messages (doing two `saveSCPHistory` calls) takes ~3.7ms longer on Postgres and ~0.9ms longer on SQLite on my machine. In relative terms, that's 31% longer for Postgres and 50% longer for SQLite.

MonsieurNicolas:
just a drive-by message wrt diff vs rewrite all of SCP state:
* right now tier1 contains I think 23 validators but we expect this to grow by 2-3x. That being said: if this becomes a problem, we could track (as a different issue) to change how/where we record SCP messages.
* on the txset front (not sure if it matters here)
   * with more validators we're more likely to encounter more (competing) TxSets
   * with TPS increase, txsets themselves are going to be bigger

bboston7:
> * right now tier1 contains I think 23 validators but we expect this to grow by 2-3x. That being said: if this becomes a problem, we could track (as a different issue) to change how/where we record SCP messages.

Good point. I opened #4139 so we don't forget about this.

> * on the txset front (not sure if it matters here)
>   * with more validators we're more likely to encounter more (competing) TxSets
>   * with TPS increase, txsets themselves are going to be bigger

I don't think the number of competing TxSets will have much of an impact, as this only records the last externalizing message from any given validator. The only impact I could see is if we switched to the diff based approach and competing TxSets resulted in more frequent diffs. 

However, larger txsets might have an impact. I assume these txsets end up in the ballots in the `Value` field, thus making the ballots themselves larger. Is that right?

bboston7:
> However, larger txsets might have an impact. I assume these txsets end up in the ballots in the `Value` field, thus making the ballots themselves larger. Is that right?

Actually, it looks like the SCP envelopes contain only the [hash of the txset](https://github.com/stellar/stellar-xdr/blob/bb54e505f814386a3f45172e0b7e95b7badbe969/Stellar-ledger.x#L29). In that case, I don't think txset size has any impact on saving envelopes to the database.

src/herder/test/HerderTests.cpp
```diff
@@ -5803,3 +5803,117 @@ TEST_CASE("exclude transactions by operation type", "[herder]")
...
+    simulation->crankUntil(
+        [&]() {
+            return A->getLedgerManager().getLastClosedLedgerNum() == 3 &&
+                   B->getLedgerManager().getLastClosedLedgerNum() == 3;
+        },
+        2 * Herder::EXP_LEDGER_TIMESPAN_SECONDS, false);
+
+    // A and B should now have 3 entries in their scphistory table for ledger 2.
+    REQUIRE(getNumSCPHistoryEntries(A, 2) == 3);
+    REQUIRE(getNumSCPHistoryEntries(B, 2) == 3);
```
SirTyson, line 5918: Should we check that C does not have extra copies of messages? I might have missed something but I think this may be possible.

SirTyson (commented):
Looks good! Just a few nit pics and questions.

src/herder/HerderPersistenceImpl.cpp
```diff
@@ -76,21 +81,28 @@ HerderPersistenceImpl::saveSCPHistory(uint32_t seq,
...
 
+        nodeIDs.push_back(nodeIDStrKey);
+        seqs.push_back(seq);
+        envelopes.push_back(envelopeEncoded);
+    }
+
+    if (!envs.empty())
+    {
+        // Perform multi-row insert into scphistory
         auto prepEnv =
```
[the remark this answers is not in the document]
bboston7: This handles the collision case by [deleting all entries from `currentLedger` before performing the `INSERT`](https://github.com/bboston7/stellar-core/blob/07defe8ceb992f810e47f5047ae60fd7dfb58dfe/src/herder/HerderPersistenceImpl.cpp#L54). This also resolves the `CONFIRM` vs `EXTERNALIZE` issue.

src/herder/test/HerderTests.cpp
bboston7: Good idea to explicitly check this case. I added a test for this in d39400b30d1c195b1968a090fe6fb033a8752b08.

src/herder/HerderImpl.cpp
```diff
@@ -307,10 +307,28 @@ HerderImpl::processExternalized(uint64 slotIndex, StellarValue const& value,
     // save the SCP messages in the database
     if (mApp.getConfig().MODE_STORES_HISTORY_MISC)
     {
+        ZoneNamedN(updateSCPHistoryZone, "update SCP history", true);
+
+        // NOTE: Consolidating the two saveSCPHistory calls into one transaction
+        // provides modest performance increases for sqlite and insignificant
+        // performance increases for postgres.
+        soci::transaction txscope(mApp.getDatabase().getSession());
```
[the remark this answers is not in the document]
bboston7: Done in c950e1c06e44f7234e7cc1bc7e1ab7c5f98b08ea

src/herder/test/HerderTests.cpp
```diff
@@ -5803,3 +5803,117 @@ TEST_CASE("exclude transactions by operation type", "[herder]")
...
+    };
+
+    // A and B should have 2 entries in their scphistory table for ledger 2. C
+    // should have 0.
+    REQUIRE(getNumSCPHistoryEntries(A, 2) == 2);
+    REQUIRE(getNumSCPHistoryEntries(B, 2) == 2);
+    REQUIRE(getNumSCPHistoryEntries(C, 2) == 0);
+
+    // Get messages from A and B
+    HerderImpl* herderA = dynamic_cast<HerderImpl*>(&A->getHerder());
```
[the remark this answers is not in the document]
bboston7: Done in c950e1c06e44f7234e7cc1bc7e1ab7c5f98b08ea

src/herder/test/HerderTests.cpp
```diff
@@ -5803,3 +5803,126 @@ TEST_CASE("exclude transactions by operation type", "[herder]")
...
+
+    // Crank A and B until they're on ledger 2
+    simulation->crankUntil(
+        [&]() {
+            return A->getLedgerManager().getLastClosedLedgerNum() == 2 &&
+                   B->getLedgerManager().getLastClosedLedgerNum() == 2;
+        },
+        2 * Herder::EXP_LEDGER_TIMESPAN_SECONDS, false);
+
+    // Return the number of entires in a node's scphistory table for the given
```
SirTyson, line 5851: Nit: "entires"

SirTyson (approved):
Minor typo, but LGTM, thanks!

src/herder/test/HerderTests.cpp
bboston7: Fixed in 72b875763825ddd0d230e819847c625f8c04999b

bboston7:
I just rebased this on top of `master` and squashed down to a single commit

marta-lokhova (changes requested):
Thanks for the change! Left a few suggestions. Let's plan to ship this change in 20.2.0

src/herder/HerderImpl.cpp
```diff
@@ -307,10 +307,28 @@ HerderImpl::processExternalized(uint64 slotIndex, StellarValue const& value,
     // save the SCP messages in the database
     if (mApp.getConfig().MODE_STORES_HISTORY_MISC)
     {
+        ZoneNamedN(updateSCPHistoryZone, "update SCP history", true);
+
+        // NOTE: Consolidating the two saveSCPHistory calls into one transaction
+        // provides modest performance increases for sqlite and insignificant
+        // performance increases for postgres.
+        soci::transaction txScope(mApp.getDatabase().getSession());
```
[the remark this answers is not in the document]
bboston7: Done in a8139eda1f54ce59df7f1bf04b89c71d862e72fa

src/herder/test/HerderTests.cpp
```diff
@@ -5803,3 +5803,126 @@ TEST_CASE("exclude transactions by operation type", "[herder]")
...
+
+    // Crank A and B until they're on ledger 3
+    simulation->crankUntil(
+        [&]() {
+            return A->getLedgerManager().getLastClosedLedgerNum() == 3 &&
+                   B->getLedgerManager().getLastClosedLedgerNum() == 3;
+        },
+        2 * Herder::EXP_LEDGER_TIMESPAN_SECONDS, false);
+
+    // A and B should now have 3 entries in their scphistory table for ledger 2.
```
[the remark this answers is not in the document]
bboston7: Done in a8139eda1f54ce59df7f1bf04b89c71d862e72fa

marta-lokhova (approved):

marta-lokhova:
r+ a8139eda1f54ce59df7f1bf04b89c71d862e72fa

bboston7:
The test failed because it was overly sensitive to the rng seed. I fixed it by increasing some timeouts in the test and also better specifying the state of `A` and `B` to be insensitive to leader election.

marta-lokhova:
r+ 26bba350feed57eb19241f22d932eacf266a41a3

marta-lokhova:
r+ 26bba350feed57eb19241f22d932eacf266a41a3

PR #4144, opened 2024-01-17
Add eviction cycle metrics

### LINKED ISSUE #4137: Add metric for evicted entry liveUntilLedger
We should add a metric for how long an entry has lived in the BucketList past its `liveUntilLedger`. Specifically, whenever an entry is evicted, we should record the delta between its `liveUntilLedger` and the current ledger. We should report this metric as an average per eviction cycle (i.e. the average delta per entry over one complete eviction scan cycle).

SirTyson:
# Description

Resolves #4137

Adds additional metrics for eviction scans. Specifically, records the period of eviction scans in ledgers and the average amount of time it takes for an entry to get evicted after expiring.

# Checklist
- [x] Compiles
- [x] Ran all tests

### DIFF
--- a/src/bucket/Bucket.cpp
+++ b/src/bucket/Bucket.cpp
@@ -841,7 +841,8 @@ Bucket::scanForEviction(AbstractLedgerTxn& ltx, EvictionIterator& iter,
                         uint64_t& bytesToScan, uint32_t& maxEntriesToEvict,
                         uint32_t ledgerSeq,
                         medida::Counter& entriesEvictedCounter,
-                        medida::Counter& bytesScannedForEvictionCounter)
+                        medida::Counter& bytesScannedForEvictionCounter,
+                        std::optional<EvictionMetrics>& metrics)
 {
     ZoneScoped;
     if (isEmpty())
@@ -876,6 +877,7 @@ Bucket::scanForEviction(AbstractLedgerTxn& ltx, EvictionIterator& iter,
                 auto initialStreamPos = stream.pos();
 
                 auto ttlKey = getTTLKey(le);
+                uint32_t liveUntilLedger = 0;
                 auto shouldEvict = [&] {
                     auto entryLtxe = ltx.loadWithoutRecord(LedgerEntryKey(le));
                     auto ttlLtxe = ltx.loadWithoutRecord(ttlKey);
@@ -888,13 +890,21 @@ Bucket::scanForEviction(AbstractLedgerTxn& ltx, EvictionIterator& iter,
                     }
 
                     releaseAssert(ttlLtxe);
-
+                    liveUntilLedger =
+                        ttlLtxe.current().data.ttl().liveUntilLedgerSeq;
                     return !isLive(ttlLtxe.current(), ledgerSeq);
                 };
 
                 if (shouldEvict())
                 {
                     ZoneNamedN(evict, "evict entry", true);
+                    if (metrics.has_value())
+                    {
+                        ++metrics->numEntriesEvicted;
+                        metrics->evictedEntriesAgeSum +=
+                            ledgerSeq - liveUntilLedger;
+                    }
+
                     ltx.erase(ttlKey);
                     ltx.erase(LedgerEntryKey(le));
                     entriesEvictedCounter.inc();
--- a/src/bucket/BucketList.cpp
+++ b/src/bucket/BucketList.cpp
@@ -829,9 +829,7 @@ BucketList::addBatch(Application& app, uint32_t currLedger,
 void
 BucketList::scanForEviction(Application& app, AbstractLedgerTxn& ltx,
                             uint32_t ledgerSeq,
-                            medida::Counter& entriesEvictedCounter,
-                            medida::Counter& bytesScannedForEvictionCounter,
-                            medida::Counter& incompleteBucketScanCounter)
+                            BucketListEvictionCounters& counters)
 {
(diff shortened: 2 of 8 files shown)


src/bucket/Bucket.cpp
```diff
@@ -888,13 +890,21 @@ Bucket::scanForEviction(AbstractLedgerTxn& ltx, EvictionIterator& iter,
...
                     return !isLive(ttlLtxe.current(), ledgerSeq);
                 };
 
                 if (shouldEvict())
                 {
                     ZoneNamedN(evict, "evict entry", true);
+                    if (metrics.has_value())
+                    {
+                        ++metrics->numEntriesEvicted;
+                        metrics->evictedEntriesAgeSum +=
```
marta-lokhova, line 904: why do we need to sum it up like this? we could use a histrogram (that would live inside `Bucket`), and update it every time we evict a single entry instead of manually aggregating evictions.

src/bucket/BucketList.h
```diff
@@ -402,10 +402,24 @@ class BucketListDepth
     friend class testutil::BucketListDepthModifier;
 };
 
+struct EvictionMetrics
+{
+    // Evicted entry "age" is the delta between its liveUntilLedger and the
+    // ledger when the entry is actually evicted
+    uint64_t evictedEntriesAgeSum;
```
marta-lokhova, line 409: uninitialized variables

marta-lokhova: (also see my other comment regarding average age, I think you can remove this struct completely, and only keep evictionCycleStartLedger)

src/bucket/BucketList.cpp
```diff
@@ -916,6 +919,26 @@ BucketList::scanForEviction(Application& app, AbstractLedgerTxn& ltx,
...
+                            mEvictionMetrics->numEntriesEvicted == 0
+                                ? 0
+                                : mEvictionMetrics->evictedEntriesAgeSum /
+                                      mEvictionMetrics->numEntriesEvicted;
+                        averageEvictedEntryAgeCounter.set_count(averageAge);
+                    }
+
+                    // Reset metrics at beginning of new eviction cycle
+                    mEvictionMetrics = std::make_optional<EvictionMetrics>();
+                    mEvictionMetrics->evictionCycleStartLedger = ledgerSeq;
```
marta-lokhova, line 941: this won't record the first eviction scan; we could initialize evictionCycleStartLedger at the beginning of eviction scan to fix that, right?

marta-lokhova (changes requested):

src/bucket/BucketManagerImpl.cpp
```diff
@@ -904,9 +908,10 @@ BucketManagerImpl::scanForEviction(AbstractLedgerTxn& ltx, uint32_t ledgerSeq)
     if (protocolVersionStartsFrom(ltx.getHeader().ledgerVersion,
                                   SOROBAN_PROTOCOL_VERSION))
     {
-        mBucketList->scanForEviction(mApp, ltx, ledgerSeq, mEntriesEvicted,
-                                     mBytesScannedForEviction,
-                                     mIncompleteBucketScans);
+        mBucketList->scanForEviction(
+            mApp, ltx, ledgerSeq, mEntriesEvicted, mBytesScannedForEviction,
```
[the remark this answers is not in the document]
SirTyson: This is the standard interface we use for all BucketList related metrics. The BucketList itself is a very light weight class (the only member is `std::vector<BucketLevel> mLevels;`). We could do this, but we would also have to change the BucketList constructor from `BucketList()` to `BucketList(Application)`, which would be annoying for tests and a bit of a refactor.

src/bucket/Bucket.cpp
SirTyson: We actually don't want a histogram and specifically want the per-cycle aggregation. A histogram for this metric is misleading. Given the nature of the eviction scan, entries evicted in level 8 will almost certainly be "older" than entries evicted in level 6. Within a given cycle, the number of evicted entries is very noisy from level to level. We want to abstract away that noise. We only care if the average age increases cycle-to-cycle. This would be pretty involved to graph in Grafana if we use a histogram, as the cycle length itself is variable. This seemed significantly easier, as it logs the only metric we actually care about (average cycle-to-cycle) in a way that is easy to analyze in Grafana.

src/bucket/BucketList.cpp
SirTyson: We don't actually have a code path that checks if we're at the start of an eviction scan. Personally, I prefer it as is. I like that we record and re-initialize our metrics at the same point, and would rather not add a branch that will only every be called once when we upgrade to 20. Also, the first eviction cycle metrics will be pretty meaningless anyway since there will not yet be any entries eligible for eviction.

marta-lokhova: I don't feel strongly, but I thought it's possible to evict entries in the first cycle if some buckets spill faster and cause eviction scan to get reset. I'm fine with keeping the current approximation though.

src/bucket/BucketManagerImpl.cpp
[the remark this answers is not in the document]
marta-lokhova: Ok, I flagged this because this function signature is getting a bit out of hand. Could we use a struct here?

marta-lokhova:
r+ 407d8364020ec3c03c97b2a99ae9e45a8f28d243

PR #4152, opened 2024-01-21
Add test vectors for ed25519 edge cases

graydon:
This is the libsodium-testing variant of the dalek-testing test vectors in https://github.com/stellar/rs-soroban-env/pull/1331 (they behave the same way)


dmkozh:
r+ 81ded0c

PR #4162, opened 2024-01-25
Fix startup race condition

marta-lokhova:


src/main/ApplicationUtils.cpp
```diff
@@ -195,7 +195,16 @@ setupApp(Config& cfg, VirtualClock& clock, uint32_t startAtLedger,
         return nullptr;
     }
 
-    app->getLedgerManager().loadLastKnownLedger(nullptr);
+    auto loadConfig = [app]() {
+        LedgerTxn ltx(app->getLedgerTxnRoot());
+        if (protocolVersionStartsFrom(ltx.loadHeader().current().ledgerVersion,
+                                      SOROBAN_PROTOCOL_VERSION))
+        {
+            app->getLedgerManager().updateNetworkConfig(ltx);
```
dmkozh, line 203: Are we sure we should pass callback instead of loading the config after loading LCL? IIUC we do a bunch more stuff when handler is not passed to `loadLastKnownLedger`.

dmkozh:
r+ c748935

dmkozh:
r+ 574241b

PR #4183, opened 2024-02-07
add support for debug-tx-set in dump-xdr

MonsieurNicolas:
# Description

adds support for printing `debug-tx-set.xdr` files

### DIFF
--- a/src/main/dumpxdr.cpp
+++ b/src/main/dumpxdr.cpp
@@ -14,6 +14,7 @@
 #include "util/XDROperators.h"
 #include "util/XDRStream.h"
 #include "util/types.h"
+#include "xdr/Stellar-internal.h"
 #include <cereal/archives/json.hpp>
 #include <cereal/cereal.hpp>
 #include <fmt/format.h>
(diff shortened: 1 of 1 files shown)


src/main/dumpxdr.cpp
```diff
@@ -64,37 +65,53 @@ dumpstream(XDRInputFileStream& in, bool compact)
...
+            }
+            else if (m2 == "transactions")
+            {
+                dumpstream<TransactionHistoryEntry>(in, compact);
+            }
+            else if (m2 == "results")
+            {
+                dumpstream<TransactionHistoryResultEntry>(in, compact);
+            }
+            else if (m2 == "meta-debug")
```
marta-lokhova, line 99: I think just "meta" is a bit more intuitive (and less verbose)

marta-lokhova (approved):
Thanks, looks good. Just one small suggestion

src/main/dumpxdr.cpp
MonsieurNicolas: this is the file name pattern, we could rename files, but this is out of scope for this PR

PR #4194, opened 2024-02-15
fix compile error (Visual C++)

MonsieurNicolas:
missing capture in a lambda (not sure why this compiles)

### DIFF
--- a/src/herder/test/HerderTests.cpp
+++ b/src/herder/test/HerderTests.cpp
@@ -5814,7 +5814,7 @@ TEST_CASE("SCP message capture from previous ledger", "[herder]")
     // Initialize simulation
     auto networkID = sha256(getTestConfig().NETWORK_PASSPHRASE);
     auto simulation = std::make_shared<Simulation>(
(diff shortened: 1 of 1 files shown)


anupsdf:
r+ e850caa6ea147744a6bebfd35544cf5598b8217b

bboston7:
I went down a bit of a rabbit hole (partially because I wrote this code) and [the code is valid, but MSVC doesn't comply with the standard in this case](https://stackoverflow.com/questions/42610429/must-constexpr-expressions-be-captured-by-a-lambda-in-c), though it's admittedly a bit of a strange edge case.  It makes me wonder, should we have an MSVC build in our CI system to catch these things automatically?

MonsieurNicolas:
> I went down a bit of a rabbit hole (partially because I wrote this code) and [the code is valid, but MSVC doesn't comply with the standard in this case](https://stackoverflow.com/questions/42610429/must-constexpr-expressions-be-captured-by-a-lambda-in-c), though it's admittedly a bit of a strange edge case. It makes me wonder, should we have an MSVC build in our CI system to catch these things automatically?

the standard works in mysterious ways!

Issue #3846, opened 2023-07-20
tech debt: cleanup or remove transactions "stacks"


marta-lokhova:
To be done after https://github.com/stellar/stellar-core/issues/3782 is completed and rolled out on the network.

The change above makes the notion of "tx stacks" (aka chains of transactions for the same source account) obsolete. 
At a minimum, we can change the API so callers like the transaction queue and transaction limiter don't need to be aware of the tx stacks used underneath. We could also potentially simplify the underlying surge pricing ordering mechanism to remove the notion of tx stacks completely, but this might be harder due to the need to validate pre-protocol-20 transaction sets, which still contain tx stacks. This requires investigation to determine code to clean up/remove.

marta-lokhova:
As part of this cleanup, we should also remove any logic inside the TransactionQueue module that deals with chains of transactions per source account.

MonsieurNicolas:
+1 on doing this work. When we do it, we should also revisit tests as all this complexity is not needed and in some cases the intent of the test may not be valid anymore.

I don't think we need to validate pre-20 txsets: when replaying transactions from history, the only thing we need to ensure is that hashes are correct (we don't call `checkValid` on tx sets and only apply transactions).

marta-lokhova:
Bumping the priority of this issue: I think we need the cleanup to avoid maintenance burden when we implement parallelization changes.

marta-lokhova:
Note that @dmkozh's https://github.com/stellar/stellar-core/pull/4383 PR will close this issue

PR #4196, opened 2024-02-15
Simplify tx queue, remove transaction chains support

marta-lokhova:
partially resolves https://github.com/stellar/stellar-core/issues/3846

### DIFF
--- a/src/herder/TransactionQueue.cpp
+++ b/src/herder/TransactionQueue.cpp
@@ -135,56 +135,6 @@ canReplaceByFee(TransactionFrameBasePtr tx, TransactionFrameBasePtr oldTx,
     return res;
 }
 
-// This method will update iter to point to the tx with seqNum == seq if it is
-// found. It also returns a bool that will be false if it determines seq cannot
-// be added to the current queue, allowing the user of this function to make a
-// decision early if desired.
-static bool
-findBySeq(int64_t seq, std::optional<SequenceNumber const> minSeqNum,
-          TransactionQueue::TimestampedTransactions& transactions,
-          TransactionQueue::TimestampedTransactions::iterator& iter)
-{
-    int64_t firstSeq = transactions.front().mTx->getSeqNum();
-    int64_t lastSeq = transactions.back().mTx->getSeqNum();
-
-    // check if seq is too low
-    if (seq < firstSeq)
-    {
-        iter = transactions.end();
-        return false;
-    }
-
-    // check if seq is new, and if it is, if minSeqNum would make it valid
-    if (seq > lastSeq)
-    {
-        iter = transactions.end();
-        return minSeqNum ? *minSeqNum <= lastSeq : seq == lastSeq + 1;
-    }
-
-    // by this point we're expecting to find an existing transaction, and if we
-    // don't it's a gap that can't get filled
-    iter = std::lower_bound(transactions.begin(), transactions.end(), seq,
-                            [](TransactionQueue::TimestampedTx const& a,
-                               int64_t b) { return a.mTx->getSeqNum() < b; });
-
-    releaseAssert(iter == transactions.end() || iter->mTx->getSeqNum() >= seq);
-    if (iter != transactions.end() && iter->mTx->getSeqNum() != seq)
-    {
-        // found a gap
-        iter = transactions.end();
-        return false;
-    }
-    return true;
-}
-
-static bool
-findBySeq(TransactionFrameBasePtr tx,
-          TransactionQueue::TimestampedTransactions& transactions,
-          TransactionQueue::TimestampedTransactions::iterator& iter)
-{
-    return findBySeq(tx->getSeqNum(), tx->getMinSeqNum(), transactions, iter);
-}
-
 static bool
 isDuplicateTx(TransactionFrameBasePtr oldTx, TransactionFrameBasePtr newTx)
 {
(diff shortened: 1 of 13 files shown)


src/herder/TransactionQueue.cpp
```diff
@@ -752,11 +579,12 @@ TransactionQueue::ban(Transactions const& banTxs)
...
     // Group the transactions by source account and ban all the transactions
     // that are explicitly listed
-    std::map<AccountID, Transactions> transactionsByAccount;
+    std::map<AccountID, TransactionFrameBasePtr> transactionsByAccount;
     for (auto const& tx : banTxs)
     {
-        auto& transactions = transactionsByAccount[tx->getSourceID()];
-        transactions.emplace_back(tx);
+        // Must be a new transaction for this account
+        releaseAssert(
```
dmkozh, line 586: This condition is new and not trivial. Are we sure we never trigger this assertion?

src/herder/TransactionQueue.cpp
```diff
@@ -771,69 +599,31 @@ TransactionQueue::ban(Transactions const& banTxs)
         auto stateIter = mAccountStates.find(kv.first);
         if (stateIter != mAccountStates.end())
         {
-            // If there are no transactions in the queue for this source
-            // account, then there is nothing to do
-            auto& transactions = stateIter->second.mTransactions;
-            if (!transactions.empty())
+            auto transaction = stateIter->second.mTransaction;
```
dmkozh, line 602: This should probably be const reference.

src/herder/TransactionQueue.cpp
```diff
@@ -771,69 +599,31 @@ TransactionQueue::ban(Transactions const& banTxs)
...
-                        mBannedTransactionsCounter.inc();
-                    }
-                }
-                mSizeByAge[stateIter->second.mAge]->dec(transactions.end() -
-                                                        txIter);
-
-                // Drop all of the transactions, release fees (which can
-                // cause other accounts to be removed from mAccountStates),
-                // and potentially remove this account from mAccountStates.
-                // WARNING: stateIter and everything that references it may
```
dmkozh, line 813: Should we keep this warning?

src/herder/TransactionQueue.cpp
```diff
@@ -602,148 +485,92 @@ TransactionQueue::tryAdd(TransactionFrameBasePtr tx, bool submittedFromSelf)
 }
 
 void
-TransactionQueue::dropTransactions(AccountStates::iterator stateIter,
-                                   TimestampedTransactions::iterator begin,
-                                   TimestampedTransactions::iterator end)
+TransactionQueue::dropTransactions(AccountStates::iterator stateIter)
```
dmkozh, line 488: Should we name this `dropTransaction` now?

src/herder/TransactionQueue.cpp
marta-lokhova: Actually, it doesn't make things that much better, since we need to construct maps and validate that keys map to the right transactions. I left this assert as-is - I audited the code and confirmed that we only ban txs with unique source accounts.

marta-lokhova:
thanks for taking a look @dmkozh, I implemented your suggestions

src/herder/TransactionQueue.cpp
```diff
@@ -602,148 +489,92 @@ TransactionQueue::tryAdd(TransactionFrameBasePtr tx, bool submittedFromSelf)
...
-                // than the highest applied sequence number for this source
-                // account, then there is nothing to do because sequence numbers
-                // are monotonic (this shouldn't happen)
-                if (transactions.front().mTx->getSeqNum() <= kv.second)
+                // We care about matching the sequence number rather than
+                // the hash, because any transaction with a sequence number
+                // less-than-or-equal to the highest applied sequence number
+                // for this source account has either (1) been applied, or
+                // (2) become invalid.
+                if (transaction->mTx->getSeqNum() <= appliedTx->getSeqNum())
```
dmkozh, line 541: Do this condition and comment still make sense? IIUC we should only have one tx/account now, so seq num should be strictly the same, right?

src/herder/TransactionQueue.cpp
```diff
@@ -771,69 +603,33 @@ TransactionQueue::ban(Transactions const& banTxs)
         auto stateIter = mAccountStates.find(kv.first);
         if (stateIter != mAccountStates.end())
         {
-            // If there are no transactions in the queue for this source
-            // account, then there is nothing to do
-            auto& transactions = stateIter->second.mTransactions;
-            if (!transactions.empty())
+            auto const& transaction = stateIter->second.mTransaction;
+            // Only ban transactions that are actually present in the queue.
+            // Transactions with higher sequence numbers than banned
```
dmkozh, line 608: How is this possible? Do we ever ban txs after accepting new ones for the same account?

src/herder/TransactionQueue.cpp
marta-lokhova: yeah, the condition is up-to-date: we can't make assumptions about sequence numbers here, because what the node applies is not guaranteed to match what's in the queue. So here if we applied a tx with seqnum N, we need to ban all earlier txs as well.

src/herder/TransactionQueue.cpp
marta-lokhova: Similarly here, at apply time, we need to reconcile the block contents with the queue, so banned transactions might not overlap with the queue.

marta-lokhova: (note that this assumption has been there prior to this change, but in this PR the logic is rewritten a bit to simplify it)

dmkozh: Thanks for the explanation. I think this logic makes sense and should indeed be preserved.

dmkozh:
r+ c06ba8e

PR #4205, opened 2024-02-21
Bump overlay min version to 32

marta-lokhova:


anupsdf:
r+ 52e47b05f3b9aeeb58da4a279fead2b7c0b342a8

PR #4224, opened 2024-02-28
Adds poolshare index for BucketsDB

### LINKED ISSUE #4223: Improve loadPoolShareTrustLinesByAccountAndAsset in BucketListDB
Currently, `loadPoolShareTrustLinesByAccountAndAsset` is slow in BucketListDB. We should improve this.

SirTyson:
# Description

Resolves #4223

This extends `BucketIndex` to include a map of `Asset -> poolID` in order to optimize `loadPoolShareTrustLinesByAccountAndAsset`. This change improves query performance by around ~2500x and minimally impacts memory overhead. I suspect the memory increase is approximately 1.5 MB, but this is so small that I was not able to measure a change during testing. 

# Checklist
- [x] Compiles
- [x] Ran all tests

### DIFF
--- a/src/bucket/BucketIndex.h
+++ b/src/bucket/BucketIndex.h
@@ -74,7 +74,7 @@ class BucketIndex : public NonMovableOrCopyable
                                   IndividualIndex::const_iterator>;
 
     inline static const std::string DB_BACKEND_STATE = "bl";
-    inline static const uint32_t BUCKET_INDEX_VERSION = 1;
+    inline static const uint32_t BUCKET_INDEX_VERSION = 2;
 
     // Returns true if LedgerEntryType not supported by BucketListDB
     static bool typeNotSupported(LedgerEntryType t);
@@ -112,6 +112,11 @@ class BucketIndex : public NonMovableOrCopyable
     // returns nullopt
     virtual std::optional<std::pair<std::streamoff, std::streamoff>>
     getPoolshareTrustlineRange(AccountID const& accountID) const = 0;
+
+    // Return all PoolIDs that contain the given asset on either side of the
+    // pool
+    virtual std::vector<PoolID> const&
+    getPoolIDsByAsset(Asset const& asset) const = 0;
 
     // Returns page size for index. InidividualIndex returns 0 for page size
     virtual std::streamoff getPageSize() const = 0;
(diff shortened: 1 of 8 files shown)


src/bucket/BucketIndex.h
```diff
@@ -37,6 +37,9 @@ class BucketManager;
 // BucketIndex abstract interface
 class BucketIndex : public NonMovableOrCopyable
 {
+  protected:
+    static const inline std::vector<PoolID> kEmptyVec = {};
```
dmkozh, line 41: This doesn't follow the naming style. Also I don't think it needs to be a class member as it's only used in a single function (so could be defined right there).

src/bucket/BucketIndexImpl.h
```diff
@@ -45,7 +49,7 @@ template <class IndexT> class BucketIndexImpl : public BucketIndex
         void
         load(Archive& ar)
         {
-            ar(keysToOffset, filter);
+            ar(assetToPoolID, keysToOffset, filter);
```
dmkozh, line 52: Is there a risk of loading the old version with the new code (or vice versa)?

SirTyson: No, in `BucketIndex::load`, we load the version and pageSize first. These two values are always serialized first no matter the index version. Then we check the index version, and if the on-disk version number doesn't match the expected version number, we drop the on-disk index and reindex from scratch.

graydon (approved):
LGTM!

src/bucket/BucketIndexImpl.h
marta-lokhova: should we delete the outdated index right away instead of keeping both until the bucket gets cleaned up? (so that we don't accidentally use it)

marta-lokhova (approved):
LGTM!

src/bucket/BucketIndexImpl.h
SirTyson: We don't actually have both indexes persisted, but we generate a new index, then rename the new index the the same name as the old index. On linux, this deletes the old file. However, I looked up in the standard and this behavior is implementation defined, so I agree we should delete explicitly here.

marta-lokhova (approved):

PR #4226, opened 2024-03-02
Bump dependencies March 2024

MonsieurNicolas:
@SirTyson was asking me something about sqlite and I realized that we didn't bump it for a while.

This brings core up to date.

Moving to the 3.x Catch2 release seems to be more involved. We may want to do this at some point.

### DIFF
--- a/src/util/xdrquery/XDRQueryEval.cpp
+++ b/src/util/xdrquery/XDRQueryEval.cpp
@@ -472,6 +472,12 @@ FieldList::getFieldNames() const
     return names;
 }
 
+inline std::string
+format_as(const NullField&)
+{
+    return "<notset>";
+}
+
 std::string
 resultToString(ResultValueType const& result)
 {
--- a/src/util/xdrquery/XDRQueryEval.cpp
+++ b/src/util/xdrquery/XDRQueryEval.cpp
@@ -6,19 +6,6 @@
 #include "fmt/format.h"
 #include "util/xdrquery/XDRQueryError.h"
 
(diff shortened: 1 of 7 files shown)


src/util/xdrquery/XDRQueryEval.cpp
```diff
@@ -472,6 +472,12 @@ FieldList::getFieldNames() const
     return names;
 }
 
+inline std::string
+format_as(const NullField&)
```
dmkozh, line 476: What is this change for?

MonsieurNicolas: without this it won't compile with the new version of the fmt library
They fixed a bug in the way `fmt::format` works, and we're supposed to provide custom formatters for custom types

dmkozh: Interesting, do we still need the old formatter defined on line 11?

MonsieurNicolas: yes done, didn't see it

MonsieurNicolas:
@SirTyson please r+ this one when you get a chance (I think you took a dependency on it)

anupsdf:
r+ 2fbd233a324e4426a28d3dc3e8395c0263a0323c

anupsdf:
> @SirTyson please r+ this one when you get a chance (I think you took a dependency on it)

r plussing on behalf of Garand since he doesn't have r plus access yet.

Issue #4211, opened 2024-02-23
Reduce history tables for typical deployments


MonsieurNicolas:
Right now some historical tables contain:
* information needed to publish to history archives
* information needed for supporting legacy (non captive core) meta consumers

We should probably look into optimizing for the common cases
* dedicated validators on one hand
* captive-core on the other

In both cases, the "meta" does not need to be stored in the database, yet it represents 90% of the data stored in the database!

As meta represents significant stress on the database, this would help reduce churn that is triggered by the "maintenance" process (which in turn gets translated into vacuum operations in databases).

There is a `MODE_STORES_HISTORY_MISC` internal config that could be used for captive-core case, but this one disables historical tables entirely (which is not compatible with publishing to archives -- we could just disallow publish to archives from captive-core).

On validators, when archives are enabled more granular control is needed as to not store meta
* `storeTransaction` --> pass a default init `TransactionMeta` (minimal meta) or wire up as to store an empty string in the `txmeta` column
* do not call `storeTransactionFee` and `Upgrades::storeUpgradeHistory` at all

As some people may still depend on this kind of meta, there is a need to continue to emit it until fully deprecated.

graydon:
I think in general all we _need_ to use the DB for is queueing the txsets / txresultsets going into the next checkpoint. the fine-grained per-tx stuff is unnecessary, as is meta, yeah.

anupsdf:
Closing this issue since a replacement [issue](https://github.com/stellar/stellar-core-internal/issues/263) has been filed to track removal of SQL backend support.

PR #4230, opened 2024-03-04
Drop TxMeta column from TxHistory table

SirTyson:
# Description

Rebased on #4226.

This change drops `txmeta` from the `txhistory` SQL table when BucketListDB is enabled.

Unfortunately, SQLite does not support conditional column deletion, so I've added a new persistent state field indicating whether or not `txmeta` is supported. This allows us to check if the `txmeta` column exists before executing the drop SQL transaction. 

Addresses #4211.

# Checklist
- [x] Compiles
- [x] Ran all tests

### DIFF
--- a/src/database/Database.cpp
+++ b/src/database/Database.cpp
@@ -250,8 +250,51 @@ Database::upgradeToCurrentSchema()
         applySchemaUpgrade(vers);
         putSchemaVersion(vers);
     }
+
+    // While not really a schema upgrade, we need to upgrade the DB when
+    // BucketListDB is enabled.
+    if (mApp.getConfig().isUsingBucketListDB())
+    {
+        // Tx meta column no longer supported in BucketListDB
+        dropTxMetaIfExists();
+    }
+
     CLOG_INFO(Database, "DB schema is in current version");
     releaseAssert(vers == SCHEMA_VERSION);
+}
+
+void
+Database::dropTxMetaIfExists()
+{
+    int txMetaExists{};
+    std::string selectStr;
+    if (isSqlite())
+    {
+        selectStr = "SELECT EXISTS ("
+                    "SELECT 1 "
+                    "FROM pragma_table_info('txhistory') "
+                    "WHERE name = 'txmeta');";
+    }
+    else
+    {
+        selectStr = "SELECT EXISTS ("
+                    "SELECT 1 "
+                    "FROM information_schema.columns "
+                    "WHERE "
+                    "table_name = 'txhistory' AND "
+                    "column_name = 'txmeta');";
+    }
+
+    auto& st = getPreparedStatement(selectStr).statement();
+    st.exchange(soci::into(txMetaExists));
+    st.define_and_bind();
+    st.execute(true);
+
+    if (txMetaExists)
+    {
+        CLOG_INFO(Database, "Dropping txmeta column from txhistory table");
+        getSession() << "ALTER TABLE txhistory DROP COLUMN txmeta;";
+    }
 }
 
 void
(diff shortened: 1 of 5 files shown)


lib/catch.hpp
dmkozh, line 1: Was this formatted by accident?

lib/sqlite/sqlite3.h
```diff
@@ -146,9 +146,9 @@ extern "C" {
 ** [sqlite3_libversion_number()], [sqlite3_sourceid()],
 ** [sqlite_version()] and [sqlite_source_id()].
 */
-#define SQLITE_VERSION        "3.41.2"
-#define SQLITE_VERSION_NUMBER 3041002
-#define SQLITE_SOURCE_ID      "2023-03-22 11:56:21 0d1fc92f94cb6b76bffe3ec34d69cffde2924203304e8ffc4155597af0c191da"
+#define SQLITE_VERSION        "3.45.1"
```
dmkozh, line 149: Can we make a separate PR with deps update and no other logic changes?

src/util/xdrquery/XDRQueryEval.cpp
```diff
@@ -472,6 +472,12 @@ FieldList::getFieldNames() const
     return names;
 }
 
+inline std::string
+format_as(const NullField&)
```
dmkozh, line 476: What is this? Doesn't seem like it's used anywhere.

lib/sqlite/sqlite3.h
SirTyson: There is, as mentioned this is rebased on top off https://github.com/stellar/stellar-core/pull/4226. I think your comments refer to changes in that PR.

dmkozh: Sorry, missed the description somehow.

graydon (changes requested):
I believe you can reflect on the table structure directly and check to see if it has a given column. For example in sqlite it's something like this:

```
SELECT EXISTS (
  SELECT 1
  FROM pragma_table_info('txhistory')
  WHERE name = 'txmeta'
);
```

and for postgres I think it's something like this:

```
SELECT EXISTS (
  SELECT 1
  FROM information_schema.columns 
  WHERE 
    table_name = 'txhistory' AND 
    column_name = 'txmeta'
);
```

SirTyson:
> I believe you can reflect on the table structure directly and check to see if it has a given column. For example in sqlite it's something like this:
> 
> ```
> SELECT EXISTS (
>   SELECT 1
>   FROM pragma_table_info('txhistory')
>   WHERE name = 'txmeta'
> );
> ```
> 
> and for postgres I think it's something like this:
> 
> ```
> SELECT EXISTS (
>   SELECT 1
>   FROM information_schema.columns 
>   WHERE 
>     table_name = 'txhistory' AND 
>     column_name = 'txmeta'
> );
> ```

Done

src/main/ApplicationImpl.cpp
```diff
@@ -709,6 +709,8 @@ ApplicationImpl::validateAndLogConfig()
     {
         if (mConfig.isUsingBucketListDB())
         {
+            // Tx meta column no longer supported in BucketListDB
+            mPersistentState->dropTxMetaIfExists(*mDatabase);
```
MonsieurNicolas, line 713: I think you need to make this schema change in the same place than other schema changes so that people can run the `upgrade-db` (some people don't give the account that runs core sql access to update schema)

SirTyson: Done

marta-lokhova (approved):
lgtm!

marta-lokhova:
r+ 31dabd1e366e0c764a4ae87d51c883e2e771e173

PR #4237, opened 2024-03-12
Fix tx queue metrics

### LINKED ISSUE #4229: tx queue: pending txs by age counter is buggy
Looks like we (I) forgot to update `herder.pending-txs.ageX` metric and make it separate for classic vs soroban queues. As a result, currently reported values are incorrect, since both queues update/override the same metric (which may cause us to see negative values).

Note that there are a couple of other metrics shared in both queues: banned txs, arb txs, and tx confirmation delay. Those strictly increment counters or add new samples, so we report correct values. We should still split those anyway to make it less error-prone, and easier to reason about different types of traffic.

marta-lokhova:
Resolves #4229


dmkozh:
r+ d1bf10d

PR #4240, opened 2024-03-12
Stability fixes

marta-lokhova:


graydon (approved):
LGTM! Though you don't need to seed the sub-PRNG from the same seed as the global one, you can just ask the global PRNG for a pseudo-random number drawn from its stream, to use as the sub-PRNG's seed (if you want to avoid the change to Math.{cpp,h})

graydon:
r+ 8adb15c

PR #4256, opened 2024-03-21
WIP Add minimal history config option which writes an empty tx meta xdr

ThomasBrady:
# Description

Resolves github.com/stellar/stellar-core/issues/4211

# Checklist
- [ ] Compiles
- [ ] Ran all tests

### DIFF
--- a/src/ledger/LedgerManagerImpl.cpp
+++ b/src/ledger/LedgerManagerImpl.cpp
@@ -1560,6 +1560,16 @@ LedgerManagerImpl::applyTransactions(
             storeTransaction(mApp.getDatabase(), ledgerSeq, tx, tm.getXDR(),
                              txResultSet);
         }
+        else if (mApp.getConfig().MODE_STORES_HISTORY_MINIMAL)
+        {
+            auto ledgerSeq = ltx.loadHeader().current().ledgerSeq;
+            // Clear the meta as we don't want to store it.
+            tm.clearOperationMetas();
+            tm.clearTxChangesAfter();
+            releaseAssert(xdr::xdr_size(tm.getXDR()) == 0);
+            storeTransaction(mApp.getDatabase(), ledgerSeq, tx, tm.getXDR(),
+                             txResultSet);
+        }
     }
 
     mTransactionApplySucceeded.inc(txSucceeded);
--- a/src/ledger/LedgerManagerImpl.cpp
+++ b/src/ledger/LedgerManagerImpl.cpp
@@ -1563,12 +1563,10 @@ LedgerManagerImpl::applyTransactions(
         else if (mApp.getConfig().MODE_STORES_HISTORY_MINIMAL)
         {
             auto ledgerSeq = ltx.loadHeader().current().ledgerSeq;
-            // Clear the meta as we don't want to store it.
-            tm.clearOperationMetas();
-            tm.clearTxChangesAfter();
-            releaseAssert(xdr::xdr_size(tm.getXDR()) == 0);
-            storeTransaction(mApp.getDatabase(), ledgerSeq, tx, tm.getXDR(),
-                             txResultSet);
+            TransactionMetaFrame emptyTM(
+                ltx.loadHeader().current().ledgerVersion);
+            storeTransaction(mApp.getDatabase(), ledgerSeq, tx,
+                             emptyTM.getXDR(), txResultSet);
         }
     }
 
--- a/src/main/Config.h
+++ b/src/main/Config.h
@@ -335,11 +335,12 @@ class Config : public std::enable_shared_from_this<Config>
     bool EXPERIMENTAL_BUCKETLIST_DB_PERSIST_INDEX;
 
     // A config parameter that stores historical data, such as transactions,
-    // fees, and scp history in the database
+    // fees, and scp history in the database.
     bool MODE_STORES_HISTORY_MISC;
 
     // A config parameter that stores a minimal amount of historical data in the
-    // database (i.e. no meta)
+    // database (i.e. just history, no fees or meta). This is useful for
+    // validators that only want to publish to history archives.
     bool MODE_STORES_HISTORY_MINIMAL;
 
     // A config parameter that stores ledger headers in the database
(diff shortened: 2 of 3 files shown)


src/ledger/LedgerManagerImpl.cpp
```diff
@@ -1560,6 +1560,16 @@ LedgerManagerImpl::applyTransactions(
...
                              txResultSet);
         }
+        else if (mApp.getConfig().MODE_STORES_HISTORY_MINIMAL)
+        {
+            auto ledgerSeq = ltx.loadHeader().current().ledgerSeq;
+            // Clear the meta as we don't want to store it.
+            tm.clearOperationMetas();
+            tm.clearTxChangesAfter();
+            releaseAssert(xdr::xdr_size(tm.getXDR()) == 0);
+            storeTransaction(mApp.getDatabase(), ledgerSeq, tx, tm.getXDR(),
```
marta-lokhova, line 1570: Meta is pretty expensive, so it seems wasteful to first write it then immediately override it. I think we should write default-initialized TransactionMeta right away to avoid performance overhead.

src/main/Config.h
```diff
@@ -338,6 +338,10 @@ class Config : public std::enable_shared_from_this<Config>
     // fees, and scp history in the database
     bool MODE_STORES_HISTORY_MISC;
 
+    // A config parameter that stores a minimal amount of historical data in the
+    // database (i.e. no meta)
+    bool MODE_STORES_HISTORY_MINIMAL;
```
marta-lokhova, line 343: I'm not sure we need a new mode here: the proposed approach in https://github.com/stellar/stellar-core/issues/4211 recommends re-using MODE_STORES_HISTORY_MISC in captive core mode to avoid populating history tables completely (note that if we do it, we should enforce that core isn't configured to publish)

src/ledger/LedgerManagerImpl.cpp
```diff
@@ -1560,6 +1560,16 @@ LedgerManagerImpl::applyTransactions(
...
+                             txResultSet);
+        }
```
marta-lokhova, line 1572: Looks like the change is incomplete: we're still recording meta in `storeTransactionFee` and `Upgrades::storeUpgradeHistory`

marta-lokhova (changes requested):
Thanks for taking a stab at this! I left some suggestions for different modes of core, happy to discuss irl also :)

src/ledger/LedgerManagerImpl.cpp
ThomasBrady: I think `storeUpgradeHistory` and `storeTransactionFee` are only invoked when `MODE_STORES_HISTORY_MISC` is true?

src/main/Config.h
ThomasBrady: AIUI, the issue identified the two use cases for optimization, the captive core case (the requirements of which are met by the `MODE_STORES_HISTORY_MISC`)  and the validator use case (write history, but not meta)?

src/ledger/LedgerManagerImpl.cpp
ThomasBrady: Cool yeah that makes sense. I changed it to use a minimally initialized meta (with the ledger version). We still want the `tm` to be populated prior to this for streaming, right?

src/main/Config.h
ThomasBrady: cc @MonsieurNicolas

src/ledger/LedgerManagerImpl.cpp
marta-lokhova: `MODE_STORES_HISTORY_MISC` it always true on validators, so we're still storing unneeded meta at the moment

src/main/Config.h
marta-lokhova: Currently captive core still writes meta, because on typical deployments MODE_STORES_HISTORY_MISC is on by default (it's only off in in-memory mode). So to address both cases you can:
* have MODE_STORES_HISTORY_MISC off on all captive core configurations (need to verify that core isn't configured to publish history in this case and throw if it is)
* for validators, we can update storeTransaction to write empty meta (this is already the case in the PR), and not call `storeUpgradeHistory` and `storeTransactionFee`.
* For all other configurations, we need to continue writing tx meta, and probably add a deprecation warning on startup

marta-lokhova: Btw, looks like for the first bullet we already ensure that publish should be disabled if history is not stored: https://github.com/stellar/stellar-core/blob/97138128b4fc02cf296a07acc5d7145557a27782/src/main/ApplicationImpl.cpp#L770

ThomasBrady: > have MODE_STORES_HISTORY_MISC off on all captive core configurations (need to verify that core isn't configured to publish history in this case and throw if it is)

This is controlled by horizon / validator operators right? 

> for validators, we can update storeTransaction to write empty meta (this is already the case in the PR), and not call storeUpgradeHistory and storeTransactionFee.

I'm a little confused how this is exposed, these are recommendations for configuration rather than sets of configs we apply right? Ie. we would "recommend" captive core sets MODE_STORES_HISTORY_MISC = false, and that validators set `MODE_STORES_HISTORY_MINIMAL= true`. 

If that is the case, It could be helpful to provide an opinionated "default validator options" etc as there are a lot of knobs to turn.

ThomasBrady: Yeah thanks that's makes much more sense.  I was thinking it would take some attention to detail for operators to get this right so I'm glad we have higher level configs that result in these being set.

SirTyson:
Thanks for working on this Tom, and sorry I'm a little late in looking at this, but are these changes necessary? Given that the plan is to switch to BucketListDB by default in the next release and merge https://github.com/stellar/stellar-core/pull/4230, which drops the txMeta column entirely when BucketListDB is enabled, is there a strong reason to write empty meta to SQL nodes if we (hopefully) will have very few SQL nodes in this release?

SirTyson:
Closing, as this is covered by https://github.com/stellar/stellar-core/pull/4230

Issue #3781, opened 2023-06-14
Make flow control in bytes required


marta-lokhova:
Flow control in bytes must be required before the network upgrades to protocol 20.

For this to work correctly, we need to complete https://github.com/stellar/stellar-core/issues/3746 first

marta-lokhova:
reminder to also gracefully deprecate "enable flow control in bytes" config flag.

PR #4258, opened 2024-03-21
add an option to run overlay operations in the background

marta-lokhova:
First stab at running some of overlay processing in the background. At a high level, the change splits processing of raw inbound/outbound data and higher level business logic of the application. The rationale for the split is that this way we actually get a more-or-less clear cut of responsibilities between overlay and the rest of the application (SCP, tx queue, etc). 

In this change the following is moved to the background: socket reads/writes, XDR decoding, HMAC verification, signature verification of SCP messages, tx hashing, flow control outbound broadcasting. Now that we have BL snapshotting available, we can go further and do transaction validation in the background, and possibly exit early to avoid additional burden on the scheduler and the tx queue. For the sake of doing things iteratively, I'd rather do that in a subsequent PR as this change is already quite big.

There are a couple of design choices in this PR that attempt to de-risk it as much as possible. For simplicity, there's a single "overlay lock" and a single "overlay thread". Single lock simplifies synchronization and removes a whole class of deadlock bugs. A single thread seems good enough for now, as our traffic patterns are fairly heavily throttled, so most of the time we're actually idle, but periodically we receive bursts of data that main thread needs to process as fast as possible. So the value of many threads is not clear at the moment given that we're still bottle-necked by the main thread, which runs SCP that keeps the node in sync.

The code is structured such that everything before we get to a valid StellarMessage is managed by the background thread, and the business logic is done on the main thread. When main thread is done processing, some of the outbound processing, like flow control + asio queuing, is also done in the background.

I wanted to get the initial version out to get review going and kick-off discussions on de-risking these changes further, as well as minimizing the impact when the experimental flag is off. 

Resolves https://github.com/stellar/stellar-core/issues/4320
Resolves https://github.com/stellar/stellar-core/issues/4309


src/simulation/LoadGenerator.cpp
```diff
@@ -705,18 +706,27 @@ LoadGenerator::generateLoad(GeneratedLoadConfig cfg)
...
-                        uint32_t const keyOverhead = 40;
-                        auto maxWasmSize =
-                            std::min(mApp.getLedgerManager()
-                                         .getSorobanNetworkConfig()
-                                         .maxContractSizeBytes(),
-                                     static_cast<uint32>(maxPerTx.getVal(
-                                         Resource::Type::WRITE_BYTES)) -
-                                         keyOverhead);
-                        wasmSize = rand_uniform<uint32_t>(1, maxWasmSize);
+                        auto const& appCfg = mApp.getConfig();
```
marta-lokhova, line 709: Note that this change is obsolete since now we have https://github.com/stellar/stellar-core/pull/4248, I'll update this PR.

src/overlay/Peer.cpp
```diff
@@ -388,6 +424,10 @@ Peer::shutdown()
 void
 Peer::connectHandler(asio::error_code const& error)
 {
+    // Thread-safe
+    std::lock_guard<std::recursive_mutex> guard(
+        mOverlayManager.getOverlayManagerMutex());
+
```
bboston7, line 430: Why does this function need the `OverlayManager`'s mutex?

src/overlay/TCPPeer.h
```diff
@@ -17,6 +17,9 @@ namespace stellar
 {
 
 static auto const MAX_UNAUTH_MESSAGE_SIZE = 0x1000;
+// action quee name must be uniform across all postOnMainThread calls to ensure
```
bboston7, line 20:
```suggestion
// action queue name must be uniform across all postOnMainThread calls to ensure
```

src/overlay/TCPPeer.cpp
```diff
@@ -697,12 +752,29 @@ TCPPeer::readBodyHandler(asio::error_code const& error,
...
+{
+    // Write queue can only be accessed by the overlay thread
+    releaseAssert(!threadIsMain() || !mUseBackgroundThread);
+    return mWriteQueue;
+}
+
+std::vector<asio::const_buffer>&
+TCPPeer::OverlayThreadWriteQueue::getWriteBuffers()
+{
+    // Write queue can only be accessed by the overlay thread
```
bboston7, line 766:
```suggestion
    // Write buffers can only be accessed by the overlay thread
```

bboston7 (commented):
This is an exciting change! It was fun to think about possible interleavings with the backgrounded overlay thread.

src/overlay/Peer.cpp
marta-lokhova: A few things are happening in this function: we're updating peer's state, or triggering peer drop. Both actions touch data that's read and written by main and background threads. For example, dropping a connection can be triggered from background (like if reading bytes from the wire failed) or from main thread (if we received a valid message, but a higher level processor decided that peer is not following the protocol)

src/overlay/FlowControl.cpp
```diff
@@ -108,15 +123,14 @@ void
...
 
     if (msg.type() == SEND_MORE || msg.type() == SEND_MORE_EXTENDED)
     {
         if (mNoOutboundCapacity)
         {
-            mApp.getOverlayManager()
-                .getOverlayMetrics()
-                .mConnectionFloodThrottle.Update(mApp.getClock().now() -
-                                                 *mNoOutboundCapacity);
+            mOverlayManager.getOverlayMetrics().mConnectionFloodThrottle.Update(
```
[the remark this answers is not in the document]
marta-lokhova: I did a bit of refactoring and now we just store OverlayMetrics reference directly inside this class. I didn't use a mutex since OverlayMetrics is just a wrapper around a bunch of medida metrics (the metrics library we use), and those are thread-safe.

src/overlay/OverlayManagerImpl.h
```diff
@@ -63,6 +63,10 @@ class OverlayManagerImpl : public OverlayManager
 
         std::vector<Peer::pointer> mPending;
         std::map<NodeID, Peer::pointer> mAuthenticated;
+        // TODO: keep dropped peers alive, just so overlay thread can still
+        // safely perform delayed shutdown, etc; when overlay thread is done and
+        // main is the only user left, release all refeences
```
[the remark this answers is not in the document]
marta-lokhova: Fixed!

src/main/ApplicationImpl.cpp
```diff
@@ -904,19 +918,27 @@ ApplicationImpl::joinAllThreads()
...
+    }
+
+    LOG_INFO(DEFAULT_LOG, "Joining {} worker threads", mWorkerThreads.size());
     for (auto& w : mWorkerThreads)
     {
         w.join();
     }
-    LOG_DEBUG(DEFAULT_LOG, "Joined all {} threads", mWorkerThreads.size());
+    LOG_INFO(DEFAULT_LOG, "Joining the overlay thread");
+    mOverlayThread->join();
```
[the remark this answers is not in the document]
marta-lokhova: Good point! Looks like this interface has been around for a while, but I agree it's not safe. Updated this function to be protected.

src/overlay/TCPPeer.h
marta-lokhova: This code went away after the latest refactor, so I'm resolving this.

SirTyson (changes requested):
Thanks for putting this together! Definitely a tricky change but I'm looking forward to parallel core! That being said, I don't think we should merge this as is. I haven't finished reviewing this quite yet, but I got particularly stuck on the `Peer` classes.

I think currently the `Peer` interface is too difficult to reason about and is error prone. For example, we have two overloads of `recvMessage`, one that asserts it's the main thread and the other which asserts it's the background thread. We also have multiple ways to get to the same object depending on if a given function is calling from a main thread or background thread context. For example, we have a const copy of `Config` that the background thread functions use, while main thread functions just call `mApp.getConfig()`. It's fairly implicit which functions should be able to access `mApp` and which functions should not. Finally, there's some functions used in both a main thread and background context, like `drop`, making the boundary even muddier.

Personally, I don't think we should merge this change without some stronger division between threadsafe and non-threadsafe functions in `Peer`. This is my first venture into the guts of overlay, so some of the confusion may be in part because I'm learning this part of the codebase. That being said, while unfortunately it's a fairly significant amount of work, I feel like a refactor of `Peer` may be necessary. If we decide not to refactor after all that's fine, but I will need to study a bit more and do a second pass to finish the review :).

src/overlay/Peer.cpp
bboston7: That makes sense, thanks!

src/overlay/TCPPeer.h
```diff
@@ -102,6 +119,8 @@ class TCPPeer : public Peer
 
     virtual ~TCPPeer();
 
+    // Thread-safe peer drop (can be initiated from either main or overlay
+    // thread)
     virtual void drop(std::string const& reason, DropDirection dropDirection,
```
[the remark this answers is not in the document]
marta-lokhova: Drop can be initiated from either thread, because there are different reasons for dropping a connection. E.g. we may encounter a super low-level ASIO error from background, or main thread deciding that peer doesn't follow the protocol at that (higher) layer of abstraction. That being said, in the updated version there is no need to synchronize peer lists, and the responsibilities are separated such that peer list is always handled on main, while socket shutdown is handled in the background.

src/overlay/OverlayManagerImpl.cpp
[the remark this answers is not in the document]
marta-lokhova: I agree it should be static, but given that data safety isn't a concern here anymore (the overlay lock was removed in favor of more granular locking), let's make this change later as to not further expand the scope of this PR.

src/overlay/FlowControl.cpp
```diff
@@ -19,25 +19,58 @@ namespace stellar
...
+FlowControl::FlowControl(Application& app, bool useBackgroundThread)
+    : mConfig(app.getConfig())
+    , mUseBackgroundThread(useBackgroundThread)
+    , mOverlayMetrics(app.getOverlayManager().getOverlayMetrics())
+    , mOverlayManager(app.getOverlayManager())
     , mNoOutboundCapacity(
           std::make_optional<VirtualClock::time_point>(app.getClock().now()))
+    , mFlowControlCapacity(std::make_shared<FlowControlMessageCapacity>(
+          app.getConfig(), mNodeID))
+    , mGetNowCb([&app]() { return app.getClock().now(); })
```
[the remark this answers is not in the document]
marta-lokhova: Done!

src/main/ApplicationImpl.cpp
```diff
@@ -143,6 +151,8 @@ ApplicationImpl::ApplicationImpl(VirtualClock& clock, Config const& cfg)
         }};
         mWorkerThreads.emplace_back(std::move(thread));
     }
+
+    mOverlayThread = std::thread{[this]() { mHighPriorityIOContext.run(); }};
```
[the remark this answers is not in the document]
marta-lokhova: Good call, done.

src/overlay/FlowControl.h
```diff
@@ -54,7 +72,12 @@ class FlowControl
     std::shared_ptr<FlowControlCapacity> mFlowControlCapacity;
     std::shared_ptr<FlowControlByteCapacity> mFlowControlBytesCapacity;
 
-    Application& mApp;
+    OverlayMetrics& mOverlayMetrics;
+    OverlayManager& mOverlayManager;
+    Config const mConfig;
```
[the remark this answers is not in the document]
marta-lokhova: In general, in a multi-threaded environment, I prefer a copy to a reference to avoid auditing the code for thread-safety and prevent footguns. This works when performance isn't an issue of course: in this case I think it's fine since Config is a tiny const object stored in each peer.

I'd ideally like to do the same for OverlayMetrics (e.g. deal with copies and do message passing of sorts), but that's more involved, so maybe as a follow-up.

src/main/ApplicationImpl.cpp
```diff
@@ -83,6 +83,12 @@ ApplicationImpl::ApplicationImpl(VirtualClock& clock, Config const& cfg)
     , mConfig(cfg)
     , mWorkerIOContext(mConfig.WORKER_THREADS)
     , mWork(std::make_unique<asio::io_context::work>(mWorkerIOContext))
+    // start with 1 thread for simplicity, as there isn't enough overlay work to
+    // consume one whole CPU
+    , mHighPriorityIOContext(
+          !!mConfig.EXPERIMENTAL_BACKGROUND_OVERLAY_PROCESSING)
+    , mHighPriorityWork(
+          std::make_unique<asio::io_context::work>(mHighPriorityIOContext))
```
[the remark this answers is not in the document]
marta-lokhova: Yep, updated.

src/main/Config.cpp
```diff
@@ -159,6 +159,7 @@ Config::Config() : NODE_SEED(SecretKey::random())
     CATCHUP_COMPLETE = false;
     CATCHUP_RECENT = 0;
     EXPERIMENTAL_PRECAUTION_DELAY_META = false;
+    EXPERIMENTAL_BACKGROUND_OVERLAY_PROCESSING = true;
```
marta-lokhova, line 162: This is temporary for testing, as initially this flag should be `false` by default.

marta-lokhova:
Thanks all for the feedback! I pushed a refreshed version of the PR. Working through a few remaining items, like new unit tests and a small remaining todo to make peer's string address thread-safe, but here's an overview of the updates:
* Updates to peer API: explicitly require most public functions to be main-thread only, audit the rest for thread-safety. 
* Synchronize peer state with an explicit fine-grained lock, remove global locking to make it easier to reason about synchronization. As part of this, add getters and setters for Peer’s state, that all require a lock guard, hinting to the caller that the lock is needed. Hide any other access to state outside of peer. If callers need to access it, they can pass a lambda and let Peer handle the synchronization.
* Flow control: remove dependency on Peer completely by removing the send callback. The goal is to prevent flow control from ever calling any method in Peer. This is needed since we have two locks now: Peer’s state lock and flow control lock, they have to be ordered, so the order of locking should always be peer lock followed by flow control.  This way we can guarantee the two new locks never deadlock. 
* Get rid of any locking at the peer lists layer: the problem with synchronizing peer lists is that it bleeds into too many subsystems past overlay, so it's hard to reason about and review. The updated design avoids the need to synchronize peer lists completely: only the main thread maintains peer lists and officially "drops" peers. The background thread asynchronously watches for any state changes, and closes the socket. Note that this pattern already exists in master today, where we delay the actual socket shutdown on the main thread, while the Peer is dropped from the OverlayManager point of view, so I'm just re-using it but in the background.

src/overlay/TCPPeer.cpp
```diff
@@ -46,20 +48,38 @@ TCPPeer::TCPPeer(Application& app, Peer::PeerRole role,
...
-        endpoint, [result](asio::error_code const& error) {
-            // Handler is invoked asynchronously, so check if connection got
-            // dropped in the meantime
-            if (result->shouldAbort())
+        endpoint, [weak = std::weak_ptr<TCPPeer>(result),
+                   address](asio::error_code const& error) {
+            auto result = weak.lock();
+            if (!result)
+            {
+                // If peer was rejected was reasons like no pending slots
```
SirTyson, line 71: Typo: for reasons

src/overlay/TCPPeer.h
```diff
@@ -26,15 +28,80 @@ class TCPPeer : public Peer
     static constexpr size_t BUFSZ = 0x40000; // 256KB
 
   private:
+    // Helper class which provides invariance for various data structures;
+    // Each method should be called from the same thread
+    class ThreadRestrictedVars
```
SirTyson, line 33: This encapsualtion is syuper helpful!

src/overlay/TCPPeer.h
```diff
@@ -26,15 +28,80 @@ class TCPPeer : public Peer
...
+            mWriting = value;
+        }
+        bool
+        socketShutdownScheduled() const
+        {
+            releaseAssert(threadIsMain());
+            return mSocketShutdownScheduled;
+        }
+        void
+        scheduleSocketShutdown(bool value)
```
SirTyson, line 91: Remove `bool value`. This should only every be called once anyway and `true` is the only valid value, so remove parameter.

src/overlay/TCPPeer.cpp
```diff
@@ -173,105 +218,81 @@ TCPPeer::~TCPPeer()
...
 void
 TCPPeer::messageSender()
 {
     ZoneScoped;
-    releaseAssert(threadIsMain());
+    releaseAssert(!threadIsMain() || !useBackgroundThread());
 
     // if nothing to do, mark progress and return.
-    if (mWriteQueue.empty())
+    if (mThreadVars.getWriteQueue().empty())
```
SirTyson, line 292: Is this labmda being called on the main thread or background thread here?

SirTyson (commented):
Thanks for doing all the refactoring work, this is significantly more clear and easier to follow! I left a few nit pics and and questions, and still probably want to walk through it together to make sure I understand everything before signing off, but it's looking good!

src/overlay/TCPPeer.cpp
```diff
@@ -588,30 +625,33 @@ TCPPeer::startRead()
...
     length &= 0x7f; // clear the XDR 'continuation' bit
     length <<= 8;
-    length |= mIncomingHeader[1];
+    length |= mThreadVars.getIncomingHeader()[1];
     length <<= 8;
-    length |= mIncomingHeader[2];
+    length |= mThreadVars.getIncomingHeader()[2];
     length <<= 8;
-    length |= mIncomingHeader[3];
+    length |= mThreadVars.getIncomingHeader()[3];
```
bboston7, line 646: I know this was here before, but these unchecked accesses are little sketchy. In practice `mThreadVars.getIncomingHeader()` is always resized to be at least `4` elements large before calling this function, but it seems a bit error prone to me. What do you think of adding an assert that the vector is large enough? Or if there are performance concerns there, at least a comment in the header file stating that `mThreadVars.getIncomingHeader.size()` must be at least `4`?

src/overlay/test/LoopbackPeer.cpp
```diff
@@ -273,13 +273,22 @@ LoopbackPeer::recvMessage(xdr::msg_ptr const& msg)
     catch (xdr::xdr_runtime_error& e)
     {
         CLOG_ERROR(Overlay, "received corrupt xdr::msg_ptr {}", e.what());
-        drop("received corrupted message",
-             Peer::DropDirection::WE_DROPPED_REMOTE,
-             Peer::DropMode::IGNORE_WRITE_QUEUE);
+        drop("git received corrupted message",
```
bboston7, line 276: I think the `git` in this line is unintentional?

bboston7 (commented):
This is looking great! I just had a few small comments.

src/overlay/Peer.cpp
```diff
@@ -835,8 +1001,34 @@ Peer::recvMessage(StellarMessage const& stellarMsg)
...
+    type = isAuthenticated(guard) ? type : Scheduler::ActionType::NORMAL_ACTION;
+    mAppConnector.postOnMainThread(
+        [self = shared_from_this(), msgTracker]() {
+            self->recvMessage(msgTracker);
+        },
+        std::move(queueName), type);
+}
+
+void
+Peer::recvMessage(std::shared_ptr<MsgCapacityTracker> msgTracker)
```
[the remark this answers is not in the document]
marta-lokhova: I agree that'd be ideal. Unfortunately there was an underlying asio issue in passing data between IO services, where it needed to be copyable. I also had to remove the destructor, because asio doesn't provide any guarantees regarding when referenced lambdas are destroyed, so I didn't want to rely on RAII here.

src/overlay/FlowControl.h
```diff
@@ -54,6 +56,8 @@ class FlowControl
     size_t mDemandQueueTxHashCount{0};
     size_t mTxQueueByteCount{0};
 
+    // Mutex to synchronize flow control state
+    std::recursive_mutex mutable mFlowControlMutex;
```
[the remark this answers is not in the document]
marta-lokhova: From the correctness perspective, there's no requirement to hold peer lock before acquiring the flow control lock. They're logically different locks concerned with different pieces of shared data. So I'd prefer to avoid adding an unnecessary dependency here.

src/overlay/TCPPeer.cpp
marta-lokhova: Everything inside `messageSender` should be executed in the background

src/util/Timer.h
```diff
@@ -171,7 +173,7 @@ class VirtualClock
     size_t advanceToNow();
 
     // timer should be last to ensure it gets destroyed first
-    asio::basic_waitable_timer<std::chrono::steady_clock> mRealTimer;
+    RealSteadyTimer mRealTimer;
```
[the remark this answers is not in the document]
marta-lokhova: That's right! asio::steady_timer is just a typedef for basic_waitable_timer\<chrono::steady_clock\>

src/overlay/OverlayAppConnector.h
```diff
@@ -30,6 +31,8 @@ class OverlayAppConnector
     void postOnMainThread(
```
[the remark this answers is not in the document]
marta-lokhova: Sure, done.

src/overlay/Peer.h
```diff
@@ -173,45 +173,77 @@ class Peer : public std::enable_shared_from_this<Peer>,
...
+        // Since MsgCapacityTracker is passed around between thread and stored
+        // by ASIO in lambdas, do not rely on RAII to trigger completion.
+        // Instead, explicitly call `finish`.
+        void finish();
     };
 
     OverlayAppConnector mAppConnector;
 
     Hash const mNetworkID;
     std::shared_ptr<FlowControl> mFlowControl;
```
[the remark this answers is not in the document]
marta-lokhova: Good idea! there are some test dependencies sadly, which override mFlowControl. I can fix this in a follow-up though after this PR lands.

src/overlay/Peer.h
```diff
@@ -173,45 +167,78 @@ class Peer : public std::enable_shared_from_this<Peer>,
...
+                  PeerState newState);
+    bool shouldAbort(std::lock_guard<std::recursive_mutex>& stateGuard) const;
+    void shutdownAndRemovePeer(std::string const& reason,
+                               DropDirection dropDirection);
+
+    // Subclasses should only use these methods to load and modify peer state,
+    // signatured hint that callers need to grab mStateMutex.
+    PeerState
+    getState(std::lock_guard<std::recursive_mutex>& stateGuard) const
+    {
```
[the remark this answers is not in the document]
marta-lokhova: Good idea, done.

src/overlay/TCPPeer.cpp
marta-lokhova: Done!

src/overlay/test/LoopbackPeer.cpp
marta-lokhova: oops, typo. fixed.

src/main/ApplicationImpl.cpp
```diff
@@ -170,6 +181,11 @@ ApplicationImpl::ApplicationImpl(VirtualClock& clock, Config const& cfg)
         }};
         mWorkerThreads.emplace_back(std::move(thread));
     }
+
+    if (mConfig.EXPERIMENTAL_BACKGROUND_OVERLAY_PROCESSING)
+    {
+        mOverlayThread = std::thread{[this]() { mOverlayIOContext->run(); }};
```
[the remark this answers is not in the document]
marta-lokhova: As overlay processing is fairly high-priority and time-sensitive, I didn't modify priority. Added a comment though.

src/simulation/Simulation.cpp
```diff
@@ -40,20 +40,35 @@ Simulation::Simulation(Mode mode, Hash const& networkID, ConfigGen confGen,
...
 }
 
 Simulation::~Simulation()
 {
     // kills all connections
     mLoopbackConnections.clear();
+
+    // shutdown overlay on all nodes first to ensure socket closure is properly
+    // triggered in TCP mode. This shouldn't normally be needed, but because in
+    // Simulation we shutdown nodes quickly before they
```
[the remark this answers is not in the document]
marta-lokhova: Good catch! Actually, I realized this extra shutdown logic isn't needed at all anymore, so I removed it.

src/main/Config.cpp
graydon, line 162: I think this should be false, no?

```suggestion
    EXPERIMENTAL_BACKGROUND_OVERLAY_PROCESSING = false;
```

src/overlay/Peer.cpp
```diff
@@ -233,39 +245,70 @@ Peer::startRecurrentTimer()
...
+                            DropDirection dropDirection)
+{
+    releaseAssert(threadIsMain());
+    RECURSIVE_LOCK_GUARD(mStateMutex, guard);
+    auto state = getState(guard);
+
+    if (state != GOT_AUTH)
+    {
+        CLOG_DEBUG(Overlay,
+                   "TCPPeer::drop {} in state {} we called:{}, reason: {}",
```
graydon, line 264:
```suggestion
                   "Dropping peer {} with state {}, role {}, reason {}",
```

src/overlay/Peer.cpp
```diff
@@ -233,39 +245,70 @@ Peer::startRecurrentTimer()
...
+                   toString(), format_as(state), format_as(mRole), reason);
+    }
+    else if (dropDirection == Peer::DropDirection::WE_DROPPED_REMOTE)
+    {
+        CLOG_INFO(Overlay, "Dropping peer {}, reason {}", toString(), reason);
+    }
+    else
+    {
+        CLOG_INFO(Overlay, "peer {} dropped us, reason {}", toString(), reason);
```
graydon, line 273:
```suggestion
        CLOG_INFO(Overlay, "Peer {} dropped us, reason {}", toString(), reason);
```

src/overlay/Peer.cpp
```diff
@@ -389,25 +434,52 @@ Peer::clearBelow(uint32_t seq)
...
+             Peer::DropDirection::WE_DROPPED_REMOTE);
     }
     else
     {
-        CLOG_DEBUG(Overlay, "Async connect to {}", toString());
         connected();
-        setState(CONNECTED);
-        sendHello();
+        setState(guard, CONNECTED);
+        // Always send auth from main thread
```
graydon, line 447:
```suggestion
        // Always send hello from main thread
```

src/overlay/Peer.cpp
```diff
@@ -389,25 +434,52 @@ Peer::clearBelow(uint32_t seq)
...
+        if (useBackgroundThread())
+        {
+            mAppConnector.postOnMainThread(
+                [self = shared_from_this()]() { self->sendHello(); },
+                "Peer::connectHandler sendAuth");
```
graydon, line 452:
```suggestion
                "Peer::connectHandler sendHello");
```

src/overlay/Peer.cpp
```diff
@@ -699,102 +794,168 @@ Peer::sendMessage(std::shared_ptr<StellarMessage const> msg, bool log)
...
+        amsg.v0().message = *msg;
+        if (msg->type() != HELLO && msg->type() != ERROR_MSG)
+        {
+            ZoneNamedN(hmacZone, "message HMAC", true);
+            amsg.v0().sequence = self->mSendMacSeq;
+            amsg.v0().mac =
+                hmacSha256(self->mSendMacKey,
+                           xdr::xdr_to_opaque(self->mSendMacSeq.load(), *msg));
+            self->mSendMacSeq++;
+        }
```
graydon, line 824: This is perhaps a bit minor since I _think_ this is the only code path touching the MAC sequence number and I _think_ there's a level of protocol-handshake-enforced mutual exclusion between potential threads that can wind up here, but this access pattern looks like it's supposed to be logically atomic and as-written it's more of a race. Here's a sketch of a more-atomic usage:

```suggestion
            auto seq = self->mSendMacSeq.fetch_add(1) + 1;
            amsg.v0().sequence = seq;
            amsg.v0().mac =
                hmacSha256(self->mSendMacKey,
                           xdr::xdr_to_opaque(seq, *msg));
        }
```

src/overlay/Peer.cpp
```diff
@@ -699,102 +794,168 @@ Peer::sendMessage(std::shared_ptr<StellarMessage const> msg, bool log)
...
                 msg.v0().mac, mRecvMacKey,
                 xdr::xdr_to_opaque(msg.v0().sequence, msg.v0().message)))
         {
-            ++mRecvMacSeq;
-            sendErrorAndDrop(ERR_AUTH, "unexpected MAC",
-                             DropMode::IGNORE_WRITE_QUEUE);
-            return;
+            errorMsg = "unexpected MAC";
         }
         ++mRecvMacSeq;
```
graydon, line 911: This is also a logical race. Here's a sketch of a less-racing one:

```suggestion
        auto seq = mRecvMacKey.fetch_add(1);
        if (!hmacSha256Verify(
                msg.v0().mac, seq,
                xdr::xdr_to_opaque(msg.v0().sequence, msg.v0().message)))
        {
            errorMsg = "unexpected MAC";
        }
```

graydon, line 911: Also a logical race. A sketch of a less-racy version:

```suggestion
        auto seq = mRecvMacSeq.fetch_add(1);
        if (msg.v0().sequence != seq)
        {
            errorMsg = "unexpected auth sequence";
        }

        if (!hmacSha256Verify(
                msg.v0().mac, seq,
                xdr::xdr_to_opaque(msg.v0().sequence, msg.v0().message)))
        {
            errorMsg = "unexpected MAC";
        }
```

src/overlay/Peer.h
```diff
@@ -171,47 +165,84 @@ class Peer : public std::enable_shared_from_this<Peer>,
...
+    void shutdownAndRemovePeer(std::string const& reason,
+                               DropDirection dropDirection);
+
+    // Subclasses should only use these methods to load and modify peer state,
+    // signatured hint that callers need to grab mStateMutex.
+    PeerState
+    getState(LockGuard const& stateGuard) const
+    {
+        return mState;
+    }
```
graydon, line 229: Small suggestion here to reorganize the signatures and comment. Also it seems to me odd that 2/3 of the methods are implemented out-of-line -- maybe make the 3rd also out-of-line too, for symmetry? It's not like it's a performance bottleneck.

```suggestion
    void shutdownAndRemovePeer(std::string const& reason,
                               DropDirection dropDirection);

    // Subclasses should only use these methods to access peer state;
    // they all take a LockGuard that should be holding mStateMutex,
    // but do not lock that mutex themselves (to allow atomic
    // read-modify-write cycles or similar patterns in callers).

    bool shouldAbort(LockGuard const& stateGuard) const;
    void setState(LockGuard const& stateGuard, PeerState newState);
    PeerState
    getState(LockGuard const& stateGuard) const
    {
        return mState;
    }

```

src/overlay/Peer.h
```diff
@@ -171,47 +165,84 @@ class Peer : public std::enable_shared_from_this<Peer>,
...
+    void maybeExecuteInBackground(std::string const& jobName,
+                                  std::function<void(std::shared_ptr<Peer>)> f);
+    VirtualTimer&
+    getRecurrentTimer()
+    {
+        releaseAssert(threadIsMain());
+        return mRecurringTimer;
+    }
+
   private:
```
graydon, line 246:
```suggestion

  // NB: Everything below is private to minimize the chance that subclasses with methods
  // running on background threads might access this unsynchronized state. All methods
  // that access this private state should assert that they are running on the main thread.
  private:
```

src/overlay/Peer.h
```diff
@@ -171,47 +165,84 @@ class Peer : public std::enable_shared_from_this<Peer>,
   protected:
```
graydon, line 165:
```suggestion
  // NB: all Peer's protected state should have some synchronization mechanisms on it:
  // either be const, or atomic, or associated with a lock, or hidden in a helper class
  // that does one of the same. Some methods in subclasses of Peer will run on a background
  // thread, so may try to access the protected state.
  protected:
```

src/overlay/Peer.h
```diff
@@ -171,47 +165,84 @@ class Peer : public std::enable_shared_from_this<Peer>,
...
+#endif
+
     HmacSha256Key mSendMacKey;
     HmacSha256Key mRecvMacKey;
-    uint64_t mSendMacSeq{0};
-    uint64_t mRecvMacSeq{0};
-    VirtualTimer mRecurringTimer;
-
+    std::atomic<uint64_t> mSendMacSeq{0};
+    std::atomic<uint64_t> mRecvMacSeq{0};
```
graydon, line 204: I notice the mSendMacKey and mRecvMacKey aren't actually synchronized. Of course there's a sort of protocol-enforced boundary where they're only ever modified before we start running anything on a background thread, but that's an implicit contract it'd be easy to break with future refactoring. I wonder if it makes sense to try factoring-out the HMAC machinery into a helper class with its own lock, similar to the flow control class?

src/overlay/FlowControl.h
```diff
@@ -160,10 +172,7 @@ class FlowControl
 
     Json::Value getFlowControlJsonInfo(bool compact) const;
 
-    void
-    start(NodeID const& peerID,
-          std::function<void(std::shared_ptr<StellarMessage const>)> sendCb,
-          std::optional<uint32_t> enableFCBytes);
+    void start(NodeID const& peerID, std::optional<uint32_t> enableFCBytes);
```
graydon, line 175: If we don't have sendCb anymore, is it still necessary to have a start method separate from the constructor?

src/overlay/TCPPeer.h
```diff
@@ -26,15 +28,81 @@ class TCPPeer : public Peer
...
   private:
+    // Helper class which provides invariance for various data structures;
+    // Each method should be called from the same thread
+    class ThreadRestrictedVars
+    {
+        std::deque<TimestampedMessage> mWriteQueue;
+        std::vector<asio::const_buffer> mWriteBuffers;
+        bool const mUseBackgroundThread;
+        bool mWriting{false};
+        bool mSocketShutdownScheduled{false};
```
graydon, line 39: It _looks_ from the way it's used like `mSocketShutdownScheduled` is now redundant with `mDropStarted`, but I might be misreading?

src/overlay/TCPPeer.cpp
```diff
@@ -173,105 +225,81 @@ TCPPeer::~TCPPeer()
...
     {
-        mWriting = false;
-        // there is nothing to send and delayed shutdown was
-        // requested - time to perform it
-        if (mDelayedShutdown)
-        {
-            shutdown();
-        }
+        mThreadVars.setWriting(false);
+        // Shutdown will be handled automatically by the timer
```
graydon, line 302: This comment relates to a code block that was removed -- the handling of mDelayedShutdown

```suggestion
```

src/overlay/TCPPeer.cpp
```diff
@@ -719,55 +760,84 @@ TCPPeer::recvMessage()
...
+    {
+        auto drop = [errorMsg, self = shared_from_this()]() {
+            // Queue up a drop; we may still process new messages
+            // from this peer, but it'll be dropped as soon as main
+            // thread gets to it
+            self->sendErrorAndDrop(ERR_DATA, errorMsg);
+        };
+        if (!threadIsMain())
+        {
+            mAppConnector.postOnMainThread(drop, "TCPPeer::recvMessage");
```
graydon, line 782:
```suggestion
            mAppConnector.postOnMainThread(drop, "TCPPeer::recvMessage drop");
```

src/overlay/PeerDoor.cpp
```diff
@@ -59,10 +61,17 @@ PeerDoor::acceptNextPeer()
     }
 
     CLOG_DEBUG(Overlay, "PeerDoor acceptNextPeer()");
-    auto sock = make_shared<TCPPeer::SocketType>(mApp.getClock().getIOContext(),
-                                                 TCPPeer::BUFSZ);
+    // Asio guarantees it is safe to create a socket object with overlay's
+    // io_context on main (as long as the socket object isn't shared across
```
graydon, line 65: You use this phrase ("so long as the socket object isn't shared across threads") in another similar comment in TCPPeer and I find it a little confusing: the socket object seems to be "shared" across threads in this codebase, in the sense that a single TCPPeer::SocketType value is periodically accessed from one thread or the other; I think what you're saying is that the object and its methods do not perform any synchronization internally, and so must be protected from simultaneous access from multiple threads by some kind of higher level synchronization. Is that right?

src/overlay/OverlayManagerImpl.h
```diff
@@ -79,7 +83,7 @@ class OverlayManagerImpl : public OverlayManager
     PeerManager mPeerManager;
     PeerDoor mDoor;
     PeerAuth mAuth;
-    bool mShuttingDown;
+    std::atomic<bool> mShuttingDown;
```
graydon, line 86: I don't think this has to be atomic anymore, does it? Nobody else besides main thread should observe it even through methods.

graydon (changes requested):
Handful of minor changes and suggestions, definitely getting close!

I remain nervous about the additional complexity in doing all this, just because the Peer and TCPPeer classes are already such spaghetti; but absent "rewriting the whole subsystem and miraculously gaining clarity in the process" I'm not sure there's a lot more complexity to shave off this PR. Thanks for all the iterations!

src/overlay/Peer.cpp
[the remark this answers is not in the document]
marta-lokhova: Ok, I experimented a bit more, and ended up still using shared_ptr, but moved it into the lambda to ensure that background thread isn't holding any references. We still can't use unique_ptr unfortunately, because std::function doesn't allow its callable to be move-only, so I added a comment about this. I think this is still safer than manually calling `finish`.

src/overlay/FlowControl.h
marta-lokhova: As of now, yes: we still need to set mNodeID and mFlowControlBytesCapacity based on what we get during auth. That being said, https://github.com/stellar/stellar-core/issues/3781 will allow to simplify the flow and we'll be able to initialize mFlowControlBytesCapacity in the constructor. We'll still need to set mNodeID after auth, but that's minor, since we only use peer name for logging.

src/overlay/PeerDoor.cpp
marta-lokhova: That's correct. I re-phrased the comment to clarify this.

src/overlay/OverlayManagerImpl.h
marta-lokhova: I think `shouldAbort` should still check if overlay is shutting down, similar to what we do in master

graydon (approved):
LGTM!

graydon:
r+ 894ef3b

graydon:
r+ ff1acff

graydon:
r+ 894ef3b33f997e4c3214079461d3096f3a6d84d2

graydon:
latobarita: retry

graydon:
@latobarita: retry

PR #4260, opened 2024-03-26
Better error codes in tx queue

### LINKED ISSUE #4221: Core reports `TRY_AGAIN_LATER` error for txs that exceed resource limits
Core returns a blanket `TRY_AGAIN_LATER` error for any scenario (before performing any validation) when we couldn't add a transaction to the tx queue and there is no way to come up with a higher fee. This makes sense for scenarios like having trying to add a second transaction for the same account. But with Soroban such scenarios include cases where trying again later clearly wouldn't help, such as when a transactions exceeds some ledger-wide resource limit.

Ideally we should sanity check the resources before adding the transaction into queue, otherwise we won't be able to come up with a meaningful error message.

marta-lokhova:
Resolves #4221 

Not certain if pushing diagnostic events in the new codepath is problematic (since it's not throttled/validated unlike tx queue limiter), so any input would be appreciated.

### DIFF
--- a/src/herder/TransactionQueue.cpp
+++ b/src/herder/TransactionQueue.cpp
@@ -278,6 +278,20 @@ TransactionQueue::canAdd(TransactionFrameBasePtr tx,
                 return TransactionQueue::AddResult::ADD_STATUS_ERROR;
             }
 
+            // Before rejecting Soroban transactions due to source account
+            // limit, check validity of its declared resources, and return an
+            // appropriate error message
+            if (tx->isSoroban())
+            {
+                if (!tx->checkSorobanResourceAndSetError(
+                        mApp, mApp.getLedgerManager()
+                                  .getLastClosedLedgerHeader()
+                                  .header.ledgerVersion))
+                {
+                    return TransactionQueue::AddResult::ADD_STATUS_ERROR;
+                }
+            }
+
             if (tx->getEnvelope().type() != ENVELOPE_TYPE_TX_FEE_BUMP)
             {
                 // If there's already a transaction in the queue, we reject
--- a/src/herder/test/TransactionQueueTests.cpp
+++ b/src/herder/test/TransactionQueueTests.cpp
@@ -1381,10 +1381,20 @@ TEST_CASE("Soroban TransactionQueue limits",
                 resources.instructions =
                     static_cast<uint32>(conf.txMaxInstructions() + 1);
 
-                // Double the fee
-                auto tx2 =
-                    createUploadWasmTx(*app, account2, initialInclusionFee * 2,
-                                       resourceFee, resources);
+                TransactionFrameBasePtr tx2;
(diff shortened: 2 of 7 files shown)


src/herder/TransactionQueue.cpp
```diff
@@ -278,6 +278,20 @@ TransactionQueue::canAdd(TransactionFrameBasePtr tx,
                 return TransactionQueue::AddResult::ADD_STATUS_ERROR;
             }
 
+            // Before rejecting Soroban transactions due to source account
+            // limit, check validity of its declared resources, and return an
+            // appropriate error message
+            if (tx->isSoroban())
```
SirTyson, line 284: Why do we need to check this first? It seems like the root issue we're trying to solve here is when the only thing keeping a Soroban TX out of the queue is invalid resources, and we return `TRY_AGAIN_LATER`. If there are multiple issues with the TX, such as invalid resources and and multiple TXs from the same account, I think it's reasonable to return `TRY_AGAIN_LATER`, since that's a valid actionable error for one of the issues with the TX.

Not a huge deal either way, but moving this check to the end of the function might buy you a few less calls to `checkValid`, but probably not much.

SirTyson (approved):
LGTM! I'm not well versed in diagnostic event cost, so I'll leave that issue for someone else to comment on.

src/herder/TransactionQueue.cpp
marta-lokhova: mm, I think in the scenario you describe it actually makes more sense to return resource error rather than try again later. in this case, the transaction will never be valid and re-submitting later won't help, so we should probably give the user at least a rough idea about what's wrong.

SirTyson: Yeah I agree that this is a more actionable error to the end user, but if you're concerned about pushing too many diagnostic events you could do this check last instead of first and still be correct, which a few less diagnostic events. Probably not worth doing though since it would only save us in a few edge cases.

dmkozh: We always attach just a single fixed-size diagnostic event during validation, so I don't think we should be concerned about attaching it. IIUC the main reason for starting the validation from tx queue and not `checkValid` is that signature verification is relatively expensive. Everything else should be much cheaper in comparison.

dmkozh (approved):
LGTM

src/herder/TransactionQueue.cpp
marta-lokhova: yeah thanks for confirming, that's what @sisuresh concluded as well, so I think we're good to go with this change.

marta-lokhova:
Since master is now open, can anyone please r+ this PR?

SirTyson:
r+ b88ef2d13478d6e21d0b3e3450a3e9b03d5e675c

PR #4262, opened 2024-03-27
Fix a main-thread assertion in the QuorumIntersectionChecker

graydon:
Change https://github.com/stellar/stellar-core/pull/4240 introduced a new call to `assertThreadIsMain();` in the QI constructor. This is correct for the top-level QI checker, but we then run that QI checker on a background thread and, if it _does_ enjoy intersection, we call `getIntersectionCriticalGroups` which then constructs a pile of QI checkers on the background thread.

The assertion only exists to guard access to the global PRNG to get a seed (the seed used here is a very light requirement that just helps perturb the algorithm's set-splitting path). So a simple solution is to not access the global PRNG in the constructor, but to access it in the scope outside and pass-in a seed. It's harmless to use the same seed in all the sub-checkers constructed in `getIntersectionCriticalGroups`, so I do that here.

cc @SirTyson

### DIFF
--- a/src/herder/HerderImpl.cpp
+++ b/src/herder/HerderImpl.cpp
@@ -29,6 +29,7 @@
 #include "util/DebugMetaUtils.h"
 #include "util/LogSlowExecution.h"
 #include "util/Logging.h"
+#include "util/Math.h"
 #include "util/StatusManager.h"
 #include "util/Timer.h"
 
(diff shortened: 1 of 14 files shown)


SirTyson (approved):
Thanks for the fix! I'm wondering if we should wrap `gRandomEngine` in a light weight class that just checks if the current thread is main on access, it seems pretty easy to accidentally call it from a background thread as it currently stands. A refactor for another day though.

marta-lokhova (approved):

marta-lokhova:
r+ 68b7f8aa7778d5317757628c4f3edcf055663d3e

PR #4270, opened 2024-04-03
No longer serialize small indexes to disk

### LINKED ISSUE #4268: Stop Persisting Individual Indexes on BucketListDB
The individual index type on BucketListDB is limited to small Buckets (by default Buckets < 20 mb). On startup, these files are so small that there's little to gain by serializing the indexes to disk. Additionally, saving these small indexes to disk creates more disk churn, specifically on the main thread in `Bucket::addBatch`. We should only persist RangeIndexes, as these are much larger and never need to be saved to disk as part of `Bucket::addBatch`

SirTyson:
# Description

Resolves #4268

Previously, we serialized every BucketListDB index to disk, regardless of size. This created significant disk churn in small, top level buckets, where new indexes are being re-written constantly.

This change no longer serializes the `IndividualIndex` type to disk to reduce this churn. By default, all buckets with size < 20 MB have individual indexes, so this will remove disk churn on the small, top most buckets. This also removes writing indexes to disk on the main thread, as previously `Bucket::addBatch` needed to write indexes to disk as part of `ledgerClose`.

There are essentially no drawbacks to recreating individual indexes on startup. They are so small that recreating them takes about as long as deserializing them. This change adds approximately 100ms to node startup time.

# Checklist
- [x] Compiles
- [x] Ran all tests

### DIFF
--- a/src/bucket/BucketIndexImpl.cpp
+++ b/src/bucket/BucketIndexImpl.cpp
@@ -214,13 +214,25 @@ BucketIndexImpl<IndexT>::BucketIndexImpl(BucketManager& bm,
     }
 }
 
-template <class IndexT>
+// Individual indexes are associated with small buckets, so it's more efficient
+// to just always recreate them instead of serializing to disk
+template <>
 void
-BucketIndexImpl<IndexT>::saveToDisk(BucketManager& bm, Hash const& hash) const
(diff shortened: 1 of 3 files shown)


dmkozh (commented):
LGTM, let's merge after we unfreeze Core

dmkozh:
r+ 66675ae

graydon (approved):

PR #4273, opened 2024-04-05
Fix memory issue with individual indexes

SirTyson:
# Description

Fixes an issue where `IndividualIndexes` would over allocate memory.

# Checklist
- [x] Compiles
- [x] Ran all tests

### DIFF
--- a/src/bucket/BucketIndexImpl.cpp
+++ b/src/bucket/BucketIndexImpl.cpp
@@ -82,7 +82,6 @@ BucketIndexImpl<IndexT>::BucketIndexImpl(BucketManager& bm,
             xdr::xdr_traits<BucketEntry>::serial_size(BucketEntry{});
         auto fileSize = fs::size(filename.string());
         auto estimatedNumElems = fileSize / estimatedLedgerEntrySize;
-        size_t estimatedIndexEntries;
 
         // Initialize bloom filter for range index
         if constexpr (std::is_same<IndexT, RangeIndex>::value)
@@ -105,7 +104,7 @@ BucketIndexImpl<IndexT>::BucketIndexImpl(BucketManager& bm,
             params.random_seed = shortHash::getShortHashInitKey();
             params.compute_optimal_parameters();
             mData.filter = std::make_unique<bloom_filter>(params);
-            estimatedIndexEntries = fileSize / mData.pageSize;
+            auto estimatedIndexEntries = fileSize / mData.pageSize;
             CLOG_DEBUG(
                 Bucket,
                 "Bloom filter initialized with params: projected element count "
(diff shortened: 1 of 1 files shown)


src/bucket/BucketIndexImpl.cpp
```diff
@@ -105,7 +104,7 @@ BucketIndexImpl<IndexT>::BucketIndexImpl(BucketManager& bm,
             params.random_seed = shortHash::getShortHashInitKey();
             params.compute_optimal_parameters();
             mData.filter = std::make_unique<bloom_filter>(params);
-            estimatedIndexEntries = fileSize / mData.pageSize;
+            auto estimatedIndexEntries = fileSize / mData.pageSize;
             CLOG_DEBUG(
                 Bucket,
                 "Bloom filter initialized with params: projected element count "
```
anupsdf, line 110: How about logging `estimatedIndexEntries` for debugging purposes?

anupsdf:
For lower level buckets, by not calling reserve for individual indexes is there a way to see how bad it will be in terms of memory reallocations?

src/bucket/BucketIndexImpl.cpp
SirTyson: I don't think this would be particularly useful. `estimatedIndexEntries` is directly proportional to the BucketList size, which we already have a metric for. The issue with the `IndividualIndexes` was that we were using an estimate for the reserve size. This is not the case for range index we are reserving here, where we don't use an estimate but reserve exactly how much space we will need. I think this particular metric would be difficult to reason about since it would vary significantly bucket to bucket and there's not really a visualization that I think would be actionable.

I think the underlying property here that is important is `BucketIndex` memory usage. I think that would be a worthwhile metric to add, but it would be more involved than just logging `estimatedIndexEntries`. I think it's worth doing, but isn't related to this specific issue and is something we can add later.

SirTyson:
> For lower level buckets, by not calling reserve for individual indexes is there a way to see how bad it will be in terms of memory reallocations?

I ran a few tests and couldn't see any difference in metrics on Grafana (though we might be able to measure a difference in tracy). I don't think this is a big issue though, as the size of the memory reallocation is guaranteed to be small (because large indexes become `RangeIndex`). Bucket indexing is also a disk IO bound task that we do in the background thread, so I don't think a relatively small in-memory realloc would be significant compared to the amount of disk work we have to do during indexing anyway.

anupsdf (approved):
LGTM

marta-lokhova (approved):
LGTM! Let's merge this when master opens for v21.0.0rc2

marta-lokhova:
r+ e6f9b7f32ca2fc5456ee39c307e883ffd4c4f1a7

graydon (approved):
LGTM

PR #4279, opened 2024-04-10
Overlay stability fixes

marta-lokhova:
Discovered while implementing https://github.com/stellar/stellar-core/pull/4258:
* During core shutdown, we weren't gracefully dropping peers because OverlayManager was marking itself as shutdown too early (causing peers to no-op)
* We weren't checking if peer got dropped during async connect flow, causing the handler to "resurrect" the peer.


SirTyson (approved):
LGTM

SirTyson:
r+ f0deaac6ca3c5c11e830fab3a2122b1f9d62b823

PR #4291, opened 2024-04-19
Overlay improvements

marta-lokhova:
Extracting tech debt changes from the parallelization PR, to make things easier to review. This one should be pretty straightforward. Most notable change is to how we assert: previously, core would abort without any useful information about the problematic file and line number.

### DIFF
--- a/src/database/Database.cpp
+++ b/src/database/Database.cpp
@@ -447,7 +447,7 @@ soci::session&
 Database::getSession()
 {
     // global session can only be used from the main thread
-    assertThreadIsMain();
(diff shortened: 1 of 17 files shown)


SirTyson (approved):
LGTM, bu there's a compilation error that needs to be fixed.

marta-lokhova:
> LGTM, bu there's a compilation error that needs to be fixed.

ah yes, must have slipped in during rebase. fixed!

SirTyson:
r+ f4b2002910c2059ab6c235ce8d8407c335717240

PR #4297, opened 2024-04-26
April 2024 warning police

graydon:
Fix some warnings that cropped up from (a) recent work in overlay and buckets and (b) recent upgrades to my copy of clangd which now has [IWYU turned on by default](https://clangd.llvm.org/guides/include-cleaner), and I figured .. why not start listening to it and cleaning things up incrementally?


src/herder/test/HerderTests.cpp
```diff
@@ -5862,7 +5862,7 @@ TEST_CASE("SCP message capture from previous ledger", "[herder]")
     // Initialize simulation
     auto networkID = sha256(getTestConfig().NETWORK_PASSPHRASE);
     auto simulation = std::make_shared<Simulation>(
-        Simulation::OVER_LOOPBACK, networkID, [version](int i) {
+        Simulation::OVER_LOOPBACK, networkID, [](int i) {
```
bboston7, line 5865: Unless Microsoft fixed their compiler this unnecessary `version` capture is necessary due to an MSVC bug (see #4194, and specifically [this comment](https://github.com/stellar/stellar-core/pull/4194#issuecomment-1947821687))

bboston7:
If this change resolves all of the warnings, I also wonder what your thoughts are on setting `-Werror` in CI to ensure future warnings don't slip in.

src/herder/test/HerderTests.cpp
graydon: Ok, fixed (crudely: made it non-constexpr and captured it)

graydon:
> If this change resolves all of the warnings, I also wonder what your thoughts are on setting `-Werror` in CI to ensure future warnings don't slip in.

It doesn't. there are warnings in several submodules and depending on the compiler and compiler _version_ you use you'll get a different set. In general while I appreciate the moral goal of `-Werror`, I think in practice the C++ compiler ecosystem and language stability model makes it .. not very viable most of the time. It's more viable in Rust! But even then people (cough cough -- us) sometimes turn on "error on any clippy warning" and then a single github actions builder-image update breaks the build.

SirTyson (approved):
LGTM! It looks like we're not compiling though in CI, I think you might have removed a few too many includes.

bboston7:
> It doesn't. there are warnings in several submodules and depending on the compiler and compiler _version_ you use you'll get a different set. In general while I appreciate the moral goal of `-Werror`, I think in practice the C++ compiler ecosystem and language stability model makes it .. not very viable most of the time. It's more viable in Rust! But even then people (cough cough -- us) sometimes turn on "error on any clippy warning" and then a single github actions builder-image update breaks the build.

Makes sense, thanks!

graydon:
> LGTM! It looks like we're not compiling though in CI, I think you might have removed a few too many includes.

This is weird! I hate to be the "it works on my machine" guy but .. it works on my machine! I will try to figure out the cause.

src/herder/test/HerderTests.cpp
```diff
@@ -5856,13 +5856,12 @@ TEST_CASE("exclude transactions by operation type", "[herder]")
...
-        static_cast<uint32_t>(SOROBAN_PROTOCOL_VERSION);
+    uint32_t version = static_cast<uint32_t>(SOROBAN_PROTOCOL_VERSION);
 
     // Initialize simulation
     auto networkID = sha256(getTestConfig().NETWORK_PASSPHRASE);
     auto simulation = std::make_shared<Simulation>(
-        Simulation::OVER_LOOPBACK, networkID, [](int i) {
+        Simulation::OVER_LOOPBACK, networkID, [version](int i) {
             auto cfg = getTestConfig(i, Config::TESTDB_ON_DISK_SQLITE);
             cfg.TESTING_UPGRADE_LEDGER_PROTOCOL_VERSION = version;
```
marta-lokhova, line 5866: is this needed at all? I think this test should just use the most recent version for testing, which is the default, if we don't override `TESTING_UPGRADE_LEDGER_PROTOCOL_VERSION`

graydon: I have no idea why the test is hard-wired to a single version. possibly copypasta. it works fine without that line and in fact without the ConfigGen argument at all.

graydon:
>  I will try to figure out the cause

It was too embarrassingly silly to mention here. Anyway, fixed.

marta-lokhova (approved):

marta-lokhova:
r+ efb9d1ef4242ec82caf0a9e90a0eff63450ff9b1

PR #4299, opened 2024-04-30
Avoid copying StellarMessage in flooding paths

marta-lokhova:
Follow-up addressing @bboston7's comments in https://github.com/stellar/stellar-core/pull/4293

I also noticed that we make unnecessary copies of StellarMessage in places like SCP and transaction flooding; since we're flooding hundreds to thousands of messages constantly, these can add up to a tangible overhead.


bboston7 (approved):
Looks great, thanks! Good catch with those other additional copies.

graydon (approved):

graydon:
r+ dca05af

PR #4304, opened 2024-05-03
Bump overlay version to 33

marta-lokhova:
We need to bump overlay version in v21.0.0 so we can reject old clients after the protocol upgrade.


graydon (approved):

graydon:
r+ 3ae5da7

PR #4308, opened 2024-05-09
Move tracy FrameMark macro so it occurs during replay as well as online consensus

graydon:
This is a tiny change that just moves the place we mark a "frame ending" to tracy from a place that only happens during online consensus to a deeper place in the ledger-close path so that we get "frames" representing ledger boundaries when doing replay as well.


marta-lokhova:
r+ e0fdd9ecbadf2350c6dbe68be3fb0e9d23879dff

Issue #4290, opened 2024-04-18
Refactor ApplyBuckets Work


SirTyson:
Currently, `ApplyBuckets` work reasons at the `BucketLevel` instead of at the per-bucket level. Especially after the BucketListDB apply optimization, this means that there's some tricky logic required to keep our current place in the BucketList and make sure our iteration order is correct. We should refactor this to apply per-bucket instead of per-level to simplify the logic.

AungkoMCM:
[README.md](https://github.com/user-attachments/files/15541900/README.md)

marta-lokhova:
Resolved by https://github.com/stellar/stellar-core/pull/4314

PR #4314, opened 2024-05-14
ApplyBucketWork at per-bucket level

ThomasBrady:
# Description
Refactors `ApplyBucketsWork` to apply buckets per-bucket rather than per-level. Deleted a lot of book keeping code referring to curr/snap apply ordering. 

Issue:  https://github.com/stellar/stellar-core/issues/4290

# Checklist
- [x] Compiles
- [x] Ran all tests

### DIFF
--- a/src/catchup/ApplyBucketsWork.cpp
+++ b/src/catchup/ApplyBucketsWork.cpp
@@ -58,25 +58,6 @@ ApplyBucketsWork::startingLevel()
                                                   : BucketList::kNumLevels - 1;
 }
 
-bool
-ApplyBucketsWork::appliedAllLevels() const
-{
-    if (mApp.getConfig().isUsingBucketListDB())
-    {
-        return mLevel == BucketList::kNumLevels - 1;
-    }
-    else
-    {
-        return mLevel == 0;
-    }
-}
-
-uint32_t
-ApplyBucketsWork::nextLevel() const
-{
-    return mApp.getConfig().isUsingBucketListDB() ? mLevel + 1 : mLevel - 1;
-}
-
 ApplyBucketsWork::ApplyBucketsWork(
     Application& app,
     std::map<std::string, std::shared_ptr<Bucket>> const& buckets,
@@ -86,7 +67,6 @@ ApplyBucketsWork::ApplyBucketsWork(
     , mBuckets(buckets)
     , mApplyState(applyState)
     , mEntryTypeFilter(onlyApply)
-    , mApplying(false)
     , mTotalSize(0)
     , mLevel(startingLevel())
     , mMaxProtocolVersion(maxProtocolVersion)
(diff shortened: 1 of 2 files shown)


src/catchup/ApplyBucketsWork.cpp
```diff
@@ -325,62 +254,45 @@ ApplyBucketsWork::doWork()
...
+                                    mSeenKeys};
+        TempLedgerVersionSetter tlvs(mApp, mMaxProtocolVersion);
+
+        auto isCurr = isUsingBucketListDB ? mBucketsToApplyIndex % 2 == 0
+                                          : mBucketsToApplyIndex % 2 == 1;
+        // Only advance the applicator if there are entries to apply.
+        if (applicator)
         {
-            startLevel();
+            advance(isCurr ? "curr" : "snap", applicator);
```
ThomasBrady, line 276: I assume we still want log messages to print `snap` / `curr` to show more granular progress? If not I can remove this logic as well

SirTyson (changes requested):
Currently, this doesn't work because `advance` is not being called properly. The way the work classes are designed, we do long running jobs in chunks to not hog resources. The expectation is that `doWork()` will be called periodically, do a little work, then return quickly.

`BucketApplicator` assumes this interface. If you look at `BucketApplicator::advance`, you'll see that it only applies `LEDGER_ENTRY_BATCH_COMMIT_SIZE` entries then exits. This means that, especially for large buckets, `BucketApplicator::advance` must be called many times before the Bucket finishes applying. The issue is this code only calls `BucketApplicator::advance` a single time before moving on to the next bucket.

If you look at the old `ApplyBucketsWork`, we maintained `mFirstBucketApplicator` and `mSecondBucketApplicator` pointers to persist Applicators across calls to `doWork`. We only move on to the next bucket when `*mFirstBucketApplicator` is false (note that this is pointer dereference, not checking if the pointer itself is null), indicating the applicator has reached EOF. You need to implement similar logic, where you maintain an applicator pointer and only advance to the next bucket when the applicator is false. Currently you check if `applicator` is true once then call apply, but you always advance to the next bucket unconditional, even if after the first call to `advance` `applicator` is still `true`.

SirTyson (approved):
Overall looks good to me! Just a few nit pics. There's still some complexity involving `mLevel` and `isCurr` I'd like to see get factored out, but unfortunately it looks like this may be required given how we check invariants :(

SirTyson:
r+ ba9f43f939606d74752e1910fefd3e4fc76f0cec

SirTyson:
Trying again

r+ 545b721eb3a7ef518bd21a89b2e83d02d039632f

graydon:
@latobarita: retry

graydon:
r+ 545b721

ThomasBrady:
@latobarita: retry

PR #4328, opened 2024-05-17
Bump tracy to v0.10

marta-lokhova:
Bump tracy to v0.10. Most notable update is the UI overhaul, which makes it much faster and smoother (see https://wolf.nereid.pl/posts/how-tracy-faster/ for a bit more info)


graydon:
Generally I'm happy to track tracy upstream _but_ there is a bit of awkwardness around rust. We also emit tracy events from soroban and so we need a _precisely_ matched version of [tracy-client](https://crates.io/crates/tracy-client) as well as an unfortunate manual regeneration of the patched/forked [tracy-client-sys](https://github.com/stellar/stellar-core/tree/master/lib/tracy-client-sys) crate we use locally due to various silly technical requirements.

I think it should be mostly sufficient to bump [the tracy-client ref in Cargo.toml](https://github.com/stellar/stellar-core/blob/master/src/rust/Cargo.toml#L14) and then do `cargo-install --locked bindgen` and re-run the [tracy-client-sys regenerate.sh script](https://github.com/stellar/stellar-core/blob/master/lib/tracy-client-sys/src/regenerate.sh) but if you want I can handle this part. Depends how much you want to fight with rust stuff.

graydon:
Also depends on an update on the soroban side: https://github.com/stellar/rs-soroban-env/pull/1417

graydon:
Pushed a change on here that illustrates (it's not quite what needs to be here for a final change, as it references a PR git head over in rs-soroban-env, but .. roughly this is the idea!)

marta-lokhova:
thanks for the fixes @graydon! LGTM.

marta-lokhova:
r+ c3b1269e3c2244cddac8cdaecec6e20461c6ca3b

PR #4334, opened 2024-05-29
Track SCP latencies in milliseconds for overlay survey

bboston7:
# Description

Fixes #4333 by recording SCP latencies in milliseconds rather than nanoseconds, which could overflow when placed in a `uint32_t`.

I also checked the other survey data fields for potential overflow and didn't find anything else that could overflow. However, in doing so I noticed that I failed to bump
`TIME_SLICED_SURVEY_MIN_OVERLAY_PROTOCOL_VERSION` to `34` when I bumped the overlay protocol version, so I fixed that as well.

# Checklist
- [x] Reviewed the [contributing](https://github.com/stellar/stellar-core/blob/master/CONTRIBUTING.md#submitting-changes) document
- [x] Rebased on top of master (no merge commits)
- [x] Ran `clang-format` v8.0.0 (via `make format` or the Visual Studio extension)
- [x] Compiles
- [x] Ran all tests
- [ ] If change impacts performance, include supporting evidence per the [performance document](https://github.com/stellar/stellar-core/blob/master/performance-eval/performance-eval.md)

PR #4342, opened 2024-06-01
Update network survey script to support V2 survey

### LINKED ISSUE #4302: Update overlay survey script to support new time sliced survey
#4275 adds a new time sliced survey mechanism with new HTTP endpoints for running a survey. We should update the python script to use the new endpoints.

bboston7:
# Description

Resolves #4302

This change updates the network overlay survey script to support the new V2 survey API.

To test this, I also updated the simulator to simulate the new API, then I checked that the resulting graph from simulating an old pubnet survey is structurally isomorphic to the input graph. I also checked that the new fields are present in the output graph.

# Checklist
- [ ] Compiles
- [ ] Ran all tests

### DIFF
--- a/scripts/OverlaySurvey.py
+++ b/scripts/OverlaySurvey.py
@@ -89,7 +89,10 @@ MAX_COLLECT_DURATION = 30
 # requests sent to select nodes, followed by a wait period of 15 seconds,
 # followed by checking for responses and building up the next batch of requests
 # to send. Therefore, a setting of `8` is roughly 2 minutes of inactivity
-# before the script considers the survey complete.
+# before the script considers the survey complete. This is necessary because
+# it's very likely that not all surveyed nodes will respond to the survey.
+# Therefore, we need some cutoff after we which we assume those nodes will never
+# respond.
 MAX_INACTIVE_ROUNDS = 8
 
 def get_request(url, params=None):
--- a/scripts/OverlaySurvey.py
+++ b/scripts/OverlaySurvey.py
@@ -2,75 +2,105 @@
 """
 Script to trigger network survey and record results. Sample survey result:
 {
-    "topology" :
-    {
-      "GDZPYXO3HC3GC6CCNTJH7BNURRH5VJTPDP5YALGNDCTPNDCSR74QBH3H" : {
-         "inboundPeers" : [
-            {
-               "bytesRead" : 218480,
-               "bytesWritten" : 225188,
-               "duplicateFetchBytesRecv" : 0,
-               "duplicateFetchMessageRecv" : 0,
-               "duplicateFloodBytesRecv" : 173296,
-               "duplicateFloodMessageRecv" : 443,
-               "messagesRead" : 511,
-               "messagesWritten" : 522,
-               "nodeId" : "GB4HZXA2IALAMH4CIA5CBLOITQT45CC3275MNRQ4VTYJCFKJW2AHHMXP",
-               "secondsConnected" : 4,
-               "uniqueFetchBytesRecv" : 0,
-               "uniqueFetchMessageRecv" : 0,
-               "uniqueFloodBytesRecv" : 10288,
-               "uniqueFloodMessageRecv" : 26,
-               "version" : "v19.5.0-126-gcdf8d018-dirty"
-            }
-         ],
-         "maxInboundPeerCount" : 64,
-         "maxOutboundPeerCount" : 8,
-         "numTotalInboundPeers" : 6,
-         "numTotalOutboundPeers" : 8,
-         "outboundPeers" : [
-            {
-               "bytesRead" : 220540,
-               "bytesWritten" : 224720,
-               "duplicateFetchBytesRecv" : 0,
-               "duplicateFetchMessageRecv" : 0,
-               "duplicateFloodBytesRecv" : 186584,
-               "duplicateFloodMessageRecv" : 473,
-               "messagesRead" : 517,
-               "messagesWritten" : 521,
-               "nodeId" : "GARM4GDQIQ2Z4SN5OXPG4XVHWZMKYEYTHY2X4TTP2TD2FDRAQOZIVJVX",
-               "secondsConnected" : 4,
-               "uniqueFetchBytesRecv" : 0,
-               "uniqueFetchMessageRecv" : 0,
-               "uniqueFloodBytesRecv" : 10612,
-               "uniqueFloodMessageRecv" : 27,
-               "version" : "v19.5.0-126-gcdf8d018-dirty"
-            }
-         ]
-      }
+  "backlog": [],
+  "badResponseNodes": null,
+  "surveyInProgress": true,
+  "topology": {
+    "GBBNXPPGDFDUQYH6RT5VGPDSOWLZEXXFD3ACUPG5YXRHLTATTUKY42CL": null,
+    "GDEXJV6XKKLDUWKTSXOOYVOYWZGVNIKKQ7GVNR5FOV7VV5K4MGJT5US4": {
+      "inboundPeers": [
+        {
+          "averageLatencyMs": 23,
+          "bytesRead": 26392,
+          "bytesWritten": 26960,
+          "duplicateFetchBytesRecv": 0,
+          "duplicateFetchMessageRecv": 0,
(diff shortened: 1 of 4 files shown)


AungkoMCM (approved):
#4342 #4299 #4293 #4334

scripts/OverlaySurvey.py
```diff
@@ -113,53 +154,40 @@ def update_results(graph, parent_info, parent_key, results, is_inbound):
...
-        print("Send request to %s" % key)
-        global request_count
-        request_count += 1
-        if (request_count % limit) == 0:
-            print("Submitted %i queries, sleep for ~1 ledger" % limit)
-            time.sleep(5)
-
-    print("Done")
+    # Add survey results to parent node (if available)
+    field_names = [("numTotalInboundPeers", "totalInbound"),
```
marta-lokhova, line 158: thanks for cleaning this up, much nicer without all the duplication!

scripts/OverlaySurvey.py
```diff
@@ -2,75 +2,105 @@
...
+# internal limit.
+MAX_DURATION = 30
+
+# Maximum number of consecutive rounds in which the surveyor neither sent
+# requests to nor received responses from any nodes. A round contains a batch of
+# requests sent to select nodes, followed by a wait period of 15 seconds,
+# followed by checking for responses and building up the next batch of requests
+# to send. Therefore, a setting of `8` is roughly 2 minutes of inactivity
+# before the script considers the survey complete.
+MAX_INACTIVE_ROUNDS = 8
```
marta-lokhova, line 93: Do we need this for cases where we don't get complete survey back from the network? Otherwise, we'd mark the survey complete once we've discovered and surveyed all the peers, right?

scripts/OverlaySurvey.py
```diff
@@ -113,53 +154,40 @@ def update_results(graph, parent_info, parent_key, results, is_inbound):
...
+    update_node(graph, parent_info, parent_key, results, field_names)
+
+
+def send_survey_requests(peer_list, url_base):
+    """
+    Request survey data from a list of peers. `url_base` is the root HTTP
+    endpoint to send requests to.
+    """
+    request_url = url_base + "/surveytopologytimesliced"
+    logger.info("Requesting survey data from %s peers", len(peer_list))
```
marta-lokhova, line 177: No more request rate throttling? (previous version of this code limited number of requests per ledger to 10)

scripts/OverlaySurvey.py
```diff
@@ -289,28 +318,48 @@ def run_survey(args):
...
+    nonce = random.randint(0, 2**32-1)
+    logger.info("Starting survey with nonce %s", nonce)
+    response = get_request(url=start_collecting, params={'nonce': nonce})
+    if response.text != util.START_SURVEY_COLLECTING_SUCCESS_TEXT:
+        logger.critical("Failed to start survey: %s", response.text)
+        sys.exit(1)
+
+    # Sleep for duration of collecting phase
+    logger.info("Sleeping for collecting phase (%i minutes)", args.duration)
+    time.sleep(args.duration * 60)
```
marta-lokhova, line 341: do we need to sleep an extra minute just to allow the start collecting phase message to propagate?

scripts/OverlaySurvey.py
```diff
@@ -289,28 +318,48 @@ def run_survey(args):
...
+
+    # Stop survey recording
+    logger.info("Stopping survey collecting")
+    response = get_request(url=stop_collecting)
+    if response.text != util.STOP_SURVEY_COLLECTING_SUCCESS_TEXT:
+        logger.critical("Failed to stop survey: %s", response.text)
+        sys.exit(1)
+
+    # Allow time for stop message to propagate
+    sleep_time = 30
```
marta-lokhova, line 351: maybe increase it to a minute just in case we're dealing with slow nodes on the edges of the network?

marta-lokhova (changes requested):
Thanks for the changes!

scripts/OverlaySurvey.py
bboston7: [`args.duration` has a minimum value of `1`](https://github.com/bboston7/stellar-core/blob/0d339f979cfbe63e9e23f0350765286734ba5170/scripts/OverlaySurvey.py#L507), so in effect the "start collecting phase" message will have at least 1 minute to propagate.

scripts/OverlaySurvey.py
bboston7: I removed the request throttling because stellar-core has its own builtin request throttling (see [`SurveyManager::topOffRequests`](https://github.com/bboston7/stellar-core/blob/0d339f979cfbe63e9e23f0350765286734ba5170/src/overlay/SurveyManager.cpp#L1006)). Specifically, stellar-core sends 10 requests from its queue every 15 seconds.

Even with the old request rate throttling this queue would build up because the script would add 10 nodes every 5 seconds, but stellar-core would only send a batch of 10 every 15 seconds.

That being said, I'm happy to add back in request rate throttling as a extra safety measure if you think it's necessary. Thoughts?

scripts/README.md
```diff
@@ -12,15 +12,16 @@ This folder is for storing any scripts that may be helpful for using stellar-cor
 - Usage - Ex. `python3 OverlaySurvey.py -gs gs.json survey -n http://127.0.0.1:11626 -d 50 -sr sr.json -gmlw gmlw.graphml` to run the survey, `python3 OverlaySurvey.py -gs gs.json analyze -gmla gmla.graphml` to analyze an existing graph, or `python3 OverlaySurvey.py -gs gs.json augment -gmli gmlw.graphml -gmlo augmented.graphml` to augment the existing graph with data from StellarBeat.
 
     - `-gs GRAPHSTATS`, `--graphStats GRAPHSTATS` - output file for graph stats (Optional)
+    - `-v`, `--verbose` - increase log verbosity (Optional)
     - sub command `survey` - run survey and analyze
         - `-n NODE`, `--node NODE` - address of initial survey node
-        - `-d DURATION`, `--duration DURATION` - duration of survey in seconds
+        - `-d DURATION`, `--duration DURATION` - duration of survey collecting phase in minutes
```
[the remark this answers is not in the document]
bboston7: Good idea. I replaced `-d/--duration` with `-c/--collect-duration` in 1a36c486dca5149c1f619ea3e294f77604db32ec

scripts/OverlaySurvey.py
bboston7: Good point. Increased to 1 minute in 1a36c486dca5149c1f619ea3e294f77604db32ec

scripts/OverlaySurvey.py
bboston7: That's exactly right. I added an explanation to this comment to clarify why `MAX_INACTIVE_ROUNDS` is necessary.

scripts/OverlaySurvey.py
marta-lokhova: That makes sense, I'm just trying to recall why we added this throttling in the first place. I agree though that it's weird to have throttling in both places, and ideally core should properly deal with all the requests. Looks like this [PR](https://github.com/stellar/stellar-core/pull/3633) added throttling because survey on the core side used to expire too quickly, I think this was related to the old `-d` flag. This shouldn't be the case with the new-style survey, since our reporting phase is now long by default, right?

bboston7: Ah, that makes sense. Digging into it, I think without the throttling in the script there's a weird race condition that would prematurely end the survey on the stellar-core side immediately after processing a large batch of requests. Trickling in requests causes the expiration to be regularly extended and avoids the premature expiration.

The V2 survey removed the `duration` parameter and instead [uses the survey phase](https://github.com/bboston7/stellar-core/blob/bcf2e59ab1e714914c6127f28887a2c5518fa45a/src/overlay/SurveyManager.cpp#L1169) to determine when the survey is finished.

marta-lokhova: Yeah, that makes sense. I like that in this version we move the decision to end survey to the script itself after some inactivity rather than try to end in core.

marta-lokhova (approved):
LGTM!

marta-lokhova:
r+ bcf2e59ab1e714914c6127f28887a2c5518fa45a

PR #4343, opened 2024-06-03
overlay: perf and error handling improvements

marta-lokhova:
Follow-up to https://github.com/stellar/stellar-core/pull/4258 with the following tweaks:
* Only verify SCP envelope signatures in the background
* Send path: move the check for connection drop earlier in the path. Reason is that in some situations we want to queue up an ERROR message with a drop reason, which should still go through even if connection is shutting down
* Break out of the read loop if a malformed message was received. Previous behavior existed in master prior to PR4258 but with the addition of async drops in the background became more error-prone.
* Put read throttling logic under a single lock (instead of two calls, canRead and throttleRead) to prevent main thread from logically racing background when releasing read capacity.
* Add a few tests for receiving bad messages in TCPPeer


graydon (approved):
LGTM!

graydon:
r+ 7033ee2

PR #4345, opened 2024-06-04
emit warning if bl merge slower than 50 seconds

ThomasBrady:
# Description

Resolves #X

# Checklist
- [ ] Compiles
- [ ] Ran all tests


src/bucket/FutureBucket.cpp
```diff
@@ -372,14 +372,28 @@ FutureBucket::startMerge(Application& app, uint32_t maxProtocolVersion,
...
-                        "Bucket merge on level {} finished in {} seconds "
-                        "({}% of available time)",
-                        level, time.count(), timePct);
+                    auto time = timeScope.Stop();
+                    std::chrono::duration<double> doubleTime(time);
+                    double timePct =
+                        doubleTime.count() / availableTime.count() * 100;
+                    auto timeSec =
+                        std::chrono::duration_cast<std::chrono::seconds>(time);
+                    if (time > std::chrono::seconds(50))
```
marta-lokhova, line 381: Can we use `LogSlowExecution` with threshold of 50 seconds for this?

ThomasBrady: I considered that, but we currently have (and I assume want to keep?) a debug log for every merge. If I change to use LogSlowExecution with a threshold of 50, then we will get a debug log for merges at the threshold and a warn (10x threshold) log for merges at 500 seconds. I could set a threshold of 5 seconds for a debug at 5 and a warn at 50. The debug would be duplicated in that case but I guess its not a big deal.

SirTyson: Hmm I think `LogSlowExecution` with a threshold of 25 is a good middle ground, which gives us debug at 25 and warning at 250 right? BucketListDB is pretty mature so idk if we need super granular debug logs these days.

marta-lokhova: sorry what i meant is that we should only warn if bucket merge crosses 1 min mark (so adjust the threshold accordingly). we should still debug log for all merges though.

SirTyson: Sure, I might actually up that thought to 2 minutes. 1 minute is pretty standard for big buckets and doesn't indicate that anything is wrong.

marta-lokhova: yeah that seems fine

PR #4346, opened 2024-06-04
Un-downgrade tracy-client-sys accidentally caused by feature selection

graydon:
I failed to copy from upstream all the cargo features visible in tracy-client-sys when updating our [local override of it](https://github.com/stellar/stellar-core/blob/master/lib/tracy-client-sys/Cargo.toml)

As a result, when doing a recent build, cargo helpfully failed to satisfy `tracy-client`'s dependency on `tracy-client-sys` using the locked `tracy-client-sys 0.22.2` (which selects the local override) and instead added an additional downgraded dependency on `tracy-client-sys 0.22.1`. Of course this also has the unfortunate side effect of _not using our local copy of the C++ tracy client library at all_, including any local patching we've had to do to make it build (eg. [ignoring an fscanf return value as required by gcc 11](https://github.com/stellar/tracy/pull/3)).

Anyway short story long the fix is just to add the necessary features, delete the auxiliary tracy-client-sys 0.22.1 dependency, and fix any fallout.

(Note that this also causes the host-dep-tree-curr.txt entry to change to `inexact` -- because it's just pointing to an in-tree patched version of `tracy-client-sys` -- which on the one hand it's never been before, but on the other hand is _probably_ correct-ish? I.e. I think it's probably the case that the hash we had in there before for 0.20.0 was wrong, because it would have been the checksum of the crates.io package and we literally never built against that, I think it just wound up in the lockfile by accident while I was originally iterating on getting the cross-language tracy build working.)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1340,16 +1340,7 @@ checksum = "82da0d50d9df1106619b1e5b118f39de779f7d8b9c3504485b291cb16fabd20f"
 dependencies = [
  "loom",
  "once_cell",
- "tracy-client-sys 0.22.1",
-]
-
-[[package]]
-name = "tracy-client-sys"
-version = "0.22.1"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "078c7ed72141b0e4369671a7f7af0eecffe18d753bf0296adca9c7add7276c9d"
-dependencies = [
- "cc",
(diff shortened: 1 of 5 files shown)


SirTyson:
r+ 36e526f90618e4df135f9f5200931c83dffa6935

PR #4352, opened 2024-06-07
Add initial version of the debug info gathering script

jacekn:
### What

Add initial version of the debug info gathering script

### Why

We want to make it easier to gather information useful for troubleshooting

### DIFF
--- a/scripts/gather_debug_info
+++ b/scripts/gather_debug_info
@@ -30,22 +30,22 @@ class Gatherer(object):
                 func(self)
                 return True
             except:
-                print('Error calling function {}'.format(func.__name__))
+                print(f'Error calling function {func.__name__}')
                 return False
         return wrapper
 
     def __init__(self, args):
         timestamp = datetime.now().strftime("%Y-%m-%d-%H-%M-%S")
         self.base_dir = args.dest
-        self.scratch_dir = os.path.join(args.dest, '{}-{}'.format('stellar-core-debug-info', timestamp))
-        self.tgz_file = '{}.tar.gz'.format(self.scratch_dir)
+        self.scratch_dir = os.path.join(args.dest, f'stellar-core-debug-info-{timestamp}')
+        self.tgz_file = f'{self.scratch_dir}.tar.gz'
         self.core_config = args.core_config
         self.log_dir = args.log_dir
         self.header_template = '#####################\n# {}\n#####################\n'
 
     def pre_flight(self):
         if not os.path.isdir(self.scratch_dir):
-            print('Creating scratch directory: {}'.format(self.scratch_dir))
+            print(f'Creating scratch directory: {self.scratch_dir}')
             try:
                 os.mkdir(self.scratch_dir, mode=0o755)
             except:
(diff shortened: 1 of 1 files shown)


scripts/gather_debug_info
```diff
@@ -0,0 +1,113 @@
...
+            return False
+        else:
+            return True
+
+    @catch_errors
+    def gather_os_info(self):
+        print('Gathering OS information...')
+        shutil.copy('/etc/os-release', os.path.join(self.scratch_dir, 'os-info'))
+        with open(os.path.join(self.scratch_dir, 'os-info', 'disk-info'), 'w') as f:
+            f.write('#####################\n# df -h\n#####################\n')
```
xiv, line 70: Might be worth creating a template for this since its a repeated pattern {separator} {command}  {separator} thats used 3+ times

jacekn: I converted this string into a template to make it easier. This way we won't have to introduce new variables to store command.
I also converted all other strings into f-strings format

SirTyson:
Thanks for putting this together! I think it mostly looks like we have everything we need, but there are a few extra pieces that could be useful:

1. Optionally specify a core executable path. This isn't super important because all the production use cases that we're targeting should just be using `stellar-core`, but during development we often use one of builds that specify a specific path, and this script could be useful.
2. Zip DB when using a SQLite backend. I don't think this is possible with Postgres, but I think (and could be wrong) that it's relatively easy to do in SQLite. We're in the process of getting rid of Postgres support anyway, so if you can parse the SQLite DB image path from the config and include it in the final zip result, it would be useful.
3. Attempt to parse log path from the config file. I'm not sure what log systems other folks use, but during testing runs I just use a single file as my running log. If most everyone is encoding time info into their logs then you can skip this, I might be the odd one out here. This isn't super important so feel free to skip it in the interest of time, it might be a little tricky to detect when the config specifies a deterministic log name vs. a date based one.

jacekn:
> Thanks for putting this together! I think it mostly looks like we have everything we need, but there are a few extra pieces that could be useful:
> 
>     1. Optionally specify a core executable path. This isn't super important because all the production use cases that we're targeting should just be using `stellar-core`, but during development we often use one of builds that specify a specific path, and this script could be useful.

Done

>     2. Zip DB when using a SQLite backend. I don't think this is possible with Postgres, but I think (and could be wrong) that it's relatively easy to do in SQLite. We're in the process of getting rid of Postgres support anyway, so if you can parse the SQLite DB image path from the config and include it in the final zip result, it would be useful.

Done

>     3. Attempt to parse log path from the config file. I'm not sure what log systems other folks use, but during testing runs I just use a single file as my running log. If most everyone is encoding time info into their logs then you can skip this, I might be the odd one out here. This isn't super important so feel free to skip it in the interest of time, it might be a little tricky to detect when the config specifies a deterministic log name vs. a date based one.

Done. Logs with date in the filename are a bit tricky but for this reason the script tries to get last 24h worth of files with `.log` suffix. So with this this in place I believe we'll be fine.

anupsdf:
Let me post on hayashi about the security advisory https://rustsec.org/advisories/RUSTSEC-2024-0344 since we cannot switch to curve25519-dalek 4.1.3 prior to protocol 22.

SirTyson (approved):
Seems like it's working great, thanks for working on this!

SirTyson:
We're currently changing our Github CI so we can't merge anything at the moment, but should have it resolved in a day or two.

SirTyson:
CI is back online. If you rebase this off master, the last CI check should be resolved then I can merge

jacekn:
@SirTyson all checks pass now so if it looks good from our side can you merge?

PR #4353, opened 2024-06-07
Mandate flow control in bytes

### LINKED ISSUE #3781: Make flow control in bytes required

bboston7:
Resolves #3781 and #4374.

This change removes the `ENABLE_FLOW_CONTROL_BYTES` config option and always uses flow control in bytes.

# Checklist
- [x] Compiles
- [x] Ran all tests

### DIFF
--- a/src/main/Config.cpp
+++ b/src/main/Config.cpp
@@ -151,8 +151,8 @@ Config::Config() : NODE_SEED(SecretKey::random())
     LEDGER_PROTOCOL_VERSION = CURRENT_LEDGER_PROTOCOL_VERSION;
     LEDGER_PROTOCOL_MIN_VERSION_INTERNAL_ERROR_REPORT = 18;
 
-    OVERLAY_PROTOCOL_MIN_VERSION = 32;
-    OVERLAY_PROTOCOL_VERSION = 34;
+    OVERLAY_PROTOCOL_MIN_VERSION = 33;
+    OVERLAY_PROTOCOL_VERSION = 35;
 
     VERSION_STR = STELLAR_CORE_VERSION;
 
@@ -261,7 +261,6 @@ Config::Config() : NODE_SEED(SecretKey::random())
     PEER_FLOOD_READING_CAPACITY_BYTES = 0;
     FLOW_CONTROL_SEND_MORE_BATCH_SIZE_BYTES = 0;
     OUTBOUND_TX_QUEUE_BYTE_LIMIT = 1024 * 1024 * 3;
-    ENABLE_FLOW_CONTROL_BYTES = true;
 
     // WORKER_THREADS: setting this too low risks a form of priority inversion
     // where a long-running background task occupies all worker threads and
@@ -1030,10 +1029,6 @@ Config::processConfig(std::shared_ptr<cpptoml::table> t)
             {
                 FLOW_CONTROL_SEND_MORE_BATCH_SIZE_BYTES =
                     readInt<uint32_t>(item, 1);
-            }
-            else if (item.first == "ENABLE_FLOW_CONTROL_BYTES")
-            {
-                ENABLE_FLOW_CONTROL_BYTES = readBool(item);
             }
             else if (item.first == "OUTBOUND_TX_QUEUE_BYTE_LIMIT")
             {
--- a/src/overlay/Peer.cpp
+++ b/src/overlay/Peer.cpp
@@ -156,14 +156,10 @@ Peer::endMessageProcessing(StellarMessage const& msg)
     // We may release reading capacity, which gets taken by the background
     // thread immediately, so we can't assert `canRead` here
     auto res = mFlowControl->endMessageProcessing(msg);
-    if (res.second)
+    if (res.first > 0 || res.second > 0)
     {
         sendSendMore(static_cast<uint32>(res.first),
-                     static_cast<uint32>(*res.second));
-    }
-    else if (res.first > 0)
-    {
-        sendSendMore(static_cast<uint32>(res.first));
+                     static_cast<uint32>(res.second));
     }
 
     // Now that we've released some capacity, maybe schedule more reads
@@ -389,10 +385,7 @@ Peer::sendAuth()
     ZoneScoped;
     StellarMessage msg;
     msg.type(AUTH);
-    if (mAppConnector.getConfig().ENABLE_FLOW_CONTROL_BYTES)
-    {
-        msg.auth().flags = AUTH_MSG_FLAG_FLOW_CONTROL_BYTES_REQUESTED;
-    }
+    msg.auth().flags = AUTH_MSG_FLAG_FLOW_CONTROL_BYTES_REQUESTED;
     auto msgPtr = std::make_shared<StellarMessage const>(msg);
     sendMessage(msgPtr);
 }
@@ -591,18 +584,6 @@ Peer::sendErrorAndDrop(ErrorCode error, std::string const& message)
     releaseAssert(threadIsMain());
     sendError(error, message);
     drop(message, DropDirection::WE_DROPPED_REMOTE);
-}
-
-void
-Peer::sendSendMore(uint32_t numMessages)
-{
-    ZoneScoped;
-    releaseAssert(threadIsMain());
-
-    auto m = std::make_shared<StellarMessage>();
-    m->type(SEND_MORE);
-    m->sendMoreMessage().numMessages = numMessages;
-    sendMessage(m);
 }
 
 void
@@ -1597,6 +1578,7 @@ Peer::recvHello(Hello const& elo)
     mRemoteOverlayVersion = elo.overlayVersion;
     mRemoteVersion = elo.versionStr;
     mPeerID = elo.peerID;
+    mFlowControl->setPeerID(mPeerID);
     mRecvNonce = elo.nonce;
     mHmac.setSendMackey(peerAuth.getSendingMacKey(elo.cert.pubkey, mSendNonce,
                                                   mRecvNonce, mRole));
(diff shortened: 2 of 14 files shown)


src/main/Config.cpp
```diff
@@ -1033,7 +1032,23 @@ Config::processConfig(std::shared_ptr<cpptoml::table> t)
...
+                        "file.");
+                }
+                else
+                {
+                    CLOG_ERROR(Overlay,
+                               "ENABLE_FLOW_CONTROL_BYTES is now mandatory. "
+                               "Overriding to `true`. This config option will "
+                               "be removed in a future release. Please remove "
+                               "the option from your configuration file.");
+                }
```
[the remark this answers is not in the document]
SirTyson: I think ignoring is fine here. Unlike the BucketListDB config migrations, `ENABLE_FLOW_CONTROL_BYTES` being silently enabled won't break any operators, so we can just warn for a bit then remove the flag.

src/overlay/Peer.cpp
```diff
@@ -389,10 +389,15 @@ Peer::sendAuth()
...
     StellarMessage msg;
     msg.type(AUTH);
-    if (mAppConnector.getConfig().ENABLE_FLOW_CONTROL_BYTES)
+#ifdef BUILD_TESTS
+    if (!mAppConnector.getOverlayManager()
+             .isFlowControlBytesDisabledForTesting())
     {
         msg.auth().flags = AUTH_MSG_FLAG_FLOW_CONTROL_BYTES_REQUESTED;
     }
+#else
```
SirTyson, line 398: Nit: This `#else` isn't necessary, you can just wrap the if expression:
```
#ifdef BUILD_TESTS
    if (!mAppConnector.getOverlayManager()
             .isFlowControlBytesDisabledForTesting())
#endif
```

src/overlay/Peer.cpp
```diff
@@ -1740,15 +1745,17 @@ Peer::recvAuth(StellarMessage const& msg)
...
-        (mAppConnector.getConfig().OVERLAY_PROTOCOL_VERSION >=
-             Peer::FIRST_VERSION_SUPPORTING_FLOW_CONTROL_IN_BYTES &&
-         getRemoteOverlayVersion() >=
-             Peer::FIRST_VERSION_SUPPORTING_FLOW_CONTROL_IN_BYTES);
+    // NOTE: Once min overlay version is 35 we can remove this check
     bool bothWantBytes =
-        enableBytes &&
-        msg.auth().flags == AUTH_MSG_FLAG_FLOW_CONTROL_BYTES_REQUESTED &&
-        mAppConnector.getConfig().ENABLE_FLOW_CONTROL_BYTES;
+        msg.auth().flags == AUTH_MSG_FLAG_FLOW_CONTROL_BYTES_REQUESTED;
```
SirTyson, line 1750: Should we change this check to 
```
 bool bothWantBytes =
        msg.auth().flags == AUTH_MSG_FLAG_FLOW_CONTROL_BYTES_REQUESTED || getRemoteOverlayVersion() >= 35;
```

That way once the minimum protocol version >= 35, we can just stop sending/checking AUTH_MSG_FLAG_FLOW_CONTROL_BYTES_REQUESTED entirely, since the flag would be made redundant.

SirTyson (approved):
LGTM, one comment on changing a flow control bytes check to make deprecation easier.

src/overlay/Peer.cpp
bboston7: Yeah, I think that's a good idea. I made that change.

src/overlay/Peer.cpp
bboston7: Done!

src/main/Config.cpp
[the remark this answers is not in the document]
bboston7: Sounds good! I left this as-is.

SirTyson (dismissed):
LGTM, thanks for the changes!

src/main/Config.cpp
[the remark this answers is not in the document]
marta-lokhova: Tbh I doubt anyone would be affected by the removal of this flag, since it's sort of an advanced low-level config. You can probably just remove it without graceful deprecation, but we do need to make sure that our own infra isn't using the flag somewhere (to avoid crashing on startup during the release).

marta-lokhova (changes requested):
I think you can go further with this change and remove support for flow control in bytes completely:
* Minimum overlay version is far beyond `FIRST_VERSION_SUPPORTING_FLOW_CONTROL_IN_BYTES` at this point, so most (if not all) nodes on the network will have the feature enabled
* I doubt there are any nodes outside of our infra that are manually tuning this config. Looking at a snapshot of peers on one of the watchers, I don't see anyone not using the feature.
* I think we can remove any conditional enabling of flow control in bytes, and just implicitly set it to true. You might be able to get rid of the annoying FlowControl::start call at the end of authentication as well, and basically start flow control automatically when we create Peer.
* We also don't need to set `AUTH_MSG_FLAG_FLOW_CONTROL_BYTES_REQUESTED` anymore. I'd suggest we set auth flags back to 0, so we can use it for future feature rollout. `AUTH_MSG_FLAG_FLOW_CONTROL_BYTES_REQUESTED` can be deprecated.
* We still need to bump overlay version. While we're at it: the network recently upgraded to v21, so we can actually bump minimum overlay version as well to the one that supports v21 (meaning that we can drop all v20 clients now)
* Dropping support for this flag means we can drop all the tests dealing with mixed modes. Yay to code removal!

src/main/Config.cpp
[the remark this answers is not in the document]
bboston7: I removed this flag

bboston7:
I fully removed the option to disable flow control in bytes and simplified the flow control code. I also bumped the minimum overlay version. This is now ready for another round of review!

src/overlay/OverlayManagerImpl.cpp
```diff
@@ -359,8 +359,6 @@ OverlayManagerImpl::start()
 OverlayManager::AdjustedFlowControlConfig
 OverlayManagerImpl::getFlowControlBytesConfig() const
 {
-    auto const maxTxSize = mApp.getHerder().getMaxTxSize();
```
marta-lokhova, line 362: Is there a reason for this code move? i think it was placed here originally to ensure that by the time this function is called, we've properly loaded network configs

src/overlay/Peer.cpp
```diff
@@ -1746,35 +1728,13 @@ Peer::recvAuth(StellarMessage const& msg)
         return;
     }
 
-    bool enableBytes =
-        (mAppConnector.getConfig().OVERLAY_PROTOCOL_VERSION >=
-             Peer::FIRST_VERSION_SUPPORTING_FLOW_CONTROL_IN_BYTES &&
-         getRemoteOverlayVersion() >=
-             Peer::FIRST_VERSION_SUPPORTING_FLOW_CONTROL_IN_BYTES);
-    bool bothWantBytes =
```
marta-lokhova, line 1754: Should we still check if a peer has AUTH_MSG_FLAG_FLOW_CONTROL_BYTES_REQUESTED enabled, and drop it if not? This way we can add a drop reason to the error for clarity

src/main/Config.cpp
```diff
@@ -151,7 +151,7 @@ Config::Config() : NODE_SEED(SecretKey::random())
     LEDGER_PROTOCOL_VERSION = CURRENT_LEDGER_PROTOCOL_VERSION;
     LEDGER_PROTOCOL_MIN_VERSION_INTERNAL_ERROR_REPORT = 18;
 
-    OVERLAY_PROTOCOL_MIN_VERSION = 32;
+    OVERLAY_PROTOCOL_MIN_VERSION = 33;
     OVERLAY_PROTOCOL_VERSION = 34;
```
marta-lokhova, line 155: We should bump current overlay version as well

marta-lokhova (changes requested):

src/overlay/Peer.cpp
bboston7: Good idea. Done in 9e862982a98e7f04fd2ce8b72618c7fe8977334b

src/main/Config.cpp
bboston7: Done

src/overlay/OverlayManagerImpl.cpp
bboston7: Yes, I moved the assert only to the branch where `maxTxSize` is actually used because otherwise this assert fires during certain tests that start `OverlayManager` without starting `Herder`. For example, [`database is purged at overlay start`](https://github.com/bboston7/stellar-core/blob/f04f0c19a18067f6af406d5c4c0fae8e72f2d804/src/overlay/test/OverlayTests.cpp#L2201).

The reason the assert fires is that I removed `FlowControl::start`, so now [`FlowControl` calls `getFlowControlBytesConfig()` in its constructor](https://github.com/bboston7/stellar-core/blob/f04f0c19a18067f6af406d5c4c0fae8e72f2d804/src/overlay/FlowControl.cpp#L39), which itself is called in the `Peer` constructor. That means tests that start `OverlayManager` without starting `Herder` first don't know the max tx size and so `getMaxTxSize` returns `0`. I think the only reason the tests didn't run into this before is that they didn't authenticate peers (or just didn't run long enough to get to that point) and so never called `FlowControl::start`.

I fixed this by explicitly setting `cfg.PEER_FLOOD_READING_CAPACITY_BYTES` and `cfg.FLOW_CONTROL_SEND_MORE_BATCH_SIZE_BYTES` in those tests and moving the check to the case where `getFlowControlBytesConfig` needs to use `maxTxSize` to automatically set flow control parameters. I'm open to other ideas though if you'd like to restore this check. Maybe there's a different default we could use in the `FlowControl` constructor that doesn't require `Herder` to be running?

src/overlay/test/OverlayTests.cpp
```diff
@@ -569,61 +546,21 @@ TEST_CASE("loopback peer flow control activation", "[overlay][flowcontrol]")
...
-                    conn.getAcceptor()->sendSendMore(0, 0);
-                }
-                else
-                {
-                    conn.getAcceptor()->sendSendMore(0);
-                }
-                dropReason = fcInBytes ? "invalid message SEND_MORE_EXTENDED"
-                                       : "invalid message SEND_MORE";
-            }
-            SECTION("invalid message type")
```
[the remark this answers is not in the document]
bboston7: I added this section back, this time manually constructing a `SEND_MORE` in f1111b9b02a5b2936b46a1df632cc8fc7cc7ceb0. It found a bug in a log statement that assumed the message was a `SEND_MORE_EXTENDED`, so I fixed that too and also cleaned up some conditionals/asserts that were overly permissive with `SEND_MORE`.

src/overlay/test/OverlayTests.cpp
```diff
@@ -635,27 +572,17 @@ TEST_CASE("loopback peer flow control activation", "[overlay][flowcontrol]")
...
-            SECTION("one disables")
-            {
-                cfg1.ENABLE_FLOW_CONTROL_BYTES = false;
-                runTest({cfg1, cfg2}, false);
-            }
-        }
-    };
+    SECTION("basic")
+    {
+        // Flow control without illegal SEND_MORE
```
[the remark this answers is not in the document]
bboston7: It's accurate again now after re-adding the `SEND_MORE` test!

src/overlay/OverlayManagerImpl.cpp
marta-lokhova: That makes sense for the tests! In prod paths, by the time we're calling this function, we need to be sure that maxTsSize is setup (regardless of whether it's used or not), so the assert also checks that we have already loaded Soroban config. Looking at the tests, `createTestApplication` starts the app by default, and there seems to be a small number of tests that don't do that. Perhaps we could force set mMaxTxSize in those tests manually and keep the assert at the top of the function? We could introduce a new function behind BUILD_TESTS flag similar to `setMaxClassicTxSize`. Wdyt?

bboston7: I restored the assert in 80c7cba06f73f1b869fb95efe35596162405af85 and added a `setMaxTxSize` function behind `BUILD_TESTS` to fix up the tests.

marta-lokhova (approved):

PR #4358, opened 2024-06-21
Update overlay survey script with lessons learned during testnet run

bboston7:
# Description

This change makes a few tweaks to the overlay survey script to fix some small things I noticed after running it on testnet:

* Changes the script's end condition to depend only on responses, and not requests. Without this it was possible for the survey script to run for the full duration of the collecting phase (2 hours) if a node with more than 25 peers stopped responding after the surveyor received the first set of peers.
* Downgrades the severity of "node already in backlog" messages from `error` to `debug`. This is an expected condition that I simply forgot to special-case before.
* Modifies the simulator to occasionally return "node already in backlog" messages to test the script against that case.
* Adds a `--fast` option to the `simulate` mode that skips any `sleep` calls. This makes the script much nicer to test.
* Fixes naming of graphml fields to match JSON result fields.
  * I did most of this in the V2 script update, but missed a couple spots.
  * Most of this change is in the simulator to support the new field names.

# Checklist
- [ ] Compiles
- [ ] Ran all tests

### DIFF
--- a/scripts/OverlaySurvey.py
+++ b/scripts/OverlaySurvey.py
@@ -84,15 +84,14 @@ SIMULATION = None
 # internal limit.
 MAX_COLLECT_DURATION = 30
 
-# Maximum number of consecutive rounds in which the surveyor neither sent
-# requests to nor received responses from any nodes. A round contains a batch of
-# requests sent to select nodes, followed by a wait period of 15 seconds,
-# followed by checking for responses and building up the next batch of requests
-# to send. Therefore, a setting of `8` is roughly 2 minutes of inactivity
-# before the script considers the survey complete. This is necessary because
-# it's very likely that not all surveyed nodes will respond to the survey.
-# Therefore, we need some cutoff after we which we assume those nodes will never
-# respond.
+# Maximum number of consecutive rounds in which the surveyor does not receive
+# responses from any nodes. A round contains a batch of requests sent to select
+# nodes, followed by a wait period of 15 seconds, followed by checking for
+# responses and building up the next batch of requests to send. Therefore, a
+# setting of `8` is roughly 2 minutes of inactivity before the script considers
+# the survey complete. This is necessary because it's very likely that not all
+# surveyed nodes will respond to the survey.  Therefore, we need some cutoff
+# after we which we assume those nodes will never respond.
 MAX_INACTIVE_ROUNDS = 8
 
 def get_request(url, params=None):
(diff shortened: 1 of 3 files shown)


scripts/OverlaySurvey.py
```diff
@@ -405,25 +415,33 @@ def run_survey(args):
...
                         # Received a new response!
                         logger.debug("Received response from %s", key)
                         inactive_rounds = 0
                         heard_from.add(key)
+                    elif key in incomplete_responses and len(node_data) > 0:
+                        # Received additional data for a node that previously
+                        # responded
+                        logger.debug("Received additional data for %s", key)
+                        inactive_rounds = 0
+                        incomplete_responses.remove(key)
```
marta-lokhova, line 444: Do we need to make sure we received all data from the node before removing it from `incomplete_responses`?

scripts/overlay_survey/simulation.py
```diff
@@ -231,11 +252,19 @@ def _getsurveyresult(self, params):
...
                                len(out_edges),
-                               node_json["numTotalOutboundPeers"])
-                node_json["numTotalOutboundPeers"] = len(out_edges)
+                               node_json["totalOutbound"])
+                node_json["totalOutbound"] = len(out_edges)
 
             _add_v2_survey_data(node_json)
 
+            # Rename peer count fields to match stellar-core's response
+            _rename_dictionary_field(node_json,
```
marta-lokhova, line 261: Why do we need this rename? (previous version of code used numTotalInboundPeers/numTotalOutboundPeers so no rename was necessary from what I see)

marta-lokhova (changes requested):
Thanks! Just a few questions. Looks like the PR needs to be rebased also to re-trigger CI.

marta-lokhova:
any update on this change? would be nice to land it in v21.2.0 (starts today), so we can use new-style script for pubnet.

scripts/overlay_survey/simulation.py
bboston7: I changed the output fields in the graph to match the JSON output when I initially updated the script for the V2 survey. That means that `numTotalInboundPeers`/`numTotalOutboundPeers` becomes `totalInbound` and `totalOutbound`. I don't have a preference for which set of field names we use (that is, I could change it the other way around so JSON output becomes `numTotalInboundPeers`/`numTotalOutboundPeers`), but I do think the two outputs should have the same field names.

scripts/OverlaySurvey.py
bboston7: No. If there's more data the node will be re-added to `incomplete_responses` here: https://github.com/stellar/stellar-core/blob/6979693a8dab50d3efd694444048768cd211d534/scripts/OverlaySurvey.py#L478

bboston7:
> any update on this change? 

I just responded to your comments and rebased the change

scripts/OverlaySurvey.py
marta-lokhova: ah, makes sense. thanks!

marta-lokhova (approved):

PR #4361, opened 2024-06-25
Improve performance of dump-ledger

MonsieurNicolas:
# Description

This PR changes how dump-ledger operate as to use a lot less memory:

previously running a command such as `stellar-core dump-ledger --output-file out.json --group-by "data.type" --agg "count()"` against pubnet would crash nodes by using up all memory (16GB).

This version uses much less memory (peaks at 4GB) by not storing the current bucket in memory and not storing all seen keys from the bucket list in RAM.

This PR also includes a couple compile/link fixes for Windows.


dmkozh:
Looks good, thanks for the optimization. In order to get past the cargo-deny check you just need to change deny.toml like here: https://github.com/dmkozh/stellar-core/commit/aa935a95b6055c159642cd4673f6bf417e30d7bd#diff-1040309c64844eb1b6b63d8fd67938adbf9461f1b3c61f12cf738f064a02d3de

dmkozh (approved):

PR #4367, opened 2024-06-27
Rework github action caching

graydon:
Quite a lot is going wrong with our current github actions artifact caching:

  - I'm not sure if it's been getting any cache hits on PRs for a long time. A PR hit would need a master-seeded cache and those are not being generated.
  - The "hour-number-qualified cache key" we were generating was/is both inaccurate enough to represent changes we _want_ to invalidate the cache, and also redundant: github always falls back to recency-order when accessing prefix restore keys on a cache miss
  - Latobarita's auto branch was getting hits, and getting seeded, but we're not running latobarita anymore.
  - The PR branches were making their own cache entries so they'd get a hit against themselves if you did multiple pushes but only then. Otherwise they'd be filling up the cache set with cache entries that never get hits again.
  - The new merge queues are the same: never hit, fill up the cache set with stale entries.
  - We weren't caching any of the Rust stuff

So this PR overhauls this quite a bit:
  - We split cache probing, retrieving and saving
  - We key the cache on OS/toolchain/protocol as before (and use that prefix for restore-prefix fallback) but use the commit SHA as the exact-hit suffix
  - We also add a run on master on a cron schedule, every hour
  - We only save the cache at the end of a run on schedule, on master, so no pollution from other branches
  - All runs probe for a cache hit on the commit SHA, and do nothing if there is a hit: we already tested this SHA and saved its cache. So the hourly run should be very cheap/quick/quiet most of the time
  - Any cache _miss_ falls back to restoring from the last cache written, which will have been written on master on schedule and be fairly similar to whatever the runner is doing (any PR has to be fairly close to it to work)
  - We cache the Rust stuff

Hopefully this will speed up everything. Suggestions on further improvements welcome.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -136,6 +136,12 @@ jobs:
       - name: install rustup components
         if: steps.cache.outputs.cache-hit != 'true'
         run: rustup component add rustfmt
+      - name: install cargo-cache
+        if: steps.cache.outputs.cache-hit != 'true'
+        run: cargo install --locked cargo-cache --version 0.8.3
+      - name: install cargo-sweep
+        if: steps.cache.outputs.cache-hit != 'true'
+        run: cargo install --locked cargo-sweep --version 0.7.0
       - name: install dependencies
         if: steps.cache.outputs.cache-hit != 'true'
         run: sudo apt-get -y install postgresql git build-essential pkg-config autoconf automake libtool bison flex libpq-dev parallel libunwind-dev sed perl
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -4,8 +4,12 @@ on:
   pull_request:
   merge_group:
   push:
-    branches:
-      - auto
+
+  # At 17 minutes past the hour every hour we do a run to prime the cache.
+  # If nothing has changed, the commit SHA will cause a cache-probe hit and
+  # this run will conclude the cache is already primed and exit quickly.
+  schedule:
+    - cron: '17 * * * *'
 
 jobs:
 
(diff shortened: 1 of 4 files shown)


MonsieurNicolas:
The reason for not using commit SHA and instead use some "stable for an hour" ID was that there is a global limit per repo that is fairly small (like 5GB), which when there is enough churn (if a few people push to their branches it may be enough to hit the limit) to purge keys that are actually useful (like the master cache in your case).
I think each cache is about 250MB per toolchain combination so 1 GB cache addition every time (gcc/clang*cur/next).

In PRs there should have been caching advantage but only when pushing new commits. Now that we also have a cache on master it should work better for sure.

.github/workflows/build.yml
```diff
@@ -51,14 +58,22 @@ jobs:
...
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v3
     - uses: stellar/actions/rust-check-git-rev-deps@main
 
   build:
     runs-on: ubuntu-latest
-
+    env:
+      CACHED_PATHS: |
```
MonsieurNicolas, line 70: the ci script [purges the cache from time to time](https://github.com/stellar/stellar-core/blob/be1b3ee138ff75c42bfae8c90cec111feffad785/ci-build.sh#L131) (on top of ccache being configured to not grow indefinitely).
How can we make sure that this cargo cache does not grow too much?
On my dev box (before running some `cargo cache` commands, my cache was in the GBs)

graydon: Ok I've added code to both incrementally clear rust caches and join in the purge-once-a-month ccache party. This requires installing two (!) separate cargo extensions, one to manage its downloads (cargo-cache) and another to manage its build artifacts (cargo-sweep).

MonsieurNicolas (approved):

MonsieurNicolas (approved):

PR #4369, opened 2024-06-27
Bucketlist db invariant

### LINKED ISSUE #4368: BucketListIsConsistentWithDatabase broken with BucketListDB
`BucketListIsConsistentWithDatabase` does not currently work with BucketListDB. Note that the invariant is not violated, but the logic to check the invariant itself is not compatible with BucketListDB.

SirTyson:
# Description

Resolves #4368

Fixes `BucketListIsConsistentWithDatabase` when BucketListDB is enabled.

# Checklist
- [x] Compiles
- [x] Ran all tests

### DIFF
--- a/src/invariant/BucketListIsConsistentWithDatabase.cpp
+++ b/src/invariant/BucketListIsConsistentWithDatabase.cpp
@@ -5,6 +5,7 @@
 #include "invariant/BucketListIsConsistentWithDatabase.h"
 #include "bucket/Bucket.h"
 #include "bucket/BucketInputIterator.h"
+#include "bucket/BucketList.h"
 #include "bucket/BucketManager.h"
 #include "crypto/Hex.h"
 #include "history/HistoryArchive.h"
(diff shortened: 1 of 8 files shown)


src/invariant/BucketListIsConsistentWithDatabase.cpp
```diff
@@ -258,6 +259,96 @@ BucketListIsConsistentWithDatabase::checkEntireBucketlist()
...
+                }
+            }
+        }
+    }
+
+    auto range =
+        LedgerRange::inclusive(LedgerManager::GENESIS_LEDGER_SEQ, newestLedger);
+
+    // SQL only stores offers when BucketListDB is enabled
+    return counts.checkDbEntryCounts(
```
dmkozh, line 348: I'm not sure if that matters much, but could that check go to `checkOnBucketApply`? It seems like that would be a bit more consistent.

SirTyson: No, this check makes sure the number of entries in SQL DB matches the number of entries in the BucketList. This is not possible to check on a per-bucket level when applying buckets in-order. For example, suppose an offer `DEADENTRY` exists on level 0, and a `LIVEENTRY` of the same key exists on level 1. The database correctly does not contain the entry such that the offer count is 0. When performing the check on the level 1 bucket, the invariant would fail, as based on that individual bucket, the offer count should be 1 instead of 0.

Essentially we can only do individual Bucket level checks in `checkOnBucketApply`, whereas `checkAfterAssumeState` performs BucketList level checks.

dmkozh: Thanks for the explanation.

dmkozh (approved):

PR #4372, opened 2024-06-28
Fail loadgen when pending accounts are found in available accounts list

bboston7:
# Description

Although `mAccountsAvailable` shouldn't contain pending accounts, it is possible when the network is overloaded. Consider the following scenario:
1. A node generates a transaction `t` using account `a` and broadcasts it on. In doing so, loadgen marks `a` as in use, removing it from `mAccountsAvailable.
2. For whatever reason, `t` never makes it out of the queue and the node bans it.
3. After some period of time, the node unbans `t` because bans only last for so many ledgers.
4. Loadgen marks `a` available, moving it back into `mAccountsAvailable`.
5. The node hears about `t` again on the network and (as it is no longer banned) adds it back to the queue
6. getNextAvailableAccount draws `a` from `mAccountsAvailable`. However, `a` is no longer available as `t` is in the transaction queue!

In this scenario, returning `a` results in an assertion failure later. This change instead detects this and marks the loadgen run as failed.

# Checklist
- [x] Compiles
- [x] Ran all tests

### DIFF
--- a/src/simulation/LoadGenerator.cpp
+++ b/src/simulation/LoadGenerator.cpp
@@ -684,14 +684,7 @@ LoadGenerator::generateLoad(GeneratedLoadConfig cfg)
                 return;
             }
 
-            std::optional<uint64_t> maybeSourceAccountId =
-                getNextAvailableAccount(ledgerNum);
-            if (!maybeSourceAccountId.has_value())
-            {
-                // Loadgen has failed
-                break;
-            }
-            uint64_t sourceAccountId = maybeSourceAccountId.value();
+            uint64_t sourceAccountId = getNextAvailableAccount(ledgerNum);
 
             std::function<
                 std::pair<LoadGenerator::TestAccountPtr, TransactionFramePtr>()>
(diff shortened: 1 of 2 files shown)


src/simulation/LoadGenerator.cpp
```diff
@@ -988,10 +1040,21 @@ std::pair<LoadGenerator::TestAccountPtr, TransactionFramePtr>
...
-        mInitialAccountsCreated
-            ? findAccount(getNextAvailableAccount(), ledgerNum)
-            : mRoot;
+    TestAccountPtr sourceAcc = mRoot;
+    if (mInitialAccountsCreated)
+    {
+        std::optional<uint64_t> accountId = getNextAvailableAccount(ledgerNum);
+        if (!accountId.has_value())
+        {
+            // This really shouldn't happen and indicates a misconfiguration
```
marta-lokhova, line 1049: Given that this shouldn't happen, should we keep the code as-is and allow it to crash? Otherwise, the exception suggests that this can be remedied by adjusting parameters (but this just hides a bug in sheduleLoadGeneration)

src/simulation/LoadGenerator.cpp
```diff
@@ -915,6 +927,46 @@ LoadGenerator::getNextAvailableAccount()
...
+        // We chose option (2), as this scenario is indicative of an overloaded
+        // network and in practice loadgen is highly likely to fail soon anyway.
+        // Moreover, option (1) provides some false sense of having resolved the
+        // issue as this is a race condition that could still occur between
+        // returning the new account from this function and using it later on.
+        CLOG_WARNING(
+            LoadGen,
+            "mAccountsAvailable contains an account '{}' already in use",
+            sourceAccountId);
+        mFailed = true;
```
marta-lokhova, line 966: We could also just pick up another account to use to avoid failing loadgen? This edge case seems quite rare and hard to trigger, so I think we should try to keep loadgen alive if possible (I don't think we can draw conclusions on network being overloaded based on transactions re-surfacing)

marta-lokhova: (this way getNextAvailableAccount would always return something instead of an optional, so we won't need to check for nullopt in other places)

src/simulation/LoadGenerator.cpp
```diff
@@ -915,6 +927,46 @@ LoadGenerator::getNextAvailableAccount()
...
+        //
+        // In this scenario, returning `a` results in an assertion failure
+        // later. At this point we have two reasonable options:
+        // 1. Mark `a` in-use and resample new accounts until we either find one
+        //    that is available, or run out of accounts
+        // 2. Fail loadgen.
+        // We chose option (2), as this scenario is indicative of an overloaded
+        // network and in practice loadgen is highly likely to fail soon anyway.
+        // Moreover, option (1) provides some false sense of having resolved the
+        // issue as this is a race condition that could still occur between
```
marta-lokhova, line 960: we're creating and submitting transactions synchronously, so I think there's currently no race condition here, right?

marta-lokhova (changes requested):

src/simulation/LoadGenerator.cpp
bboston7: Hmm, I was thinking that in between picking `a` and submitting the new transaction, the old transaction `t` could come in on the overlay. But I think you're right this isn't a problem because there's only one thread here and the overlay can't interleave between between these two operations. I'll change this to sample a new account instead.

bboston7: I just pushed a change that resamples instead

src/simulation/LoadGenerator.cpp
bboston7: I just pushed a change that resamples instead, and dropped the `nullopt`

src/simulation/LoadGenerator.cpp
bboston7: I reverted this. This check isn't necessary anymore after getting rid of the `std::optional`

marta-lokhova (approved):
LGTM, thanks!

PR #4383, opened 2024-07-11
Cleanup outdated surge pricing logic.

dmkozh:
# Description

Cleanup outdated surge pricing logic.

- Remove tx stacks in surge pricing queue as we have only a single tx/account
- Remove lazy fee computation for legacy tx sets (not sure it was necessary before)
- Remove tests covering surge pricing logic for old protocols. Replay doesn't need to emulate surge pricing.

# Checklist
- [ ] Compiles
- [ ] Ran all tests


src/herder/TxQueueLimiter.h
```diff
@@ -13,51 +13,14 @@
...
-};
-
 class TxQueueLimiter
 {
     // number of ledgers we can pool in memory
     uint32 const mPoolLedgerMultiplier;
     LedgerManager& mLedgerManager;
 
-    UnorderedMap<TransactionFrameBasePtr, TxStackPtr> mStackForTx;
+    UnorderedMap<TransactionFrameBasePtr, SurgePricingQueueTxBasePtr>
```
marta-lokhova, line 22: isn't this whole data structure redundant now?

src/herder/TxQueueLimiter.h
```diff
@@ -73,10 +36,6 @@ class TxQueueLimiter
     // limits.
     std::shared_ptr<SurgePricingLaneConfig> mSurgePricingLaneConfig;
 
-    // Quick lookup of relevant account IDs, needed temporary to maintain
```
marta-lokhova, line 76: The comment says this can be removed when the stacks are removed, but now where are we actually enforcing 1 tx per account in tx limiter? I think this was a valid sanity check (that helped catch tx queue bugs)

src/herder/TxSetUtils.cpp
```diff
@@ -157,114 +138,33 @@ TxSetUtils::buildAccountTxQueues(TxSetTransactions const& txs)
...
+        ltx.loadHeader().current().ledgerVersion, ProtocolVersion::V_20));
+    // This is done so minSeqLedgerGap is validated against the next
+    // ledgerSeq, which is what will be used at apply time
+    ltx.loadHeader().current().ledgerSeq =
+        app.getLedgerManager().getLastClosedLedgerNum() + 1;
 
-    UnorderedMap<AccountID, int64_t> accountFeeMap;
     TxSetTransactions invalidTxs;
 
-    auto accountTxQueues = buildAccountTxQueues(txs);
```
marta-lokhova, line 178: nice!

marta-lokhova (changes requested):

src/herder/TxQueueLimiter.h
dmkozh: It's still necessary because we need to remove containers from the queue using the tx pointer. Inheritance would've helped here, but I still think it's the lesser evil.

src/herder/TxSetUtils.h
```diff
@@ -13,22 +13,19 @@
...
-    void popTopTx() override;
-    Resource getResources() const override;
-
-    std::deque<TransactionFrameBasePtr> mTxs;
+    TransactionFrameBasePtr getTopTx() const;
+    bool empty() const;
+    void popTopTx();
 
   private:
+    std::deque<TransactionFrameBasePtr> mTxs;
```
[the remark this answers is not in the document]
marta-lokhova: ah, because we need to re-sort for apply. i guess we have to keep this logic around, you're right.

src/herder/TxQueueLimiter.h
marta-lokhova: i think it was a somewhat useful sanity check, but i don't really feel strongly. i'm ok with removing it.

marta-lokhova (changes requested):

src/herder/TxQueueLimiter.h
dmkozh: There are several users of custom TxContainer implementations (2 now, 1 more for necessary parallel tx building), but these can be reasonably simply replaced with the maps. But in the end this seems to reduce the line count, so the change is probably for the better.
FWIW surge pricing queue is not quite an iterator, it's also a container (as in, it has to store data), so the fundamental issue is whether it allows storing data in a more or less flexible fashion. I think using more rigid interface + maps is fine.

src/herder/SurgePricingUtils.cpp
```diff
@@ -143,7 +131,7 @@ SurgePricingPriorityQueue::SurgePricingPriorityQueue(
     : mComparator(isHighestPriority, comparisonSeed)
     , mLaneConfig(settings)
     , mLaneLimits(mLaneConfig->getLaneLimits())
-    , mTxStackSets(mLaneLimits.size(), TxStackSet(mComparator))
+    , mTxSets(mLaneLimits.size(), TxSet(mComparator))
```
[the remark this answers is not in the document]
dmkozh: Even without containers this is indeed somewhat confusing, so renamed to `TxSortedSet` as to not confuse with `TxSet`

src/herder/TxSetFrame.cpp
```diff
@@ -1136,24 +1157,17 @@ ApplicableTxSetFrame::computeTxFees(
...
 
 std::optional<int64_t>
 ApplicableTxSetFrame::getTxBaseFee(TransactionFrameBaseConstPtr const& tx,
                                    LedgerHeader const& lclHeader) const
 {
-    if (std::any_of(mFeesComputed.begin(), mFeesComputed.end(),
-                    [](bool comp) { return !comp; }))
-    {
-        releaseAssert(!isGeneralizedTxSet());
-        computeTxFeesForNonGeneralizedSet(lclHeader);
```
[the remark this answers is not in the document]
marta-lokhova: sounds good, if you're doing the fee computation cleanup in a subsequent PR, I'll resolve this thread.

src/herder/TxSetFrame.cpp
```diff
@@ -1343,7 +1355,7 @@ ApplicableTxSetFrame::addTxsFromXdr(
     auto& phaseTxs = mTxPhases.at(static_cast<int>(phase));
     size_t oldSize = phaseTxs.size();
     phaseTxs.reserve(oldSize + txs.size());
-
+    auto& inclusionFeeMap = mPhaseInclusionFeeMap[static_cast<size_t>(phase)];
```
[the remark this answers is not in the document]
marta-lokhova: it just looks kinda verbose (also some places you access with `at` and some with [] which have different guarantees), so i was wondering about a centralized place for fee access.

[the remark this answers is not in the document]
dmkozh: I've added mutable getter for the time being, it should be easy to remove going forward.

marta-lokhova (approved):
LGTM, thanks for the updates! Could you please squash commits?

dmkozh:
> LGTM, thanks for the updates! Could you please squash commits?

Done

Issue #4362, opened 2024-06-25
outbound_queue_* metrics should track total time to batch and send a message


marta-lokhova:
https://github.com/stellar/stellar-core/pull/4258 restructured flow control queuing quite a bit, such that we first create a batch then send it. As a result, outbound_queue_* metrics now only track batch creation rather than batching+send total time. We should fix this by also counting send time.

PR #4386, opened 2024-07-15
Make outbound queue metrics more accurate

marta-lokhova:
Resolves https://github.com/stellar/stellar-core/issues/4362

### DIFF
--- a/src/overlay/FlowControl.cpp
+++ b/src/overlay/FlowControl.cpp
@@ -103,14 +103,14 @@ FlowControl::maybeReleaseCapacity(StellarMessage const& msg)
     }
 }
 
-std::vector<std::shared_ptr<StellarMessage const>>
+std::vector<FlowControl::QueuedOutboundMessage>
 FlowControl::getNextBatchToSend()
 {
     ZoneScoped;
     releaseAssert(!threadIsMain() || !mUseBackgroundThread);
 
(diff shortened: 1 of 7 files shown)


src/overlay/FlowControl.cpp
```diff
@@ -199,6 +189,42 @@ FlowControl::getNextBatchToSend()
     return batchToSend;
 }
 
+void
+FlowControl::updateMsgMetrics(std::shared_ptr<StellarMessage const> msg,
+                              VirtualClock::time_point const& timePlaced)
+{
```
SirTyson, line 195: Why do we need to acquire the lock here? The metrics class is thread safe, as well as the medida metric classes themselves. is `mAppConnector.now()` not thread safe?

SirTyson (approved):
LGTM, one question

src/overlay/FlowControl.cpp
marta-lokhova: it's not strictly needed (right now), i just put it to be consistent with the rest of the flow control class and avoid foot guns in the future in case this function gets modified.

SirTyson: Cool, could you add a comment clarifying this? I was a little confused at first.

marta-lokhova: Done!

SirTyson (approved):
LGTM, thanks!

PR #4389, opened 2024-07-18
Binary fuse filter

SirTyson:
# Description

Addresses #4393.

This PR adds a Binary Fuse Filter library and switches BucketListDB to use the new filter.

The Binary Fuse Filter library is based on [this](https://github.com/FastFilter/xor_singleheader) library, but I ended up essentially rewriting it in C++. The original library had a bug where some filter bytes were non-deterministic. While this did not affect correctness or the false positive rate, we need determinism because State Archival will hash these filters into the ledger. Additionally, the library did not support 32 bit filters, which are required for State Archival.

Currently a draft until the (non protocol breaking) associated XDR changes are merged: https://github.com/stellar/stellar-xdr/pull/195

While Binary Fuse Filters are necessary for State Archival, they also significantly improve BucketListDB. Compared to the older bloom filter implementation, Bucket indexing time has decreased by 67% (which is blocking on startup in some cases), decreases index memory size by 20%, and reduces redundant disk reads by ~130x. 

# Checklist
- [x] Compiles
- [x] Ran all tests


src/util/BinaryFuseFilter.cpp
```diff
@@ -0,0 +1,75 @@
...
+{
+
+template <typename T, typename U>
+BinaryFuseFilter<T, U>::BinaryFuseFilter(std::vector<uint64_t>& keyHashes,
+                                         binary_fuse_seed_t const& seed)
+    : mFilter(keyHashes.size()), mInputSeed(seed)
+{
+    // If too many hash collisions occur, population will fail. Retry with
+    // a different seed. This is unlikely to happen once, and is statically
+    // impossible to happen 10 times.
```
graydon, line 20: "statically" impossible or _statistically_ impossible? In the paper it looks to me that a failure is in the 1-in-10 to 1-in-100 range -- so I'll buy it's statistically unlikely! But impossible seems strong.

src/util/BinaryFuseFilter.h
```diff
@@ -0,0 +1,91 @@
...
+{
+
+// This class is a wrapper around the binary_fuse_t library that provides
+// serialization for the XDR BinaryFuseFilter type and provides a deterministic
+// LedgerKey interface.
+
+// Only allow uint8_t, uint16_t, and uint32_t
+template <typename T,
+          typename = std::enable_if_t<std::is_unsigned<T>::value &&
+                                      !std::is_same<T, uint64_t>::value>>
```
graydon, line 25: `std::is_same<T,uint8_t> || std::is_same<T, uint16_t> || std::is_same<T,uint32_t>` and then you can remove the comment!

src/util/BinaryFuseFilter.h
```diff
@@ -0,0 +1,91 @@
...
+{
+  private:
+    binary_fuse_t<T> mFilter;
+
+    // Note: as part of filter construction, the internal filter seed might
+    // rotate and no longer be the same as the input seed. The input seed must
+    // be maintained outside of the filter and used to hash input keys in the
+    // contain function to ensure deterministic hashing of input keys
+    // during both populating and querying the filter.
+    binary_fuse_seed_t const mInputSeed;
```
graydon, line 36: Assuming I understand the difference here, I suggest a change to the comment and variable name.

```suggestion
    // Note: This is the seed used to hash keys going into the filter, and we need to preserve it
    // to hash keys we're looking up in the filter. The filter also has its own seed determining its
    // overall structure, which _starts_ with this seed, but which may be rotated a bit in during
    // repeated attempts to successfully populate the filter. So we have to keep both seeds. 
    binary_fuse_seed_t const mHashSeed;
```

graydon (approved):
As far as I can tell this looks good! I would appreciate if you could back out a bunch of the gratuitous diffs from the original upstream code -- variable renames (Foo -> _foo) and reformatting (moving braces across lines, reflowing comments, etc.) -- just because they made it challenging to review the diff, but I am fairly confident the code is functionally the same.

Oh, also it looks like you didn't actually remove the previous bloom filter code from lib. If it's dead it should go.

(None of these are critical and I'll be out monday so I'm marking this approved as-is, but I'd appreciate a couple of these touchups as followups at some point, I don't think they change anything on-disk or functional)

MonsieurNicolas:
we'll have to put a note in the release notes that this will cause a reindexing of buckets on startup which will cause nodes to be offline for longer than normal

src/util/BinaryFuseFilter.h
```diff
@@ -0,0 +1,91 @@
...
+                              binary_fuse_seed_t const& seed);
+    explicit BinaryFuseFilter(SerializedBinaryFuseFilter const& xdrFilter);
+
+    bool contains(LedgerKey const& key) const;
+
+    bool operator==(BinaryFuseFilter<T> const& other) const;
+
+    template <class Archive>
+    void
+    save(Archive& archive) const
```
dmkozh, line 51: Why do we need to use custom encoding here instead of copying the fields into the XDR struct?

src/util/BinaryFuseFilter.cpp
```diff
@@ -0,0 +1,75 @@
...
+}
+
+template <typename T, typename U>
+bool
+BinaryFuseFilter<T, U>::operator==(BinaryFuseFilter<T> const& other) const
+{
+    return mFilter == other.mFilter && mInputSeed == other.mInputSeed;
+}
+
+template class BinaryFuseFilter<uint8_t, void>;
```
dmkozh, line 72: I can't immediately grasp what's the purpose of `U` type and why is it always `void`, would you mind to elaborate?

src/util/BinaryFuseFilter.h
SirTyson: Hmm I'm not sure what you mean by custom encoding? We are just copying the fields into the XDR struct. The `mInputSeed` parameter is used to provide input into the binary fuse filter library, so it is stored in `mInputSeed` instead of the library class `mFilter`. `mFilter.copyTo` does just copy all the fields into the XDR struct.

If by "custom encoding" you're refering to byte packing the `fingerprints` XDR field, that is necessary because XDR does not support vector with 16 bit elements, only 8 bit and 32 bit. For consistency I opted to always use an 8 bit XDR vector and bit pack all elem types instead of just special casing the 16 bit case and requiring a switch at the XDR level.

src/util/BinaryFuseFilter.cpp
SirTyson: It's just the syntactically ugly way to have template constraints pre C++20. When I declare the constrain in the header file via

```
template <typename T,
          typename = std::enable_if_t<std::is_unsigned<T>::value && !std::is_same<T, uint64_t>::value>>
```

this either results in a compilation error or the deduced type template <valid_type, void>. While I agree it's ugly and not-intuitive, I think it's important to enforce that this is a small collection of unsigned bytes, just because the signedness and bit width is integral to the filter. Hopefully we'll upgrade to C++20 at some point and I can just use concepts.

Here's where I found this btw: https://stackoverflow.com/questions/66649545/method-outside-template-class-with-enable-if. To make it a little more robust and following Graydon's comment, I'll change the H type definition to 
```
template <typename T,
          typename = std::is_same<T,uint8_t> || std::is_same<T, uint16_t> || std::is_same<T,uint32_t> = 0>
```

src/util/BinaryFuseFilter.h
dmkozh: Yeah, by 'custom encoding' I meant definitions of save/load functions instead of only using the autogenerated ones for the XDR struct.  I'm still not sure I understand the motivation for implementing the archive functions vs explicit construction from/conversion to `SerializedBinaryFuseFilter`; typically we use that to customize the way the data is being serialized, but IIUC we actually don't customize anything. But I'm not feeling too strongly about this.

src/util/BinaryFuseFilter.cpp
dmkozh: That would be great, but if that doesn't work just leave a small comment about this.

src/util/BinaryFuseFilter.cpp
SirTyson: Yeah, statistically impossible. The library implementation itself will rotate the key 100 times then give up. Our wrapper around the library will rotate a different byte of the key then the library and try again 10 times, for a total seed rotation of 10 * 100. According to the original implementation, 100 seed rotations was already lower than the [chance of being hit with a cosmic ray](https://github.com/FastFilter/xor_singleheader/blob/3c0fd15b1e04281b2ada00cf82ddffc4b3292dee/include/binaryfusefilter.h#L308), so I think 10 * 100 can be considered statistically impossible.

src/util/BinaryFuseFilter.h
SirTyson: This is just for convenience purposes. After start and when creating a new filter, we only want to keep the non-XDR version in memory to actually do lookups/construction. When we go to serialize, since we only have the in-memory version, it's significantly simpler to just define a skeleton save/load pair here than manually convert the saved version to the XDR version then serialize that.

Specifically, in `SerializableBucketIndex`, we want to serialize `std::unique_ptr<BinaryFuseFilter16> filter`. Cereal has support for smart pointers, where it essentially automatically does
```
ar(pointerOverhead)
if (pointer)
    ar(*pointer)
```

If we were to not specify thesave/load pairs for the non-xdr filter, we'd have to manually implement this pointer behavior instead of just using the built in cereal functionality, since Cereal would not be able to deduce how to archive the value of the dereferenced pointer.

SirTyson:
> I also was wondering if you considered a version that doesn't have the "populated" flag, but that returns an optional<> from the populate function, i.e. treat it as a sort of fallible constructor. I don't have a super strong preference for this -- it's not really idiomatic C++ the way it would be idiomatic Rust -- but I'm not thrilled with the 2-phase initialization and need to check-and-fail on the non-populated state. As far as I can tell, no client should ever have a reason to hold a non-populated filter, right? I wonder if there's a way to bake that further into the API.

I tried to enforce this with the `[[nodiscard]] bool` on the populate function. I agree the interface is not ideal, but because the caller constructing the filter may need to rotate the input seed, there is a "valid" reason that a caller may need to hold a non populated filter during construction while finding the correct seed. I'm not sure if an optional is right here. I think I'll just increase the max attempt count to a highly ridiculous number (as this is filter is part of consensus and construction should never fail) and throw in the constructor if it fails.

dmkozh (approved):

src/util/BinaryFuseFilter.cpp
graydon: ok

PR #4413, opened 2024-08-02
Soroban upgrades for simulation tests

ThomasBrady:


src/test/TestUtils.cpp
```diff
@@ -194,6 +197,78 @@ genesis(int minute, int second)
...
+{
+    auto nodes = simulation->getNodes();
+    auto& lg = nodes[0]->getLoadGenerator();
+    auto& app = *nodes[0];
+
+    auto& complete =
+        app.getMetrics().NewMeter({"loadgen", "run", "complete"}, "run");
+    auto completeCount = complete.count();
+    // Only create an account if there are none aleady created.
+    if (app.getMetrics()
```
marta-lokhova, line 212: Why do we need to create accounts here? We could just use the root account right?

src/test/TestUtils.cpp
```diff
@@ -194,6 +197,78 @@ genesis(int minute, int second)
...
+            .NewMeter({"loadgen", "account", "created"}, "account")
+            .count() == 0)
+    {
+        auto createAccountsLoadConfig =
+            GeneratedLoadConfig::createAccountsLoad(1, 1);
+        createAccountsLoadConfig.offset =
```
marta-lokhova, line 218: this doesn't look right: offsets are used if multiple nodes are creating accounts at the same time, so we use offset as to not overlap created accounts. Here we're only using one node, so we shouldn't need to configure offsets (same applies to other loadgen modes below)

src/test/TestUtils.cpp
```diff
@@ -194,6 +197,78 @@ genesis(int minute, int second)
...
+            app->getLedgerManager().getLastClosedLedgerHeader().header;
+        scheduledUpgrades.mUpgradeTime =
+            VirtualClock::from_time_t(lclHeader.scpValue.closeTime);
+        scheduledUpgrades.mConfigUpgradeSetKey = upgradeSetKey;
+        app->getHerder().setUpgrades(scheduledUpgrades);
+    }
+    // Wait for upgrade to be applied
+    simulation->crankUntil(
+        [&]() {
+            auto status = app.getStatusManager().getStatusMessage(
```
marta-lokhova, line 264: This doesn't properly wait for the upgrade to go through: setUpgrades sets status message REQUIRES_UPGRADES immediately, so this call to crankUntil is essentially a no-op. This is probably why some of the tests are failing. 

I'd suggest cranking until `cfg` equals config you get from `LedgerManager.getSorobanNetworkConfig`. Basically, you need to ensure that upgrades were applied to the ledger.

marta-lokhova (changes requested):

src/test/TestUtils.cpp
ThomasBrady: The issue is that this function is used by tests that create accounts after calling `upgradeSorobanNetworkConfig`. If an offset is not used, the test case's subsequent account creation load generation will fail as the first `CREATE_ACCOUNT` operation will return `CREATE_ACCOUNT_ALREADY_EXIST`. The the load generation will not complete and the simulation will timeout.

src/test/TestUtils.cpp
ThomasBrady: How can I use the root account? If I remove the account creation step I get: ` Account 0 must exist in the DB.`

marta-lokhova: ah I see, loadgen doesn't mark root account as "available to use" for some reason. if it did, it would simplify things since you wouldn't need account creation + offset. You could modify this path https://github.com/stellar/stellar-core/blob/ad7b1646209fcd069c067a9a7730c9b6305be667/src/simulation/LoadGenerator.cpp#L646 

Instead, it would also return root account (mRoot) if it's available. That's lesser priority though, and can be done as a follow-up.

src/test/TestUtils.cpp
ThomasBrady: As I understand it, `REQUIRES_UPGRADES` is a category of status message, `setUpgrades` [ sets the contents ](https://github.com/stellar/stellar-core/blob/master/src/herder/HerderImpl.cpp#L1513) of the `REQUIRES_UPGRADES` status message to "Armed with network upgrades...". When the upgrades are applied, the `REQUIRES_UPGRADES` status message is removed. This cranks until that status message is empty. I'm fairly certain this works as I have [assertions on the upgraded values after this completes](https://github.com/stellar/stellar-core/pull/4413/files/7275997f7876adcbc806bc99772347f0f30f8b71#diff-519110afbb88598be9bd6ca994ca41c4a80840b66febd9d58765ca96955c3389R122), but I can change it to compare with the config from `LedgerManager.getSorobanNetworkConfig`.

marta-lokhova: I see. I believe relying on status message still isn't enough to tell that upgrades went through, for a few reasons:
* Core clears upgrades when it reaches consensus on a block containing upgrades, but _before_ the block is applied. This means that your crank function is potentially racing with the actual upgrade application, so we don't know if upgrades were applied when this function exits. 
* It seems a bit error-prone to rely on an empty status message: core is a state machine so it might enter a different state, and set status to something else, which would cause the simulation to time out.

I think the most reliable approach would be to rely directly on what's in the database, e.g. use `getSorobanNetworkConfig`

src/herder/test/HerderTests.cpp
```diff
@@ -4654,31 +4656,29 @@ TEST_CASE("do not flood too many soroban transactions",
...
+
+    auto cfgUpgradeSetKey = upgradeSorobanNetworkConfig(
+        [&](SorobanNetworkConfig& cfg) {
+            setSorobanNetworkConfigForTest(cfg);
+            // Update read entries to allow flooding at most 1 tx per broadcast
+            // interval.
+            cfg.mLedgerMaxReadLedgerEntries = 40;
+            cfg.mLedgerMaxReadBytes = cfg.mTxMaxReadBytes;
+        },
+        simulation);
```
marta-lokhova, line 4681: I think this test is now failing because we start flooding a little bit later (because of the upgrade), so transaction queue counts that time towards total allowed quota it can flood. You can fix this by clearing the carry-over in the transaction queue, here's the diff:
```
diff --git a/src/herder/TransactionQueue.h b/src/herder/TransactionQueue.h
index a58573a1a..4d531a68f 100644
--- a/src/herder/TransactionQueue.h
+++ b/src/herder/TransactionQueue.h
@@ -259,6 +259,14 @@ class SorobanTransactionQueue : public TransactionQueue
     }

     size_t getMaxQueueSizeOps() const override;
+#ifdef BUILD_TESTS
+    void
+    clearBroadcastCarryover()
+    {
+        mBroadcastOpCarryover.clear();
+        mBroadcastOpCarryover.resize(1, Resource::makeEmptySoroban());
+    }
+#endif

   private:
     virtual std::pair<Resource, std::optional<Resource>>
diff --git a/src/herder/test/HerderTests.cpp b/src/herder/test/HerderTests.cpp
index 03eee093e..603e57b22 100644
--- a/src/herder/test/HerderTests.cpp
+++ b/src/herder/test/HerderTests.cpp
@@ -4769,6 +4769,8 @@ TEST_CASE("do not flood too many soroban transactions",
     {
         // no broadcast right away
         REQUIRE(numBroadcast == 0);
+        tq.clearBroadcastCarryover();
+
         // wait for a bit more than a broadcast period
         // rate per period is 100 ms
         auto broadcastPeriod =
```

marta-lokhova:
@ThomasBrady what other tests are failing?

ThomasBrady:
> @ThomasBrady what other tests are failing?
Thanks for the fix! I saw the residual carryOver but I didn't think of just clearing it 😅 .
I think they are all passing now, I fixed two other test cases today (`generate soroban load`, `overlay flow control`) as well as reading the config from the db. Running a full make check to confirm, then I'll push the update.

marta-lokhova:
@ThomasBrady looks like CI is failing?

ThomasBrady:
> @ThomasBrady looks like CI is failing?

Formatting issues 🙃 re-running now 🙏

marta-lokhova:
@ThomasBrady thanks for the updates! Could we also confirm that acceptance tests pass? You can run `[acceptance]`  tag to verify that.

src/overlay/test/OverlayTests.cpp
```diff
@@ -2141,33 +2141,32 @@ TEST_CASE("overlay flow control", "[overlay][flowcontrol]")
...
-            [&] { return simulation->haveAllExternalized(2, 1); },
-            3 * Herder::EXP_LEDGER_TIMESPAN_SECONDS, false);
         // Generate a bit of load to flood transactions, make sure nodes can
         // close ledgers properly
         auto& loadGen = node->getLoadGenerator();
-        loadGen.generateLoad(
-            GeneratedLoadConfig::createAccountsLoad(/* nAccounts */ 150,
-                                                    /* txRate */ 1));
+        loadGen.generateLoad(GeneratedLoadConfig::createAccountsLoad(
+            /* nAccount */ 150,
```
marta-lokhova, line 2168: please revert this

src/simulation/LoadGenerator.cpp
```diff
@@ -2183,9 +2220,7 @@ LoadGenerator::execute(TransactionTestFramePtr& txf, LoadGenMode mode,
                 : "";
         CLOG_INFO(LoadGen, "tx rejected '{}': ===> {}, {}",
                   TX_STATUS_STRING[static_cast<int>(addResult.code)],
-                  txf->isSoroban() ? "soroban"
-                                   : xdrToCerealString(txf->getEnvelope(),
-                                                       "TransactionEnvelope"),
+                  xdrToCerealString(txf->getEnvelope(), "TransactionEnvelope"),
```
marta-lokhova, line 2223: I think the problem is with transactions like UPLOAD_WASM, which print the whole multi-KB WASM blob. Ideally, we print the envelope contents without WASM, but that seems outside of the scope of this PR, so I'd just revert this.

src/simulation/test/LoadGeneratorTests.cpp
```diff
@@ -339,13 +376,13 @@ TEST_CASE("generate soroban load", "[loadgen][soroban]")
         }
     }
 
-    // Manually override network settings for invoke load gen
-    for (auto node : nodes)
-    {
-        overrideSorobanNetworkConfigForTest(*node);
-        modifySorobanNetworkConfig(*node, [&](SorobanNetworkConfig& cfg) {
+    // Override network settings for invoke load gen
```
marta-lokhova, line 379: Should we update this comment? We're not "overriding" configs anymore, as the network just goes through a normal upgrade flow.

src/simulation/test/LoadGeneratorTests.cpp
```diff
@@ -91,6 +91,49 @@ TEST_CASE("generate load with unique accounts", "[loadgen]")
...
+        });
+
+    simulation->startAllNodes();
+    simulation->crankUntil(
+        [&]() { return simulation->haveAllExternalized(3, 1); },
+        2 * Herder::EXP_LEDGER_TIMESPAN_SECONDS, false);
+    auto nodes = simulation->getNodes();
+    auto& app = *nodes[0]; // pick a node to generate load
+
+    // Manually override network settings for invoke load gen
```
marta-lokhova, line 110: Let's update this comment, since we're not manually overriding configs anymore

src/simulation/test/LoadGeneratorTests.cpp
```diff
@@ -339,13 +376,13 @@ TEST_CASE("generate soroban load", "[loadgen][soroban]")
...
-        overrideSorobanNetworkConfigForTest(*node);
-        modifySorobanNetworkConfig(*node, [&](SorobanNetworkConfig& cfg) {
+    // Override network settings for invoke load gen
+    upgradeSorobanNetworkConfig(
+        [&](SorobanNetworkConfig& cfg) {
+            setSorobanNetworkConfigForTest(cfg);
+
             // Entries should never expire
-            cfg.mStateArchivalSettings.maxEntryTTL = 1'000'000;
+            cfg.mStateArchivalSettings.maxEntryTTL = 2'000'000;
```
marta-lokhova, line 385: Why this change?

src/test/TestUtils.cpp
marta-lokhova: can we track this issue somewhere though? It'd be nice to get rid of account creation, and manipulations with the offset, which are quite error-prone.

src/test/TestUtils.cpp
```diff
@@ -194,6 +197,78 @@ genesis(int minute, int second)
...
+        scheduledUpgrades.mConfigUpgradeSetKey = upgradeSetKey;
+        app->getHerder().setUpgrades(scheduledUpgrades);
+    }
+    // Wait for upgrade to be applied
+    simulation->crankUntil(
+        [&]() {
+            auto netCfg = app.getLedgerManager().getSorobanNetworkConfig();
+            return netCfg == cfg;
+        },
+        300 * Herder::EXP_LEDGER_TIMESPAN_SECONDS, false);
```
marta-lokhova, line 269: 300 ledgers seems like a really long timeout, in reality you should see the upgrade going through on the next ledgers, so you can set it to 2 to be safe.

src/test/TestUtils.h
```diff
@@ -16,6 +16,8 @@ namespace stellar
 {
 
 class LoopbackPeerConnection;
+class Simulation;
+using SimulationPtr = std::shared_ptr<Simulation>;
```
marta-lokhova, line 20: Do we need this? SimulationPtr isn't used anywhere outside of this util file

src/test/TestUtils.cpp
```diff
@@ -3,11 +3,14 @@
 // of this distribution or at http://www.apache.org/licenses/LICENSE-2.0
 
 #include "TestUtils.h"
+#include "ledger/LedgerTxn.h"
 #include "overlay/test/LoopbackPeer.h"
+#include "simulation/LoadGenerator.h"
+#include "simulation/Simulation.h"
 #include "test/TxTests.h"
 #include "test/test.h"
+#include "util/StatusManager.h"
```
marta-lokhova, line 12: unused include (same with  LedgerTxn)

src/herder/test/HerderTests.cpp
```diff
@@ -50,6 +50,8 @@
 #include <numeric>
 #include <optional>
 
+#include "util/XDRCereal.h"
```
marta-lokhova, line 53: unused?

marta-lokhova (commented):
thanks, looks good overall! I left a few comments regarding code cleanup.

src/simulation/test/LoadGeneratorTests.cpp
ThomasBrady: `SorobanNetworkConfig::isValidConfigSettingEntry` [checks that ](https://github.com/stellar/stellar-core/blob/master/src/ledger/NetworkConfig.cpp#L799) `cfg.stateArchivalSettings().maxEntryTTL >= MinimumSorobanNetworkConfig::MAXIMUM_ENTRY_LIFETIME`, and `MAXIMUM_ENTRY_LIFETIME = 1'054'080;`, so 1 million is an invalid setting. The previously used ltx-based `modifySorobanNetworkConfig` function bypasses the validity check.

src/test/TestUtils.cpp
ThomasBrady: https://github.com/stellar/stellar-core/issues/4445

src/overlay/test/OverlayTests.cpp
```diff
@@ -2141,33 +2141,32 @@ TEST_CASE("overlay flow control", "[overlay][flowcontrol]")
         simulation->addPendingConnection(vNode2NodeID, vNode3NodeID);
         simulation->addPendingConnection(vNode3NodeID, vNode1NodeID);
 
-        for (auto& node2 : simulation->getNodes())
+        simulation->startAllNodes();
+        simulation->crankUntil(
```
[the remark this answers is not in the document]
ThomasBrady: I missed this one, checking now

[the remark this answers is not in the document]
ThomasBrady: deleted

src/overlay/test/OverlayTests.cpp
marta-lokhova: the typo is still there, should be nAccounts

ThomasBrady: oops I thought you meant the formatting 😅 done.

marta-lokhova (approved):

PR #4416, opened 2024-08-05
Enable background eviction by default & drop support for EXPERIMENTAL_BACKGROUND_EVICTION_SCAN

### LINKED ISSUE #4415: Enable Background Eviction by Default
Background eviction has been running successfully for some time now, so we should remove the "experimental" from the feature flag naming and default enable background eviction.

SirTyson:
# Description

Resolves #4415

Drops "experimental" from background eviction flag and enables it by default.

# Checklist
- [x] Compiles
- [x] Ran all tests

### DIFF
--- a/docs/stellar-core_example.cfg
+++ b/docs/stellar-core_example.cfg
@@ -258,10 +258,10 @@ BUCKETLIST_DB_INDEX_CUTOFF = 20
 # this value is ingnored and indexes are never persisted.
 BUCKETLIST_DB_PERSIST_INDEX = true
 
-# EXPERIMENTAL_BACKGROUND_EVICTION_SCAN (bool) default false
+# BACKGROUND_EVICTION_SCAN (bool) default true
 # Determines whether eviction scans occur in the background thread. Requires
-# that EXPERIMENTAL_BACKGROUND_EVICTION_SCAN is set to true.
-EXPERIMENTAL_BACKGROUND_EVICTION_SCAN = false
+# that DEPRECATED_SQL_LEDGER_STATE is set to false.
+BACKGROUND_EVICTION_SCAN = true
 
 # EXPERIMENTAL_BACKGROUND_OVERLAY_PROCESSING (bool) default false
 # Determines whether some of overlay processing occurs in the background
--- a/src/main/Config.cpp
+++ b/src/main/Config.cpp
@@ -167,7 +167,7 @@ Config::Config() : NODE_SEED(SecretKey::random())
     BUCKETLIST_DB_INDEX_PAGE_SIZE_EXPONENT = 14; // 2^14 == 16 kb
     BUCKETLIST_DB_INDEX_CUTOFF = 20;             // 20 mb
     BUCKETLIST_DB_PERSIST_INDEX = true;
-    EXPERIMENTAL_BACKGROUND_EVICTION_SCAN = false;
+    BACKGROUND_EVICTION_SCAN = true;
     PUBLISH_TO_ARCHIVE_DELAY = std::chrono::seconds{0};
     // automatic maintenance settings:
     // short and prime with 1 hour which will cause automatic maintenance to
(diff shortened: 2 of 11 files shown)


docs/stellar-core_example.cfg
```diff
@@ -236,10 +236,10 @@ BUCKETLIST_DB_INDEX_CUTOFF = 20
 # this value is ingnored and indexes are never persisted.
 BUCKETLIST_DB_PERSIST_INDEX = true
 
-# EXPERIMENTAL_BACKGROUND_EVICTION_SCAN (bool) default false
+# BACKGROUND_EVICTION_SCAN (bool) default true
 # Determines whether eviction scans occur in the background thread. Requires
-# that EXPERIMENTAL_BACKGROUND_EVICTION_SCAN is set to true.
-EXPERIMENTAL_BACKGROUND_EVICTION_SCAN = false
+# that USE_DEPRECATED_SQL is set to false.
```
marta-lokhova, line 241: `DEPRECATED_SQL_LEDGER_STATE`

marta-lokhova, line 241: note that this will break nodes that use EXPERIMENTAL_BACKGROUND_EVICTION_SCAN right now (so at least our infra), we should remember to remove that flag before releasing these changes.

src/main/Config.cpp
```diff
@@ -1074,9 +1074,9 @@ Config::processConfig(std::shared_ptr<cpptoml::table> t)
             {
                 EXPERIMENTAL_BACKGROUND_OVERLAY_PROCESSING = readBool(item);
             }
-            else if (item.first == "EXPERIMENTAL_BACKGROUND_EVICTION_SCAN")
+            else if (item.first == "BACKGROUND_EVICTION_SCAN")
             {
-                EXPERIMENTAL_BACKGROUND_EVICTION_SCAN = readBool(item);
+                BACKGROUND_EVICTION_SCAN = readBool(item);
```
marta-lokhova, line 1079: should we add a warning here if BucketListDB isn't enabled?

marta-lokhova (commented):
LGTM! Just a few minor comments

src/main/Config.cpp
SirTyson: We do, warning is on `src/main/ApplicationImpl.cpp:836`

docs/stellar-core_example.cfg
SirTyson: Done

SirTyson: Yeah, since it's experimental we are probably the only org using these flags, but we'll need to make sure to update the toml files on release accordingly.

src/main/Config.cpp
marta-lokhova: Oh i see, should we throw though? It's likely operators won't notice the misconfiguration otherwise

docs/stellar-core_example.cfg
marta-lokhova: sg, I updated the name of your PR to remind us about this when putting together release notes for 22.0.0

src/main/Config.cpp
SirTyson: I'd rather not. Bot this flag and the SQL flag are going away next release anyway, and I doubt anyone would have a reason to set this flag manually. Even if they did, currently there's essentially no difference between foreground and background eviction (because the scan size is still small), so I'd rather not crash over something so minor.

marta-lokhova: well it's a crash on startup right after an operator manually makes a change to the config, so crashing seems more reasonable than silently proceeding, as it indicates that there's a misconfiguration (that's typically what we do for other configs that can't co-exist together). otherwise, operators have no idea that their nodes are misconfigured and get bitten by it later.

SirTyson: Done

marta-lokhova: thanks!

marta-lokhova (dismissed):

marta-lokhova (approved):

PR #4417, opened 2024-08-05
Remove protocol-specific assert from TxSetUtils::getInvalidTxList

marta-lokhova:
This assert isn't quite right, as nodes may still call this functionality when out-of-sync/catching up. I think there are two scenarios where this can be triggered:
* Node externalized latest (p21) ledger, but it's at genesis (p0), so tx queue cleanup crashes at the assert. 
* Validator starts at genesis with force_scp=true, which causes it to crash while trying to nominate a block for ledger 2 (which gets ignored by the network). It's not crashed right now given how the code is written (the function is not called if the transaction set is empty), but it's still a footgun.

I removed the assert - for out-of-sync nodes, this path _should_ be harmless, as we don't accept any invalid/unsupported transactions in the transaction queue to begin with. I believe this is how it was handled in previous protocols as well.

### DIFF
--- a/src/herder/TxSetUtils.cpp
+++ b/src/herder/TxSetUtils.cpp
@@ -144,15 +144,6 @@ TxSetUtils::getInvalidTxList(TxSetTransactions const& txs, Application& app,
     ZoneScoped;
     LedgerTxn ltx(app.getLedgerTxnRoot(), /* shouldUpdateLastModified */ true,
                   TransactionMode::READ_ONLY_WITHOUT_SQL_TXN);
-#ifdef BUILD_TESTS
-    if (protocolVersionIsBefore(ltx.loadHeader().current().ledgerVersion,
-                                ProtocolVersion::V_20))
-    {
-        return {};
-    }
-#endif
-    releaseAssert(protocolVersionStartsFrom(
-        ltx.loadHeader().current().ledgerVersion, ProtocolVersion::V_20));
     // This is done so minSeqLedgerGap is validated against the next
     // ledgerSeq, which is what will be used at apply time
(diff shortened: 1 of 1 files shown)


src/herder/TxSetUtils.cpp
```diff
@@ -144,15 +144,6 @@ TxSetUtils::getInvalidTxList(TxSetTransactions const& txs, Application& app,
...
     LedgerTxn ltx(app.getLedgerTxnRoot(), /* shouldUpdateLastModified */ true,
                   TransactionMode::READ_ONLY_WITHOUT_SQL_TXN);
-#ifdef BUILD_TESTS
-    if (protocolVersionIsBefore(ltx.loadHeader().current().ledgerVersion,
-                                ProtocolVersion::V_20))
-    {
-        return {};
-    }
-#endif
-    releaseAssert(protocolVersionStartsFrom(
```
dmkozh, line 154: I'm ok with removing this, but can we make sure this can actually happen? Specifically, it is a bit surprising to me that a node tries to nominate a tx set that is targeting the old protocol. Do out-of-sync nodes think that they're in-sync and thus should try to nominate tx sets?

marta-lokhova: Yea, it depends on how a node starts. If a validator is started without `--wait-for-conseus`, it'll consider itself "synced", and attempt nominating a value for LCL+1 to potentially unstick the network. For ledger 2, it doesn't matter of course, so we just timeout and start buffering ledgers from the network.

I should add that right now the code is written such that we won't trigger getInvalidTxs crash (as it's not called on empty tx sets), but I think this is still a valid scenario to consider.

dmkozh (approved):

PR #4423, opened 2024-08-07
harden simulation tests

marta-lokhova:
* Minor follow up to #4353 to harden tests
* Bump medida


bboston7 (approved):
Looks good! Thanks for fixing these tests up and documenting how the `Simulation` class is meant to be used!

sisuresh (approved):

PR #4428, opened 2024-08-12
non-existent entries should not be metered

ThomasBrady:
# Description

Resolves https://github.com/stellar/stellar-core-internal/issues/282

Remove optimization that checks whether there is quota remaining for a ledger key before attempting to load the entry from the bucketlist. This ensures that non-existent entries are not metered and result in a `nullptr` entry in the cache. This will prevent point loads when writing newly created entries post-contract invocation.


SirTyson (approved):
LGTM, thanks for the fix and the test! I'll add this to the merge queue once we finish the release.

Issue #4426, opened 2024-08-08
Add per-bucket Entry size Metrics


SirTyson:
Currently, we only track the size of the whole BucketList. This doesn't give us much insight into the makeup of the BucketList. We should add a size and count metric grouped by `LedgerEntry` type (and persistent vs. temporary for soroban types) to each Bucket. This could be efficiently calculated during Bucket merges. We should then aggregate BucketList level metrics and log these to grafana, as well as output them in meta.

ThomasBrady:
- [ ] Code changes
- [ ] Grafana dashboard change
- [ ] Hubble ingestion / coordinate with data team

ThomasBrady:
https://github.com/stellar/puppet-v4/pull/4081 grafana changes

ThomasBrady:
XDR changes https://github.com/stellar/stellar-xdr/pull/219

PR #4436, opened 2024-08-23
Bucket entry counters and metrics

ThomasBrady:
# Description

Resolves https://github.com/stellar/stellar-core/issues/4426

* Adds `BucketEntryCounters` to `BucketIndexImpl` to count entry types and duration. 
* Adds functionality to `BucketList` to accumulate the counters for all buckets
* Adds functionality to LedgerManager to emit these metrics to medida 

Still TODO:
* Local testing of serialization of indexes with counters 

Out of scope:
* XDR Changes for including counters in ledger close meta
* grafana updates

# Checklist
- [ ] Compiles
- [ ] Ran all tests

### DIFF
--- a/src/bucket/Bucket.cpp
+++ b/src/bucket/Bucket.cpp
@@ -863,4 +863,38 @@ Bucket::getBucketVersion(std::shared_ptr<Bucket const> const& bucket)
     BucketInputIterator it(bucket);
     return it.getMetadata().ledgerVersion;
 }
+
+BucketEntryCounters const&
+Bucket::getBucketEntryCounters() const
+{
+    releaseAssert(mIndex);
+    return mIndex->getBucketEntryCounters();
+}
+
+BucketEntryCounters&
+BucketEntryCounters::operator+=(BucketEntryCounters const& other)
+{
+    for (auto [type, count] : other.entryTypeCounts)
+    {
+        this->entryTypeCounts[type] += count;
+    }
+    for (auto [type, size] : other.entryTypeSizes)
+    {
+        this->entryTypeSizes[type] += size;
+    }
+    return *this;
+}
+
+bool
+BucketEntryCounters::operator==(BucketEntryCounters const& other) const
+{
+    return this->entryTypeCounts == other.entryTypeCounts &&
+           this->entryTypeSizes == other.entryTypeSizes;
+}
+
+bool
+BucketEntryCounters::operator!=(BucketEntryCounters const& other) const
+{
+    return !(*this == other);
+}
 }
--- a/src/bucket/Bucket.h
+++ b/src/bucket/Bucket.h
@@ -9,6 +9,7 @@
 #include "util/ProtocolVersion.h"
 #include "xdr/Stellar-ledger.h"
 #include <list>
+#include <map>
 #include <optional>
 #include <string>
 
@@ -43,6 +44,7 @@ class BucketManager;
 class SearchableBucketListSnapshot;
 struct EvictionResultEntry;
 class EvictionStatistics;
+struct BucketEntryCounters;
 
 class Bucket : public std::enable_shared_from_this<Bucket>,
                public NonMovableOrCopyable
@@ -179,7 +181,25 @@ class Bucket : public std::enable_shared_from_this<Bucket>,
     static uint32_t getBucketVersion(std::shared_ptr<Bucket> const& bucket);
     static uint32_t
     getBucketVersion(std::shared_ptr<Bucket const> const& bucket);
-
+    BucketEntryCounters const& getBucketEntryCounters() const;
     friend class BucketSnapshot;
+};
+
+enum class LedgerEntryTypeAndDurability : uint32_t;
+struct BucketEntryCounters
+{
+    std::map<LedgerEntryTypeAndDurability, size_t> entryTypeCounts;
+    std::map<LedgerEntryTypeAndDurability, size_t> entryTypeSizes;
+
+    BucketEntryCounters& operator+=(BucketEntryCounters const& other);
+    bool operator==(BucketEntryCounters const& other) const;
+    bool operator!=(BucketEntryCounters const& other) const;
+
+    template <class Archive>
+    void
+    serialize(Archive& ar)
+    {
+        ar(entryTypeCounts, entryTypeSizes);
+    }
 };
 }
--- a/src/bucket/BucketIndexImpl.cpp
+++ b/src/bucket/BucketIndexImpl.cpp
@@ -7,6 +7,7 @@
 #include "bucket/BucketManager.h"
 #include "crypto/Hex.h"
 #include "crypto/ShortHash.h"
+#include "ledger/LedgerTypeUtils.h"
 #include "main/Config.h"
 #include "util/BinaryFuseFilter.h"
 #include "util/Fs.h"
@@ -100,6 +101,17 @@ BucketIndexImpl<IndexT>::BucketIndexImpl(BucketManager& bm,
         std::vector<uint64_t> keyHashes;
         auto seed = shortHash::getShortHashInitKey();
 
+        auto countEntry = [&](BucketEntry const& be) {
+            if (be.type() == METAENTRY)
+            {
+                // Do not count meta entries.
+                return;
+            }
+            auto ledt = bucketEntryToLedgerEntryAndDurabilityType(be);
+            mData.counters.entryTypeCounts[ledt]++;
+            mData.counters.entryTypeSizes[ledt] += xdr::xdr_size(be);
+        };
+
         while (in && in.readOne(be))
         {
             // peridocially check if bucket manager is exiting to stop indexing
@@ -170,6 +182,7 @@ BucketIndexImpl<IndexT>::BucketIndexImpl(BucketManager& bm,
                 {
                     mData.keysToOffset.emplace_back(key, pos);
                 }
+                countEntry(be);
             }
 
             pos = in.pos();
@@ -551,6 +564,11 @@ BucketIndexImpl<IndexT>::operator==(BucketIndex const& inRaw) const
         }
     }
 
+    if (mData.counters != in.mData.counters)
+    {
+        return false;
+    }
+
     return true;
 }
 #endif
@@ -579,5 +597,12 @@ void
 BucketIndexImpl<BucketIndex::RangeIndex>::markBloomLookup() const
 {
     mBloomLookupMeter.Mark();
+}
+
+template <class IndexT>
+BucketEntryCounters const&
+BucketIndexImpl<IndexT>::getBucketEntryCounters() const
+{
+    return mData.counters;
 }
 }
--- a/src/bucket/BucketIndexImpl.h
+++ b/src/bucket/BucketIndexImpl.h
@@ -4,6 +4,7 @@
 // under the Apache License, Version 2.0. See the COPYING file at the root
 // of this distribution or at http://www.apache.org/licenses/LICENSE-2.0
 
+#include "bucket/Bucket.h"
 #include "bucket/BucketIndex.h"
 #include "medida/meter.h"
 #include "util/BinaryFuseFilter.h"
@@ -15,7 +16,6 @@
 
 namespace stellar
 {
-
 // Index maps either individual keys or a key range of BucketEntry's to the
 // associated offset within the bucket file. Index stored as vector of pairs:
 // First: LedgerKey/Key ranges sorted in the same scheme as LedgerEntryCmp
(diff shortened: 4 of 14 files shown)


SirTyson (dismissed):
Looks great! A lot of super small nit picky things. I'll go ahead and approve now so you can merge after a few fixups next week when I'm out.

src/bucket/Bucket.cpp
```diff
@@ -854,4 +855,14 @@ Bucket::getBucketVersion(std::shared_ptr<Bucket const> const& bucket)
     BucketInputIterator it(bucket);
     return it.getMetadata().ledgerVersion;
 }
+
+BucketEntryCounters
```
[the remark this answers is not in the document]
SirTyson: This still should be `BucketEntryCounters const&` in `Bucket.h`.

src/bucket/Bucket.cpp
```diff
@@ -854,4 +854,28 @@ Bucket::getBucketVersion(std::shared_ptr<Bucket const> const& bucket)
...
+Bucket::getBucketEntryCounters() const
+{
+    if (mIndex)
```
SirTyson, line 861: I think this should probably just `releaseAssert(mIndex)`. The reason being if we return empty here, downstream might run into confusing bugs where metrics are returning as all zeros because of a misconfiguration, and it's not clear to downstream that these metrics would be tied to SQL database config flags. I would instead assert and protect calls to the metrics endpoint with `isUsingBucketListDB` checks.

src/bucket/BucketIndexImpl.cpp
```diff
@@ -100,6 +101,79 @@ BucketIndexImpl<IndexT>::BucketIndexImpl(BucketManager& bm,
         std::vector<uint64_t> keyHashes;
         auto seed = shortHash::getShortHashInitKey();
 
+        auto countEntry = [&](BucketEntry const& be) {
+            auto ledgerEntryToLedgerEntryAndDurabilityType =
```
SirTyson, line 105: Can we move this inner lambda to be a static function instead? Generally speaking I'd like to avoid nested lambdas for clarity. It might make the most sense to move this static function and `LedgerEntryTypeAndDurability` definition to `LedgerTypeUtils`. Also, I don't think we need to pass in `isTemporaryEntry` at all here. The simplest interface would be something like `static LedgerEntryTypeAndDurability getLedgerEntryAndDurabilityType(BucketEntry const& be)`.

src/bucket/BucketIndexImpl.h
```diff
@@ -33,13 +33,15 @@ template <class IndexT> class BucketIndexImpl : public BucketIndex
...
+        BucketEntryCounters counters{};
 
         template <class Archive>
         void
         save(Archive& ar) const
         {
             auto version = BUCKET_INDEX_VERSION;
-            ar(version, pageSize, assetToPoolID, keysToOffset, filter);
+            ar(version, pageSize, assetToPoolID, keysToOffset, filter,
+               counters);
```
SirTyson, line 44: Counters must also be added to the `load` functions as well. It might be helpful to add metric tests to the "serialize bucket indexes" test to catch issues like this. I'd also recommend running this on testnet with a few restarts to make sure the metrics work as expected. There's a command flag you can pass to dump a file with all the metrics.

src/bucket/BucketManagerImpl.h
```diff
@@ -59,6 +62,10 @@ class BucketManagerImpl : public BucketManager
     EvictionCounters mBucketListEvictionCounters;
     MergeCounters mMergeCounters;
     std::shared_ptr<EvictionStatistics> mEvictionStatistics{};
+    std::map<LedgerEntryTypeAndDurability, medida::Counter&>
+        mBucketListEntryCounts;
```
SirTyson, line 66: Nit: `mBucketListEntryCountCounters`

src/bucket/BucketManagerImpl.h
```diff
@@ -59,6 +62,10 @@ class BucketManagerImpl : public BucketManager
...
+    std::map<LedgerEntryTypeAndDurability, medida::Counter&>
+        mBucketListEntrySizes;
```
SirTyson, line 68: Nit: `mBucketListEntrySizeCounters`

src/bucket/Bucket.h
```diff
@@ -179,7 +181,37 @@ class Bucket : public std::enable_shared_from_this<Bucket>,
...
+    LIQUIDITY_POOL = 5,
+    TEMPORARY_CONTRACT_DATA = 6,
+    PERSISTENT_CONTRACT_DATA = 7,
+    CONTRACT_CODE = 8,
+    CONFIG_SETTING = 9,
+    TTL = 10
+};
+
+struct BucketEntryCounters
+{
```
SirTyson, line 204: Nit: Since this is a small struct where we're directly accessing the member, drop the m prefix and just say `entryTypeCounts`

src/bucket/BucketManagerImpl.cpp
```diff
@@ -1521,4 +1523,72 @@ BucketManagerImpl::getConfig() const
...
+            fmt::format(FMT_STRING("unknown LedgerEntryTypeAndDurability {:d}"),
+                        static_cast<uint32_t>(type)));
+    }
+}
+
+void
+BucketManagerImpl::reportBucketEntryCountMetrics()
+{
+    auto bucketEntryCounters = mBucketList->sumBucketEntryCounters();
+    for (auto& [type, count] : bucketEntryCounters.mEntryTypeCounts)
```
SirTyson, line 1565: Nit: These are primitive types that shouldn't change, just use `auto [type, count]`

src/bucket/BucketManagerImpl.cpp
```diff
@@ -1521,4 +1523,72 @@ BucketManagerImpl::getConfig() const
...
+    {
+        auto countsCounter = mBucketListEntryCounts.find(type);
+        if (countsCounter == mBucketListEntryCounts.end())
```
SirTyson, line 1568: This lazy construction is a little difficult to keep up with all the iterators and pairs floating around. I think it might be simpler to just construct all the metrics in the BucketManager constructor instead of having lazy instantiation logic here.

src/bucket/BucketManagerImpl.cpp
```diff
@@ -1521,4 +1523,72 @@ BucketManagerImpl::getConfig() const
 {
     return mApp.getConfig();
 }
+
+std::string
+to_string(LedgerEntryTypeAndDurability const type)
```
SirTyson, line 1528: Nit: `toString`. Also this would be a good utility function to move to the `LedgerTypeUtils` file as well.

SirTyson (changes requested):
A few more changes. Also, these new metrics should be added to `metrics.md`.

src/bucket/Bucket.h
```diff
@@ -181,37 +181,29 @@ class Bucket : public std::enable_shared_from_this<Bucket>,
...
-    std::map<LedgerEntryTypeAndDurability, size_t> mEntryTypeSizes;
+    std::map<LedgerEntryTypeAndDurability, size_t> entryTypeCounts;
+    std::map<LedgerEntryTypeAndDurability, size_t> entryTypeSizes;
 
     BucketEntryCounters& operator+=(BucketEntryCounters const& other);
 
     template <class Archive>
     void
-    serialize(Archive& ar) const
+    save(Archive& ar) const
```
SirTyson, line 198: Why did this change from the serialize function?

src/bucket/BucketIndexImpl.cpp
```diff
@@ -102,76 +102,15 @@ BucketIndexImpl<IndexT>::BucketIndexImpl(BucketManager& bm,
...
-            else if (bet == DEADENTRY)
-            {
-                let = be.deadEntry().type();
-                if (let == CONTRACT_DATA)
-                {
-                    isTemp = isTemporaryEntry(be.deadEntry());
-                }
-            }
-            else
+            if (bet == METAENTRY)
```
SirTyson, line 106: Nit: `be.type() == METAENTRY`

src/bucket/BucketIndexImpl.cpp
```diff
@@ -626,6 +565,44 @@ BucketIndexImpl<IndexT>::operator==(BucketIndex const& inRaw) const
         }
     }
 
+    if (mData.counters.entryTypeCounts.size() !=
```
SirTyson, line 568: Nit: `std::map` supports `==` so you can use that here instead.

src/bucket/BucketManagerImpl.cpp
```diff
@@ -188,6 +188,21 @@ BucketManagerImpl::BucketManagerImpl(Application& app)
     , mDeleteEntireBucketDirInDtor(
           app.getConfig().isInMemoryModeWithoutMinimalDB())
 {
+    for (uint32_t t =
```
SirTyson, line 191: Nit: An old C style trick is to add a last place holder value in an enum for easier, less footgunny iteration. See `enum TestDbMode` for an example, but the last element is usually called something like `NUM_TYPES`.

src/bucket/BucketManagerImpl.cpp
```diff
@@ -188,6 +188,21 @@ BucketManagerImpl::BucketManagerImpl(Application& app)
...
+             static_cast<uint32_t>(LedgerEntryTypeAndDurability::ACCOUNT);
+         t <= static_cast<uint32_t>(LedgerEntryTypeAndDurability::TTL); ++t)
+    {
+        auto typeString =
```
SirTyson, line 195: Nit: you can just static cast a local variable once here at the top.

src/bucket/BucketManagerImpl.cpp
```diff
@@ -1524,71 +1539,19 @@ BucketManagerImpl::getConfig() const
...
-                    .emplace(type,
-                             mApp.getMetrics().NewCounter(
-                                 {"bucket", "entrySizes", to_string(type)}))
-                    .first;
-        }
-        sizesCounter->second.set_count(
-            bucketEntryCounters.mEntryTypeSizes[type]);
+        mBucketListEntryCountCounters.find(type)->second.set_count(count);
+        mBucketListEntrySizeCounters.find(type)->second.set_count(
+            bucketEntryCounters.entryTypeSizes[type]);
```
SirTyson, line 1554: Nit: since this isn't asigning a value, use `at` instead of `[]`

src/ledger/LedgerTypeUtils.cpp
```diff
@@ -34,4 +35,103 @@ getTTLKey(LedgerKey const& e)
...
+    {
+        auto label = xdr::xdr_traits<BucketEntryType>::enum_name(bet);
+        throw std::runtime_error(
+            fmt::format("Unsupported BucketEntryType {}", label));
+    }
+    switch (let)
+    {
+    case ACCOUNT:
+        return LedgerEntryTypeAndDurability::ACCOUNT;
+        break;
```
SirTyson, line 67: Nit: redundant breaks.

SirTyson (approved):
LGTM! A couple last super nit picky things then we can merge.

src/bucket/Bucket.h
ThomasBrady: I had some cereal errors after adding counters to the `load` function of `BucketIndexImpl`, which I think I eventually solved by adding the counter map comparison to `BucketIndexImpl<T>::operator==`. I split `serialize` into `save` and `load` as part of that process but it wasn't necessary. I've reverted that change now.

ThomasBrady:
This PR is the dashboard changes FYI https://github.com/stellar/puppet-v4/pull/4081

SirTyson:
Last round of changes looks good! Once we have a serialization test, we'll be ready to merge.

ThomasBrady:
> Last round of changes looks good! Once we have a serialization test, we'll be ready to merge.

I'm pretty sure this is already test by `"serialize bucket indexes"`, but I added some extra assertions to make sure the counters are non-empty before the equality comparison.

SirTyson (dismissed):

src/bucket/BucketManagerImpl.cpp
```diff
@@ -1521,4 +1537,20 @@ BucketManagerImpl::getConfig() const
...
+BucketManagerImpl::reportBucketEntryCountMetrics()
+{
+    if (!mApp.getConfig().isUsingBucketListDB())
+    {
+        return;
+    }
+    auto bucketEntryCounters = mBucketList->sumBucketEntryCounters();
+    for (auto [type, count] : bucketEntryCounters.entryTypeCounts)
+    {
+        mBucketListEntryCountCounters.find(type)->second.set_count(count);
```
marta-lokhova, line 1551: Can we be more defensive and not potentially deference `end` here if type isn't found? These are just metrics, so we should handle bugs/failures as gracefully as possible, without crashing the node.

SirTyson (approved):

PR #4449, opened 2024-08-30
Windows fixes

MonsieurNicolas:
(refreshed previous PR that was conflicting with other changes in master)

@SirTyson -- I got a warning that `1UL << 36` is UB because the type is 32 bits

### DIFF
--- a/src/main/Config.cpp
+++ b/src/main/Config.cpp
@@ -996,17 +996,6 @@ Config::processConfig(std::shared_ptr<cpptoml::table> t)
         }
         std::vector<ValidatorEntry> validators;
         UnorderedMap<std::string, ValidatorQuality> domainQualityMap;
-        UnorderedMap<std::string, uint32_t*> uint32WithMax1Flags = {
-            {"PEER_READING_CAPACITY", &PEER_READING_CAPACITY},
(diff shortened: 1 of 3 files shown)


SirTyson (approved):
LGTM, I double checked the paper and this should always be 64 bits.

src/main/Config.cpp
```diff
@@ -1023,659 +1012,609 @@ Config::processConfig(std::shared_ptr<cpptoml::table> t)
...
-                    value == 0 ? std::nullopt : std::make_optional(value);
-            }
-            else if (item.first == "EMIT_SOROBAN_TRANSACTION_META_EXT_V1")
-            {
-                EMIT_SOROBAN_TRANSACTION_META_EXT_V1 = readBool(item);
-            }
-            else if (item.first == "EMIT_LEDGER_CLOSE_META_EXT_V1")
-            {
-                EMIT_LEDGER_CLOSE_META_EXT_V1 = readBool(item);
+            std::map<std::string, std::function<void()>> confProcessor = {
```
SirTyson, line 1016: Nit: `unordered_map`?

MonsieurNicolas: I don't think we care either way here. I went with the shortest version

Issue #4387, opened 2024-07-16
Consider updating the nomination leader algorithm to not bias towards larger orgs


MonsieurNicolas:
Looking at the current configuration of the network, it looks like there is some bias happening during nomination towards the "Lobstr" organization. This is probably caused by the different number of validators: all orgs except lobstr have a "2 out of 3" setup, where as lobstr has ["3 out of 5"](https://www.stellarbeat.io/nodes/GCFONE23AB7Y6C5YZOMKUKGETPIAJA4QOYLS5VNS4JHBGKRZCPYHDLW7?center=1&no-scroll=0).

In [the IETF draft paper](https://datatracker.ietf.org/doc/html/draft-mazieres-dinrg-scp-05#section-3.4), the algorithm used to select a leader during nomination is described as:
 *  Let `Gi(m) = SHA-256(i || m)`, where "||" denotes the concatenation of serialized XDR values.  Treat the output of "Gi" as a 256-bit binary number in big-endian format.
 * For each peer "v", define "weight(v)" as the fraction of quorum
      slices containing "v".
 * Define the set of nodes "neighbors(n)" as the set of nodes v for which `Gi(1 || n || v) < 2^{256} * weight(v)`, where "1" and "n"
      are both 32-bit XDR "int" values.  Note that a node is always its own neighbor because conceptually a node belongs to all of its own
      quorum slices.
 * Define `priority(n, v)` as `Gi(2 || n || v)`, where "2" and "n" are both 32-bit XDR "int" values
 *  For each round "n" until nomination has finished (see below), a node starts _echoing_ the available peer "v" with the highest value of
   `priority(n, v)` from among the nodes in `neighbors(n)`.  To echo "v", the node merges any valid values from "v"'s "voted" and "accepted" sets into its own "voted" set.

The implementation of the "weight" function used in core is slightly different and takes advantage of the quorum set structure:
https://github.com/stellar/stellar-core/blob/d78f48eacabb51753e34443de7618b956e61c59f/src/scp/LocalNode.cpp#L108-L134

Note as an aside that the implementation is wrong `n` should be `d-qset.threshold+1` (as the function is approximating the number of slices) but in the "2 out of 3" and "3 out of 5", the values end up being the same (so no impact at the org level), but at the top level, "5 out of 7" evaluates to `n=5` instead of `n=7-5+1=3` (this issue does not introduce bias towards any particular org).

[...]

PR #4453, opened 2024-09-04
Update weight function to fix bias in nomination

bboston7:
# Description

Part of #4387

This change adds an application-specific
`HerderSCPDriver::getNodeWeight` function that computes a validator's weight based on its quality as described in the config file. Much of this change is in the testing infrastructure, which stresses the new algorithm to demonstrate that it avoids biasing towards larger orgs.

The remaining work (to be done in follow up PRs) is:
* Documentation explaining the new algorithm in `src/herder/readme.md`
* A supercluster mission testing the upgrade from the old weight function to the new one.

# Checklist
- [x] Compiles
- [x] Ran all tests

### DIFF
--- a/docs/stellar-core_example.cfg
+++ b/docs/stellar-core_example.cfg
@@ -634,6 +634,10 @@ EMIT_SOROBAN_TRANSACTION_META_EXT_V1=false
 # https://github.com/stellar/stellar-xdr/commit/cdc339f5e74a75e8e558fd1a853397da71f1659a
 EMIT_LEDGER_CLOSE_META_EXT_V1=false
 
+# When set to true, Core will revert to using the old, application-agnostic
+# nomination weight function for SCP leader election.
+FORCE_OLD_STYLE_LEADER_ELECTION=false
+
 # EXCLUDE_TRANSACTIONS_CONTAINING_OPERATION_TYPE (list of strings) default is empty
 # Setting this will cause the node to reject transactions that it receives if
 # they contain any operation in this list. It will not, however, stop the node
--- a/src/herder/HerderSCPDriver.cpp
+++ b/src/herder/HerderSCPDriver.cpp
@@ -25,6 +25,7 @@
 #include "xdr/Stellar-ledger.h"
 #include <Tracy.hpp>
 #include <algorithm>
+#include <cmath>
 #include <fmt/format.h>
 #include <medida/metrics_registry.h>
 #include <numeric>
@@ -1283,5 +1284,72 @@ HerderSCPDriver::TxSetValidityKeyHash::operator()(
     hashMix(res, std::get<2>(key));
     hashMix(res, std::get<3>(key));
     return res;
+}
+
+uint64
+HerderSCPDriver::getNodeWeight(NodeID const& nodeID, SCPQuorumSet const& qset,
+                               bool const isLocalNode) const
+{
+    Config const& cfg = mApp.getConfig();
+    bool const unsupportedProtocol = protocolVersionIsBefore(
+        mApp.getLedgerManager()
+            .getLastClosedLedgerHeader()
+            .header.ledgerVersion,
+        APPLICATION_SPECIFIC_NOMINATION_LEADER_ELECTION_PROTOCOL_VERSION);
+    if (unsupportedProtocol || !cfg.VALIDATOR_WEIGHT_CONFIG.has_value() ||
+        cfg.FORCE_OLD_STYLE_LEADER_ELECTION)
+    {
+        // Fall back on old weight algorithm if any of the following are true:
+        // 1. The network has not yet upgraded to
+        //    APPLICATION_SPECIFIC_NOMINATION_LEADER_ELECTION_PROTOCOL_VERSION,
+        // 2. The node is using manual quorum set configuration, or
+        // 3. The node has the FORCE_OLD_STYLE_LEADER_ELECTION flag
+        //    set
+        return SCPDriver::getNodeWeight(nodeID, qset, isLocalNode);
+    }
+
+    ValidatorWeightConfig const& vwc =
+        mApp.getConfig().VALIDATOR_WEIGHT_CONFIG.value();
+
+    auto entryIt = vwc.mValidatorEntries.find(nodeID);
+    if (entryIt == vwc.mValidatorEntries.end())
+    {
+        // This shouldn't be possible as the validator entries should contain
+        // all validators in the config. For this to happen, `getNodeWeight`
+        // would have to be called with a non-validator `nodeID`. Throw if
+        // building tests, and otherwise fall back on the old algorithm.
+        throw std::runtime_error(
+            fmt::format(FMT_STRING("Validator entry not found for node {}"),
+                        toShortString(nodeID)));
+    }
+
+    ValidatorEntry const& entry = entryIt->second;
+    auto homeDomainSizeIt = vwc.mHomeDomainSizes.find(entry.mHomeDomain);
+    if (homeDomainSizeIt == vwc.mHomeDomainSizes.end())
+    {
+        // This shouldn't be possible as the home domain sizes should contain
+        // all home domains in the config. For this to happen, `getNodeWeight`
+        // would have to be called with a non-validator, or the config parser
+        // would have to allow a validator without a home domain. Throw if
+        // building tests, and otherwise fall back on the old algorithm.
+        throw std::runtime_error(
+            fmt::format(FMT_STRING("Home domain size not found for domain {}"),
+                        entry.mHomeDomain));
+    }
+
+    auto qualityWeightIt = vwc.mQualityWeights.find(entry.mQuality);
+    if (qualityWeightIt == vwc.mQualityWeights.end())
+    {
+        // This shouldn't be possible as the quality weights should contain all
+        // quality levels in the config.
+        throw std::runtime_error(
+            fmt::format(FMT_STRING("Quality weight not found for quality {}"),
+                        static_cast<int>(entry.mQuality)));
+    }
+
+    // Node's weight is its quality's weight divided by the number of nodes in
+    // its home domain
+    releaseAssert(homeDomainSizeIt->second > 0);
+    return qualityWeightIt->second / homeDomainSizeIt->second;
 }
 }
--- a/src/herder/HerderSCPDriver.h
+++ b/src/herder/HerderSCPDriver.h
@@ -9,6 +9,7 @@
 #include "herder/TxSetUtils.h"
 #include "medida/timer.h"
 #include "scp/SCPDriver.h"
+#include "util/ProtocolVersion.h"
 #include "util/RandomEvictionCache.h"
 #include "xdr/Stellar-ledger.h"
 #include <optional>
@@ -31,6 +32,11 @@ class Upgrades;
 class VirtualTimer;
 struct StellarValue;
 struct SCPEnvelope;
+
+// First protocol version supporting the application-specific weight function
+// for SCP leader election.
+ProtocolVersion constexpr APPLICATION_SPECIFIC_NOMINATION_LEADER_ELECTION_PROTOCOL_VERSION =
+    ProtocolVersion::V_22;
 
 class HerderSCPDriver : public SCPDriver
 {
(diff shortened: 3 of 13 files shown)


src/herder/HerderSCPDriver.h
```diff
@@ -32,6 +33,11 @@ class VirtualTimer;
 struct StellarValue;
 struct SCPEnvelope;
 
+// First protocol version supporting the application-specific weight function
+// for SCP leader election.
+ProtocolVersion constexpr APPLICATION_SPECIFIC_WEIGHT_PROTOCOL_VERSION =
```
marta-lokhova, line 38: nit: could we name this parameter such that it's clear we're talking about nomination protocol and leader election?

src/herder/HerderSCPDriver.cpp
```diff
@@ -1285,4 +1286,84 @@ HerderSCPDriver::TxSetValidityKeyHash::operator()(
...
+            fmt::format(FMT_STRING("Validator entry not found for node {}"),
+                        toShortString(nodeID)));
+#endif
+        return SCPDriver::getNodeWeight(nodeID, qset, isLocalNode);
+    }
+
+    ValidatorEntry const& entry = entryIt->second;
+    if (entry.mQuality == ValidatorQuality::VALIDATOR_LOW_QUALITY)
+    {
+        // Special case: low quality validators have weight `0`
```
marta-lokhova, line 1332: I know this was in the original design, but I'm now wondering if there's a reason to not give a chance to LOW quality nodes at all?

src/herder/HerderSCPDriver.cpp
```diff
@@ -1285,4 +1286,84 @@ HerderSCPDriver::TxSetValidityKeyHash::operator()(
...
+        // all home domains in the config. For this to happen, `getNodeWeight`
+        // would have to be called with a non-validator, or the config parser
+        // would have to allow a validator without a home domain. Throw if
+        // building tests, and otherwise fall back on the old algorithm.
+#ifdef BUILD_TESTS
+        throw std::runtime_error(
+            fmt::format(FMT_STRING("Home domain size not found for domain {}"),
+                        entry.mHomeDomain));
+#endif
+        return SCPDriver::getNodeWeight(nodeID, qset, isLocalNode);
```
marta-lokhova, line 1349: In general, I'd prefer not to change behavior in prod vs tests (we recently had some bugs related to this). This shouldn't happen in prod either, right?

src/herder/test/HerderTests.cpp
```diff
@@ -5706,3 +5707,416 @@ TEST_CASE("SCP message capture from previous ledger", "[herder]")
...
+
+// Spin up a simulation of validators and run for `numLedgers`. After running,
+// check that the win percentages of each node and org are within 5% of the
+// expected win percentages. On large runs with higher `numLedgers` counts, this
+// can be quite slow. It's a more thorough test of the fairness of the
+// nomination algorithm than the `getNodeWeight` test, but at the expense of not
+// being capable of testing as many configurations due to the time it takes to
+// run.
+static void
+testWinProbabilities(std::vector<SecretKey> const& sks,
```
marta-lokhova, line 5961: Do we need full Simulation here? We should be able to just run leader election (similar to SCP tests) on its own without the rest of consensus (which I think is why the test is taking a long time).

marta-lokhova (changes requested):

src/herder/HerderSCPDriver.cpp
bboston7: This allows for evaluation of new validators who will perform most SCP tasks except for choosing txsets.

CCing @MonsieurNicolas because this was originally his idea.

docs/stellar-core_example.cfg
```diff
@@ -634,6 +634,10 @@ EMIT_SOROBAN_TRANSACTION_META_EXT_V1=false
 # https://github.com/stellar/stellar-xdr/commit/cdc339f5e74a75e8e558fd1a853397da71f1659a
 EMIT_LEDGER_CLOSE_META_EXT_V1=false
 
+# When set to true, Core will revert to using the old, application-agnostic
+# weight function for SCP leader election.
+FORCE_APPLICATION_AGNOSTIC_WEIGHT_FUNCTION=false
```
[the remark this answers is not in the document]
bboston7: Renamed!

src/herder/HerderSCPDriver.cpp
```diff
@@ -1285,4 +1286,84 @@ HerderSCPDriver::TxSetValidityKeyHash::operator()(
     hashMix(res, std::get<3>(key));
     return res;
 }
+
+uint64
+HerderSCPDriver::getNodeWeight(NodeID const& nodeID, SCPQuorumSet const& qset,
+                               bool const isLocalNode) const
+{
+    Config const& cfg = mApp.getConfig();
+    if (protocolVersionIsBefore(mApp.getLedgerManager()
```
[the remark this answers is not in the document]
bboston7: Good point. I extracted the protocol version check into a variable.

src/simulation/Simulation.h
```diff
@@ -47,9 +47,11 @@ class Simulation
...
     void setCurrentVirtualTime(VirtualClock::system_time_point t);
 
-    // Add new node to the simulation. This function does not start the node.
-    // Callers are expected to call `start` or `startAllNodes` manually.
-    Application::pointer addNode(SecretKey nodeKey, SCPQuorumSet qSet,
+    // Add new node to the simulation. `cfg.QUORUM_SET` is not modified if
+    // `qSet` is `nullopt`.  This function does not start the node.  Callers are
+    // expected to call `start` or `startAllNodes` manually.
+    Application::pointer addNode(SecretKey nodeKey,
+                                 std::optional<SCPQuorumSet> qSet,
```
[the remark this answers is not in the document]
bboston7: > I think I'll end up reverting this because I'm going to try to rewrite the test to avoid using `Simulation`.

This is exactly what I wound up doing.

src/herder/HerderSCPDriver.h
bboston7: Renamed!

src/herder/HerderSCPDriver.cpp
bboston7: No, it shouldn't happen in prod either. I changed this to `throw` regardless of `BUILD_TESTS`

src/herder/test/HerderTests.cpp
bboston7: I rewrote this test to remove `Simulation` and just test `NomiationProtocol`. In doing so I also bumped up the number of slots tested for each scenario to 10,000.

src/scp/NominationProtocol.cpp
```diff
@@ -254,6 +254,15 @@ NominationProtocol::updateRoundLeaders()
...
             return true;
         });
+
+        if (topPriority == 0)
+        {
+            // No one had priority, so all nodes would choose themselves
+            // resulting in a timeout. Clear newRoundLeaders, allowing the
+            // algorithm to fast timeout and try again.
+            newRoundLeaders.clear();
+        }
```
bboston7, line 264: I want to call this out again to double check my thinking about the worst-case scenario with this change. If some nodes are running the new algorithm and some are running the old algorithm (because they have `QUORUM_SET` manually defined), then it's possible that the nodes running the new algorithm hit this case and fast timeout, while the nodes running the old algorithm do not fast timeout. In that case, their `mRoundNumber` values diverge. This has an impact in a few places:

1. timeout lengths differ. I don't think this is a big problem as it's equivalent to some nodes having a worse connection than others.
2.  `hashNode` values differ. `hashNode` is used for the priority function, which already differs between the two algorithms. We've already reasoned about the impact of nodes running each algorithm choosing different leaders (potential timeouts) and are OK with them.
3. `hashValue` values differ. The impact of this is that nodes running each algorithm may vote for different values initially. However, over time nodes expand their sets of voted for values and they should still eventually converge on some values with enough votes to cross the threshold. The impact here is potential additional timeouts while waiting for convergence.

Simulation has demonstrated that this `topPriority == 0` case is rare when modeling Tier 1 (affecting ~0.04% of slots). It will become even rarer as we expand the validator set. Nevertheless, part of rolling this change out is communication with validators to ensure they're using automatic quorum set configuration, and therefore will end up using the new algorithm after the vote.

Does this reasoning sound correct? @MonsieurNicolas @marta-lokhova

bboston7:
This is ready for another round of review!

src/simulation/Simulation.h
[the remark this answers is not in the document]
marta-lokhova: Ah nice, glad we ended up not needing any changes in Simulation.

src/herder/test/HerderTests.cpp
marta-lokhova: nice, thanks! Did it speed up the test?

src/main/Config.h
```diff
@@ -37,24 +37,39 @@ enum class ValidationThresholdLevels : int
...
+struct ValidatorWeightConfig
+{
+    // Mapping from node ids to info about each validator
+    UnorderedMap<NodeID, ValidatorEntry> mValidatorEntries;
+
+    // Mapping from org names to the number of validators in that org
+    UnorderedMap<std::string, size_t> mHomeDomainSizes;
+
+    // Highest quality level of any validator in the config
+    ValidatorQuality mHighestQuality;
```
marta-lokhova, line 68: we need to initialize this enum, otherwise it's not safe to use in the max function below (UB)

src/herder/HerderSCPDriver.cpp
```diff
@@ -1285,4 +1286,79 @@ HerderSCPDriver::TxSetValidityKeyHash::operator()(
...
+    // Then, compute the number of quality levels between the highest quality
+    // org and this node's org
+    int const qualityDelta = static_cast<int>(vwc.mHighestQuality) -
+                             static_cast<int>(entry.mQuality);
+    releaseAssert(qualityDelta >= 0);
+
+    // Finally, divide the theoretical max weight for nodes in this org
+    // (`numerator`) by `10^{qualityDelta}` to get the final weight.
+    uint64 const denominator =
+        static_cast<uint64>(std::pow(10.0, qualityDelta));
```
marta-lokhova, line 1360: The new weight function looks good to me, but going back to the design doc, it's actually different from the one proposed, isn't it? Is it intended?

src/herder/HerderSCPDriver.cpp
```diff
@@ -1285,4 +1286,79 @@ HerderSCPDriver::TxSetValidityKeyHash::operator()(
...
+        // would have to allow a validator without a home domain. Throw if
+        // building tests, and otherwise fall back on the old algorithm.
+        throw std::runtime_error(
+            fmt::format(FMT_STRING("Home domain size not found for domain {}"),
+                        entry.mHomeDomain));
+    }
+
+    // First, divide the theoretical maximum organization weight by the total
+    // number of nodes in that organization
+    uint64 const numerator = UINT64_MAX / homeDomainSizeIt->second;
```
marta-lokhova, line 1349: I'd prefer to keep the types consistent if possible (size_t vs uint64)

marta-lokhova (commented):
one issue that needs addressing, otherwise this change looks good to me!

marta-lokhova:
Feel free to squash fixup commits and rebase also to prepare for merging

src/herder/HerderSCPDriver.cpp
bboston7: Oh wow, this is embarrassing. Dropping the `|O_q|` term from the design doc was not intentional.

When we added the multiplicative factor in the denominator, the importance of the `|O_q|` term was significantly reduced, to the point that I completely forgot about it.  As you noted, the weight function without it still satisfies the fairness criteria we care about.

However, dropping the term overweights `MEDIUM` quality validators a little bit beyond what we originally planned, so I felt it was worth fixing. Adding the term back involved calculating quality-level weights (`w_q` in the design doc) in `Config::setValidatorWeightConfig` and tweaking `HerderSCPDriver::getNodeWeight` to use those new weights. Please take another look at those two functions now that this is fixed.

src/main/Config.h
bboston7: Good catch. This enum is no longer in the struct with the change to bring the implementation in line with the design doc.

src/herder/test/HerderTests.cpp
bboston7: It made a huge difference! A single simulation used to take up to 2 hours, but now the test runs 13 simulations in under a minute.

bboston7:
> Feel free to squash fixup commits and rebase also to prepare for merging

Done!

src/herder/HerderSCPDriver.cpp
MonsieurNicolas: yes exactly -- this allows an org to add as LOW validators to their own qset without risking their own reputation. We'll need to document this

src/scp/NominationProtocol.cpp
MonsieurNicolas: yeah I agree that round numbers diverging is not the worst problem as rounds are supposed to be "additive" (as in: if a node is behind, it may actually unstick a node that is ahead as the ones ahead are supposed to listen to the union of leaders from previous rounds).

For the divergence in hashing, maybe we should have a flag to force using the old behavior (based on the generated quorum set) for situations where this would be a problem?

MonsieurNicolas: nm -- the flag is just `FORCE_OLD_STYLE_LEADER_ELECTION` so I think we're good

marta-lokhova (approved):

Issue #4451, opened 2024-09-03
Soroban tests need to work with `closeLedger`


dmkozh:
Currently almost all the Soroban tests rely on directly committing LTXs, which will no longer work with BucketListDB. There is currently a 'workaround' that just closes the ledger, but skips the tx result and tx meta and both are necessary for Soroban tests to function properly. Thus before we can disable SQL DB for the Soroban tests, we have to populate all the necessary data and fix all the tests that break due to using `closeLedger` (some of the tests also rely on the ledger sequence not being bumped between applied transactions).

PR #4455, opened 2024-09-04
Switch Soroban tests to always use closeLedger to apply txs.

dmkozh:
# Description

Resolves https://github.com/stellar/stellar-core/issues/4451

Switch Soroban tests to always use closeLedger to apply txs. 

This is a necessary change for the BucketListDB support as it doesn't materialize changes on LTX commit.  I've also changed a single test ('basic contract invocation') to only use in-memory LTX in order to cover lower-level logic.

# Checklist
- [ ] Compiles
- [ ] Ran all tests


src/transactions/test/SorobanTxTestUtils.cpp
```diff
@@ -1097,28 +1106,24 @@ SorobanTest::isTxValid(TransactionTestFramePtr tx)
     return ret;
 }
 
-bool
+TransactionResult
```
SirTyson, line 1109: There's an interface inconsistency here with returning a `TransactionResult` object while taking in a `TransactionTestFramePtr` parameter. The idea behind `TransactionTestFramePtr` is that the `TransactionResult` is always bundled with the tx object, allowing for calls like `tx->getResult`. The issue is, here we call `ledgerClose({tx})`. This will convert `{tx}` into a XDR TxSet object, and since we're calling the prod `ledgerClose` path, the transaction inside the XDR TxSet get converted to non-test `TransactionFrame` objects. The result is, the `TransactionResult` returned by `invokeTx` is different than the `TransactionResult` stored inside the test `tx` object. This seems like a footgun, as many tx tests rely on `TransactionTestFrame::getResult`. Sorry, this was a short sight on my end and something I ran into on my test refactors as well.

What I've been doing is using the result set from `closeLedger` to set the cached `TransactionResult` inside `TransactionTestFrame` ([see `applyTx`in my branch](https://github.com/SirTyson/stellar-core/blob/cd7f3100208eb13606ada0018199f032796a23fb/src/test/TxTests.cpp#L388)). I think it's going to be kinda hacky either way given the nature of `TransactionTestFrame` (my bad), but I prefer keeping `TransactionTestFrame->getResult` correct over returning the result to be consistent with other test interfaces, open to suggestions though.

src/transactions/test/InvokeHostFunctionTests.cpp
```diff
@@ -402,12 +402,23 @@ TEST_CASE("basic contract invocation", "[tx][soroban]")
                       SorobanInvocationSpec const& spec,
                       std::optional<uint32_t> expectedRefund = std::nullopt,
                       bool addContractKeys = true) {
-        int64_t initBalance = test.getRoot().getBalance();
+        auto rootAccount = test.getRoot();
```
SirTyson, line 405: Note: I've added a new TestDBMode so you don't have to do this manually, I'll update in my branch after rebasing.

src/transactions/test/InvokeHostFunctionTests.cpp
```diff
@@ -2065,17 +2072,21 @@ TEST_CASE("state archival", "[tx][soroban]")
...
+        REQUIRE(test.getTTL(contractKeys[1]) ==
+                originalExpectedLiveUntilLedger);
 
         // Contract instance and code are expired, any TX should fail
         REQUIRE(client.put("temp", ContractDataDurability::TEMPORARY, 0) ==
                 INVOKE_HOST_FUNCTION_ENTRY_ARCHIVED);
 
         auto newExpectedLiveUntilLedger =
-            stateArchivalSettings.minPersistentTTL + ledgerSeq - 1;
+            (test.getLCLSeq() + 1) + stateArchivalSettings.minPersistentTTL - 1;
```
SirTyson, line 2089: Comment here would be helpful, as well as getting rid of the +1 followed by -1.

src/transactions/test/InvokeHostFunctionTests.cpp
```diff
@@ -2514,9 +2544,9 @@ TEST_CASE("temp entry eviction", "[tx][soroban]")
         cfg.DEPRECATED_SQL_LEDGER_STATE = !enableBucketListDB;
         cfg.BACKGROUND_EVICTION_SCAN = backgroundEviction;
 
-        // overrideSorobanNetworkConfigForTest commits directly to the database,
-        // will not work if BucketListDB is enabled so we must use the cfg
-        // override
+        // overrideSorobanNetworkConfigForTest commits directly to the
```
SirTyson, line 2547: I think we can remove this now.

src/transactions/test/InvokeHostFunctionTests.cpp
```diff
@@ -2532,44 +2562,40 @@ TEST_CASE("temp entry eviction", "[tx][soroban]")
...
-        // When BucketListDB is enabled, invoking any operation closes a ledger,
-        // messing up the eviction boundary we are trying to test. SQL commits
-        // directly to the DB without advancing the LCL, so only test this when
-        // SQL is enabled
-        if (!enableBucketListDB)
-        {
-            // This should be a noop
-            test.invokeExtendOp({lk}, 10'000, 0);
-            REQUIRE(test.getTTL(lk) == expectedLiveUntilLedger);
+        // When BucketListDB is enabled, invoking any operation closes a
```
SirTyson, line 2583: Remove comment out commented out if statement.

SirTyson (commented):
Thanks for this change, looks like you had to wade through off by one hell for this one...

Overall LGTM and is very helpful for my BucketList test changes! A few small nit picks and on unfortunate interface inconsistency with `TransactionTestFrame`.

src/transactions/test/InvokeHostFunctionTests.cpp
dmkozh: Moved the code around to follow the usual pattern of LCL + min TTL - 1, so hopefully it's more clear now.

src/transactions/test/InvokeHostFunctionTests.cpp
dmkozh: #Hmm, I haven't changed anything around that logic, have I? Or should this be generally addressed in some other change set?

src/transactions/test/InvokeHostFunctionTests.cpp
dmkozh: Done

src/transactions/test/SorobanTxTestUtils.cpp
dmkozh: We only use test frame (anti-)pattern for the legacy tests, here it's easy enough to just use the proper tx frame. The same goes for the loadgen (which happens to use this function), so I just got rid of test frames for both.

PR #4458, opened 2024-09-05
bugfix: guard access to mNoOutboundCapacity

marta-lokhova:
Fix issue discovered by static analysis (data race on `mNoOutboundCapacity`). This variable is used to trigger recovery in case flow control gets stuck. In practice, this triggers very rarely (if at all), which is probably why we did not catch this before.

### DIFF
--- a/src/overlay/FlowControl.h
+++ b/src/overlay/FlowControl.h
@@ -164,14 +164,10 @@ class FlowControl
     SendMoreCapacity endMessageProcessing(StellarMessage const& msg);
     bool canRead() const;
 
-    // This method return last timestamp (if any) when peer had no available
-    // outbound capacity (useful to diagnose if the connection is stuck for any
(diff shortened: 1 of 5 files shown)


src/overlay/FlowControl.h
```diff
@@ -167,11 +167,8 @@ class FlowControl
     // This method return last timestamp (if any) when peer had no available
     // outbound capacity (useful to diagnose if the connection is stuck for any
```
SirTyson, line 168: Nit: Update comment for new return type.

SirTyson (approved):
LGTM, one nitpick.

src/overlay/FlowControl.h
marta-lokhova: thanks, it's fixed now.

SirTyson (approved):

PR #4462, opened 2024-09-09
Change default test backend to BucketListDB

### LINKED ISSUE #4433: BucketListDB test coverage
- [ ] do an audit of the test suite, identify any other tests where coverage is missing (there are more, e.g. upgrades)
- [ ] leave tests that don’t need buckets but need in-memory ledger txn as-is, explicitly enable SQL on them such that in tests.cpp we can still enable BL DB.

The goal of this issue is to remove DEPRECATED_SQL_LEDGER_STATE=true from test.cpp

SirTyson:
# Description

Resolves #4433

This switches the default testing backend to BucketListDB. While some test still disable BucketListDB and commit directly via `LedgerTxn` to test lower level subsystem, such as operation application, the majority of tests and all end-to-end tests no use BucketListDB. This is rebased on top of https://github.com/stellar/stellar-core/pull/4455.

# Checklist
- [x] Compiles
- [x] Ran all tests


src/main/Config.h
```diff
@@ -116,6 +116,23 @@ class Config : public std::enable_shared_from_this<Config>
 
     typedef std::shared_ptr<Config> pointer;
 
+    // These test modes should be used in the following contexts:
+    // 1. TESTDB_DEFAULT / TESTDB_BUCKET_DB_VOLATILE: provides the most
+    //    comprehensive end-to-end test, but does not allow arbitrary ledger
```
dmkozh, line 121: I'm not sure I understand what does 'arbitrary ledger writes' mean. I think it would be more actionable if you explained the actual test implementation implications. IIUC this mode doesn't persist writes between LTXs, but I'm not sure if this is what you're trying to convey with 'arbitrary ledger writes' here.

src/ledger/LedgerManagerImpl.cpp
```diff
@@ -1725,8 +1726,9 @@ LedgerManagerImpl::transferLedgerEntriesToBucketList(
     ltx.getAllEntries(initEntries, liveEntries, deadEntries);
     if (blEnabled)
     {
-        mApp.getBucketManager().addBatch(mApp, lh, initEntries, liveEntries,
-                                         deadEntries);
+        BucketTestUtils::addBatchAndUpdateSnapshot(
```
dmkozh, line 1729: This looks really suspicious, how come we're using a test utility in (seemingly) the prod code path?

src/testdata/ledger-close-meta-v0-protocol-0.json
```diff
@@ -3,24 +3,24 @@
         "v": 0,
         "v0": {
             "ledgerHeader": {
-                "hash": "6c275540e30a8b79994b45bb4a9792fea4d1a1129b017307009e15972f2fdd7f",
+                "hash": "f4ce7471b7c9b6f8be7ab97d06c7d0d832825d4b5bbbf4674b1067582d44ee49",
```
dmkozh, line 6: Why did the hashes change?

src/ledger/LedgerManagerImpl.cpp
SirTyson: You're right, I messed up the rebase, this is supposed to be on the test-only variant,  `LedgerManagerForTesting::transferLedgerEntriesToBucketList`

src/bucket/test/BucketListTests.cpp
```diff
@@ -664,7 +673,9 @@ TEST_CASE("BucketList check bucket sizes", "[bucket][bucketlist][count]")
     Application::pointer app = createTestApplication(clock, cfg);
     BucketList& bl = app->getBucketManager().getBucketList();
     std::vector<LedgerKey> emptySet;
-    auto ledgers = LedgerTestUtils::generateValidUniqueLedgerEntries(256);
+    auto ledgers =
+        LedgerTestUtils::generateValidUniqueLedgerEntriesWithExclusions(
```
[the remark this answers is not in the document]
SirTyson: There we several BucketList tests that would occasionally fail with bad Network Config asserts when random network config entries were generated. This happened in the past (like in the Bucket test failure a couple weeks ago) as well as when I was working on this refactor, so I changed all the Bucket tests to by default not populate Network Config settings. The exception is the "merging bucket entries," which is a low level tests that checks merge behavior for all entry types, including Config Setting entries. This should give us enough coverage of type specific BucketList behavior without having to worry about assert failures in higher level tests.

src/testdata/ledger-close-meta-v0-protocol-0.json
SirTyson: Previously, all the test setup, such as creating accounts, assets, etc, committed directly to SQL but did not add entries to the BucketList. Now we actually commit these to the BucketList, so BucketList hash and ledger header hashes have changed in all meta tests. I did an eye check, and the actual meta contents does not appear to change outside the hashes.

src/main/Config.h
SirTyson: I've updated the comment, as well as changed the test mode names to make this more clear.

src/ledger/LedgerManagerImpl.cpp
```diff
@@ -5,6 +5,7 @@
 #include "ledger/LedgerManagerImpl.h"
 #include "bucket/BucketList.h"
 #include "bucket/BucketManager.h"
+#include "bucket/test/BucketTestUtils.h"
```
marta-lokhova, line 8: unused

src/main/Config.h
```diff
@@ -116,14 +116,40 @@ class Config : public std::enable_shared_from_this<Config>
...
     typedef std::shared_ptr<Config> pointer;
 
+    // These test modes should be used in the following contexts:
+    // 1. TESTDB_DEFAULT / TESTDB_BUCKET_DB_VOLATILE: provides the most
+    //    comprehensive end-to-end test, but does not support arbitrary ledger
+    //    state writes via ltx root commits. All ledger state changes must occur
+    //    via applying valid TXs or manually adding entries to the BucketList.
+    //    BucketList state is not preserved over restarts. If this mode can be
+    //    used, it should be.
+    // 2. TESTDB_IN_MEMORY_NO_OFFERS: allows arbitrary ledger state writes via
```
marta-lokhova, line 126: do we need both `TESTDB_IN_MEMORY_NO_OFFERS` and `TESTDB_IN_MEMORY_OFFERS`? seems like we can get away with just a single mode that stores everything, no?

marta-lokhova (approved):
LGTM, just one question. Thanks for this change, really happy to see improvements to test coverage!

marta-lokhova (approved):

PR #4482, opened 2024-09-24
Bump env to version with wasmi 0.36.1

graydon:
This also brings in p22 updates for k256 and dalek libraries.

### DIFF
--- a/src/rust/src/dep-trees/p22-expect.txt
+++ b/src/rust/src/dep-trees/p22-expect.txt
@@ -70,7 +70,7 @@ soroban-env-host v22.0.0 (src/rust/soroban/p22/soroban-env-host)
 │   │   │   │   │   ├── block-buffer v0.10.4
 │   │   │   │   │   │   └── generic-array v0.14.7
 │   │   │   │   │   │       ├── typenum v1.16.0
-│   │   │   │   │   │       └── zeroize v1.6.0
+│   │   │   │   │   │       └── zeroize v1.8.1
 │   │   │   │   │   │           └── zeroize_derive v1.4.2 (proc-macro)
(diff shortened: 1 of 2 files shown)


src/rust/src/dep-trees/p22-expect.txt
```diff
@@ -169,35 +168,36 @@ soroban-env-host v22.0.0 (src/rust/soroban/p22/soroban-env-host)
...
-├── elliptic-curve v0.13.5 (*)
+│   ├── subtle v2.5.0
+│   └── zeroize v1.8.1 (*)
+├── elliptic-curve v0.13.8 (*)
 ├── generic-array v0.14.7 (*)
 ├── getrandom v0.2.11 (*)
 ├── hex-literal v0.4.1
 ├── hmac v0.12.1 (*)
-├── k256 v0.13.1
+├── k256 v0.13.4
```
anupsdf, line 197: Shouldn't this be v0.13.3 like we have in latest env?

graydon: the latest env specifies "0.13.3" in its Cargo.toml file but as mentioned that is implicitly a loose (non-locked) caret-type constraint -- equivalent to "^0.13.3" which resolved to 0.13.4 in the committed env lockfile. Which is why it's 0.13.4 here. This file is literally generated from the env lockfile directly now. It's just another copy of it as a check to catch unanticipated changes.

graydon: See: https://github.com/stellar/rs-soroban-env/blob/main/soroban-env-host/Cargo.toml#L36 and https://github.com/stellar/rs-soroban-env/blob/main/Cargo.lock#L857

anupsdf (approved):

PR #4492, opened 2024-10-01
Convert live entries to init entries in lowest level of bucketlist

ThomasBrady:
# Description
Resolves [#193](https://github.com/stellar/stellar-core-internal/issues/193)

Changes `BucketOutputIterator::put` to convert `LIVEENTRY` to `INITENTRY` if `mKeepDeadEntries == true`. 

Perhaps this variable should be renamed to `mIsLowestLevel` or something now that it affects dead entry persistence and init entry to live entry conversion. 

# Checklist
- [ ] Compiles
- [ ] Ran all tests

### DIFF
--- a/src/bucket/BucketOutputIterator.cpp
+++ b/src/bucket/BucketOutputIterator.cpp
@@ -66,17 +66,32 @@ BucketOutputIterator::put(BucketEntry const& e)
         ++mMergeCounters.mOutputIteratorTombstoneElisions;
         return;
     }
+    std::optional<BucketEntry> maybeInitEntry;
+    if (!mKeepDeadEntries && e.type() == LIVEENTRY)
+    {
+        // If mKeepDeadEntries is false (lowest level), 
+        // we also want to convert the LIVEENTRY to an INITENTRY.
+        // This is because each level of the bucket list contains 
+        // only one entry per key, and per CAP-0020, INIT ENTRY
+        // implies that either no entry with the same ledger key 
+        // exists in an older bucket. Therefore, all entries of type 
+        // LIVEENTRY in the lowest level are also of type INITENTRY.
+        ++mMergeCounters.mOutputIteratorLiveToInitConversions;
+        // Make a copy of e and set the type of the new entry to INITENTRY.
+        maybeInitEntry.emplace(e);
+        maybeInitEntry->type(INITENTRY);
+    }
 
     // Check to see if there's an existing buffered entry.
     if (mBuf)
     {
         // mCmp(e, *mBuf) means e < *mBuf; this should never be true since
         // it would mean that we're getting entries out of order.
-        releaseAssert(!mCmp(e, *mBuf));
+        releaseAssert(!mCmp(maybeInitEntry.has_value() ? *maybeInitEntry : e, *mBuf));
 
         // Check to see if the new entry should flush (greater identity), or
         // merely replace (same identity), the buffered entry.
-        if (mCmp(*mBuf, e))
+        if (mCmp(*mBuf, maybeInitEntry.has_value() ? *maybeInitEntry : e))
         {
             ++mMergeCounters.mOutputIteratorActualWrites;
             mOut.writeOne(*mBuf, &mHasher, &mBytesPut);
@@ -90,7 +105,7 @@ BucketOutputIterator::put(BucketEntry const& e)
 
     // In any case, replace *mBuf with e.
     ++mMergeCounters.mOutputIteratorBufferUpdates;
-    *mBuf = e;
+    *mBuf = maybeInitEntry.has_value() ? *maybeInitEntry : e;
 }
 
 std::shared_ptr<Bucket>
--- a/src/bucket/test/BucketTests.cpp
+++ b/src/bucket/test/BucketTests.cpp
@@ -399,6 +399,36 @@ TEST_CASE("merges proceed old-style despite newer shadows",
     }
 }
 
+TEST_CASE("lowest level merge converts live to init", "[bucket][bucketinit]")
+{
+    VirtualClock clock;
+    Config const& cfg = getTestConfig();
+    Application::pointer app = createTestApplication(clock, cfg);
+    auto& bm = app->getBucketManager();
+    auto vers = getAppLedgerVersion(app);
+
+    LedgerEntry entry = generateAccount();
+    auto b1 = Bucket::fresh(bm, vers, {}, {entry}, {},
+                               /*countMergeEvents=*/true, clock.getIOContext(),
+                               /*doFsync=*/true);
+    EntryCounts e1(b1);
+    CHECK(e1.nInit == 0);
+    CHECK(e1.nLive == 1);
+    auto b2 = Bucket::fresh(bm, vers, {}, {}, {},
+                               /*countMergeEvents=*/true, clock.getIOContext(),
+                               /*doFsync=*/true);
+
+    // Merge b1 and b2 into a new, lowest-level bucket.
+    // keepDeadEntries = false signfies the lowest level of the bucket list
+    auto bMerged = Bucket::merge(bm, vers, b1, b2, /*shadows=*/{},
+                            /*keepDeadEntries=*/false,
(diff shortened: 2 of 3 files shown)


ThomasBrady:
@SirTyson can you PTAL?

src/bucket/BucketOutputIterator.cpp
```diff
@@ -67,17 +67,36 @@ BucketOutputIterator::put(BucketEntry const& e)
         ++mMergeCounters.mOutputIteratorTombstoneElisions;
         return;
     }
+    std::optional<BucketEntry> maybeInitEntry;
```
SirTyson, line 70: I don't think you need to do this check here. If you move this to the bottom right before `*mBuf = ...`, you could get rid of the optional the complicated referencing later on and just say
```
if (shouldRewriteAsInit)
{
auto copy = e;
copy = init;
*mBuf = init;
}
else 
{
*mBuf = e;
}
```

src/bucket/test/BucketTests.cpp
```diff
@@ -399,6 +399,37 @@ TEST_CASE("merges proceed old-style despite newer shadows",
     }
 }
 
+TEST_CASE("lowest level merge converts live to init", "[bucket][bucketinit]")
```
SirTyson, line 402: I think this test would be better as a BucketList level test instead of a `Bucket` level test. While I'm confident your code works today, there's an assumption here that `keepDeadEntries = false` is always the case and only the case for the bottom level bucket. While this is true today, it's an unstated assumption that could change in the future (as in will change with my state archival stuff), so a higher level test would test that:

1. Your code works (helpful for this PR)
2. This assumption in the BucketList logic is upheld (useful for future PRs)

SirTyson (changes requested):
Overall LGTM. Just one logic cleanup and I think a higher level test would be more helpful here vs. a low level one.

src/bucket/Bucket.h
```diff
@@ -0,0 +1,205 @@
+#pragma once
```
ThomasBrady, line 1: oops bad rebase, I'll delete this

src/bucket/BucketOutputIterator.cpp
```diff
@@ -157,9 +156,31 @@ BucketOutputIterator<BucketT>::put(typename BucketT::EntryT const& e)
         mBuf = std::make_unique<typename BucketT::EntryT>();
     }
 
+    if constexpr (std::is_same_v<BucketT, LiveBucket>)
```
ThomasBrady, line 159: @SirTyson is this correct? Do we want to do this conversion on other BucketT types?

src/bucket/BucketBase.cpp
```diff
@@ -306,7 +306,8 @@ BucketBase::merge(BucketManager& bucketManager, uint32_t maxProtocolVersion,
                   std::shared_ptr<BucketT> const& newBucket,
                   std::vector<std::shared_ptr<BucketT>> const& shadows,
                   bool keepTombstoneEntries, bool countMergeEvents,
-                  asio::io_context& ctx, bool doFsync)
+                  asio::io_context& ctx, bool doFsync,
+                  bool rewriteLiveToInitEntries)
```
SirTyson, line 310: Sorry, I don't think I was clear in my original comment. I don't think we need another bool for `rewriteLiveToInitEntries`. This leads to footguns, as it should always be the case (currently) that when `keepTombstoneEntries == true`, `rewriteLiveToInitEntries == true`. When I was talking about the implicit assumption, I didn't mean to add a flag, as this adds more complexity for future maintainers. Rather, we should keep the interface as is, but add a test that will trip. Sorry about that.

src/bucket/BucketBase.h
```diff
@@ -65,6 +65,9 @@ class BucketBase : public NonMovableOrCopyable
   public:
     static constexpr ProtocolVersion
         FIRST_PROTOCOL_SUPPORTING_PERSISTENT_EVICTION = ProtocolVersion::V_23;
+    static constexpr ProtocolVersion
+        FIRST_PROTOCOL_SUPPORTING_INITENTRY_AND_METAENTRY =
```
SirTyson, line 69: This shouldn't be redefined here. It's only relevant for `LiveBucket`, so we should define it in `LiveBucket` only (which it currently is). It's static so we can use it everywhere via `LiveBucket::FIRST_PROTOCOL_SUPPORTING_INITENTRY_AND_METAENTRY`.

src/bucket/BucketOutputIterator.cpp
SirTyson: Ya, the constexpr check is correct, since this only applies to `LiveBucket` types. We can still simplify this though. Try
```
// Comment about rewrite here
bool shouldRewrite = std::is_same_v<BucketT, LiveBucket> && mRewriteLiveToInitEntries && e.type() == LIVEENTRY &&
            protocolVersionStartsFrom(
                mMeta.ledgerVersion,
                BucketT::FIRST_PROTOCOL_SUPPORTING_INITENTRY_AND_METAENTRY);
if (shouldRewrite)
{
            ++mMergeCounters.mOutputIteratorLiveToInitRewrites;
            auto eCopy = e;
            eCopy.type(INITENTRY);
            *mBuf = eCopy;
}
else
{
            *mBuf = e;
}

++mMergeCounters.mOutputIteratorBufferUpdates;
```

src/bucket/test/BucketListTests.cpp
```diff
@@ -521,6 +522,22 @@ TEST_CASE_VERSIONS("live bucket tombstones expire at bottom level",
         REQUIRE(e0.nDead != 0);
         REQUIRE(e1.nDead != 0);
         REQUIRE(e2.nDead == 0);
+        if (protocolVersionStartsFrom(
```
SirTyson, line 525: This doesn't actually test our fix. The problem is, in old bucketlist code that has long since been fixed, live entries where written to the bottom level. The current code doesn't exhibit this behavior. What you're testing here is if the current BucketList logic works, but not that your conversion code from the buggy BucketList -> good bucketlist actually works. What we should do is populate a BucketList, but specifically populate it with buggy contents such that the bottom level has live entries (see "live bucket tombstones expire at bottom level" for populating a bucketList without `addBatch`). Then, add empty batched to your BucketList until all entries have coallesced in the bottom bucket (see "hot archive bucket tombstones expire at bottom level" for this example). Finally, scan the bottom bucket and make sure they're all init.

SirTyson (changes requested):
Sorry, I think I was a little unclear in my last comment. A little cleanup on the interface, and I think the test case still needs improving.

src/bucket/BucketOutputIterator.cpp
ThomasBrady: I need the block to be conditional on `constexpr (std::is_same_v<BucketT, LiveBucket>)` in order to compile, as other `BucketT::EntryT` have different types, no?

For example, when  `BucketT` is a `HotArchiveBucket`, `e.type() == LIVEENTRY` is a comparison of different enumeration types (`BucketEntryType` to `HotArchiveBucketEntryType`), and `eCopy.type(INITENTRY)` doesn't compile as there's no `HotArchiveBucketEntry::type(BucketEntryType t)`.

SirTyson: Ahh, rip `constexpr`. This is fine thing, but I'd still like to see the typename replaced with auto.

SirTyson (changes requested):
Test looks great! One issue that I didn't notice (until your test case, which was super helpful), we need to gate this feature on protocol 23. Currently, we gate on `FIRST_PROTOCOL_SUPPORTING_INITENTRY_AND_METAENTRY`, which is before p23. This means catchup will fail, as we'll change the bucket hashes for p22 for example. We need to add a new constexpr, like `FIRST_PROTOCOL_CONVERTING_BOTTOM_LEVEL_LIVE_TO_INIT` or something like that which is equal to 23. Sorry I didn't catch this before!

SirTyson (approved):

ThomasBrady:
@SirTyson I had to rebase, can you re-enable automerge?

PR #4500, opened 2024-10-06
Fix tracy

graydon:
This change returns tracy to kinda-working on soroban. It does not work on soroban v21 (you still just get an opaque span for such InvokeHostFunction calls) but does:

  - Build for soroban v22
  - Even when running with `CXX=clang++-12 CXXFLAGS=-stdlib=libc++` as in CI
  - Work for soroban v22 (not-crash and actually produce tracy sub-spans for soroban v22)
  - Roll back the stale cost-adjustments made for SOROBAN_TEST_EXTRA_PROTOCOL=22 mode
  - As a bonus, simplify lifetime/startup and eliminate a layer of locking in the memory-tracking code

There were three major issues to deal with:

  - We weren't passing relevant toolchain variables from the Makefile through to cargo invocations, at all.
  - We weren't _setting_ CXXSTDLIB which is a special one that build.rs needs to build C++ code.
  - We were using tracy in `manual-lifetime` mode because older versions of the rust `tracy-client` crate required it, and this just can't really work when there are multiple copies of `tracy-client` linked into the final executable, they fight over who gets to initialize tracy.

By abandoning those old versions of `tracy-client` and only enabling new ones that allow _turning off_ `manual-lifetime`, we're in a much better state. The multiple copies of `tracy-client` all expect tracy to initialize itself, and it does so, and everything's fine.

I did _not_ yet conditionalize-out the memory-tracking code. I might push a further commit here to do so. It seems like it's fairly expensive to have on by default when using tracy, and I'm not sure anyone benefits from it most of the time.


sisuresh (approved):

graydon:
This was accidentally closed.

PR #4511, opened 2024-10-21
Revert downgrading of log error->warn->error

ThomasBrady:
# Description
Reverts downgrading of error to warning in `CheckSingleLedgerHeaderWork`

https://github.com/stellar/stellar-core/pull/4508#discussion_r1809573062

# Checklist
- [ ] Compiles
- [ ] Ran all tests

### DIFF
--- a/src/historywork/CheckSingleLedgerHeaderWork.cpp
+++ b/src/historywork/CheckSingleLedgerHeaderWork.cpp
@@ -7,6 +7,7 @@
 #include "history/HistoryArchive.h"
 #include "history/HistoryManager.h"
 #include "historywork/GetAndUnzipRemoteFileWork.h"
+#include "main/ErrorMessages.h"
 #include "util/GlobalChecks.h"
 #include "util/Logging.h"
 #include "util/TmpDir.h"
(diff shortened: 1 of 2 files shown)


src/historywork/CheckSingleLedgerHeaderWork.cpp
```diff
@@ -84,9 +84,17 @@ CheckSingleLedgerHeaderWork::doWork()
     }
     else if (mGetLedgerFileWork->getState() != State::WORK_SUCCESS)
     {
-        CLOG_WARNING(History,
-                     "Failed to download ledger checkpoint {} from archive {}",
-                     mFt->baseName_gz(), mArchive->getName());
+        CLOG_ERROR(
+            History,
+            "Failed to download ledger checkpoint {} from archive {}. This may "
```
marta-lokhova, line 89: I think this error should not be transient, since we wait until checkpoint is published (this code is used for health checks, not catchup). Also, could we use the "possibly corrupt archive" error message const? It could be followed by the suggestion to focus on fixing the archives  (but not other configs, like quorum set, etc)

src/main/ErrorMessages.h
```diff
@@ -11,7 +11,12 @@ constexpr auto const REPORT_INTERNAL_BUG =
     "Please report this bug along with this log file if this was not expected";
 constexpr auto const POSSIBLY_CORRUPTED_HISTORY =
     "One or more of history archives may be corrupted. Update HISTORY "
-    "configuration entry to only contain valid ones";
+    "configuration entry to only contain valid ones. If you have a valid "
```
marta-lokhova, line 14: Sorry, I wasn't clear: we can't modify POSSIBLY_CORRUPTED_HISTORY directly, because it's used in other scenarios, where the rest of this message isn't relevant/confusing. Please keep this message as-is, then in `CheckSingleLedgerWork` add a log:
```
If this occurs often, consider notifying the archive owner. As long as your configuration has any valid history archives, this error does NOT mean your node is unhealthy.
```

marta-lokhova (approved):

PR #4528, opened 2024-10-30
Error and exit on validator configs that contain only `LOW` validators

### LINKED ISSUE #4526: Detect and error out on configurations that will cause leader election to fail
If a node is a validator and all validators in its config file are assigned quality level `LOW` (including itself), then the node will spin forever on leader election because no validator is eligible to be the leader. We should add check to the config file validator that checks for this and fails to start with a useful error message if the check fails.

Although there has been some discussion around adding a maximum number of retries to the leader election algorithm, I do not recommend that because the algorithm is probabilistic and should eventually produce a leader. Moreover, it's difficult to decide what an "acceptable" number of retries is and when to give up. Given this, I believe continuing to try is preferable to halting.

bboston7:
Closes #4526

This change raises an error when a node is a validator and all of the validators in its config (including itself) are marked `LOW` quality. This represents a misconfiguration because `LOW` quality validators cannot win leader election. If every validator is `LOW` quality, then there are no eligible nomination leaders and nomination will stall.

# Checklist
- [x] Compiles
- [x] Ran all tests

### DIFF
--- a/src/main/Config.cpp
+++ b/src/main/Config.cpp
@@ -2467,6 +2467,14 @@ Config::setValidatorWeightConfig(std::vector<ValidatorEntry> const& validators)
         homeDomainsByQuality[v.mQuality].insert(v.mHomeDomain);
     }
 
+    if (NODE_IS_VALIDATOR &&
+        highestQuality == ValidatorQuality::VALIDATOR_LOW_QUALITY)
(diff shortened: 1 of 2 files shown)


marta-lokhova (approved):
Looks great, thank you!

PR #4533, opened 2024-11-04
Split core's database into "main" and "misc" to allow concurrent read/writes

### LINKED ISSUE #4319: background ledger close: concurrent database access. Switch to using connection pools, so ledger close and things like publishing and peer management can run concurrently. A few observations here

marta-lokhova:
Resolves #4319 

Currently, the split is only supported for on-disk SQLite database (if we decide to support Postgres long-term, we can add support for it as well)


marta-lokhova:
Closing in favor of https://github.com/stellar/stellar-core/pull/4533

PR #4540, opened 2024-11-12
Update resource calculator with latest soroban resource limits

anupsdf:
### What
Update resource calculator with latest soroban tx write byte limit of 129 KiB


MonsieurNicolas:
you need to rerun the notebook so that the markdown contained in this doc gets updated with the latest results

anupsdf:
> you need to rerun the notebook so that the markdown contained in this doc gets updated with the latest results

Good point! Re-ran and markdown has the latest results now.

MonsieurNicolas (approved):

PR #4544, opened 2024-11-15
De-dupe inbound SCP traffic

marta-lokhova:
The primary goal of this change is to reduce burden on the main thread scheduler by hashing and de-duping inbound SCP traffic. In simulations, we've observed duplicate SCP message consume quite a bit of main thread/scheduler capacity. With this change, the amount of duplicate SCP traffic core processes is reduced by ~50%. Another small change in this PR is to broadcast SCP messages immediately instead of scheduling them asynchronously - SCP traffic is high priority, so there's no reason to delay it like we do in master.
Attaching scheduler load comparison plus duplicate traffic inbound rate for comparison
Screenshot 2024-11-14 at 16 25 25
Screenshot 2024-11-14 at 16 28 54


src/overlay/OverlayManagerImpl.cpp
```diff
@@ -373,17 +375,27 @@ OverlayManagerImpl::getFlowControlBytesConfig() const
                   INITIAL_FLOW_CONTROL_SEND_MORE_BATCH_SIZE_BYTES >=
               maxTxSize))
         {
-            return {static_cast<uint32_t>(maxTxSize) +
-                        INITIAL_FLOW_CONTROL_SEND_MORE_BATCH_SIZE_BYTES,
-                    INITIAL_FLOW_CONTROL_SEND_MORE_BATCH_SIZE_BYTES};
+            static_cast<uint32_t>(maxTxSize) +
```
bboston7, line 378: This result just gets thrown away. Should it be returned? Also, `maxTxSize` is already a `uint32_t`, so the `static_cast` is unnecessary.

src/overlay/Peer.h
```diff
@@ -500,4 +491,17 @@ class Peer : public std::enable_shared_from_this<Peer>,
         }
     }
 };
+
+class MsgCapacityTracker : private NonMovableOrCopyable
```
bboston7, line 495: Can you please add a class-level comment explaining what a `MsgCapacityTracker` is?

src/overlay/OverlayManagerImpl.cpp
marta-lokhova: oh gosh, great catch! Looks like this was accidentally introduced in the last cleanup commit (so after perf evaluation was done). Tests didn't catch it since they always override these configs. I'll add a test case that uses default configs.

src/main/AppConnector.h
```diff
@@ -51,5 +52,6 @@ class AppConnector
     Config const& getConfig() const;
     bool overlayShuttingDown() const;
     OverlayMetrics& getOverlayMetrics();
+    bool isScheduled(std::shared_ptr<MsgCapacityTracker> msgTracker);
```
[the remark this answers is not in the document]
marta-lokhova: Done.

src/overlay/Floodgate.cpp
```diff
@@ -145,21 +145,7 @@ Floodgate::broadcast(std::shared_ptr<StellarMessage const> msg,
...
-                    [msg, weak, log = !broadcasted]() {
-                        auto strong = weak.lock();
-                        if (strong)
-                        {
-                            strong->sendMessage(msg, log);
-                        }
-                    },
-                    fmt::format(FMT_STRING("broadcast to {}"),
-                                peer.second->toString()));
+                peer.second->sendMessage(msg, !broadcasted);
```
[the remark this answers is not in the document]
marta-lokhova: Fixed!

src/overlay/Peer.h
marta-lokhova: Added.

graydon: Given the elevated role of `MsgCapacityTracker` in this PR -- as _the_ way messages get passed around now -- I wonder if it ought to get a more central-sounding name like `CapacityTrackedMessage` or something?

graydon (changes requested):
Just a couple renaming suggestions, looks good!

graydon (dismissed):

graydon (approved):

bboston7 (approved):

PR #4549, opened 2024-11-20
fsync on BucketIndex writes

### LINKED ISSUE #4546: Bucket index files corruption
Observed on a validator, the problem occurred when core shutdown while re-indexing buckets
```
current exception: std::runtime_error("Failed to read 4 bytes from input stream! Read 0")
stellar-core.service: Main process exited, code=killed, status=6
```

SirTyson:
# Description

Resolves #4546

This change calls fsync after writing `BucketIndex` files. This change was slightly more involved than expected. There's no way to fall fsync on files opened via `std::fstream`, so I had to implement a new cereal archive type that accepts our own `asio::buffered_write_streamasio::buffered_write_stream` backed stream type.

# Checklist
- [x] Compiles
- [x] Ran all tests

### DIFF
--- a/src/catchup/IndexBucketsWork.h
+++ b/src/catchup/IndexBucketsWork.h
@@ -4,6 +4,7 @@
 
 #pragma once
 
+#include "work/BasicWork.h"
 #include "work/Work.h"
 #include <memory>
 
@@ -20,7 +21,7 @@ class IndexBucketsWork : public Work
     {
         std::shared_ptr<Bucket> mBucket;
         std::unique_ptr<BucketIndex const> mIndex;
-        bool mDone{false};
+        BasicWork::State mState{BasicWork::State::WORK_WAITING};
 
         void postWork();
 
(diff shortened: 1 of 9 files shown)


src/catchup/IndexBucketsWork.h
```diff
@@ -21,6 +21,7 @@ class IndexBucketsWork : public Work
         std::shared_ptr<Bucket> mBucket;
         std::unique_ptr<BucketIndex const> mIndex;
         bool mDone{false};
+        bool mFailed{false};
```
marta-lokhova, line 24: nit: use one BasicWork::State variable to convey final state instead of mDone and mFailed. Also it needs to be reset in doReset (right now this works because we override the retry policy, but it's a footgun)

marta-lokhova (commented):
thanks! one small comment, otherwise lgtm

marta-lokhova (approved):

PR #4553, opened 2024-11-21
Improves stellar-core-debug-info script and adds docs

### LINKED ISSUE #4545: Document stellar-core-debug-info
Including documentation in the repo as well as stellar-docs

SirTyson:
# Description

Resolves #4545

This PR updates documentation regarding the `stellar-core-debug-info`.

Additionally, while helping people debug nodes, the script was difficult to use and had many default values specific only to SDF infrastructure. I've updated the script to be easier to use. Specifically, it requires an output directory argument, and creates the directory automatically if it does not exist. The script also automatically detects the stellar-core executable path and config via the `stellar-core.service` file. Finally, I've added additional error checking around `offline-info` and better path resolution, which previously was buggy.

# Checklist
- [x] Compiles
- [x] Ran all tests

### DIFF
--- a/scripts/stellar-core-debug-info
+++ b/scripts/stellar-core-debug-info
@@ -15,25 +15,78 @@ import time
 
 def parse_args():
     parser = argparse.ArgumentParser(description='Gathers information about host and stellar-core')
-    parser.add_argument('-d', '--dest', required=False, type=str, help='Pre-existing path to use for scratch space and.'
-                        'storing results. The script will create new subdirectory under this path.',
-                        default='/var/lib/stellar/')
-    parser.add_argument('-c', '--core-config', required=False, type=str, help='Path to the stellar-core config file',
-                        default='/etc/stellar/stellar-core.cfg')
-    parser.add_argument('-l', '--log-dir', required=False, type=str, help='Path where logs are written to.'
-                        'If not set we will try to find it in the config or use /var/log/stellar/ location.'
+    parser.add_argument('outputDir', type=str, help='Path to directory to use for scratch space and '
+                        'storing results. The script will create the directory if it does not exist and a new subdirectory under this path.')
+    parser.add_argument('-c', '--core-config', required=False, type=str, help='Path to the stellar-core config file. '
+                        'If not set we will try to find it in the service file.')
+    parser.add_argument('-l', '--log-dir', required=False, type=str, help='Path where logs are written to. '
+                        'If not set we will try to find it in the config. '
                         'Set to string "disabled" to exclude logs.')
-    parser.add_argument('-b', '--bucket-dir', required=False, type=str, help='Path where buckets are written to.'
-                        'If not set we will try to find it in the config or use /var/lib/stellar/buckets location.'
+    parser.add_argument('-b', '--bucket-dir', required=False, type=str, help='Path where buckets are written to. '
+                        'If not set we will try to find it in the config. '
                         'Set to string "disabled" to exclude buckets directory.')
-    parser.add_argument('-p', '--core-path', required=False, type=str, help='Path to the stellar-core binary'
-                        'If not set "stellar-core" will be used.',
-                        default='stellar-core')
-    parser.add_argument('-s', '--sqlite-path', required=False, type=str, help='Path to the sqlite database.'
-                        'If not set we will try to find it in the config or use /var/lib/stellar/stellar.db location.'
+    parser.add_argument('-p', '--core-path', required=False, type=str, help='Path to the stellar-core binary. '
+                        'If not set "stellar-core" will be used.', default='stellar-core')
+    parser.add_argument('-s', '--sqlite-path', required=False, type=str, help='Path to the sqlite database. '
+                        'If not set we will try to find it in the config. '
                         'Set to string "disabled" to exclude sqlite.')
     return parser.parse_args()
 
+def get_service_exec_start():
+    service_name = "stellar-core.service"
+    try:
+        # Use systemctl to retrieve the service file content
+        result = subprocess.run(
+            ["systemctl", "cat", service_name],
+            stdout=subprocess.PIPE,
+            stderr=subprocess.PIPE,
+            text=True
+        )
+
+        if result.returncode != 0:
+            raise Exception(result.stderr.strip())
+
+        # Parse the service file content
+        exec_start = None
+        for line in result.stdout.splitlines():
+            if line.strip().startswith("ExecStart="):
+                exec_start = line.split("=", 1)[1].strip()
+                break
+
+        if exec_start:
+            return exec_start
+        else:
+            raise ValueError(f"No 'ExecStart' found in {service_name} service file.")
+    except Exception as e:
+        return f"Error: {e}"
+
+def extract_paths(exec_start):
+    try:
+        # Extract the first path (the command)
+        first_path = re.search(r"^([^\s]+)", exec_start).group(1)
+
+        # Extract the config file path after the --conf flag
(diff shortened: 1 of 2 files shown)


scripts/stellar-core-debug-info
```diff
@@ -15,25 +15,78 @@ import time
...
+def get_service_exec_start():
+    service_name = "stellar-core.service"
+    try:
+        # Use systemctl to retrieve the service file content
+        result = subprocess.run(
+            ["systemctl", "cat", service_name],
+            stdout=subprocess.PIPE,
+            stderr=subprocess.PIPE,
+            text=True
+        )
```
jacekn, line 44: :bulb: `subprocess.check_output` might allow you to reduce code needed to handle errors

scripts/stellar-core-debug-info
```diff
@@ -15,25 +15,78 @@ import time
...
+                        'If not set we will try to find it in the config. '
                         'Set to string "disabled" to exclude sqlite.')
     return parser.parse_args()
 
+def get_service_exec_start():
+    service_name = "stellar-core.service"
+    try:
+        # Use systemctl to retrieve the service file content
+        result = subprocess.run(
+            ["systemctl", "cat", service_name],
```
jacekn, line 40: :question: it would be worth testing this inside docker too. We use `/usr/bin/stellar-core` as entrypoint so systemd will not be running inside docker containers.
Perhaps this needs to be handled separately?

scripts/stellar-core-debug-info
SirTyson: Done

scripts/stellar-core-debug-info
SirTyson: Added a function to see if we're running in docker and used the docker default paths if we are.

anupsdf:
Here is the console log from running this on my mac. 
- gathering core info has this vm space error
- Core was not running when I ran this so its strange that offline-info complained but it gave me [this log](https://gist.github.com/anupsdf/47e2b7eb3db7d4678e68151b7e20aee7) where it was having trouble getting db schema version
- gather_os_info is not mac friendly i guess
- i was able to get the buckets directory 
- gathering sqllite db info also complained

```
 ./scripts/stellar-core-debug-info -c ../pubnet_watcher.cfg -p /Users/anuppani/sdf_git/stellar-core/src/stellar-core -s . -b buckets  ../core_logs
Getting get_full_path_for_command /Users/anuppani/sdf_git/stellar-core/src/stellar-core
Gathering OS information...
Error calling function gather_os_info
Gathering stellar-core version and config...
stellar-core(40716,0x1feb14f40) malloc: nano zone abandoned due to inability to reserve vm space.
Warning: running non-release version v22.0.0rc2-85-g4b4cd3657 of stellar-core
Error calling function gather_core_info
Gathering stellar-core offline-info...
Warning: offline-info command failed. Maybe stellar-core is still running? For more information check /Users/anuppani/sdf_git/core_logs/stellar-core-debug-info-2024-11-22-08-18-09/offline-info/output
Gathering logs...
Error calling function gather_logs
Gathering buckets directory
Gathering sqlite DB
Error calling function gather_sqlite_db
Results stored in /Users/anuppani/sdf_git/core_logs/stellar-core-debug-info-2024-11-22-08-18-09.tar.gz
Encountered some errors when gathering data
```

SirTyson:
> Here is the console log from running this on my mac.
> 
> * gathering core info has this vm space error
> * Core was not running when I ran this so its strange that offline-info complained but it gave me [this log](https://gist.github.com/anupsdf/47e2b7eb3db7d4678e68151b7e20aee7) where it was having trouble getting db schema version
> * gather_os_info is not mac friendly i guess
> * i was able to get the buckets directory
> * gathering sqllite db info also complained
> 
> ```
>  ./scripts/stellar-core-debug-info -c ../pubnet_watcher.cfg -p /Users/anuppani/sdf_git/stellar-core/src/stellar-core -s . -b buckets  ../core_logs
> Getting get_full_path_for_command /Users/anuppani/sdf_git/stellar-core/src/stellar-core
> Gathering OS information...
> Error calling function gather_os_info
> Gathering stellar-core version and config...
> stellar-core(40716,0x1feb14f40) malloc: nano zone abandoned due to inability to reserve vm space.
> Warning: running non-release version v22.0.0rc2-85-g4b4cd3657 of stellar-core
> Error calling function gather_core_info
> Gathering stellar-core offline-info...
> Warning: offline-info command failed. Maybe stellar-core is still running? For more information check /Users/anuppani/sdf_git/core_logs/stellar-core-debug-info-2024-11-22-08-18-09/offline-info/output
> Gathering logs...
> Error calling function gather_logs
> Gathering buckets directory
> Gathering sqlite DB
> Error calling function gather_sqlite_db
> Results stored in /Users/anuppani/sdf_git/core_logs/stellar-core-debug-info-2024-11-22-08-18-09.tar.gz
> Encountered some errors when gathering data
> ```

Hmm this looks like an issue with your core build. Can you run `version` command normally without the script? It looks like the malloc error is coming from within stellar-core, not the script.

Wrt os_info, that's only supported on linux. This is fine, as the script is intended for production environments. Finally, I don't think gathering sqllite info worked because you fed the script a bad path via the `-s .` flag, this should be something like `-s ./sql.db`. You should be able to just run ` ./scripts/stellar-core-debug-info -c ../pubnet_watcher.cfg -p /Users/anuppani/sdf_git/stellar-core/src/stellar-core ../core_logs` and the db and buckets path will be automatically pulled from the provided config.

anupsdf:
> 
> Hmm this looks like an issue with your core build. Can you run `version` command normally without the script? It looks like the malloc error is coming from within stellar-core, not the script.
> 
> Wrt os_info, that's only supported on linux. This is fine, as the script is intended for production environments. Finally, I don't think gathering sqllite info worked because you fed the script a bad path via the `-s .` flag, this should be something like `-s ./sql.db`. You should be able to just run ` ./scripts/stellar-core-debug-info -c ../pubnet_watcher.cfg -p /Users/anuppani/sdf_git/stellar-core/src/stellar-core ../core_logs` and the db and buckets path will be automatically pulled from the provided config.

My config file didn't have the db and buckets path, I will add them. For now, `-s ./stellar.db` option worked for SQLite.
My version command [output](https://gist.github.com/anupsdf/641cbc79e1b9e56b13384c6cba234d8a) also throws this malloc error but does print the details afterwards.

SirTyson:
> My config file didn't have the db and buckets path, I will add them. For now, `-s ./stellar.db` option worked for SQLite.
> My version command [output](https://gist.github.com/anupsdf/641cbc79e1b9e56b13384c6cba234d8a) also throws this malloc error but does print the details afterwards.

I think the script is working as intended then. I'm not sure why your core image is throwing and still reporting version info, but it's definitely failing, and the correct behavior script wise is probably to just give up on processing any sort of output if the stellar-core invocation is returning a non zero exit code. In this particular instance it looks like we could still parse output on failure, but I don't think that can be generalized.

Is this build the current master? This might be a mac specific issue, the `offline-info` and `version` commands run fine for me on linux.

anupsdf:
> 
> I think the script is working as intended then. I'm not sure why your core image is throwing and still reporting version info, but it's definitely failing, and the correct behavior script wise is probably to just give up on processing any sort of output if the stellar-core invocation is returning a non zero exit code. In this particular instance it looks like we could still parse output on failure, but I don't think that can be generalized.
> 
> Is this build the current master? This might be a mac specific issue, the `offline-info` and `version` commands run fine for me on linux.

The malloc error was because I had asan enabled. The error went away with core build without `--enable-asan`. 
Yeah, this doesn't seem like a normal processing when encountering error.

anupsdf (approved):
lgtm! Thanks for fixing.

PR #4555, opened 2024-11-22
Harden publishing

marta-lokhova:
This PR contains two changes:
* Properly fsync HAS files that act as a "publish queue". Note that the checkpoints themselves are using XDR streams with fsync turned on already, so this change impacts HAS files only. 
* Gracefully handle situations where nodes enable publish mid-checkpoint. Core will wait until the _next_ checkpoint to begin publishing. Without this change, nodes would need to rebuild state before enabling publishing (although it's unclear how often partial archives occur in practice).

Additionally, I added a few more thing to aid debugging, such as printing checkpoint files with dump-xdr and an offline command to dump all checkpoints scheduled to publish.

### DIFF
--- a/src/history/HistoryManagerImpl.cpp
+++ b/src/history/HistoryManagerImpl.cpp
@@ -80,13 +80,13 @@ HistoryManager::createPublishQueueDir(Config const& cfg)
 std::filesystem::path
 publishQueueFileName(uint32_t seq)
 {
-    return fs::hexStr(seq) + ".json";
+    return fs::hexStr(seq) + ".checkpoint";
 }
 
 std::filesystem::path
 publishQueueTmpFileName(uint32_t seq)
 {
-    return fs::hexStr(seq) + ".json.dirty";
+    return fs::hexStr(seq) + ".checkpoint.dirty";
 }
 
 void
@@ -271,7 +271,7 @@ HistoryManagerImpl::logAndUpdatePublishStatus()
 bool
 isPublishFile(std::string const& name)
 {
-    std::regex re("^[a-z0-9]{8}\\.json$");
+    std::regex re("^[a-z0-9]{8}\\.checkpoint$");
     auto a = regex_match(name, re);
     return a;
 }
@@ -279,7 +279,7 @@ isPublishFile(std::string const& name)
 bool
 isPublishTmpFile(std::string const& name)
 {
-    std::regex re("^[a-z0-9]{8}\\.json.dirty$");
+    std::regex re("^[a-z0-9]{8}\\.checkpoint.dirty$");
     auto a = regex_match(name, re);
     return a;
 }
--- a/src/history/HistoryManagerImpl.cpp
+++ b/src/history/HistoryManagerImpl.cpp
@@ -103,8 +103,16 @@ writeCheckpointFile(Application& app, HistoryArchiveState const& has,
     auto filename = publishQueueFileName(has.currentLedger);
     auto tmpOut = app.getHistoryManager().getTmpDir() / filename;
     {
-        OutputFileStream out(app.getClock().getIOContext(),
-                             /* fsyncOnClose */ true);
+        // Always fsync in prod paths, but allow disabling for tests for
+        // performance
+        OutputFileStream out(
+            app.getClock().getIOContext(),
+#ifdef BUILD_TESTS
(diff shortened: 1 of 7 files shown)


graydon (approved):

graydon (approved):

graydon (dismissed):

src/history/HistoryManagerImpl.cpp
```diff
@@ -12,6 +12,11 @@
 #include "crypto/Hex.h"
 #include "crypto/SHA.h"
 #include "herder/HerderImpl.h"
+#include <cereal/archives/binary.hpp>
+#include <cereal/archives/json.hpp>
```
SirTyson, line 16: Nit: json include not required.

src/history/HistoryManagerImpl.cpp
```diff
@@ -97,7 +103,13 @@ writeCheckpointFile(Application& app, HistoryArchiveState const& has,
         app.getHistoryManager().isLastLedgerInCheckpoint(has.currentLedger));
     auto filename = publishQueueFileName(has.currentLedger);
     auto tmpOut = app.getHistoryManager().getTmpDir() / filename;
-    has.save(tmpOut.string());
+    {
+        OutputFileStream out(app.getClock().getIOContext(),
+                             /* fsyncOnClose */ true);
```
SirTyson, line 108: Use `MODE_DISABLES_FSYNC` flag here instead of hardcoding to true.

marta-lokhova: I purposely didn't want to do that: we need durable writes here regardless as to not break publish.

marta-lokhova: maybe we should deprecate MODE_DISABLES_FSYNC. It's been enabled by default for a while, and is absolutely needed for critical paths like updating ledger state.

src/history/HistoryManagerImpl.cpp
marta-lokhova: Fixed.

src/history/HistoryManagerImpl.cpp
```diff
@@ -97,7 +102,13 @@ writeCheckpointFile(Application& app, HistoryArchiveState const& has,
         app.getHistoryManager().isLastLedgerInCheckpoint(has.currentLedger));
     auto filename = publishQueueFileName(has.currentLedger);
     auto tmpOut = app.getHistoryManager().getTmpDir() / filename;
-    has.save(tmpOut.string());
+    {
+        OutputFileStream out(app.getClock().getIOContext(),
```
SirTyson, line 106: I think this might break `dropSQLBasedPublish`. That's the only function where `finalize == true`. It looks like we've changed from writing a JSON file to a binary file. This is fine for the temporary dirty files, but I think if `finalize == true`, this temp binary file will become the "canonical JSON file", but not have a JSON type. It might also help to clarify more with comments and naming that the temporary internal checkpoint file has a different type than the external JSON file we publish. I think operationally this makes sense, we just need to be careful about it.

src/history/HistoryManagerImpl.cpp
SirTyson: We should keep it for testing, but I agree it shouldn't be exposed in prod paths. During my test refactor, i played around with fsync, and some tests ran literally 100x slower when it was enabled, so we definitely need to keep the config around and pass it around to streams correctly.

src/history/HistoryManagerImpl.cpp
marta-lokhova: this is a good point to double check, but I think we're okay here. "Finalized" checkpoints essentially represent a queue of ledgers to publish (previously we did this with a sql table), and can be in any format. Later down the road we load these files into HAS objects and use them to create actual snapshots that are going to be published (https://github.com/stellar/stellar-core/pull/4555/files#diff-55b0f3e8b5f28727e3fd43b6af8569da025006279be1082bfcf0aeec44d20e08R522).

SirTyson (dismissed):
LGTM after the `MODE_DISABLES_FSYNC` change.

src/history/HistoryManagerImpl.cpp
marta-lokhova: sounds good, fixed (behind BUILD_TESTS)

SirTyson (approved):

Issue #4473, opened 2024-09-19
handle soroban loadgen modes misconfiguration gracefully


marta-lokhova:
Looks like core crashes if a soroban invoke is triggered on the node that didn't do invoke setup before. We should probably gracefully handle such a misconfiguration and propagate a useful error message back to the user instead of crashing

```
2024-09-19T00:34:08.389 GBL23 [Ledger INFO] Got consensus: [seq=32, prev=585e84, txs=0, ops=0, sv: [ SIGNED@bd-0 txH: e622ed, ct: 1726706047, upgrades: [ ] ]]
2024-09-19T00:34:08.392 GBL23 [Ledger INFO] Closed ledger: [seq=32, hash=c9c58d]
2024-09-19T00:34:13.019 GBL23 [Herder INFO] Armed with network upgrades: upgradetime=2024-09-19T00:34:00Z, {
    "configupgradesetkey": {
        "contractID": "CD2SNKASQKTLBUNWMZKZGOUXBYLY7RL5RUTMXTQRIZP7JCPLCX3L6ILS",
        "contentHash": "bf07addb011481ae5874f9eb95e07d738885f8830d2d923d1b16be2c175a4641"
    }
}
2024-09-19T00:34:14.625 GBL23 [Herder INFO] Quorum information for 31 : {"agree":23,"cost":318500,"delayed":0,"disagree":0,"fail_at":6,"hash":"39d898","lag_ms":103,"ledger":31,"missing":0,"phase":"EXTERNALIZE","validated":true}

2024-09-19T00:34:14.627 GBL23 [Ledger INFO] Got consensus: [seq=33, prev=c9c58d, txs=0, ops=0, sv: [ SIGNED@wx-2 txH: efb6eb, ct: 1726706053, upgrades: [ ] ]]
2024-09-19T00:34:14.629 GBL23 [Ledger INFO] Closed ledger: [seq=33, hash=466edb]
2024-09-19T00:34:15.538 GBL23 [default INFO] Performing maintenance
2024-09-19T00:34:15.538 GBL23 [History INFO] Trimming history <= ledger 0 (rmin=4294967295, qmin=33, lmin=0)
2024-09-19T00:34:19.875 GBL23 [Herder INFO] Quorum information for 32 : {"agree":23,"cost":210620,"delayed":0,"disagree":0,"fail_at":6,"hash":"39d898","lag_ms":103,"ledger":32,"missing":0,"phase":"EXTERNALIZE","validated":true}

[...]

PR #4559, opened 2024-11-25
Handle soroban loadgen mode misconfig gracefully

ThomasBrady:
# Description

Resolves https://github.com/stellar/stellar-core/issues/4473

# Checklist
- [ ] Compiles
- [ ] Ran all tests

### DIFF
--- a/src/simulation/LoadGenerator.cpp
+++ b/src/simulation/LoadGenerator.cpp
@@ -348,8 +348,17 @@ LoadGenerator::start(GeneratedLoadConfig& cfg)
         if (cfg.modeInvokes())
         {
             auto const& sorobanLoadCfg = cfg.getSorobanConfig();
+            if (!mCodeKey)
+            {
+                // start is incomplete, so reset to avoid leaving the
+                // LoadGenerator in an invalid state.
+                reset();
+                throw std::runtime_error(
+                    "Before running MODE::SOROBAN_INVOKE, please run "
+                    "MODE::SOROBAN_INVOKE_SETUP to set up your contract "
+                    "first.");
+            }
             releaseAssert(mContractInstances.empty());
-            releaseAssert(mCodeKey);
             releaseAssert(mAccountsAvailable.size() >= cfg.nAccounts);
             releaseAssert(mContractInstanceKeys.size() >=
                           sorobanLoadCfg.nInstances);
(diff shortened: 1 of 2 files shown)


src/simulation/LoadGenerator.cpp
```diff
@@ -348,6 +348,13 @@ LoadGenerator::start(GeneratedLoadConfig& cfg)
         if (cfg.modeInvokes())
         {
             auto const& sorobanLoadCfg = cfg.getSorobanConfig();
+            if (!mCodeKey)
```
SirTyson, line 351: Nit: Remove `mCodeKey` assert below.

SirTyson (dismissed):
LGTM, super minor cleanup

src/simulation/test/LoadGeneratorTests.cpp
```diff
@@ -125,6 +125,36 @@ TEST_CASE("generate load with unique accounts", "[loadgen]")
             10 * Herder::EXP_LEDGER_TIMESPAN_SECONDS, false);
     }
 }
+TEST_CASE("handle soroban loadgen mode misconfiguration gracefully",
```
SirTyson, line 128: Nit: This doesn't have to be it's own test, but could just be a `SECTION` in "generate soroban load"

SirTyson (approved):
LGTM! Helpful change :)

SirTyson (approved):

ThomasBrady:
> Unless I'm missing where we catch the exception, this change doesn't handle the failure gracefully, does it? The node still crashes. I think loadgen should gracefully fail and the node should continue operating normally (otherwise ssc will be stuck in pod restart loop without displaying any useful info)

Sorry you are right, this isn't "graceful", however I see there being two problems with this flow: the silent/misleading failure (`"Aborted (core dumped)"`) and gracefully propagating/handling the failure. This PR addressed the first part: if someone invokes without running setup they will now get an exception with the relevant error message. An exception is not exactly graceful however. I'm not sure how we can gracefully recover from this while also notifying the user.  `LoadGenerator::generateLoad` returns `void`; we could print an error and reset the load generator, but if the user is depending on this invocation to succeed, their test will fail regardless. Wouldn't it be better to raise an exception closer to the problematic action?

marta-lokhova:
Load generator sets loadgen.run.failed on failure, which is used across our test suite and in supercluster (e.g. supercluster observed the metrics and fails the mission properly on failure)

PR #4560, opened 2024-11-26
Visual C++ - don't pass the release flag to cargo for debug builds

MonsieurNicolas:
This makes it consistent with the way C++ code is build.

Makes debug builds faster to build and easier to use in with debuggers.


anupsdf (approved):
Makes sense to remove the `release` profile from the Debug configurations and only use them for release configurations.

PR #4504, opened 2024-10-10
Remove SQL support for Ledger State

SirTyson:
# Description

Resolves #4501, #4502, and #4503

Rebased on https://github.com/stellar/stellar-core/pull/4403. This PR

1. Removes the flag for background eviction scan, making it mandatory
2. Removes the flag for SQL ledger state, making BucketListDB mandatory
3. Removes the `--in-memory` command line option.

Because of ongoing State Archival work that requires these changes, I've opted for a single branch to avoid rebasing complexity. That being said, I can open more granular PRs to assist in the reviewing process. In particular, 1 can be seperated into it's own PR. 

Unfortunately, in order to make tests work without SQL ledger state tables, I've cannibalized `--in-memory` mode to be a testing only mode. This reuses most of the `--in-memory` code, but modifies it such that a node would not be able to sync with the network in this mode. For this reason, I think we need to deprecate SQL and `--in-memory` mode at the same time.

Finally, all  tests pass except for "herder externalizes values". While I changed the TestDB mode for that test (as the old SQL mode is no longer supported), no other changes should have been made. I haven't been able to figure out why it's failing, but perhaps someone more familiar with Herder like @marta-lokhova could take a look.

# Checklist
- [x] Reviewed the [contributing](https://github.com/stellar/stellar-core/blob/master/CONTRIBUTING.md#submitting-changes) document
- [x] Rebased on top of master (no merge commits)
- [x] Ran `clang-format` v8.0.0 (via `make format` or the Visual Studio extension)
- [x] Compiles
- [ ] Ran all tests
- [ ] If change impacts performance, include supporting evidence per the [performance document](https://github.com/stellar/stellar-core/blob/master/performance-eval/performance-eval.md)

Issue #4012, opened 2023-11-03
[test] herder handles soroban traffic correctly


marta-lokhova:
- [ ] while in v19: receive Soroban tx, receive v20 generalized tx set with soroban transactions
- [ ] while in v20: receive future tx set with network config upgrade (currently such a test is included in https://github.com/stellar/stellar-core/pull/3977), receive future tx set with different average BL size
- [ ] receiving old Soroban data (for slots already externalized) is a no-op

PR #4562, opened 2024-11-27
In memory testing

### LINKED ISSUE #4501: Deprecate `--in-memory` mode
Upcoming changes in protocol 23 are not compatible with `--in-memory` mode, and it has been deprecated for some time. We should remove support.

### LINKED ISSUE #4502: Deprecate SQL for Ledger state
SQL support for ledger state should be removed, as it has been deprecated and is not compatible with upcoming changes in protocol 23.

SirTyson:
# Description

Resolves #4501 and partially resolves #4502

This PR removes `--in-memory`. Specifically, core will not crash if this flag is passed in for Horizon compatibility reasons, but will ignore it and related flags.

Additionally, `--in-memory` has been cannibalized to use in testing. Previously, any test that committed changes directly to the root ltx was incompatible with BucketListDB and required SQL. This is an issue, as we are deprecating SQL for ledger state, and it would not make sense to maintain this SQL overhead for testing only. While I've removed most of the `in-memory` code, I've saved the bits regarding the never committing ledger root to use for the direct ltx commit test cases. Thus we remove the test dependency for SQL.

In order to properly test the orderbook, the `in-memory` testing mode still maintains and queries a SQL database just for offers (similar to what BucketListDB does).

Originally I did one PR for removing SQL, `--in-memory` and mandating background eviction. This was way to big to merge but may be helpful in providing context on what changes are in flight after this PR: https://github.com/stellar/stellar-core/pull/4504.

Currently everything works except for the "herder externalizes values" test. With BucketListDB testing mode, nodes can be restarted, but direct ltx commits are not supported. With the in-memory mode testing, ltx can be committed to directly, but nodes can not be restarted.

~~For all tests except for "herder externalizes values", this if fine. However, "herder externalizes values" is the only test that both directly commits to the ltx and restarts nodes. In particular, the `setupUpgradeAtNextLedger` function directly commits to the ltx. I've seperated the test into two helper functions, one that doesn't require restart and one that does. I _think_ that the restart test doesn't actually require an upgrade to occur. I think the test just used upgrades in order to populate ledgers with some meaningful, detectable work to make sure externalized ledgers aren't applied several times. If the upgrade operation is not integral to this test, we can probably find a workaround. Alternatively, we can maybe go through the full upgrade arming process with BucketListDB support instead of cutting corners with direct ltx commits. I'm not well versed in herder tests though, so I've left the failing test as is. If we can remove the upgrade and check the condition some other way, this is the path of least resistance.~~

Test has been fixed to use upgrades compatible with BucketListDB.

# Checklist
- [x] Compiles
- [x] Ran all tests


dmkozh:
> I think the test just used upgrades in order to populate ledgers with some meaningful,

I don't think that's the reason, the upgrade is there to make sure that ledgers with upgrades can be externalized out of order (that's important due to how config changes impact the internal core state). https://github.com/stellar/stellar-core/issues/4012 is the issue for which this has been introduced. So I don't think we should remove this.

> Alternatively, we can maybe go through the full upgrade arming process with BucketListDB support instead of cutting corners with direct ltx commits.

Yeah, I think that's the right course of the action.

bboston7 (commented):
I agree with @dmkozh that the upgrade-specific logic seems important to test here.

> Alternatively, we can maybe go through the full upgrade arming process with BucketListDB support instead of cutting corners with direct ltx commits.

I think this is a good idea even independent of the details of this specific test. When possible, avoiding the corner-cutting should lead to more realistic and better overall test IMO.

src/main/CommandLine.cpp
```diff
@@ -419,23 +371,21 @@ clara::Opt
 inMemoryParser(bool& inMemory)
 {
     return clara::Opt{inMemory}["--in-memory"](
-        "(DEPRECATED) store working ledger in memory rather than database");
+        "(DEPRECATED) flag is ignored and will be removed soon.");
```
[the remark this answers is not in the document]
SirTyson: This was specifically a request from platform. there are some Horizon configs that use this flag that are niche and no longer supported, but still exist in the codebase. They wanted us to just ignore the flags for now and not crash so that we don't have a breaking point release.

src/main/CommandLine.cpp
```diff
@@ -1535,10 +1476,10 @@ run(CommandLineArgs const& args)
     CommandLine::ConfigOption configOption;
     auto disableBucketGC = false;
     std::string stream;
-    bool inMemory = false;
     bool waitForConsensus = false;
-    uint32_t startAtLedger = 0;
-    std::string startAtHash;
+    [[maybe_unused]] bool inMemory = false;
+    [[maybe_unused]] uint32_t startAtLedger = 0;
+    [[maybe_unused]] std::string startAtHash;
```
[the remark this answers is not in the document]
SirTyson: Ya, I want clara to be able to parse the flags so we don't throw and crash, but I don't want to actually use the variables. It takes a reference and idk what clara is doing under the hood, so I used separate variables like the library intended to be on the safe side.

src/main/CommandLine.cpp
```diff
@@ -1221,25 +1168,19 @@ int
 runNewDB(CommandLineArgs const& args)
 {
     CommandLine::ConfigOption configOption;
-    bool minimalForInMemoryMode = false;
+    [[maybe_unused]] bool minimalForInMemoryMode = false;
```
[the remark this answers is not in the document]
SirTyson: See below

src/ledger/InMemoryLedgerTxn.h
```diff
@@ -44,6 +44,12 @@ class InMemoryLedgerTxn : public LedgerTxn
     Database& mDb;
     std::unique_ptr<soci::transaction> mTransaction;
 
+    // For some tests, we need to bypass ledger close and commit directly to the
+    // in-memory ltx. However, we still want to test SQL backed offers. The
+    // "never" committing root sets this flag to true such that offer-related
+    // calls get based to the real SQL backed root
+    AbstractLedgerTxnParent* const mRealRootForOffers;
```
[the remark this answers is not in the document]
SirTyson: `InMemoryLedgerTxn` classes now behind `BUILD_TESTS`. The raw pointer was left over from the initial prod implementation, but has been refactored to a reference.

SirTyson:
Got the test to work with BucketList friendly upgrades and pushed some cleanups around `BUILD_TESTS` flags, should be good to go now.

src/main/CommandLine.cpp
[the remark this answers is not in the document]
bboston7: Ah, that makes sense. Thanks

bboston7 (approved):
Thanks for the fixes! I don't have permissions to mark comment threads as "resolved" in this repo, but you can consider all of my comments resolved.

src/herder/test/HerderTests.cpp
```diff
@@ -3801,22 +3752,28 @@ herderExternalizesValuesWithProtocol(uint32_t version)
     REQUIRE(currentALedger() >= currentLedger);
     REQUIRE(currentCLedger() == currentLedger);
 
+    // Arm the upgrade, but don't close the upgrade ledger yet
```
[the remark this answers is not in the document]
SirTyson: Yeah, I should clarify that this doesn't close the upgrade ledger, but does close ledger in order to set up the upgrade.

That being said, I think this test does the same thing the previous test did? In particular, the `setupUpgradeAtNextLedger` function previously commited the upgrade directly to the DB, then called `getHerder().setUpgrades()` on the commited upgrade set. The test then disconnected C and closed a few ledgers., then reconnected C.

This does the same thing, except it uses load gen to generate the `ConfigUpgradeSet` instead of directly comiting. This is fine though, as we set the upgrade, but don't close the upgrade ledger until after we disconnect C.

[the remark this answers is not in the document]
dmkozh: > This is fine though, as we set the upgrade, but don't close the upgrade ledger until after we disconnect C.

But that's not what `upgradeSorobanNetworkConfig` does, is it? `upgradeSorobanNetworkConfig` waits for the upgrade to be applied after arming it.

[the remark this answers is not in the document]
SirTyson: No, I added the `applyUpgrade` flag to determine if `upgradeSorobanNetworkConfig` closes ledgers after arming the upgrade. In this test it's set to false so `upgradeSorobanNetworkConfig` and arms the upgrade, then returns before closing any additional ledgers.

src/ledger/test/InMemoryLedgerTxn.h
```diff
@@ -44,6 +44,12 @@ class InMemoryLedgerTxn : public LedgerTxn
     Database& mDb;
     std::unique_ptr<soci::transaction> mTransaction;
 
+    // For some tests, we need to bypass ledger close and commit directly to the
+    // in-memory ltx. However, we still want to test SQL backed offers. The
+    // "never" committing root sets this flag to true such that offer-related
```
graydon, line 49: I don't know what "this flag" means in the context of this comment -- the config flag saying it's in-memory mode?

src/ledger/test/InMemoryLedgerTxn.h
```diff
@@ -5,8 +5,8 @@
 // of this distribution or at http://www.apache.org/licenses/LICENSE-2.0
 
 #include "database/Database.h"
-#include "ledger/InMemoryLedgerTxnRoot.h"
 #include "ledger/LedgerTxn.h"
+#include "ledger/test/InMemoryLedgerTxnRoot.h"
 
 // This is a (very small) extension of LedgerTxn to help implement in-memory
```
graydon, line 11: I think this comment could do with an update that clarifies:

  - In-memory mode was _originally_ for production uses but is now _just_ a helper mechanism for old tests that committed directly to the SQL database (bypassing the BL), that emulates the existence of a full SQL database, and that if we had the energy to change all those tests we could theoretically remove it. This is a potential future project! Maybe even file a (low priority for now) bug about it.
  - It is not entirely in-memory, but also writes-through to the offers table in SQL. This is surprising and contradicts the existing comment.

graydon: Also I think you can ifdef-guard the whole thing on BUILD_TESTS?

src/main/ApplicationImpl.h
```diff
@@ -194,8 +192,10 @@ class ApplicationImpl : public Application
     // is held in the never-committing LedgerTxn in its entirety -- so if it
```
graydon, line 192: This comment could also use an update emphasizing that this exists only to support old tests that commit directly to the LTX and bypass the BL, and should be removed in future sometime (and linking to bug).

src/main/Config.h
```diff
@@ -705,6 +692,11 @@ class Config : public std::enable_shared_from_this<Config>
     // doing a graceful shutdown
     bool TEST_CASES_ENABLED;
 
+    // A config parameter that uses a never-committing ledger. This means that
```
graydon, line 695: Add note: that also writes through to the offers table in SQL

src/herder/test/HerderTests.cpp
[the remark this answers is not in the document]
dmkozh: Ah, sorry, I've been looking at the current implementation.

dmkozh (dismissed):

src/main/test/ExternalQueueTests.cpp
```diff
@@ -1,44 +0,0 @@
...
-#include "simulation/Simulation.h"
-#include "test/TestUtils.h"
-#include "test/test.h"
-
-using namespace stellar;
-
-TEST_CASE("cursors", "[externalqueue]")
-{
-    VirtualClock clock;
-    Config const& cfg = getTestConfig(0, Config::TESTDB_ON_DISK_SQLITE);
```
graydon, line 19: I _think_ you can keep this pointed at PG? though I don't especially care, I would dearly love to get rid of cursors.

src/main/test/ApplicationUtilsTests.cpp
```diff
@@ -488,146 +375,14 @@ TEST_CASE("offline self-check works", "[applicationutils][selfcheck]")
...
-        damage.damageVictim();
-        REQUIRE(selfCheck(chkConfig) == 1);
-    }
 }
 
 TEST_CASE("application setup", "[applicationutils]")
 {
     VirtualClock clock;
-
-    SECTION("SQL DB mode")
```
graydon, line 503: you could also keep this on PG, right? (not that I'm insisting, just trying to understand)

src/ledger/LedgerManagerImpl.cpp
```diff
@@ -348,41 +348,35 @@ LedgerManagerImpl::loadLastKnownLedger(bool restoreBucketlist,
 
     releaseAssert(latestLedgerHeader.has_value());
 
-    // Step 3. Restore BucketList if we're doing a full core startup
-    // (startServices=true), OR when using BucketListDB
-    if (restoreBucketlist || mApp.getConfig().isUsingBucketListDB())
```
graydon, line 353: I'm not 100% sure this guard can be dropped -- is it always true?

graydon (changes requested):
In general this looks good -- it's hard to understand the motive exactly and the exact nature of the repurposing that's going on here (had to chat with you) so I think some additional comments would be good, but I think that's about all. A couple other nits I was curious about noted along the way. Happy to try to get it landed quickly so you don't need to keep rebasing.

src/herder/test/HerderTests.cpp
```diff
@@ -3720,14 +3679,6 @@ herderExternalizesValuesWithProtocol(uint32_t version)
             Herder::State::HERDER_BOOTING_STATE);
 
     simulation->startAllNodes();
-    if (protocolVersionStartsFrom(version, SOROBAN_PROTOCOL_VERSION))
```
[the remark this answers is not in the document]
SirTyson: The loadgen upgrade helper function is a little hacky and assumes it will only be called once. While it's probably a good idea to fix eventually, this PR was big enough as it was. Everything before this point is set up anyway, so there was no reason (as far as I could tell) why we couldn't combine the upgrades into a single upgrade to play better with our test helpers. It doesn't need to be protocol gated anymore since this is herder tests and we'll never externalize a pre p20 ledger again.

src/herder/test/HerderTests.cpp
```diff
@@ -3801,22 +3752,28 @@ herderExternalizesValuesWithProtocol(uint32_t version)
...
     // Advance A and B a bit further, and collect externalize messages
     std::map<uint32_t, std::pair<SCPEnvelope, StellarMessage>>
         validatorSCPMessagesA;
     std::map<uint32_t, std::pair<SCPEnvelope, StellarMessage>>
         validatorSCPMessagesB;
 
-    for (auto& node : {A, B, getC()})
-    {
-        // C won't upgrade until it's on the right LCL
-        setupUpgradeAtNextLedger(*node);
```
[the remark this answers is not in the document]
SirTyson: Dima had the same question, so refer back to his last comment where we had a back and forth. TLDR: This test upgraded via direct commit to ltx root, set the upgrade, then disconnected node C. This breaks BucketListDB, so I modified the `upgradeSorobanNetworkConfig` to optionally set the upgrade but not actually apply the upgrade ledger. This should replicate the test correctly, but instead of direct ltx commit, we use loadgen to properly build the upgrade.

src/invariant/test/BucketListIsConsistentWithDatabaseTests.cpp
```diff
@@ -43,44 +44,10 @@ struct BucketListGenerator
   public:
     BucketListGenerator() : mLedgerSeq(1)
     {
-        auto cfg = getTestConfig(0, Config::TESTDB_IN_MEMORY_OFFERS);
+        auto cfg = getTestConfig();
         cfg.OVERRIDE_EVICTION_PARAMS_FOR_TESTING = true;
         cfg.TESTING_STARTING_EVICTION_SCAN_LEVEL = 1;
         mAppGenerate = createTestApplication(mClock, cfg);
-
-        auto skey = SecretKey::fromSeed(mAppGenerate->getNetworkID());
```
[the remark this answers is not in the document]
SirTyson: Correct, since the invariant now only applies to offer types.

src/ledger/test/InMemoryLedgerTxn.h
SirTyson: I don't think an ifdef is required given that I moved all the in memory mode files to the test directory and guard non-test file references to the in-memory mode with ifdefs.

src/herder/test/HerderTests.cpp
```diff
@@ -3258,7 +3253,7 @@ TEST_CASE("accept soroban txs after network upgrade", "[soroban][herder]")
 
     auto simulation =
         Topologies::core(4, 1, Simulation::OVER_LOOPBACK, networkID, [](int i) {
-            auto cfg = getTestConfig(i, Config::TESTDB_ON_DISK_SQLITE);
+            auto cfg = getTestConfig(i, Config::TESTDB_IN_MEMORY);
```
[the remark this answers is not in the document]
SirTyson: The test arms a Soroban upgrade with a protocol 19 ledger, which makes the loadgen upgrade helper throw. Looking at this though, I think we can remove this test entirely. It's testing herder receiving soroban TXs post p19 upgrade. We don't have to test the upgrade since this is herder an we'll never see that behavior again. This is a super basic "do soroban txs make it into the block" test, and we have coverage elsewhere, so I'll just delete it.

src/herder/test/HerderTests.cpp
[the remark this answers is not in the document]
SirTyson: Hmm it was actually pretty quick to fix the upgrade function so I'll return back to the original test with two different upgrades.

src/ledger/test/InMemoryLedgerTxn.h
SirTyson: Ahh, this comment is stale from an earlier version I had where we had a flag and only conditionally wrote to SQL, I'll update it.

src/main/test/ApplicationUtilsTests.cpp
SirTyson: This test is currently pretty meaningless for PG, given that the bulk of this test is just a call to `BucketListIsConsistentWithDatabase`, and that invariant now only checks for offers because it assumes BucketListDB is enabled. This won't break anything, but you're not gonna get better coverage than what we already have.

src/ledger/LedgerManagerImpl.cpp
SirTyson: Ah yeah this is a little premature since this doesn't entirely drop SQL.

graydon (approved):
LGTM, thanks!

src/main/ApplicationImpl.cpp
```diff
@@ -190,105 +190,112 @@ maybeRebuildLedger(Application& app, bool applyBuckets)
...
+    }
+    else
+#endif
+    {
+        for (auto let : xdr::xdr_traits<LedgerEntryType>::enum_values())
         {
-            toDrop.emplace(t);
-            continue;
-        }
+            // If BucketListDB is enabled, drop all tables except for offers
```
marta-lokhova, line 205: I know it was here before, but shouldn't it be the other way around? Only rebuild offers, because those are in SQL? the rebuild functionality was added to safely deal with SQL schema upgrades, so only offers are impacted

src/main/ApplicationUtils.cpp
```diff
@@ -195,13 +125,22 @@ setupApp(Config& cfg, VirtualClock& clock, uint32_t startAtLedger,
...
+        /* isLedgerStateReady */ !cfg.MODE_USES_IN_MEMORY_LEDGER
+#else
+        /* isLedgerStateReady */ true
+#endif
+    );
     auto lcl = app->getLedgerManager().getLastClosedLedgerHeader();
 
-    if (cfg.isInMemoryMode() &&
+    if (
+#ifdef BUILD_TESTS
```
marta-lokhova, line 141: hm, this looks suspicious - in prod builds you don't want to be rebuilding state under any circumstances. setupApp shouldn't be called anywhere outside of tests, so probably needs to be behind BUILD_TESTS

src/main/test/ApplicationUtilsTests.cpp
```diff
@@ -488,146 +375,14 @@ TEST_CASE("offline self-check works", "[applicationutils][selfcheck]")
...
     }
-    {
-        // Damage the SQL ledger.
-        TemporarySQLiteDBDamager damage(chkConfig);
-        damage.damageVictim();
-        REQUIRE(selfCheck(chkConfig) == 1);
-    }
 }
 
 TEST_CASE("application setup", "[applicationutils]")
```
marta-lokhova, line 380: this test can probably go away completely

src/main/ApplicationImpl.cpp
```diff
@@ -190,105 +190,112 @@ maybeRebuildLedger(Application& app, bool applyBuckets)
     std::set<LedgerEntryType> toRebuild;
     auto& ps = app.getPersistentState();
     auto bucketListDBEnabled = app.getConfig().isUsingBucketListDB();
-    for (auto let : xdr::xdr_traits<LedgerEntryType>::enum_values())
+
+#ifdef BUILD_TESTS
+    if (app.getConfig().MODE_USES_IN_MEMORY_LEDGER)
```
marta-lokhova, line 195: I'm a bit confused by this, in-memory mode shouldn't have anything to do with the rebuild logic here. It was added just to deal with buggy schema upgrades: https://github.com/stellar/stellar-core/pull/3031, so this shouldn't impact testing?

src/herder/test/HerderTests.cpp
```diff
@@ -2564,11 +2564,6 @@ TEST_CASE("SCP State", "[herder]")
         };
 
     auto doTest = [&](bool forceSCP) {
-        SECTION("sqlite")
```
marta-lokhova, line 2567: Why did you remove this section? It deals with persisting SCP stuff, not ledger state. We still support SQLite for misc data like consensus

src/main/ApplicationUtils.cpp
```diff
@@ -211,67 +150,6 @@ setupApp(Config& cfg, VirtualClock& clock, uint32_t startAtLedger,
         }
     }
 
-    bool doCatchupForInMemoryMode =
```
marta-lokhova, line 214: was looking forward to seeing all of this go away!

src/main/Config.cpp
```diff
@@ -157,7 +156,6 @@ Config::Config() : NODE_SEED(SecretKey::random())
     MANUAL_CLOSE = false;
     CATCHUP_COMPLETE = false;
     CATCHUP_RECENT = 0;
-    EXPERIMENTAL_PRECAUTION_DELAY_META = false;
```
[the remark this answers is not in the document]
SirTyson: I'm pretty sure this was only supported with `in-memory` mode. Master has the current check in `ApplciationImpl.cpp`:

```c++
    if (mConfig.METADATA_OUTPUT_STREAM != "" &&
        mConfig.EXPERIMENTAL_PRECAUTION_DELAY_META && !mConfig.isInMemoryMode())
    {
        throw std::invalid_argument(
            "Using a METADATA_OUTPUT_STREAM with "
            "EXPERIMENTAL_PRECAUTION_DELAY_META set to true "
            "requires --in-memory");
    }
```

`METADATA_OUTPUT_STREAM` has to be nonempty to enable any meta right?

src/main/ApplicationImpl.cpp
SirTyson: If a recall correctly this is a hacky workaround given that enabling BucketListDB is a psuedo schema upgrade, but because it's optional and based on a config flag, couldn't be a proper schema upgrade. I'd like to leave this as is here because it works, and the follow up SQL removal PR addresses this.

src/main/Config.cpp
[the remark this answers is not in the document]
marta-lokhova: it was supported for in-memory mode when the in-memory mode was the only option :) when everyone switched to on-disk, looks like we didn't update it. The flag itself doesn't care about the DB backend used (it's about when to emit meta). But given that it's not supported for on-disk deployments, _and_ nobody really ever complained about needing it, I agree we can drop it.

src/main/ApplicationImpl.cpp
marta-lokhova: well, the whole thing would still work if you removed the rebuild logic completely. Right now the logic here doesn't quite make sense: we're rebuilding everything except offers (but there's nothing to rebuild but offers because all other tables are dropped already in favor of bucketlistDB). The point of the rebuild is to guard against buggy schema upgrades (something we got bitten by in the past which super hard to recover from). I agree this is outside of the scope of this PR, but could we open an issue to fix this?

src/main/ApplicationUtils.cpp
SirTyson: Good point, I'll harden this and remove the ledgerStateReady flag.

src/main/ApplicationImpl.cpp
SirTyson: This is necessary, as the new version of "in memory mode" actually does persist and commit to a SQL DB for the offers table. Perhaps `MODE_USES_IN_MEMORY_LEDGER` is a poor named flag now, but with new in memory mode keeps all state in memory like the original prod version did in addition to writing to a SQL offers table. Reason being there was no reason we couldn't forward offer related writes and queries to the actual SQL table and it gives us better coverage.

SirTyson: In particular there were some test cases failing without this. I don't remember exactly which ones, but I think it was in the upgrades path.

src/herder/test/HerderTests.cpp
SirTyson: `TESTDB_BUCKET_DB_PERSISTENT` is the best, most accurate to prod testing mode we have. It uses BucketListDB for ledger state queries, but still maintains all the same sqlite tables that prod does, so this test mode is still hitting all the misc sqlite tables. This in conjunction with the postgres section is complete coverage.

marta-lokhova: Oh I see, and `postgres` is left to test scp persistence on postgres, even though it's technically a deprecated backend for ledger state.

src/main/test/ApplicationUtilsTests.cpp
marta-lokhova: bump

src/main/ApplicationImpl.cpp
marta-lokhova: wdyt?

SirTyson: This is going away in the next PR, which I'll open right after this gets merged :)

src/main/ApplicationUtils.cpp
```diff
@@ -195,13 +125,13 @@ setupApp(Config& cfg, VirtualClock& clock, uint32_t startAtLedger,
         return nullptr;
     }
 
-    // With in-memory mode, ledger state is not yet ready during this setup step
-    app->getLedgerManager().loadLastKnownLedger(
-        /* restoreBucketlist */ false,
-        /* isLedgerStateReady */ !cfg.isInMemoryMode());
+    app->getLedgerManager().loadLastKnownLedger(/* restoreBucketlist */ false);
```
marta-lokhova, line 128: shouldn't this go away now? IIRC the only reason we needed to load state here was to support in-memory mode (we call this function later in `start` anyway)

src/main/ApplicationUtils.cpp
marta-lokhova: sorry this is still not quite right: in prod paths we shouldn't be rebuilding state, ever. Right now you'd rebuild state at genesis, which I'm not sure would even work (I guess it does given that tests pass). I think this whole if statement should be behind BUILD_TESTS

src/main/ApplicationImpl.cpp
marta-lokhova: sounds good, thanks! just be careful and remember to still rebuild offers given that we use sql for them still.

src/main/ApplicationImpl.cpp
marta-lokhova: Sorry, I still find it confusing - if you persist offers, why does not rebuilding them break the state? Rebuilding should end up in the same state that you started with (I guess I'm just surprised that this code, which was supposed to be functionally a no-op, breaks something)

SirTyson: `maybeRebuildLedger`  doesn't actually "rebuild" state as in populating SQL tables. All it does is drop the tables, and if rebuild==true, recreates an empty version of the table. This is the only function used for creating ledger based SQL table, so if we don't support in-memory mode here, in-memory mode never actually creates the offers table.

marta-lokhova (approved):

PR #4565, opened 2024-12-06
BucketList cache

### LINKED ISSUE #3696: Add persistent cache to BucketListDB

SirTyson:
# Description

Resolves #3696

This adds a persistent cache to BucketListDB at the bucket level. If a bucket is small, the entire bucket contents are cached in memory. For larger buckets, a random eviction cache is used instead.

Currently in draft form for max TPS tests, as I need to clean up the BucketIndex classes given that there's a fair bit of dead code now that caches are used.

# Checklist
- [x] Compiles
- [x] Ran all tests


SirTyson:
Closing prototype, #4630 has a prod ready verison.

PR #4574, opened 2024-12-13
Extend nomination monte carlo simulations to analyze timeouts

### LINKED ISSUE #4561: Extend nomination Monte Carlo simulations to cover asymmetric quorums
Stellar-core contains a suite of tests that check for fair win probabilities and low timeouts in the new application specific nomination algorithm. These tests use Monte Carlo simulations to generate large numbers of random quorums and simulate nomination for thousands of ledgers. However, these tests assume that quorum sets are symmetric across validators. We should add another series of Monte Carlo simulations that test the following asymmetric cases:

1. One set of validators has some quorum set `A` and another has some quorum set `B`. `A` and `B` are almost identical, but `B` has replaced 1 org in `A` (or a small number of orgs in `A`) with a different org.
2. One set of validators has some quorum set `A` and another has some quorum set `B`. `A` is a strict subset of `B`, where `B` contains a few more orgs than `A`. The number of additional orgs should be small (say, 3) so that the two quorum sets largely overlap.

The tests should focus on timeouts, ensuring that they are infrequent, and that rounds with multiple timeouts are very rare. This is the expected outcome from the new nomination algorithm's design, but simulation will help to confirm our analysis.

bboston7:
# Description

Closes #4561

This change adds a couple more tests to the nomination tests for the following scenarios:
1. Asymmetric quorums
2. Unresponsive validators

It also removes some protocol gating in the test suite that's no longer necessary now that P22 is the current protocol in the build.

# Checklist
- [x] Compiles
- [x] Ran all tests

### DIFF
--- a/src/herder/test/HerderTests.cpp
+++ b/src/herder/test/HerderTests.cpp
@@ -5806,53 +5806,48 @@ testWeights(std::vector<ValidatorEntry> const& validators)
 
     VirtualClock clock;
     Application::pointer app = createTestApplication(clock, cfg);
-    for_versions_from(
-        static_cast<uint32>(
-            APPLICATION_SPECIFIC_NOMINATION_LEADER_ELECTION_PROTOCOL_VERSION),
-        *app, [&]() {
-            // Collect info about orgs
-            ValidatorQuality maxQuality;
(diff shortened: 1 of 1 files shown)


marta-lokhova (approved):
thanks!

PR #4575, opened 2024-12-13
Remove SQL for (most) Ledger State

### LINKED ISSUE #4503: Make Background Eviction Mandatory
Starting in protocol 23, all validators need to run background eviction. There is no reason to run main thread eviction, so we should remove the flag and make background eviction required.

SirTyson:
# Description

Resolves #4503 and #4502

This change removes SQL support for all ledger state, except for the offers table. It also enforces background eviction. In order to not break Horizon, core will not crash if any of the old flags or config settings are used. Instead it will print a warning and ignore the flags. The flags should be removed in the next protocol release.

# Checklist
- [x] Compiles
- [x] Ran all tests


src/catchup/ApplyBucketsWork.cpp
```diff
@@ -120,32 +101,10 @@ ApplyBucketsWork::doReset()
...
-                {
-                    rootAcc.erase();
-                }
-                ltx.commit();
-            }
-        }
+        // The current size of this set is 1.6 million during BucketApply
+        // (as of 12/20/23). There's not a great way to estimate this, so
+        // reserving with some extra wiggle room
+        mSeenKeys.reserve(2'000'000);
```
marta-lokhova, line 107: nit: avoid magic numbers, a bit clearer to use a static const

src/database/Database.cpp
```diff
@@ -252,13 +252,8 @@ Database::upgradeToCurrentSchema()
...
 
-    // While not really a schema upgrade, we need to upgrade the DB when
-    // BucketListDB is enabled.
-    if (mApp.getConfig().isUsingBucketListDB())
-    {
-        // Tx meta column no longer supported in BucketListDB
-        dropTxMetaIfExists();
-    }
+    // Tx meta column no longer supported
+    dropTxMetaIfExists();
```
marta-lokhova, line 256: this can go away now: DB schema version 23 drops txhistory table completely!

src/main/ApplicationImpl.cpp
```diff
@@ -734,67 +642,35 @@ ApplicationImpl::validateAndLogConfig()
...
-    if (mConfig.DEPRECATED_SQL_LEDGER_STATE)
+    auto pageSizeExp = mConfig.BUCKETLIST_DB_INDEX_PAGE_SIZE_EXPONENT;
+    if (pageSizeExp != 0)
     {
-        if (mPersistentState->getState(PersistentState::kDBBackend) ==
-            BucketIndex::DB_BACKEND_STATE)
+        // If the page size is less than 256 bytes, it is essentially
+        // indexing individual keys, so page size should be set to 0
+        // instead.
+        if (pageSizeExp < 8)
```
marta-lokhova, line 650: nit: avoid magic numbers, same for line 660

src/main/Config.cpp
```diff
@@ -1082,16 +1081,20 @@ Config::processConfig(std::shared_ptr<cpptoml::table> t)
...
                 // Still support EXPERIMENTAL_BUCKETLIST_DB* flags for
                 // captive-core for 21.0 release, remove in 21.1 release
                 {"EXPERIMENTAL_BUCKETLIST_DB",
                  [&]() {
-                     DEPRECATED_SQL_LEDGER_STATE = !readBool(item);
                      CLOG_WARNING(
                          Bucket,
-                         "EXPERIMENTAL_BUCKETLIST_DB flag is deprecated, "
-                         "use DEPRECATED_SQL_LEDGER_STATE=false instead.");
+                         "EXPERIMENTAL_BUCKETLIST_DB flag is deprecated. "
```
marta-lokhova, line 1096: the comment says  EXPERIMENTAL_BUCKETLIST_DB can be removed after v21.1, can we remove those or is the comment inaccurate?

src/main/Config.cpp
```diff
@@ -1082,16 +1081,20 @@ Config::processConfig(std::shared_ptr<cpptoml::table> t)
                          "is ignored. Please remove from config");
                  }},
                 {"DEPRECATED_SQL_LEDGER_STATE",
-                 [&]() { DEPRECATED_SQL_LEDGER_STATE = readBool(item); }},
+                 [&]() {
+                     CLOG_WARNING(
```
marta-lokhova, line 1085: Are we ok with silently migrating to BucketListDB? (and deleting ledger state tables without explicitly flagging this to operators)

src/main/PersistentState.h
```diff
@@ -46,9 +46,9 @@ class PersistentState
     setSCPStateV1ForSlot(uint64 slot, std::string const& value,
                          std::unordered_map<Hash, std::string> const& txSets);
 
-    bool shouldRebuildForType(LedgerEntryType let);
-    void clearRebuildForType(LedgerEntryType let);
-    void setRebuildForType(LedgerEntryType let);
+    bool shouldRebuildForOfferTable();
```
marta-lokhova, line 49: `kDBBackend` can be removed now (this doesn't need to be a migration, but we need to delete that row)

src/database/Database.cpp
```diff
@@ -293,6 +294,50 @@ Database::dropTxMetaIfExists()
...
+        CLOG_INFO(Database, "Dropping table liquiditypool");
+        getSession() << "DROP TABLE IF EXISTS liquiditypool;";
+
+        CLOG_INFO(Database, "Dropping table trustlines");
+        getSession() << "DROP TABLE IF EXISTS trustlines;";
+
+        CLOG_INFO(Database, "Dropping table ttl");
+        getSession() << "DROP TABLE IF EXISTS ttl;";
+
+        mApp.getPersistentState().setState(PersistentState::kDBBackend,
```
marta-lokhova, line 336: Once nodes do this "migration", does BucketListDB just work or do operators need to run some command like new-db? Or I guess it can work right away, just need to stall for a bit to rebuild the indexes, right?

marta-lokhova (approved):
very nice to see so much legacy code go away, thank you! I left a few questions wrt the migration, and a couple of minor nits. Otherwise happy to merge this asap.

src/main/ApplicationImpl.cpp
```diff
@@ -802,24 +795,6 @@ ApplicationImpl::validateAndLogConfig()
...
-        if (!mConfig.isUsingBucketListDB())
-        {
-            throw std::invalid_argument(
-                "BACKGROUND_EVICTION_SCAN set to true but "
-                "DEPRECATED_SQL_LEDGER_STATE is set to true. "
-                "DEPRECATED_SQL_LEDGER_STATE must be set to false to enable "
-                "background eviction.");
-        }
-
-        if (mConfig.WORKER_THREADS < 2)
```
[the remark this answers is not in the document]
SirTyson: I removed this check because it's redundant, we already enforce the minimum of two in Config.cpp:1337
```
{"WORKER_THREADS",
 [&]() { WORKER_THREADS = readInt<int>(item, 2, 1000); }},
```

src/main/Config.cpp
SirTyson: Given that Horizon only wants to remove flags in protocol upgrades, this will have to stick around until v23. I'll update the comment accordingly and link an issue.

src/main/Config.cpp
SirTyson: Yeah, we've given as much warning as we can, and the initial `DEPRECATED_SQL_LEDGER_STATE` was intended to be the warning flag. It doesn't matter at this point if state tables get dropped since the node would fall of the network soon anyway.

src/database/Database.cpp
SirTyson: It depends. If the `run` user has DB admin privileges, it will just work. If not, the operator will have to call `upgrade-db` with the correct user, so it pretty much works like a DB schema upgrade.

src/main/PersistentState.h
SirTyson: I'd like to keep it in until the next release. Reason is `maybeUpgradeToBucketListDB` gets called on the startup path and checks `kDBBackend`. If it's not BucketListDB, it drops a bunch of tables. We can't do this unconditionally on startup as some operators have separate users for DB maintenance and call `upgrade-db` from a separate user. After the migration, we can drop this.

src/main/Config.cpp
```diff
@@ -1066,16 +1065,21 @@ Config::processConfig(std::shared_ptr<cpptoml::table> t)
                  }},
                 {"BACKGROUND_OVERLAY_PROCESSING",
                  [&]() { BACKGROUND_OVERLAY_PROCESSING = readBool(item); }},
+                // TODO: Flags are no longer supported, remove in next release.
```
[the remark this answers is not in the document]
marta-lokhova: thanks! the todo comment can be removed now

src/main/Config.cpp
marta-lokhova: thanks. let's stick with issues then rather than a TODO (as it suggests unfinished code)

marta-lokhova (approved):

Issue #4488, opened 2024-09-26
loadgen: add a new mode to stop an existing run


marta-lokhova:
This would be helpful in simulations where we generate load from multiple nodes, and need to restart load generation with a different tx rate.

PR #4579, opened 2024-12-17
Add loadgen "stop" mode

marta-lokhova:
Resolves https://github.com/stellar/stellar-core/issues/4488

This should help speed up max TPS runs

### DIFF
--- a/src/simulation/test/LoadGeneratorTests.cpp
+++ b/src/simulation/test/LoadGeneratorTests.cpp
@@ -124,6 +124,26 @@ TEST_CASE("generate load with unique accounts", "[loadgen]")
             },
             10 * Herder::EXP_LEDGER_TIMESPAN_SECONDS, false);
     }
+    SECTION("stop loadgen")
+    {
+        loadGen.generateLoad(GeneratedLoadConfig::createAccountsLoad(
+            /* nAccounts */ 10000,
+            /* txRate */ 1));
+        simulation->crankForAtLeast(std::chrono::seconds(10), false);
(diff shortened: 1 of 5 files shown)


src/simulation/test/LoadGeneratorTests.cpp
```diff
@@ -124,6 +124,26 @@ TEST_CASE("generate load with unique accounts", "[loadgen]")
...
     }
+    SECTION("stop loadgen")
+    {
+        loadGen.generateLoad(GeneratedLoadConfig::createAccountsLoad(
+            /* nAccounts */ 10000,
+            /* txRate */ 1));
+        simulation->crankForAtLeast(std::chrono::seconds(10), false);
+        auto& acc = app.getMetrics().NewMeter({"loadgen", "account", "created"},
+                                              "account");
+        auto numAccounts = acc.count();
```
SirTyson, line 135: Nit: Check that completed loadgen runs == 0 here to make sure we didn't just complete the accounts load before running stop.

SirTyson (approved):
Small test nit but LGTM

src/simulation/test/LoadGeneratorTests.cpp
SirTyson: Also is 10 seconds of wall clock a little excessive here?

marta-lokhova: >Nit: Check that completed loadgen runs == 0 here to make sure we didn't just complete the accounts load before running stop.

isn't it what the line below does? it goes from 0->1 on failed runs. 

> Also is 10 seconds of wall clock a little excessive here?
this is loopback mode, so not real time

PR #4585, opened 2024-12-18
Evict persistent entries to Hot Archive

### LINKED ISSUE #4395: Evict persistent entries to Hot Archive during eviction scans

SirTyson:
# Description

Resolves #4395

This PR evicts persitent entries to the Hot Archive in p23. Specifically, the background eviction scan will delete expired persistent entries from the live bucketlist and add them to the hot archive. This change also includes eviction meta for these types.

Operations have not yet been modified to work with the evicted entries. This means that `InvokeHostFunctionOp` does not check the hot archive for the existence of archived entries, and `RestoreFootprintOp` will not be able to restore evicted entries. These changes will be added in a follow up PR.

# Checklist
- [x] Compiles
- [x] Ran all tests

### DIFF
--- a/src/bucket/BucketIndexImpl.cpp
+++ b/src/bucket/BucketIndexImpl.cpp
@@ -589,11 +589,22 @@ BucketIndexImpl<IndexT>::operator==(BucketIndex const& inRaw) const
 
     if constexpr (std::is_same_v<IndexT, RangeIndex>)
     {
-        releaseAssert(mData.filter);
-        releaseAssert(in.mData.filter);
-        if (!(*(mData.filter) == *(in.mData.filter)))
+        // If both indexes have a filter, check if they are equal
+        if (mData.filter && in.mData.filter)
         {
-            return false;
+            if (!(*(mData.filter) == *(in.mData.filter)))
+            {
+                return false;
+            }
+        }
+        else
+        {
+            // If both indexes don't fave a filter, check that each filter is
+            // null
+            if (mData.filter || in.mData.filter)
+            {
+                return false;
+            }
         }
     }
     else
--- a/src/bucket/BucketManager.h
+++ b/src/bucket/BucketManager.h
@@ -102,7 +102,7 @@ class BucketManager : NonMovableOrCopyable
     std::map<LedgerEntryTypeAndDurability, medida::Counter&>
         mBucketListEntrySizeCounters;
 
-    std::future<EvictionResult> mEvictionFuture{};
+    std::future<EvictionResultCandidates> mEvictionFuture{};
 
     // Copy app's config for thread-safe access
     Config const mConfig;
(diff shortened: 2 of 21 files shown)


src/ledger/LedgerCloseMetaFrame.cpp
```diff
@@ -145,28 +146,22 @@ LedgerCloseMetaFrame::populateTxSet(TxSetXDRFrame const& txSet)
...
-        }
+        releaseAssertOrThrow(isTemporaryEntry(key) || key.type() == TTL);
+        mLedgerCloseMeta.v1().evictedTemporaryLedgerKeys.emplace_back(key);
+    }
+    for (auto const& entry : evictedEntries.second)
+    {
+        releaseAssertOrThrow(isPersistentEntry(entry.data));
+        // Unfortunately, for legacy purposes, evictedTemporaryLedgerKeys is
+        // misnamed and stores all evicted keys, both temp and persistent.
+        mLedgerCloseMeta.v1().evictedTemporaryLedgerKeys.emplace_back(
```
dmkozh, line 163: Are we still planning to rename this field?

src/bucket/BucketManager.h
```diff
@@ -292,10 +292,17 @@ class BucketManager : NonMovableOrCopyable
...
     // have been evicted or maxEvictionScanSize bytes have been scanned.
-    void startBackgroundEvictionScan(uint32_t ledgerSeq);
-    void resolveBackgroundEvictionScan(AbstractLedgerTxn& ltx,
-                                       uint32_t ledgerSeq,
-                                       LedgerKeySet const& modifiedKeys);
+    void startBackgroundEvictionScan(uint32_t ledgerSeq, uint32_t ledgerVers);
+
+    // Returns a pair of vectors representing entries evicted this ledger, where
+    // the first vector constains all deleted keys (TTL and temporary), and
+    // the second vector contains all archived keys (persistent and
```
dmkozh, line 299: nit: Archived entries.

src/bucket/BucketManager.h
```diff
@@ -292,10 +292,17 @@ class BucketManager : NonMovableOrCopyable
...
+    // ContractCode). Note that when an entry is archived, its TTL key will be
+    // included in the deleted keys vector.
+    std::pair<std::vector<LedgerKey>, std::vector<LedgerEntry>>
```
dmkozh, line 302: nit: I may be overly cautious here, but I feel like it would be safer to return a struct with named fields here. Even though the types are different, there is still some risk of confusion. At least it definitely would make the caller code easier to read.

src/bucket/BucketIndexImpl.cpp
```diff
@@ -589,9 +589,18 @@ BucketIndexImpl<IndexT>::operator==(BucketIndex const& inRaw) const
 
     if constexpr (std::is_same_v<IndexT, RangeIndex>)
     {
-        releaseAssert(mData.filter);
-        releaseAssert(in.mData.filter);
-        if (!(*(mData.filter) == *(in.mData.filter)))
+        if (!mData.filter)
```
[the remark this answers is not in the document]
SirTyson: This is checking that each index either:
1. Doesn't have a filter pointer
2. Has the same filter as the other index

What you provided won't work, as there are more checks later in the function for other index fields, so we can't return true after checking the filter. I've update this function though and added some comments to hopefully make it more clear.

src/ledger/LedgerCloseMetaFrame.cpp
SirTyson: Yes, but I haven't pushed the XDR changes yet.

dmkozh (dismissed):

dmkozh (approved):

Issue #4318, opened 2024-05-14
background ledger close: cleanup core interfaces and add invariance (tech debt)


anupsdf:
- [x] TransactionFrame and OperationFrame access to Application during validation and apply time makes applying in the background more dangerous. In reality, we just need access to medida metrics, and some configs (soroban and core config). Resolved in https://github.com/stellar/stellar-core/pull/4471
- [ ] Similarly, BucketManager uses Application where it doesn't need. We should refactor it such that we can do addBatch atomically in the background during ledgerClose.
- [ ] https://github.com/stellar/stellar-core/issues/4607
- [ ] #4592

marta-lokhova:
Merge after https://github.com/stellar/stellar-core/issues/4316 is done

marta-lokhova:
Closing this: we're tracking the remaining https://github.com/stellar/stellar-core/issues/4592 and https://github.com/stellar/stellar-core/issues/4607 individually

PR #4602, opened 2025-01-02
Make read-only ledger state const

marta-lokhova:
Improve access to ledger state, to better support parallelization changes in #4543 

Note that management of SorobanNetworkConfig is still not great, as currently LM manages multiple versions of the config. Ideally, soroban network config lives inside of the state snapshot (either BucketList snapshot or LedgerTxn), but this was too tricky to implement at this time due to how network config is currently implemented. We may need to clean this up later. 

This change also partially addresses https://github.com/stellar/stellar-core/issues/4318

### DIFF
--- a/src/bucket/BucketListSnapshotBase.cpp
+++ b/src/bucket/BucketListSnapshotBase.cpp
@@ -51,10 +51,9 @@ BucketListSnapshot<BucketT>::getLedgerSeq() const
 
 template <class BucketT>
 LedgerHeader const&
-SearchableBucketListSnapshotBase<BucketT>::getLedgerHeader()
+SearchableBucketListSnapshotBase<BucketT>::getLedgerHeader() const
 {
     releaseAssert(mSnapshot);
-    mSnapshotManager.maybeUpdateSnapshot(mSnapshot, mHistoricalSnapshots);
     return mSnapshot->getLedgerHeader();
 }
 
@@ -85,7 +84,7 @@ SearchableBucketListSnapshotBase<BucketT>::loopAllBuckets(
 
 template <class BucketT>
 std::shared_ptr<typename BucketT::LoadT>
-SearchableBucketListSnapshotBase<BucketT>::load(LedgerKey const& k)
+SearchableBucketListSnapshotBase<BucketT>::load(LedgerKey const& k) const
 {
     ZoneScoped;
 
@@ -108,7 +107,6 @@ SearchableBucketListSnapshotBase<BucketT>::load(LedgerKey const& k)
         }
     };
 
-    mSnapshotManager.maybeUpdateSnapshot(mSnapshot, mHistoricalSnapshots);
     if (threadIsMain())
     {
         mSnapshotManager.startPointLoadTimer();
@@ -127,7 +125,8 @@ SearchableBucketListSnapshotBase<BucketT>::load(LedgerKey const& k)
 template <class BucketT>
 std::optional<std::vector<typename BucketT::LoadT>>
 SearchableBucketListSnapshotBase<BucketT>::loadKeysFromLedger(
-    std::set<LedgerKey, LedgerEntryIdCmp> const& inKeys, uint32_t ledgerSeq)
+    std::set<LedgerKey, LedgerEntryIdCmp> const& inKeys,
+    uint32_t ledgerSeq) const
 {
     return loadKeysInternal(inKeys, /*lkMeter=*/nullptr, ledgerSeq);
 }
@@ -141,11 +140,13 @@ BucketLevelSnapshot<BucketT>::BucketLevelSnapshot(
 
 template <class BucketT>
 SearchableBucketListSnapshotBase<BucketT>::SearchableBucketListSnapshotBase(
-    BucketSnapshotManager const& snapshotManager)
-    : mSnapshotManager(snapshotManager), mHistoricalSnapshots()
+    BucketSnapshotManager const& snapshotManager,
+    SnapshotPtrT<BucketT>&& snapshot,
+    std::map<uint32_t, SnapshotPtrT<BucketT>>&& historicalSnapshots)
+    : mSnapshotManager(snapshotManager)
+    , mSnapshot(std::move(snapshot))
+    , mHistoricalSnapshots(std::move(historicalSnapshots))
 {
-
-    mSnapshotManager.maybeUpdateSnapshot(mSnapshot, mHistoricalSnapshots);
 }
 
 template <class BucketT>
--- a/src/bucket/BucketSnapshotManager.cpp
+++ b/src/bucket/BucketSnapshotManager.cpp
@@ -52,21 +52,44 @@ BucketSnapshotManager::BucketSnapshotManager(
     }
 }
 
-std::shared_ptr<SearchableLiveBucketListSnapshot>
+template <class BucketT>
+std::map<uint32_t, SnapshotPtrT<BucketT>>
+copyHistoricalSnapshots(
+    std::map<uint32_t, SnapshotPtrT<BucketT>> const& snapshots)
+{
+    std::map<uint32_t, SnapshotPtrT<BucketT>> copiedSnapshots;
+    for (auto const& [ledgerSeq, snap] : snapshots)
+    {
+        copiedSnapshots.emplace(
+            ledgerSeq,
+            std::make_unique<BucketListSnapshot<BucketT> const>(*snap));
+    }
+    return copiedSnapshots;
+}
+
+SearchableSnapshotConstPtr
 BucketSnapshotManager::copySearchableLiveBucketListSnapshot() const
 {
     // Can't use std::make_shared due to private constructor
     return std::shared_ptr<SearchableLiveBucketListSnapshot>(
-        new SearchableLiveBucketListSnapshot(*this));
+        new SearchableLiveBucketListSnapshot(
+            *this,
+            std::make_unique<BucketListSnapshot<LiveBucket>>(
+                *mCurrLiveSnapshot),
+            copyHistoricalSnapshots(mLiveHistoricalSnapshots)));
 }
 
-std::shared_ptr<SearchableHotArchiveBucketListSnapshot>
+SearchableHotArchiveSnapshotConstPtr
 BucketSnapshotManager::copySearchableHotArchiveBucketListSnapshot() const
 {
     releaseAssert(mCurrHotArchiveSnapshot);
     // Can't use std::make_shared due to private constructor
     return std::shared_ptr<SearchableHotArchiveBucketListSnapshot>(
-        new SearchableHotArchiveBucketListSnapshot(*this));
+        new SearchableHotArchiveBucketListSnapshot(
+            *this,
+            std::make_unique<BucketListSnapshot<HotArchiveBucket>>(
+                *mCurrHotArchiveSnapshot),
+            copyHistoricalSnapshots(mHotArchiveHistoricalSnapshots)));
 }
 
 medida::Timer&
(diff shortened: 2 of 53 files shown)


src/ledger/LedgerManagerImpl.h
```diff
@@ -50,7 +50,13 @@ class LedgerManagerImpl : public LedgerManager
 
   private:
     LedgerHeaderHistoryEntry mLastClosedLedger;
-    std::optional<SorobanNetworkConfig> mSorobanNetworkConfig;
+
+    // Read-only Soroban network configuration, accessible by main thread only.
+    std::shared_ptr<SorobanNetworkConfig const> mSorobanNetworkConfigReadOnly;
```
dmkozh, line 55: The comments here seem a bit confusing. Here is my understanding:

- The config used during apply has to be derived from the LCL. It can indeed be modified but only after applying all the transactions. Basically it's conceptually just a local copy used during apply, it just happens to be a member of LM for simpler implementation. Also IIUC the reason for it to be 'half-valid' is in fact a race condition, right? Maybe we could say that the access to this is not synchronized, as it's meant to be used by a single (apply) thread only.

- The config used in all the remaining contexts does not *have* to match LCL, it rather should match the respective state snapshot. It's fine that today we achieve this via using it only on the main thread, but it's not a hard requirement, so maybe it's worth mentioning that the underlying requirement here is consistency with the snapshot (and main-thread-only access is just the current implementation).

If these are valid, it would be nice to reflect some points in the comments as well.

src/main/QueryServer.cpp
```diff
@@ -132,7 +133,15 @@ QueryServer::getLedgerEntryRaw(std::string const& params,
 
     if (!keys.empty())
     {
-        auto& bl = *mBucketListSnapshots.at(std::this_thread::get_id());
+        auto& snapshotPtr = mBucketListSnapshots.at(std::this_thread::get_id());
+        if (snapshotPtr->getLedgerSeq() <
```
dmkozh, line 137: This block is non-atomic, which is probably not a big deal here, but could be problematic if this pattern reused in the place where this does matter. It would be safer to provide a single function `maybeUpdateSnapshot` that atomically verifies the ledger sequence and assigns the new snapshot if necessary. At the very least that would be more efficient as only 1 lock would be necessary.

src/testdata/ledger-close-meta-v1-protocol-20.json
```diff
@@ -6,24 +6,24 @@
                 "v": 0
             },
             "ledgerHeader": {
-                "hash": "a50c06f398679584d4acff68de5130502114df913f41c270a4bce9e1d6cd5c36",
+                "hash": "96bbf541d371216daf6d52718c6c80c40cbfa67b51e578df821c238af37fcf6b",
```
dmkozh, line 9: Do we understand the reason for the LCM changes?

src/transactions/TransactionUtils.h
```diff
@@ -262,7 +262,8 @@ bool accountFlagMaskCheckIsValid(uint32_t flag, uint32_t ledgerVersion);
 bool hasMuxedAccount(TransactionEnvelope const& e);
 
 bool
-isTransactionXDRValidForCurrentProtocol(AppConnector& app,
+isTransactionXDRValidForCurrentProtocol(uint32_t currProtocol,
```
dmkozh, line 265: nit: This function should probably be renamed to just `isTransactionXDRValidForProtocol` (as we pass the protocol explicitly).

Also it seems like we can pass `Config` instead of the whole `AppConnector` now

src/testdata/ledger-close-meta-v1-protocol-20.json
SirTyson: If I understand correctly, when we manually do soroban network config upgrades, we call `LiveBucketList::addBatch` twice in the same ledger, so we have two different snapshots with the same ledgerSeq. LedgerSeq is what we use to determine if a snapshot is out of date previously in `maybeUpdateSnapshot`, so when doing manual upgrades, the child snapshot would be stale (the child snapshot is based off of the 1st addBatch call, not the 2nd), but would not be updated since the ledgerSeq of the stale snapshot is still "up to date". This PR changes is such that `maybeUpdateSnapshot` just checks for a null pointer but doesn't actually check the ledgerSeq, so child snapshots are always updated when they call `maybeUpdateSnapshot`.

src/bucket/BucketListSnapshotBase.cpp
```diff
@@ -144,7 +143,6 @@ SearchableBucketListSnapshotBase<BucketT>::SearchableBucketListSnapshotBase(
     BucketSnapshotManager const& snapshotManager)
     : mSnapshotManager(snapshotManager), mHistoricalSnapshots()
 {
-
     mSnapshotManager.maybeUpdateSnapshot(mSnapshot, mHistoricalSnapshots);
```
SirTyson, line 146: I don't think the `maybeUpdateSnapshot` interface makes sense anymore, as the snapshot is always updated and this function is basically just a `SearchableBucketListSnapshotBase` constructor now. I think the interface that makes the most sense is having two getter functions in the snapshot manager:

`ConstSnapshotPtr BucketSnapshotManager::copySearchableBucketListSnapshot()` (as it exists currently). This is used for getting a new snapshot when you don't have one already or when you know you definitely need a new one, such as in `LedgerTxnRoot`

`ConstSnapshotPtr BucketSnapshotManager::maybeCopyNewSearchableBucketListSnapshot(ConstSnapshotPtr&& oldPtr)`. This function would be used in places like `QueryServer` where you want to atomically check and maybe replace your snapshot. This could potentially be combined into a singular get snapshot interface or not. 

`maybeUpdateSnapshot` should be removed and inlined to the `SearchableBucketListSnapshotBase` constructor or at least renamed.

src/bucket/BucketSnapshotManager.cpp
```diff
@@ -131,16 +131,10 @@ BucketSnapshotManager::maybeUpdateSnapshotInternal(
...
-        snapshot->getLedgerSeq() != managerSnapshot->getLedgerSeq())
-    {
-        // Should only update with a newer snapshot
-        releaseAssert(!snapshot || snapshot->getLedgerSeq() <
-                                       managerSnapshot->getLedgerSeq());
-        snapshot = std::make_unique<BucketListSnapshot<BucketT> const>(
-            *managerSnapshot);
-    }
+    // Should only update on snapshot creation
+    releaseAssert(!snapshot);
```
SirTyson, line 135: Given this assert, we always update the snapshot. This function would make more sense as a constructor of `SearchableBucketListSnapshotBase`

src/bucket/BucketSnapshotManager.cpp
```diff
@@ -231,4 +225,11 @@ BucketSnapshotManager::endPointLoadTimer(LedgerEntryType t,
         iter->second.Update(duration);
     }
 }
+
+uint32_t
+BucketSnapshotManager::getCurrentLedgerSeq() const
```
SirTyson, line 230: +1 to Dima. I think this function is general is a little unsafe as it leads to non-atomic updates.

src/ledger/LedgerManager.h
```diff
@@ -129,6 +133,8 @@ class LedgerManager
     // Ledger txn here is needed for the sake of lazy load; it won't be
     // used most of the time.
     virtual SorobanNetworkConfig const& getSorobanNetworkConfig() = 0;
+    virtual SorobanNetworkConfig const& getSorobanNetworkConfigForApply() = 0;
```
SirTyson, line 136: `resolveBackgroundEvictionScan` needs to call `getSorobanNetworkConfigForApply`. If the network config settings changes mid ledger, we need to invalidate the previous scan and start over based on the new settings. Currently this won't happen since we're using the RO snapshot and not the apply time snapshot. `startBackgroundEvictionScan` should also probably use `getSorobanNetworkConfigForApply` just for consistency (although `getSorobanNetworkConfig` should have the same value at that point anyway). Note that while the eviction scan does most it's work in the background, when it's joined at the resolution step, it should be considered part of TX apply.

SirTyson (changes requested):
I'm not a huge fan of the interface where we maintain two separate copies of `SorobanNetworkConfigSettings` at the `LedgerManager` level, as it sometimes unclear which method to use and can lead to tricky bugs (such as with background eviction). Given that the network settings are part of ledger state, I would still prefer threads to maintain a canonical snapshot and route all ledger based queries through that snapshot, including LedgerEntry, ledger header, and network setting accesses. That being said, I wanna get you unstuck, and am fine with this approach in the interim as we move towards more thread safety in general. To merge this I do think it would be good to clean up the new snapshot interface a bit, specifically with the `maybeUpdateSnapshot` issues I raised + atomic snapshot swaps.

src/ledger/LedgerManagerImpl.cpp
```diff
@@ -1266,37 +1287,43 @@ LedgerManagerImpl::advanceLedgerPointers(LedgerHeader const& header,
                    ledgerAbbrev(header, ledgerHash));
     }
 
-    auto prevLedgerSeq = mLastClosedLedger.header.ledgerSeq;
+    // NB: with parallel ledger close, this will have to be called strictly from
```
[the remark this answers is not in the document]
marta-lokhova: yeah, this is already the case in parallel ledger close changes on top of this PR.

src/ledger/LedgerManagerImpl.h
marta-lokhova: Yeah, this is mostly correct. The only thing to clarify is this bit:
> Also IIUC the reason for it to be 'half-valid' is in fact a race condition, right?
There is no race condition, since this variable is always accessed by the same thread. What I meant by "half-valid" is that this variable is being actively modified, so it's not safe to use it outside of apply. I'll add the comments you suggested.

src/testdata/ledger-close-meta-v1-protocol-20.json
marta-lokhova: The main reason the meta test broke was because I removed snapshot auto-updates, and reset the snapshot for each new LedgerTxn (previously it had a long-living pointer that called into maybeUpdateSnapshot). So because of the reset, LedgerTxn now always picks up the most up-to-date snapshot (in this case, the one from the 2nd call to addBatch). This was not the case previously, as that test used the first snapshot. Note that the meta breaks regardless of the ledgerSeq check in `maybeUpdateSnapshot` (that change followed later).

marta-lokhova:
@dmkozh @SirTyson thanks for taking a look! I believe your comments are addressed now.

src/main/QueryServer.cpp
```diff
@@ -132,7 +133,11 @@ QueryServer::getLedgerEntryRaw(std::string const& params,
 
     if (!keys.empty())
     {
-        auto& bl = *mBucketListSnapshots.at(std::this_thread::get_id());
+        auto snapshotPtr = mBucketListSnapshots.at(std::this_thread::get_id());
```
SirTyson, line 136: This needs to be `auto& snapshotPtr`, since currently we never actually update the snapshot in the vector and would have to get a new copy on every call.

src/transactions/TransactionFrame.cpp
```diff
@@ -875,7 +875,8 @@ TransactionFrame::isTooEarlyForAccount(LedgerHeaderWrapper const& header,
 
 bool
 TransactionFrame::commonValidPreSeqNum(
-    AppConnector& app, LedgerSnapshot const& ls, bool chargeFee,
+    AppConnector& app, std::optional<SorobanNetworkConfig> cfg,
```
SirTyson, line 878: Nit: Can we make  these `std::optional<SorobanNetworkConfig>& cfg` to avoid extra copies?

src/main/QueryServer.cpp
marta-lokhova: yeah sorry, it was as of last commit, but got removed by accident. Fixed now.

SirTyson (approved):
LGTM

SirTyson (approved):

PR #4603, opened 2025-01-02
Hot archive tx apply

### LINKED ISSUE #4584: Support the Hot Archive in TX apply
For protocol 23, we need to check both the live BucketList and the Hot Archive BucketList when checking for Archived keys. We also need to allow the restore operation to restore state from the Archived BucketList.

SirTyson:
# Description

Resolves #4584

This PR adds support for the Hot Archive to TX apply. Specifically, `InvokeHostFunctionOp` will check both the live BucketList and Hot Archive BucketList for archived entries. The `RestoreFootprintOp` can now restore entries from both the live BucketList and Hot Archive.

Additionally, restore meta has also changed for p23. The `LEDGER_ENTRY_RESTORED` type has been added for the restore op. When an entry is restored. both the `LedgerEntry` of the restored data and the corresponding `TTL` entry are emitted as `LEDGER_ENTRY_RESTORED LedgerEntryChangeType`. If the entry has not yet been evicted (such that the entry and it's TTL still exist in the live BucketList), the preexisting TTL value will be emitted as `LEDGER_ENTRY_STATE`. If the entry has been evicted such that the TTL value has been previously deleted, only a single change type of `LEDGER_ENTRY_RESTORED` will be emitted for the new value. For the data being restored, `LEDGER_ENTRY_STATE` is never emitted. 

Rebased on #4585.

# Checklist
- [x] Compiles
- [x] Ran all tests


src/transactions/TransactionFrame.cpp
```diff
@@ -57,6 +59,83 @@ namespace
...
 int64_t const MAX_RESOURCE_FEE = 1LL << 50;
+
+#ifdef ENABLE_NEXT_PROTOCOL_VERSION_UNSAFE_FOR_PRODUCTION
+// Starting in protocol 23, some operation meta needs to be modified
+// to be consumed by downstream systems. In particular, restoration is
+// logically a new entry creation from the perspective of ltx and stellar-core
+// as a whole, but this change type is reclassified to LEDGER_ENTRY_RESTORED
+// for easier consumption downstream.
+LedgerEntryChanges
+processOpLedgerEntryChanges(std::shared_ptr<OperationFrame const> op,
```
dmkozh, line 70: While auto-restore is out of scope of this PR, I find it a bit concerning that this whole function is hardcoded to only work with restore op. Is there a generic way of converting the changes that is independent of the operation that has updated them (it's ok to have a top-level condition for the sake of performance though). I wonder if adding a restoration helper to ltx (as I've suggested above) could help with that.

src/transactions/TransactionFrame.cpp
```diff
@@ -75,50 +74,77 @@ processOpLedgerEntryChanges(std::shared_ptr<OperationFrame const> op,
...
     for (auto& change : changes)
     {
+        // For entry creation meta, we only need to check for Hot Archive
+        // restores
         if (change.type() == LEDGER_ENTRY_CREATED)
         {
             auto le = change.created();
-            change.type(LEDGER_ENTRY_RESTORED);
-            change.restored() = le;
+            if (hotArchiveRestores.find(LedgerEntryKey(le)) !=
```
dmkozh, line 111: What happens if an entry is restored, then updated? Are we going to have 2 changes, or just 1? I wonder if we'll need to do more work for autorestore.

src/transactions/TransactionFrame.cpp
```diff
@@ -57,6 +58,112 @@ namespace
...
+                    // Update the TTL change from LEDGER_ENTRY_UPDATED to
+                    // LEDGER_ENTRY_RESTORED.
+                    change.type(LEDGER_ENTRY_RESTORED);
+                    change.restored() = ttlLe;
+                }
+            }
+        }
+    }
+
+    // Now we need to insert all the LEDGER_ENTRY_RESTORED changes for the
```
dmkozh, line 138: I don't quite understand this, aren't we adding duplicate changes here?

dmkozh (dismissed):

src/ledger/LedgerTxn.cpp
```diff
@@ -638,15 +637,25 @@ LedgerTxn::Impl::commitChild(
...
         }
     }
 
+    for (auto const& key : restoredKeys.liveBucketList)
+    {
+        auto [_, inserted] = mRestoredKeys.liveBucketList.emplace(key);
+        if (!inserted)
+        {
+            printErrorAndAbort(
+                "restored hlive BucketList entry already exists");
```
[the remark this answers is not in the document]
SirTyson: Fixed

src/transactions/TransactionFrame.cpp
SirTyson: There will be some changes, yes, but they should be minimal and work in this framework. If we edit an entry normally, the meta is
```
LEDGER_ENTRY_STATE(oldValue), LEDGER_ENTRY_UPDATED(newValue)
```

For auto restore if an entry is restored and modified, we'll need to change this to
```
LEDGER_ENTRY_RESTORE(oldValue), LEDGER_ENTRY_UPDATED(newValue)
```

By default, I think currently in the restore and update case the ltx will just have a single event type of `LEDGER_ENTRY_CREATE(newValue)`, so we'd have to manually add the restoration with the old value. This should be pretty easy though, for example we could save full ledger entires instead of just keys for the `RestoredKeys` object. I think the interface is close enough to working with auto restore as is and we can address these issues in the auto restore PR.

src/transactions/TransactionFrame.cpp
SirTyson: I've updated the comment explain this a bit more. When an entry still exists on the BucketList, all the restoration needs to do is change the TTL value, so we never actually change the code/data entry. This means that the ltx only has meta events for the TTL change. In this case, platform requested that we also emit the entry being restored in meta even though it hasn't changed, so we add it manually here. Otherwise, they'd have to do DB lookups during meta ingest to determine what TTL restore events map to which entries.

dmkozh: Thanks, I've missed that we have only TTL meta above.

src/transactions/TransactionFrame.cpp
dmkozh: Ok, for sure we don't need to address this now, I just want to make sure we have an extensible approach to meta in general.

dmkozh (commented):
LGTM, could you please rebase and squash?

SirTyson:
> LGTM, could you please rebase and squash?

Done

dmkozh (approved):

Issue #4607, opened 2025-01-06
tech debt: ledger state snapshots should contain complete state


marta-lokhova:
We're moving towards individual subsystems in core relying on read-only snapshots to query ledger state. There are several data structures that together comprise a complete "ledger state": the bucketlist, ledger header, and soroban config. The problem is that currently all these pieces of ledger state live in different places. Specifically, ledger header and soroban config live inside the LedgerManager, while the bucketlist lives inside of the bucketlist snapshot. This is quite error-prone and it definitely a footgun. It would be much easier to reason about consistency of ledger state snapshots if all relevant pieces lived inside of a single snapshot object. There is some WIP changes in https://github.com/stellar/stellar-core/pull/4597 already, which we can probably revive when we prioritize this.

marta-lokhova:
We should understand if this change is important for the parallel apply changes as well.

PR #4617, opened 2025-01-09
WIP: Background `tryAdd` functionality in TransactionQueue

### LINKED ISSUE #4316: background overlay: move doing transaction validation in the background before forwarding to tx queue, fast failing invalid transactions

bboston7:
This is a draft change that will resolve #4316 when it is complete. The change makes `TransactionQueue` thread safe and runs the `tryAdd` function in the background when the feature is enabled. The implementation closely follows the
[design document](https://docs.google.com/document/d/1pU__XfEp-rR-17TNsuj-VhY6JfyendaFSYLTiq6tIj4/edit?usp=sharing) I wrote.  The implementation still requires the main thread to re-broadcast the transactions (for now). I've opened this PR for visibility / early feedback on the implementation.

This change is very much a work in progress, with the following tasks remaining:

* [ ] Fix catchup. I seem to have broken catchup in rebasing these changes on master. I need to figure out what is going on there and fix it.
* [ ] Fix failing tests. These are failing because they don't update `TransactionQueue`s new snapshots correctly.
* [ ] Rigorous testing, both for correctness and performance.
* [ ] Address all TODOs I added
* [ ] Improve documentation
* [ ] I'd like to take a look at pushing the cut-point out a bit to enable flooding in the background as well. If this is a relatively simple change, I'd like to roll it into this PR. If it looks hairy, then I'll leave it for a separate change later.


src/herder/TransactionQueue.cpp
```diff
@@ -448,20 +448,21 @@ TransactionQueue::canAdd(
...
-                         .getLastClosedLedgerHeader()
-                         .header.scpValue.closeTime;
+    auto closeTime = lh.scpValue.closeTime;
     if (protocolVersionStartsFrom(ledgerVersion, ProtocolVersion::V_19))
     {
         // This is done so minSeqLedgerGap is validated against the next
         // ledgerSeq, which is what will be used at apply time
-        ls.getLedgerHeader().currentToModify().ledgerSeq =
-            mApp.getLedgerManager().getLastClosedLedgerNum() + 1;
+        ++ls.getLedgerHeader().currentToModify().ledgerSeq;
```
marta-lokhova, line 456: this should be correct, yeah

src/herder/TransactionQueue.cpp
```diff
@@ -380,13 +388,7 @@ TransactionQueue::canAdd(
...
                 auto txResult = tx->createSuccessResult();
                 if (!tx->checkSorobanResourceAndSetError(
-                        mApp.getAppConnector(),
-                        mApp.getLedgerManager()
-                            .getSorobanNetworkConfigReadOnly(),
-                        mApp.getLedgerManager()
-                            .getLastClosedLedgerHeader()
-                            .header.ledgerVersion,
-                        txResult))
+                        *mValidationSnapshot, ledgerVersion, txResult))
```
marta-lokhova, line 391: isn't ledgerVersion stored inside mValidationSnapshot already?

src/herder/TransactionQueue.h
```diff
@@ -170,7 +172,6 @@ class TransactionQueue
      */
     using BannedTransactions = std::deque<UnorderedSet<Hash>>;
 
-    Application& mApp;
```
marta-lokhova, line 173: Glad to see mApp removed! Seems like it's actually not too painful to get rid of it, please correct me if I'm wrong: it was mostly used to grab the LCL from LedgerManager, which is now done via mValidationSnapshot (and hopefully later can be done with just the mBucketSnapshot object). For the rest of the functionality, we use mAppConnector which either requires calling certain methods from the main thread, and forces users to implement thread-safe functions. From what I can tell, besides the timestamp getter, no new thread-safe methods needed to be introduced, which is definitely good news.

src/main/AppConnector.cpp
```diff
@@ -116,6 +133,17 @@ AppConnector::now() const
     return mApp.getClock().now();
 }
 
+VirtualClock::system_time_point
+AppConnector::system_now() const
+{
```
marta-lokhova, line 138: system clock access via chrono should be thread-safe. Virtual time wasn't thread-safe, but it should be now as of latest master (added in #4543 )

marta-lokhova: I realize the comment in AppConnector about thread-safety was a bit misleading, because it doesn't account for virtual time (it was initially used for overlay TCP peer only, which doesn't use virtual time, but now that we're expanding its use cases, the comment is no longer accurate)

src/herder/TransactionQueue.h
```diff
@@ -230,15 +234,37 @@ class TransactionQueue
...
+    SearchableSnapshotConstPtr mBucketSnapshot;
+
+    TxQueueLimiter mTxQueueLimiter;
     UnorderedMap<AssetPair, uint32_t, AssetPairHash> mArbitrageFloodDamping;
 
     UnorderedMap<Hash, TransactionFrameBasePtr> mKnownTxHashes;
 
     size_t mBroadcastSeed;
 
+    mutable std::recursive_mutex mTxQueueMutex;
```
marta-lokhova, line 248: is recursive_mutex required here? perhaps mutex is more appropriate?

src/herder/TransactionQueue.cpp
```diff
@@ -1167,24 +1184,20 @@ SorobanTransactionQueue::broadcastSome()
...
-        auto res = mTxQueueLimiter->maxScaledLedgerResources(true);
-        releaseAssert(res.size() == NUM_SOROBAN_TX_RESOURCES);
-        return res.getVal(Resource::Type::OPERATIONS);
-    }
-    else
-    {
-        return 0;
-    }
+    std::lock_guard<std::recursive_mutex> guard(mTxQueueMutex);
+    // TODO: I removed a conditional checking that the protocol version is
```
marta-lokhova, line 1188: I _think_ this could cause a crash if we start a node from scratch and try to catchup, since it'll be in protocol 0. Any reason to remove it at this time?

src/transactions/TransactionFrameBase.h
```diff
@@ -35,6 +35,57 @@ using TransactionFrameBasePtr = std::shared_ptr<TransactionFrameBase const>;
...
+    SorobanNetworkConfig const& getSorobanNetworkConfig() const override;
+    uint32_t getCurrentProtocolVersion() const override;
+
+  private:
+    AppConnector const& mApp;
+    bool const mForApply;
+};
+
+// TODO: Docs
+class ImmutableValidationSnapshot : public ValidationConnector
```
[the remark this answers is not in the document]
marta-lokhova: Note: if we go the route described above, I think that also means that we should be able to remove ImmutableValidationSnapshot argument from the whole checkValid flow, and use LedgerSnapshot instead to retrieve SorobanConfig/LCL/etc.

marta-lokhova (changes requested):

src/herder/TransactionQueue.cpp
```diff
@@ -266,6 +270,7 @@ isDuplicateTx(TransactionFrameBasePtr oldTx, TransactionFrameBasePtr newTx)
 bool
 TransactionQueue::sourceAccountPending(AccountID const& accountID) const
 {
+    std::lock_guard<std::recursive_mutex> guard(mTxQueueMutex);
```
marta-lokhova, line 273: Any insights into lock contention with the current implementation of tx queue synchronization? Might be useful to turn these into RECURSIVE_LOCK_GUARD, so you can observe thread blocking in Tracy.

marta-lokhova:
thanks for putting this together! I'm excited about moving more of transaction processing to the background. It also seems like making tx queue thread-safe is not as dangerous/scary as we thought, which is great. My two main suggestions are to consolidate all read-only ledger state into a single object to avoid footguns, and re-work recvTransaction flow a little bit to remove main thread completely (this way we'll have a synchronous flow independent on main thread, from when the message is popped off the wire all the way to tx queue insertion)

src/herder/TransactionQueue.h
bboston7: I used a `recursive_mutex` because the timer asynchronously calls the private function `broadcast`, so `broadcast` needs to grab the lock. That causes a problem because other public functions also call `broadcast`. That being said, I think I can refactor these functions to fix this. The async call really shouldn't invoke a private function, as it makes control flow harder to reason about.

src/herder/TransactionQueue.h
bboston7: That's exactly right!

src/herder/TransactionQueue.h
marta-lokhova: I see. If it's not too painful to refactor, I think it might be cleaner. I suggested a simple mutex, because it gives you more control and visibility of the code flow. For example, passing a lock guard reference as an argument is less error-prone than grabbing a recursive mutex (something we also learned with background overlay).

src/herder/TransactionQueue.cpp
bboston7: Ah, good point. I re-added the check. No particular reason to remove, other than that I was auditing uses of `mApp` and failed to think about the "catchup from 0" case here.

src/herder/TransactionQueue.h
bboston7: Done in 7bb9da0cac81c5793add740e9dfd5a5eeb342871

src/herder/TransactionQueue.cpp
bboston7: Yep. I removed passing around `ledgerVersion` in 16979788576cf86a4314dfb96be3751ef4233d9b

src/herder/HerderImpl.cpp
```diff
@@ -608,11 +608,31 @@ HerderImpl::recvTransaction(TransactionFrameBasePtr tx, bool submittedFromSelf)
     }
     else if (!tx->isSoroban())
     {
-        result = mTransactionQueue.tryAdd(tx, submittedFromSelf);
+        if (mApp.getConfig().BACKGROUND_TX_QUEUE && !submittedFromSelf)
+        {
+            mApp.postOnOverlayThread(
```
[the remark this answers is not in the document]
bboston7: Commenting for posterity: It turns out this approach doesn't offer much additional performance benefit, but greatly increases complexity. We saw in perf testing that the vast majority of the performance gains come from transaction validation in the background. Therefore, we decided to simplify by just doing that portion in the background.

src/transactions/TransactionFrameBase.h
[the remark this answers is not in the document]
bboston7: I added a subclass of `LedgerSnapshot` called `ExtendedLedgerSnapshot` that now contains everything needed for the `checkValid` flow. With this, I was able to remove `ValidationConnector` and all of its implementations.

src/herder/TransactionQueue.cpp
bboston7: I switched over the mutex I'm using now to be a tracy-compatible one.

bboston7:
I reworked this PR to do only validation in the background, as that's what our perf testing showed had the largest impact. It's now ready for review.

src/ledger/LedgerStateSnapshot.h
```diff
@@ -158,12 +165,43 @@ class LedgerSnapshot : public NonMovableOrCopyable
         return mGetter->getAccount(header, tx, AccountID);
     }
     LedgerEntryWrapper load(LedgerKey const& key) const;
+};
+
+// A subclass of LedgerSnapshot that contains additional snapshots necessary for
+// transaction validation.
+class ExtendedLedgerSnapshot : public LedgerSnapshot
```
marta-lokhova, line 172: Is there a reason to keep both LedgerSnapshot and ExtendedLedgerSnapshot? If it's just scope, it's fine to clean this up in the follow-up, but wanted to confirm.

src/ledger/LedgerStateSnapshot.h
```diff
@@ -158,12 +165,43 @@ class LedgerSnapshot : public NonMovableOrCopyable
...
-};
+        std::function<void(ExtendedLedgerSnapshot const&)> f) const;
 
+    // Getters for the additional snapshots
+    Config const& getConfig() const;
+    SorobanNetworkConfig const& getSorobanNetworkConfig() const;
+    uint32_t getCurrentProtocolVersion() const;
+
+  private:
+    std::shared_ptr<const Config> const mConfig;
```
marta-lokhova, line 203: mConfig doesn't belong to a snapshot, since it's `const` right? Typically, we should be able to access config from either mApp or mAppConnector.

src/overlay/OverlayManagerImpl.cpp
```diff
@@ -1182,54 +1199,109 @@ OverlayManagerImpl::checkScheduledAndCache(
...
+                                 .header.scpValue.closeTime;
+            auto offset =
+                getUpperBoundCloseTimeOffset(mApp.getAppConnector(), closeTime);
+            mApp.postOnTxValidationThread(
+                [this, transaction, offset, peer, index]() {
+                    ZoneScopedN("background checkValid");
+                    LOCK_GUARD(mSnapshotMutex, guard);
+                    releaseAssert(mLedgerSnapshot);
+
+                    auto txResult =
```
marta-lokhova, line 1267: I think we're missing some tests: for transaction validation purposes, ledger state needs to be slightly modified, similar to how we do it in `TransactionQueue::canAdd`
```
    if (protocolVersionStartsFrom(ledgerVersion, ProtocolVersion::V_19))
    {
        // This is done so minSeqLedgerGap is validated against the next
        // ledgerSeq, which is what will be used at apply time
        ls.getLedgerHeader().currentToModify().ledgerSeq =
            mApp.getLedgerManager().getLastClosedLedgerNum() + 1;
    }
```

src/ledger/LedgerStateSnapshot.h
```diff
@@ -158,12 +165,43 @@ class LedgerSnapshot : public NonMovableOrCopyable
...
+    std::optional<const SorobanNetworkConfig> const mSorobanNetworkConfig;
+    uint32_t const mCurrentProtocolVersion;
```
marta-lokhova, line 205: protocol version can be obtained from the ledger header stored by the snapshot, right?

src/ledger/LedgerStateSnapshot.h
```diff
@@ -158,12 +165,43 @@ class LedgerSnapshot : public NonMovableOrCopyable
...
+        std::function<void(ExtendedLedgerSnapshot const&)> f) const;
 
+    // Getters for the additional snapshots
+    Config const& getConfig() const;
+    SorobanNetworkConfig const& getSorobanNetworkConfig() const;
+    uint32_t getCurrentProtocolVersion() const;
+
+  private:
+    std::shared_ptr<const Config> const mConfig;
+    std::optional<const SorobanNetworkConfig> const mSorobanNetworkConfig;
```
marta-lokhova, line 204: I feel a bit uneasy SorobanNetworkConfig lives here: LedgerSnapshot was meant to be simply a ledger state wrapper, that supports both LedgerTxn and BucketListSnapshot underneath. I was hoping to not cache anything at this layer to avoid ledger state inconsistencies and other footguns. The only exception is mLedgerHeader, which has to be modified for validation purposes. SorobanNetworkConfig really belongs to immutable bucketlist snapshots, since the struct is just a different representation of what lives in the bucketlist already. Could we move it there?

src/overlay/OverlayManagerImpl.h
```diff
@@ -184,6 +187,18 @@ class OverlayManagerImpl : public OverlayManager
     RandomEvictionCache<Hash, std::weak_ptr<CapacityTrackedMessage>>
         mScheduledMessages;
 
+    // Ledger snapshot used for background transaction validation. Must be
+    // updated (replaced) after each new ledger via the `updateSnapshots`
+    // function.
+    std::unique_ptr<ExtendedLedgerSnapshot const> mLedgerSnapshot;
```
marta-lokhova, line 193: It's safer to create a new LedgerSnapshot every time you need to use it to avoid caching inconsistencies.

src/overlay/OverlayManagerImpl.h
```diff
@@ -164,6 +165,8 @@ class OverlayManagerImpl : public OverlayManager
     void start() override;
     void shutdown() override;
 
+    void updateSnapshots() override;
```
marta-lokhova, line 168: I don't think this is needed: the underlying bucket snapshot in LedgerSnapshot is managed by LedgerManager already, so if you create a new LedgerSnapshot every time instead of caching it, you'll automatically get the most recent version of the ledger. No need for snapshot mutex either, since bucketlist are already refreshed in a thread-safe manner.

src/herder/TransactionQueue.cpp
```diff
@@ -469,13 +463,11 @@ TransactionQueue::canAdd(
...
     auto txResult = tx->createSuccessResult();
-#ifdef BUILD_TESTS
-    if (!isLoadgenTx)
-#endif
+    if (!isPreValidated)
     {
-        txResult =
-            tx->checkValid(mApp.getAppConnector(), ls, 0, 0,
-                           getUpperBoundCloseTimeOffset(mApp, closeTime));
+        txResult = tx->checkValid(
```
marta-lokhova, line 468: Are there visible performance benefits when doing checkValid vs just calling verifySig on all the signers to populate the existing cache? Reason I'm asking is because skipping checkValid here means that we might be potentially adding invalid transactions to the queue (if ledger state changed while the transaction was being processed). This _can_ be handled, but I wanted to make sure we'd actually get the benefits compared to a much simpler verifySig option. Note that in case of verifySig, we could simply call it in `Peer::recvAuthenticatedMessage` similar how it's done for [SCP](https://github.com/stellar/stellar-core/blob/c1746e764983020a725415b017ba0863e15292ec/src/overlay/Peer.cpp#L976) messages, and then still do checkValid on main thread. Because verifySig uses the global cache, we would just get a cache hit and avoid the extra cost of signature verification.

src/ledger/LedgerStateSnapshot.cpp
```diff
@@ -257,10 +259,66 @@ LedgerSnapshot::load(LedgerKey const& key) const
...
+{
+    releaseAssert(threadIsMain());
+}
+
+ExtendedLedgerSnapshot::ExtendedLedgerSnapshot(AbstractLedgerTxn& ltx,
+                                               AppConnector const& app,
+                                               bool forApply)
+    : LedgerSnapshot(ltx)
+    , mConfig(app.getConfigPtr())
+    , mSorobanNetworkConfig(forApply
```
marta-lokhova, line 279: I think something here isn't right: ExtendedLedgerSnapshot is supposed to be "the canonical representation of ledger state". Other subsystems are meant to use the snapshot and query it for data, such as ledger entries, soroban config, ledger header etc. But here we still have a dependency on LM to maintain correctness of the config. In https://github.com/stellar/stellar-core/issues/4607 I was hoping to remove cached `mSorobanNetworkConfig` completely from LedgerManager, and have it rely on the snapshot instead. The nice thing about having all of the state inside the snapshot is that it's a lot less error-prone; the snapshot will always contain correct read-only state. If some parts of the state are managed by other classes like LedgerManager, that creates a risk of divergence if not updated correctly.

marta-lokhova (changes requested):

src/herder/TransactionQueue.cpp
bboston7: From tracy testing, signature verification makes up 2/3s to 3/4s of `checkValid`. So it's the largest component, but it's also not everything. I'd expect that if we were to scope this down to just signature verifcation, we'd get results that are slightly worse than 2/3s of 3/4s of what we're seeing here, just due to potential  lock contention within the cache and the small possibility for bad luck with eviction (it's a random eviction cache, right?). 

As far as invalid transactions in the queue go, I was under the impression that the validity of transactions in the queue was already "best effort", especially with Soroban where we can't validate much anyway? That's not to say we should make the problem "worse" without a good reason, but surely we already handle invalid transactions in the queue, right?

src/overlay/Peer.h
```diff
@@ -503,12 +503,12 @@ class CapacityTrackedMessage : private NonMovableOrCopyable
 {
     std::weak_ptr<Peer> const mWeakPeer;
     StellarMessage const mMsg;
-    std::optional<Hash> mMaybeHash;
+    std::shared_ptr<Hash const> mMaybeHash;
```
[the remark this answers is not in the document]
bboston7: This helps to avoid copying `Hash` objects between various threads. This way we can just copy the pointer instead.

src/ledger/LedgerStateSnapshot.h
bboston7: Part of the design for this change was to make it impossible for `checkValid` to interact with other subsystems directly (outside of snapshots). I removed `AppConnector` from the validation flow entirely to ensure this was the case. As a result, I folded the config into `ExtendedLedgerSnapshot`, but if we don't like it there than I can manually pass it through through validation flow.

PR #4622, opened 2025-01-13
Bump minimum overlay version to 35

### LINKED ISSUE #4620: Bump minimum overlay version to 35
Now that P22 is live on the network there is no reason to support connections with clients older than v22.0.0.

bboston7:
Closes #4620

This changes bumps the minimum overlay version to match the version in `v22.0.0`. Now that the network has upgraded to protocol 22, there's no reason to keep older clients connected.


SirTyson (approved):
LGTM

PR #4630, opened 2025-01-28
BucketListDB in-memory Buckets

### LINKED ISSUE #3696: Add persistent cache to BucketListDB

SirTyson:
# Description

Partially resolves #3696

This change refactors the `BucketIndex` parts of BucketListDB to be more friendly towards the Hot Archive BucketList. This includes cleanups to metrics, which previously only tracked metrics for the main thread access of the BucketList. Now, both BucketList types and background threads record metrics properly.

Additionally, this change removes the `IndividualIndex` and instead caches small Buckets entirely in-memory so we never read from disk. `RangeIndex` is largely unchanged but has been renamed to `DiskIndex`. 

A follow up PR will add a random eviction cache to the `DiskIndex`. I tried to break this up as much as possible, but it was easiest to do the refactor + in-memory buckets at the same time so I didn't have to refactor `IndividualIndex`.

The `BUCKETLIST_DB_INDEX_CUTOFF` config setting determines the maximum size at which we keep bucket in-memory. I've set this to 250 MB, which is approximately the first 4-5 levels of the BucketList. This increases total memory consumption of stellar-core from 2.2 GB to 3 GB. This seems reasonable, and we could probably go even higher, but I'm holding off for now as the random eviction cache will further increase memory requirements.

# Checklist
- [x] Compiles
- [x] Ran all tests


src/bucket/BucketIndexUtils.h
```diff
@@ -0,0 +1,110 @@
...
+};
+
+class IndexReturnT
+{
+  private:
+    // Payload maps to the possible return states:
+    // CACHE_HIT: IndexPtrT
+    // FILE_OFFSET: std::streamoff
+    // NOT_FOUND: nullopt
+    using PayloadT = std::optional<std::variant<IndexPtrT, std::streamoff>>;
```
dmkozh, line 54: nit: using `std::monostate` in variant would make this type more in line with the enum.

src/bucket/BucketIndexUtils.h
```diff
@@ -0,0 +1,110 @@
...
+    IndexReturnState
+    getState() const
+    {
+        return mState;
+    }
+    IndexPtrT
+    cacheHit() const
+    {
+        releaseAssertOrThrow(mState == IndexReturnState::CACHE_HIT);
+        releaseAssertOrThrow(mPayload);
```
dmkozh, line 81: nit: The assertion should probably be on `std::holds_alternative` instead.

src/bucket/InMemoryIndex.h
```diff
@@ -0,0 +1,116 @@
...
+    // Entries sorted by LedgerKey. INIT/LIVE entries stored as
+    // LedgerEntry, DEADENTRY stored as LedgerKey.
+    std::vector<IndexPtrT> mEntries;
+
+  public:
+    using IterT = std::vector<IndexPtrT>::const_iterator;
+
+    // Insert a LedgerEntry (INIT/LIVE) into the ordered container. Entries must
+    // be ordered.
+    void push_back(BucketEntry const& be);
```
dmkozh, line 31: nit: `pushBack` to follow the style?

src/bucket/SearchableBucketList.cpp
```diff
@@ -241,36 +241,29 @@ SearchableLiveBucketListSnapshot::loadKeysWithLimits(
...
-    {
-        auto timer =
-            mSnapshotManager.recordBulkLoadMetrics("prefetch", inKeys.size())
-                .TimeScope();
-        auto op = loadKeysInternal(inKeys, lkMeter, std::nullopt);
-        releaseAssertOrThrow(op);
-        return std::move(*op);
-    }
-
+    auto timer = getBulkLoadTimer("prefetch", inKeys.size()).TimeScope();
```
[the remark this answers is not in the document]
SirTyson: When we originally did the background snapshotting, metrics weren't thread safe. Marta has since audited and made a couple changes to the library so the metrics should be thread safe.

src/bucket/BucketManager.cpp
```diff
@@ -330,6 +326,24 @@ BucketManager::getMergeTimer()
     return mBucketSnapMerge;
 }
 
+template <class BucketT>
+medida::Meter&
+BucketManager::getBloomMissMeter() const
+{
+    BUCKET_TYPE_ASSERT(BucketT);
+    return mApp.getMetrics().NewMeter(
```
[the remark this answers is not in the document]
SirTyson: Yeah this is fine, we only call this function when we initialize a new `DiskIndex`, which happens very occasionally.

src/bucket/HotArchiveBucketIndex.cpp
```diff
@@ -0,0 +1,79 @@
...
+
+HotArchiveBucketIndex::HotArchiveBucketIndex(
+    BucketManager& bm, std::filesystem::path const& filename, Hash const& hash,
+    asio::io_context& ctx)
+    : mDiskIndex(bm, filename, getPageSize(bm.getConfig(), 0), hash, ctx)
+{
+    ZoneScoped;
+    releaseAssert(!filename.empty());
+
+    // HotArchive only supports disk indexes
```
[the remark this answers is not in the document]
SirTyson: I think this is left over from before I refactored `DiskIndex` and `InMemory` index into separate classes. I agree the assert doesn't make sense and will remove it.

src/bucket/BucketIndexUtils.h
SirTyson: I think I'll keep both, just because the value of `mState` is how the user interacts with the wrapper class.

dmkozh (approved):

dmkozh (approved):

PR #4632, opened 2025-01-29
BucketListDB Random Eviction Cache

### LINKED ISSUE #3696: Add persistent cache to BucketListDB

SirTyson:
# Description

Resolves #3696

Adds a random eviction cache to `LiveBucket` using `DiskIndex`. Currently, I've set the default to cache up to 25% of Bucket entries, which puts total stellar-core memory usage at about 5-6 GB by default.

# Checklist
- [x] Compiles
- [x] Ran all tests

### DIFF
--- a/docs/stellar-core_example.cfg
+++ b/docs/stellar-core_example.cfg
@@ -239,7 +239,8 @@ BUCKETLIST_DB_INDEX_PAGE_SIZE_EXPONENT = 14
 # Percentage of entries cached by BucketListDB when Bucket size is larger
 # than BUCKETLIST_DB_INDEX_CUTOFF. Note that this value does not impact
 # Buckets smaller than BUCKETLIST_DB_INDEX_CUTOFF, as they are always
-# completely held in memory.
+# completely held in memory. Roughly speaking, RAM usage for BucketList
+# cache == BucketListSize * (BUCKETLIST_DB_CACHED_PERCENT / 100).
 BUCKETLIST_DB_CACHED_PERCENT = 25
 
 # BUCKETLIST_DB_INDEX_CUTOFF (Integer) default 250
--- a/docs/stellar-core_example.cfg
+++ b/docs/stellar-core_example.cfg
@@ -235,6 +235,13 @@ MAX_DEX_TX_OPERATIONS_IN_TX_SET = 0
 # 0, indiviudal index is always used. Default page size 16 kb.
 BUCKETLIST_DB_INDEX_PAGE_SIZE_EXPONENT = 14
 
+# BUCKETLIST_DB_CACHED_PERCENT (Integer) default 25
+# Percentage of entries cached by BucketListDB when Bucket size is larger
+# than BUCKETLIST_DB_INDEX_CUTOFF. Note that this value does not impact
+# Buckets smaller than BUCKETLIST_DB_INDEX_CUTOFF, as they are always
+# completely held in memory.
+BUCKETLIST_DB_CACHED_PERCENT = 25
+
 # BUCKETLIST_DB_INDEX_CUTOFF (Integer) default 250
 # Size, in MB, determining whether a bucket should have an individual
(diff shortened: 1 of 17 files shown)


src/bucket/HotArchiveBucketIndex.h
```diff
@@ -74,6 +74,13 @@ class HotArchiveBucketIndex : public NonMovableOrCopyable
         return mDiskIndex.scan(mDiskIndex.begin(), k).first;
     }
 
+    // Hot Archive does not support the cache, so define empty function for
+    // consistency with LiveBucketIndex
+    void
+    maybeAddToCache(std::shared_ptr<HotArchiveBucketEntry const> entry) const
```
dmkozh, line 80: nit: this might be a micro-optimization, but since we don't always put an entry into a cache, we end up creating unnecessary shared ptr copies that are somewhat expensive. I don't see any significant downside to passing a const reference instead.

docs/stellar-core_example.cfg
```diff
@@ -235,6 +235,13 @@ MAX_DEX_TX_OPERATIONS_IN_TX_SET = 0
 # 0, indiviudal index is always used. Default page size 16 kb.
 BUCKETLIST_DB_INDEX_PAGE_SIZE_EXPONENT = 14
 
+# BUCKETLIST_DB_CACHED_PERCENT (Integer) default 25
+# Percentage of entries cached by BucketListDB when Bucket size is larger
+# than BUCKETLIST_DB_INDEX_CUTOFF. Note that this value does not impact
+# Buckets smaller than BUCKETLIST_DB_INDEX_CUTOFF, as they are always
+# completely held in memory.
+BUCKETLIST_DB_CACHED_PERCENT = 25
```
dmkozh, line 243: I wonder how is one supposed to reason about this value. Are the RAM requirements roughly BL size * cached percent? Maybe it's worth adding a comment regarding this.
I'm also curious why do we use percentage here instead of e.g. the number of entries or even the rough RAM boundary.

SirTyson: Yeah, usage is basically BL size * cached percentage. The reason I did percentage is primarily for easier implementation. Max number of items to cache is a poor metric, since we now have high variance in LedgerEntry size because of Soroban entries.

RAM based caps are more useful, but pose some implementation challenges. For example,  suppose we need to cache a large WASM entry. If we use a random eviction cache, we might need to evict many entries to free up enough space for the WASM entry. It is also a little complicated to determine how you allocate RAM among the different buckets. Suppose you have a limit of 1 GB, how do we divy this up among the different buckets? You could do something like each bucket gets a cache size relative to how much of the total BucketList size the bucket makes up. However, we now have to initialize caches wrt total BucketList size, which is challenging since Bucket objects are created individually in a background thread well before they are required.

All of these are probably solvable issues, but I doubt anyone will actually modify this value. In practice we can probably manage this value ourselves as needed by changing the default value in point release should bucketList increase significantly. In practice this ratio is just the upper bound. Especially for smaller buckets, I imagine they will churn before filling up their cache anyway.

dmkozh: > Max number of items to cache is a poor metric, since we now have high variance in LedgerEntry size because of Soroban entries.

But this value is set based on the entry count as well, isn't it? I just think it's much easier to argue about the RAM usage when you deal with a constant number of entries - given that this number is large enough and given the random eviction safe, I think it's safe to just multiply this by the mean entry size to get the RAM estimate. While I agree that this is unlikely to be changed, I find it hard to argue about this value even for ourselves, especially if BL grows and we just forget about this.

dmkozh (approved):

dmkozh (approved):
Forgot to approve this; I think we can change the config semantics later if we want. This still needs to be rebased though.

SirTyson:
> Forgot to approve this; I think we can change the config semantics later if we want. This still needs to be rebased though.

Done

PR #4642, opened 2025-02-06
WIP Add LedgerStateCache

ThomasBrady:
Adds LedgerStateCache to store all **Soroban** entries:
* Populates (`LedgerEntryCache::addEntry`) the cache in `PopulateLedgerCacheWork` (executed in `LedgerManagerImpl::loadLastKnownLedger`). 
* Updates the cache (`LedgerEntryCache:::addEntries`) each ledger from `LedgerManagerImpl::transferLedgerEntriesToBucketList`. 
* Reads the cache (`LedgerEntryCache::readEntry`) in LedgerTxnRoot::Impl:::getNewestVersion.

Notes:
* `addEntry` and `addEntries` are private and only accessible from `friend` classes `LedgerManagerImpl` and `PopulateLedgerCacheWork`
* `addEntry` and `addEntries` acquire a unique lock on the cache
* Currently, the `LeddgerStateCache` only supports contract entries. To access them in the index, I've added `Bucket / Index::getContractEntryRange`.
* The `LedgerStateCache` is enabled via `Config::IN_MEMORY_SOROBAN_STATE_FOR_TESTING` (default `true`)

# Description

Resolves #X

# Checklist
- [ ] Compiles
- [ ] Ran all tests

### DIFF
--- a/src/catchup/PopulateLedgerCacheWork.cpp
+++ b/src/catchup/PopulateLedgerCacheWork.cpp
@@ -1,0 +1,115 @@
+// Copyright 2025 Stellar Development Foundation and contributors. Licensed
+// under the Apache License, Version 2.0. See the COPYING file at the root
+// of this distribution or at http://www.apache.org/licenses/LICENSE-2.0
+
+#include "catchup/PopulateLedgerCacheWork.h"
+#include "bucket/BucketInputIterator.h"
+#include "bucket/BucketManager.h"
+#include "bucket/LiveBucket.h"
+#include "bucket/LiveBucketIndex.h"
+#include "bucket/LiveBucketList.h"
+#include "ledger/LedgerManager.h"
+#include "ledger/LedgerStateCache.h"
+#include "xdrpp/printer.h"
+
+namespace stellar
+{
+PopulateLedgerCacheWork::PopulateLedgerCacheWork(Application& app)
+    : Work(app, "populate-ledger-cache", BasicWork::RETRY_NEVER)
+    , mBucketsToProcess{}
+    , mBucketToProcessIndex(0)
+    , mDeadKeys{}
+{
+    auto& bm = mApp.getBucketManager();
+    auto& bl = bm.getLiveBucketList();
+    int counter = 0;
+    for (uint32_t i = 0; i < LiveBucketList::kNumLevels; ++i)
+    {
+        auto const& level = bl.getLevel(i);
+        for (auto const& bucket : {level.getCurr(), level.getSnap()})
+        {
+            mBucketsToProcess.push_back(bucket);
+            CLOG_DEBUG(Ledger, "Adding bucket {} to mBucketsToProcess[{}]",
+                       xdr::xdr_to_string(bucket->getHash()), counter);
+            counter++;
+        }
+    }
+}
+// TODO possibly refactor BucketApplicator to have modular
+// application logic for:
+// - selecting entries to apply (e.g. offers, contract entries)
+// - applying the entry (e.g. add to ltx or ledgerStateCache)
+// It seems the main motivation for BucketApplicator is to limit
+// applications to batches of LEDGER_ENTRY_BATCH_COMMIT_SIZE entries.
+// No such limit is necessary here, but it would deduplicate
+// some shared logic.
+BasicWork::State
+PopulateLedgerCacheWork::advance()
+{
+    CLOG_INFO(Ledger, "Added bucket {}/{} to LedgerStateCache",
+              mBucketToProcessIndex, mBucketsToProcess.size());
+    mBucketToProcessIndex++;
+    if (mBucketToProcessIndex >= mBucketsToProcess.size())
+    {
+        mBucketsToProcess.clear();
+        return State::WORK_SUCCESS;
+    }
+    return State::WORK_RUNNING;
+}
+
+BasicWork::State
+PopulateLedgerCacheWork::doWork()
+{
+    auto const& bucket = mBucketsToProcess.at(mBucketToProcessIndex);
+    // Hacky way to skip empty buckets
+    if (!bucket || bucket->getFilename().empty())
+    {
+        CLOG_DEBUG(Ledger,
+                   "PopulateLedgerCacheWork: Advancing past empty bucket {}",
+                   mBucketToProcessIndex);
+        return advance();
+    }
+
+    auto contractEntryRange = bucket->getContractEntryRange();
+    auto ledgerStateCache = mApp.getLedgerManager().getLedgerStateCache();
+    if (ledgerStateCache && contractEntryRange)
+    {
+        auto cache = ledgerStateCache.value();
+        auto [lowerBound, upperBound] = *contractEntryRange;
+        for (LiveBucketInputIterator iter(bucket);
+             iter && iter.pos() < upperBound; ++iter)
+        {
+            if (iter.pos() < lowerBound)
+            {
+                iter.seek(lowerBound);
+            }
+            BucketEntry const& entry = *iter;
+            if (entry.type() == LIVEENTRY || entry.type() == INITENTRY)
+            {
+                auto const& e = entry.liveEntry();
+                auto const& k = LedgerEntryKey(e);
+                // If the key is not in the dead keys set and not already in the
+                // cache, add it.
+                if (mDeadKeys.find(k) == mDeadKeys.end() && !cache->getEntry(k))
+                {
+                    cache->addEntry(e);
+                }
+            }
+            else if (entry.type() == DEADENTRY)
+            {
+                if (!cache->getEntry(entry.deadEntry()))
+                {
+                    mDeadKeys.insert(entry.deadEntry());
+                }
+            }
+            else
+            {
+                releaseAssert(false);
+                continue;
+            }
+        }
+    }
+    return advance();
+}
+
+}
--- a/src/ledger/LedgerManagerImpl.cpp
+++ b/src/ledger/LedgerManagerImpl.cpp
@@ -7,6 +7,7 @@
 #include "bucket/HotArchiveBucketList.h"
 #include "bucket/LiveBucketList.h"
 #include "catchup/AssumeStateWork.h"
+#include "catchup/PopulateLedgerCacheWork.h"
 #include "crypto/Hex.h"
 #include "crypto/KeyUtils.h"
 #include "crypto/SHA.h"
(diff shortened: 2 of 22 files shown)


src/ledger/LedgerStateCache.h
```diff
@@ -0,0 +1,82 @@
...
+#include <shared_mutex>
+#include <unordered_set>
+
+namespace stellar
+{
+// TODO Add ifdef
+struct LedgerEntryHash
+{
+    size_t operator()(LedgerEntry const& entry) const;
+};
```
SirTyson, line 19: This is a bit of a dangerous struct, since it's not obvious that this doesn't hash the whole key. I'd change the name to `LedgerEntryKeyHash` and maybe make it a public member of `LedgerStateCache`.

src/ledger/LedgerStateCache.h
```diff
@@ -0,0 +1,82 @@
...
+{
+    enum class Mode
+    {
+        ALL_ENTRIES,
+        SOROBAN_ONLY
+    };
+
+  private:
+    Mode mMode;
+    std::unordered_set<LedgerEntry, LedgerEntryHash, LedgerEntryEqual> mState;
```
SirTyson, line 43: C++ 20 supports heterogenous unordered_set, but alas, C++17 does not. We might wanna benchmark this implementation where we create throw away dummy keys against just a plain old std::set with heterogenous lookups, set may be better.

src/ledger/LedgerStateCache.h
```diff
@@ -0,0 +1,82 @@
...
+    // LedgerEntry in initEntries and liveEntries are added to the cache,
+    // while LedgerEntry in deadEntries are removed from the cache.
+    void addEntries(std::vector<LedgerEntry> const& initEntries,
+                    std::vector<LedgerEntry> const& liveEntries,
+                    std::vector<LedgerKey> const& deadEntries);
+
+    // Add a single entry to the cache
+    // Acquires a unique lock on the cache. Called
+    // when populating the cache.
+    void addEntry(LedgerEntry const& entry);
```
SirTyson, line 58: Nit: change name to populate so it's a little more distinct from `addEntries`.

src/ledger/LedgerStateCache.h
```diff
@@ -0,0 +1,82 @@
...
+    Mode mMode;
+    std::unordered_set<LedgerEntry, LedgerEntryHash, LedgerEntryEqual> mState;
+    mutable std::shared_mutex mMutex;
+
+    // Add entries to the cache
+    // Acquires a unique lock on the cache. Should be called
+    // once per ledger, when transfering entries to the bucket list.
+    // LedgerEntry in initEntries and liveEntries are added to the cache,
+    // while LedgerEntry in deadEntries are removed from the cache.
+    void addEntries(std::vector<LedgerEntry> const& initEntries,
```
SirTyson, line 51: Nit: `addBucketEntries` for better symmetry, since this takes the same argument.

src/ledger/LedgerStateCache.h
```diff
@@ -0,0 +1,82 @@
...
+    enum class Mode
+    {
+        ALL_ENTRIES,
+        SOROBAN_ONLY
+    };
+
+  private:
+    Mode mMode;
+    std::unordered_set<LedgerEntry, LedgerEntryHash, LedgerEntryEqual> mState;
+    mutable std::shared_mutex mMutex;
```
SirTyson, line 44: I _think_ it's probably better to store `std::shared_ptr<LedgerEntry const>` in this set. That way, we can return a shared_ptr and never have to copy actual data after populating the cache.

src/ledger/LedgerStateCache.cpp
```diff
@@ -0,0 +1,122 @@
...
+    const LedgerKey& key = LedgerEntryKey(entry);
+    return std::hash<stellar::LedgerKey>()(key);
+}
+
+std::optional<LedgerEntry>
+LedgerStateCache::getEntry(LedgerKey const& key) const
+{
+    if (!supportedKeyType(key.type()))
+    {
+        return std::nullopt;
```
SirTyson, line 27: This should be `releaseAssertOrThrow`. Otherwise, this return value is ambiguous, did we just look up a wrong type, or did we look up the correct type, but the key does not exist?

src/ledger/LedgerStateCache.cpp
```diff
@@ -0,0 +1,122 @@
...
+    }
+
+    std::shared_lock lock(mMutex);
+    if (mState.empty())
+    {
```
SirTyson, line 32: This should also be a releaseAssertOrThrow for the same reason.

src/ledger/LedgerStateCache.cpp
```diff
@@ -0,0 +1,122 @@
...
+        return std::optional<LedgerEntry>(*it);
+    }
+    else
+    {
+        // TODO potentially return an optional if we cannot find the entry, but
+        // the cache should contain all entries that are in the ledger so
+        // perhaps a runtime error is appropriate. For now we will just return
+        // nullopt.
+        CLOG_ERROR(Ledger, "LedgerStateCache does not contain entry for key {}",
```
SirTyson, line 48: This shouldn't be an error. Since we cache everything there are no "misses" such that returning null indicates the entry does not exist.

src/ledger/LedgerStateCache.cpp
```diff
@@ -0,0 +1,122 @@
...
+                       entry.data.type());
+            continue;
+        }
+        mState.erase(entry); // Remove old version if it exists
+        mState.emplace(entry);
+    }
+}
+
+void
+LedgerStateCache::addEntry(LedgerEntry const& entry)
```
SirTyson, line 86: I think I would like to see some stronger safety around querying an incomplete cache. Currently, there's nothing preventing a background thread from accidentally querying an in-progress cache. It would be nice to have some asserts and throw if we do lookups before we've finished construction.

src/ledger/LedgerStateCache.cpp
```diff
@@ -0,0 +1,122 @@
...
+{
+    std::unique_lock lock(mMutex);
+    if (!supportedKeyType(entry.data.type()))
+    {
+        CLOG_DEBUG(Ledger, "Skipping unsupported key type {}",
+                   entry.data.type());
+        return;
+    }
+    mState.erase(entry);
+    mState.emplace(entry);
```
SirTyson, line 96: since this is population, we should just call emplace and assert it actually got inserted. There should be no duplicate inserts on construction.

src/ledger/LedgerManagerImpl.cpp
```diff
@@ -385,6 +400,23 @@ LedgerManagerImpl::loadLastKnownLedger(bool restoreBucketlist)
             // Work should only fail during graceful shutdown
             releaseAssertOrThrow(mApp.isStopping());
         }
+        if (mLedgerStateCache)
+        {
+            auto populateLedgerCacheWork =
+                mApp.getWorkScheduler().executeWork<PopulateLedgerCacheWork>();
```
SirTyson, line 406: I think this would be better as a step in `AssumeStateWork`. That's where we do all our "in-memory matches current state" work, such as setting the BucketList, indexing buckets, etc.

src/ledger/LedgerTxn.cpp
```diff
@@ -534,6 +535,8 @@ LedgerTxn::Impl::commitChild(EntryIterator iter,
                              RestoredKeys const& restoredKeys,
                              LedgerTxnConsistency cons) noexcept
 {
+    // We will want to acquire the write lock to the cache in this function.
```
SirTyson, line 538: Remove comment

src/catchup/PopulateLedgerCacheWork.cpp
```diff
@@ -0,0 +1,132 @@
...
+    int counter = 0;
+    for (uint32_t i = 0; i < LiveBucketList::kNumLevels; ++i)
+    {
+        auto const& level = bl.getLevel(i);
+        for (auto const& bucket : {level.getCurr(), level.getSnap()})
+        {
+            mBucketsToProcess.push_back(bucket);
+            CLOG_DEBUG(Ledger, "Adding bucket {} to mBucketsToProcess[{}]",
+                       xdr::xdr_to_string(bucket->getHash()), counter);
+            counter++;
```
SirTyson, line 34: Nit: ++counter

src/catchup/PopulateLedgerCacheWork.cpp
```diff
@@ -0,0 +1,132 @@
...
+    auto& bl = bm.getLiveBucketList();
+    int counter = 0;
+    for (uint32_t i = 0; i < LiveBucketList::kNumLevels; ++i)
+    {
+        auto const& level = bl.getLevel(i);
+        for (auto const& bucket : {level.getCurr(), level.getSnap()})
+        {
+            mBucketsToProcess.push_back(bucket);
+            CLOG_DEBUG(Ledger, "Adding bucket {} to mBucketsToProcess[{}]",
+                       xdr::xdr_to_string(bucket->getHash()), counter);
```
SirTyson, line 33: Nit: use `binToHex()` or `hexAbbrev()` instead of `xdr_to_string`

SirTyson (changes requested):
I think in general this looks good! I think we need explicit cache tests though. Specifically, we need population tests (which isn't covered in any unit tests currently I think). While we might have transitive coverage of actual cache usage, a few explicit unit tests would be useful as well.

I have two sorta high level comments. First, given that this is our base apply time cache and we're making protocol assumptions based on its efficiency, we need to be memory conscious and worry about things like copy semantics, probably preffering const shared_ptrs where we can.

2nd, I think we need to be more defensive in general about who has access with this cache. Many different threads and subsystems access state now after Marta'a background apply changes. However, each thread/subsystem has different expectations wrt that that access looks like. This cache in particular is an apply-time only cache, so I think it should be very limited to the apply time thread. I think to accomplish this we should have `LedgerTxnRoot` hold a `unique_ptr` to the cache instead of spreading around shared_ptrs. I think this is probably easiest if you construct a uniqe_ptr of the cache via your work class, then call some function like `LedgerManager::setApplyTimeCache(unique_ptr&&)`. LedgerManager will then set the cache pointer in the ltx root. I think this is important as we want to make sure that only ltx has access to the cache. We might need to add an invalidation function too, such that the `addbucketEntriesToCache` function is a member of `LedgerTxnRoot` if that makes sense. This would also allow us to get rid of locks, since we'd be guaranteed everything is happening from the apply time thread. This might be different than what was discussed in the original design doc, but I think makes more sense after Marta's changes, where there is a more clear distinction between "apply time only state" and "state snapshots used by overlay and other subsystems"

PR #4643, opened 2025-02-07
move template class member to correct TU for explicit instantiation

graydon:
During recent refactoring a single method of the `template <class BucketT> class SearchableBucketListSnapshotBase` -- the method `SearchableBucketListSnapshotBase<BucketT>::loadKeysInternal` -- got left behind in `SearchableBucketList.cpp` while the rest of the methods got moved to `BucketListSnapshotBase.cpp`.

Unfortunately all explicit template instantiations of `SearchableBucketListSnapshotBase` occur in `BucketListSnapshotBase.cpp` and they only cause instantiation of methods _defined in_ the same TU. So the method didn't get instantiated. At least on g++.

This was covered up in CI because enabling extrachecks winds up triggering instantiations anyways. But if you turn off extrachecks, master doesn't currently compile on g++.

This fixes the bug by moving the method to the right TU.

### DIFF
--- a/src/bucket/BucketListSnapshotBase.cpp
+++ b/src/bucket/BucketListSnapshotBase.cpp
@@ -120,6 +120,41 @@ SearchableBucketListSnapshotBase<BucketT>::load(LedgerKey const& k) const
 
 template <class BucketT>
 std::optional<std::vector<typename BucketT::LoadT>>
+SearchableBucketListSnapshotBase<BucketT>::loadKeysInternal(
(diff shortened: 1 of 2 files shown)


SirTyson (approved):
LGTM, thanks for tracking this down! Apologies for introducing some templated complexity with this one...

graydon:
@SirTyson I appreciate your ongoing efforts to keep well-factored what has turned into a real house of mirrors! At some point when the dust has settled it might be useful to draw an ascii-art diagram of all the classes in bucket/ just to help orient future readers :)

PR #4645, opened 2025-02-11
Add function to allow checking thread type

### LINKED ISSUE #4613: Extending `threadIsMain()` assert
Given the amount of parallelism being added to core with #4543, we now have many different types of threads responsible for different tasks. Binary asserts of `!threadIsMain()` and `threadIsMain()` no longer hold as much meaning, and we've had to loosen our thread asserts in a lot of places. We should extend the `threadIsMain` assert to all different "thread types," such as overlay thread, apply thread, etc. This shouldn't be too difficult. For each thread type we cal declare a 0 initialized `thread::id` like we currently do for `mainThread`, then we can initialize the value when spinning up the thread. It may also be useful to rename the main thread to something like "consensus thread" or "scheduling thread" so it carries a little more meaning.

bboston7:
Closes #4613

This change adds a function `threadIsType` function to `Application` that allows a thread to check its type. This is intended to support more detailed assertions than the usual `releaseAssert(threadIsMain())` assertions we're currently using everywhere.

The implementation differs a bit from the proposed solution in #4613 as it uses a mapping in `ApplicationImpl` to track thread types, rather than using static variables. I chose this approach because as far as I can tell, it's not possible to assign a thread an id.  Given that, `ApplicationImpl` would need to set these variables in its constructor, and the variables would hold meaningless values prior to that. I figure it's safer to ensure that thread types can only be reasoned about after the creation of the threads themselves in `ApplicationImpl`'s constructor. However, if it's important that thread types be reasoned about without an `Application` or `AppConnector`, then I'm open to changing the design.

# Checklist
- [x] Compiles
- [x] Ran all tests

### DIFF
--- a/src/main/Application.h
+++ b/src/main/Application.h
@@ -164,6 +164,16 @@ class Application
         APP_NUM_STATE
     };
 
+    // Types of threads that may be running
+    enum class ThreadType
+    {
+        MAIN,
+        WORKER,
+        EVICTION,
+        OVERLAY,
+        APPLY
+    };
+
     virtual ~Application(){};
 
     virtual void initialize(bool createNewDB, bool forceRebuild) = 0;
@@ -329,6 +339,9 @@ class Application
 
         return ret;
     }
+
+    // Returns true iff the calling thread has the same type as `type`
+    virtual bool threadIsType(ThreadType type) const = 0;
 
     virtual AppConnector& getAppConnector() = 0;
 
--- a/src/main/ApplicationImpl.cpp
+++ b/src/main/ApplicationImpl.cpp
@@ -160,11 +160,14 @@ ApplicationImpl::ApplicationImpl(VirtualClock& clock, Config const& cfg)
     releaseAssert(mConfig.WORKER_THREADS > 0);
     releaseAssert(mEvictionIOContext);
 
+    mThreadTypes[std::this_thread::get_id()] = ThreadType::MAIN;
+
     // Allocate one thread for Eviction scan
     mEvictionThread = std::thread{[this]() {
         runCurrentThreadWithMediumPriority();
         mEvictionIOContext->run();
     }};
+    mThreadTypes[mEvictionThread->get_id()] = ThreadType::EVICTION;
 
     --t;
 
@@ -174,6 +177,7 @@ ApplicationImpl::ApplicationImpl(VirtualClock& clock, Config const& cfg)
             runCurrentThreadWithLowPriority();
             mWorkerIOContext.run();
         }};
+        mThreadTypes[thread.get_id()] = ThreadType::WORKER;
         mWorkerThreads.emplace_back(std::move(thread));
     }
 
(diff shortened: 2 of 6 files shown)


graydon (approved):

src/main/ApplicationImpl.cpp
```diff
@@ -1197,6 +1203,14 @@ ApplicationImpl::getMetrics()
     return *mMetrics;
 }
 
+bool
+ApplicationImpl::threadIsType(ThreadType type) const
```
SirTyson, line 1207: I don't think this should assert. For example, we currently don't always initialize a LEDGER_CLOSE thread since it's behind an experimental flag, but something like `releaseAssert(!isLedgerCloseThread)` should probably be valid whether the flag is set or not.

SirTyson (changes requested):

src/main/ApplicationImpl.cpp
graydon: Oh, yeah, this is also true!

src/main/Application.h
```diff
@@ -164,6 +164,16 @@ class Application
...
     };
 
+    // Types of threads that may be running
+    enum class ThreadType
+    {
+        MAIN,
+        WORKER,
+        EVICTION,
+        OVERLAY,
+        LEDGER_CLOSE
```
graydon, line 174: In conversation with @SirTyson a few minutes ago we were trying to figure out terminology that works best for differentiating the state-bits being touched by the "close thread" vs. the lagging LCL, and we came to the conclusion that a good naming scheme would be to keep LCL as the lagging state (it's already used everywhere) and just change the _new_ concept of the database-and-buckets (which can get ahead of LCL) to be the last _applied_ ledger. So "close" or "closed" terminology continues to mean "data fully published back to the main thread state" and "apply" is reserved for the actions of this extra thread. And the thread doing it is therefore the "apply" thread, not the "close" thread. (And I will follow up with a few method-renames inside `LedgerManagerImpl` to reflect this also). 
```suggestion
        APPLY
```

src/main/ApplicationImpl.cpp
```diff
@@ -174,19 +177,22 @@ ApplicationImpl::ApplicationImpl(VirtualClock& clock, Config const& cfg)
...
         // Keep priority unchanged as overlay processes time-sensitive tasks
         mOverlayThread = std::thread{[this]() { mOverlayIOContext->run(); }};
+        mThreadTypes[mOverlayThread->get_id()] = ThreadType::OVERLAY;
     }
 
     if (mConfig.parallelLedgerClose())
     {
         mLedgerCloseThread =
             std::thread{[this]() { mLedgerCloseIOContext->run(); }};
+        mThreadTypes[mLedgerCloseThread->get_id()] = ThreadType::LEDGER_CLOSE;
```
graydon, line 195:
```suggestion
        mThreadTypes[mLedgerCloseThread->get_id()] = ThreadType::APPLY;
```

graydon (changes requested):

src/main/ApplicationImpl.cpp
bboston7: That's not what this is asserting. Your example would be fine here. The assertion is checking that the calling thread's ID is in the map. This ensures that the map has been properly constructed. It wouldn't be possible for a nonexistent thread to trigger this.

That is to say, this is a map from thread IDs to thread types, not thread types to thread ids. It should contain all running thread IDs. if it's missing one, then that's a bug.

Am I missing something?

SirTyson: Shoot, ya you're right, sorry I flipped this key-value in my head for some reason/

SirTyson (approved):
LGTM, can you squash?

bboston7:
> can you squash?

Done

graydon (approved):

PR #4647, opened 2025-02-12
Rename various close things to apply, add comment.

graydon:
This does a few things:
  1. Adds a big block comment to LedgerManager.h explaining the flow of ledgers through LedgerApplyManager and LedgerManager in more detail, as well as showing/describing the threading behaviour (assuming parallel apply).
  2. Renames various bits named "close" to "apply" when they are things-done-by-the-apply-thread or variables-touched-by-the-apply-thread, in order to more clearly differentiate "close" stuff (on main thread, tracking LCL) from "apply" stuff (on apply thread, possibly ahead of LCL).
  3. Partitions state between two structs in the LM, one for apply-state and one for LCL state.
  4. Recycles the output-from-the-apply-phase struct to _store_ the LCL state, since they had the same content.
  5. Moves the ledger metrics into a sub-struct just so they're a little less cluttered.
  6. Renames a few of the later-stages-of-apply/close from generic terms like `ledgerApplied` to more specific terms like `sealLedgerTxnAndStoreInBucketsAndDB`. It's a mouthful but you at least don't have to go re-read the function body each time you see a call to remember what it does.

The change is _nearly_ a no-op. Functionality-wise I only added two asserts that ought to (and appear to) always be true.

### DIFF
--- a/src/ledger/LedgerManager.h
+++ b/src/ledger/LedgerManager.h
@@ -243,18 +243,18 @@ class LedgerManager
 
     virtual Resource maxLedgerResources(bool isSoroban) = 0;
     virtual Resource maxSorobanTransactionResources() = 0;
-    virtual void updateNetworkConfig(AbstractLedgerTxn& ltx) = 0;
+    virtual void updateSorobanNetworkConfigForApply(AbstractLedgerTxn& ltx) = 0;
     // Return the network config for Soroban.
     // The config is automatically refreshed on protocol upgrades.
     // Ledger txn here is needed for the sake of lazy load; it won't be
     // used most of the time.
-    virtual SorobanNetworkConfig const& getSorobanNetworkConfigReadOnly() = 0;
+    virtual SorobanNetworkConfig const& getLastClosedSorobanNetworkConfig() = 0;
     virtual SorobanNetworkConfig const& getSorobanNetworkConfigForApply() = 0;
 
-    virtual bool hasSorobanNetworkConfig() const = 0;
+    virtual bool hasLastClosedSorobanNetworkConfig() const = 0;
 
 #ifdef BUILD_TESTS
-    virtual SorobanNetworkConfig& getMutableSorobanNetworkConfig() = 0;
+    virtual SorobanNetworkConfig& getMutableSorobanNetworkConfigForApply() = 0;
     virtual std::vector<TransactionMetaFrame> const&
     getLastClosedLedgerTxMeta() = 0;
     virtual void storeCurrentLedgerForTest(LedgerHeader const& header) = 0;
--- a/src/ledger/LedgerManager.h
+++ b/src/ledger/LedgerManager.h
@@ -148,7 +148,7 @@ class SorobanMetrics;
 // ledger" is fairly atomic, and there is no distinction made between "apply"
 // and "close". But now that we have two threads, at least _within these
 // classes_ we endeavour to use the term "close" and "last closed" to refer to
-// the actions taken and varaibles updated by the main thread _after_ apply, and
+// the actions taken and variables updated by the main thread _after_ apply, and
 // the term "apply" to refer to actions taken and variables updated by the apply
 // thread.
 class LedgerManager
--- a/src/ledger/LedgerManager.h
+++ b/src/ledger/LedgerManager.h
@@ -25,7 +25,7 @@ class SorobanMetrics;
 // A "logical" ledger takes on more structure as it flows through the system. It
 // begins as LedgerCloseData, which is just a txset and hash reference to a
 // ledger, coming out of SCP. But during apply it is transformed into an
-// ApplyLedgerOutput, which carries with it a full shapshot of the bucketlist.
+// ApplyLedgerOutput, which carries with it a full snapshot of the bucketlist.
 //
 // The tricky part of this flow involves the fact that work is being done by
 // _two threads_: on the left there is work by the main thread, and on the right
--- a/src/ledger/LedgerManager.h
+++ b/src/ledger/LedgerManager.h
@@ -16,27 +16,141 @@ class LedgerCloseData;
 class Database;
 class SorobanMetrics;
 
-/**
- * LedgerManager maintains, in memory, a logical pair of ledgers:
- *
- *     1. The "current ledger", to which the transactions currently accumulating
- *        in memory (in the Herder) will be applied on the next ledger-close.
- *
- *     2. The "last-closed ledger" (LCL), which is the result of the most recent
- *        ledger-close. This ledger is the state most recently committed to the
- *        SQL store, is immutable, and is the basis for the transactions applied
- *        in "current". In other words, "current" begins as a copy of LCL and
- *        then is mutated by the transactions applied to it.
- *
- * LedgerManager also coordinates the incremental advance of these ledgers: the
- * process of closing ledgers, turning the current ledger into the next LCL, and
- * producing a new current ledger.
- *
- * Finally, LedgerManager triggers and responds to boundary conditions in the
- * life cycle of the application that "suddenly" alter the notions of "current"
- * and "LCL": process startup, desynchronization and resynchronization of the
- * current process with the rest of the network.
- */
+// This diagram provides a schematic of the flow of (logical) ledgers coming in
+// from the SCP-and-Herder consensus complex, passing through the
+// LedgerApplyManager (LAM) and LedgerManager (LM), and finally coming to rest
+// in the "last closed ledger" (LCL) variables in the LM. These will
+// subsequently be published to history but that's not relevant to this diagram.
+//
+// A "logical" ledger takes on more structure as it flows through the system. It
+// begins as LedgerCloseData, which is just a txset and hash reference to a
+// ledger, coming out of SCP. But during apply it is transformed into an
+// ApplyLedgerOutput, which carries with it a full shapshot of the bucketlist.
+//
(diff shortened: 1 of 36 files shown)


graydon:
@SirTyson I think this is about as far as I want to take this for now. We can't, unfortunately, assert on various functions being-the-apply-thread because we allow apply to happen from either main _or_ apply. I guess one option would be to store which thread-type _started_ each apply call and then continue to check throughout that the same thread is continuing to make calls that belong to apply?

(Also curiously now that I have the state more clearly partitioned, I am not really sure the mutex on the LM is doing anything. It seems to be protecting apply from itself, and nothing else. But maybe? Anyway I did not change it.)

bboston7:
Also, out of curiosity, did you use a tool to generate those ASCII diagrams?

src/ledger/LedgerManager.h
```diff
@@ -16,27 +16,141 @@ class LedgerCloseData;
...
+// LCL variables, are for the main thread. One of the LCL state variables is a
+// snapshot of the bucketlist. The bucketlist snapshots are immutable and
+// therefore trivially threadsafe, but may be lagging behind the newest
+// bucketlist formed by the apply thread.
+//
+// In more precise terms, 4 points on the diagram are labeled: L, Q, A, and LCL.
+// These are points where we can identify and relate some state variables and an
+// invariant order between them:
+//
+//     L -- LedgerApplyManagerImpl::mLargestLedgerSeqHeard tracks this, it is
```
SirTyson, line 94: Nit: I think changing L -> H makes a little more sense to me. This disambiguates it more with LCL and also gives us:
H -> (largest) Heard
Q -> (largest) Queued
A -> Applied
LCL -> Last Closed (which should probably really be C, but LCL is maintained for legacy reasons).

src/ledger/LedgerManagerImpl.cpp
```diff
@@ -388,11 +394,12 @@ LedgerManagerImpl::loadLastKnownLedger(bool restoreBucketlist)
     }
 
     // Step 4. Restore LedgerManager's internal state
```
SirTyson, line 396: Nit: "Restore LedgerManager's LCL state" now that the LedgerManager technically stores different state variables.

src/ledger/LedgerManagerImpl.cpp
```diff
@@ -993,22 +1000,22 @@ LedgerManagerImpl::closeLedger(LedgerCloseData const& ledgerData,
...
+        *ledgerData.getExpectedHash() != appliedLedgerState.ledgerHeader.hash)
     {
         throw std::runtime_error("Local node's ledger corrupted during close");
     }
 
     if (mMetaStream || mMetaDebugStream)
     {
         releaseAssert(ledgerCloseMeta);
-        ledgerCloseMeta->ledgerHeader() = closeLedgerResult.ledgerHeader;
+        ledgerCloseMeta->ledgerHeader() = appliedLedgerState.ledgerHeader;
```
SirTyson, line 1018: Nit: Maybe `ledgerCloseMeta` -> `ledgerApplyMeta`. I know the XDR struct is called `ledgerCloseMeta` itself, but this is really the meta from the current apply ledger, not the LCL ledger, which this kinda sounds like at the moment. But maybe it's more confusing to not align with the XDR struct name here, I'm not sure

src/ledger/LedgerManagerImpl.cpp
```diff
@@ -1073,10 +1080,10 @@ LedgerManagerImpl::closeLedger(LedgerCloseData const& ledgerData,
     // Invoke completion handler on the _main_ thread: kick off publishing,
     // cleanup bucket files, notify herder to trigger next ledger
     auto completionHandler = [this, ledgerSeq, calledViaExternalize, ledgerData,
```
SirTyson, line 1082: Perhaps a more descriptive name/comment here would be nice, like `advanceLCLtoAppliedLedgerHandler` or something along those lines.

src/ledger/LedgerManagerImpl.h
```diff
@@ -42,58 +42,64 @@ class BasicWork;
...
-    struct CloseLedgerOutput
+    Application& mApp;
+    std::unique_ptr<XDROutputFileStream> mMetaStream;
+    std::unique_ptr<XDROutputFileStream> mMetaDebugStream;
+    std::weak_ptr<BasicWork> mFlushAndRotateMetaDebugWork;
+    std::filesystem::path mMetaDebugPath;
+
+  private:
+    // Output of the apply process, also what gets held as "LCL".
+    struct LedgerState
```
SirTyson, line 53: Good idea, I like the inner struct breakdown in the interim of hopefully making this class less icky.

src/ledger/LedgerManagerImpl.h
```diff
@@ -42,58 +42,64 @@ class BasicWork;
...
-
-    // Latest Soroban config during apply (should not be used outside of
-    // application, as it may be in half-valid state). Note that access to this
-    // variable is not synchronized, since it should only be used by one thread
-    // (main or ledger close).
-    std::shared_ptr<SorobanNetworkConfig> mSorobanNetworkConfigForApply;
-
-    // Cache most recent HAS, accessible only from main thread
-    HistoryArchiveState mLastClosedLedgerHAS;
+    struct LedgerMetrics
```
SirTyson, line 73: Nit: `LedgerApplyMetrics`, as these are only used by the apply thread

src/ledger/LedgerManagerImpl.h
```diff
@@ -42,58 +42,64 @@ class BasicWork;
...
+        medida::Counter& mSorobanTransactionApplySucceeded;
+        medida::Counter& mSorobanTransactionApplyFailed;
+        medida::Meter& mMetaStreamBytes;
+        medida::Timer& mMetaStreamWriteTime;
+        LedgerMetrics(medida::MetricsRegistry& registry);
+    };
+
+    // This state is private to the apply thread and holds work-in-progress
+    // that gets accessed via the AppConnector, from inside transactions.
+    ApplyState mApplyState;
```
SirTyson, line 93: Might be nice to make access to this a `getter` wrapper function that asserts threadIsApply. That might be too much work now though because we'd have to pass in the CFG to see if background apply is enabled or not, but would be nice to see in the future.

src/ledger/LedgerManagerImpl.h
```diff
@@ -42,58 +42,64 @@ class BasicWork;
...
 
+    // Cached LCL state output from last apply (or loaded from DB on startup).
+    LedgerState mLastClosedLedgerState;
```
SirTyson, line 96: Same here, a getter wrapper that assert threadIsMain might be helpful

SirTyson (approved):
Overall some helpful documentation! A few nits on naming and some possible suggestions for better thread asserts, but good as is regardless.

src/ledger/LedgerManager.h
graydon: done

src/ledger/LedgerManagerImpl.cpp
graydon: done

src/ledger/LedgerManagerImpl.cpp
graydon: yeah I considered it but as with LedgerCloseData I decided this terminology has bled too far outside of LM to maintain the apply-vs-close distinction there. I'd like to leave this as is.

src/ledger/LedgerManagerImpl.cpp
graydon: done

src/ledger/LedgerManagerImpl.h
graydon: I wanted to, but apply can happen on main thread of apply thread, even when background apply is enabled. This is still fairly messy and we might want to prohibit both possibilities, but as of today it's how it is.

src/ledger/LedgerManagerImpl.h
graydon: this one I can do. adding.

src/ledger/LedgerManagerImpl.h
graydon: _Mostly_ .. at least enough to pretend. There's a little main thread action but it's minor. Changed.

src/ledger/LedgerManagerImpl.h
```diff
@@ -42,58 +42,64 @@ class BasicWork;
...
+        medida::Timer& mMetaStreamWriteTime;
+        LedgerApplyMetrics(medida::MetricsRegistry& registry);
+    };
+
+    // This state is private to the apply thread and holds work-in-progress
+    // that gets accessed via the AppConnector, from inside transactions.
+    ApplyState mApplyState;
 
+    // Cached LCL state output from last apply (or loaded from DB on startup).
+    LedgerState mLastClosedLedgerState;
```
SirTyson, line 96: If we wanted to get _really_ pedantic, we could do

```
class LedgerStateGetter {
    LedgerState state;

public:
    LedgerState& get() {
        assert(main);
        return state;
    }

} mLastClosedLedgerState;
```

but I don't feel too strongly.

SirTyson (approved):
LGTM

src/ledger/LedgerManagerImpl.h
graydon: eh .. it gets constructed on the apply thread and passed back to main. I think I am happy with the assert on main only accessing the LCL state.

PR #4650, opened 2025-02-19
Changed random eviction cache to accounts only

SirTyson:
# Description

This change adds a configurable memory based cap to the LiveBucketList random eviction cache. It also modifies the cache such that only accounts are stored.

Following initial testing, the random eviction cache improved TX flooding performance, but had virtually no effect on apply time. This makes sense, since apply time already leverages the ltx cache and only makes at most one disk read per ledger entry. TX flooding on the other hand has no such queue and constantly re reads entries off disk. Given that we already have an ltx cache for apply time, it makes sense to only cache the entry types that are used by transaction flooding (i.e. `ACCOUNT`).

My initial default setting is to cache 3 GB. Currently accounts take up about 2.5 GB in the BucketList, so this allows us to cache all accounts with some room for growth, and 3 GB seems like a modest increase in RAM requirements for validators.

# Checklist
- [x] Compiles
- [x] Ran all tests

### DIFF
--- a/src/bucket/LiveBucketIndex.cpp
+++ b/src/bucket/LiveBucketIndex.cpp
@@ -5,6 +5,7 @@
 #include "bucket/LiveBucketIndex.h"
 #include "bucket/BucketIndexUtils.h"
 #include "bucket/BucketManager.h"
+#include "bucket/BucketUtils.h"
 #include "bucket/DiskIndex.h"
 #include "util/Fs.h"
 #include "util/GlobalChecks.h"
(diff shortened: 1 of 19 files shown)


src/bucket/LiveBucketIndex.cpp
```diff
@@ -95,6 +81,44 @@ LiveBucketIndex::LiveBucketIndex(BucketManager const& bm, Archive& ar,
...
+    if (bucketListTotalAccounts < maxBucketListAccountsToCache)
+    {
+        // We can cache the entire bucket
+        mCache = std::make_unique<CacheT>(bucketListTotalAccounts);
+    }
+    else
+    {
+        double percentAccountsInBucket =
+            static_cast<double>(accountsInThisBucket) / bucketListTotalAccounts;
+        auto cacheSize = static_cast<size_t>(bucketListTotalAccounts *
```
dmkozh, line 116: Currently `cacheSize == accountsInThisBucket`. You probably meant to use `maxBucketListAccountsToCache`.
Maybe it's worth adding a meter for this and writing a simple smoke test that ensures that the cache size is as expected given the BL size and cache size setting.

SirTyson: Good catch, I've added robust unit tests now.

src/bucket/LiveBucketList.cpp
```diff
@@ -31,7 +34,7 @@ LiveBucketList::sumBucketEntryCounters() const
     {
         for (auto const& b : {lev.getCurr(), lev.getSnap()})
         {
-            if (b->isIndexed())
+            if (!b->isEmpty())
```
dmkozh, line 37: I don't quite understand this change. Where are non-indexed, non-empty buckets coming from? Are we interested in these?

SirTyson: It doesn't really matter, as we have the invariant that every BucketList referenced bucket is either indexed or empty. I made this change because outside of bucket creation and GC, no class should know/care whether or not a bucket is indexed. Ideally, `isIndexed` would be private, but we have to call it in BucketManager for gc and the output iterator for initial creation. Generally speaking I think it's better for most classes to reason about a bucket being empty or non-empty instead of reasoning about whether or not it is indexed, but it's a minor point.

dmkozh (approved):

PR #4651, opened 2025-02-19
Fix race in DownloadBucketsWork

SirTyson:
# Description

Fixes a potential race in `DownloadBucketsWork`, where the parent `Work` could modify a map while the child `Work` on a potentially different thread held an iterator to that map.

# Checklist
- [x] Compiles
- [x] Ran all tests

### DIFF
--- a/src/historywork/DownloadBucketsWork.cpp
+++ b/src/historywork/DownloadBucketsWork.cpp
@@ -12,6 +12,7 @@
 #include "work/WorkWithCallback.h"
 #include <Tracy.hpp>
 #include <fmt/format.h>
+#include <mutex>
 
 namespace stellar
 {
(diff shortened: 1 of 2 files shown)


dmkozh (approved):

src/historywork/DownloadBucketsWork.cpp
```diff
@@ -92,7 +93,9 @@ DownloadBucketsWork::yieldMoreWork()
         std::static_pointer_cast<DownloadBucketsWork>(shared_from_this()));
 
     auto currId = mIndexId++;
+    mIndexMapMutex.lock();
     auto [indexIter, inserted] = mIndexMap.emplace(currId, nullptr);
+    mIndexMapMutex.unlock();
```
greg7mdp, line 98: In general, I think it is a better practice to use `std::lock_guard` rather than `lock` / `unlock` pairs, as the former guarantees that the mutex is unlocked even if there is an exception thrown in between.

src/historywork/DownloadBucketsWork.cpp
```diff
@@ -102,17 +105,20 @@ DownloadBucketsWork::yieldMoreWork()
             // To avoid dangling references, maintain a map of index pointers
             // and do a lookup inside the callback instead of capturing anything
             // by reference.
+            self->mIndexMapMutex.lock();
             auto indexIter = self->mIndexMap.find(currId);
             releaseAssertOrThrow(indexIter != self->mIndexMap.end());
             releaseAssertOrThrow(indexIter->second);
```
greg7mdp, line 111: If one of these throws, the mutex will not be unlocked.

PR #4652, opened 2025-02-21
re-test sorobans against their observations with core-chosen rust compiler

graydon:
# Description

This just adds an entry to `make check` to re-run the tests of each soroban module using the toolchain that stellar-core is building soroban with. This will catch early any divergence from observation caused by differences between the rust toolchain soroban tested itself against in CI vs. the one core is being built with. It's a small window of difference, but possible. 

Resolves https://github.com/stellar/rs-soroban-env/issues/1518


dmkozh (approved):

anupsdf:
How long does it take to run this new test today? (thinking of the impact when we have a lot of old Sorobans)

graydon:
> How long does it take to run this new test today? (thinking of the impact when we have a lot of old Sorobans)

Takes about 1 minute per soroban. I'm not worried.

Issue #4586, opened 2024-12-18
drop LOCAL_BUCKETS_ONLY Catchup mode


marta-lokhova:
Following removal of in-memory mode, we can remove LOCAL_BUCKETS_ONLY catchup mode which was needed to rebuild state. Related functions in LedgerManager should be dropped too.

PR #4657, opened 2025-02-28
Cleanups and fixes to maintenance

marta-lokhova:
Only the last commit has functional changes, which are fixes to maintenance introduced by #4543. Changes are as follows, by commit:
- Resolves https://github.com/stellar/stellar-core/issues/4586
- Remove an old history mode which became obsolete after `in-memory` got removed
- Cleanup various interfaces, there were several unused variables
- Two main changes are: (1) restore maintenance of SCP history, which got removed by accident, and (2) run ledger header maintenance on the ledger close thread to avoid potential serialization conflicts during ledger close.


src/catchup/CatchupConfiguration.cpp
```diff
@@ -13,16 +13,6 @@ namespace stellar
 void
 CatchupConfiguration::checkInvariants() const
```
SirTyson, line 14: Remove `checkInvariants`

src/main/Config.h
```diff
@@ -492,9 +492,6 @@ class Config : public std::enable_shared_from_this<Config>
     // fees, and scp history in the database
     bool MODE_STORES_HISTORY_MISC;
 
-    // A config parameter that stores ledger headers in the database
-    bool MODE_STORES_HISTORY_LEDGERHEADERS;
```
SirTyson, line 496: We need to check with platform to make sure they don't do anything with this flag, otherwise we need to wait for the next major release.

src/main/Config.h
```diff
@@ -805,8 +802,7 @@ class Config : public std::enable_shared_from_this<Config>
 
     bool modeDoesCatchupWithBucketList() const;
     bool isPersistingBucketListDBIndexes() const;
-    bool modeStoresAllHistory() const;
-    bool modeStoresAnyHistory() const;
+    bool modeStoresHistory() const;
```
SirTyson, line 805: Why not just use the config option directly?

SirTyson (dismissed):
LGTM

src/catchup/CatchupConfiguration.cpp
marta-lokhova: Fixed! (and your other comment as well)

SirTyson (approved):

SirTyson (approved):

PR #4660, opened 2025-03-04
Remove support for old V1 survey

### LINKED ISSUE #4332: Remove V1 survey XDR and supporting code
We have deprecated the V1 network survey with the addition of the V2 survey in #4275. Once the network minimum overlay version is 34, we can be sure that all nodes support the V2 survey and we should remove the XDR definitions and supporting code for the V1 survey.

bboston7:
Closes #4332.

Now that all stellar-core nodes on the network support the V2 survey, and there are no node stellar-core nodes that will generate V1 survey messages, it's safe to cleanup the codebase a bit and remove support for the old V1 survey entirely. This has the benefit of greatly simplifying `SurveyManager`.

# Checklist
- [x] Compiles
- [x] Ran all tests


src/protocol-curr/xdr
bboston7, line 1: NOTE to self: This is currently pointing to a downstream branch of the XDR repo. I'll need to update this once the relevant XDR PR merges.

src/protocol-next/xdr
bboston7, line 1: NOTE to self: This is currently pointing to a downstream branch of the XDR repo. I'll need to update this once the relevant XDR PR merges.

src/overlay/test/SurveyMessageLimiterTests.cpp
```diff
@@ -36,12 +36,12 @@ TEST_CASE("messagelimiter", "[overlay][survey][messagelimiter]")
     auto ledgerNum = app->getHerder().trackingConsensusLedgerIndex();
     SurveyRequestMessage firstRequest(v0SecretKey.getPublicKey(),
                                       v1SecretKey.getPublicKey(), ledgerNum,
-                                      temp, SURVEY_TOPOLOGY);
+                                      temp, TIME_SLICED_SURVEY_TOPOLOGY);
```
marta-lokhova, line 39: Shouldn't this fail in validation path given that the test doesn't setup the recording phase? Or do we treat the request as valid, but produce an empty response?

src/overlay/SurveyMessageLimiter.cpp
```diff
@@ -24,6 +24,12 @@ SurveyMessageLimiter::addAndValidateRequest(
     SurveyRequestMessage const& request,
     std::function<bool()> onSuccessValidation)
 {
+    if (request.commandType != TIME_SLICED_SURVEY_TOPOLOGY)
```
marta-lokhova, line 27: Will this if ever be true? Given that we're dropping support at the XDR level, the decoder should always reject any unsupported types. Perhaps this should be a sanity check assert instead?

src/main/CommandHandler.cpp
```diff
@@ -1051,51 +1048,6 @@ CommandHandler::checkBooted() const
...
-    surveyManager.addNodeToRunningSurveyBacklog(
-        SurveyMessageCommandType::SURVEY_TOPOLOGY, duration, id, std::nullopt,
-        std::nullopt);
-    retStr = "Adding node.";
-
-    retStr += success ? "Survey started " : "Survey already running!";
-}
-
-void
-CommandHandler::stopSurvey(std::string const&, std::string& retStr)
```
marta-lokhova, line 1088: We use `stopSurvey` in the survey script for retry purposes, so we should keep this endpoint (and probably remove the deprecation warning. You can rename the command if you'd like, but I think we should keep the ability to stop the reporting phase)
https://github.com/stellar/stellar-core/blob/69cb5cb2e8cb5a6d8c09bf4df7c3ef8505331357/scripts/OverlaySurvey.py#L408

src/overlay/SurveyManager.h
```diff
@@ -86,24 +77,27 @@ class SurveyManager : public std::enable_shared_from_this<SurveyManager>,
...
+    void stopSurveyReporting();
     // topology specific methods
     void sendTopologyRequest(NodeID const& nodeToSurvey);
-    void processOldStyleTopologyResponse(NodeID const& surveyedPeerID,
-                                         SurveyResponseBody const& body);
-    void
-    processOldStyleTopologyRequest(SurveyRequestMessage const& request) const;
     void processTimeSlicedTopologyResponse(NodeID const& surveyedPeerID,
-                                           SurveyResponseBody const& body);
+                                           TopologyResponseBodyV2 const& body);
```
marta-lokhova, line 94: nit: might be nicer to keep SurveyResponseBody argument to make extensions easier in the future

src/overlay/test/SurveyManagerTests.cpp
```diff
@@ -438,8 +434,21 @@ TEST_CASE("survey request process order", "[overlay][survey][topology]")
             false);
     };
 
+    // Get survey into reporting mode
+    auto startSurveyReportingFrom = [&](PublicKey const& node) {
```
marta-lokhova, line 438: Let's avoid duplication of logic with line 200, and extract it into a static function.

src/overlay/SurveyManager.cpp
```diff
@@ -863,78 +613,8 @@ SurveyManager::processTimeSlicedTopologyRequest(
...
-
-void
-SurveyManager::recordResults(Json::Value& jsonResultList,
-                             PeerStatList const& peerList) const
-{
-    for (auto const& peer : peerList)
-    {
-        jsonResultList.append(peerStatsToJson(peer));
-    }
+        std::make_shared<StellarMessage const>(msg));
```
marta-lokhova, line 617: Further opportunity for cleanup: minOverlayVersion can now be removed from broadcastMessage as well as Floodgate.

marta-lokhova (changes requested):
Looks good overall, just a few questions on survey API support and tests.

src/overlay/test/SurveyMessageLimiterTests.cpp
bboston7: That test is evaluating `SurveyMessageLimiter`, but [the check for an active survey occurs in `SurveyManager` as a lambda passed in to `SurveyMessageLimiter`](https://github.com/bboston7/stellar-core/blob/66bcfa7cd3a1531e250ed535d30d0336c21f8eb7/src/overlay/SurveyManager.cpp#L318). This helps to keep `SurveyMessageLimiter` small, while `SurveyManager` does the heavier lifting of communicating with other subsystems.

There are tests in `SurveyManagerTests` that check that validation fails for requests outside of the reporting phase.

src/overlay/SurveyMessageLimiter.cpp
bboston7: Good point. I changed this to an assert. I was thinking someone could craft a message with `commandType` set to something other than `TIME_SLICED_SURVEY_TOPOLOGY`, but you're right that the XDR decoder should reject that as an invalid enum.

src/main/CommandHandler.cpp
bboston7: Ah, good catch! I removed it because it was deprecated, not realizing that we were still using it. Turns out I added that functionality to the script *after* deprecating the endpoint. I should have un-deprecated it then. Whoops. I added the endpoint back and removed the deprecation warning. I agree it's important to keep.

src/overlay/SurveyManager.h
bboston7: Good point. Done

src/overlay/test/SurveyManagerTests.cpp
bboston7: Done

src/overlay/SurveyManager.cpp
bboston7: Done

marta-lokhova:
LGTM! I think you need to rebase on top of master to pickup cargo-deny fix, and refresh XDR pointers after the prerequisite PRs are merged.

marta-lokhova (approved):

PR #4661, opened 2025-03-05
Reduce BucketListDB configs

SirTyson:
# Description

Reduces `BUCKETLIST_DB_INDEX_CUTOFF` to 50 and disables account caching by default.

During testing, I determined `BUCKETLIST_DB_INDEX_CUTOFF` to be a reasonable value wrt memory consumption. Unfortunately I was actually setting the default to 250 despite the comments and docs saying 100. While I think 100 is still fine, I've turned it down a bit more to be on the safe side and unblock the release.

# Checklist
- [x] Compiles
- [x] Ran all tests

### DIFF
--- a/docs/stellar-core_example.cfg
+++ b/docs/stellar-core_example.cfg
@@ -235,20 +235,20 @@ MAX_DEX_TX_OPERATIONS_IN_TX_SET = 0
 # 0, indiviudal index is always used. Default page size 16 kb.
 BUCKETLIST_DB_INDEX_PAGE_SIZE_EXPONENT = 14
 
-# BUCKETLIST_DB_MEMORY_FOR_CACHING (Integer) default 3000
+# BUCKETLIST_DB_MEMORY_FOR_CACHING (Integer) default 0
 # Memory used for caching entries by BucketListDB when Bucket size is
 # larger than BUCKETLIST_DB_INDEX_CUTOFF, in MB. Note that this value does
 # not impact Buckets smaller than BUCKETLIST_DB_INDEX_CUTOFF, as they are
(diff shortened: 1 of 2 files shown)


docs/stellar-core_example.cfg
```diff
@@ -235,20 +235,20 @@ MAX_DEX_TX_OPERATIONS_IN_TX_SET = 0
...
 
-# BUCKETLIST_DB_INDEX_CUTOFF (Integer) default 100
+# BUCKETLIST_DB_INDEX_CUTOFF (Integer) default 50
 # Size, in MB, determining whether a bucket should have an individual
 # key index or a key range index. If bucket size is below this value, range
 # based index will be used. If set to 0, all buckets are range indexed. If
 # BUCKETLIST_DB_INDEX_PAGE_SIZE_EXPONENT == 0, value ignored and all
 # buckets have individual key index.
-BUCKETLIST_DB_INDEX_CUTOFF = 100
+BUCKETLIST_DB_INDEX_CUTOFF = 50
```
marta-lokhova, line 251: How can we reason about these values? I realize I don't quite understand the implications of going from 100 MB to 50 MB, or the other way around, from 100 to 250MB. I know it depends heavily on the data, but can we estimate worst case scenario, so we at least know what the upper bound is?

SirTyson: Worst case scenario is  ~`3 * BUCKETLIST_DB_INDEX_CUTOFF * numLevelsInBucketList`, where `numLevelsInBucketList == 11`. This gives us about 1.65 GB with the 50 MB setting, or 3.3 GB with the 100 MB setting.

marta-lokhova (approved):

PR #4662, opened 2025-03-07
fix example config

marta-lokhova:


anupsdf (approved):

marta-lokhova:
this change will be included in a bigger docs overhaul

PR #4666, opened 2025-03-13
Adapt Core to the CAP-67 XDR.

dmkozh:
# Description

Adapt Core to the CAP-67 XDR.

This is unfortunately rather hacky in order to allow for these changes to go in a minor release. Specifically, we're only bumping XDR for the Core itself, and not for the Rust env. That's why we need to temporarily disable a few XDR integrity checks. There is also a point of interaction of the new and old XDR(ScAddress changes), and I've added some tests to ensure that Core won't be able to accidentally accept XDR that's not supported by the current protocol.

# Checklist
- [ ] Compiles
- [ ] Ran all tests


src/main/main.cpp
```diff
@@ -186,6 +186,18 @@ checkXDRFileIdentity()
...
+        // are mostly binary compatible with the Rust XDR, and the incompatible
+        // part (ScAddress change) is actually supposed to be not decodable on
+        // the Rust side.
+        if (cpp.first.filename() == "Stellar-types.x" ||
+            cpp.first.filename() == "Stellar-contract.x" ||
+            cpp.first.filename() == "Stellar-ledger-entries.x" ||
+            cpp.first.filename() == "Stellar-ledger.x")
+        {
+            continue;
+        }
```
jayz22, line 200: instead of skipping, would it be better to statically assert the hashes of these files, on both sides?

PR #4670, opened 2025-03-21
Change expensive timer metrics to trend counters

SirTyson:
# Description

Resolves https://github.com/stellar/stellar-core-internal/issues/336

Changes expensive timer metrics to counters.

For timer and histogram metrics, we store data points in a buffer. When this buffer "fills up" we must to a compaction step, which involves many copies. For metrics that measure short lived functions (runtime in the ns or us range) that get called very frequently, the cost of this compaction step dominates the function runtime. For example, for BucketList in memory loads, metric overhead accounted for 75% of the load's total runtime.

This changes several short lived metrics to trends instead of timer based histograms. Specifically, we no longer track individual data samples, but keep a running sum of runtime instead. This still gives us insight into trends, but no longer gives us info on individual calls. However, it is significantly cheaper runtime wise.

# Checklist
- [x] Compiles
- [x] Ran all tests

### DIFF
--- a/docs/metrics.md
+++ b/docs/metrics.md
@@ -47,7 +47,8 @@ bucketlistDB-live.bulk.poolshareTrustlines  | timer     | time to load poolshare
 bucketlistDB-live.bulk.prefetch             | timer     | time to prefetch
 bucketlistDB-live.bulk.eviction           | timer     | time to load for eviction scan
 bucketlistDB-live.bulk.query              | timer     | time to load for query server
-bucketlistDB-<X>.point.<Y>                | timer     | time to load single entry of type <Y> on BucketList <X> (live/hotArchive)
+bucketlistDB-<X>.<Y>.sum                  | counter   | sum of time (microseconds) to load single entry of type <Y> on BucketList <X> (live/hotArchive)
+bucketlistDB-<X>.<Y>.count                | counter   | number of times single entry of type <Y> on BucketList <X> (live/hotArchive) is loaded
 bucketlistDB-cache.hit                    | meter     | number of cache hits on Live BucketList Disk random eviction cache
 bucketlistDB-cache.miss                   | meter     | number of cache misses on Live BucketList Disk random eviction cache
 crypto.verify.hit                         | meter     | number of signature cache hits
@@ -58,8 +59,10 @@ herder.pending[-soroban]-txs.age1         | counter   | number of gen1 pending t
 herder.pending[-soroban]-txs.age2         | counter   | number of gen2 pending transactions
 herder.pending[-soroban]-txs.age3         | counter   | number of gen3 pending transactions
 herder.pending[-soroban]-txs.banned       | counter   | number of transactions that got banned
-herder.pending[-soroban]-txs.delay        | timer     | time for transactions to be included in a ledger
-herder.pending[-soroban]-txs.self-delay   | timer     | time for transactions submitted from this node to be included in a ledger
+herder.pending[-soroban]-txs.sum          | counter   | sum of time (milliseconds) for transactions to be included in a ledger
+herder.pending[-soroban]-txs.count        | counter   | number of transactions to be included in a ledger
+herder.pending[-soroban]-txs.self-sum     | counter   | sum of time (milliseconds) for transactions submitted from this node to be included in a ledger
+herder.pending[-soroban]-txs.self-count   | counter   | number of transactions submitted from this node to be included in a ledger
 history.check.failure                     | meter     | history archive status checks failed
 history.check.success                     | meter     | history archive status checks succeeded
 history.publish.failure                   | meter     | published failed
@@ -126,7 +129,6 @@ overlay.flood.peer-tx-pull-latency        | timer     | time to pull a transacti
 overlay.demand.timeout                    | meter     | pull mode timeouts
 overlay.flood.relevant-txs                | meter     | relevant transactions pulled from peers
 overlay.flood.irrelevant-txs              | meter     | irrelevant transactions pulled from peers
-overlay.flood.advert-delay                | timer     | time each advert sits in the inbound queue
 overlay.flood.abandoned-demands           | meter     | tx hash pull demands that no peers responded
 overlay.flood.broadcast                   | meter     | message sent as broadcast per peer
 overlay.flood.duplicate_recv              | meter     | number of bytes of flooded messages that have already been received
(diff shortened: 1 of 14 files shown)


docs/metrics.md
```diff
@@ -57,8 +58,10 @@ herder.pending[-soroban]-txs.age1         | counter   | number of gen1 pending t
 herder.pending[-soroban]-txs.age2         | counter   | number of gen2 pending transactions
 herder.pending[-soroban]-txs.age3         | counter   | number of gen3 pending transactions
 herder.pending[-soroban]-txs.banned       | counter   | number of transactions that got banned
-herder.pending[-soroban]-txs.delay        | timer     | time for transactions to be included in a ledger
-herder.pending[-soroban]-txs.self-delay   | timer     | time for transactions submitted from this node to be included in a ledger
+herder.pending[-soroban]-txs.delay-time   | counter     | sum of time (microseconds) for transactions to be included in a ledger
```
marta-lokhova, line 61: `delay-time` --> `sum`
`delay-count` --> `count`

docs/metrics.md
```diff
@@ -46,7 +46,8 @@ bucketlistDB-live.bulk.poolshareTrustlines  | timer     | time to load poolshare
 bucketlistDB-live.bulk.prefetch             | timer     | time to prefetch
 bucketlistDB-live.bulk.eviction           | timer     | time to load for eviction scan
 bucketlistDB-live.bulk.query              | timer     | time to load for query server
-bucketlistDB-<X>.point.<Y>                | timer     | time to load single entry of type <Y> on BucketList <X> (live/hotArchive)
+bucketlistDB-<X>.point-time.<Y>           | counter   | sum of time (microseconds) to load single entry of type <Y> on BucketList <X> (live/hotArchive)
```
marta-lokhova, line 49: what do you think about `bucketlistDB-<X>.point-time.<Y>` --> `bucketlistDB-<X>.<Y>.sum`? (and similar with count)

marta-lokhova (changes requested):

src/herder/TransactionQueue.cpp
```diff
@@ -1075,9 +1084,14 @@ SorobanTransactionQueue::SorobanTransactionQueue(Application& app,
         sizeByAge,
         app.getMetrics().NewCounter(
             {"herder", "pending-soroban-txs", "banned"}),
-        app.getMetrics().NewTimer({"herder", "pending-soroban-txs", "delay"}),
-        app.getMetrics().NewTimer(
-            {"herder", "pending-soroban-txs", "self-delay"}));
+        app.getMetrics().NewCounter(
```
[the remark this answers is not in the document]
SirTyson: I'll open an issue for this. I think it might be easier to merge this first, then do the grafana changes. That way we actually have some of the new metrics to look at on a dev box to make sure the changes are right.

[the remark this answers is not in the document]
marta-lokhova: that makes sense, yeah

src/herder/TransactionQueue.cpp
```diff
@@ -111,8 +111,10 @@ ClassicTransactionQueue::ClassicTransactionQueue(Application& app,
     mQueueMetrics = std::make_unique<QueueMetrics>(
         sizeByAge,
         app.getMetrics().NewCounter({"herder", "pending-txs", "banned"}),
-        app.getMetrics().NewTimer({"herder", "pending-txs", "delay"}),
-        app.getMetrics().NewTimer({"herder", "pending-txs", "self-delay"}));
+        app.getMetrics().NewCounter({"herder", "pending-txs", "sum"}),
```
marta-lokhova, line 114: forgot to rename delay-counter --> count

src/herder/TransactionQueue.cpp
```diff
@@ -771,12 +773,17 @@ TransactionQueue::removeApplied(Transactions const& appliedTxs)
                     if (transaction->mTx->getFullHash() ==
                         appliedTx->getFullHash())
                     {
-                        auto elapsed = now - transaction->mInsertionTime;
-                        mQueueMetrics->mTransactionsDelay.Update(elapsed);
+                        auto elapsed = std::chrono::duration_cast<
```
marta-lokhova, line 776: tx queue reasons about txs on like, seconds scale. I think milliseconds is more appropriate here

marta-lokhova (changes requested):

docs/metrics.md
```diff
@@ -46,7 +46,8 @@ bucketlistDB-live.bulk.poolshareTrustlines  | timer     | time to load poolshare
 bucketlistDB-live.bulk.prefetch             | timer     | time to prefetch
 bucketlistDB-live.bulk.eviction           | timer     | time to load for eviction scan
 bucketlistDB-live.bulk.query              | timer     | time to load for query server
-bucketlistDB-<X>.point.<Y>                | timer     | time to load single entry of type <Y> on BucketList <X> (live/hotArchive)
+bucketlistDB-<X>.<Y>.sum                  | counter   | sum of time (microseconds) to load single entry of type <Y> on BucketList <X> (live/hotArchive)
```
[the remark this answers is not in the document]
SirTyson: I think historical data is probably correct. Until recently, all these metrics were in the ms range since we didn't have the in-memory bucket optimization. Even after the in-memory change, calls are still in the ms range until this pr gets merged.

Also, it doesn't look like we'd lose too much resolution here regardless. Medida tracks all raw data in ns regardless of your duration unit. It only uses the duration_unit value to return the correct unit when reporting. However, it returns a double, so even if your duration_unit is in ms but your query is actually in the us range, medida would report a fraction of a ms via a double, so I don't think we lose any accuracy (other than maybe some float rounding).

src/herder/TransactionQueue.cpp
SirTyson: Ah sorry, fixed

src/herder/TransactionQueue.cpp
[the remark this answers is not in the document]
marta-lokhova: Note that the simulation dashboard need to be updated as well

src/herder/TransactionQueue.h
```diff
@@ -181,18 +181,29 @@ class TransactionQueue
...
+            , mTransactionsDelayCounter(transactionsDelayCounter)
+            , mTransactionsSelfDelayCounter(transactionsSelfDelayCounter)
         {
         }
         std::vector<medida::Counter*> mSizeByAge;
         medida::Counter& mBannedTransactionsCounter;
-        medida::Timer& mTransactionsDelay;
-        medida::Timer& mTransactionsSelfDelay;
+
+        // Sum of total delay, in microseconds
```
marta-lokhova, line 200: milliseconds

marta-lokhova (approved):
LGTM, just one small typo. Feel free to squash the last commit and rebase.

SirTyson:
> LGTM, just one small typo. Feel free to squash the last commit and rebase.

Done

marta-lokhova (approved):

PR #4671, opened 2025-03-22
Load generator: add a new mode to pre-generate load offline

### LINKED ISSUE #2833: load generator may perform too much work on the main thread
Right now it looks like `loadgen` performs a lot (too much?) of heavy work on the main thread:
* it fetches account data for _all accounts_ from the database when waiting for the simulation to end using point loads instead of batching.
     * Maybe we could both limit the number of queries performed and remove accounts from that list when they reach the right sequence number (to avoid querying them all the time)?
* it fetches accounts from the database for the current "step" using "point loads" (ie, not using batch loading)
* it's generating transactions, in particular digital signatures, for each transaction for the current "step"

As loadgen is scheduled like anything else, I wonder if there is some weird scheduling interaction between the node doing "real work" and performing loadgen related work.

I don't know if it's easy to quantify that effect, or if we should just mitigate all of those issues somehow.

marta-lokhova:
Resolves stellar/stellar-core-internal#345 
Resolves https://github.com/stellar/stellar-core-internal/issues/339

Offline load generation has a few nice properties:
- It eliminates account creation phase in our simulations completely. We can now pre-fill the DB by configuring GENESIS_TEST_ACCOUNT_COUNT. On my machine, creating 1,000,000 accounts takes ~20 seconds. This should speed up our iteration speed when working with simulations.
- It eliminates the overhead from creating transactions in real-time, and allows for a more realistic performance measurement. Transactions are pre-generated into LOADGEN_PREGENERATED_TRANSACTIONS_FILE via command line.
- It removes mempool race conditions that are making the simulation fail incorrectly (because we can create arbitrarily many accounts very quickly now, we can wait a long time before re-using the same account)

In addition, the new loadgen skips certain expensive validity checks, and adds pre-generated transactions directly into the tx queue, this further reduces the overhead of a load generation step.


src/main/CommandLine.cpp
```diff
@@ -1870,6 +1871,53 @@ runApplyLoad(CommandLineArgs const& args)
         return 0;
     });
 }
+
+int
+runGenerateSyntheticLoad(CommandLineArgs const& args)
+{
+    std::string outputFile = "stellar-load-transactions.xdr";
+    uint32_t numTransactions = 1000;
```
anupsdf, line 1879: Should we do higher than 1000 default txns since the upper bound of binary search might be higher than 1000?

anupsdf: Spoke with Marta. This default is fine since it will be overridden.

src/herder/TransactionQueue.cpp
```diff
@@ -459,9 +460,15 @@ TransactionQueue::canAdd(
             mApp.getLedgerManager().getLastClosedLedgerNum() + 1;
     }
 
-    auto txResult =
-        tx->checkValid(mApp.getAppConnector(), ls, 0, 0,
-                       getUpperBoundCloseTimeOffset(mApp, closeTime));
+    // Loadgen txs were generated by this local node, and therefore can skip
+    // validation, and be added directly to the queue.
+    auto txResult = tx->createSuccessResult();
```
SirTyson, line 465: Can we refactor this logic behind an ifdef? It seems fairly dangerous to have logic to skip verification available in non-test builds. I think we can pipe the flag everywhere without the ifdef, but only use it inside the ifdef.

src/simulation/LoadGenerator.cpp
```diff
@@ -1743,4 +1813,32 @@ GeneratedLoadConfig::modeUploads() const
...
+LoadGenerator::readTransactionFromFile(GeneratedLoadConfig const& cfg)
+{
+    ZoneScoped;
+
+    // Read the next transaction from the file
+    TransactionEnvelope txEnv;
+    if (!mPreloadedTransactionsFile.readOne(txEnv))
+    {
+        // End of file reached, start again from the beginning
+        throw std::runtime_error("LoadGenerator: End of file reached");
```
SirTyson, line 1827: Hmm I don't think the comment is correct. Currently, if we throw it looks like loadgen will just fail because we don't catch this exception anywhere. Also, would this actually work? I think we'd have invalid sequence numbers if we started over.

src/simulation/LoadGenerator.cpp
```diff
@@ -896,8 +943,11 @@ LoadGenerator::submitTx(GeneratedLoadConfig const& cfg,
                       tx->getInclusionFee());
             return false;
         }
+
+        // No re-submission in PAY_PREGENERATED mode
         if (++numTries >= TX_SUBMIT_MAX_TRIES ||
-            status != TransactionQueue::AddResultCode::ADD_STATUS_ERROR)
+            status != TransactionQueue::AddResultCode::ADD_STATUS_ERROR ||
+            cfg.mode == LoadGenMode::PAY_PREGENERATED)
```
SirTyson, line 950: I think we need to guard the `maybeHandleFailedTx` and only call it if mode != PAY_PREGENERATED. It looks like that function makes unchecked dereferences to the `from` account pointer, which is always null for PAY_PREGENERATED.

src/simulation/LoadGenerator.cpp
```diff
@@ -1485,7 +1536,9 @@ LoadGenerator::execute(TransactionFrameBasePtr txf, LoadGenMode mode,
     auto msg = txf->toStellarMessage();
     txm.mTxnBytes.Mark(xdr::xdr_argpack_size(*msg));
 
-    auto addResult = mApp.getHerder().recvTransaction(txf, true);
+    // Skip certain checks for pregenerated transactions
+    bool isLoadgenTx = (mode == LoadGenMode::PAY_PREGENERATED);
```
SirTyson, line 1540: Nit: I think the `isLoadgenTx` variable name is confusing here, since all TXs are coming from loadgen. Maybe rename to `isPregeneratedTx`?

src/simulation/LoadGenerator.cpp
```diff
@@ -1743,4 +1813,32 @@ GeneratedLoadConfig::modeUploads() const
...
+
+    // Increment the sequence number of the source account
+    auto idx = mCurrPreloadedTransaction % cfg.nAccounts;
+    auto acc = mTxGenerator.getAccount(idx + cfg.offset);
+    releaseAssert(acc);
+    acc->setSequenceNumber(txFrame->getSeqNum());
+    ++mCurrPreloadedTransaction;
+
+    // Do not provide an account
+    return std::make_pair(nullptr, txFrame);
```
SirTyson, line 1842: Is there a specific reason we don't just return acc here? It looks like consumers of this pair assume that the `from` account is non-null, and we're just lucky we don't happen to hit those paths.

src/simulation/LoadGenerator.cpp
```diff
@@ -791,6 +831,13 @@ LoadGenerator::generateLoad(GeneratedLoadConfig cfg)
                         ledgerNum, sourceAccountId, cfg);
                 };
                 break;
+            case LoadGenMode::PAY_PREGENERATED:
+                generateTx = [&]() {
+                    // Each transaction is for a unique source account, so we
```
SirTyson, line 836: Helpful comment, but I think it would be more relevant in `submitTx`

src/simulation/LoadGenerator.cpp
```diff
@@ -412,6 +424,26 @@ LoadGenerator::start(GeneratedLoadConfig& cfg)
     mPreLoadgenApplySorobanFailure =
         mTxGenerator.getApplySorobanFailure().count();
 
+    if (cfg.mode == LoadGenMode::PAY_PREGENERATED)
```
SirTyson, line 427: Do we need to assert/fail if `GENESIS_TEST_ACCOUNT_COUNT` is 0 here? Similarly, should we fail is `GENESIS_TEST_ACCOUNT_COUNT` is set but someone attempts to invoke create accounts load?

src/test/TestUtils.h
```diff
@@ -133,4 +133,8 @@ bool appProtocolVersionStartsFrom(Application& app,
 
 // Large enough fee to cover most of the Soroban transactions.
 constexpr uint32_t DEFAULT_TEST_RESOURCE_FEE = 1'000'000;
+
+void generateTransactions(Application& app, std::string const& outputFile,
```
SirTyson, line 137: Nit: std::filesystem::path preferred over string

src/simulation/test/LoadGeneratorTests.cpp
```diff
@@ -70,6 +75,12 @@ TEST_CASE("generate load with unique accounts", "[loadgen]")
     auto nodes = simulation->getNodes();
     auto& app = *nodes[0]; // pick a node to generate load
 
+    std::string fileName = "stellar-load-transactions.xdr";
+    auto cleanup = gsl::finally([&]() { std::remove(fileName.c_str()); });
+
+    generateTransactions(app, fileName, nTxs, nAccounts,
+                         /* offset */ nAccounts);
```
SirTyson, line 82: Nit: use `LOADGEN_PREGENERATED_TRANSACTIONS_FILE` here instead of hard code, you can just copy the path by value in the cleanup lambda.

src/simulation/test/LoadGeneratorTests.cpp
```diff
@@ -53,12 +54,16 @@ TEST_CASE("generate load in protocol 1")
 TEST_CASE("generate load with unique accounts", "[loadgen]")
 {
     Hash networkID = sha256(getTestConfig().NETWORK_PASSPHRASE);
+    uint32_t const nAccounts = 1000;
```
SirTyson, line 57: I think this tests looks good, but it kinda canibalizes our only explicit create accounts load test. Should we break this off into a separate test to preserve the legacy account generation until it's deprecated (which I assume is the plan)?

SirTyson (dismissed):
Overall a very helpful change! A few comments from me, mostly minor stuff.

src/simulation/LoadGenerator.cpp
marta-lokhova: This `if` case already handles that by failing before we can get to `maybeHandleFailedTx`

src/simulation/LoadGenerator.cpp
marta-lokhova: Good catch, the comment is wrong (leftover from the first iteration of this change). Yeah, we should not be re-using the same file. The idea is that we generate enough transactions at the beginning of the mission. Removed the comment.

src/simulation/LoadGenerator.cpp
marta-lokhova: The account is only needed for submission retries. In pre-generated mode, there's no retries, hence the account is not used. I didn't want to return the account here to avoid situations where we accidentally use incorrect cached info. Instead, we should fully rely on account information from the generated file (so if there's a bug somewhere, I'd prefer to crash).

src/simulation/TxGenerator.h
```diff
@@ -157,6 +157,9 @@ class TxGenerator
 
     void reset();
 
+    TestAccountPtr getAccount(uint64_t accountId);
```
bboston7, line 160: Nitpick: This function can be `const`

bboston7 (approved):

src/simulation/LoadGenerator.cpp
marta-lokhova: Yeah, extra checks sound good, added. In the change I left the door open for flexible uses like old-style account creation + pre-generated txs, or new-style account creation + old-style account creation with an offset. But those are weird/niche use cases I don't think we need to support anyway.

src/simulation/test/LoadGeneratorTests.cpp
SirTyson: Ya that's probably good enough, especially since we want to get rid of the old account generation soon anyway.

src/herder/TransactionQueue.cpp
marta-lokhova: Agreed, done

src/main/Config.cpp
```diff
@@ -1170,6 +1174,10 @@ Config::processConfig(std::shared_ptr<cpptoml::table> t)
                  [&]() {
                      ARTIFICIALLY_GENERATE_LOAD_FOR_TESTING = readBool(item);
                  }},
+                {"LOADGEN_PREGENERATED_TRANSACTIONS_FILE",
```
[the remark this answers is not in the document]
marta-lokhova: Fixed!

src/simulation/TxGenerator.h
marta-lokhova: Done.

SirTyson (dismissed):
LGTM

SirTyson (approved):

PR #4675, opened 2025-03-26
Use raii lock in DownloadBucketsWork

### LINKED ISSUE #4674: Use RAII-style mechanism for owning a mutex in a scoped block
### What:
Thanks @greg7mdp for the [recommendation here](https://github.com/stellar/stellar-core/pull/4651/files/b71c80738bba644d3376c06b6cc911eb51abe804#r2011131932)

[DownloadBucketsWork::yieldMoreWork()](https://github.com/SirTyson/stellar-core/blob/b71c80738bba644d3376c06b6cc911eb51abe804/src/historywork/DownloadBucketsWork.cpp#L108) could use a RAII-style mechanism for owning a mutex in a scoped block.  We could either start a new scope using curly brackets or a `if` condition for starting a new scope like this,
```
if(std::lock_guard<std::mutex> lock(mIndexMapMutex); true)
{
   ....
}

```

We should also scan the codebase for other places where we could do this improvement.

EDIT: Scoped block is better than the if condition.

SirTyson:
# Description

Resolves #4674

Note that we are still using lock/unlock at `DownloadBucketsWork.cpp:97`, given that the type definition is incredibly verbose to forward declare, and it's unlikely std::map::emplace will throw. Even if it does, it doesn't matter that we hold the lock since core and the `DownloadBucketsWork` will crash anyway.

# Checklist
- [x] Compiles
- [x] Ran all tests

### DIFF
--- a/src/historywork/DownloadBucketsWork.cpp
+++ b/src/historywork/DownloadBucketsWork.cpp
@@ -105,13 +105,15 @@ DownloadBucketsWork::yieldMoreWork()
             // To avoid dangling references, maintain a map of index pointers
             // and do a lookup inside the callback instead of capturing anything
             // by reference.
-            self->mIndexMapMutex.lock();
-            auto indexIter = self->mIndexMap.find(currId);
-            releaseAssertOrThrow(indexIter != self->mIndexMap.end());
-            releaseAssertOrThrow(indexIter->second);
-            auto index = std::move(indexIter->second);
-            self->mIndexMap.erase(indexIter);
-            self->mIndexMapMutex.unlock();
+            std::unique_ptr<LiveBucketIndex const> index{};
+            {
+                std::lock_guard<std::mutex> lock(self->mIndexMapMutex);
+                auto indexIter = self->mIndexMap.find(currId);
+                releaseAssertOrThrow(indexIter != self->mIndexMap.end());
+                releaseAssertOrThrow(indexIter->second);
+                index = std::move(indexIter->second);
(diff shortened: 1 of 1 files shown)


anupsdf (approved):
lgtm

greg7mdp:
> Note that we are still using lock/unlock at DownloadBucketsWork.cpp:97, given that the type definition is incredibly verbose to forward declare

Not sure what C++ version core is using, but you shouldn't have to provide the type to `std::lock_guard`.

SirTyson:
> > Note that we are still using lock/unlock at DownloadBucketsWork.cpp:97, given that the type definition is incredibly verbose to forward declare
> 
> Not sure what C++ version core is using, but you shouldn't have to provide the type to `std::lock_guard`.

I'm referring to the type returned by `emplace`, which must be assigned in the scope of the lock_guard but live outside the lock_guard scope.

anupsdf:
> Not sure what C++ version core is using

stellar-core is on C++ 17, https://github.com/stellar/stellar-core/blob/v22.2.0/configure.ac#L229

greg7mdp:
> I'm referring to the type returned by `emplace`, which must be assigned in the scope of the lock_guard but live outside the lock_guard scope.

Got it. You could use a lambda:

```c++
    auto [indexIter, inserted] = [&]() {
       std::lock_guard g(mIndexMapMutex);
       return mIndexMap.emplace(currId, nullptr);
    }();
```

PR #4680, opened 2025-03-31
Removed peer authenticated assert in Floodgate::broadcast

SirTyson:
# Description

Resolves https://github.com/stellar/stellar-core-internal/issues/335

This PR removes an expensive assert in `Floodgate::broadcast`. This assert had to acquire `mStateMutex` from the main thread, but would often be blocked by the background thread. During periods of high load (1850 TPS), waiting for this mutex would double the median time of broadcast. Because a peer's authenticated state can only be changed from the mainthread anyway, the assert was replaced with a lockless main thread assert. During testing at 1850 TPS, this change resulted in a 15% decrease of txset fetch time, as well as a 10% decrease in nomination latency.

# Checklist
- [x] Compiles
- [x] Ran all tests

### DIFF
--- a/src/overlay/Floodgate.cpp
+++ b/src/overlay/Floodgate.cpp
@@ -86,6 +86,7 @@ bool
 Floodgate::broadcast(std::shared_ptr<StellarMessage const> msg,
                      std::optional<Hash> const& hash)
 {
+    releaseAssert(threadIsMain());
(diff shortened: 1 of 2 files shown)


marta-lokhova (approved):
LGTM! We can merge this in the public repo

PR #4686, opened 2025-04-10
Add rust CKMS backend

graydon:
This is a sketch of a hookup between medida and a rust library that provides CKMS estimators. There are two reasons:

1. The rust code is much faster than the prometheus C++ client code we copied into medida previously
2. The rust code is more correct! Apparently the CKMS paper had two versions, one of which corrected a bunch of mistakes in the first, and the prometheus code is based on the first, where the rust is based on the second.

I don't claim to understand point 2 super well, but it is a claim made by the rust implementation. I can empirically validate point 1 -- it is faster! -- and that is the original motivation here, since CKMS metrics updates are expensive enough to risk lagging core if done accidentally in a hot loop.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -311,6 +311,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "quantiles"
+version = "0.7.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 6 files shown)


graydon:
closing -- we just made medida's CKMS faster https://github.com/stellar/stellar-core/pull/5338

